PR #128, opened 2020-04-24
Remove dd4hep warnings

asalzburger:
This is a cherry-pick plus add-ons and supersedes #118.

PR #127, opened 2020-04-24
Fix ITk navigation

asalzburger:
Cherry-picked version of #113 (see description there) as a start, in order to make everything working, I have had to:

- [x] change the `ProtoLayer` to use the `Extent` object of the `Polyhedron`
- [x] add UnitTests to Surfaces for the correct extent
- [x] introduce a method to find out if you are on a hyperplane (and thus radial distance needs to be checked)

Ad a consequence, I removed A LOT of code from the `LayerCreator`, which makes it way more readable, particularly the envelope handling is now done at one single space.

PR #126, opened 2020-04-24
Fix bug in CKF and add script for KF timing test

XiaocongAi:
- Fix bug in CKF: before adding the track state, the last tip should be removed from the active tips. But when creating multiple states on ONE surface, the update of active tips should be done only ONCE so that states have the same previous state. In the current master branch, each call to `addSourcelinkState` will remove the last tip, which is not correct.
- Fix a bug in Pythia8 option: include the `pileup` generators as well instead of including the `hard` configurations twice.
- Clean the KF smoothing
- Add script for KF timing test (To produce plot as below)
image

PR #1099, opened 2021-12-06
feat: global chi2 fitter / GX2F

r4lv:
Dear all,
please have a look at my implementation of the global χ² fitter. It consists of the _core_ `Chi2Fitter.hpp`, an Algorithm, an Example (in cpp...) and a simple Unit Test.

I'd mark this as _WIP_, as there is still a lot of debug output and comments which might need to be removed before this gets merged.


XiaocongAi:
Hi @r4lv , could you remind me if this is ready to be reviewed or not?

Core/include/Acts/TrackFitting/Chi2Fitter.hpp
```diff
@@ -0,0 +1,734 @@
...
+  size_t measurementHoles = 0;
+
+  // Counter for handled states
+  size_t processedStates = 0;
+
+  // Indicator if track fitting has been done
+  bool finished = false;
+  // we do not have smoothed, reset, reversed, like the KF does.
+
+  bool isFinalized = false;
```
XiaocongAi, line 176: Is this used somewhere? We can remove it if not used.

Core/include/Acts/TrackFitting/Chi2Fitter.hpp
```diff
@@ -0,0 +1,734 @@
...
+
+  // Measurement surfaces without hits
+  std::vector<const Surface*> missedActiveSurfaces;
+
+  // collectors
+  std::vector<ActsScalar>
+      m_measurements_collector;  // pixel layers add 2 entries, strip layers
```
XiaocongAi, line 183: I think these varaibles should be renamed following the naming conventions in Acts.

Core/include/Acts/TrackFitting/Chi2Fitter.hpp
```diff
@@ -0,0 +1,734 @@
...
+                   std::reference_wrapper<const SourceLink>>*
+        inputMeasurements = nullptr;
+
+    /// Whether to consider multiple scattering.
+    bool multipleScattering = false;  // TODO: add later
+
+    /// Whether to consider energy loss.
+    bool energyLoss = false;  // TODO: add later
+
+    // /// The measurement calibrator
```
XiaocongAi, line 243: We can just remove those commented out code if not used anywhere.

Core/include/Acts/TrackFitting/Chi2Fitter.hpp
```diff
@@ -0,0 +1,734 @@
...
+        trackStateProxy.predicted() = std::move(boundParams.parameters());
+        if (boundParams.covariance().has_value()) {
+          trackStateProxy.predictedCovariance() =
+              std::move(*boundParams.covariance());
+        }
+        trackStateProxy.jacobian() = std::move(jacobian);
+        trackStateProxy.pathLength() = std::move(pathLength);
+
+        result.jacobianFromStart = jacobian * result.jacobianFromStart;
+        // TODO: we need to update result.jacobianFromStart also when we have no
```
XiaocongAi, line 373: Is this already done?

Core/include/Acts/TrackFitting/Chi2Fitter.hpp
```diff
@@ -0,0 +1,734 @@
...
+        result.d1_deriv_sum_collector += deriv1;
+        result.d2_deriv_sum_collector += deriv2;
+
+        result.m_measurements_collector.push_back(parameters(0));
+        result.mcov_measurement_covariance_collector.push_back(
+            covariance(0, 0));
+        result.r_residual_collector.push_back(residuals(0));
+
+        // KF: filtered = predicted + K * (calibrated - H * predicted);
+        trackStateProxy.filtered() =
```
XiaocongAi, line 412: We don't have to set the filtered track parameters for the chi2fitter.

Core/include/Acts/TrackFitting/Chi2Fitter.hpp
```diff
@@ -0,0 +1,734 @@
...
+            trackStateProxy.predicted();  // in KF, this is set in
+                                          // GainMatrixUpdater.hpp:102, called
+                                          // from filter():
+        trackStateProxy.filteredCovariance() =
+            trackStateProxy.predictedCovariance();
+
+        if (parameters.rows() == 2) {
```
XiaocongAi, line 419: What if the dimension is not 2?

Core/include/Acts/TrackFitting/Chi2Fitter.hpp
```diff
@@ -0,0 +1,734 @@
...
+        result.measurementHoles = result.missedActiveSurfaces.size();  // TODO
+        // Since we encountered a measurment update the lastMeasurementIndex to
+        // the lastTrackIndex.
+        result.lastMeasurementIndex = result.lastTrackIndex;
+
+      } else if (surface->associatedDetectorElement() != nullptr ||
+                 surface->surfaceMaterial() != nullptr) {
+        // We only create track states here if there is already measurement
+        // detected or if the surface has material (no holes before the first
+        // measurement
```
XiaocongAi, line 460: We need to consider the material effects even if there is no measurement on the surface.

Examples/Algorithms/TrackFittingChi2/include/ActsExamples/TrackFittingChi2/TrackFittingChi2Algorithm.hpp
```diff
@@ -0,0 +1,131 @@
...
+    /// Input initial track parameter estimates for for each proto track.
+    std::string inputInitialTrackParameters;
+    /// Output fitted trajectories collection.
+    std::string outputTrajectories;
+    /// Type erased fitter function.
+    unsigned int nUpdates;
+    /// number of update steps
+    std::shared_ptr<TrackFitterChi2Function> fit;
+    /// Type erased direct navigation fitter function
+    // std::shared_ptr<DirectedTrackFitterFunction> dFit;
```
XiaocongAi, line 77: We can remove this option.

XiaocongAi (commented):
Hi @r4lv , thank you for implementing this new feature. I think we can clean up the code a bit. If there is something to be done, we can explicitly have the comment for that. Otherwise, we can just remove the comments. Also, we need to do some renaming of some variables.

Examples/Algorithms/TrackFittingChi2/src/TrackFittingChi2Options.cpp
```diff
@@ -0,0 +1,19 @@
...
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#include "ActsExamples/TrackFittingChi2/TrackFittingChi2Options.hpp"
+
+#include "ActsExamples/Utilities/Options.hpp"
+
+using namespace boost::program_options;
```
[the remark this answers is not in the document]
r4lv: I see that the other `addOptions()` in the code base use `ActsExamples::Options::Description` instead of `boost::program_options::options_description`. But as we would also need to change it in the KF, this should go in another PR, I think…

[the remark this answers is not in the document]
AJPfleger: I see, thank you. Since we have similar `using namespace` cases in numerous other places, I think we could keep it as it is for now and postpone the decision to a future PR.

Core/include/Acts/TrackFitting/Chi2Fitter.hpp
```diff
@@ -0,0 +1,725 @@
...
+      }
+
+      c2rCurrent.residuals =
+          Eigen::Map<ActsDynamicVector>(c2rCurrent.collectorResiduals.data(),
+                                        c2rCurrent.collectorResiduals.size());
+      // TODO: is this safe? from Stackoverflow: "dangerous! Because the Eigen
+      // object will NOT create its own memory. It will operate on the memory
+      // provided by "data". In other words, working with the Eigen object
+      // when the "data" object is out of scope will result in a segmentation
+      // fault (or memory access violation)."
```
AJPfleger, line 662: I think this may be important to check or at least to catch it in some way.

Unfortunately I don't have expertise in that field.

paulgessinger: The comment is correct, if the lifetime of `c2rCurrent.residuals` is longer than `c2rCurrent.collectorResidual`, it will be a problem. Is this the case? Other options would be to

1. Create a copy into `.residuals` (so dropping the map type) which might be more expensive
2. Creating the map type "on demand" when it is used, and discarding it immediately after

I think I would probably go with option 2.

r4lv:
With the recent addition of the GSF, the `Examples/Algorithms/TrackFitting/src/TrackFittingAlgorithm.cpp` is now using `GeneralFitterOptions` instead of `KalmanFitterOptions`. Can this be reused for the Gχ²F? For now, the Gχ²F is in its separate `TrackFittingChi2` folder, I think it would make sense to merge it with the KF+GSF, e.g. create a `Examples/Algorithms/TrackFitting/src/TrackFittingAlgorithmFunctionsChi2.cpp`.

Also the `Examples/Python/src/TrackFitting.cpp` ([diff](https://github.com/acts-project/acts/compare/b7d70c4a..83880e0#diff-a0346d9527c4778d50b91394c01430aad0f56675c5f533ca09503f85c7c827a1)), `Examples/Run/Reconstruction/Common/RecTruthTracks.cpp`([diff](https://github.com/acts-project/acts/compare/b7d70c4a..83880e0#diff-73c202264d82dd9f542bd17f902483b817567940258daea3cb146a4d323f453d)) and others have been updated with the new Fitter Options.

paulgessinger:
I'll have a look at the macOS CI failure.

AJPfleger:
macOS CI failure:
I ran the unit test on `Ubuntu 20.04` with `Clang 10.0.0` and could NOT reproduce the failure. 
~~~~~
:
:
        Start 126: GsfComponentMerging
126/212 Test #126: GsfComponentMerging ..................   Passed    0.02 sec
        Start 127: Chi2Fitter
127/212 Test #127: Chi2Fitter ...........................   Passed    0.06 sec
        Start 128: AnnealingUtility
128/212 Test #128: AnnealingUtility .....................   Passed    0.01 sec
:
:
100% tests passed, 0 tests failed out of 212
~~~~~

AJPfleger:
A followup on this topic is made with PR #1580 .

PR #1101, opened 2021-12-07
fix: Warning in navigation when targeting perigee

paulgessinger:
This warning popped up in a number of places, but mostly when targeting
parigee surfaces, e.g. in the vertexing or in the KF when the backward
smoothing propagation is enabled. It's possible this only occurs with a
beampipe layer.

I **think** what's happening is two things:
- The on-surface tolerance wasn't applied correctly in one location (I think we do this in other spots, like when we're looking for compatible layers in volumes, and compatible surfaces in layers.)
- If we actually have a target surface (and it's still reachable), this
  navigation state is probably still valid. We are on the final layer,
  the navigator realizes it won't find the perigee surfaces through the
  tracking geometry. It looks up the volume from the target surface
  position, and determines that we are in fact in the target volume
  already, so it only needs to *proceed to target* to be done.

As a tentative solution I've included the `s_onSurfaceTolerance` in the
check in `SteppingHelper` (@asalzburger can you advise if that seems
correct here?), and split the output into a warning in case there is no
target surface (which I still think is a bug/edge case we should warn on), and a verbose messagge in case we do have it.

One could supplement the last change by checking if at this point the
target surface is still reachable, but I'm not sure this should be done
only for reason of providing output.

### DIFF
--- a/Core/include/Acts/Propagator/Navigator.hpp
+++ b/Core/include/Acts/Propagator/Navigator.hpp
@@ -389,8 +389,14 @@ class Navigator {
       }
     } else if (state.navigation.currentVolume ==
                state.navigation.targetVolume) {
-      ACTS_WARNING(volInfo(state) << "No further navigation action, proceed to "
-                                     "target. This is very likely an error");
+      if (state.navigation.targetSurface == nullptr) {
+        ACTS_WARNING(volInfo(state)
+                     << "No further navigation action, proceed to "
+                        "target. This is very likely an error");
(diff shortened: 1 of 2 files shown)


asalzburger (approved):
Looks good!

Corentin-Allaire (approved):
I just tested it with the ATLAS track refitting and I confirm it solves the issue there

paulgessinger:
Fantastic!

PR #1102, opened 2021-12-07
fix: Added wait statements to kernels and changed the atomic operation to use VecMem

konradkusiak97:
While testing the Sycl seedfinder on Intel's Arctic Sound P card, we found  with @krasznaa that the code is missing some explicit `wait` statements in a few places that would result in waiting for the kernel to finish before executing later part of the code. 

This hasn't appeared before in our tests probably because, in none of the devices that we run the code on, the kernels were run asynchronously. 

This is now fixed with this PR. 
I also changed the outdated handling of the atomic operation to make use of `vecmem::atomic` which we discussed with @krasznaa already some time ago.


krasznaa:
Konrad, you should've left out the switch to a different memory resource from this PR. Let's keep this one focused on just fixing the scheduling bugs.

What you need to do now is to fix up the CI. We can discuss offline if you don't know how to do that, but I'd let you take a crack at it yourself at first. As you can see, the formatting of your code changes does not pass the repository's formatting rules. You will need to fix that with `clang-format`.

paulgessinger:
@krasznaa can you review & approve this?

Plugins/Sycl/src/Seeding/CreateSeedsForGroupSycl.cpp
```diff
@@ -546,9 +541,24 @@ void createSeedsForGroupSycl(
 
       // Counting the seeds in the second kernel allows us to copy back the
       // right number of seeds, and no more.
-
       seeds.resize(M);
-      std::vector<uint32_t> deviceCountTriplets(edgesBottom, 0);
+      vecmem::vector<uint32_t> countTriplets(&resource);
+      countTriplets.resize(edgesBottom, 0);
```
krasznaa, line 546: This is more of a PSA for Konrad's benefit, it does not necessarily need to be changed here.

VecMem vectors can still be created with a single line, to be initialised to some size/value. Like:

```c++
vecmem::vector<uint32_t> countTriplets(edgesBottom, 0, &resource);
```

This uses the same `std::vector` constructor ("number 3" from https://en.cppreference.com/w/cpp/container/vector/vector) as the previous incantation used. We just rely on the memory resource pointer being "auto converted" into [std::pmr::polymorphic_allocator](https://en.cppreference.com/w/cpp/memory/polymorphic_allocator).

Just something to keep in mind for future coding... :wink:

Plugins/Sycl/src/Seeding/TripletSearch.hpp
```diff
@@ -204,7 +204,10 @@ class TripletSearch {
           const auto top = deviceIndTopDuplets[it];
           // this will be the t-th top space point for
           // fixed middle and bottom SP
-          auto t = m_countTripletsAcc[ib].fetch_add(1);
+          vecmem::device_vector<uint32_t> deviceCountTriplets(
+              m_countTripletsView);
+          vecmem::atomic obj(&deviceCountTriplets[ib]);
```
krasznaa, line 209: Wow... This actually works? :smile: That's great!

As you know, [vecmem::atomic](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/memory/atomic.hpp) is a template. I know that C\+\+17 (and oneAPI in particular) can do this. (https://en.cppreference.com/w/cpp/language/class_template_argument_deduction) I just never tried...

I was pondering for a moment whether to allow this to go through, or to ask you to change `vecmem::atomic` to `vecmem::atomic<uint32_t>`. But you know what, let's keep it as-is. Let's not worry about SYCL implementations that do not support CTAD...

krasznaa (approved):
Yeah, looks good on my end. :+1:

krasznaa:
Wait... I'm not allowed to resolve/close discussions? :confused: I only opened those two to point out two "interesting things" in the code. They should not be holding up this PR.

paulgessinger:
That's odd indeed. 🤔

krasznaa:
The whole interface is just a lot more "jerky" today than I'd like. I don't know if it's some disagreement between Chrome and GitHub, or what... :confused:

But I'm not keen on investigating it much further today...

paulgessinger:
Weird.

Kodiak should hopefully take it from here.

Issue #1077, opened 2021-11-16
Vertexing triggers propagation warnings frequently


paulgessinger:
We noticed in ATLAS recently that the vertexing frequently runs into 

```
08:51:03 08:51:03 IPEstProp WARNING Pixel::Barrel | No further navigation action, proceed to target. This is very likely an error
```

which is triggered [here](https://github.com/acts-project/acts/blob/2dc809f7b3077abf8fd862b41c27e08caedace88/Core/include/Acts/Propagator/Navigator.hpp#L392-L393)

```cpp
} else if (state.navigation.currentVolume ==
               state.navigation.targetVolume) {
      ACTS_WARNING(volInfo(state) << "No further navigation action, proceed to "
                                     "target. This is very likely an error");
      // Set navigation break and release the navigation step size
      state.navigation.navigationBreak = true;
      stepper.releaseStepSize(state.stepping);
    } else {
```

I think I've observed this in the past also in ACTS standalone vertexing examples, but I'll have to go back and check if I can reproduce.

PR #1103, opened 2021-12-08
feat: Allow setting reverse filter p threshold

paulgessinger:
We previously didn't set up the reverse filtering at all in the examples track fitting algorithm, so the WARNING in #1101 and #1077 was never observed.

This PR exposes a momentum threshold variable through the config struct and into python. The pytests run it with a threshold of 0 (no reverse filtering at all) and 1 TeV (almost always reverse filtering), so we should at least see this warning here until #1101 is merged.


Examples/Scripts/Python/truth_tracking.py
```diff
@@ -14,6 +14,7 @@ def runTruthTracking(
     outputDir: Path,
     digiConfigFile: Path,
     directNavigation=False,
+    reverseFilteringMomThreshold=0*u.GeV,
```
asalzburger, line 17: This effectively switches reverse filtering off, right?

asalzburger (approved):
Makes totally sense.

asalzburger:
There's still a failure & the format checks are running.

Examples/Scripts/Python/truth_tracking.py
paulgessinger: Exactly, which is what the previous default was.

paulgessinger:
> There's still a failure & the format checks are running.

The failure is because #1101 is not in this branch yet. I'll update and it should resolve the failure (as intended).

paulgessinger:
I also fixed the format now.

paulgessinger:
Ok, everything seems to succeed.
@asalzburger can you approve again?

paulgessinger:
The CI failure was likely caused by a regression introduced in #1103. This should be fixed in #1115. This branch now includes that commit. The CI should run through now (I hope), but I'll mark this one WIP again until #1115 is merged.

paulgessinger:
Tests succeeds now it seems. @asalzburger can you approve one more time?

paulgessinger:
@asalzburger The bot isn't merging because the approval itself is not valid yet.

asalzburger:
Merged by hand.

PR #1104, opened 2021-12-08
feat: Adding option for CylinderBounds to have a bevelled edges

noraemi:
PR to introduce a modification of the CylinderBounds to include an angle to tilt both edges of the cylinder.   
WIP for now to receive initial feedback on the implementation, e.g. if this should rather be an extension
- Before updating all the boundary checks / isInside / etc that I haven't done yet, I wonder if its fine to update the VerticesHelper as I have done or if its preferred to do this tilt purely by a transform (which proved to be quite difficult)

Current implementation keeps the length in z at the mid-point of the cylinder and tilt the edges with an defined angle
image


Core/include/Acts/Surfaces/CylinderSurface.hpp
```diff
@@ -49,7 +49,7 @@ class CylinderSurface : public Surface {
   /// @param halfphi The half opening angle
   /// @param avphi The phi value from which the opening angle spans (both sides)
   CylinderSurface(const Transform3& transform, double radius, double halfz,
-                  double halfphi = M_PI, double avphi = 0.);
+                  double halfphi = M_PI, double avphi = 0., double angle = 0.);
```
paulgessinger, line 52: I might be missing something here, but I would have thought this `angle` parameter could be restricted to `CylinderBounds` (and probably `CylinderVolumeBounds`).

Core/include/Acts/Surfaces/RadialBounds.hpp
```diff
@@ -43,8 +44,8 @@ class RadialBounds : public DiscBounds {
   /// @param halfPhi The half opening angle (Pi for full angular coverage)
   /// @param avgPhi The average phi for the disc/ring sector
   RadialBounds(double minR, double maxR, double halfPhi = M_PI,
-               double avgPhi = 0.) noexcept(false)
-      : m_values({minR, maxR, halfPhi, avgPhi}) {
+               double avgPhi = 0., double angle = 0.) noexcept(false)
+      : m_values({minR, maxR, halfPhi, avgPhi, angle}) {
```
paulgessinger, line 48: I'm not sure I understand what the meaning of `angle` for `RadialBounds` is?

asalzburger:
Screenshot 2021-12-09 at 09 44 42

The word is 'bevel' - had to look it up as well.

Core/include/Acts/Surfaces/detail/VerticesHelper.hpp
```diff
@@ -62,6 +62,12 @@ void createSegment(std::vector<vertex_t>& vertices,
     vertex_t vertex = vertex_t::Zero();
     vertex(0) = rxy.first * std::cos(phi);
     vertex(1) = rxy.second * std::sin(phi);
+
+    if(angle != 0. && vertex.size() == 3) {
+      double sign = offset(2) != 0. ? std::fabs(offset(2))/offset(2) : 1.;
+      vertex(2) = vertex(2) + sign * vertex(1) * std::tan(angle);
+    }
+
```
paulgessinger, line 70: Is this actually called anywhere in the current changeset?

What I was originally thinking (and please correct me if I'm missing something), was to move this part:

https://github.com/acts-project/acts/blob/98597a7191e142015964c6dba5a3f535b6ace905/Core/src/Surfaces/CylinderSurface.cpp

into the bounds itself. Basically: `CylinderSurface` would call on the bounds to deliver (or maybe fill via `push_back`) the `vertices` vector. 

The reason for this is that in principle, `CylinderSurface` handles the coordinate system, and doesn't really care about the specific bounds that have been set.

With this approach, we could confine the fraction of the code that needs to be aware of this `angle` parameter, and also we would have the option of using a separate class for this.

paulgessinger (commented):
Thanks @noraemi!

I left a few comments. One general one: I **think** we'll need the possibility to have only one of the sides of the cylinder have an inclined cutoff. I guess this would mean having two `angle` parameters?

Core/include/Acts/Geometry/CylinderVolumeBounds.hpp
```diff
@@ -77,6 +77,7 @@ class CylinderVolumeBounds : public VolumeBounds {
     eHalfLengthZ = 2,
     eHalfPhiSector = 3,
     eAveragePhi = 4,
+    eAngle = 5,
```
asalzburger, line 80: * can we name this `eBevelAngle` ?

I think we need different angles on both sides, i.e.

`eBevelMinZ`
`eBevelMaxZ`

Because we will need different use cases for the ATLAS MS and the Athena-EIC beam pipe:

Screenshot 2021-12-09 at 09 51 29

Core/include/Acts/Surfaces/RadialBounds.hpp
```diff
@@ -31,7 +31,8 @@ class RadialBounds : public DiscBounds {
     eMaxR = 1,
     eHalfPhiSector = 2,
     eAveragePhi = 3,
-    eSize = 4
+    eAngle = 4,
```
asalzburger, line 34: Why do the `RadialBounds` change ?

For the new CylinderShape we might need instead something like `RadialEllipsoidBounds` or can we get them into one hood (actually I think we could live with `EllipsoidBounds` and keep the coordinate system cartesian).

Core/include/Acts/Surfaces/detail/VerticesHelper.hpp
noraemi: Thanks for the feedback, this is exactly why I made the draft to get comments whether I was doing something non-optimal. OK, will try this approach instead

Core/src/Geometry/CylinderVolumeBounds.cpp
```diff
@@ -33,6 +33,7 @@ Acts::CylinderVolumeBounds::CylinderVolumeBounds(
   m_values[eHalfLengthZ] = cBounds.get(CylinderBounds::eHalfLengthZ);
   m_values[eHalfPhiSector] = cBounds.get(CylinderBounds::eHalfPhiSector);
   m_values[eAveragePhi] = cBounds.get(CylinderBounds::eAveragePhi);
+  m_values[eAngle] = cBounds.get(CylinderBounds::eAngle);
```
asalzburger, line 36: Same as above, I think we need two angles, one on either side ...

Core/src/Geometry/CylinderVolumeBounds.cpp
```diff
@@ -48,6 +49,7 @@ Acts::CylinderVolumeBounds::CylinderVolumeBounds(
   m_values[eHalfLengthZ] = 0.5 * thickness;
   m_values[eHalfPhiSector] = rBounds.get(RadialBounds::eHalfPhiSector);
   m_values[eAveragePhi] = rBounds.get(RadialBounds::eAveragePhi);
+  m_values[eAngle] = rBounds.get(RadialBounds::eAngle);
   buildSurfaceBounds();
 }
```
asalzburger, line 55: In ATLAS we have solved this with `BevelledCylinderVolumeBounds` in order to not have too many if statements ... not sure what's the better solution here.

asalzburger (changes requested):
Great to get this started, @noraemi !

I think we need two different angles, and -- what I totally forgot to mention -- of course we need to update the 

* `SurfaceBounds::inside()`
* `VolumeBounds::inside()`

methods, there's a lot of code we can just take from the equivalent ATLAS classes.

Then, to complete the PR, this needs UnitTests for the added code to check consistency.

asalzburger:
As for the format checks, we are using clang-format, and if the correct/right version is installed locally, one can run 

```shell
./Ci/check_format .
```

on the repository and this should fix the format checker.

Core/include/Acts/Surfaces/detail/VerticesHelper.hpp
paulgessinger: I messed up the code snippet linked in the comment above. Fixed now.

Core/include/Acts/Surfaces/RadialBounds.hpp
noraemi: This was a mistake, left over by a trial implementation of bevelling of the "side"-disc but in the end I didn't need it but I forgot to remove it

Core/include/Acts/Geometry/CylinderVolumeBounds.hpp
noraemi: Done!

Core/include/Acts/Surfaces/detail/VerticesHelper.hpp
noraemi: An attempt has been made to move this to `CylinderBounds`

Core/include/Acts/Surfaces/RadialBounds.hpp
noraemi: Same answer as to Andi: This was a mistake, left over by a trial implementation of bevelling of the "side"-disc but in the end I didn't need it but I forgot to remove it

Core/include/Acts/Surfaces/CylinderSurface.hpp
noraemi: Should be gone now

noraemi:
I ran into a snag trying to finish up the last parts, and some feedback would be helpful. The issue is with the "inside" check for the cylinderbounds. This check does box, with the phi/Z, check as far as  I can tell. For the bevel part this would obviously not work. Would the right approach here be to extend the functionality in BoundaryCheck to have say an "isInsideRhombus" function or similar? Where I could define the tilted box rather. And call this function instead of isInside for the cases where the bevel parameters are set.

```
bool Acts::CylinderBounds::inside(const Vector2& lposition,
                                  const BoundaryCheck& bcheck) const {
  return bcheck.transformed(jacobian())
      .isInside(shifted(lposition),
                Vector2(-get(eHalfPhiSector), -get(eHalfLengthZ)),
                Vector2(get(eHalfPhiSector), get(eHalfLengthZ)));
}
```

paulgessinger:
image

image

asalzburger:
Regarding the `BoundaryCheck` - as a start, it will be good enough with the fixed tolerance, although we should be able to use the generic code when unfolded.

paulgessinger:
I think since in local coordinates the edges are straight, even the covariance check should be ok: convert to local with jacobian und use the vertex inside check in `BoundaryCheck`.

noraemi:
> Regarding the `BoundaryCheck` - as a start, it will be good enough with the fixed tolerance, although we should be able to use the generic code when unfolded.

Isn't this already handled by `boundaryCheck.isTolerated(closestPoint - lposition);` ? It performs the type check there whether to use the absolute or more

Core/src/Geometry/CylinderVolumeBounds.cpp
asalzburger: Let's keep it though.

asalzburger:
Hi @noraemi - one quick question: can the bevelling also be in the same direction on both sides?

The ATHENA-EIC guys need a tilted beam pipe inlay into a cylinder volume (we can specify that volume shape later), but that would need some sort of parallel cut off on both sides.

Core/src/Geometry/CylinderVolumeBounds.cpp
noraemi: > Hi @noraemi - one quick question: can the bevelling also be in the same direction on both sides?
> 
> The ATHENA-EIC guys need a tilted beam pipe inlay into a cylinder volume (we can specify that volume shape later), but that would need some sort of parallel cut off on both sides.

Yes! the bevel can be done in both directions e.g: \\___    or    /___
the two sides are independent as well. Will add a comment about what the bevel is w.r.t somewhere in the CylinderBounds

noraemi:
> Hi @noraemi - one quick question: can the bevelling also be in the same direction on both sides?
> 
> The ATHENA-EIC guys need a tilted beam pipe inlay into a cylinder volume (we can specify that volume shape later), but that would need some sort of parallel cut off on both sides.

Yes! the bevel can be done in both directions e.g: \___ or /___
the two sides are independent as well. Will add a comment about what the bevel is w.r.t somewhere in the CylinderBounds

Core/src/Geometry/CylinderVolumeBounds.cpp
noraemi: Hmm, I think I forgot this. In anycase, wouldn't this mean quite a lot of duplicated code with minor additions? But perhaps that's the preferred option..

asalzburger (approved):
This looks ok to me now, is there more to do here from your side, Nora?

asalzburger:
Updating branch to check merge-ability,

noraemi:
> This looks ok to me now, is there more to do here from your side, Nora?

I'm just adding the unit-tests for the bevel, after that I think its done

noraemi:
This PR should be ready now, if the CI doesn't fail.

paulgessinger (approved):

PR #1105, opened 2021-12-08
feat: Add ostream-operator for IntersectionStatus

benjaminhuth:
This is the first part of my GSF pull-request series, where I often need to print the Intersection status for debugging. I split up this PR, since it is independent on the following ones.

Basically this pulls the nested `enum class Status` out of the `Intersection<DIM>` template struct to a `enum class IntersectionStatus` at namespace scope. The `IntersectionStatus` is broadcasted into the `Intersection<DIM>` with a `using Status = IntersectionStatus` to remain compatability.

This is done because otherwise the `Status`-enum would depend on the template-parameter `DIM` and thus the `operator<<` would also needed to be templated on `DIM` which makes things more complicated than they need to be.

### DIFF
--- a/Core/include/Acts/Utilities/Intersection.hpp
+++ b/Core/include/Acts/Utilities/Intersection.hpp
@@ -25,13 +25,12 @@ enum class IntersectionStatus : int {
 };
 
 /// Ostream-operator for the IntersectionStatus enum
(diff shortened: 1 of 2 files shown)


paulgessinger (approved):
LGTM!

(CI complains about the format, should be easy to fix)

paulgessinger (approved):

PR #1107, opened 2021-12-09
fix: Comparison bug in updateSingleSurfaceStatus

benjaminhuth:
I think I found actually two small bugs in the `updateSingleSurfaceStatus` function:

1) The first one is older I think: when `pLimit == cLimit + s_onSurfaceTolerance`, then the intersection is not accepted, but there is also no message printed out, because it is only checkt for `<` and `>`, but one should be `<=` or `>=` in this case I think.

2) With the `s_onSurfaceTolerance`, I think we do not compare what we accually want here anymore:

   What we check is something like `a*a < b*b + c`, but what we (I think) want to check is `|a| < |b| + c`. This is obviously not the same (e.g. with a=4, b=3, c=2).

   So I know also added `std::abs` in the computation of the `accept`-boolean. Not sure if that is the best way, since it seems that was not done for performance reasons or something like that before. One could also change the verbose message to print out the actual criterion...


paulgessinger:
You're right, I didn't think this through in `SteppingHelper`. I very recently changed this in #1101 to fix a warning we had here. I think we use very similar comparisons in a few other spots, could you check / fix those too?

paulgessinger:
Btw, let's just use `abs`, I really don't think this is something where performance matters for our use-case.

benjaminhuth:
Ok, I have no done a few more things...

Since this functionality is used at at least 3 places, I thought it might be worth to put it in a separate function. This function I put in the namespace `detail` in the `Utilities/Intersection.hpp`.

This function does just the checking and return a `bool`, while the action what should be done when the intersection is valid (e.g., set the stepsize or return another intersection-object) I pulled out of the function. 

The function does take a `LoggerWrapper`-object, but once could also think about a template-parameter-boolean to prevent the generation of the Logging-Code in places where this is not desired.

Since it is also a template-function, I think it will be inlined...

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -175,4 +177,60 @@ struct SameSurfaceIntersection {
...
+/// prevent the generation of logging code
+///
+/// @param intersection The intersection to check
+/// @param pLimit The path-limit
+/// @param oLimit The overstep-limit
+/// @param logger A optionally supplied logger which prints out a lot of infos
+/// at VERBOSE level
+template <typename intersection_t, typename logger_t = std::false_type>
+bool checkIntersection(const intersection_t& intersection, double pLimit,
+                       double oLimit, logger_t logger = logger_t{}) {
```
paulgessinger, line 196: I would probably prefer having the tolerance be an explicit parameter here. What do you think @benjaminhuth?

benjaminhuth: Good idea!

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -194,7 +194,7 @@ namespace detail {
 template <typename intersection_t, typename logger_t = std::false_type>
 bool checkIntersection(const intersection_t& intersection, double pLimit,
                        double oLimit, double tolerance,
-                       logger_t logger = logger_t{}) {
+                       [[maybe_unused]] logger_t logger = logger_t{}) {
```
paulgessinger, line 197: Unfortunately I don't think this works everywhere yet.

In other spaces we had to use 
```cpp
(void)logger;
```
in the function body to suppress the error.

benjaminhuth: Hmm but it seems to build with the `[[maybe_unused]]` attribute...

paulgessinger: Ok. If the CI is happy, then let's not bother.

paulgessinger (approved):

PR #1109, opened 2021-12-10
fix: missing array include

Corentin-Allaire:
The `#include <array>` was missing from `Intersection.hpp`. I have no idea why it build in the CI and not on my machine but in any case it is probably better to have it here


paulgessinger (approved):
OK.

PR #1110, opened 2021-12-13
feat: MultiStepper for the GSF

benjaminhuth:
Requires #1113 

This PR adds mainly the class `MultiEigenStepperLoop` with some helper code. 

It mainly adds new, independent code, the only case where it interferes with other propagation code is the stepper-concept.

The contained UnitTests cover most functionality of the `MultiEigenStepperLoop` and are built mainly as template-functions, so they can later be reused for the other MultiStepper-implementation (which will come later). The functionality is mostly tested against the `EigenStepper`.


paulgessinger:
Alright, #1113 is merged. Is this ready to review now, or are additional changes needed, @benjaminhuth?

benjaminhuth:
This should be ready to review I think!

Core/include/Acts/EventData/MultiComponentBoundTrackParameters.hpp
```diff
@@ -0,0 +1,240 @@
...
+  MultiComponentBoundTrackParameters& operator=(
+      MultiComponentBoundTrackParameters&&) = default;
+
+  /// Access the parameters
+  const auto& components() const { return m_components; }
+
+  /// Reference surface onto which the parameters are bound.
+  const Surface& referenceSurface() const { return *m_surface; }
+
+  /// Get a SingleBoundTrackParameters object for one component
```
asalzburger, line 160: This returns at the weight and the single component ?
If so, please indicate both in the comment

Core/include/Acts/EventData/MultiComponentBoundTrackParameters.hpp
```diff
@@ -0,0 +1,240 @@
...
+  Vector3 momentum() const {
+    return reduce([](const SingleParameters& p) { return p.momentum(); });
+  }
+
+  /// Particle electric charge.
+  Scalar charge() const {
+    return reduce([](const SingleParameters& p) { return p.charge(); });
+  }
+};
+
```
asalzburger, line 239: A question here, for potential future optimisation ... in the Navigation, there's often the successive call `position()`, `direction()` ... would it make sense to implement a tuple call, such that the reduction has to be done only once ?

Core/include/Acts/Propagator/MultiEigenStepperLoop.hpp
```diff
@@ -0,0 +1,800 @@
...
+
+  template <typename stepper_state_t>
+  static ActsScalar momentum(const stepper_state_t& s) {
+    return std::accumulate(
+        s.components.begin(), s.components.end(), ActsScalar{0.},
+        [](const auto& sum, const auto& cmp) -> ActsScalar {
+          return sum +
+                 cmp.weight * (1 / (cmp.state.pars[eFreeQOverP] / cmp.state.q));
+        });
+  }
```
asalzburger, line 73: Se my comment from before - do you have a feeling if we could gain from something like this?

Core/include/Acts/Propagator/MultiEigenStepperLoop.hpp
```diff
@@ -0,0 +1,800 @@
...
+    auto& status() { return cmp.status; }
+    auto& weight() { return cmp.weight; }
+    auto& charge() { return cmp.state.q; }
+    auto& pathAccumulated() { return cmp.state.pathAccumulated; }
+    auto& pars() { return cmp.state.pars; }
+    auto& derivative() { return cmp.state.derivative; }
+    auto& jacTransport() { return cmp.state.jacTransport; }
+    auto& cov() { return cmp.state.cov; }
+    auto& jacobian() { return cmp.state.jacobian; }
+    auto& jacToGlobal() { return cmp.state.jacToGlobal; }
```
asalzburger, line 407: Do we really need `auto` keywords here everywhere - I think these are explicit types, aren't they ?

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
```diff
@@ -0,0 +1,200 @@
...
+    int failedBoundTransforms = 0;
+
+    for (auto i = 0ul; i < numberComponents(state); ++i) {
+      auto bs = SingleStepper::boundState(state.components[i].state, surface,
+                                          transportCov);
+
+      if (bs.ok()) {
+        states.push_back(
+            {state.components[i].weight, std::get<BoundTrackParameters>(*bs)});
+        accumulatedPathLength +=
```
asalzburger, line 31: This implicitly implies that the navigation path is a weighted mean of all paths, will this hold for all merging policies? 
(probably)

I guess the accumulatedPathLength is anyway only needed/used for the path length aborter and the path limit settings in this scope, so I *guess* this should be fine.

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
```diff
@@ -0,0 +1,200 @@
...
+      ACTS_ERROR("Multi component bound state: "
+                 << failedBoundTransforms << " of " << numberComponents(state)
+                 << " transforms failed");
+    }
+
+    if (states.size() == 0) {
+      return MultiStepperError::AllComponentsConversionToBoundFailed;
+    }
+
+    // TODO also implement a method of using the mode of the mixture
```
asalzburger, line 48: I don't understand the comment here - do you want to make this configurable?

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
```diff
@@ -0,0 +1,200 @@
...
+    BoundSymMatrix cov = BoundSymMatrix::Zero();
+    ActsScalar pathLenth = 0.0;
+
+    for (auto i = 0ul; i < numberComponents(state); ++i) {
+      const auto [cp, jac, pl] = SingleStepper::curvilinearState(
+          state.components[i].state, transportCov);
+
+      pos4 += state.components[i].weight * cp.fourPosition(state.geoContext);
+      dir += state.components[i].weight * cp.unitDirection();
+      qop += state.components[i].weight * (cp.charge() / cp.absoluteMomentum());
```
asalzburger, line 80: Couldn't one do the pos4, mom4 merging first, then distil `pos4, dir, q/p` out of it?

Somehow I think this would be more stable, but I can wrong.

Core/include/Acts/Utilities/Identity.hpp
```diff
@@ -0,0 +1,22 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#include <utility>
+
+namespace Acts {
+
+/// @brief Function object which maps a value to itself by perfect forwarding
+/// TODO Replace Identity with std::identity on C++20
```
asalzburger, line 14: If `std::identity` implements the same interface, we could make this with a pragma already and alias of std::identity,  I always fear that we'll forget to remove this sort of code.

asalzburger (approved):
I have a view points which would be great to have some clarification on.

Core/include/Acts/Propagator/MultiEigenStepperLoop.hpp
asalzburger: It's not against, it sort of hides whether the return type is e.g. a `Object&` or a `const Object&` which has a big consequence to the caller, I just wanted tot make this clear. That can be just done with a comment, then we're good, as they are accessors it's clear then.

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
benjaminhuth: Yeah, in all places where a single `pathLength`-representation for all components is needed, I did a weighted sum (e.g. for the return value of the `MultiEigenStepperLoop::step(...)`-function.
The pathlength of such a multi-component-propagation isn't really defined I think, and this is the value which comes closest to it. 
One could also use the pathlenght of the maximum-weight component e.g., but in normal use cases this should not make a huge difference.

Core/include/Acts/Propagator/MultiEigenStepperLoop.hpp
benjaminhuth: Yeah I see that it might be not quite clear, that the `const`-qualified accessors are inherited from the `ComponentProxyBase`. I did this to avoid code duplications between the const and the non-const `ComponentProxy`...

I'll add a comment to clarify this.

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
benjaminhuth: That was one thing that came up in the meeting with Anthony and Christos, they said that at ATLAS they use the mode of the mixture for the final parameters, not the mean. Currently the bound states are combined by computing the mean. This probably should be tried out during the validation phase?

However, I will clarify this comment with a better formulation

Core/include/Acts/EventData/MultiComponentBoundTrackParameters.hpp
benjaminhuth: This particular class (the `MultiComponentBoundTrackParameters<charge_t>`) should never interact with the navigation. All these accessors with the reductions are required by the `BoundTrackParametersConcept` in `TrackParametersConcept.hpp`.

However (as you also commented below), for the interaction between the `Stepper::State` and the `Navigator`, this could potentially save some computations.

Core/include/Acts/Propagator/MultiEigenStepperLoop.hpp
benjaminhuth: As mentioned above: Yes this could potentially reduce both code size and increase performance here...

However, I'm not sure if I would implement this immediately or do a big performance measurement some time and keep it in mind...

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
benjaminhuth: Rigth now  I'm not sure how one would do this with the mom4... 

But now where I look at it, this function (and some others...) outputs garbage if the charges of the components are not equal, right? I will think this through...

paulgessinger:
Will merge once the threads are resolved. @asalzburger, @benjaminhuth can you resolve the threads when complete?

Core/include/Acts/Propagator/MultiEigenStepperLoop.hpp
benjaminhuth: So to resume on this:

You mean something like this, right? (taken arbitrarily from Navigator.hpp:366)

```c++
state.navigation.currentVolume = boundary->attachedVolume(
     state.geoContext, stepper.position(state.stepping),
     stepper.direction(state.stepping), state.stepping.navDir);
```

And here do something like:

```c++

const auto [pos, dir] = stepper.posdir(state.stepping);

state.navigation.currentVolume = boundary->attachedVolume(
     state.geoContext, pos, dir, state.stepping.navDir);
```

But this would change the whole stepper-navigator-interaction, right? Or do I misunderstand what your idea is?

asalzburger: We can leave this as an optimisation step for later.

asalzburger:
I am satisfied with the answers, we can do the optional optimisation step in a second round.

asalzburger:
I will not add the kodiac auto-merge, as this PR is very big, it may need another rebase before merging.

asalzburger (approved):
Thanks for adding the 'identity' change.

PR #1114, opened 2021-12-14
fix: adding missing averagePhi in CylinderVolumeBounds

asalzburger:
Cylindrical volumes can cover only a sectoral part, if a `phi` opening angle and a `averagePhi` is defined.

The sectoral planes were only correct for `averagePhi=0`, as this parameter was missing in the rotation.


Corentin-Allaire (approved):
Looks good ! So if I understand properly eAveragePhi was actually never used ?

asalzburger:
> Looks good ! So if I understand properly eAveragePhi was actually never used ?

We don't use these type of volumes in regular detectors yet, but we will try for the Muon System navigation, so this was not caught,

asalzburger:
Since this has been approved, I merge it.

PR #1115, opened 2021-12-14
fix: TrackingVolume intersection check didn't copy sign

paulgessinger:
I noticed in #1103 that the navigation failed in backward mode after we merged #1107.

Indeed, during the change in #1107, `TrackingVolume` lost a line that copied the sign from the navigation direction onto the path length in `compatibleBoundaries`. This made the navigation go haywire when it tried to target boundaries in reverse propagation mode.

This PR adds that back in, and should fix #1103's test runs.


benjaminhuth (approved):
Ah I lost these two lines somehow. Thanks, this looks good to me!

PR #1117, opened 2021-12-14
fix: Missing include in EigenStepper.hpp

paulgessinger:
Self explanatory. This doesn't give a compilation error because we never compile the Eigen Stepper header on it's own.


Corentin-Allaire (approved):
Thanks for finding this !

PR #1118, opened 2021-12-14
refactor: Optimize `TrapezoidBounds` inside check

paulgessinger:
This was previously delegating to `BoundaryCheck::isInside` with a vector of vertices. This PR:

- Does a number of early checks for the trapezoid shape: inside core rectangle, outside y, outside larger x halflength
- Uses an `std::array` for the vertices it passes to the `isInside` call, which reduces the number of heap allocations.


paulgessinger:
Actually, I don't think this is correct in case of a boundary check with tolerances. Marking WIP until I've implemented that.

paulgessinger:
Should be correct now.

PR #1119, opened 2021-12-14
refactor: Move Layer resolution functions to `.cpp`, reduce heap allocations

paulgessinger:
With a simple forward declaration, `Layer::compatibleSurfaces` and `Layer::surfaceOnApproach` can be moved to the `Layer.cpp` file. 

I also changed the way `compatibleSurfaces` resolves duplicates: instead of filling an auxiliary `map`, the vector of surface intersections is sorted by associated object first to remove duplicates, and only then sorted by path length.


robertlangenberg (approved):
Zealous Paul at it again :)

PR #1120, opened 2021-12-14
refactor: Use `small_vector` in `surfaceOnApproach`

paulgessinger:
Reduces the number of heap allocations in that method. As the number of approach surfaces is *almost always* 2, I think this makes sense to use a `small_vector` for.


asalzburger (approved):
uncontroversial

PR #1122, opened 2021-12-16
test: Refactor kalman-fitter test so it can be reused for other fitters

benjaminhuth:
This PR restructures the testing code for the `KalmanFitter` so that it can be reused for other fitters. This especially targets the GSF, but could also be useful for the GX2F #1099.

This is mainly done by putting the code inside a header file and templating the functions on `fitter_t`, `fitter_options_t` and `parameters_t`.

This should not change the behavior or the results of the tests.


Tests/UnitTests/Core/TrackFitting/KalmanFitterTests.cpp
```diff
@@ -150,482 +65,124 @@ Acts::CurvilinearTrackParameters makeParameters() {
...
-const auto kfZeroPropagator = makeConstantFieldPropagator(geometry, 0_T);
+const auto kfZeroPropagator =
+    makeConstantFieldPropagator<ConstantFieldStepper>(tester.geometry, 0_T);
 const auto kfZero = KalmanFitter(kfZeroPropagator);
 
 std::default_random_engine rng(42);
 
+auto makeDefaultKalmanFitterOptions() {
+  return KalmanFitterOptions(tester.geoCtx, tester.magCtx, tester.calCtx,
+                             getExtensions(), LoggerWrapper{*kfLogger},
```
paulgessinger, line 81: I guess if there's this function, there's no need for a separate `getExtensions` anymore is there? You consider manually inlining that here.

paulgessinger (approved):
Ok, let's go.

PR #1123, opened 2021-12-16
refactor: Hard coded matrix operations

paulgessinger:
I ran some tests and it seems to me like we spend about 20% of our time in this expression:

https://github.com/acts-project/acts/blob/d62c2f9a22a1cb4e2376c9aa2fcb488f256a0cad/Core/src/Propagator/detail/CovarianceEngine.cpp#L156-L159

along with some other heavy matrix expressions.

My guess is Eigen is not super efficient for 6x6 and 8x8 matrices. This PR replaces these expressions with hard-coded calculations, generated by a python script. This then looks like:

https://github.com/acts-project/acts/blob/29c995088f098ff8ac370ca1acf818900c44fa0c/Core/src/Propagator/detail/CovarianceEngine.cpp#L161-L164

This is meant as a discussion prompt, I'm not sure hard-coding this math is a great idea necessarily.

Some benchmarking leads me to believe that it has an overall measurable impact on the performance. I haven't run any microbenchmarks.

On `main` I get 

```
AtlasStepper (cov=True): 19.686031s / 1015949 tracks per second
AtlasStepper (cov=False): 13.950861s / 1433603 tracks per second
EigenStepper (cov=True): 30.063474s / 665259 tracks per second
EigenStepper (cov=False): 13.740095s / 1455594 tracks per second
```

while on this branch I get

```
AtlasStepper (cov=True): 18.965792s / 1054530 tracks per second
AtlasStepper (cov=False): 13.505438s / 1480885 tracks per second
EigenStepper (cov=True): 27.210936s / 734999 tracks per second
EigenStepper (cov=False): 13.808752s / 1448357 tracks per second
```

(tested with 1000 tracks per event and 20000 events each)


paulgessinger:
@HadrienG2 @asalzburger what do you think?

Core/include/Acts/Utilities/HelpersMatrixGenerated.ipp
```diff
@@ -0,0 +1,825 @@
+//////////////////////////////////////////////////
+// This function is AUTOGENERATED. Do not edit! //
+//////////////////////////////////////////////////
+inline ActsMatrix<8, 8> multiply(const ActsMatrix<8, 1>& A,
+                                 const ActsMatrix<1, 8>& B) {
```
HadrienG2, line 5: You may want to take matrices A and B by value. The compiler should be able to elide the copy, and this will eliminate any risk of losing compiler optimizations due to pointer aliasing issues.

Core/include/Acts/Utilities/HelpersMatrixGenerated.ipp
```diff
@@ -0,0 +1,825 @@
...
+           pA[55] * pB[38] + pA[63] * pB[39];
+  pC[47] = pA[7] * pB[40] + pA[15] * pB[41] + pA[23] * pB[42] +
+           pA[31] * pB[43] + pA[39] * pB[44] + pA[47] * pB[45] +
+           pA[55] * pB[46] + pA[63] * pB[47];
+  pC[55] = pA[7] * pB[48] + pA[15] * pB[49] + pA[23] * pB[50] +
+           pA[31] * pB[51] + pA[39] * pB[52] + pA[47] * pB[53] +
+           pA[55] * pB[54] + pA[63] * pB[55];
+  pC[63] = pA[7] * pB[56] + pA[15] * pB[57] + pA[23] * pB[58] +
+           pA[31] * pB[59] + pA[39] * pB[60] + pA[47] * pB[61] +
+           pA[55] * pB[62] + pA[63] * pB[63];
```
HadrienG2, line 422: You may want to cross-check the effect of this on compilation overhead (in time and memory) as compilers have a hard time churning through a lot of code. Some compilation algorithms have highly nonlinear overhead wrt input size.

Also, C compilers by default force summation to occur in the order written in the code, which is very unlikely to be optimal here. This is, in particular, very likely to hamper good code autovectorization. So you may want to disable this mechanism by compiling affected compilation units with `-ffast-math`, just to see how much the performance would change with a more optimal (according to the compiler) summation order.

Core/src/Propagator/detail/CovarianceEngine.cpp
```diff
@@ -366,8 +371,10 @@ void transportCovarianceToCurvilinear(BoundSymMatrix& boundCovariance,
 
   // Apply the actual covariance transport to get covariance of the current
   // curvilinear parameters
-  boundCovariance = fullTransportJacobian * boundCovariance *
-                    fullTransportJacobian.transpose();
+  using namespace MatrixHelpers;
+  boundCovariance =
+      multiply(fullTransportJacobian,
+               multiply(boundCovariance, transpose(fullTransportJacobian)));
```
HadrienG2, line 377: Transposes are both memory-bound and accessing data in the wrong order, so they are expensive unless successfully elided by the compiler. Which it may not do here due to the sheer amount of code that it is already diving through.

To avoid the transpose here, you can try either of two strategies:

- Implement the minimal composite operation that avoids the transpose, `multiplyByTranspose`. This is the route that Lucas took back in the day.
- Bite the bullet and implement the entire `multiply(T, multiply(U, transpose(T)))` round trip as a single composite operation, as previously suggested by @asalzburger .

Core/include/Acts/Utilities/HelpersMatrixGenerated.ipp
```diff
@@ -0,0 +1,825 @@
+//////////////////////////////////////////////////
+// This function is AUTOGENERATED. Do not edit! //
+//////////////////////////////////////////////////
+inline ActsMatrix<8, 8> multiply(const ActsMatrix<8, 1>& A,
+                                 const ActsMatrix<1, 8>& B) {
+  ActsMatrix<8, 8> C;
+  double* pC = C.data();
+  const double* pA = A.data();
+  const double* pB = B.data();
```
HadrienG2, line 9: For highest-quality vectorized code when using wider hardware vectorization (which you can try out by building affected compilation units in `-march=native` mode), you will likely need to tell the compiler that these matrices have been properly aligned on the stack by Eigen. You can do this with the `__builtin_assume_aligned` intrinsic, available in both clang and GCC, which I tend to wrap in the following way for convenience:

```c++
  template <typename T>
  const T* assume_aligned(const T* ptr, size_t alignment) {
    return static_cast<const T*>(__builtin_assume_aligned(ptr, alignment));
  }
  
  template <typename T>
  T* assume_aligned(T* ptr, size_t alignment) {
    return static_cast<T*>(__builtin_assume_aligned(ptr, alignment));
  }
  ```

Eigen should impose an alignment that matches the largest hardware vector size, so 64 on AVX-512 CPUs and 32 on AVX-256 CPUs (which is the vast majority of modern hardware).

HadrienG2 (commented):
Just to state the obvious, this is only useful as one more proof that Eigen is not the end of the performance story (in the sense that we can easily outperform it by an estimated factor of 2x without spending a lot of time optimizing). I do not see this as something we would actually want to merge for an overall 10% propagation perf improvement, but rather as a motivation to make algebra-plugins good enough to replace Eigen in hot parts of the ACTS code.

Nevertheless, for the purpose of pushing this experiment all the way to its logical conclusion, you will find below a number of easy-ish suggestions for tuning up this autogenerated code's performance by helping the compiler optimize things as much as it is able to.

HadrienG2:
Another completely unrelated family of optimizations that you could try out is to take into account known zeroes in the matrices when specializing code. From memory, there's a fair number of elements in the matrices at hand that are known to be zero, and every zero you tell the compiler about removes O(N) multiply-add operations in the matrix multiplication code. This may not seem very useful in an operation that is overall O(N^3), but with enough zeroes it can add up to non-negligible time savings if N is small as is the case here.

beomki-yeo:
I like the changes while you can also discuss with Eigen developers to add this to the Eigen repository. They do have some hard coded operations for small matrices

Core/include/Acts/Utilities/HelpersMatrixGenerated.ipp
paulgessinger: This has a negative effect. I'm guessing the copy elision fails.

PR #1124, opened 2021-12-16
refactor!: Reduce heap allocations in navigation

paulgessinger:
This PR changes the storage values in `Navigator` to use `boost::container::small_vector` for layers, sensitive and boundary surfaces. At the same time `Layer::compatibleSurface`, `TrackingVolume::combatibleBoundaries` and `TrackingVolume::compatibleLayers` are modified to return this container.

This PR contains the changes from #1119, so should be merged afterwards.

I benchmarked this with the EigenStepper **without** covariance transport, as this shouldn't be affected by this change.

On main I get:
```
  Time (mean ± σ):     14.518 s ±  0.222 s    [User: 882.433 s, System: 0.663 s]
  Range (min … max):   14.249 s … 14.836 s    5 runs
```

while on this branch I get
```
  Time (mean ± σ):     13.981 s ±  0.141 s    [User: 849.680 s, System: 0.482 s]
  Range (min … max):   13.879 s … 14.228 s    5 runs
```

for 1000 tracks per event and 20000 events each.

The measurements seem to be just outside about 2sigma, if we take the mean at face-value, it goes down by about 3%.


paulgessinger:
@asalzburger what do you think?

HadrienG2:
@paulgessinger Try re-running the benchmark with more runs. "just outside 2 sigmas" on 5 runs is not enough to claim a significant difference on benchmark timings, whose probability law is not normal.

paulgessinger:
Sure. This is what I get with 30 runs on main

```
  Time (mean ± σ):     14.444 s ±  0.183 s    [User: 878.311 s, System: 0.630 s]
  Range (min … max):   14.141 s … 14.867 s    30 runs
```
and on this PR
```
  Time (mean ± σ):     13.977 s ±  0.235 s    [User: 848.936 s, System: 0.520 s]
  Range (min … max):   13.653 s … 14.603 s    30 runs
```

The means seem roughly consistent with before. I can run again with even more samples.

Overall, this isn't a huge performance improvement, I agree. But looking at the propagation profile, at least I can't see any great singular optimization opportunities, where we could get a large increase in performance in one go. It rather seems like there's a number of smaller optimization opportunities that in total could be a net positive.

robertlangenberg:
The changes look good, can you resolve the conflicts in Layer.hpp and Layer.cpp?

robertlangenberg (approved):

PR #1125, opened 2021-12-16
fix: Add smearing options to ckf tracking example

niermann999:
The CKF tracking example crashes due to missing particle smearing options. The initial inflation option clashes with an option of the same name in the fitting options and has been renamed.


niermann999:
Or should this rather be dependent on the ```fit-initial-variance-inflation``` option and therefore the inclusion of ```Options::addFittingOptions(desc)```?

paulgessinger:
Honestly, this is a great example of why we want to remove these executable. 

If you confirm this change fixes the crash, I'm ok to merge this as is.

paulgessinger (approved):
Let's go then.

PR #1126, opened 2021-12-17
feat - Make binning explicitly configurable in TGeo example

niermann999:
This PR makes the surface binning in the TGeo Detector example explicitly configurable. In the json configuration a vector of bins for ```loc0``` and ```loc1``` is added for the negative, central and positive detector regions respectively. The binning must be set per layer, meaning that the number of layers must be known when configuring the detector building. Alternatively, a single binning entry of ```0``` can be used to trigger autobinning.

### DIFF
--- a/Examples/Detectors/TGeoDetector/include/ActsExamples/TGeoDetector/TGeoDetector.hpp
+++ b/Examples/Detectors/TGeoDetector/include/ActsExamples/TGeoDetector/TGeoDetector.hpp
@@ -11,6 +11,7 @@
 #include "Acts/Utilities/Logger.hpp"
 #include "ActsExamples/Detector/IBaseDetector.hpp"
 #include "ActsExamples/Utilities/Options.hpp"
+#include "Acts/Utilities/BinningType.hpp"
 
 #include <map>
 #include <memory>
@@ -99,6 +100,8 @@ struct TGeoDetector : public ActsExamples::IBaseDetector {
       LayerTriplet<Options::Interval> zRange;
       LayerTriplet<double> splitTolR{0};
       LayerTriplet<double> splitTolZ{0};
+      LayerTriplet<std::vector<std::pair<int, Acts::BinningType>>> binning0;
+      LayerTriplet<std::vector<std::pair<int, Acts::BinningType>>> binning1;
 
       Options::Interval binToleranceR;
       Options::Interval binTolerancePhi;
--- a/Plugins/TGeo/src/TGeoLayerBuilder.cpp
+++ b/Plugins/TGeo/src/TGeoLayerBuilder.cpp
@@ -94,18 +94,30 @@ void Acts::TGeoLayerBuilder::buildLayers(const GeometryContext& gctx,
 
   // Helper function to fill the layer
   auto fillLayer = [&](const LayerSurfaceVector lSurfaces,
-                       const LayerConfig& lCfg) -> void {
-    // Set binning by hand if nb0 > 0 and nb1 > 0
-    auto nb0 = std::get<int>(lCfg.binning0);
-    auto nb1 = std::get<int>(lCfg.binning1);
-    // Or use the binning type
-    auto nt0 = std::get<BinningType>(lCfg.binning0);
-    auto nt1 = std::get<BinningType>(lCfg.binning1);
+                       const LayerConfig& lCfg, unsigned int pl_id = 0) -> void {
+
+    int nb0 = 0, nb1 = 0, nt0 = 0, nt1 = 0;
+    if (not (lCfg.binning0.size() > pl_id)) {
+      // Has not been configured for auto-binning
+      if (lCfg.binning0.empty() or (not (std::get<int>(lCfg.binning0.at(0)) <= 0) or not(std::get<int>(lCfg.binning1.at(0)) <= 0))) {
+        ACTS_WARNING("No binning configuration found for protolayer #" << pl_id
+                     << ". Either no configuration or too few configurations were provided. Using auto-binning instead.");
+      }
+    }
+    else {
+      // Set binning by hand if nb0 > 0 and nb1 > 0
+      nb0 = std::get<int>(lCfg.binning0.at(pl_id));
+      nb1 = std::get<int>(lCfg.binning1.at(pl_id));
+      // For a binning type
+      nt0 = std::get<BinningType>(lCfg.binning0.at(pl_id));
+      nt1 = std::get<BinningType>(lCfg.binning1.at(pl_id));
+    }
 
     if (type == 0) {
       ProtoLayer pl(gctx, lSurfaces);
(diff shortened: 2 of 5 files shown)


niermann999:
Some more testing is needed, but the trackstate plot looks better:

Examples/Detectors/TGeoDetector/include/ActsExamples/TGeoDetector/TGeoDetector.hpp
```diff
@@ -99,6 +100,8 @@ struct TGeoDetector : public ActsExamples::IBaseDetector {
       LayerTriplet<Options::Interval> zRange;
       LayerTriplet<double> splitTolR{0};
       LayerTriplet<double> splitTolZ{0};
+      LayerTriplet<std::vector<std::pair<int, Acts::BinningType>>> binning0;
+      LayerTriplet<std::vector<std::pair<int, Acts::BinningType>>> binning1;
```
paulgessinger, line 104: Can you add python bindings for these two added properties? I think that should be all that's needed on the python side.

niermann999:
trackstates_r_z

paulgessinger:
Btw, I think root can actually evaluate `sqrt(g_x_hit^2 + g_y_hit^2)`.

niermann999:
> Btw, I think root can actually evaluate `sqrt(g_x_hit^2 + g_y_hit^2)`.
Good point!

Examples/Detectors/TGeoDetector/include/ActsExamples/TGeoDetector/TGeoDetector.hpp
niermann999: Like this? Examples/Python/src/Detector.cpp#L140-L141

paulgessinger: Yeah I think that should work.

Plugins/TGeo/src/TGeoLayerBuilder.cpp
```diff
@@ -94,18 +94,47 @@ void Acts::TGeoLayerBuilder::buildLayers(const GeometryContext& gctx,
...
-    auto nb1 = std::get<int>(lCfg.binning1);
-    // Or use the binning type
-    auto nt0 = std::get<BinningType>(lCfg.binning0);
-    auto nt1 = std::get<BinningType>(lCfg.binning1);
+                       const LayerConfig& lCfg,
+                       unsigned int pl_id = 0) -> void {
+    int nb0 = 0, nt0 = 0;
+    if (lCfg.binning0.size() <= pl_id) {
+      // Has not been configured for auto-binning
+      if (lCfg.binning0.empty() or (std::get<int>(lCfg.binning0.at(0)) > 0)) {
```
paulgessinger, line 102: I don't quite understand this logic. Is this trying to allow partially auto-binning the outer layers while manually binning the inner ones? I would probably just say that you either have to specify bins for all layers or use auto-binning for all of them.

Plugins/TGeo/src/TGeoLayerBuilder.cpp
```diff
@@ -94,18 +94,47 @@ void Acts::TGeoLayerBuilder::buildLayers(const GeometryContext& gctx,
...
+      // Set binning by hand if nb0 > 0 and nb1 > 0
+      nb0 = std::get<int>(lCfg.binning0.at(pl_id));
+      // For a binning type
+      nt0 = std::get<BinningType>(lCfg.binning0.at(pl_id));
+    }
+
+    int nb1 = 0, nt1 = 0;
+    if (lCfg.binning1.size() <= pl_id) {
+      // Has not been configured for auto-binning
+      if (lCfg.binning1.empty() or (std::get<int>(lCfg.binning1.at(0)) > 0)) {
```
paulgessinger, line 119: Same here.

Plugins/TGeo/src/TGeoLayerBuilder.cpp
niermann999: Well yes, if you provide too few configurations because you guessed the wrong number of protolayers, it will try to autobin the rest while issuing a warning. It was actually not meant as a feature, but I can set a misconfiguration to trigger autobinning for all layers and check for all valid configuration to not contain any zeros. Or should this become a failure outright?

paulgessinger: I would probably make this a failure, so that you can correct your number of configs. Ultimately up to you though.

timadye:
Thanks for this update. I tested it with your ITk JSON in [acts-detector-examples!11](https://gitlab.cern.ch/acts/acts-detector-examples/-/merge_requests/11) and it works perfectly for me too - just as you show above.

With the Python bindings, we should also update the `Examples/Scripts/Python/itk.py` example. That does a `TGeoDetector.create()` with the whole ITk configuration specified in Python, rather than JSON. I think it makes sense to include it with this PR, so fix and test are consistent. At the moment `itk.py` produces lots of warning messages like:
```
11:06:01    InnerPixelsL   WARNING   Incorrect binning configuration found for loc1 protolayer #0. Either no configuration or too few configurations were provided. Using auto-binning for this layer instead.
```

I guess the change needed is to add `binning0` and `binning1` settings to the `TGeoDetector.create()` call in `itk.py`? Does it also need other changes, eg. for the module splitting?

niermann999:
That's a good point. Yes, I think it needs both configurations added. The binning option is at least already registered in this PR, I will add the configuration

Plugins/TGeo/src/TGeoLayerBuilder.cpp
niermann999: Done. I left the option to autobin only one of the two loc coordinates though, since I am using it. Don't know if we are expecting problems from that?

paulgessinger: No, that sounds fine to me. Thanks @niermann999!

Examples/Scripts/Python/itk.py
```diff
@@ -97,6 +96,8 @@ def runITk(
 def buildITkGeometry(geo_dir: Path, material: bool = True):
     Volume = TGeoDetector.Config.Volume
     LayerTriplet = TGeoDetector.Config.LayerTriplet
+    equidistant = TGeoDetector.Config.BinningType(0)
+    arbritrary = TGeoDetector.Config.BinningType(1)
```
paulgessinger, line 100: Wouldn't this work?

```suggestion
    equidistant = TGeoDetector.Config.BinningType.equidistant
    arbritrary = TGeoDetector.Config.BinningType.arbitrary
```

niermann999: Sure, looks good!

paulgessinger (approved):

PR #1127, opened 2021-12-17
fix: Enable copyFrom const track proxy for MultiTrajectory

benjaminhuth:
This adds a `friend class` statement in the `MultiTrajectory` that enables `copyFrom(...)` from a `ConstTrackStateProxy` to a mutable `TrackStateProxy`.

I added this to the UnitTest, but this for now only checks if this compiles.

### DIFF
--- a/Tests/UnitTests/Core/EventData/MultiTrajectoryTests.cpp
+++ b/Tests/UnitTests/Core/EventData/MultiTrajectoryTests.cpp
@@ -868,4 +868,19 @@ BOOST_AUTO_TEST_CASE(ProxyAssignment) {
   // MultiTrajectory::TrackStateProxy tp4{tp3};
 }
 
+// Check if the copy from const does compile, assume the copy is done correctly
+BOOST_AUTO_TEST_CASE(CopyFromConst) {
+  using PM = TrackStatePropMask;
+  MultiTrajectory mj;
+
+  const auto idx_a = mj.addTrackState(PM::All);
+  const auto idx_b = mj.addTrackState(PM::All);
+
+  auto mutableProxy = mj.getTrackState(idx_a);
+  auto constProxy =
+      static_cast<const MultiTrajectory&>(mj).getTrackState(idx_b);
+
+  mutableProxy.copyFrom(constProxy);
+}
+
 BOOST_AUTO_TEST_SUITE_END()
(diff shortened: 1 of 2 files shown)


Tests/UnitTests/Core/EventData/MultiTrajectoryTests.cpp
```diff
@@ -868,4 +868,19 @@ BOOST_AUTO_TEST_CASE(ProxyAssignment) {
...
+BOOST_AUTO_TEST_CASE(CopyFromConst) {
+  using PM = TrackStatePropMask;
+  MultiTrajectory mj;
+
+  const auto idx_a = mj.addTrackState(PM::All);
+  const auto idx_b = mj.addTrackState(PM::All);
+
+  auto mutableProxy = mj.getTrackState(idx_a);
+  auto constProxy =
+      static_cast<const MultiTrajectory&>(mj).getTrackState(idx_b);
```
paulgessinger, line 881: Does a static cast do the right thing here? I'd probably created an auxiliary variable `const MultiTrajectory&` to test this.

Tests/UnitTests/Core/EventData/MultiTrajectoryTests.cpp
```diff
@@ -868,4 +868,19 @@ BOOST_AUTO_TEST_CASE(ProxyAssignment) {
   // MultiTrajectory::TrackStateProxy tp4{tp3};
```
paulgessinger, line 868: Did you check that this line still doesn't compile? (the compilation failure is what is expected).

Also did you test that `copyFrom` does **NOT** work when called on a const track state proxy?

benjaminhuth: Yes, both lines still do not compile.

I also quickly tried to find a way to somehow turn such a compile error in a checkable `constexpr bool` but that seems not to be so easy unfortunately...

Tests/UnitTests/Core/EventData/MultiTrajectoryTests.cpp
benjaminhuth: It should be equivalent, but maybe its better to be more explicit here

Tests/UnitTests/Core/EventData/MultiTrajectoryTests.cpp
paulgessinger: Ok perfect!

paulgessinger (approved):
Ok, LGTM!

PR #1129, opened 2021-12-23
fix: hash_root_file AttributeError: no field named 'to_numpy'

timadye:
With my installation
```
ROOT_HASH_CHECKS=on pytest
```
fails with errors like
```
FAILED src/Examples/Python/tests/test_examples.py::test_particle_gun - AttributeError: no field named 'to_numpy'
/cvmfs/sft.cern.ch/lcg/views/LCG_101/x86_64-centos7-gcc11-opt/lib/python3.9/site-packages/awkward/highlevel.py:1123: AttributeError
```
I'm using `LCG_101` which has Python 3.9 and `awkward` 1.0.2. It seems that `obj.to_numpy()` is not supported in this version, but `awkward.to_numpy(obj)` is.

Can someone advise whether this is likely OK for other installations?


timadye:
By failing, #1130 demonstrated that the CI (ubuntu2004) was running the hash_root_file. Since this PR passed the test, I hope we can conclude that this one does no harm. Since it fixes other environments (like mine), it should at least be a step forward.

paulgessinger:
I think if it runs on Ubuntu 20.04 in the CI, this should probably be good enough for us. On another note: I would probably not use any python dependencies from LCG. Most of the time, creating a local venv with the LCG python binary but no packages works well.

paulgessinger (approved):
🚀

PR #1130, opened 2021-12-23
fix: update root_file_hashes

timadye:
Some past update to `main` has changed the `pytest` ROOT file output such that
```
ROOT_HASH_CHECKS=on pytest
```
shows failures in `test_seeding` and `test_ckf_tracks_example`.

This update changes `root_file_hashes.txt` to match the output of the current (486d6fede, 2021-12-17) `main`.

I didn't track down what changed, or if the new output actually is any good. If anyone knows about this, then it would be good to check. Otherwise, it's useful to have a working baseline for future tests, eg. to test #1128.

Note, that I needed #1129 to get `ROOT_HASH_CHECKS=on pytest` working for me.


timadye:
So it looks like the root_file_hashes were being tested in the CI. And they seem to have been correct all along - at least when run in the CI.
When I tested on my installation I got different hashes for a few tests, which is what I attempted to fix with this PR. I'll try to understand the difference after the break.
In the meantime, since this clearly isn't a good fix, I'll close this PR.

HadrienG2:
> When I tested on my installation I got different hashes for a few tests, which is what I attempted to fix with this PR. I'll try to understand the difference after the break.

@timadye This is expected, or at least it was observed by @paulgessinger during development of the hashing tool. Since the hashes do reproduce when running in a Docker container that matches the system configuration of CI, our current understanding is that this happens due to differences in math libraries changing the last few digits of some floating-point results.

PR #1132, opened 2022-01-07
ci: Don't run builds on docs-only changes

paulgessinger:
This PR uses [a feature](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-ignoring-paths) of GitHub actions to skip our C++ builds in case **only** files in `docs/` change. This requires moving the doc CI job to a separate workflow.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -2,7 +2,11 @@ name: Builds
 
 on:
   push:
+    paths-ignore:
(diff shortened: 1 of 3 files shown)


paulgessinger:
Ah, I forgot that we can't conditionally require CI jobs to pass... which means docs-only PRs would be red because the build jobs never run.

PR #1133, opened 2022-01-10
refactor: Move definitions of two functions to .cpp file

benjaminhuth:
This PR moves the definitions of two functions in the `/Tests/CommonHelpers` directory to their `.cpp` file. 

I did this because I tried to add to more than one source file to a unit test, and then got linker errors...


paulgessinger (approved):
This could probably have been solved by adding `inline` to the functions, but moving them to `.cpp` if they don't have to be in a header is preferable I think.

PR #1134, opened 2022-01-11
feat: Gaussian Sum Filter

benjaminhuth:
Finally this is the PR which contains the GSF algorithm. It consists of the files necessary in `Core` and two `UnitTests`. There are no example executables or scripts contained, these will be part of a future PR.

The GSF passes all unit tests which are also used for the `KalmanFitter`, except for one, which does not makes sense for the GSF. The other test checks the component-merging and the KL-mixture-reduction. 

Probably there should be added some more tests, but for the start I think this is ready for a first review.

This PR depends on #1110 (The `MultiEigenStepperLoop`)


Core/include/Acts/TrackFitting/detail/KLMixtureReduction.hpp
```diff
@@ -0,0 +1,203 @@
...
+
+    // Columns
+    for (auto i = n + 1; i < m_N; ++i) {
+      m_data[(i - 1) * i / 2 + n] = value;
+    }
+  }
+
+  auto minDistancePair() const {
+    // TODO Eigen minCoeff does not work for some reason??
+    // return m_mapToPair.at(m_data.minCoeff());
```
asalzburger, line 131: interesting, what's the actual problem with the `minCoeff` call?

Core/include/Acts/Utilities/Overload.hpp
```diff
@@ -0,0 +1,24 @@
...
+//
+// Copyright (C) 2021 CERN for the benefit of the Acts project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+namespace Acts {
+
+/// TODO This probably belongs to Acts/Core/Utilities
```
asalzburger, line 11: Yes, it does.

asalzburger (changes requested):
My first set of comments, I guess we will have a quick chat about a few points that may need clarification.

Core/include/Acts/TrackFitting/GaussianSumFitter.hpp
```diff
@@ -0,0 +1,503 @@
...
+#include "Acts/TrackFitting/detail/GsfActor.hpp"
+
+#include <fstream>
+
+#define RETURN_ERROR_OR_ABORT_FIT(error) \
+  if (options.abortOnError) {            \
+    std::abort();                        \
+  } else {                               \
+    return error;                        \
+  }
```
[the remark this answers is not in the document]
benjaminhuth: Ah yeah, I think that can go away, seems like I forgot to clean this up in the end...

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -0,0 +1,738 @@
...
+    return;                                        \
+  }
+
+namespace Acts {
+namespace detail {
+
+struct GsfResult {
+  /// The multi-trajectory which stores the graph of components
+  MultiTrajectory fittedStates;
+  std::map<size_t, ActsScalar> weightsOfStates;
```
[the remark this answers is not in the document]
benjaminhuth: This was only meant to be a workaround. But maybe I change this with a later PR, to not further blow up this one...

Core/include/Acts/TrackFitting/GaussianSumFitter.hpp
```diff
@@ -0,0 +1,503 @@
...
+                  proxy.referenceSurface().geometryId());
+            }
+
+            result.measurementStates++;
+            result.processedStates++;
+          };
+
+      bwdPropOptions.direction = gsfBackward;
+
+      // TODO somehow this proagation fails if we target the first
```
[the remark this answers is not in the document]
benjaminhuth: follow-up issue

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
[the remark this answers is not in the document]
benjaminhuth: follow up issue

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -0,0 +1,738 @@
...
+
+    /// Bethe Heitler Approximator pointer. The fitter holds the approximator
+    /// instance TODO if we somehow could initialize a reference here...
+    const detail::BHApprox* bethe_heitler_approx = nullptr;
+
+    /// Number of processed states
+    std::size_t processedStates = 0;
+
+    /// Allow to configure surfaces to skip
+    std::set<GeometryIdentifier> surfacesToSkip;
```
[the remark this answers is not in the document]
benjaminhuth: better comment

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -0,0 +1,738 @@
...
+        result.missedActiveSurfaces.push_back(&surface);
+      }
+
+      // Early return if nothing happens
+      if (not haveMaterial && not haveMeasurement) {
+        return;
+      }
+
+      // Create Cache
+      thread_local std::vector<ComponentCache> componentCache;
```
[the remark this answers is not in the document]
benjaminhuth: remove `thread_local`

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -0,0 +1,738 @@
...
+      }
+      // If we have no material, just compute the bound state
+      else {
+        detail::extractComponents(state, stepper, result.parentTips,
+                                  detail::ComponentForwarder{},
+                                  m_cfg.doCovTransport, componentCache);
+      }
+
+      // If we have a measuerement, perform a kalman update
+      if (haveMeasurement) {
```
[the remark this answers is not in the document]
benjaminhuth: -> kalman before splitting, split factor consider afterwards

Core/include/Acts/TrackFitting/detail/GsfUtils.hpp
```diff
@@ -0,0 +1,331 @@
...
+    const auto &logger = state.options.logger;
+    const auto &surface = *state.navigation.currentSurface;
+    const auto p_prev = old_bound.absoluteMomentum();
+
+    // Evaluate material slab
+    auto slab = surface.surfaceMaterial()->materialSlab(
+        old_bound.position(state.stepping.geoContext), state.stepping.navDir,
+        MaterialUpdateStage::fullUpdate);
+
+    //     auto pathCorrection = surface.pathCorrection(
```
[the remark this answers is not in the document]
benjaminhuth: -> remove coment

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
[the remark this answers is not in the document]
benjaminhuth: Okay, I removed this mechanism and replaced it by using the `std::function` below, which is also used to initialize the result-object of the `GsfActor`.

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -0,0 +1,738 @@
...
+    /// When to discard components
+    double weightCutoff = 1.0e-4;
+
+    /// A not so nice workaround to get the first backward state in the
+    /// MultiTrajectory for the DirectNavigator
+    std::function<void(result_type&, const LoggerWrapper&)>
+        multiTrajectoryInitializer;
+
+    /// We can disable component splitting for debugging or so
+    bool applyMaterialEffects = true;
```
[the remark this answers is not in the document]
benjaminhuth: I renamed the option and removed the above option to disable `energyLoss` since that makes no real sense for the Gsf.

asalzburger:
@benjaminhuth - is that ready for another round of review?

benjaminhuth:
not quite yet, but I'm close. I will notify you when its ready (probably tomorrow).

benjaminhuth:
Some of the current build errors are caused by `boost::tuple` which does not support structured bindings (these `auto [a,b] = tuple` things) until Boost 1.74.0
@asalzburger Probably I should rewrite this, since it is only syntactic sugar, right? Or would it be also possible to bump the required boost version?

Core/include/Acts/TrackFitting/GaussianSumFitter.hpp
```diff
@@ -0,0 +1,495 @@
...
+
+  PropagatorPlainOptions propagatorPlainOptions;
+
+  const Surface* referenceSurface = nullptr;
+
+  bool abortOnError = true;
+
+  std::size_t maxComponents = 4;
+
+  bool disableAllMaterialHandling = false;
```
asalzburger, line 60: I like that way more.

Core/include/Acts/TrackFitting/GaussianSumFitter.hpp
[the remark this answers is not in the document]
asalzburger: Sounds good.

Core/include/Acts/TrackFitting/GaussianSumFitter.hpp
```diff
@@ -0,0 +1,503 @@
...
+
+    if (bwdGsfResult.processedStates == 0) {
+      RETURN_ERROR_OR_ABORT_FIT(GsfError::NoStatesCreated);
+    }
+
+    ////////////////////////////////////
+    // Smooth and create Kalman Result
+    ////////////////////////////////////
+    ACTS_VERBOSE("Gsf: Do smoothing");
+    ACTS_VERBOSE("- Fwd measurement states: " << fwdGsfResult.measurementStates
```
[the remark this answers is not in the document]
asalzburger: We discussed that.

Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -517,61 +518,21 @@ class KalmanFitter {
...
-        // now get track state proxy back
-        auto trackStateProxy =
-            result.fittedStates.getTrackState(result.lastTrackIndex);
-
-        trackStateProxy.setReferenceSurface(surface->getSharedPtr());
-
-        // assign the source link to the track state
-        trackStateProxy.setUncalibrated(sourcelink_it->second);
+        // do the kalman update
+        auto trackStateProxyRes = detail::kalmanHandleMeasurement(
```
asalzburger, line 522: Nice, I like that, re-using the code in the both fitter options.

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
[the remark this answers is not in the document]
asalzburger: Thanks.

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -0,0 +1,728 @@
...
+    /// Maximum number of components which the GSF should handle
+    std::size_t maxComponents = 16;
+
+    /// Input measurements
+    std::map<GeometryIdentifier, std::reference_wrapper<const SourceLink>>
+        inputMeasurements;
+
+    /// Bethe Heitler Approximator pointer. The fitter holds the approximator
+    /// instance TODO if we somehow could initialize a reference here...
+    const bethe_heitler_approx_t* bethe_heitler_approx = nullptr;
```
asalzburger, line 88: Could that be a case we use Paul's `Delegate<>()` class?

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -0,0 +1,738 @@
...
+      // Check what we have on this surface
+      const auto found_source_link =
+          m_cfg.inputMeasurements.find(surface.geometryId());
+      const bool haveMaterial =
+          state.navigation.currentSurface->surfaceMaterial() &&
+          m_cfg.applyMaterialEffects;
+      const bool haveMeasurement =
+          found_source_link != m_cfg.inputMeasurements.end();
+
+      ACTS_VERBOSE(std::boolalpha << "haveMaterial " << haveMaterial
```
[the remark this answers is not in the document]
asalzburger: Thanks, learned something.

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -0,0 +1,728 @@
...
+    const auto final_cmp_number = std::min(
+        static_cast<std::size_t>(stepper.maxComponents), m_cfg.maxComponents);
+
+    auto proj = [](auto& a) -> decltype(auto) { return std::get<0>(a); };
+
+    // We must differ between surface types, since there can be different
+    // local coordinates
+    // TODO add other surface types
+    switch (surface.type()) {
+      case Surface::Cylinder: {
```
asalzburger, line 427: This has been a long-standing problem, for the new geometry, I will add an option to approximate cylinders by multiple surfaces as well ...

Core/include/Acts/TrackFitting/detail/GsfSmoothing.hpp
```diff
@@ -0,0 +1,274 @@
...
+    bwdTips.erase(std::unique(bwdTips.begin(), bwdTips.end()), bwdTips.end());
+
+    auto invalid_it = std::find(bwdTips.begin(), bwdTips.end(),
+                                std::numeric_limits<uint16_t>::max());
+    if (invalid_it != bwdTips.end()) {
+      bwdTips.erase(invalid_it);
+    }
+  };
+
+  sort_unique_validate_bwd_tips();
```
asalzburger, line 146: That's not entirely according to our coding standards, but I don't mind too much.

asalzburger (approved):
Let's get this in like this, such that we can start validating, there are few follow-ups to do ...

Core/include/Acts/TrackFitting/detail/GsfSmoothing.hpp
paulgessinger: Should be a simple fix though. Up to you @benjaminhuth

cmake/ActsCompilerOptions.cmake
```diff
@@ -4,7 +4,7 @@ if (NOT CMAKE_BUILD_TYPE)
   message(STATUS "Setting default build type: ${CMAKE_BUILD_TYPE}")
 endif() 
 
-set(ACTS_CXX_FLAGS "-Wall -Wextra -Wpedantic -Wshadow -Wunused-local-typedefs")
+set(ACTS_CXX_FLAGS "-Wall -Wextra -Wpedantic -Wshadow -Wno-unused-local-typedefs")
```
paulgessinger, line 7: Is there a reason we need to disable this warning?

Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -517,61 +518,21 @@ class KalmanFitter {
...
-            return updateRes.error();
-          }
-          // Set the measurement type flag
-          typeFlags.set(TrackStateFlag::MeasurementFlag);
+        // Update the stepper if it is not an outlier
+        if (trackStateProxy.typeFlags().test(
+                Acts::TrackStateFlag::MeasurementFlag)) {
           // Update the stepping state with filtered parameters
           ACTS_VERBOSE("Filtering step successful, updated parameters are : \n"
                        << trackStateProxy.filtered().transpose());
```
paulgessinger, line 538: This log message is then not shared via `kalmanHandleMeasurement`?

Core/include/Acts/TrackFitting/detail/GsfSmoothing.hpp
benjaminhuth: You mean the not-CamelCase or the use of the lambda?

Core/include/Acts/Utilities/Zip.hpp
```diff
@@ -0,0 +1,50 @@
...
+  using ItB = decltype(rb.begin());
+
+  struct It {
+    ItA a;
+    ItB b;
+
+    using reference =
+        std::tuple<typename ItA::reference, typename ItB::reference>;
+
+    // clang-format off
```
paulgessinger, line 33: Why?

cmake/ActsCompilerOptions.cmake
benjaminhuth: It gives false positives at `MultiEigenStepperLoop.hpp:452`, where I have a local struct with typedefs which than is returned from the function...

Core/include/Acts/TrackFitting/detail/KalmanUpdateHelpers.hpp
```diff
@@ -0,0 +1,185 @@
...
+                             const SourceLink &source_link,
+                             MultiTrajectory &fittedStates,
+                             const size_t lastTrackIndex, bool doCovTransport)
+    -> Result<MultiTrajectory::TrackStateProxy> {
+  const auto &logger = state.options.logger;
+
+  // Bind the transported state to the current surface
+  auto res = stepper.boundState(state.stepping, surface, doCovTransport);
+  if (!res.ok()) {
+    return res.error();
```
paulgessinger, line 48: Log message?

paulgessinger:
Yeah looks great I think.

I guess changes to the KF would show up in the repro tests and maybe physics monitoring if they changed the KF output, so should be ok.

Core/include/Acts/TrackFitting/detail/GsfSmoothing.hpp
paulgessinger: `snake_case`, yes. The lambda is fine!

Core/include/Acts/Utilities/Zip.hpp
benjaminhuth: yeah, so this originiates from the `std::iterator_traits`, which defines some typedefs that iterators should have (https://en.cppreference.com/w/cpp/iterator/iterator_traits)
So the thing is, an iterator can represent a proxy-class (which we have with the `ComponentProxy`) with reference-semantics, than this type is not a real reference type, and we want to pass it by-value (otherwise we would probably get an "cannot bind reference to tempory"-like error)
Hope this makes sense^^

Probably one should also define the other typedefs required by the iterator-traits, but this is not necessary for a range-for loop...

asalzburger:
Benjamin, can you please rebase?

Core/include/Acts/Utilities/Zip.hpp
paulgessinger: Sure, I got the typedefs, but what about the format off? 😅

benjaminhuth: Ah sorry I missed that its related to the clang-format...
Yeah I somehow wanted these function to fit in one line each, but since there is no real reason for this, I changed it

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
benjaminhuth: Hmm not sure if  this is that easy, since the return type of the `.mixture()` function depends on the `NComponents` integer-template-parameter (it is a `std::array<..., NComponents>`), and the `Delegate` needs a stable return-type, right?
I also don't think this template instatiation is expensive...

Core/include/Acts/TrackFitting/KalmanFitter.hpp
benjaminhuth: I thought it might be better leave the decision what is logged to the caller of this, but this obviously could be changed...

asalzburger:
There's still a format failure for python which we need to fix, but then I think this should go in.

paulgessinger:
@asalzburger format failure seems to be a technical issue, see #1212

asalzburger (approved):

Issue #1135, opened 2022-01-17
Autodiff plugin is not compatible with current versions of autodiff


HadrienG2:
Trying to build the ACTS autodiff plugin against a current version of autodiff will fail because the autodiff 0.6.x release series moved around several headers, including the autodiff/forward.hpp header that the ACTS autodiff plugin relies on.

cc @benjaminhuth

benjaminhuth:
I did not follow the developements in the `autodiff` library, but I'll work out a PR to make it compatible again...

PR #1136, opened 2022-01-17
build: upgrade autodiff version

benjaminhuth:
This PR upgrades the bundled `autodiff` version to 0.6.4, and sets the minimum required version to 0.6 (which introduced breaking changes).

This should resolve issue #1135


HadrienG2 (approved):
Thanks for the prompt fix!

PR #811, opened 2021-05-21
refactor: Material surface handling in CKF

Corentin-Allaire:
Implement a couple of changes to the CKF: 
- Properly check if material is present on the surface before turning on the material flag.

- Doesn't require anymore the present of material on a surface for a Hole to be added.

- Replace the TrackTips by a lastTrackIndices and a lastMeasurementIndices (similarly to what was implemented in the fitter in #729) this allow us to avoid doing the filtering on the many holes/passive surface that could be after the mast measurement.

PR #1137, opened 2022-01-17
fix: CKF flags material surfaces as holes

paulgessinger:
I think after #811, the CKF started to flag all passive material surfaces as holes in the track tip, because both passive surfaces and sensitive surfaces without measurements started to be handled by `addNonSourceLinkState`.

This PR changes this to only flag states as holes if the current surface is in fact sensitive.

/cc @asalzburger @Corentin-Allaire @timadye @noemina


Corentin-Allaire (approved):
Great to see ! But this mean the large number of holes shouldn't affect the performances since it is just mistagged material right ?

paulgessinger:
Exactly.

paulgessinger:
Updated with new output file hashes. I think the changes make sense given reclassification of holes.

Corentin-Allaire (approved):

PR #1055, opened 2021-10-27
WIP: feat: SP builders in Core

toyamaza:
This PR moves SingleHitSpacePointBuilder and DoubleHitSpacePointBuilder from Plugins/Digitization to Core in order to use them in athena.

The previous SP builders used PlanarModuleCluster as input. The segmentation of the strip is needed for the double space point builder, so I'm adding a simple ModuleCluster class that contains Measurement and segment in Examples, but it's open for discussion.

The calculation of the variances in R and Z is introduced as they are required in seeding.

The strip SP builder is not supported in ActsExamples yet. I think the strip digitization needs to be added.

PR #1138, opened 2022-01-19
refactor: Move Digitization from Plugins to Core

toyamaza:
This PR moves the SP builders and associated files from Plugins/Digitization to Core. 
PR #1055  is split into two: reorganization of files (this PR) and updating the SP builder (will be opened shortly).
The whole Plugins/Digitization is moved to Core because the current SP builders depend on all files in Digitization, such as the DigitizationModule, PlannarModuleCluster, etc.


paulgessinger (commented):
Could you update the [getting started doc page](https://github.com/toyamaza/acts/blob/main/docs/getting_started.md) to remove the now obsolete option `ACTS_BUILD_PLUGIN_DIGITIZATION`?

Also, I believe the following piece of code would need updating:

https://github.com/acts-project/acts/blob/3d17c325f23d2593a65b1e137e9ec5073664b95d/Examples/Python/CMakeLists.txt#L86-L91

I guess it would be ok to reduce the if to only check for the existence of the identification plugin.

toyamaza:
Thanks @paulgessinger for your comment. I removed `ACTS_BUILD_PLUGIN_DIGITIZATION`.

paulgessinger (approved):

PR #1141, opened 2022-01-20
fix: Delegate checks exact compatibility of signatures

### LINKED ISSUE #1140: bug: Delegate accepts approximate signatures
@Corentin-Allaire had to find out that 

```cpp
using Smoother = Delegate<Result<void>(
    const GeometryContext&, MultiTrajectory&, size_t, LoggerWrapper)>;
```

actually accepts a function with a signature of 

```cpp
Result<void>(const GeometryContext&, MultiTrajectory, size_t, LoggerWrapper)
```

(`MultiTrajectory` by value) which is obviously a problem.

Ideally, Delegate should reject this connection at compile time.

paulgessinger:
As described in #1140, signatures are not checked for **exact** compatibility by `Delegate`.

This PR introduces an additional check where it will try to assign a given callable to an expected function pointer type generated from the signature given to `Delegate`. This assignment only succeeds if the signatures match exactly. This is enforced by a `static_assert`.

Fixes #1140

/cc @Corentin-Allaire


Corentin-Allaire (approved):
Looks good to me ! It is great that we will be able avoid thing similar to issue #1140 in future

PR #1142, opened 2022-01-20
fix: Prevent FPE in AMVF if covariance is not invertible

paulgessinger:
It's unclear to me how this covariance can end up not being invertible. I briefly discussed with @baschlag whether returning `true`, i.e. treating the vertex as *merged* is correct, and we think it's safer to do so, and therefore discard the vertex one level up the call chain.

I'm going to cherry pick this PR into branch `release/v15.x` to create a bugfix tag for the `v15` release.


baschlag:
Hi, as Paul mentioned I also think that treating these vertices as merged and thus discard them is the safest way to go here.

baschlag (approved):

PR #1084, opened 2021-11-25
feat: ITK SeedFinder integration 

LuisFelipeCoelho:
This PR introduces several changes to the `SeedFinder` algorithm:

1. Introduced a variable radius range (`rMinMiddleSP`, `rMaxMiddleSP`) for the middle SP:
- Obtained by a floor approximation to an even number of the minimum and maximum SP radius found plus a ΔR (`deltaRMiddleSPRange`) defined by the user
- An Extent `rRangeSPExtent` is used to get the min and max values
- This cut is only applied if a boolean `useVariableMiddleSPRange` defined by the user is true
- If `useVariableMiddleSPRange` is set to false, the vector `rRangeMiddleSP` can be used to define a fixed r range. If the vector is empty the cuts won't be applied.

2. Check if middle SP is on the first or last disk with `zMax` and `zMin`

3. Introduced `seedConfirmation` step that requires a minimum number of top SP depending on whether the middle SP is in the central or forward region:
- If the boolean `seedConfirmation` is set to true the cut will be applied

4. Changed the order of the loops over the top and bottom space points:
- Looping over the top SP first. The algorithm doesn’t need to unnecessarily loop over the bottom SP if the minimum number of top SP is not greater than `nTopSeedConf` when `seedConfirmation` is set to true

5. Introduced `deltaRMinBottomSP`, `deltaRMaxBottomSP`, `deltaRMinTopSP` and `deltaRMaxTopSP`:
- ΔR minimum and maximum between middle and top/bottom SP can now be defined separately, as opposed to having the same value for all

6. Added a cut on A and B coefficients for the top and bottom SP

**During SP comparison:**

7. `linCircleBottom` and `linCircleTop` are sorted by `cotTheta` of the bottom and top space points inside `transformCoordinates`

8. Added the `t0` index

9. In the cut on the compatibility between the r-z slope of the two seed segments, `deltaCotTheta2 - error2 > 0` was changed to `deltaCotTheta2 - error2 > sigmaSquaredScatteringMinPt`, where `sigmaSquaredScatteringMinPt` is an approximate worst case multiple scattering term assuming the lowest pT we allow and the estimated theta angle
- `dCotThetaMinusError2 > scatteringInRegion2` cut **removed**
- if `deltaCotTheta < 0` → break
- else → `t0=t+1` and continue 

[...]

PR #1143, opened 2022-01-21
fix: Bug when sorting the SP in SeedFinder

LuisFelipeCoelho:
In PR #1084 I mentioned that we were seeing a degradation in tracking efficiency when sorting the SP in cotangent of theta in the `SeedFinder`. However I realised that I was sorting only the `linCircleVec` vector in `SeedFinder` and the `compatBottomSP` and `compatTopSP` vectors should also be sorted. 

This PR solves the bug and now the efficiency and the output should be the same as without the sorting.

@paulgessinger @robertlangenberg @noemina

### DIFF
--- a/Core/include/Acts/Seeding/SeedFinderUtils.ipp
+++ b/Core/include/Acts/Seeding/SeedFinderUtils.ipp
@@ -96,6 +96,11 @@ void transformCoordinates(
   }
   // sort the SP in order of cotTheta
   if (enableCutsForSortedSP) {
+    std::sort(vec.begin(), vec.end(),
+              [](const InternalSpacePoint<external_spacepoint_t>* a,
+                 const InternalSpacePoint<external_spacepoint_t>* b) -> bool {
+                return (a->cotTheta() < b->cotTheta());
+              });
     std::sort(linCircleVec.begin(), linCircleVec.end(),
               [](const LinCircle& a, const LinCircle& b) -> bool {
                 return (a.cotTheta < b.cotTheta);
--- a/Core/include/Acts/Seeding/InternalSpacePoint.hpp
+++ b/Core/include/Acts/Seeding/InternalSpacePoint.hpp
@@ -41,15 +41,18 @@ class InternalSpacePoint {
   const float& varianceR() const { return m_varianceR; }
   const float& varianceZ() const { return m_varianceZ; }
   const SpacePoint& sp() const { return m_sp; }
+  const float& cotTheta() const { return m_cotTheta; }
+  void setCotTheta(float& cotTheta) const { m_cotTheta = cotTheta; }
 
  protected:
-  float m_x;               // x-coordinate in beam system coordinates
-  float m_y;               // y-coordinate in beam system coordinates
-  float m_z;               // z-coordinate in beam system coordinetes
-  float m_r;               // radius       in beam system coordinates
-  float m_varianceR;       //
-  float m_varianceZ;       //
-  const SpacePoint& m_sp;  // external space point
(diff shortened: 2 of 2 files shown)


paulgessinger:
Does that mean we could get rid of `enableCutsForSortedSP` again?

LuisFelipeCoelho:
I realised that the output is still not the same, so I will have to investigate this a bit further before removing the `enableCutsForSortedSP`

LuisFelipeCoelho:
I believe that now we can get rid of `enableCutsForSortedS`, or we can keep it and add a python-level test that runs the seeding with the `cotTheta` sorting.

Core/include/Acts/Seeding/SeedFinderUtils.ipp
```diff
@@ -92,10 +92,21 @@ void transformCoordinates(
...
     linCircleVec.push_back(l);
   }
   // sort the SP in order of cotTheta
   if (enableCutsForSortedSP) {
+    std::sort(vec.begin(), vec.end(),
+              [](const InternalSpacePoint<external_spacepoint_t>* a,
+                 const InternalSpacePoint<external_spacepoint_t>* b) -> bool {
+                return (a->cotTheta() < b->cotTheta());
+              });
     std::sort(linCircleVec.begin(), linCircleVec.end(),
```
noemina, line 110: Do we need to sort both the space point vector and the `LinCircleVec`? Can you check if the `LinCircleVec` gets sorted automatically if you sort the space point vector before looping on its content?

LuisFelipeCoelho: I am filling the space points with `cotTheta` during the loop, and using these values to sort the space point vector. In order to sort the space point vector before the loop I will need to calculate the values of `cotTheta` before the loop, as well. Or maybe there is a better way to sort these vectors?

LuisFelipeCoelho:
One alternative would be to create an index vector and sort this vector based on `cotTheta` values from `linCircleVec`. Then create a new vector of pointers `newVec`, fill it with the new order from the index vector, and replace `vec` with the new one:

```
std::vector<size_t> idx(vec.size());
std::iota(idx.begin(), idx.end(), 0);

std::sort(idx.begin(), idx.end(),
	[&linCircleVec](size_t i1, size_t i2) {
	return linCircleVec[i1].cotTheta < linCircleVec[i2].cotTheta;
});

std::vector<const InternalSpacePoint<external_spacepoint_t>*> newVec;
for (size_t i = 0; i<vec.size();i++) {
	newVec.push_back(vec[idx[i]]);
}
vec = newVec;
```

@paulgessinger @robertlangenberg @noemina

paulgessinger:
Can't we just have `transformCoordinates` take the spacepoints as non-const?

Right now we have 

```cpp
void transformCoordinates(
    std::vector<const InternalSpacePoint<external_spacepoint_t>*>& vec,
    const InternalSpacePoint<external_spacepoint_t>& spM, bool bottom,
bool enableCutsForSortedSP, std::vector<LinCircle>& linCircleVec);
```

Does `transformCoordinates` get executed on the same `vec` concurrently? Otherwise, maybe this could be changed to something like `std::vector<InternalSpacePoint<external_spacepoint_t>*>&`?

noemina:
> Does `transformCoordinates` get executed on the same `vec` concurrently? Otherwise, maybe this could be changed to something like `std::vector<InternalSpacePoint<external_spacepoint_t>*>&`?

I don't think this should happen. @robertlangenberg, can you please confirm it?

noemina (approved):
Changes look good to me. Approving.

PR #1147, opened 2022-02-04
feat: Scale backward filtering covariance

Corentin-Allaire:
Add an option to scale up the covariance matrix before performing the backward filtering for the Kalman fitter. Scaling this covariance might be necessary to obtain a reasonable smoothed covariance at the end of the backward filtering and shouldn't affect the performance.


Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -152,6 +153,10 @@ struct KalmanFitterOptions {
   /// ReverseFilteringLogic
   bool reversedFiltering = false;
 
+  /// Factor by which the covariance of the input of the reversed filtering is
+  /// scaled
+  double reversedFilteringCovarianceScaling = 1.0;
```
paulgessinger, line 158: Can we make this absolutely clear that it's only used in case we actually do the backward propagation.

PR #1148, opened 2022-02-07
feat:add clusters to trackstatewriter

niermann999:
Debug my cluster writing


paulgessinger:
Right. The cluster sizes are of course not the same as the measurement resolutions (which I believe are filled in the `hits` branches). Do we need the actual cluster sizes or would the measurement resolutions do?

niermann999:
Well, the Athena sample uses cluster sizes (at least the variables are named accordingly). The ```m_err_x_hit``` hit variables seem to be the covariances of the resolutions?

niermann999:
What I need are the projected cluster sizes of every hit in a track, since these are compared directly. And then the resolutions are plotted per projected cluster size, too.

PR #1150, opened 2022-02-08
perf: replace `std::pow(x, 0.25)` with std::sqrt(std::sqrt(x))

paulgessinger:
`powf64` is relatively slow. This change improves the performance of `ActsBenchmarkEigenStepper` by about 10%, the performance impact on the more real-world propagation with navigation (as in the propagation example with the generic detector), seems negligible.

Overall I think it's probably still worth adding.


HadrienG2 (approved):
It's a bit puzzling to me that we have two almost but not quite identical versions of the step size adaptation formula. But I'm happy to see this optimization land.

paulgessinger:
Copy paste error? 

Also, I'm a bit surprised that the root file output changes. I'd have thought that a slight numerical change in the step size scaling would have a negligible effect on the output...

asalzburger:
> It's a bit puzzling to me that we have two almost but not quite identical versions of the step size adaptation formula. But I'm happy to see this optimization land.

This doesn't surprise me, I think this is really a numerical difference from slight change of step size.

benjaminhuth:
Out of interest I've played a bit around with different versions of this, and I found that 

```c++
std::sqrt(std::sqrt(static_cast<float>(x)));
```
is about twice as fast as using the `double` version. As I understand we do not require large precision here, so that could speed this even up a bit?

(The values I measured by just computing the result out of many `sqrts(sqrts(...))` in a row, but without vectorization and with -O2)

HadrienG2:
@benjaminhuth This throughput difference indeed expected, https://www.agner.org/optimize/instruction_tables.pdf states that on modern CPUs the underlying `SQRTSS` and `SQRTSD` hardware instructions exhibit a 2x throughput difference in the worst-case scenario. I also agree that some computations here could likely be safely moved to single precision.

paulgessinger:
I'd probably say using this static cast to `float` is a good idea. I don't think we care about numerical precision in these cases.

stephenswat:
Depending on the frequency at which the clamping condition between 0.25 and 4.0 occurs, it might also be beneficial to rewrite the clamp explicitly:

```cpp
float r = state.options.tolerance / std::abs(2. * error_estimate);

if (r <= 0.00390625f) { // 0.25^4
    stepSizeScaling = 0.25;
else if (r >= 256.f) { // 4.0^4
    stepSizeScaling = 4.0;
} else {
    stepSizeScaling = std::sqrt(std::sqrt(r));
}
```

Hard to say if you'll benefit without knowing how often this fires, though. :man_shrugging:

paulgessinger:
@stephenswat that could be an improvement, but might indeed be overkill.

asalzburger:
This has conflicts now because of the updated file I suppose ...

paulgessinger:
Updated the hashes again. Let's see.

paulgessinger:
This is green now. Do we merge (/ can you approve)? @benjaminhuth @HadrienG2 @stephenswat ?

stephenswat (approved):

PR #1152, opened 2022-02-09
fix: Reset the stepper Jacobian in the KF smoothing

Corentin-Allaire:
This PR solves the low Pt issue that was occurring when using the KF and the smoothing. The Jacobian of the stepper is now reset before performing the last propagation from the last state to the perigee after the smoothing. 

This is the equivalent of replacing the `stepper.update` by a `stepper.resetState` but this implementation allows us to avoid computing the free parameters twice.


Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -1111,12 +1111,20 @@ class CombinatorialKalmanFilter {
                        firstCreatedState.smoothedCovariance(),
                        firstCreatedState.referenceSurface());
         reverseDirection = (firstIntersection.intersection.pathLength < 0);
+        // Reinitialize the stepping jacobian
+        state.stepping.jacToGlobal =
```
XiaocongAi, line 1115: After a second thought, I'm not sure if we really need to reset the `jabToGlobal` after calling the stepper `update` as the `jacToGlobal` is already reset there.

Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -965,12 +965,20 @@ class KalmanFitter {
                        firstCreatedState.smoothedCovariance(),
                        firstCreatedState.referenceSurface());
         reverseDirection = (firstIntersection.intersection.pathLength < 0);
+        // Reinitialize the stepping jacobian
+        state.stepping.jacToGlobal =
```
XiaocongAi, line 969: Same here. Do we need to reset the `jacToGlobal` here?

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
Corentin-Allaire: Your right, it is already reset in the `update` I will remove it

XiaocongAi (approved):

paulgessinger:
Root output hashes are still changed. Can you look at that @Corentin-Allaire?

Corentin-Allaire:
@paulgessinger I have updated the hashes.

Corentin-Allaire:
@XiaocongAi could you reapprove ? I had to update the hashes of the python test

PR #1155, opened 2022-02-10
fix: Updating itk description in python, fixes issue #1149

### LINKED ISSUE #1149: buildITkGeometry in itk.py doesn't produce hits
I used [`buildITkGeometry`](https://github.com/acts-project/acts/blob/main/Examples/Scripts/Python/itk.py#L97) to create a `TGeoDetector` geometry for the ATLAS ITk. This looks OK, except when used by Fatras, it doesn't generate any hits.

If instead, I specify the geometry definition with
```
TGeoDetector.create(jsonFile="acts-detector-examples/atlas/itk-hgtd/tgeo-atlas-itk-hgtd.json", ...)
```
hits are generated without problem.

I'll continue to try to find a difference in the returned `detector` and `trackingGeometry` objects from these two methods, but thought it worth flagging up now in case someone has a better idea.

noemina:
Improving the description of the ITk geometry parameters in the python file.
This fixes as well the issue #1149, caused by the wrong configuration of the HGTD extends.

Fixes #1149

### DIFF
--- a/Examples/Scripts/Python/itk.py
+++ b/Examples/Scripts/Python/itk.py
@@ -125,11 +125,11 @@ def buildITkGeometry(geo_dir: Path, material: bool = True):
         volumes=[
             Volume(
                 name="InnerPixels",
-                layers=LayerTriplet(True),
-                subVolumeName=LayerTriplet("Pixel::Pixel"),
                 binToleranceR=(5 * u.mm, 5 * u.mm),
                 binToleranceZ=(5 * u.mm, 5 * u.mm),
                 binTolerancePhi=(0.025 * u.mm, 0.025 * u.mm),
+                layers=LayerTriplet(True),
+                subVolumeName=LayerTriplet("Pixel::Pixel"),
                 sensitiveNames=LayerTriplet(["Pixel::siLog"]),
                 sensitiveAxes=LayerTriplet("YZX"),
                 rRange=LayerTriplet((0 * u.mm, 135 * u.mm)),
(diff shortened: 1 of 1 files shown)


Examples/Scripts/Python/itk.py
```diff
@@ -258,24 +266,33 @@ def buildITkGeometry(geo_dir: Path, material: bool = True):
...
                 },
             ),
             Volume(
                 name="HGTD",
-                layers=LayerTriplet(positive=True, central=False, negative=True),
-                subVolumeName=LayerTriplet("HGTD::HGTD"),
                 binToleranceR=(15 * u.mm, 15 * u.mm),
-                binToleranceZ=(15 * u.mm, 15 * u.mm),
+                binToleranceZ=(5 * u.mm, 5 * u.mm),
                 binTolerancePhi=(0.025 * u.mm, 0.025 * u.mm),
```
timadye, line 293: I noticed also that the HGTD `binTolerancePhi` is 0.025 in Python, but `geo-tgeo-sfbin-phi-tolerance` is 0.25 in JSON. If I change the Python to 0.25, I get identical results from the two methods.
Do you know which is better?

noemina: I have changed also this parameter to use the same we have in the JSON file.
Not sure where this number come from though.

timadye (approved):
looks good to me

paulgessinger:
The python format check fails, can you fix that @noemina?

paulgessinger (approved):

PR #1156, opened 2022-02-10
fix: Make ITk splitting json entry optional

### LINKED ISSUE #1153: ITk Information expected in TGeo Example
There is currently ITk information expected in the TGeometry JSON reader for example [here](https://github.com/acts-project/acts/blob/7a4814b4d375db8d43da2637aa773bfeef5c200a/Examples/Detectors/TGeoDetector/include/ActsExamples/TGeoDetector/JsonTGeoDetectorConfig.hpp?_pjax=%23js-repo-pjax-container#L144). Is this intended to be present given that these examples can be used to do the material mapping? The TGeometry standalone example can't run unless the JSON configuration has some expected ITk geometry info.

niermann999:
This PR makes the ITk specific module splitting entry in the TGeo example json config file optional. Fixes #1153


paulgessinger (approved):
LGTM!

PR #1158, opened 2022-02-13
feat: Implement iterator support for k-d trees

stephenswat:
Since the elements in our k-d tree implementation are stored consecutively in memory, it is trivial (and performant) to provide iterator-based access to them, such that it becomes easier to loop over the elements of the tree in a flat fashion. Of course, the order of the elements will be somewhat unpredictable.


paulgessinger (approved):
Good idea!

PR #1159, opened 2022-02-13
fix: Add constant qualifiers to seed finder utilities

stephenswat:
Some of the input vectors used in the seed finder utilities are currently not marked as `const` where they should be. This limits their usability in contexts where vectors are more strictly cv-qualified. We should be able to safely set these inputs as constant, which is what this commit does.


stephenswat:
Should also mention this is totally my own fault #1035. :sunglasses:

paulgessinger (approved):
:+1:

PR #1163, opened 2022-02-14
ci: Lock dependencies for python-level tests

paulgessinger:
They are currently not locked, and I have a slight suspicion that this is the cause for the vertexing reproducibility test failing sporadically at the moment.


stephenswat (approved):
Looks great to me. Perhaps, at some point, a more full-fledged Python dependency like Poetry could be helpful, but this should be fine for now.

paulgessinger:
I use poetry a lot, but for this project I feel we should stick to just pip and `requirements.txt`. I use `pip-compile` from `pip-tools` to get (some of the) locking benefits that poetry provides.

PR #1164, opened 2022-02-15
fix: Covariance expansion with expander

paulgessinger:
I believe this would have to be `E * C * E.T`, otherwise the output matrix is not actually 6x6, but 6x2.


asalzburger (approved):
Makes sense.

PR #1166, opened 2022-02-16
feat: Improvements to Cuboid Volume Builder

pbutti:
Adds a number of changes to the CuboidVolumeBuilder to make it more robust.


paulgessinger:
Thanks @pbutti! I set the milestone and the CI should run now.

paulgessinger:
I'll fix the build errors.

Core/src/Geometry/CuboidVolumeBuilder.cpp
```diff
@@ -35,8 +35,7 @@
 #include <limits>
 #include <optional>
```
niermann999, line 36: That is already included in the header

Core/src/Geometry/CuboidVolumeBuilder.cpp
```diff
@@ -61,32 +60,63 @@ std::shared_ptr<const Acts::Layer> Acts::CuboidVolumeBuilder::buildLayer(
...
+  }
+
+  centroid /= cfg.surfaces.size();
+
+  Transform3 trafo = Transform3::Identity();
+  trafo.translation() = centroid;
+  if (cfg.rotation) {
+    trafo.linear() = *cfg.rotation;
+  } else {
+    trafo.linear() = cfg.surfaces.front()->transform(gctx).rotation();
```
niermann999, line 82: Dumb question, but what makes the first surface special enough, that we can use its rotation? Calling ```front``` is also undefined behavior in case there are no ```surfaceCfg```s given or is this case excluded somewhere else?

Core/src/Geometry/CuboidVolumeBuilder.cpp
```diff
@@ -183,6 +218,14 @@ Acts::MutableTrackingVolumePtr Acts::CuboidVolumeBuilder::trackingVolume(
     volumes.push_back(buildVolume(gctx, volCfg));
   }
 
+  // Sort the volumes vectors according to the center location, otherwise the
+  // binning boundaries will fail
+
```
niermann999, line 223: One line too many it seems :)

niermann999 (changes requested):
I am not sure what the expected behavior of ```CuboidVolumeBuilder``` is in detail, but the code makes sense to me

Core/src/Geometry/CuboidVolumeBuilder.cpp
paulgessinger: We have to make some choice about the rotation. The assumption is that this is a telescope-like geometry, where the sensors are located linearly but you're right that they might be rotated with respect to each other. We average the positions, but I don't think there's a good way to average the rotations accessible here.

niermann999: I see, thanks for clarifying. Maybe this should be commented then?

paulgessinger: That's a good idea. We should also add a check to see if `cfg.surfaceCfg.empty()`

pbutti:
Opened a PR to the base branch in Paul's fork. When that is merged, the commits will be picked up.

Core/src/Geometry/CuboidVolumeBuilder.cpp
paulgessinger: Should be added now.

Core/src/Geometry/CuboidVolumeBuilder.cpp
paulgessinger: Removed.

niermann999 (approved):

PR #1167, opened 2022-02-16
fix: Prevent unphysical number of phi bins (#1161)

### LINKED ISSUE #1161: Seed finding C++ example throws exceptions
As I mentioned briefly in this morning's meeting, the seed finding examples no longer seem to work. There is a `std::length_error` being thrown somewhere. Here's what that looks like:

```
$ ./build/bin/ActsExampleSeedingGeneric \
    --input-dir data/sim/pions \
    --bf-constant-tesla=0:0:2 \
    --digi-config-file Examples/Algorithms/Digitization/share/default-smearing-config-generic.json \
    --geo-selection-config-file Examples/Algorithms/TrackFinding/share/geoSelection-genericDetector.json               


======================= DEPRECATION NOTICE ========================
The examples executables is deprecated. They will be removed in a
future version.
Consider using the python bindings for the example algorithms: 
https://acts.readthedocs.io/en/latest/examples/python_bindings.html
===================================================================


10:13:42    MagneticFiel   INFO      Use a constant magnetic field
10:13:42    Sequencer      INFO      Added reader 'CsvSimHitReader'
10:13:42    Sequencer      INFO      Added reader 'CsvParticleReader'
10:13:42    Sequencer      INFO      Added algorithm 'DigitizationAlgorithm'
10:13:42    Sequencer      INFO      Added algorithm 'TruthSeedSelector'
10:13:42    SpacePointMa   INFO      Space point geometry selection:
10:13:42    SpacePointMa   INFO        vol=7|lay=8
10:13:42    SpacePointMa   INFO        vol=7|lay=10
10:13:42    SpacePointMa   INFO        vol=7|lay=12
10:13:42    SpacePointMa   INFO        vol=7|lay=14
10:13:42    SpacePointMa   INFO        vol=8|lay=2
10:13:42    SpacePointMa   INFO        vol=8|lay=4
10:13:42    SpacePointMa   INFO        vol=8|lay=6
10:13:42    SpacePointMa   INFO        vol=8|lay=8
10:13:42    SpacePointMa   INFO        vol=9|lay=2
10:13:42    SpacePointMa   INFO        vol=9|lay=4
10:13:42    SpacePointMa   INFO        vol=9|lay=6
10:13:42    SpacePointMa   INFO        vol=9|lay=8
10:13:42    Sequencer      INFO      Added algorithm 'SpacePointMaker'
10:13:42    Sequencer      INFO      Added algorithm 'SeedingAlgorithm'
10:13:42    Sequencer      INFO      Added algorithm 'TrackParamsEstimationAlgorithm'
10:13:42    Sequencer      INFO      Added writer 'TrackFinderPerformanceWriter'
10:13:42    Sequencer      INFO      Added writer 'SeedingPerformanceWriter'
10:13:42    Sequencer      INFO      Added writer 'RootTrackParameterWriter'
10:13:42    Sequencer      INFO      Processing events [0, 100)
10:13:42    Sequencer      INFO      Starting event loop with -1 threads
10:13:42    Sequencer      INFO        0 services
10:13:42    Sequencer      INFO        0 context decorators
10:13:42    Sequencer      INFO        2 readers
10:13:42    Sequencer      INFO        5 algorithms
10:13:42    Sequencer      INFO        3 writers
terminate called after throwing an instance of 'std::length_error'
  what():  vector::_M_default_append
[1]    548061 IOT instruction  ./build/bin/ActsExampleSeedingGeneric --input-dir data/sim/pions 
```

Through bisection, I have determined that this error is introduced in c8ef092c4dacb09971bc512701d7d662deede902 from #1038. Before this change, everything works fine, after that the exception is thrown.

Pinging @paulgessinger and @LuisFelipeCoelho.

stephenswat:
The bug described in issue #1161 is caused by the creation of negative or (near) infinite number of phi bins, which cause the internal mechanisms of `std::vector` to crash. This effect is caused by the fact that the number of phi neighbours is not properly initialized, causing it to take on uninitialized values.

This commit does two things to prevent this issue. First of all, it adds a sensible default value for the number of phi neighbours, ensuring that it never takes on uninitialized values. Secondly, a sanity check is added to the space point grid to ensure that the phi delta value has a physically meaningful value; if this is not the case, an exception is thrown.

Resolves #1161.


paulgessinger:
Looks fine from here, @LuisFelipeCoelho can you review this PR?

LuisFelipeCoelho:
Looks fine to me. Thanks for finding the problem and fixing it!

LuisFelipeCoelho (approved):

stephenswat:
Hmm, it looks like some of the tests fail.

stephenswat:
Okay, so one of the Python tests appears to be losing tracks. I would suspect then, that one phi bin neighbour is not enough. Does anyone have a suggestion for a better default value? Otherwise, we might want to look into a mechanism that can ensure that this value is set.

robertlangenberg:
This may be related to #1172 as the size of the phi bins reduced. In principle a single phi bin should cover the maximum deflection in phi such that looking at a single neighbor should be enough. This didn't change, right @noemina @LuisFelipeCoelho ?

noemina:
> This may be related to #1172 as the size of the phi bins reduced. In principle a single phi bin should cover the maximum deflection in phi such that looking at a single neighbor should be enough as long as numPhiNeighbors is 1. This didn't change, right @noemina @LuisFelipeCoelho ?

That's exactly what it does. We didn't change the logic there.

stephenswat:
From the discussion in #1172 I gather that 0 is probably a better default value than 1 in this case.

robertlangenberg:
I just had a quick chat with Noemi, and I think I understand the issue. There are configuration parameters called numPhiNeighbors for Grid and BinFinder which don't mean the same. The numPhiNeighbors for the Grid means "how many phiBin neighbors (plus the current bin) should cover the full deflection of a minimum pT particle". Setting this to zero means "no phi bin neighbors needed to cover the full deflection", i.e. 1 phi-bin covers the full deflection. 

What has been newly added is the consideration of the max impact parameters, which increases the size of the bins a bit and is probably uncontroversial, but may slightly change results.

I think this should be renamed and even numbers of phiBins should be allowed to cover the full deflection. 

The numPhiNeighbors parameter of the BinFinder sets how many neighboring phi bins (plus the current one) are returned for each call to "findBins" (which I think is aptly named).

To get the behavior the seeding test expects, (assuming the newly added d_ZERO doesn't change the results as well), the numPhiNeighbors of the Grid has to be set to zero, while the numPhiNeighbors of the BinFinder must be set to 1.

robertlangenberg:
The reason this was changed like this is because for ATLAS at µ=60 the configuration with numPhiNeighbors=1 for both Grid and BinFinder has shown to give acceptable results and reduces CPU usage (due to the reduced combinatorics, this corresponds to a 66% reduction of the phi-range).

LuisFelipeCoelho (approved):

PR #1168, opened 2022-02-17
feat: Curvature sorting in SeedFilter

LuisFelipeCoelho:
This PR introduces an option to sort the Seed by curvature in the `SeedFilter`. 

If the boolean `curvatureSortingInFilter` is true, the seeds will be sorted by curvature and if a seed has a curvature greater that the maximum (`upperLimitCurv`) it will break out of loop.

`curvatureSortingInFilter` is set to false by default.

@noemina @robertlangenberg @paulgessinger


robertlangenberg:
delete comments in SeedFilter.ipp line 70 & 71
      // TODO: how much slower than sorting all vectors by curvature
      // and breaking out of loop? i.e. is vector size large (e.g. in jets?)

as this PR addresses this :)

Did you test if this is faster respectively not significantly slower for small topSP vectors? If so, no need to make it conditional on `curvatureSortingInFilter`, and just always run it.

robertlangenberg (approved):
Hey, looks good! Did you do tests if this speeds up performance? If not or if tests weren't showing it's speeding up or maintaining performance in +- all cases, we should probably just push this in. If you have tests that DO say it's always beneficial, we can also get rid of the switch curvatureSortingInFilter.

LuisFelipeCoelho:
@robertlangenberg these plots compare the time per event with and without the curvature sorting. In the first one, I ran the seeding 100 times for 500 single muon events (100 GeV) and the second one 100 times for 20 ttbar events (14 TeV).
The performance seems to be the same.

[compare_curvature_sorting_100MeV_singleMu_500evts.pdf](https://github.com/acts-project/acts/files/8326038/compare_curvature_sorting_100MeV_singleMu_500evts.pdf)
 
[compare_curvature_sorting_ttbar_20evts.pdf](https://github.com/acts-project/acts/files/8326062/compare_curvature_sorting_ttbar_20evts.pdf)

paulgessinger (approved):
Only a clang format change after @robertlangenberg's approval, so let's merge.

PR #1170, opened 2022-02-18
Feat restrict json surface wrtier

asalzburger:
Minor update to allow the JsonSurface writer to optionally be restricted to only one layer for testing purposes of binning.


Corentin-Allaire (approved):
Looks good to me

asalzburger:
Actually, I might close this one, one can easily catch this with json manipulation, does not need extra C++ code.

PR #1171, opened 2022-02-18
refactor: Add method to reset jacobian in stepper

Corentin-Allaire:
When trying to use the ATLAS stepper with the KF I noticed that it wouldn't work since we use to set the stepping state jacobian to a BoundMatrix::Identity() which isn't compatible with the ATLAS stepper.
This PR add a method to the stepper to set the jacobian to the identity, and thus solve this issue. 
Similarly I had to transform the update to a reset for the propagation to the Perigee as it relied on manually setting some matrix afterward (which was not compatible with the ATLAS stepper).

### DIFF
--- a/Core/include/Acts/Propagator/AtlasStepper.hpp
+++ b/Core/include/Acts/Propagator/AtlasStepper.hpp
@@ -1376,6 +1376,54 @@ class AtlasStepper {
     return h;
   }
 
+  /// Method that reset the Jacobian to the Identity for when no bound state are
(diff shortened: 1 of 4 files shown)


paulgessinger:
This looks good to me generally. I'm very curious to see if the CI reports changed outputs on this.

paulgessinger:
CI reports no output changes (as expected I guess). So let's go.

paulgessinger (approved):

PR #1173, opened 2022-02-21
refactor: transformCoordinates methods without InternalSpacePoint class

CarloVarni:
Adding `transformCoordinates` functions that does not rely only on `InternalSpacePoint`. These methods will accept any space point format and require the user to define a function for retrieving coordinate and covariance values (i.e. `std::function<std::array<float, 6>(const external_spacepoint_t&)>`). This will allow users to use it externally to ACTS.

This does not remove the previous `transformCoordinates` functions in order to be backward compatible. However, they are adapted to use these new implementations.


Core/include/Acts/Seeding/SeedFinderUtils.hpp
```diff
@@ -36,6 +36,13 @@ LinCircle transformCoordinates(
     const InternalSpacePoint<external_spacepoint_t>& sp,
     const InternalSpacePoint<external_spacepoint_t>& spM, bool bottom);
 
+template <typename external_spacepoint_t>
+LinCircle transformCoordinates(
+    const external_spacepoint_t& sp, const external_spacepoint_t& spM,
+    bool bottom,
+    std::function<std::array<float, 6>(const external_spacepoint_t&)>
```
paulgessinger, line 43: Since this is a template function anyway, you could go ahead and replace the `std::function` with a blanket template to allow passing a lambda as is, without the overhead of `std::function`.

Core/include/Acts/Seeding/SeedFinderUtils.ipp
```diff
@@ -11,21 +11,37 @@ template <typename external_spacepoint_t>
...
   // vectorized version of this function, except that it operates on a single
   // spacepoint. Please see the other version of this function for more
   // detailed comments.
-  float xM = spM.x();
-  float yM = spM.y();
-  float zM = spM.z();
-  float rM = spM.radius();
-  float varianceZM = spM.varianceZ();
-  float varianceRM = spM.varianceR();
+
```
paulgessinger, line 36: If you replace `std::function`, you could consider using SFINAE to add a `static_assert` here that gives a helpful error message.

Core/include/Acts/Seeding/SeedFinderUtils.hpp
CarloVarni: Moved to `typename callable_t`

Core/include/Acts/Seeding/SeedFinderUtils.ipp
CarloVarni: I should have addressed this now

robertlangenberg (approved):
This looks good, sorry for taking so long with the review!

paulgessinger:
@CarloVarni can you fix the CI?

CarloVarni:
@paulgessinger I should have fixed the compilation issues now. I also found an inconsistency between the `.ipp` and `.hpp` files in a function declaration already present in main, which I solved (a `const`)

paulgessinger (approved):
Let's go then.

Issue #1098, opened 2021-12-06
python config for example chains


timadye:
Discuss the best way to organise the Python in `Examples/Scripts/Python` to allow configuration of various full-chain examples.

Some ideas already in #1093 and #1096. @asalzburger suggested he and I post further ideas as comments, and we can discuss at the dev meeting.

CC @paulgessinger

timadye:
Here are some of my thoughts...

Most of the current examples in `Examples/Scripts/Python` seem to be complete examples that run

1. evgen
2. geometry
3. fatras
4. digitization
5. seeding
6. track finding

at least up to the step being tested. I think it would be easier and clearer to have each step defined in a stand-alone routine (in its own `import`) and have the current `runParticleGun`, `runFatras` functions call them. Some of those might want different settings for the preceding steps (eg. different number of particles for evgen), so the `run*` routines may have to change some defaults.

Orthogonal to this, it would also be useful to have one or more full-chain tests, where the different options can be specified with command-line options. Probably it is simpler to have separate Python main commands for Generic, ODD (or all DD4hep?) , and ATLAS (or all TGeo?) detectors. It would be good if these could be small scripts that set the appropriate defaults and call the other modules. It would also be clearer if the `argparse` option setup and overall defaults were defined in the appropriate modules.

Eg. `fatras.py` could contain the following:

* `addFatras(s,**opts)`: configure `ParticleSelector` and `FatrasSimulation`, and add them to the sequence, `s`. Or should it return the algorithms and let the user add them? Perhaps complicated for the user if there are optional additional algorithms, like printing or dumping the output.
* `optFatras(p)`: contains several `p.add_argument()`s to define all the Fatras options.
* `runFatras()`: same as current definition, but uses `addParticleGun()` and `addFatras()`
* main: same as current definition

One problem with this is that `addFatras()` can be passed Python function arguments or the result of `p.parse_args()` (a simple `argparse.Namespace` object, convertible to a `dict` with `vars()`). Do we want `addFatras()` to merge common options, or require the user to convert to one or the other? I don't know if it is too complicated, but there is some advantage to making all options be specified with `argparse`. That gives automatic documentation and greatest flexibility, but probably makes the Python more complicated than it need be.

Here is an example of how `runFatras` would work if we made all options potential command-line options:
```
def runFatras(outputDir):
  import particle_gun, generic_detector

  p = argparse.ArgumentParser()   # or have something in common.py to add common options like --loglevel
  optParticleGun(p)
[...]

PR #1093, opened 2021-12-03
feat: configurable particle gun

asalzburger:
This is my first attempt of updating the python scripts to be more flexible if needed.

Also tagging: @timadye

PR #1174, opened 2022-02-25
feat: remaining Python examples can be chained

timadye:
Update the Python in `Examples/Scripts/Python` to allow configuration of various full-chain examples.

Continuing the work of #1098 in order to complete the proposal originally discussed in #1093.


timadye:
I open this PR to track ongoing development and stop #1098 going stale 😄. So far I have mostly completed the digitization and seeding examples, though the latter will need more customisation options.

Examples/Scripts/Python/fatras.py
```diff
@@ -36,16 +37,22 @@ def addFatras(
...
-    )
+    if preselectParticles:
+        particles_selected = "particles_selected"
+        s.addAlgorithm(
+            acts.examples.ParticleSelector(
+                level=s.config.logLevel,
+                inputParticles="particles_input",
+                outputParticles=particles_selected,
+            )
+        )
```
timadye, line 48: Hi @paulgessinger,

Do you know what `ParticleSelector` does when no selections are specified? Naively I would have thought it did nothing, so why is this algorithm included here? It was there in the original `runFatras` example.

This is an issue because the seeding example doesn't call `ParticleSelector` before Fatras. If I add it (by calling `addFatras`), then I get a slightly different number of hits and `pytest` fails. That's why I added a flag to disable it, but it would be good to understand this oddity.

Alternatively, it might be more useful to allow more arbitrary selections specified by the caller. What do you think?

I hope this is a good place to put in random questions about this development, and @ you in my PR.

Thanks, Tim.

paulgessinger: Oh that's a good point. It's possible I skipped something when transcribing the original example to python. In principle the `ParticleSelector` should do nothing, but I could imagine due to numerical effects that it does throw out a few particles in edge cases. 

Did you by any chance have a look at the resulting written particles? Do they change noticeably?

timadye: That's the odd thing: I can't see any difference between the initial and selected particles. The same number, values, and order. Fatras particles and hits are identical. Only when the seeding happens do things change: `track_finder_particles` has +/-1 or 2 tracks for some (23%) particles.
One other clue: even if I don't use the output, running `ParticleSelector` still changes the seeding. I guess this means there is something non-deterministic or state-dependent in the seeding algorithm - even single-threaded. Or just a bug in how we're using it. (I checked it's not just an output collection name clash.)
Could it be a problem with one of the mapping objects, if they use pointers that will be different if moved?

paulgessinger: Hm, that sounds weird indeed. So you don't see differences at the particle level, but they do change the seeds/tracks... 

If there's anything state-dependent, it also being present in single-threaded mode indicates that it's likely not a thread-safety issue. But other than that I don't know.

Examples/Scripts/Python/seeding.py
```diff
@@ -9,97 +9,54 @@
...
+        the output folder for the Root output, None triggers no output
+    logLevel : acts.logging.Level, None
+        logging level to override setting given in `s`
+    """
+
+    def customLogLevel(custom: acts.logging.Level = acts.logging.INFO):
+        """override logging level"""
+        if logLevel is None:
+            return s.config.logLevel
+        return acts.logging.Level(max(custom.value, logLevel.value))
```
paulgessinger, line 40: Not entirely sure I get the logic here. This is either

- The Sequencer log level, and it that case all components get this exact level
- An explicit loglevel, and in that case, the components use the more verbose out of that and what's given to this function.

Did I get this right, and if so, does that make sense?

Examples/Scripts/Python/fatras.py
timadye: Yes, that's what I see. I can't think of anything else I can test. Maybe make a mininal script to try to demonstrate the issue? Or see if it does the same with the example binaries? Neither a trivial amount of work, so let me know if it's worth it.

Examples/Scripts/Python/seeding.py
timadye: I agree, this is pretty arbitrary. What I was trying to do was allow to reproduce the various logging levels that were present in the old `runSeeding`. That's perhaps useful, but don't want to mess up the simple case where you just keep the same logging level specified in the `Sequencer`.

> An explicit loglevel, and in that case, the components use the more verbose out of that and what's given to this function.

Less verbose, I think. Lower numbers are more verbose, right?

paulgessinger: You're right yes.

Examples/Scripts/Python/ckf_tracks.py
```diff
@@ -22,313 +135,108 @@ def runCKFTracks(
...
+            maxSeedsPerSpM=1,
+            sigmaScattering=50,
+            radLengthPerSeed=0.1,
+            minPt=500 * u.MeV,
+            bFieldInZ=1.99724 * u.T,
+            impactMax=3 * u.mm,
         ),
+        TrackParamsEstimationConfig(deltaR=(10.0 * u.mm, None)),
+        acts.logging.VERBOSE,
+        inputParticles="particles_initial",
```
timadye, line 221: Hi @paulgessinger,
`seeding.py` uses input `particles_final`, while `ckf_tracks.py` uses  input `particles_initial`. Why the difference, and what's best?
I guess it should be `particles_final`, so I made that the `addSeeding` default and override it here to keep the same output for `pytest`. What do you think?
Thanks, Tim.

timadye: @paulgessinger, any ideas on this discrepancy between `seeding.py` and `ckf_tracks.py`? Should we keep it or change to one or the other?

CI/physmon/phys_perf_mon.sh
```diff
@@ -35,26 +36,26 @@ function run() {
 
 
 run \
-    physmon/performance_ckf_tracks_truth_smeared.root \
+    $outdir/performance_ckf_tracks_truth_smeared.root \
```
paulgessinger, line 39: Whoops 😅

Examples/Scripts/Python/common.py
```diff
@@ -44,48 +44,23 @@ def addPythia8(
     vertexStddev: acts.Vector4 = acts.Vector4(0, 0, 0, 0),
     vertexMean: acts.Vector4 = acts.Vector4(0, 0, 0, 0),
 ):
-    vertexGenerator = acts.examples.GaussianVertexGenerator(
-        stddev=vertexStddev, mean=vertexMean
+    """This function steers the particle generation using Pythia8
+
+    NB. this version is included here only for compatibility. Please use pythia8.addPythia8 instead.
```
paulgessinger, line 49: Should we do a deprecation warning then? Or maybe just remove this one?

In any case, if it's just a forward, one could use `*args, **kwargs` to forward instead of listing all the defaults again here.

timadye: I think we can remove this one, but lets do that in a separate PR.

Unfortunately a few of the kwargs are a bit different (`beam` and `vtxGen`) to fit with our new style.

Examples/Python/tests/helpers/__init__.py
```diff
@@ -5,6 +5,11 @@
 from acts.examples import BareAlgorithm
 
 geant4Enabled = any(v.startswith("G4") for v in os.environ.keys())
+if geant4Enabled:
+    try:
+        import acts.examples.geant4
+    except ImportError:
+        geant4Enabled = False
```
timadye, line 12: I added this to get the pytest to work. It gave errors when DD4HEP was enabled but not Geant4. Is this OK, @paulgessinger? I don't know why `geant4Enabled` is handled differently from `dd4hepEnabled`.

Examples/Python/tests/test_examples.py
```diff
@@ -923,7 +922,7 @@ def test_ckf_tracks_example_truth_smeared(tmp_path, assert_root_hash):
 
 @pytest.mark.skipif(not dd4hepEnabled, reason="DD4hep not set up")
 @pytest.mark.slow
-# @pytest.mark.filterwarnings("ignore::UserWarning")
+@pytest.mark.filterwarnings("ignore::UserWarning")
```
timadye, line 925: @paulgessinger, why was this line commented out? It gives a warning in pytest. (The warning changed a bit with this PR, but it's still there.)
Another way to fix this would be to disable ROOT output, since the output is not checked. However it's perhaps useful to check that the output doesn't crash.

Examples/Scripts/Python/seeding.py
```diff
@@ -1,266 +1,385 @@
...
+    field: acts.MagneticFieldProvider,
+    geoSelectionConfigFile: Optional[Union[Path, str]] = None,
+    seedingAlgorithm: SeedingAlgorithm = SeedingAlgorithm.Default,
+    truthSeedRanges: Optional[TruthSeedRanges] = TruthSeedRanges(),
+    particleSmearingSigmas: Optional[ParticleSmearingSigmas] = ParticleSmearingSigmas(),
+    initialVarInflation: Optional[list] = None,
+    seedfinderConfigArg: Optional[SeedfinderConfigArg] = SeedfinderConfigArg(),
+    trackParamsEstimationConfig: Optional[
+        TrackParamsEstimationConfig
+    ] = TrackParamsEstimationConfig(),
```
timadye, line 74: @paulgessinger,
These parameters all now take their defaults from the C++, so defaults are only specified in one place.
The main disadvantage of this is that they aren't always trivial to find.
I did wonder whether it would help to show the defaults in the comments (either here, or in the - currently incomplete - help text below), or if the comments getting out of sync is equally problematic.
Anyway, I suggest to deal with this in a follow-up PR.

[similarly for parameters in other routines]

Examples/Python/tests/helpers/__init__.py
paulgessinger: The import check checks if the module is importable. I don't remember exactly what I was thinking, but the other check checks if the user has sourced the G4 file, meaning that the necessary environment variables are set. I guess I assumed if that is the case, the import should be safe to do.

Adding this extra check is probably fine!

Examples/Scripts/Python/common.py
paulgessinger: Ok

Examples/Scripts/Python/seeding.py
paulgessinger: I think it's good that the defaults are only in one place. We could potentially add comments pointing to where to find these defaults.

On another note: the use of `Optional` for e.g. `truthSeedRanges` implies that you can pass `None` to these arguments. Does that actually work?

Examples/Scripts/Python/vertex_fitting.py
```diff
@@ -32,6 +32,109 @@ class VertexFinder(enum.Enum):
...
+
+    if outputDirRoot is not None:
+        outputDirRoot = Path(outputDirRoot)
+        if not outputDirRoot.exists():
+            outputDirRoot.mkdir()
+        if associatedParticles == selectedParticles:
+            warnings.warn(
+                "Using RootVertexPerformanceWriter with smeared particles is not necessarily supported. "
+                "Please get in touch with us"
+            )
```
timadye, line 120: @paulgessinger,
This test is a bit different from the previous one (`inputTrackSummary == None with outputRoot`) in `runVertexFitting`, but I think this is what is intended. Does it look reasonable?

Examples/Python/tests/test_examples.py
paulgessinger: I don't remember. I think this will discard the warning though. Do we want to display it?

paulgessinger: So that means we should remove this line altogether, right?

Examples/Scripts/Python/seeding.py
paulgessinger: Right. I guess what I'm saying is if you're meant to either not provide the argument at all (= use default) or provide a valid value, `Optional` isn't needed here.

Examples/Python/tests/test_examples.py
timadye: This line supresses the expected warning in the pytest. Isn't that what we want?

Examples/Scripts/Python/seeding.py
timadye: Ah, maybe I misunderstand what `Optional` means. Where should it still be used?

Examples/Scripts/Python/ckf_tracks.py
timadye: We decided to keep this discrepancy in for this PR to reproduce the original root_file_hashes, but correct it in a future PR.
I'll just check if I need to change the default in `addSeeding` and can resolve this.

Examples/Python/tests/test_examples.py
paulgessinger: Should be alright.

Examples/Scripts/Python/ckf_tracks.py
```diff
@@ -1,12 +1,126 @@
...
+
+        # write track summary from CKF
+        trackSummaryWriter = acts.examples.RootTrajectorySummaryWriter(
+            level=s.config.logLevel,
+            inputTrajectories=trackFinder.config.outputTrajectories,
+            # @note The full particles collection is used here to avoid lots of warnings
+            # since the unselected CKF track might have a majority particle not in the
+            # filtered particle collection. This could be avoided when a seperate track
+            # selection algorithm is used.
+            inputParticles="particles_selected",
```
paulgessinger, line 90: The comment says it's the 'full particles collection' but it's hard coded to `particles_selected`. That comes from FATRAS' `preselectParticles`?

timadye: Right, `particles_selected` comes from the output of Fatras' initial `ParticleSelector`. At the moment, the selector passes all particles, so it shouldn't make any difference. We run the selector anyway to maintain exact consistency for the python checks, but we could drop that later just for speed.

I don't know if this comment is referring to the "full particles collection" as the particles before the Fatras preselection (`ParticleSelector`), or the particles before the seeding preselection (`TruthSeedSelector`). If the latter, then the comment still makes sense.

Maybe worth pointing out, that this setting and the comment are unchanged by this PR. It was just moved to `addCKFTracks` and hardcoded here, rather than in `selector`.

Examples/Scripts/Python/pythia8.py
```diff
@@ -1,51 +1,173 @@
...
+        hard and pileup processes
+    vtxGen : VertexGenerator, None
+        vertex generator module
+    outputDirCsv : Path|str, path, None
+        the output folder for the Csv output, None triggers no output
+    outputDirRoot : Path|str, path, None
+        the output folder for the Root output, None triggers no output
+    printParticles : bool, False
+        print generated particles
+    """
```
CouthuresJeremy, line 55: Missing `returnEvGen` description in docstring

Examples/Scripts/Python/seeding.py
```diff
@@ -1,266 +1,384 @@
...
+    field : magnetic field
+    geoSelectionConfigFile : Path|str, path, None
+        Json file for space point geometry selection. Not required for SeedingAlgorithm.TruthSmeared.
+    outputDirRoot : Path|str, path, None
+        the output folder for the Root output, None triggers no output
+    logLevel : acts.logging.Level, None
+        logging level to override setting given in `s`
+    rnd : RandomNumbers, None
+        random number generator. Only used by SeedingAlgorithm.TruthSmeared.
+    """
```
CouthuresJeremy, line 94: - [ ] Missing `seedingAlgorithm`  description in docstring
- [ ] Missing `truthSeedRanges`  description in docstring
- [ ] Missing `particleSmearingSigmas`  description in docstring
- [ ] Missing `initialVarInflation`  description in docstring
- [ ] Missing `seedfinderConfigArg`  description in docstring
- [ ] Missing `trackParamsEstimationConfig`  description in docstring
- [ ] Missing `inputParticles`  description in docstring

Examples/Scripts/Python/ckf_tracks.py
```diff
@@ -1,12 +1,126 @@
...
+    ----------
+    s: Sequencer
+        the sequencer module to which we add the Seeding steps (returned from addSeeding)
+    trackingGeometry : tracking geometry
+    field : magnetic field
+    outputDirCsv : Path|str, path, None
+        the output folder for the Csv output, None triggers no output
+    outputDirRoot : Path|str, path, None
+        the output folder for the Root output, None triggers no output
+    """
```
CouthuresJeremy, line 37: - [ ] Missing `truthSeedRanges` description in docstring
- [ ] Missing `selectedParticles` description in docstring

Examples/Scripts/Python/particle_gun.py
```diff
@@ -97,47 +79,36 @@ def addParticleGun(
         random number generator
     """
```
CouthuresJeremy, line 80: - [ ] Missing `phiConfig` description in docstring
- [ ] Missing `multiplicity` description in docstring

Examples/Scripts/Python/vertex_fitting.py
```diff
@@ -32,6 +32,109 @@ class VertexFinder(enum.Enum):
...
+    Parameters
+    ----------
+    s: Sequencer
+        the sequencer module to which we add the Seeding steps (returned from addVertexFitting)
+    field : magnetic field
+    outputDirRoot : Path|str, path, None
+        the output folder for the Root output, None triggers no output
+    logLevel : acts.logging.Level, None
+        logging level to override setting given in `s`
+    """
```
CouthuresJeremy, line 54: - [ ] Missing `associatedParticles` in docstring
- [ ] Missing `vertexFinder` in docstring

Examples/Scripts/Python/pythia8.py
timadye: I'll add:
```
    returnEvGen: bool, False
        returns EventGenerator instead of Sequencer.
        This option  is included for compatibility and will be removed when common.addPythia8 is removed.
```

I mention this now, because this reminds me of a minor, but general issue with splitting our Python examples up into modules. There is no way for the caller to inspect or modify the algorithms/readers/writers, since they are just added to the `Sequencer` and not returned to the caller.

We could do that, but many modules set up several algorithms so we'd have to think a bit how best to do it. But the caller already has the `Sequencer`, which has this information - just no way to inspect it as far as I could easily tell. Could we look at a way of querying the algorithms, readers, writers etc? Eg. allowing something like:

```
outputParticles = s.lastEntry(acts.examples.EventGenerator).config.outputParticles
```

`acts.examples.Sequencer.lastEntry(type)` would return the last entry that matches the type. If that's difficult, we could search by name. Actually the C++ would probably require separate separate methods for `lastAlgorithm`, `lastReader`, `lastWriter` etc, so maybe that's better. We could do something more general and complicated, but that would probably suit this use case well.

Sorry for the digression. What do you and @paulgessinger think?

timadye: PS. related to the above: normally (unless `returnEvgen=True`) we return the `Sequencer`, although this is also passed in by the user. This is perhaps useful to make clear that the the `Sequencer` has been modified, but isn't actually useful.

Would it help to have the examples just call without resetting `s`?

timadye: BTW, all my discussion is about ideas for the future. I hope it won't delay this PR.

paulgessinger: I'm not sure. I would try to keep this meta-level somewhat simple, and try to encourage people to copy the actual algorithm constructions into their own scripts if they want to more thoroughly customize. That was kind of the original idea of these basic scripts.

I feel like with this PR we have a good middle ground: use the `addXXX` function if you just want to compose what's already there, and fall back to a "full" script if you want to control more closely what's being set up.

CouthuresJeremy: > I'll add:
> 
> ```
>     returnEvGen: bool, False
>         returns EventGenerator instead of Sequencer.
>         This option  is included for compatibility and will be removed when common.addPythia8 is removed.
> ```
From the docstring point of view, this is enough for me.

> I mention this now, because this reminds me of a minor, but general issue with splitting our Python examples up into modules. There is no way for the caller to inspect or modify the algorithms/readers/writers, since they are just added to the `Sequencer` and not returned to the caller.
> 
> We could do that, but many modules set up several algorithms so we'd have to think a bit how best to do it. But the caller already has the `Sequencer`, which has this information - just no way to inspect it as far as I could easily tell. 

I also tried to split the Python examples into modules before seeing this PR and my first thought was also to extract info from the algorithm added into the Sequencer but didn't find a way to do so.

Allowing something like:
> 
> ```
> outputParticles = s.lastEntry(acts.examples.EventGenerator).config.outputParticles
> ```
would make easier the chaining of new algorithms. 

> BTW, all my discussion is about ideas for the future. I hope it won't delay this PR.

I think that such change should'nt be done in this PR as it is more related on how to chain new algorithm. As long as everything works with the current algorithm, I would delay that into another PR.

timadye: I'll resolve this thread when I commit this (and other docstring fixes)

CouthuresJeremy: > @paulgessinger had suggested maybe later moving the `add*` functions to separate files in the Python module tree. That might be a good time to consider this sort of thing.

Yes, I think that examples should be kept simple and not callable, something ```full_chain_itk.py```-like.

Examples/Scripts/Python/seeding.py
timadye: Added. Thanks for spotting this.

Examples/Scripts/Python/vertex_fitting.py
timadye: Added. Thanks for reporting.

Examples/Scripts/Python/ckf_tracks.py
paulgessinger: Hm. Ok I don't remember where this is from then. I don't think I wrote it 😅

paulgessinger (approved):
Ok let's merge this bad boy then.

PR #1182, opened 2022-03-03
build: Allow forcing assertions

paulgessinger:
In the CI, we currently run the examples in Release only, which adds the `-DNDEBUG` flag and consequently disables `assert`. We can't run these in the CI with Debug, since that implies `-O0` and runs forever. 

This PR adds a new CMake option `ACTS_FORCE_ASSERTIONS` which defaults to `OFF`. If enabled, this will prepend a system include directory `cmake/assert_include` to all compilation units, which contains a wrapper around `cassert`, `Eigen/{Core,Dense,Geometry}` which enables `assert` (and `eigen_assert`) regardless of `NDEBUG`.

I verified that this would have triggered the assertion from #1180 for example.

The CI jobs are configured to run with `ACTS_FORCE_ASSERTIONS=ON`.


stephenswat (approved):
I think this looks extremely useful. :+1:

PR #1181, opened 2022-03-03
perf: Tile 8×8 covariance matrix multiplication

stephenswat:
Currently, we are multiplying an 8x8 covariance matrix with an 8x8 transport matrix, and we see that Eigen is failing to optimize this properly, because it is calling a generalized GEMM method rather than an optimized small matrix method. In order to resolve this, we change the code to use a tiled multiplication method which splits the matrices into 4x4 sub-matrices which can be multiplied and added to achieve the desired effect. This has two advantages:

  1. It allows Eigen to use its hand-rolled optimized 4x4 matrix multiplication methods.
  2. It allows us to perform some trickery with matrix identities to reduce the number of floating point operations.

PR #1183, opened 2022-03-03
perf: Optimize Eigen usage in covariance engine

stephenswat:
This commit optimizes some of the Eigen usage in the covariance engine, specifically in the critical path for the propagation examples. The first optimisation we make is to introduce a tiled matrix multiplication method, which takes 2i×2j matrices, and performs four i×j multiplications instead, which Eigen can optimize far more easily. Secondly, we reduce the number of floating point operations performed by working with smaller submatrices wherever possible.

On my machine, the following performance is achieved in the propagation example before this patch: 53.555595 ms/event. After this patch, we take 43.750143 ms/event. This performance gain is independent from the performance gain of #1181.

### DIFF
--- a/Core/src/Propagator/detail/CovarianceEngine.cpp
+++ b/Core/src/Propagator/detail/CovarianceEngine.cpp
@@ -10,6 +10,7 @@
 
 #include "Acts/EventData/detail/TransformationBoundToFree.hpp"
 #include "Acts/EventData/detail/TransformationFreeToBound.hpp"
+#include "Acts/Utilities/Helpers.hpp"
 #include "Acts/Utilities/Logger.hpp"
 #include "Acts/Utilities/Result.hpp"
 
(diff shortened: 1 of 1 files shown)


Core/src/Propagator/detail/CovarianceEngine.cpp
```diff
@@ -15,6 +15,53 @@
...
+/// @tparam N Second dimension of first matrix, and first dimension of second
+/// matrix
+/// @tparam P Second dimension of second matrix
+///
+/// @param[in] a MxN matrix
+/// @param[in] b NxP matrix
+///
+/// @returns The product ab
+template <std::size_t M, std::size_t N, std::size_t P>
+inline ActsMatrix<M, P> tiledMult(const ActsMatrix<M, N>& a,
```
paulgessinger, line 30: How useful do you think is is to pull out of this compilation unit into a helper function somewhere, and potentially add a unit test against a regular multiplication (if we think this should be generally correct)?

Core/src/Propagator/detail/CovarianceEngine.cpp
```diff
@@ -15,6 +15,53 @@
...
+/// @tparam M First dimension of first matrix
+/// @tparam N Second dimension of first matrix, and first dimension of second
+/// matrix
+/// @tparam P Second dimension of second matrix
+///
+/// @param[in] a MxN matrix
+/// @param[in] b NxP matrix
+///
+/// @returns The product ab
+template <std::size_t M, std::size_t N, std::size_t P>
```
paulgessinger, line 29: I guess these could be deduced from the matrix arguments?

Core/src/Propagator/detail/CovarianceEngine.cpp
```diff
@@ -143,20 +190,18 @@ void boundToCurvilinearJacobian(const Vector3& direction,
                                 const FreeMatrix& freeTransportJacobian,
                                 const FreeVector& freeToPathDerivatives,
                                 BoundMatrix& fullTransportJacobian) {
-  // Calculate the derivative of path length at the the curvilinear surface
-  // w.r.t. free parameters
-  FreeToPathMatrix freeToPath = FreeToPathMatrix::Zero();
-  freeToPath.segment<3>(eFreePos0) = -1.0 * direction;
-  // Calculate the jacobian from global to local at the curvilinear surface
```
paulgessinger, line 150: I guess this comment still applies?

Core/src/Propagator/detail/CovarianceEngine.cpp
stephenswat: Yeah, that would be a solid idea. Where do you think would be the best place to put this?

Core/src/Propagator/detail/CovarianceEngine.cpp
stephenswat: I thought so too, but surprisingly it struggled a little. However, we should be able to accept the matrix types themselves (not their sizes) as the template types, and _then_ it should be automatically deducible.

Core/src/Propagator/detail/CovarianceEngine.cpp
stephenswat: Yeah, I'll restore it.

paulgessinger:
On my end I see wall-time for the generic propagation example go from ~26s to 22s for 500 events. Good stuff!

Core/src/Propagator/detail/CovarianceEngine.cpp
paulgessinger: Yeah right. That coupled with a few `static_assert`s and it should be good. Up to you, could also do this in a follow-up PR.

Core/src/Propagator/detail/CovarianceEngine.cpp
paulgessinger: I'd put it here https://github.com/acts-project/acts/blob/main/Core/include/Acts/Utilities/Helpers.hpp.

Also this could be done in a follow-up PR, up to you.

stephenswat: Nah to hell with it I'll do it now. Clean git history and all.

stephenswat: Moved to #1184.

stephenswat:
This is now ready to go in.

paulgessinger (approved):

PR #1184, opened 2022-03-04
feat: Implement blocked matrix multiplication

stephenswat:
Eigen uses a built-in cost model to determine how it should perform certain matrix operations. In general, this works quite well, but there are some edge cases where Eigen can struggle to output efficient code. An especially apparent instance of this is when multiplying 8x8 matrices, which we do very commonly in operations such as stepping. In these cases, we observe that Eigen emits code for generalized matrix-matrix multiplication, rather than specialized code for small matrices. This adds significant overhead to the matrix multiplication, and it slows down some parts of our code a lot.

Our proposed solution is to implement a wrapper around Eigen's matrix multiplication code that breaks a multiplication down into smaller operations through a technique called blocking. A matrix of size 2ix2j can be split into four matrices of sizes ixj, which can then be multiplied and added to provide a correct multiplication of the parent matrices. In this way, we can encourage Eigen to emit code for its hand-optimized small-matrix operations, which should give speed-up in some cases.

The operation is also supported for odd-sized matrices.

### DIFF
--- a/Tests/UnitTests/Core/Utilities/HelpersTests.cpp
+++ b/Tests/UnitTests/Core/Utilities/HelpersTests.cpp
@@ -155,6 +155,48 @@ BOOST_AUTO_TEST_CASE(test_matrix_dimension_switch) {
   }
 }
 
+typedef std::tuple<
+    // Square matrix tests
+    std::pair<ActsMatrix<4, 4>, ActsMatrix<4, 4>>,
+    std::pair<ActsMatrix<8, 8>, ActsMatrix<8, 8>>,
+
+    // Square odd matrix tests
+    std::pair<ActsMatrix<3, 3>, ActsMatrix<3, 3>>,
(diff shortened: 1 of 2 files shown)


Tests/UnitTests/Core/Utilities/HelpersTests.cpp
```diff
@@ -13,6 +13,8 @@
 
 #include <bitset>
 
+#include <boost/mpl/list.hpp>
```
paulgessinger, line 16: For a test it's probably fine, but we phased out usage of boost mpl at some point.

stephenswat: Ah, sorry! I can change it to `std::tuple` instead, which is also supported by Boost test. :smile:

stephenswat: Resolved.

paulgessinger (approved):

PR #1185, opened 2022-03-07
fix: Fix issue with applying transformation to bevelled cylinders

noraemi:
Applying transforms to the cylinder did not work as intended. These fixes should place the side disks at the right place as well as fixing the visualization of the cylinder length.

### DIFF
--- a/Core/src/Surfaces/CylinderBounds.cpp
+++ b/Core/src/Surfaces/CylinderBounds.cpp
@@ -167,7 +167,12 @@ std::vector<Acts::Vector3> Acts::CylinderBounds::createCircles(
   if ((bevelMinZ != 0. || bevelMaxZ != 0.) && vertices.size() % 2 == 0) {
     auto halfWay = vertices.end() - vertices.size() / 2;
     double mult{1};
-    auto func = [&mult](Vector3& v) { v(2) += v(1) * mult; };
+    auto func = [&mult, &ctrans](Vector3& v) 
+	{ 
+	  v = ctrans.inverse() * v;
+          v(2) += v(1) * mult; 
+	  v = ctrans * v;
+	};
     if (bevelMinZ != 0.) {
       mult = std::tan(-bevelMinZ);
       std::for_each(vertices.begin(), halfWay, func);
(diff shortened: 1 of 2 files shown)


Core/src/Surfaces/CylinderBounds.cpp
```diff
@@ -167,7 +167,11 @@ std::vector<Acts::Vector3> Acts::CylinderBounds::createCircles(
   if ((bevelMinZ != 0. || bevelMaxZ != 0.) && vertices.size() % 2 == 0) {
     auto halfWay = vertices.end() - vertices.size() / 2;
     double mult{1};
-    auto func = [&mult](Vector3& v) { v(2) += v(1) * mult; };
+    auto func = [&mult, &ctrans](Vector3& v) {
+      v = ctrans.inverse() * v;
```
paulgessinger, line 171: Consider caching the inverse here and capturing it alongside `ctrans`.

noraemi: fixed

Core/src/Surfaces/CylinderBounds.cpp
```diff
@@ -167,7 +167,12 @@ std::vector<Acts::Vector3> Acts::CylinderBounds::createCircles(
   if ((bevelMinZ != 0. || bevelMaxZ != 0.) && vertices.size() % 2 == 0) {
     auto halfWay = vertices.end() - vertices.size() / 2;
     double mult{1};
-    auto func = [&mult](Vector3& v) { v(2) += v(1) * mult; };
+    auto invCtrans = ctrans.inverse();
```
paulgessinger, line 170: Sorry to comment again, but I think if you use `auto` here, it won't actually do the inversion here (might be wrong). You can force this by either using a specific type or using `inverse().eval()`.

noraemi: Seems to work for me, at least dumping a cylinder (or even multiple) gives the expected result
image

paulgessinger (approved):

PR #1052, opened 2021-10-26
feat!: changed Grid interface of neighborHoodIndices

robertlangenberg:
Changed `Grid` interface of `neighborHoodIndices` with differing lower and upper neighbors. This now allows to specify the number of neighbors desired per axis, as opposed to having one number of neighbors for all axes. It also inverts the way neighbors are interpreted:

BREAKING CHANGE: The `Grid` interface changes: f you want 1 neighbor on the left and two on the right, you now have to write `{-1,2}` instead of `{1,2}` as before. This is necessary as this also allows to give "negative" neighbors, i.e. not starting from the current bin, but e.g. only 2 bins left of the current bin: `{-2,-1}`.

PR #1038, opened 2021-10-11
feat: change BinFinder logic

LuisFelipeCoelho:
This PR changes the `BinFinder` algorithm that returns the connected bins used to build seeds. Instead of always returning the 8 surrounding bins, the user may define a vector containing a map of z bins in the top and bottom sets of space points for each middle space point bin. The algorithm will only search for SP in the bins defined in the map. In case no map is defined, the algorithm returns the 8 surrounding bins for top and bottom layers.
That will allow us to configure the top and bottom maps independently form each other and also to
return the bins based on geometrical assumptions as it is done by the ATLAS algorithm, for example:

```
//  1     2     3     4     5     6     7     8     9     10     11     z bin index
// -------------------------------------------------------------------> Z[mm]
// vector containing the map of z bins in the top layer
seedingCfg.binFinderConfigTop = {{1},{1, 2},{2, 3},{3, 4},{4, 5},{5, 6, 7},{7, 8},{8, 9},{9, 10},{10, 11},{11}};
// vector containing the map of z bins in the bottom layer
seedingCfg.binFinderConfigBottom = {{1, 2},{2, 3},{3, 4},{4, 5},{5, 6},{6},{6, 7},{7, 8},{8, 9},{9, 10},{10, 11}};
```

these vectors are passed to `BinFinder` by the `SeedingAlgorithm`.

@noemina

PR #1005, opened 2021-09-22
feat: Add variable axis type for SpacePointGrid - Follow-up 

LuisFelipeCoelho:
I lost access to the PR #976 so I am creating this follow-up PR in order to fix some minor comments.

PR #1186, opened 2022-03-09
feat: ITk seeding example

LuisFelipeCoelho:
This PR adds the `ITkSeedingExample.cpp` to run the ITk seeding.
It shows how to configure the non-equidistant binning in z (#1005), the seed confirmation cuts (#1084), the radial range for middle SP cut (#1084) and the vector containing the map of z neighbours (#1052 and #1038).

It contains all the parameters to run the seeding for ITk **pixel** space points (I intend to extend this to ITk strip SPs soon).

@noemina @paulgessinger

### DIFF
--- a/Core/include/Acts/Seeding/SeedfinderConfig.hpp
+++ b/Core/include/Acts/Seeding/SeedfinderConfig.hpp
@@ -21,16 +21,16 @@ struct SeedConfirmationRange {
   float rMaxSeedConf;
   size_t nTopForLargeR;
   size_t nTopForSmallR;
-	
-	SeedConfirmationRange(float zMinSC = std::numeric_limits<float>::min(),
-												float zMaxSC = std::numeric_limits<float>::max(),
-												float rMaxSC = std::numeric_limits<float>::max(),
-												size_t nTopLargeR = 0, size_t nTopSmallR = 0)
-	: zMinSeedConf(zMinSC),
-	zMaxSeedConf(zMaxSC),
-	rMaxSeedConf(rMaxSC),
-	nTopForLargeR(nTopLargeR),
-	nTopForSmallR(nTopSmallR) {}
+
+  SeedConfirmationRange(float zMinSC = std::numeric_limits<float>::min(),
+                        float zMaxSC = std::numeric_limits<float>::max(),
+                        float rMaxSC = std::numeric_limits<float>::max(),
+                        size_t nTopLargeR = 0, size_t nTopSmallR = 0)
+      : zMinSeedConf(zMinSC),
+        zMaxSeedConf(zMaxSC),
+        rMaxSeedConf(rMaxSC),
+        nTopForLargeR(nTopLargeR),
+        nTopForSmallR(nTopSmallR) {}
 };
 
 // forward declaration to avoid cyclic dependence
--- a/Core/include/Acts/Seeding/SeedfinderConfig.hpp
+++ b/Core/include/Acts/Seeding/SeedfinderConfig.hpp
@@ -19,9 +19,18 @@ struct SeedConfirmationRange {
   float zMinSeedConf;
   float zMaxSeedConf;
   float rMaxSeedConf;
-  size_t nTopSeedConf;
   size_t nTopForLargeR;
   size_t nTopForSmallR;
+	
+	SeedConfirmationRange(float zMinSC = std::numeric_limits<float>::min(),
+												float zMaxSC = std::numeric_limits<float>::max(),
+												float rMaxSC = std::numeric_limits<float>::max(),
+												size_t nTopLargeR = 0, size_t nTopSmallR = 0)
+	: zMinSeedConf(zMinSC),
+	zMaxSeedConf(zMaxSC),
+	rMaxSeedConf(rMaxSC),
+	nTopForLargeR(nTopLargeR),
+	nTopForSmallR(nTopSmallR) {}
 };
 
 // forward declaration to avoid cyclic dependence
--- a/Examples/Run/Reconstruction/CMakeLists.txt
+++ b/Examples/Run/Reconstruction/CMakeLists.txt
@@ -119,6 +119,17 @@ target_link_libraries(
       ActsExamplesRecTracksCommon)
 # =================================================
 
+# Seeding for ITk detector
+add_executable(
+  ActsExampleSeedingITk
+  ITkSeedingExample.cpp)
+target_link_libraries(
+  ActsExampleSeedingITk
+  PRIVATE
+    ActsExamplesRecTracksCommon
+    ActsExamplesDetectorGeneric
+)
+
 install(
   TARGETS
     ActsExampleTruthTracksGeneric
@@ -129,6 +140,7 @@ install(
     ActsExampleCKFTracksTelescope
     ActsExampleSeedingGeneric
     ActsExampleSeedingTGeo
+    ActsExampleSeedingITk
   RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
 
 add_subdirectory_if(DD4hep ACTS_BUILD_EXAMPLES_DD4HEP)
--- a/Examples/Run/Reconstruction/Common/CMakeLists.txt
+++ b/Examples/Run/Reconstruction/Common/CMakeLists.txt
@@ -1,6 +1,7 @@
 add_library(
   ActsExamplesRecTracksCommon SHARED
   RecTruthTracks.cpp
+  RecInput.cpp
   RecCKFTracks.cpp
   SeedingExample.cpp
   MeasurementsToSpacepoints.cpp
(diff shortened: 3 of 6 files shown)


noemina:
Not sure we want to add the "old-style" example in the repository, also because it includes a specific ITk implementation of it. Can we try to have the example implemented using the python-binding framework?

LuisFelipeCoelho:
We should discuss if an example containing all the configuration to run the ITk seeding should be included into ACTS or just an example showing how to configure the new seeding features.

In any case, this example still does not contain the full configuration to produce a 1 to 1 comparison to ITk pixel seeds.

@noemina @paulgessinger

Examples/Scripts/Python/itk_seeding.py
```diff
@@ -0,0 +1,155 @@
...
+    )
+
+    gridConfig = acts.SpacePointGridConfig(
+        bFieldInZ=1.997244311 * u.T,
+        minPt=900 * u.MeV,
+        rMax=320 * u.mm,  # pixel: 320 mm, strip: 1000 mm
+        zMax=3000 * u.mm,
+        zMin=-3000 * u.mm,
+        deltaRMax=280 * u.mm,  # pixel: 280 mm, strip: 600 mm
+        cotThetaMax=27.2899,  # pixel: 27.2899 (4 eta), strip: 900
```
robertlangenberg, line 31: what does strip:900 mean?

Examples/Scripts/Python/itk_seeding.py
```diff
@@ -0,0 +1,155 @@
...
+        deltaRMaxTopSP=280 * u.mm,  # pixel: 280 mm, strip: 3000 mm
+        deltaRMaxBottomSP=120 * u.mm,  # pixel: 120 mm, strip: 3000 mm
+        collisionRegionMin=-200 * u.mm,
+        collisionRegionMax=200 * u.mm,
+        zMin=gridConfig.zMin,
+        zMax=gridConfig.zMax,
+        maxSeedsPerSpM=4,
+        cotThetaMax=gridConfig.cotThetaMax,
+        sigmaScattering=2,
+        radLengthPerSeed=0.09804522341059585,
```
robertlangenberg, line 65: quite... surprising precision. the amount of material depends on the forward region and other factors, so in central region is probably less, in more forward regions higher.

what is the reason for having this instead of 0.1?

Examples/Scripts/Python/itk_seeding.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def runITkSeeding(field, outputDir, s=None):
+
+    # Read input space points from input csv files
+    evReader = CsvSpacePointReader(
+        level=acts.logging.INFO,
+        inputStem="spacepoints",
+        inputCollection="pixel",
+        inputDir="acts/Examples/Scripts/Python/data_itk_seeding/CsvSpacePointsOutput_singleMu_100GeV_300evnts/",
```
paulgessinger, line 20: Consider using a temporary file. If you want to try this, i can provide a snippet.

Examples/Scripts/Python/itk_seeding.py
```diff
@@ -0,0 +1,155 @@
...
+        zMax=gridConfig.zMax,
+        maxSeedsPerSpM=4,
+        cotThetaMax=gridConfig.cotThetaMax,
+        sigmaScattering=2,
+        radLengthPerSeed=0.09804522341059585,
+        minPt=gridConfig.minPt,
+        bFieldInZ=gridConfig.bFieldInZ,
+        beamPos=acts.Vector2(0 * u.mm, 0 * u.mm),
+        impactMax=gridConfig.impactMax,
+        maxPtScattering=1000000 * u.GeV,
```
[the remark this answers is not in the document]
LuisFelipeCoelho: In Athena they don't use the maxPtScattering cut, so I just added a large value. I will change that to `float("inf")`

Examples/Scripts/Python/itk_seeding.py
LuisFelipeCoelho: I dumped that value from Athena. I will set it to 0.1, which is probably fine

Examples/Scripts/Python/itk_seeding.py
LuisFelipeCoelho: yes, please send me the snippet

Examples/Scripts/Python/itk_seeding.py
LuisFelipeCoelho: In Athena the eta cut is not applied for the strip SPs, so the strip:900 is also just a large value. I will remove this part of the comment and when I start checking the strip and overlap SPs I can set a better value for this

Core/include/Acts/Seeding/SeedfinderConfig.hpp
```diff
@@ -19,9 +19,18 @@ struct SeedConfirmationRange {
...
+
+  SeedConfirmationRange(float zMinSC = std::numeric_limits<float>::min(),
+                        float zMaxSC = std::numeric_limits<float>::max(),
+                        float rMaxSC = std::numeric_limits<float>::max(),
+                        size_t nTopLargeR = 0, size_t nTopSmallR = 0)
+      : zMinSeedConf(zMinSC),
+        zMaxSeedConf(zMaxSC),
+        rMaxSeedConf(rMaxSC),
+        nTopForLargeR(nTopLargeR),
+        nTopForSmallR(nTopSmallR) {}
```
paulgessinger, line 33: Is there a need for a constructor here? Can the defaults not be put directly in the struct member declarations?

Examples/Python/src/TrackFinding.cpp
```diff
@@ -102,8 +102,26 @@ void addTrackFinding(Context& ctx) {
...
+    ACTS_PYTHON_MEMBER(useVariableMiddleSPRange);
+    ACTS_PYTHON_MEMBER(deltaRMiddleSPRange);
+    ACTS_PYTHON_MEMBER(seedConfirmation);
+    ACTS_PYTHON_MEMBER(centralSeedConfirmationRange);
+    ACTS_PYTHON_MEMBER(forwardSeedConfirmationRange);
     ACTS_PYTHON_STRUCT_END();
     patchKwargsConstructor(c);
+
+    py::class_<SeedConfirmationRange>(m, "SeedConfirmationRange")
+        .def(py::init<float, float, float, size_t, size_t>())
```
paulgessinger, line 119: If you remove the constructor, this here would just be the default constructor.

Examples/Scripts/Python/itk_seeding.py
```diff
@@ -0,0 +1,171 @@
...
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        temp = open(tmpdirname + "/event000000000-spacepoints_pixel.csv", "w+b")
+        print(
+            "created temporary file: "
+            + tmpdirname
+            + "/event000000000-spacepoints_pixel.csv"
+        )
+        temp.write(
+            b"measurement_id,sp_type,module_idhash,sp_x,sp_y,sp_z,sp_radius,sp_covr,sp_covz\n 1,0,3139,32.67557144165039,-5.311902523040771,-47.65000152587891,33.10452270507812,0.05999999865889549,0.02999880164861679\n 2,0,3422,95.14442443847656,-15.46361255645752,-52.125,96.39286804199219,0.05999999865889549,0.01687432639300823\n 3,0,3650,102.8257064819336,-16.71612739562988,-52.67499923706055,104.1755981445312,0.05999999865889549,0.001875000074505806\n 4,0,4223,159.4266204833984,-25.91166687011719,-56.75,161.5186157226562,0.05999999865889549,0.02999880164861679\n 5,0,5015,224.07958984375,-36.37123107910156,-61.40000152587891,227.0121765136719,0.05999999865889549,0.007499700412154198\n 6,0,6023,284.1485595703125,-46.0638542175293,-65.72499847412109,287.8580932617188,0.05999999865889549,0.001875000074505806"
+        )
```
paulgessinger, line 164: Neat! If you don't want to use `b""` here, I think you can use

```python
temp = open(tmpdirname + "/event000000000-spacepoints_pixel.csv", "w+t")
```

in line 156 to open the file in text mode directly.

Core/include/Acts/Seeding/SeedfinderConfig.hpp
LuisFelipeCoelho: done, removed the constructor as you said

Examples/Python/src/TrackFinding.cpp
LuisFelipeCoelho: done

Examples/Scripts/Python/itk_seeding.py
LuisFelipeCoelho: done

paulgessinger (approved):
Ok let's go. Would you be able to follow up this PR with another one which adds this example to the python level automatic tests (so we can ensure this keeps running)?

LuisFelipeCoelho:
ok! I will work on that @paulgessinger

PR #1187, opened 2022-03-09
feat!: Allow regional seeding configuration

Corentin-Allaire:
This PR introduce the option to specify different seeding configuration for different region of the detector, those region can be defined by specifying the position in Z (`zboundaries`) and R (`rboundaries`) of the boundary between the different regions. For each of those region a different set of `RegionalParameters` will need to be defined (including a different seed filter).

This Pr also perform some clean up of the Seeding config moving/ removing some comments and removing the `deltaRMin`, `deltaRMax` and `maxSeedsPerSpM` as they where not used anywhere and would have been confusing to new users.

Finally this makes the needed changes to the python bindings to accommodate the uses of the regional parameters there (I noted that some of the lattest option that have been added to the seeding are not in python bindings, should we add them too ?)


Examples/Run/Reconstruction/Common/SeedingExample.cpp
```diff
@@ -118,23 +118,22 @@ int runSeedingExample(int argc, char* argv[],
...
   seedingCfg.gridConfig.rMax = 200._mm;
   seedingCfg.seedFinderConfig.rMax = seedingCfg.gridConfig.rMax;
 
-  seedingCfg.seedFilterConfig.deltaRMin = 1_mm;
-  seedingCfg.seedFinderConfig.deltaRMin = seedingCfg.seedFilterConfig.deltaRMin;
-  seedingCfg.seedFinderConfig.deltaRMinTopSP =
-      seedingCfg.seedFilterConfig.deltaRMin;
-  seedingCfg.seedFinderConfig.deltaRMinBottomSP =
-      seedingCfg.seedFilterConfig.deltaRMin;
+  filterConfig.deltaRMin = 1_mm;
```
robertlangenberg, line 130: isn't this parameter removed?

robertlangenberg: nvm, found my mistake. Not sure if it's actually reasonable to not distinguish between bottom and top in the SeedFilter if we do in the SeedFinder

Corentin-Allaire: I removed it from the finder since it wasn't used anymore. But this is the filter config

Plugins/Cuda/include/Acts/Plugins/Cuda/Seeding/Seedfinder.ipp
```diff
@@ -16,22 +16,43 @@ namespace Acts {
...
+  // fill the remaning region with the default RegionalParameters
+  if (m_config.regionalParameters.size() <
+      (m_config.zboundaries.size() + 1) * (m_config.zboundaries.size() + 1)) {
+    while (m_config.regionalParameters.size() <
+           (m_config.zboundaries.size() + 1) *
+               (m_config.zboundaries.size() + 1)) {
+      m_config.regionalParameters.push_back(
+          Acts::RegionalParameters<external_spacepoint_t>());
+    }
+  }
```
robertlangenberg, line 30: This isn't necessary since the cuda Seeding is hard-coded to only use regionalParamters[0], right?

Plugins/Cuda/include/Acts/Plugins/Cuda/Seeding2/SeedFinder.ipp
```diff
@@ -34,27 +34,49 @@ SeedFinder<external_spacepoint_t>::SeedFinder(
...
+  if (m_commonConfig.regionalParameters.size() <
+      (m_commonConfig.zboundaries.size() + 1) *
+          (m_commonConfig.zboundaries.size() + 1)) {
+    while (m_commonConfig.regionalParameters.size() <
+           (m_commonConfig.zboundaries.size() + 1) *
+               (m_commonConfig.zboundaries.size() + 1)) {
+      m_commonConfig.regionalParameters.push_back(
+          Acts::RegionalParameters<external_spacepoint_t>());
+    }
+  }
```
robertlangenberg, line 53: as the other CUDA code, not necessary since it's always [0]?

Plugins/Sycl/include/Acts/Plugins/Sycl/Seeding/Seedfinder.ipp
```diff
@@ -29,35 +29,52 @@ Seedfinder<external_spacepoint_t>::Seedfinder(
...
+  // fill the remaning region with the default RegionalParameters
+  if (m_config.regionalParameters.size() <
+      (m_config.zboundaries.size() + 1) * (m_config.zboundaries.size() + 1)) {
+    while (m_config.regionalParameters.size() <
+           (m_config.zboundaries.size() + 1) *
+               (m_config.zboundaries.size() + 1)) {
+      m_config.regionalParameters.push_back(
+          Acts::RegionalParameters<external_spacepoint_t>());
+    }
+  }
```
robertlangenberg, line 47: see above

Tests/UnitTests/Core/Seeding/SeedfinderTest.cpp
```diff
@@ -115,30 +115,31 @@ int main(int argc, char** argv) {
             << elapsed_read.count() << "s" << std::endl;
 
   Acts::SeedfinderConfig<SpacePoint> config;
+  Acts::RegionalParameters<SpacePoint> regionalParameters;
```
robertlangenberg, line 118: Can you add regional parameters for more than 1 region in this or another test?

robertlangenberg (changes requested):
Looks good, I'd add the regional parameters for more than one region in a test to showcase how to use this.

Plugins/Cuda/include/Acts/Plugins/Cuda/Seeding/Seedfinder.ipp
Corentin-Allaire: Your right I wanted to already add this initialisation mechanism in case someone wanted to implement the use of regionalParamters in the Cuda and Sycl plugin but I could also remove it

LuisFelipeCoelho (changes requested):
I think the itk_seeding.py example needs to be changed as well

Corentin-Allaire:
I think this is now completely outdated compared to all the recent changes in the seeding so we can probably close this .

PR #1188, opened 2022-03-09
feat: Adding inner and outer tube covers for disc approach descriptor

noemina:
This allows to fix propagation problems seen in the inclined section of the ITk tracking geometry.

FYI: @timadye @niermann999 @paulgessinger @asalzburger


paulgessinger:
I post here the two diagnostics plots from @noemina for documentation.

image_2022_03_09T16_05_48_514Z
image_2022_03_09T16_14_27_554Z

noemina:
Failures are expected as this change affects the construction of disc layers.

paulgessinger:
At least the CKF histograms seem to be identical for the ODD (that's all I have implemented right now):
[ckf.html.gz](https://github.com/acts-project/acts/files/8224592/ckf.html.gz)

asalzburger (approved):
Looks ok to me.

paulgessinger:
I think the CI failures might have been transient due to GitHub's outage yesterday.

timadye:
The failure seems to be due to changes in the root file hashes. Those tests are for the generic detector. Clearly this will change the ITk reco output. Could it also change the generic detector?

paulgessinger:
Very likely, yes. In principle it's ok if the files change, we should just do some sanity checks that everything looks reasonable.

PR #1191, opened 2022-03-11
ci: Turn on FORCE_ASSERTIONS

paulgessinger:
In #1182 I unfortunately set the wrong CMake option to turn the forces
assertions on.

This PR should fix this.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -181,7 +181,7 @@ jobs:
              && CI/check_boost_eigen_versions.sh ./build-downstream-nodeps/bin/ShowActsVersion
 
   macos:
-    runs-on: macos-10.15
+    runs-on: macos-11
     env:
       INSTALL_DIR: ${{ github.workspace }}/install
     steps:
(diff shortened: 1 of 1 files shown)


stephenswat:
I looked into this, and it seems like the ROOT people have solved the issue in [this commit](https://github.com/root-project/root/commit/acfe6dc8476f1a06aab3d31b25e4e2bfa3003a02). Updating the ROOT version on the MacOS CI machine should fix this issue.

paulgessinger:
> I looked into this, and it seems like the ROOT people have solved the issue in [this commit](https://github.com/root-project/root/commit/acfe6dc8476f1a06aab3d31b25e4e2bfa3003a02). Updating the ROOT version on the MacOS CI machine should fix this issue.
> 
> Edit: Version `510ff2e` of the MacOS dependencies (the latest one) still has an older ROOT without these fixes.

Unfortunately that is more complicated than I care to admit 😅

asalzburger (approved):
Trivial, approved.

paulgessinger:
I managed to bump the macOS dependencies, let's see if that fixes the CI.

paulgessinger:
Working now, can you reapprove @asalzburger ?

stephenswat (approved):
Going to go ahead and stamp this.

PR #1193, opened 2022-03-14
ci: Physics performance monitoring

paulgessinger:
This PR introduces a set of "physics" monitoring CI jobs based on histogram comparisons. 
Currently this runs

- CKF truth estimated
- CKF truth smeared
- CKF seeded
- truth tracking

I intend to run this for a bit and see how it behaves. If it works well, we should extend it to cover other areas as well.


paulgessinger:
Test

paulgessinger:
I think this is good to go now.

paulgessinger:
@asalzburger @robertlangenberg can we get this in? I think it would help #1188 as well as #1181.

robertlangenberg (approved):
Very clean, nice. We can adjust the thresholds as we go if they're too tight or too loose.

PR #1194, opened 2022-03-15
build: Allow python bindings to be used from install

paulgessinger:
This PR adds some changes to enable the python bindings to be used from an install. Instead of `source $BUILD/python/setup.sh` you can now do `source $INSTALL/python/setup.sh`. 

The folder `python` won't make sense if ACTS is installed into a directory alongside other projects, but I would argue in that case you would not want to enable the python bindings anyway.

### DIFF
--- a/Examples/Python/CMakeLists.txt
+++ b/Examples/Python/CMakeLists.txt
@@ -5,6 +5,8 @@ set(_python_dir "${CMAKE_BINARY_DIR}/python")
 file(MAKE_DIRECTORY ${_python_dir})
 file(MAKE_DIRECTORY ${_python_dir}/acts)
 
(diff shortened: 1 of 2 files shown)


paulgessinger:
@stephenswat can you look at this maybe? Shouldn't be overly complex.

stephenswat (approved):
Looks like a solid change, I'll stamp it. Is it still possible to run Python code from builds (rather than installs)? Not that this is necessarily such an important feature.

paulgessinger:
Yes it is, this should not affect that. Main difference is that from a build, the python files should be symlinked (to allow changing them without rebuilding) whereas they are copied for an install.

PR #1197, opened 2022-03-22
feat: Reintroduce default bin finder constructor

stephenswat:
In Acts version 15.0.0, we see support being added for dynamic bins in the bin finder. However, this also means that we no longer have useful default constructors for this class. This has caused me a few issues in terms of backwards compatibility. I think it would be a good idea to reintroduce some sane default configuration for the bin finder, and I discussed with @LuisFelipeCoelho that having an empty list of _z_ neighbours with a single _φ_ neighbour would be a good default.


LuisFelipeCoelho (approved):

paulgessinger:
Format fails.

stephenswat:
On code that I didn't touch :cry:

paulgessinger:
That is pretty weird. The CI is not setup to only format changes, this was merged the way it's formatted now in #1038, and there clang-format was happy... and has been on every commit to main since then. But here it fails, and locally on my machine clang-format also reformats that file...

What the heck?

paulgessinger (approved):
@LuisFelipeCoelho had approved before, no changes since then other than formatting.

LuisFelipeCoelho (approved):

PR #1202, opened 2022-03-23
build: Add `assert.h` override

paulgessinger:
I noticed in a few cases, if `assert.h` was not included through
`cassert`, `ACTS_FORCE_ASSERTIONS` would not do it's job because `assert` would then already be defined.
This commit thus enables a couple of additional assertions.


stephenswat (approved):
Looks good to me, can go in as is. Just for reference, if you didn't know about this you might like the macro stack that most good C pre-processors support: https://gcc.gnu.org/onlinedocs/gcc/Push_002fPop-Macro-Pragmas.html

paulgessinger:
I'm not sure how that would help here though?

PR #1203, opened 2022-03-23
refactor!: Simplify CKF SourceLinkAccessor and make fewer assumptions

paulgessinger:
This PR migrates the CKF's source link accessor away from a class with 3 necessary methods to a single delegate that is given as part of the options. At the same time it changes the lookup key from the raw geometry identifier to the current surface, which will allow navigating to experiment specific detector elements without the need to go through a map lookup.

The CKF options move from being templated on the source link accessor itself to the source link accessor **iterator type**. The delegate is then defined as returning a range (a pair of) source link iterators, and those are used. Previously, source link accessors needed to implement three methods:

```cpp
size_t count(GeometryIdentifier id) const;
std::pair<Iterator, Iterator> range(GeometryIdentifier id) const;
const SourceLink& at(Iterator it) const;
```

which is now changed to a callable like

```cpp
std::pair<Iterator, Iterator> range(const Surface&) const;
```

which is accessed through the delegate. This PR also removes the obsolete methods. In addition the accessor was required to have a public *container* member which the `findTracks` method would accept and then set on the accessor. This is now assumed to happen before calling, the delegate can contain pointer to the accessor instance with the container already set. We believe this is better suited to environments with complex access call chains (e.g. ATLAS).

BREAKING CHANGE: This changes the way source links are passed to the CKF `findTracks` method. Instead of taking a container of source links, it now assumed the CKF options have a source link accessor connected which is configured to be able to access that container.


paulgessinger:
@noemina could you have a look at this?

noemina (approved):
It looks very clear to me. Approving.

PR #1204, opened 2022-03-24
build: Download nlohmann/json zip instead of git clone

paulgessinger:
Switches from getting `nlohmann/json` sources via git to a zip file download from GitHub. For some reason, the git repo is very large and takes a while to clone, so this method is much quicker in my testing.


AJPfleger (approved):
Looks good to me

PR #1205, opened 2022-03-24
refactor: Use enum class for MaterialUpdateStage

paulgessinger:
This change is motivated by arithmetic between non-class enums and other types that are not the same enum being deprecated in C++20. This mainly affects `NavigationDirection`, but I took the chance to update the `MaterialUpdateStage` enum to make it an enum class.


Core/src/Definitions/Common.cpp
```diff
@@ -0,0 +1,33 @@
...
+    case MaterialUpdateStage::PreUpdate:
+      os << "PreUpdate (-1)";
+      break;
+    case MaterialUpdateStage::PostUpdate:
+      os << "PostUpdate (1)";
+      break;
+    case MaterialUpdateStage::FullUpdate:
+      os << "FullUpdate (0)";
+      break;
+    default:
```
stephenswat, line 26: It might be useful to mark this code as unreachable to the compiler.

Core/include/Acts/Material/ISurfaceMaterial.hpp
```diff
@@ -130,10 +130,14 @@ class ISurfaceMaterial {
...
-  if (mStage == Acts::fullUpdate) {
+  if (mStage == Acts::MaterialUpdateStage::FullUpdate) {
     return 1.;
   }
-  return (pDir * mStage > 0 ? m_splitFactor : 1. - m_splitFactor);
+  return (
+      pDir * static_cast<std::underlying_type_t<MaterialUpdateStage>>(mStage) >
+              0
+          ? m_splitFactor
+          : 1. - m_splitFactor);
```
stephenswat, line 140: This is the only thing I don't really like, it smells a little unmaintainable to me. In my opinion it would be clearer to write:

```cpp
if (mStage == Acts::MaterialUpdateStage::PreUpdate) {
    return pDir < 0 ? m_splitFactor : 1. - m_splitFactor;
} else if (mStage == Acts::MaterialUpdateStage::PostUpdate) {
    return pDir > 0 ? m_splitFactor : 1. - m_splitFactor;
} else {
    return 1.;
}
```

stephenswat (approved):
This one looks good to go as it - there is one thing I would change, but it's more of an opinion matter, so I'll stamp it and then you can decide where to go with it. :+1:

Core/include/Acts/Material/ISurfaceMaterial.hpp
paulgessinger: Fair enough, here it's not actually more concise with the multiplication. I'll change this.

Core/src/Definitions/Common.cpp
paulgessinger: That's a valid point. That wouldn't result in a compilation error with an unhandled value though right?

stephenswat: It shouldn't, no. But thinking about it, marking it as unreachable might be a little heavy-handed. Sure, the compiler can optimize your code better that way, but it can also create undefined behaviour if someone managed to pass an invalid enum to your function... They probably shouldn't, but it's possible... Maybe we should just leave it like this.

Core/include/Acts/Material/ISurfaceMaterial.hpp
paulgessinger: Updated, and also added a test for this behavior.

Core/src/Definitions/Common.cpp
paulgessinger: We try not to use these output functions in hot code paths anyway, so any potential optimization should have negligible effect.

stephenswat (approved):
Fantastic, thanks!

paulgessinger:
Updated after GSF went in, @stephenswat can you reapprove?

paulgessinger:
Actually, I now also removed the arithmetic operator overload, since that's not needed anymore after the change requested by @stephenswat.

stephenswat (approved):

PR #1206, opened 2022-03-24
refactor!: Use enum class for NavigationDirection

paulgessinger:
This change is motivated by arithmetic between non-class enums and other types that are not the same enum is deprecated in C++20. I overload the operators we use these enums in to support the same use cases.

I think this is breaking since it's part of the public propagator interface.

BREAKING CHANGE: The `NavigationDirection` enum changes to `enum class`. The values are renamed `NavigationDiretcion::forward` -> `NavigationDirection::Forward` and `NavigationDirection::backward` -> `NavigationDirection::Backward`.


Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -105,7 +105,7 @@ struct KalmanFitterOptions {
   /// @param mScattering Whether to include multiple scattering
   /// @param eLoss Whether to include energy loss
   /// @param rFiltering Whether to run filtering in reversed direction as smoothing
-  /// @param rfScaling Scale factor for the covariance matrix before the backward filtering
+  /// @param rfScaling Scale factor for the covariance matrix before the NavigationDirection::Backward filtering
```
stephenswat, line 108: This is probably an unintended find-replace artifact.

Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -206,7 +206,8 @@ struct KalmanFitterResult {
   // Measurement surfaces without hits
   std::vector<const Surface*> missedActiveSurfaces;
 
-  // Measurement surfaces handled in both forward and backward filtering
+  // Measurement surfaces handled in both NavigationDirection::Forward and
+  // NavigationDirection::Backward filtering
```
stephenswat, line 210: Same here, bad replace.

Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -319,7 +320,9 @@ class KalmanFitter {
       // - Waiting for a current surface
       auto surface = state.navigation.currentSurface;
       std::string direction =
-          (state.stepping.navDir == forward) ? "forward" : "backward";
+          (state.stepping.navDir == NavigationDirection::Forward)
+              ? "NavigationDirection::Forward"
+              : "NavigationDirection::Backward";
```
stephenswat, line 325: Same here.

Core/src/Definitions/Common.cpp
```diff
@@ -0,0 +1,30 @@
...
+  switch (navDir) {
+    case NavigationDirection::Forward:
+      os << "forward";
+      break;
+    case NavigationDirection::Backward:
+      os << "backward";
+      break;
+    default:
+      assert(false && "Invalid direction (shouldn't happen)");
+      std::abort();
```
stephenswat, line 25: Maybe add a `__builtin_unreachable` here?

Core/src/Geometry/Layer.cpp
```diff
@@ -105,7 +105,7 @@ Acts::Layer::compatibleSurfaces(
   boost::container::small_vector<SurfaceIntersection, 10> sIntersections;
 
   // fast exit - there is nothing to
-  if (!m_surfaceArray || !m_approachDescriptor || !options.navDir) {
+  if (!m_surfaceArray || !m_approachDescriptor) {
```
stephenswat, line 108: Shouldn't this be replaced by `|| options.navDir == Forward`?

Core/src/Geometry/Layer.cpp
```diff
@@ -180,7 +180,7 @@ Acts::Layer::compatibleSurfaces(
     if (sfi && detail::checkIntersection(sfi.intersection, pathLimit,
                                          overstepLimit, s_onSurfaceTolerance)) {
       // Now put the right sign on it
-      sfi.intersection.pathLength *= std::copysign(1., options.navDir);
+      sfi.intersection.pathLength *= options.navDir;
```
stephenswat, line 183: Doesn't this need to be explicitly converted to an integer (or underlying type) in this case?

Core/src/Geometry/TrackingVolume.cpp
```diff
@@ -480,8 +482,7 @@ Acts::TrackingVolume::compatibleBoundaries(
                  << bSurface->surfaceRepresentation().geometryId());
     if (detail::checkIntersection(sIntersection.intersection, pLimit, oLimit,
                                   s_onSurfaceTolerance, logger)) {
-      sIntersection.intersection.pathLength *=
-          std::copysign(1., options.navDir);
+      sIntersection.intersection.pathLength *= options.navDir;
```
stephenswat, line 485: Why can we skip the conversion here?

Core/src/Geometry/TrackingVolume.cpp
```diff
@@ -490,8 +491,7 @@ Acts::TrackingVolume::compatibleBoundaries(
       ACTS_VERBOSE("Consider alternative");
       if (detail::checkIntersection(sIntersection.alternative, pLimit, oLimit,
                                     s_onSurfaceTolerance, logger)) {
-        sIntersection.alternative.pathLength *=
-            std::copysign(1., options.navDir);
+        sIntersection.alternative.pathLength *= options.navDir;
```
stephenswat, line 494: And here?

Core/src/Geometry/TrackingVolume.cpp
```diff
@@ -656,7 +656,7 @@ Acts::TrackingVolume::compatibleSurfacesFromHierarchy(
   double pLimit = options.pathLimit;
   double oLimit = options.overstepLimit;
 
-  if (m_bvhTop == nullptr || !options.navDir) {
+  if (m_bvhTop == nullptr) {
```
stephenswat, line 659: Same comment as before, is this valid?

stephenswat (changes requested):
Massive support for going to type-safe enums here, just a few comments where I think find-replace might have gone wrong.

Core/src/Geometry/TrackingVolume.cpp
stephenswat: Never mind, I see how this works now.

Core/src/Geometry/Layer.cpp
stephenswat: No, it shouldn't -- never mind. Why can't I resolve my own comments?

Core/include/Acts/TrackFitting/KalmanFitter.hpp
asalzburger: oh, yes, you're right Stephen!

asalzburger (dismissed):
There have been some replace errors that can/should be fixed.

Core/src/Geometry/Layer.cpp
paulgessinger: I defined the operators for these types.

Core/src/Definitions/Common.cpp
paulgessinger: For string output, so you think it's necessary? Would also have to be behind a compiler aware macro, right?

Core/include/Acts/TrackFitting/KalmanFitter.hpp
paulgessinger: Should be fixed now.

paulgessinger:
I updated the branch and fixed the bad replaces.

PR #1207, opened 2022-03-25
feat: make interaction point cut optional in seedFinder

LuisFelipeCoelho:
A cut on the compatibility between the SP and the interaction point was added to the seedFinder in PR #1084

The cut may not always be necessary or wanted so this PR makes the cut optional by introducing a boolean `interactionPointCut`

This is also useful for the ITk implementation because this cut is used for the pixel SPs but not for the strip SPs


stephenswat (approved):
This looks sensible and safe to me. :+1:

robertlangenberg:
fix format in python and this goes in

paulgessinger:
Format job is broken see #1212.

PR #1208, opened 2022-03-28
ci: Clang tidy 2022

paulgessinger:
This adds a clang-tidy static analysis job to our CI. On GitHub Actions, I tested this to run for about 3h, which is clearly too long. 

I leverage the mechanism that I originally intended for the GPU CI to run this job on CERN GitLab. Here it only takes about 35min.

This is behind additional permission checks, where we maintain an allow-list of people who the CI acts for. Otherwise this check should remain neutral, allowing merging without overriding.

The CMake configuration should allow us to relatively flexibly define which checks to include in the report, and if we want we can even fail the build for others. I would propose we go about like this: 

1. This PR enables a number of checks, not all of which we intend to completely fix. 
2. We see how this behaves in practice. This shouldn't block anything from merging for now, other than due to compilation failures (i.e. clang-tidy warnings are reported but ignored)
3. If this seems to make sense, I'll add a restricted set of checks that will fail the build, and add a PR to fix these checks.

Thoughts?


paulgessinger:
@andiwand maybe you could look at this too?

andiwand (approved):

andiwand:
I fear I cannot really add anything. I think it is good to be able to run it from `cmake` because you can check locally as well. in my setup I only used `cmake` to generate and index of files that should be checked and then directly called `clang-tidy` https://github.com/opendocument-app/OpenDocument.core/blob/master/.github/workflows/tidy.yml

paulgessinger:
> I fear I cannot really add anything. I think it is good to be able to run it from `cmake` because you can check locally as well. in my setup I only used `cmake` to generate and index of files that should be checked and then directly called `clang-tidy` https://github.com/opendocument-app/OpenDocument.core/blob/master/.github/workflows/tidy.yml

I guess the main reason for putting it into cmake is that we can reasily canonically configure it and it's integrated directly with the build, thus less moving parts.

This runs as an extra (for now: optional) job alongside what we have so far.

paulgessinger:
I had to push another update to the `limits.yml` file to ignore `readability-inconsistent-declaration-parameter-name`. A PR to actually fix these will come at a later point in time.

andiwand (approved):

PR #1214, opened 2022-03-30
fix: Match seed finder util .hpp and .ipp types

stephenswat:
I've discovered that #1143 introduces a subtle but rather nefarious bug in the seed finder utilities: the `const` qualifier on one of the types was removed in the `.ipp` file, but not in the corresponding `.hpp` file. This means that any code importing the `.hpp` will gladly compile the code, but because a matching implementation is not present in the `.ipp` file, it will rely on this implementation being made available at link time. In this particular case, that never happens, and we run into rather hard to debug linking errors. This commit resolves the issue by making sure that the function signatures between the `.hpp` file and the `.ipp` file are the same.

### DIFF
--- a/Core/include/Acts/Seeding/SeedFinderUtils.hpp
+++ b/Core/include/Acts/Seeding/SeedFinderUtils.hpp
@@ -52,7 +52,7 @@ LinCircle transformCoordinates(InternalSpacePoint<external_spacepoint_t>& sp,
 /// @param[out] linCircleVec The output vector to write to.
 template <typename external_spacepoint_t>
 void transformCoordinates(
-    const std::vector<InternalSpacePoint<external_spacepoint_t>*>& vec,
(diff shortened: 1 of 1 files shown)


paulgessinger:
That is indeed interesting (and not good). Thanks!

paulgessinger (approved):

paulgessinger:
Ah I'm noticing this is changed (fixed) in https://github.com/acts-project/acts/pull/1173/ too.

@stephenswat @CarloVarni what do you prefer we do here?

stephenswat:
That would only fix the issue for Carlo's new callable-based implementation of these functions, right?

paulgessinger:
IIUC the PR also contains this fix.

stephenswat:
Oh so it does, let's go with Carlo's pull request then.

PR #1216, opened 2022-03-30
ci: Add ODD file hash checks, parametrize CKF example test

paulgessinger:
This requires a couple of changes:
1. Turn off the log failure thresholds for the Python based tests
   (otherwise setup becomes too complicated)
2. Run the physmon step **before** the python based tests: if the tests
   fail, the physmon html page will still be uploaded and accessible
   this way

I also change the implementation of the CKF tracks example test: it's
now parametrized via pytest to run the three configurations (seeded,
truth estimated, truth smeared) from before, and in addition runs it
with the ODD and the generic detector. I had to drop the numerical event
count checks for that to work, but that should be covered by the hash
check anyway (if the count changes the hash changes).

I'm also bumping the number of events for these checks from 10 to 100.

Finally, sparked by https://github.com/acts-project/acts/pull/1174, I'm
switching the operation mode of the physics monitoring to not fail the
job even if the comparisons fail. Let's try this workflow instead:

1. We try to cover the physics monitoring scenarios with hash checks
   also
2. The hash checks notify us if the output changes
3. We manually check the physics monitoring HTML file to get an idea
   whether the output changes are expected / tolerable


paulgessinger:
Let's hold this until https://github.com/acts-project/acts/pull/1174 is merged.

paulgessinger:
@asalzburger can you have a quick look at this?

CI/physmon/phys_perf_mon.sh
```diff
@@ -27,7 +27,7 @@ function run() {
     histcmp \
         --label-reference=$refcommit \
         --label-monitored=$commit \
-        "$@"
+        "$@" || true
```
stephenswat, line 30: Just out of curiosity, why is this needed if `set +e` is set?

stephenswat (approved):
Agreed that this is sensible.

CI/physmon/phys_perf_mon.sh
paulgessinger: Because I used to collect the exit codes and combine them for the script exit code. I didn't want to completely remove that combination logic, and keep it for now, so I write 1 into `$ec` here. I could also do this differently, you're correct.

PR #1217, opened 2022-03-30
refactor: Move Identification from Plugins to Core

toyamaza:
This PR moves Identification from Plugins to Core in order to use IdentifiedDetectorElement in the strip SP builder


paulgessinger:
Hmm, I don't think we want this Identifier in core actually. 

We originally removed this to ensure that ACTS code does not depend on anything specific in the Identifier. 

Can we work around this and achieve what you need a different way?

toyamaza:
Hi @paulgessinger , OK, let me open another PR for the SP builder to see what we can do instead of this PR.

toyamaza:
Closing this PR as we discussed at the meeting.

Issue #1172, opened 2022-02-18
Seeding grid phi binning has changed 


osbornjd:
The seeding algorithm has changed in a nontrivial way in regards to the way that the grid binning is determined. In previous versions, e.g. 1.1, the space point grid binning was determined by the maximum deflection angle as seen [here](https://github.com/acts-project/acts/blob/836fddd02c3eff33825833ff97d6abda5b5c20a0/Core/include/Acts/Seeding/SpacePointGrid.ipp#L37). In the latest version, there are additional components that lead to different phi binning for the same input configuration parameters due to the addition of `deltaAngleWithMaxD0` as well as the dividing by the number of phi neighbors as seen [here](https://github.com/acts-project/acts/blob/b8abfb54cbb6561337be82c6ad0dd1c12609f35b/Core/include/Acts/Seeding/SpacePointGrid.ipp#L58). Is the effect of this change completely understood? It changes the underlying binning of the grid, which may or may not impact physics and/or CPU timing performance (that is still to be determined for the case of sPHENIX, not sure if other experiments have seen changes).

paulgessinger:
This was largely the work of @LuisFelipeCoelho towards making the ACTS seeding perform (or able to be configured to perform) identically to the ATLAS seeding for ITk. 

I believe the changes are in fact understood overall (I'll let @LuisFelipeCoelho and @noemina comment on this).

We unfortunately still don't have detailed distribution level physics performance monitoring, but we do have the ROOT hash monitoring, and changes to the output were investigated.

robertlangenberg:
Joe had told me the parameter deltaAngleWithMaxD0 leads to MORE phi-bins, which would be counter-intuitive. it should be less than before. If numPhiNeighbors=1 there shouldn't be an increase in the number of phi-bins if i'm not mistaken

noemina:
As you say, the parameter deltaAngleWithMaxD0 leads to a larger expected deflection in phi and should reduce the number of bins. I think is the change in the denominator that increases the number of bins (there was no denominator in the previous version of the code).

> If numPhiNeighbors=1 there shouldn't be an increase in the number of phi-bins if i'm not mistaken

In terms of number of bins, the previous value is obtainable if numPhiNeighbors=0, not 1. This has been indeed added to configure consistently the number of bins in the grid and the bin finder.

robertlangenberg:
Hi @noemina sorry for having this discussion spread out, I'll try to recap what I said and your response on #1167. 
I said
[...]

PR #1220, opened 2022-03-31
feat: change name of numPhiNeighbors in the grid

LuisFelipeCoelho:
As @robertlangenberg and @noemina have addressed in PR #1167 and issue #1172, `numPhiNeighbors` for Grid and BinFinder don't mean the same. The `numPhiNeighbors` for the Grid is the number of phiBin neighbors that should cover the full deflection of a minimum pT particle, while `numPhiNeighbors` in the BinFinder is the number of phi bin neighbors (including the current one) that are used to search for SPs. 

This PR renames `numPhiNeighbors` for the Grid to `phiBinDeflectionCoverage` and allows for even numbers of phiBins.


Core/include/Acts/Seeding/SpacePointGrid.hpp
```diff
@@ -37,8 +37,12 @@ struct SpacePointGridConfig {
   float cotThetaMax;
   // maximum impact parameter in mm
   float impactMax;
-  // sets of consecutive phi bins in the seed making step
-  int numPhiNeighbors = 0;
+  // number of phiBin neighbors (plus the current bin) that covers the full
+  // deflection of a minimum pT particle
+  int phiBinDeflectionCoverage = 1;
```
robertlangenberg, line 42: @LuisFelipeCoelho, thanks for the name change!

You changed the calculation that uses phiBinDeflectionCoverage, which is fine, it needs to be carefully explained though: I suggest a comment like this: Multiplicator for the number of phi-bins. The minimum number of phi-bins depends on min_pt, magnetic field: 2*M_PI/(minPT particle phi-deflection). phiBinDeflectionCoverage is a multiplier for this number. If numPhiNeighbors (is this the correct naming? where is this configured?) is configured to return 1 neighbor on either side of the current phi-bin (and you want to cover the full phi-range of minPT), leave this at 3.

Core/include/Acts/Seeding/SpacePointGrid.hpp
```diff
@@ -37,8 +37,12 @@ struct SpacePointGridConfig {
...
+  // number of phiBin neighbors at each side of the current bin that will be
+  // used to search for SPs
+  int numPhiNeighbors;
```
robertlangenberg, line 45: is this still used (i don't think it is?)? if yes, set a default. if no, delete it.

Core/include/Acts/Seeding/SpacePointGrid.hpp
robertlangenberg: I'd also set the default to 3.

robertlangenberg: also, please run the seeding examples and check if they still work with the new config.

robertlangenberg (changes requested):
please address comments

Core/include/Acts/Seeding/SpacePointGrid.hpp
robertlangenberg: Thank you very much for the additional explanation and for moving  :)

could you run the seeding example (not the itk one) to see if it still works? my expectation is that it won't work as it expects the numPhiNeighbors configuration (which now doesn't exist anymore)

LuisFelipeCoelho: I tested the `seeding.py` and it worked. I think it is using the default value since there was no configuration for numPhiNeighbors in the seeding example. Should I add one configuration for numPhiNeighbors and phiBinDeflectionCoverage?

robertlangenberg (approved):

PR #1222, opened 2022-03-31
ci: Decouple LCG and Ubuntu builds

paulgessinger:
This allows getting rid of some `if/else` branching between these, and allows splitting up the follow up checks (examples+pytest and phys monitoring) into separate jobs that depend on the build job result. This means they can run independently of each other and if either fails, the other will still run and provide output.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -121,10 +121,13 @@ jobs:
       - name: Install
         run: cmake --build build -- install
 
+      - name: Package build
+        run: tar czf ../build.tar.gz -C build . 
+
       - uses: actions/upload-artifact@v3
         with:
           name: acts-linux-ubuntu
-          path: build
+          path: build.tar.gz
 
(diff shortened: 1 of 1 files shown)


stephenswat:
Hmm, so it looks like the CI is still trying to run the old jobs, which it now no longer has the configuration for anymore? Or how does that work?

paulgessinger:
"Required checks" in the PR UI is a UI thing, that's not defined by the job description. I'll have to override this and change the settings if we merge this.

stephenswat:
Ah, that seems... Confusing on Github's part. :thinking:

paulgessinger:
Ok, fixed conflicts. Can we go again @stephenswat ?

stephenswat (approved):
Re-approving.

paulgessinger:
Ok, required old checks are still listed, new ones have passed. Overriding check requirement to merge this then.

PR #1223, opened 2022-04-04
fix: Kalman Filter with ATLAS Stepper

Corentin-Allaire:
Fix the issues related to the ATLASStepper so that it is usable with the KF :
- The pVector is the updated in `update` instead of `resetState` it seems this is the equivalent of the jacToGlobal in the EigenStepper
- The pVector is not updated when calling `transportCovarianceToBound` this would let to the covariance changing depending on how many time this method was called
- The ATLAS Stepper now create state with the transpose of its jacobian (which correspond to the same jacobian as the Eigen Stepper one)
- With this implementation the stepper curently doesn't work when encountering non-measurement surfaces, this is probably due to some kind of reinitialisation not being perform correctly ? This should be investigated.

And the `PropagationCompareAtlasEigenConstant` now also compares the covariances.


asalzburger:
We may override the codecov, this is literally shifting code around, weird why this results in a negative codecov.

PR #1224, opened 2022-04-05
ci: GPU-CI for Exa.TrkX plugin

benjaminhuth:
This PR tries to get the GPU-CI for the Exa.TrkX plugin working.

For now the ONNX-model-files are retrieved by the job via HTTP.


.gitlab-ci.yml
```diff
@@ -0,0 +1,41 @@
...
+  image: ghcr.io/acts-project/ubuntu2004_exatrkx:v17
+  artifacts:
+    paths:
+      - build/
+      - src/Examples/Python
+      - src/Examples/Scripts/Python
+      - src/Examples/Algorithms/TrackFinding/share
+      - src/Examples/Algorithms/Digitization/share
+    exclude:
+      - build/**/*.o
```
paulgessinger, line 12: This works, another option would be to add an install step and package up the install directory, and then running from that.

Up to you though.

.gitlab-ci.yml
```diff
@@ -0,0 +1,41 @@
+build:
+  stage: build
+  image: ghcr.io/acts-project/ubuntu2004_exatrkx:v17
```
paulgessinger, line 3: I think I'd probably also add

```yml
  tags:
    - docker
```

otherwise I believe it isn't *guaranteed* to run on a docker runner.

paulgessinger:
Would it be useful to add an output hash check on the output root file? That would maybe give us a heads-up if it changes? Doesn't have to be through pytest, but could be. This can also be done in a separate PR to get this in sooner.

benjaminhuth:
I think since the current ONNX files are still the "dummy" ones, that do not produce useful output. Once this is fixed, we should definitively add the hash checks, but I would do this in a follow-up PR

paulgessinger (approved):
Ok let's go.

PR #1226, opened 2022-04-06
feat: Python bindings for Telescope Detector

### LINKED ISSUE #1225: feat: Python bindings for Telescope Detector
Dear all,
I noticed that there are no Python bindings for the Telescope Detector. I'll open a PR shortly.

r4lv:
closes #1225.

I also included a small example script, which uses the `runTruthTracking()` from `truth_tracking.py`.

### DIFF
--- a/Examples/Scripts/Python/truth_tracking_telescope.py
+++ b/Examples/Scripts/Python/truth_tracking_telescope.py
@@ -1,0 +1,33 @@
+#!/usr/bin/env python3
+import os
+from pathlib import Path
+
+import acts
+import acts.examples
+
+from truth_tracking import runTruthTracking
+
+u = acts.UnitConstants
+
+if "__main__" == __name__:
+
+    srcdir = Path(__file__).resolve().parent.parent.parent.parent
+
+    # detector, trackingGeometry, _ = getOpenDataDetector()
+    # detector, trackingGeometry, decorators = acts.examples.GenericDetector.create()
+    detector, trackingGeometry, decorators = acts.examples.TelescopeDetector.create(
+        bounds=[200, 200],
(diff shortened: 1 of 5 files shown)


Examples/Scripts/Python/truth_tracking_telescope.py
```diff
@@ -0,0 +1,32 @@
...
+
+    # detector, trackingGeometry, _ = getOpenDataDetector()
+    # detector, trackingGeometry, decorators = acts.examples.GenericDetector.create()
+    detector, trackingGeometry, decorators = acts.examples.TelescopeDetector.create(
+        bounds=[200, 200], positions=[30, 60, 90, 120, 150, 180, 210, 240, 270]
+    )
+
+    field = acts.ConstantBField(acts.Vector3(0, 0, 2 * u.T))
+
+    runTruthTracking(
```
paulgessinger, line 24: So for the truth tracking there is no pre-configured `addTruthTracking`. I guess for a use case like this, reusing `runTruthTracking` makes sense. What do you think @timadye?

timadye: Sorry, I didn't yet look into the truth tracking, so didn't make an `addTruthTracking`. Maybe I should add that to the list of improvements.
But since `runTruthTracking` does everything needed, maybe we can keep this with a `runTruthTracking` call even when there is the option to use `addTruthTracking`. Or we can switch over.

robertlangenberg: I'll interpret this as "no strong preference" and go ahead with this PR

robertlangenberg (approved):

PR #1227, opened 2022-04-06
feat: change rMax invalid_argument to warning

LuisFelipeCoelho:
This PR allows for rMax in the grid to be different than the one used in `BinnedSPGroup`. Instead of throwing std::invalid_argument it just dumps a warning.

This is useful for ITk strip SPs. But in case we don't want to change the invalid_argument check, another option would be to add another parameter in the calculation of `numRBins = (config.rMax + config.beamPos.norm())` that allows for an increase or decrease of this value.

@noemina @robertlangenberg @paulgessinger


robertlangenberg (approved):

robertlangenberg:
Shis makes sense if you want to use the same grid for both pixel and strips, which you could (at least after this PR), but it would make the execution slower, as the grid would return SpacePoints which aren't used by the Seedfinder. The idea was to have a grid with only the needed SP.

Though I guess the performance overhead is low if the SP are ordered in r and you break out of loops when you reach SP outside of rMax.

robertlangenberg (changes requested):
Sorry, I rescind the approval, I think a "correct" configuration shouldn't produce a warning, your suggestion to add the beamSpot.norm() to the rBins is a much nicer approach.

robertlangenberg:
Do the ITK strips only require to increase the rBins by the beamSpot.norm? (since you also talked about reducing the number of rBins)

LuisFelipeCoelho:
@robertlangenberg the beamSpot.norm() is already there, I thought I could add one more parameter besides beamSpot.norm(): `numRBins = (config.rMax + config.beamPos.norm() + config.newParameter)`

Yes, the ITK strips only require an increase to the rBins.

Examples/Algorithms/TrackFinding/src/SeedingAlgorithm.cpp
```diff
@@ -40,7 +40,9 @@ ActsExamples::SeedingAlgorithm::SeedingAlgorithm(
   }
 
   if (m_cfg.gridConfig.rMax != m_cfg.seedFinderConfig.rMax) {
-    throw std::invalid_argument("Inconsistent config rMax");
+    ACTS_WARNING(
```
paulgessinger, line 43: One consideration is that in **principle** we want the CI and automated tests to fail if there's any log message of level WARNING or above (we currently have some exceptions but in principle we don't want to add more).

Could we make this an extra flag like `m_cfg.allowSeparateRMax` and if that is true, no output happens and if false we print this warning (or go back to the exception)?

paulgessinger (changes requested):

Examples/Algorithms/TrackFinding/src/SeedingAlgorithm.cpp
robertlangenberg: I think then we could just remove the warning altogether, this warning/exception only helps someone who didn't expect that this should be the same (in most cases), if there's a parameter you'd need to switch to find out, you may as well just check your configuration.

@LuisFelipeCoelho, maybe you can explain under which circumstances a higher rMax in the seedFinder than in the Grid makes sense, given that the Grid's rMax removes all SP outside of this area.

LuisFelipeCoelho: The rMax in the grid is only used to calculate deltaPhi of the Grid in `SpacePointGrid.ipp`. Then, rMax in the seedFinder is used in `BinnedSPGroup.ipp` to remove all SP outside the region defined by rMax.

In Athena, rMax is only used to calculate deltaPhi of the Grid and there is no r limit to SPs that are used by the Seedfinder. I am not sure why Athena does that, but If we set them equal and limit the SPs we are removing measurements that are used by Athena, and if we increase the grid rMax we change the grid and the seeds are different. 

I don't mind adding an extra flag `m_cfg.allowSeparateRMax`

robertlangenberg: sounds good, let's add the additional flag m_cfg.allowSeparateRMax and only throw the warning if that's not True.

robertlangenberg (approved):

paulgessinger (approved):

PR #1229, opened 2022-04-08
feat: read ITk strip measurement details with CsvSpacePointReader

LuisFelipeCoelho:
This PR adds an option to read strip coordinate information with CsvSpacePointReader by setting `extendCollection` to true.

These parameters are passed to a struct `DoubleMeasurementDetails` defined in InternalSpacePoints.

I will follow up this PR with another one which use this strip coordinate information in the seedFinder.

@noemina @paulgessinger @robertlangenberg


LuisFelipeCoelho:
Hey @konradkusiak97, could you help me dealing with CUDA and SYCL here?

konradkusiak97:
> Hey @konradkusiak97, could you help me dealing with CUDA and SYCL here?

Sure! I will have a look at it

konradkusiak97:
I think that there are still some issues while building just the UnitTests from Core. I get the following error message without enabling the CUDA and SYCL plugins: 
```cpp
[ 32%] Building CXX object Tests/CommonHelpers/CMakeFiles/ActsTestsCommonHelpers.dir/Acts/Tests/CommonHelpers/TestSourceLink.cpp.o
In file included from /home/kkusiak/acts_test/acts/Tests/UnitTests/Core/Seeding/SeedfinderTest.cpp:9:
In file included from /home/kkusiak/acts_test/acts/Core/include/Acts/Seeding/BinFinder.hpp:11:
In file included from /home/kkusiak/acts_test/acts/Core/include/Acts/Seeding/SpacePointGrid.hpp:12:
/home/kkusiak/acts_test/acts/Core/include/Acts/Seeding/InternalSpacePoint.hpp:122:47: error: no member named 'topHalfStripLength' in 'SpacePoint'
  m_doubleMeasurement.topHalfStripLength = sp.topHalfStripLength();
                                           ~~ ^
/usr/lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/bits/unique_ptr.h:835:34: note: in instantiation of member function 'Acts::InternalSpacePoint<SpacePoint>::InternalSpacePoint' requested here
    { return unique_ptr<_Tp>(new _Tp(std::forward<_Args>(__args)...)); }
                                 ^
```
so that's only with `ACTS_BUILD_UNITTESTS=ON`. 
The errors while building the code with SYCL and CUDA plugins seemed to have a similar source of errors so if this will be fixed, I think that it might also work for SYCL and CUDA. Let me know if I'm missing something here :stuck_out_tongue:

Core/include/Acts/Seeding/InternalSpacePoint.hpp
```diff
@@ -42,6 +53,27 @@ class InternalSpacePoint {
...
+  }
+  const Acts::Vector3& bottomStripDirection() const {
+    return m_doubleMeasurement.bottomStripDirection;
+  }
+  const Acts::Vector3& stripCenterDistance() const {
+    return m_doubleMeasurement.stripCenterDistance;
+  }
+  const Acts::Vector3& bottomStripCenterPosition() const {
+    return m_doubleMeasurement.bottomStripCenterPosition;
+  }
```
stephenswat, line 73: We will probably want to return these as non-reference objects; the performance benefit of returning such small objects by reference is probably near zero.

LuisFelipeCoelho: done!

Core/include/Acts/Seeding/InternalSpacePoint.hpp
```diff
@@ -42,6 +53,27 @@ class InternalSpacePoint {
...
+  const Acts::Vector3 bottomStripDirection() const {
+    return m_doubleMeasurement.bottomStripDirection;
+  }
+  const Acts::Vector3 stripCenterDistance() const {
+    return m_doubleMeasurement.stripCenterDistance;
+  }
+  const Acts::Vector3 bottomStripCenterPosition() const {
+    return m_doubleMeasurement.bottomStripCenterPosition;
+  }
```
paulgessinger, line 73: I don't think returning `const` values is necessary.

Core/include/Acts/Seeding/InternalSpacePoint.hpp
```diff
@@ -66,6 +98,9 @@ class InternalSpacePoint {
                             // for. Quality can be changed if the space point is
                             // used for a better quality seed.
   const SpacePoint& m_sp;   // external space point
+
+  DoubleMeasurementDetails
+      m_doubleMeasurement;  // double measurement details for strip coordinates
```
paulgessinger, line 103: It seems pretty heavy to have this attached to every single SP, even if it's not a strip SP. Is there a better way to do this? Maybe there's some kind of type-erasure mechanism that could be used.

paulgessinger (commented):
I have some reservations about this design. Especially since we don't intend to do SPP or SSP seeds (I think), it seems like a pretty big overhead to attach strip SP information to all internal space points.

Core/include/Acts/Seeding/InternalSpacePoint.hpp
noemina: Another option would be to add a measurement details container, and having the InternalSpacePoint storing the pointer to the container (if any) and an integer with the location to access.
Would this be a better approach?

Core/include/Acts/Seeding/InternalSpacePoint.hpp
```diff
@@ -83,6 +118,16 @@ inline InternalSpacePoint<SpacePoint>::InternalSpacePoint(
   m_r = std::sqrt(m_x * m_x + m_y * m_y);
   m_varianceR = variance.x();
   m_varianceZ = variance.y();
+
+  m_doubleMeasurement.topHalfStripLength = sp.topHalfStripLength();
```
noemina, line 122: Was wondering about these lines... you read directly the following quantities from the space point sp.
Maybe we can use the same approach as done for the global position and the variance using a std::function that is passed to the BinnedSPGroup to extract the additional details from the space point:

https://github.com/acts-project/acts/blob/db7f497303712fff32cf418646737a205a6e1adc/Examples/Algorithms/TrackFinding/src/SeedingAlgorithm.cpp#L140-L158

Of course we need a way to save and return the quantities.

Core/include/Acts/Seeding/InternalSpacePoint.hpp
LuisFelipeCoelho: done

Core/include/Acts/Seeding/InternalSpacePoint.hpp
robertlangenberg: I'm working on replacing the internal SP with a Delegate extractors set in the configuration, which retrieve information (e.g. covariance and x,y,z coordinates) from the external input SP. To support the functionality in this PR we could add an extra Delegate that retrieves this information from the external SP.

LuisFelipeCoelho:
I am closing this PR since these changes are already included in #1255

PR #1230, opened 2022-04-11
feat: adding python based full chain example for the ODD

asalzburger:
This PR adds a python based full chain example for the Open Data Detector, very along the lines of the ITk example.


asalzburger:
I need to change the |eta| range +/- 4 does not make sense

Examples/Scripts/Python/seeding.py
```diff
@@ -280,6 +280,7 @@ def customLogLevel(custom: acts.logging.Level = acts.logging.INFO):
             outputProtoTracks="prototracks_estimated",
             trackingGeometry=trackingGeometry,
             magneticField=field,
+            initialVarInflation=initialVarInflation,            
             **acts.examples.defaultKWArgs(
```
timadye, line 284: The default `initialVarInflation` is `None`, which probably isn't understood by `TrackParamsEstimationAlgorithm()`. You can easily fix it by moving the setting inside the `defaultKWArgs()`.
```suggestion
            **acts.examples.defaultKWArgs(
               initialVarInflation=initialVarInflation,            
```

timadye:
If we are adding option parsing now, should we also do it for full_chain_itk.py, or as a separate PR?

Other options to consider:
* set `logLevel`
* Allow Pythia8 or evgen input files (eg. options for HS process, pileup, and input filename)

paulgessinger:
Personally I would restrict it to be the barest possible minimum. I can remove it again if you feel like this should be consistent across the "full chain" examples.

timadye:
I had another idea. Could we have a `common.py` function to define `Sequence` and/or evgen inputs (particle gun, Pythia8, or file input) based on command-line options. That could be used in `full_chain_itk.py` and `full_chain_odd.py` without adding much extra code to distract from the example they give. It would still be "slide-ware" 😄

Examples/Scripts/Python/seeding.py
paulgessinger: Ok I moved it.

paulgessinger: Ok I moved it.

paulgessinger:
I removed the command line arguments and did some cleanup.

timadye (approved):
looks good :)

PR #1232, opened 2022-04-20
feat: GSF python example

benjaminhuth:
This PR proposes a first version of an GSF example. The implementation of this required some bigger changes in the `TrackFittingAlgorithm`, since I wanted the TrackFittingAlgorithm to support different fitters with different options.

- I split the "algorithm-specific" options from the "fitter-specific" options. This now looks (in python) like this (for the Kalman Fitter):
```python
  kalmanOptions = {
      "multipleScattering": True,
      "energyLoss": True,
      "reverseFilteringMomThreshold": reverseFilteringMomThreshold,
  }

  fitAlg = acts.examples.TrackFittingAlgorithm(
      level=acts.logging.INFO,
      inputMeasurements=digiAlg.config.outputMeasurements,
      inputSourceLinks=digiAlg.config.outputSourceLinks,
      inputProtoTracks=inputProtoTracks,
      inputInitialTrackParameters=smearAlg.config.outputTrackParameters,
      outputTrajectories="trajectories",
      directNavigation=directNavigation,
      pickTrack=-1,
      dFit=acts.examples.TrackFittingAlgorithm.makeKalmanFitterFunction(
          field, **kalmanOptions
      ),
      fit=acts.examples.TrackFittingAlgorithm.makeKalmanFitterFunction(
          trackingGeometry, field, **kalmanOptions
      ),
  )
  s.addAlgorithm(fitAlg)
```

- The extensions (with exception of the `MeasurementCalibrator`) live now inside the `TrackFitterFunction`, since there may be different extensions required by different fitters. The `MeasurementCalibrator` still must live in the TrackFittingAlgorithm, since it must be constructed with the measurements, which are obviously not known when constructing the `TrackFitterFunction`.

Besides this there are some minor changes in the GSF (moved some config structs from `Acts::detail` to `Acts`, ...), but should be uncontroversial


Examples/Python/src/TrackFitting.cpp
```diff
@@ -56,15 +56,37 @@ void addTrackFitting(Context& ctx) {
             .def(py::init<const Alg::Config&, Acts::Logging::Level>(),
                  py::arg("config"), py::arg("level"))
             .def_property_readonly("config", &Alg::config)
-            .def_static("makeTrackFitterFunction",
+            .def_static("makeKalmanFitterFunction",
                         py::overload_cast<
```
paulgessinger, line 60: So I used overloads to distinguish between directed and navigation with tracking geometry. Since this now doubles the number functions, do you think it would be better to check if the tracking geometry pointer is null, and use that to decide which navigator to set up?

Don't have a strong opinion on this, just a thought.

Examples/Scripts/Python/gaussian_sum_filter.py
```diff
@@ -0,0 +1,168 @@
...
+from typing import Optional, Union
+
+from acts.examples import Sequencer, GenericDetector, RootParticleReader
+
+import acts
+
+from acts import UnitConstants as u
+
+
+def runCKFTracks(
```
paulgessinger, line 12: Following the `addXXX`/`runXXX` pattern, you could add a `addGSFTracks` function that only configures the GSF, and then a `runGSFTracks` function which configures and example chain (and would be used in a test)

Examples/Scripts/Python/gaussian_sum_filter.py
```diff
@@ -0,0 +1,168 @@
...
+        digiConfigFile=digiConfigFile,
+        rnd=rnd,
+    )
+
+    selAlg = acts.examples.TruthSeedSelector(
+        level=acts.logging.INFO,
+        ptMin=500 * u.MeV,
+        nHitsMin=9,
+        inputParticles="particles_initial",  # fatrasCfg.outputParticlesInitial,
+        inputMeasurementParticlesMap="measurement_particles_map",  # digiCfg.outputMeasurementParticlesMap,
```
paulgessinger, line 83: Comments probably not needed.

Examples/Scripts/Python/gaussian_sum_filter.py
```diff
@@ -0,0 +1,168 @@
...
+        outputProtoTracks="prototracks",
+    )
+    s.addAlgorithm(truthTrkFndAlg)
+
+    gsfOptions = {
+        "maxComponents": 12,
+        "abortOnError": False,
+        "disableAllMaterialHandling": False,
+    }
```
paulgessinger, line 122: These could probably just go directly into the `makeGsfFitterFunction`?

Examples/Scripts/Python/gaussian_sum_filter.py
```diff
@@ -0,0 +1,168 @@
...
+
+    s = addDigitization(
+        s,
+        trackingGeometry,
+        field,
+        digiConfigFile=digiConfigFile,
+        rnd=rnd,
+    )
+
+    selAlg = acts.examples.TruthSeedSelector(
```
paulgessinger, line 78: Could this use `addSeeding` and use its arguments to configure this seeding strategy?

Examples/Scripts/Python/gaussian_sum_filter.py
```diff
@@ -0,0 +1,168 @@
...
+
+    s.addAlgorithm(gsfAlg)
+
+    return s
+
+
+if "__main__" == __name__:
+    srcdir = Path(__file__).resolve().parent.parent.parent.parent
+
+    detector, trackingGeometry, decorators = GenericDetector.create()
```
paulgessinger, line 147: Have you tested this on ODD at this point? (GenericDetector as default here is fine though)

paulgessinger (commented):
Looks reasonable overall, I've added a few comments here and there.

At the python level, there could be some cleanup. I'm also wondering if at this point we should rename `truth_tracking.py` to make it obvious it runs the KF.

Aside from that: do you think you could look into running the GSF example from  a pytest test, so we run it in the CI and do basic checks on the output? Do you have any kind of code / tooling to produce validation plots for the GSF? If so, that could potentially also be added and monitored by the CI.

Examples/Algorithms/TrackFitting/include/ActsExamples/TrackFitting/TrackFittingAlgorithm.hpp
```diff
@@ -34,6 +34,17 @@ class TrackFittingAlgorithm final : public BareAlgorithm {
 
   using TrackFitterResult = Acts::Result<Acts::KalmanFitterResult>;
 
+  /// General options that do not depend on the fitter type, but need to be
+  /// handed over by the algorithm
+  struct GeneralFitterOptions {
```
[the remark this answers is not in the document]
benjaminhuth: `FitterBaseOptions` sounds good to me!

Examples/Scripts/Python/gaussian_sum_filter.py
benjaminhuth: they are leftovers from a former version, I'll remove them

Examples/Scripts/Python/gaussian_sum_filter.py
benjaminhuth: With the binaries, yes. With the python version not yet...

Examples/Python/src/TrackFitting.cpp
benjaminhuth: Hmm with the current setup this is difficult on the C++ side, because these functions have different return types and thus cannot be overloaded.
Or do you have something else in mind?

benjaminhuth:
I did some renaming at this point as @paulgessinger mentioned above (`truth_tracking.py` -> `kalman_tracks.py`, `gaussian_sum_filter.py`-> `gsf_tracks.py`), and added the functions `addKalmanTracks(...)` as well as `addGsfTracks(...)`.

I think this is now more in line with  the other examples

paulgessinger:
Changes look good to me in principle. There's some build failures right now though.

paulgessinger:
Note that there are some CI issues that should be fixed by #1243.

paulgessinger (approved):

PR #1233, opened 2022-04-21
refactor: Add nonlinear correction in KF

XiaocongAi:
This PR refactors a few methods of the stepper related with transforming from free to bound parameters. It adds the option for correction of non-linear effects during the transformation using a few sigma points (https://doi.org/10.1117/12.280797).


XiaocongAi:
Hi @asalzburger , it's the PR for the non-linear KF. Could you help have a look?

Core/include/Acts/EventData/detail/CorrectedTransformationFreeToBound.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include "Acts/Definitions/Common.hpp"
+#include "Acts/EventData/detail/TransformationFreeToBound.hpp"
+
+namespace Acts {
+namespace detail {
+
+/// @brief Parameters sampling based on covariance matrix sqrt root in UKF: https://doi.org/10.1117/12.280797
+///
+///@todo make the parameters configurable
+struct CorrectedFreeToBoundTransformer {
```
asalzburger, line 20: The Idea here is to construct this 
`CorrectedFreeToBoundTransform{0.2,4.}` I guess?

Core/include/Acts/Propagator/AtlasStepper.hpp
```diff
@@ -580,13 +580,15 @@ class AtlasStepper {
   /// @param [in] state State that will be presented as @c BoundState
   /// @param [in] surface The surface to which we bind the state
   /// @param [in] transportCov Flag steering covariance transport
+  /// @param [in] doFreeToBoundCorrection Flag steering non-linearity correction during global to local transform
```
asalzburger, line 583: as a simple naming, I think we usually do not call things `bool doFooCorrection` but rather `bool applyFooCorrection` or simply `bool correctFoo`.

Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -107,14 +107,16 @@ struct KalmanFitterOptions {
...
   KalmanFitterOptions(const GeometryContext& gctx,
                       const MagneticFieldContext& mctx,
                       std::reference_wrapper<const CalibrationContext> cctx,
                       KalmanFitterExtensions extensions_, LoggerWrapper logger_,
                       const PropagatorPlainOptions& pOptions,
                       const Surface* rSurface = nullptr,
                       bool mScattering = true, bool eLoss = true,
-                      bool rFiltering = false, double rfScaling = 1.0)
+                      bool rFiltering = false, double rfScaling = 1.0,
+                      bool gToLCorrection = false)
```
asalzburger, line 119: also here, probably we should take one name and carry it through consistently.

Core/src/EventData/CorrectedTransformationFreeToBound.cpp
```diff
@@ -0,0 +1,132 @@
...
+    const Acts::GeometryContext& geoContext, NavigationDirection navDir) {
+  // Get the incidence angle
+  Vector3 dir = freeParams.segment<3>(eFreeDir0);
+  Vector3 normal = surface.normal(geoContext);
+  ActsScalar absCosIncidenceAng = std::abs(dir.dot(normal));
+  // No correction if the incidentAngle is small enough (not necessary ) or too
+  // large (correction could be invalid).
+  if (absCosIncidenceAng < cosIncidentAngleMinCutoff or
+      absCosIncidenceAng > cosIncidentAngleMaxCutoff) {
+    return std::nullopt;
```
asalzburger, line 27: I guess the `nullopt` means to fall back to the normal freeToBound transformation?

Core/src/EventData/CorrectedTransformationFreeToBound.cpp
```diff
@@ -0,0 +1,132 @@
...
+  size_t sampleSize = 2 * eFreeSize + 1;
+  std::vector<std::tuple<FreeVector, ActsScalar, ActsScalar>> sampledFreeParams;
+  sampledFreeParams.reserve(sampleSize);
+
+  // Initialize the covariance sqrt root matrix
+  FreeSymMatrix covSqrt = FreeSymMatrix::Zero();
+  // Get the covariance sqrt root matrix
+  Eigen::JacobiSVD<FreeSymMatrix> svd(
+      freeCovariance, Eigen::ComputeThinU | Eigen::ComputeThinV);
+  // U*S*V^-1
```
asalzburger, line 39: Let's just make sure the variable names here map to the paper, that would be great as a documentation help.

Core/src/EventData/CorrectedTransformationFreeToBound.cpp
```diff
@@ -0,0 +1,132 @@
...
+    }
+    D(i, i) = std::sqrt(S(i));
+  }
+
+  FreeMatrix UP = FreeMatrix::Zero();
+  for (unsigned i = 0; i < eFreeSize; ++i) {
+    for (unsigned j = 0; j < eFreeSize; ++j) {
+      UP(i, j) = U(i, j);
+    }
+  }
```
asalzburger, line 55: Isn't that just an assignment, or at least a block-assignment? From `auto U = svg.matrixU()` I haven't seen what type `U` is, but I suppose what can always set `UP = U.block<eFreeSize,eFreeSize>(0,0);`.

Core/src/EventData/CorrectedTransformationFreeToBound.cpp
```diff
@@ -0,0 +1,132 @@
...
+    sampledFreeParams.push_back(
+        {freeParams - covSqrt.col(i) * gamma, 0.5 / kappa, 0.5 / kappa});
+  }
+
+  // Initialize the mean of the bound parameters
+  BoundVector bpMean = BoundVector::Zero();
+  // Initialize the bound covariance
+  BoundSymMatrix bv = BoundSymMatrix::Zero();
+
+  // The transformed bound parameters
```
asalzburger, line 79: That's transform bound + weight I suppose? Let's comment as such.

Core/src/Propagator/detail/CovarianceEngine.cpp
```diff
@@ -334,17 +333,46 @@ void transportCovarianceToBound(
...
-                    fullTransportJacobian.transpose();
+  bool correction = false;
+  if (doFreeToBoundCorrection) {
+    BoundToFreeMatrix startBoundToFinalFreeJacobian =
+        freeTransportJacobian * boundToFreeJacobian;
+    FreeSymMatrix freeCovariance = startBoundToFinalFreeJacobian *
+                                   boundCovariance *
+                                   startBoundToFinalFreeJacobian.transpose();
+
+    auto transformer = detail::CorrectedFreeToBoundTransformer();
```
asalzburger, line 352: I one would like to change the tuning parameters, this would be the place, right?

If that's so, we'd probably need to introduce a small struct 
```c++

struct FreeToBoundCorrection {

   bool apply = false;
  
 ActsScalar alpha = 0.1;
 
 ActsScalar beta = 2;

 ActsScalar& operator=(bool applyIt) {
  apply = applyIt;
   return (*this); 
}

 bool operator()() const {
   return apply;
}

};
```

Then this can be default instantiated in the method signatures with `const FreeToBoundCorrection& tfb = false` (hence we can drop the lang name because the struct type tells you what it is), but can be used to create a corrector as well, extracting the tunable parameters.

asalzburger (changes requested):
Very nicely integrated, please check if the introduction of a small struct that carries a bool plus the unable parameters wouldn't make it even a step clearer - I suppose so!

Core/src/Propagator/detail/CovarianceEngine.cpp
asalzburger: To clarify, this struct could/should replace the `bool doFreeToBoundCorrecction` with a boolean to indicate if the correction is to be done, and with the parameters for the correction (in case the boolean is true).

XiaocongAi: I see! Where should we define this struct?

asalzburger: I think it should be defined either in a dedicated file, or in the `CorrectedFreeToBound` file.

Core/include/Acts/EventData/detail/CorrectedTransformationFreeToBound.hpp
XiaocongAi: I have added the constructor using the parameters explicitly.

Core/include/Acts/Propagator/AtlasStepper.hpp
XiaocongAi: Changed to `freeToBoundCorrection` following the struct type `FreeToBoundCorrection`.

Core/src/EventData/CorrectedTransformationFreeToBound.cpp
XiaocongAi: Yes, it meas the normal freeToBound transformation will be used. Comment has been added.

Core/src/EventData/CorrectedTransformationFreeToBound.cpp
XiaocongAi: OK. Fixed now.

Core/src/EventData/CorrectedTransformationFreeToBound.cpp
XiaocongAi: This was a bit stupid. Actually,  we can just use the `U`.

Core/src/EventData/CorrectedTransformationFreeToBound.cpp
XiaocongAi: Fixed.

Core/src/Propagator/detail/CovarianceEngine.cpp
XiaocongAi: Thanks. I added a small struct `FreeToBoundCorrection` as suggested.

XiaocongAi:
Hi @asalzburger , thank you a lot for the suggestions. I should have resolved all the comments.

Examples/Algorithms/TrackFitting/include/ActsExamples/TrackFitting/TrackFittingAlgorithm.hpp
```diff
@@ -92,6 +93,8 @@ class TrackFittingAlgorithm final : public BareAlgorithm {
     bool multipleScattering = true;
     /// Some more detailed steering - correct for e-loss
     bool energyLoss = true;
+    // Correction for non-linearity effect during transform from free to bound
+    Acts::detail::FreeToBoundCorrection freeToBoundCorrection = false;
```
paulgessinger, line 97: Honestly, I don't like this auto-construction from a boolean. We've had this in various spots in the code, but my understanding was we want to try to reduce this. Personally, I think an explicit constructor (so you have to type out the name) with (potentially optional) arguments would be preferrable.

@asalzburger did you have an opinion on this pattern?

Core/include/Acts/EventData/detail/CorrectedTransformationFreeToBound.hpp
asalzburger: Yes, please move it out of the `detail` namespace though.

Examples/Algorithms/TrackFitting/include/ActsExamples/TrackFitting/TrackFittingAlgorithm.hpp
asalzburger: Fine with me if that's the consensus, I found it useful for simple on/off flags, but I agree it can be misleading as it is not explicit.

So, we can remove the assignment operator and make a default constructor that sets the internal boolean to false.

asalzburger (changes requested):
A few more things: Paul's suggestion of not auto-constructing from pool & the moving out of the detail namespace.

Core/include/Acts/EventData/detail/CorrectedTransformationFreeToBound.hpp
XiaocongAi: I have removed the `FreeToBoundCorrection` out of the `detail`.

Examples/Algorithms/TrackFitting/include/ActsExamples/TrackFitting/TrackFittingAlgorithm.hpp
XiaocongAi: OK. Changed to use the constructor now.

Core/include/Acts/Propagator/AtlasStepper.hpp
```diff
@@ -580,13 +581,16 @@ class AtlasStepper {
...
+  /// @param [in] freeToBoundCorrection Correction for non-linearity effect during transform from free to bound
   ///
   /// @return A bound state:
   ///   - the parameters at the surface
   ///   - the stepwise jacobian towards it
   ///   - and the path length (from start - for ordering)
   Result<BoundState> boundState(State& state, const Surface& surface,
-                                bool transportCov = true) const {
+                                bool transportCov = true,
+                                const detail::FreeToBoundCorrection&
```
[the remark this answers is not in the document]
XiaocongAi: Fixed now.

Core/include/Acts/EventData/detail/CorrectedTransformationFreeToBound.hpp
```diff
@@ -0,0 +1,110 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+#include "Acts/Definitions/Common.hpp"
+#include "Acts/EventData/detail/TransformationFreeToBound.hpp"
+#include "Acts/Utilities/Logger.hpp"
+
+namespace Acts {
+namespace detail {
```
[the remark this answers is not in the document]
XiaocongAi: Done.

XiaocongAi:
Hi @asalzburger @paulgessinger , I have resolved your comments. Could you have another look?

asalzburger (approved):
Comments are addressed, there is still a failure in the CI to check.

asalzburger:
Hi @XiaocongAi  - there are build failures, due to a missing test update & a mismatch in the concept, please resolve these ones and I will reapprove.

paulgessinger:
Note there are also some CI issues that should be fixed with #1243.

Core/include/Acts/EventData/detail/CorrectedTransformationFreeToBound.hpp
```diff
@@ -0,0 +1,111 @@
...
+  ///
+  /// @param apply_ Wheter to apply correction
+  /// @param alpha_ The UKF tuning parameter alpha
+  /// @param beta_ The UKF tuning parameter beta
+  FreeToBoundCorrection(bool apply_, ActsScalar alpha_, ActsScalar beta_);
+
+  /// Construct from boolean only
+  ///
+  /// @param apply_ Wheter to apply correction
+  FreeToBoundCorrection(bool apply_);
```
paulgessinger, line 44: @asalzburger @XiaocongAi I'm sorry maybe our previous discussion wasn't entirely clear. This constructor is what I was unhappy with. Can we make this

```cpp
explicit FreeToBoundCorrection(bool apply_);
```

so the autoconversion from `bool` will be rejected by the compiler right away?

Sorry I didn't see earlier that you only changed the other location I commented on.

paulgessinger (changes requested):
I would strongly prefer we make the `FreeToBoundCorrection` constructor from `bool` explicit like 

```cpp
explicit FreeToBoundCorrection(bool apply_);
```

See comment in the changeset.

Core/include/Acts/EventData/detail/CorrectedTransformationFreeToBound.hpp
XiaocongAi: Thank you! Fixed now.

asalzburger:
From my side this is ok, let's see if the CI runs through and @paulgessinger is happy as well.

XiaocongAi:
Hi @paulgessinger , thank you for your further comments. I have made the change. Do you think you can approve and get it in?

paulgessinger (approved):
LGTM as far as I can tell at this point!

PR #1234, opened 2022-04-26
fix: Material composition script: return 0 on success

paulgessinger:
The material composition script now returns `0` on success and `1` on error.


AJPfleger:
Looks good to me.

AJPfleger (approved):
Looks good to me.

PR #1235, opened 2022-04-26
ci: Use semantic PR action instead of app

paulgessinger:
The app installation was needed to enable this on PRs from forks. With `pull_request_target` it should now be possible to run it in an action.

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,1 +1,0 @@
-titleOnly: true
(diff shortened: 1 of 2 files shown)


paulgessinger:
Ok, switched to `pull_request_target` so it should run on PRs from forks.

@robertlangenberg can you approve this?

robertlangenberg (approved):

PR #1243, opened 2022-05-03
ci: Bump docker image versions

paulgessinger:
Bumps to v21 of the docker images for the CI, this should fix the failures seen in the ExaTrkX job.

Also works around [this](https://github.blog/2022-04-12-git-security-vulnerability-announced/) security vulnerability in git, which now requires you to make folders as safe for the search for a top-level `.git` folder.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -18,7 +18,7 @@ env:
 jobs:
   build_debug:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


paulgessinger:
Ok this is finally passing. Could you take a look @stephenswat ?

stephenswat (approved):
Looks sensible! :slightly_smiling_face:

PR #1245, opened 2022-05-06
ci: Add ccache to CI builds

paulgessinger:
After @andiwand suggested it, I looked at this again, and it could potentially improve our build times significantly.

The caching here follows GitHubs caching implementation, where the cache is scoped to the branch in addition to what we configure. New PR branches (also from forks apparently) should be seeded from the main branch's cache, which is nice. 

This branch currently installs ccache on each job individually, but I have a PR running to add ccache directly to all of our docker images: https://github.com/acts-project/machines/pull/52.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -2,8 +2,8 @@ name: Analysis
 
 on:
   push:
-    branches:
-      - main
+    # branches:
+      # - main
   pull_request:
     branches:
       - main
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -19,12 +19,25 @@ jobs:
   build_debug:
     runs-on: ubuntu-latest
     container: ghcr.io/acts-project/ubuntu2004:v21
+    env:
+      CCACHE_DIR: ${{ github.workspace }}/ccache
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+
+      - name: Install ccache
+        run: apt-get install -y ccache
+
+      - name: Cache build
+        uses: actions/cache@v3
+        with:
+          path: ${{ github.workspace }}/ccache
+          key: ${{ runner.os  }}-ccache-linux_ubuntu_debug
+
       - name: Configure
         run: >
           cmake -B build -S .
           -GNinja
+          -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
           -DCMAKE_BUILD_TYPE=Debug
           -DCMAKE_CXX_FLAGS="-Werror -gz -g1"
           -DACTS_BUILD_UNITTESTS=on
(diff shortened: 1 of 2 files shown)


jmasik:
I've been using it with athena/cmake and even before with root from 90s - if the build system can keep a cache of relevant files it makes a huge saving.

paulgessinger:
Jup. I think I tested this at some point, but either GH actions cache wasn't ready yet or I couldn't get it to work. Not sure anymore.

paulgessinger:
Ok, the most recent CI run was pretty fast again. I guess we can merge this and see how it behaves in practice.

@andiwand can you review (and approve) this PR?

.github/workflows/analysis.yml
```diff
@@ -2,29 +2,39 @@ name: Analysis
 
 on:
   push:
-    branches:
-      - main
+    # branches:
+      # - main
```
andiwand: was that a temporary change?

.github/workflows/builds.yml
```diff
@@ -37,6 +45,7 @@ jobs:
           ${SETUP} &&
           cmake -B build -S .
           -GNinja
+          -DCMAKE_CXX_COMPILER_LAUNCHER=$(find / -type f -name "ccache")
```
andiwand, line 48: would `which ccache` do the job too?

.github/workflows/builds.yml
```diff
@@ -81,26 +90,33 @@ jobs:
...
       - uses: actions/checkout@v3
         with:
           submodules: true
           lfs: true
 
+      - name: Cache build
+        uses: actions/cache@v3
+        with:
+          path: ${{ github.workspace }}/ccache
+          key: ${{ runner.os  }}-ccache-linux_ubuntu_r1
```
andiwand, line 110: after seeing `r1` again - maybe it would be cool to have a global ENV variable that captures the ccache revision? since they will most likely roll the same time

.github/workflows/builds.yml
andiwand: or just `ccache` like below

.github/workflows/builds.yml
```diff
@@ -248,11 +264,19 @@ jobs:
...
+        uses: actions/cache@v3
+        with:
+          path: ${{ github.workspace }}/ccache
+          key: ${{ runner.os  }}-ccache-linux-nodeps_r1
+
       - name: Configure
         run: >
           cmake -B build -S .
           -GNinja
+          -DCMAKE_CXX_COMPILER_LAUNCHER=$(find / -type f -name "ccache")
```
andiwand, line 279: same as above

andiwand (approved):
looks good :+1: thanks for bringing this in!

.github/workflows/analysis.yml
paulgessinger: Yeah, this should be reverted now.

.github/workflows/builds.yml
paulgessinger: In the CentOS7/8 + LCG images, ccache is installed in a funky location somewhere. I realize now though, that after the LCG view setup script, it is actually in `$PATH`, so `ccache` should be enough. I'll change that.

.github/workflows/builds.yml
paulgessinger: Changed.

paulgessinger:
@andiwand once you're happy you'll have to hit approve one more time so the PR goes green.

PR #1247, opened 2022-05-09
ci: Add full chain test

andiwand:
execute `OpenDataDetector/ci/full_chain_odd.py` in CI and compare against reference `CI/physmon/reference/acts_analysis_residuals_and_pulls.root` with `histcmp`


paulgessinger:
Thanks!

Do you think it would make sense to integrate this step into `CI/physmon/phys_perf_mon.sh`?

.github/workflows/builds.yml
```diff
@@ -224,6 +244,28 @@ jobs:
...
+          && test -d /usr/local/share/Geant4-10.6.1/data || /usr/local/bin/download_geant4_data.sh
+          && source /usr/local/bin/thisroot.sh
+          && source /usr/local/bin/thisdd4hep_only.sh
+          && source /usr/local/bin/geant4.sh
+          && source build/python/setup.sh
+          && echo "::endgroup::"
+          && export PYTHONPATH="${PYTHONPATH}":"${GITHUB_WORKSPACE}/Examples/Scripts/Python"
+          && thirdparty/OpenDataDetector/ci/full_chain_odd.py
+          && build/bin/ActsAnalysisResidualsAndPulls -i trackstates_ckf.root -o acts_analysis_residuals_and_pulls.root --predicted --filtered --smoothed --silent || true
+          && histcmp acts_analysis_residuals_and_pulls.root CI/physmon/reference/acts_analysis_residuals_and_pulls.root
```
paulgessinger, line 261: `histcmp -o report.html` should produce the output HTML report, and `-p $folder` will dump all the comparison plots as PDFs in `$folder`. Both of these ideally would go into the folder that get's attached as an artifact. 

Can you check how large this gets? I think with the large number of plots that we get out of the root files, this might be a bit too much.

andiwand:
> Do you think it would make sense to integrate this step into `CI/physmon/phys_perf_mon.sh`?

good point! will do

.github/workflows/builds.yml
andiwand: looks like it is getting really big https://github.com/andiwand/acts/actions/runs/2293280651

are there any options to filter what we want to compare / plot?

andiwand:
I integrated into `CI/physmon/phys_perf_mon.sh` @paulgessinger if you want to take another look!

CI/physmon/phys_perf_mon.sh
```diff
@@ -66,5 +66,25 @@ run \
...
+echo "::group::full_chain_odd"
+
+thirdparty/OpenDataDetector/ci/full_chain_odd.py -o $outdir
+
+build/bin/ActsAnalysisResidualsAndPulls \
+    --predicted --filtered --smoothed --silent \
+    -i $outdir/trackstates_ckf.root \
+    -o $outdir/acts_analysis_residuals_and_pulls.root \
+
+histcmp \
```
paulgessinger, line 78: If you use the same `run` bash function that's used for the other comparisons, you won't have to duplicate the error code handling.

CI/physmon/phys_perf_mon.sh
```diff
@@ -66,5 +66,25 @@ run \
...
+    --title "full_chain_odd" \
+    --label-reference=$refcommit \
+    --label-monitored=$commit \
+    $outdir/acts_analysis_residuals_and_pulls.root \
+    CI/physmon/reference/acts_analysis_residuals_and_pulls.root \
```
paulgessinger, line 83: Is the HTML output too large?

andiwand: yeah it was like 200mb uncompressed and 100mb compressed

paulgessinger: Hmm. Maybe create it anyway and adjust the expiry time? Although it would be good to have this for a while in principle.

CI/physmon/phys_perf_mon.sh
andiwand: good point thanks!

CI/physmon/phys_perf_mon.sh
andiwand: I will leave it out for now. the root files for the plots are in the artifacts and could be plotted manually if necessary. but we can adopt that at any time if necessary

PR #1248, opened 2022-05-10
fix: fix scattering calculation in SeedFinderOrthogonal

LuisFelipeCoelho:
In the seedFinder we check the compatibility between the (r,z) slope of the two seed segments with a scattering term calculated assuming the minimum pT particle expected to be reconstructed and a second check using a scattering term scaled by the actual measured pT.

As discussed here in PR #1213, there was an error in these cuts. This PR fixes that same error in SeedFinderOrthogonal.


stephenswat (approved):
Looks good to me, thanks a lot for spotting this. :smile:

PR #1250, opened 2022-05-11
ci: Full chain ODD monitoring follow-up

paulgessinger:
Follow-up to #1247.

This moves the extra invocation of `thirdparty/OpenDataDetector/ci/full_chain_odd.py` into the main `CI/physmon/physmon.py` script where we basically run this chain anyway. This saves some time since we can generate the residual and pulls output from that invocation. I also change the `ActsAnalysisResidualsAndPulls` to be called directly from the python script, so we can reuse the temporary directory from there.

This also changes `ActsAnalysisResidualsAndPulls` to return UNIX-style error codes (0: ok, != 0, failure)


andiwand (commented):
cool :+1: 

I am just missing the setup and call of the sequencer https://gitlab.cern.ch/acts/OpenDataDetector/-/blob/main/ci/full_chain_odd.py#L69-107. is that not necessary anymore?

paulgessinger:
> cool 👍
> 
> I am just missing the setup and call of the sequencer https://gitlab.cern.ch/acts/OpenDataDetector/-/blob/main/ci/full_chain_odd.py#L69-107. is that not necessary anymore?

It should be using the one [here](https://github.com/acts-project/acts/pull/1250/files#diff-41761138103f525af36136ca099352e12affb064937ef430b4e7b80179245601R80-R82). Or do you mean something else?

paulgessinger:
Seems fine now, consider approving @andiwand.

andiwand (approved):

andiwand (approved):

PR #699, opened 2021-02-05
feat: fix adding obj options, adding possibility to write empty json map

asalzburger:
This PR fixes:
- missing `obj` options for `GeometryExample` leading to `boost::bad_any_cast`

It introduces:
- options for `json` geometry writing
- possibility to write an empty geometry map on demand to be filled in

The map would look like, e.g. for `--json-write-sf-layer` and the track ml detector:

```json
{
  "acts-geometry-hierarchy-map": {
    "format-version": 0,
    "value-identifier": "surface_types"
  },
  "entries": [
    {
      "volume": 5,
      "layer": 2,
      "value": "vol=5|lay=2"
    },
    {
      "volume": 7,
      "layer": 2,
      "value": "vol=7|lay=2"
    },
    {
      "volume": 7,
      "layer": 4,
      "value": "vol=7|lay=4"
    },
    {
      "volume": 7,
      "layer": 6,
      "value": "vol=7|lay=6"
    },
    {
      "volume": 7,
      "layer": 8,
      "value": "vol=7|lay=8"
    },
    {
      "volume": 7,
      "layer": 10,
      "value": "vol=7|lay=10"
    },
    {
      "volume": 7,
      "layer": 12,
      "value": "vol=7|lay=12"
    },
    {
      "volume": 7,
      "layer": 14,
      "value": "vol=7|lay=14"
    },
    {
      "volume": 8,
      "layer": 2,
      "value": "vol=8|lay=2"
    },
    {
      "volume": 8,
      "layer": 4,
      "value": "vol=8|lay=4"
    },
    {
      "volume": 8,
      "layer": 6,
      "value": "vol=8|lay=6"
    },
    {
      "volume": 8,
      "layer": 8,
      "value": "vol=8|lay=8"
    },
    {
      "volume": 9,
      "layer": 2,
      "value": "vol=9|lay=2"
    },
    {
      "volume": 9,
      "layer": 4,
      "value": "vol=9|lay=4"
    },
    {
      "volume": 9,
      "layer": 6,
      "value": "vol=9|lay=6"
    },
    {
      "volume": 9,
      "layer": 8,
      "value": "vol=9|lay=8"
    },
    {
      "volume": 9,
      "layer": 10,
      "value": "vol=9|lay=10"
    },
    {
      "volume": 9,
      "layer": 12,
      "value": "vol=9|lay=12"
    },
    {
      "volume": 9,
      "layer": 14,
      "value": "vol=9|lay=14"
    },
    {
      "volume": 10,
[...]

Issue #669, opened 2021-01-25
2D r-z Magnetic Field Map Improperly Read


osbornjd:
When sPHENIX runs Acts with a constant 1.4 T magnetic field, we get a reasonably good momentum resolution; when trying to run the same thing with our field map, the tracking resolution degrades significantly. We use essentially the same code as [GeometryExampleBase](https://github.com/acts-project/acts/blob/4bd3a88df4f0926593ab1401c0041bb6123bf045/Examples/Run/Common/src/GeometryExampleBase.cpp#L31) lines 31-56 to build our geometry, with the following options:

```
 -n1 -l0 --response-file sphenixResponse.response --mat-input-type file --mat-input-file materialFile --bf-map fieldMap.root --bf-name fieldmap --bf-lscalor 10 --bf-bscalor 9.3333e-05 --bf-rz true
```

The below plots show some tests of 20 pi+ 20 pi- events.

Running with a constant 1.4T field gives us the nominal momentum/phi/eta resolution seen here:
[constFieldEtaRes.pdf](https://github.com/acts-project/acts/files/5867707/constFieldEtaRes.pdf)
[constFieldPhiRes.pdf](https://github.com/acts-project/acts/files/5867708/constFieldPhiRes.pdf)
[constFieldpTRes.pdf](https://github.com/acts-project/acts/files/5867709/constFieldpTRes.pdf)

Note that the arguments to run this are as follows:
```
 -n1 -l0 --response-file sphenixResponse.response --mat-input-type file --mat-input-file materialFile -bf-values 0 0 1.5 --bf-bscalor 0.9333333
```

Running with the same code but with the field map gives the degraded resolution seen here, which seems to be correlated with a phi offset that is charge dependent:
[fieldMapEtaRes.pdf](https://github.com/acts-project/acts/files/5867711/fieldMapEtaRes.pdf)
[fieldMapPhiRes.pdf](https://github.com/acts-project/acts/files/5867712/fieldMapPhiRes.pdf)
[fieldMappTRes.pdf](https://github.com/acts-project/acts/files/5867713/fieldMappTRes.pdf)

The phi separation indicates that there must be some charge dependent way the field is being interpreted incorrectly, I suspect. Note that I tried flipping the field sign for Acts (this leads to many failed track fits) and flipping the charge of the tracks going into the fitter (Acts just sees the tracks as the opposite sign but is still able to fit them) so it seems like Acts is interpreting the direction of the field correctly.

[...]

PR #1251, opened 2022-05-11
feat: add Python bindings for MagneticFieldMap

timadye:
* Adds missing Python bindings for `MagneticFieldMapXyz` and `MagneticFieldMapRz` to allow the field map to be read from a file.
* Add example use in `full_chain_itk.py`. Previously used a constant B-field.
* Also show different seeding options, and debug setting, commented out in `full_chain_itk.py`.


Examples/Python/src/MagneticField.cpp
```diff
@@ -14,16 +14,84 @@
...
 #include <pybind11/pybind11.h>
 
 namespace py = pybind11;
 using namespace pybind11::literals;
 
-namespace Acts::Python {
+namespace ActsExamples {
+// MagneticFieldMapRz/Xyz are subclasses of InterpolatedMagneticField2/3 used
+// here simply to keep them separate for the Python bindings
+class MagneticFieldMapRz final
```
paulgessinger, line 31: What's the advantage of this class + factory and constructor over a free function that you expose to python in isolation?

timadye: I tried the free function, but couldn't get it to work at the time because the returned `InterpolatedMagneticField3` was not recognised as a `MagneticFieldProvider`.
I have since discovered that you can specify different base classes in `py::class_<D,B1,B2>` (the `pybind11` docs I saw only mentioned one base class). I subsequently used that for my `MagneticFieldMapXyz` so it could also be used by `RootBFieldWriter` (requires an `InterpolatedMagneticField`), so maybe the same trick would work for `InterpolatedMagneticField3`.
Do you think that would be better?

paulgessinger:
I think if we can get around having to add additional classes, that would be better.
I think it should be possible to let pybind know that the interpolated field classes inherit from MagneticFieldProvider, and then I believe python/pybind should be able to do the conversions automatically.

Examples/Python/src/MagneticField.cpp
timadye: It seems to work everywhere, so I committed this change.
The PR is still WiP because I get very bad physics performance results, which I'm investigating.

timadye:
Hi @paulgessinger,

I've fixed a bug where there field was read with in the wrong units (`BFieldUnit=Acts::UnitConstants::T`) and now the physics performance looks good. I'm making final checks now, so hopefully I can remove the WiP.

While investigating, I tried writing out the field map file with `RootBFieldWriter` and comparing with the original. This now looks similar, except for a small change in the binning. I don't think this is an issue for the Python bindings, because I see the same effect with `ActsExampleMagneticField`. I think I also saw this issue with the Rz map from `solenoidFieldMap`.

I guess there is a problem with how `Acts::fieldMapXYZ` calculates the binning from the field points. I don't think it's a problem with the `RootBFieldWriter`, because the first-written solenoid map looks good.

Is this a bug, not serious, or expected? If a bug, I could submit an Issue.

Thanks,
Tim.

timadye:
I'll just record here the change in binning I mentioned above.

* `acts-itk/bfield/ATLAS-BField-xyz.root` has `(x,y,z)` bins `(200, -10000, 10000, 200, -10000, 10000, 300, -15000, 15000)`, so the bin width is `(100,100,100)`, and the first and last field points are at `(-10000,-10000,-15000)` and `(10000,10000,15000)`.
* read in and written out it has `(x,y,z)` bins `(200, -10000, 10099.502, 200, -10000, 10099.502, 302, -15000, 15099.667)`, so the bin width is `(99.502,99.502,99.667)`, and the first and **penultimate** field points are at `(-10000,-10000,-15000)` and `(10000,10000,15000)`. There is one plane of field points in each dimension outside the original positive range.

paulgessinger:
@timadye could it be the same issue as this one: https://github.com/acts-project/acts/issues/669 ?

timadye:
> @timadye could it be the same issue as this one: #669 ?

Yes, it looks like the same issue. We don't see such a bad effect for the ATLAS B-field as sPHENIX do, but they are both binning issues in `fieldMapXYZ`.

paulgessinger:
This looks to me like an off-by-one error due to the fact that we're using the lower left corner of the grid cells as the interpolation points.

timadye:
It's probably better if we discuss this in the Issue #699, rather than this PR. Since it isn't directly connected to adding the Python binding, and might simplify testing the problem, can we merge this PR - if there aren't any other problems? We can add a test once the issue is resolved.

paulgessinger (approved):
Let's do it!

PR #1254, opened 2022-05-13
fix: prevent debug printout crashing Python TGeoDetector.create()

timadye:
* Prevent `dump_args_calls` from instrumenting `Interval` and screwing up `TGeoDetector.create()`
* Add a couple of commented-out options to `full_chain_itk.py`:
  * show different seeding options
  * add `dump_args_calls` to debug Python bindings


paulgessinger (approved):
LGTM

PR #1258, opened 2022-05-17
fix: ODD with macOS

Corentin-Allaire:
This fix the common.py to be able to use the ODD with macOs. It also make the error message a bit more explicite if the path to the ODD factory was not set.

### DIFF
--- a/Examples/Scripts/Python/common.py
+++ b/Examples/Scripts/Python/common.py
@@ -30,7 +30,6 @@ def getOpenDataDetector(mdecorator=None):
         env_vars = ["DYLD_LIBRARY_PATH", "DD4HEP_LIBRARY_PATH"]
         lib_name = "libOpenDataDetector.dylib"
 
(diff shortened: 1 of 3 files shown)


Examples/Scripts/Python/common.py
```diff
@@ -41,7 +42,7 @@ def getOpenDataDetector(mdecorator=None):
         if not found:
             msg = (
                 "Unable to find OpenDataDetector factory library. "
-                f"You might need to point {'/'.join(env_vars)} at it"
+                f"You might need to point {'/'.join(env_vars)} to build/thirdparty/OpenDataDetector/factory"
```
paulgessinger, line 45: If you're already changing this, can you add `or other ODD install location`?

Corentin-Allaire: Sure ! And I will also add it to the doc

PR #1260, opened 2022-05-18
feat: EDM4hep reader

andiwand:
**WIP**

- add EDM4hep reader


Examples/Python/tests/test_reader.py
```diff
@@ -303,3 +307,29 @@ def test_csv_clusters_reader(tmp_path, fatras, conf_const, trk_geo, rng):
...
+
+    detector, _, _ = getOpenDataDetector()
+
+    s = Sequencer(numThreads=1)
+
+    s.addReader(
+        conf_const(
+            EDM4hepSimHitReader,
+            level=acts.logging.INFO,
+            inputPath=str("/home/andreas/cern/output_edm4hep.root"),  # TODO
```
andiwand, line 324: add file to repo

Examples/Io/EDM4hep/src/EDM4hepSimHitReader.cpp
```diff
@@ -0,0 +1,116 @@
...
+              },
+              [&](std::uint64_t cellId) {
+                auto detElement = m_cfg.dd4hepGeometryService->lcdd()
+                                      ->volumeManager()
+                                      .lookupDetElement(cellId);
+                Acts::GeometryIdentifier result = detElement.volumeID();
+                return result;
+              });
+          unordered.push_back(std::move(hit));
+        } catch (...) {
```
paulgessinger, line 98: I didn't know this was a thing... cool.

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,305 @@
...
+  ActsFatras::Barcode particleId = particleMapper(from);
+
+  ActsFatras::Particle to(particleId, (Acts::PdgParticle)from.getPDG(),
+                          from.getCharge() * Acts::UnitConstants::e,
+                          from.getMass() * Acts::UnitConstants::GeV);
+
+  // TODO do we have that in EDM4hep?
+  // particle.setProcess(static_cast<ActsFatras::ProcessType>(data.process));
+
+  to.setPosition4(from.getVertex()[0] * Acts::UnitConstants::mm,
```
paulgessinger, line 36: Are the input units fixed from edm4hep? Otherwise we'd have to divide by their unit constants first.

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,305 @@
...
+                                    from.getMomentum()[1],
+                                    from.getMomentum()[2]) *
+                         Acts::UnitConstants::GeV);
+
+  return to;
+}
+
+void EDM4hepUtil::writeParticle(const ActsFatras::Particle& from,
+                                edm4hep::MutableMCParticle to) {
+  // TODO what about particleId?
```
paulgessinger, line 55: So we can't writer our barcode in there? Maybe an input to edm4hep as well: add a user defined particle identifier?

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+  to.setPDG(from.pdg());
+  to.setCharge(from.charge() / Acts::UnitConstants::e);
+  to.setMass(from.mass() / Acts::UnitConstants::GeV);
+  to.setVertex({from.position().x(), from.position().y(), from.position().z()});
+  to.setMomentum({(float)from.fourMomentum().x(),
```
paulgessinger, line 61: Would prefer `static_cast` here.

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,305 @@
...
+                  (float)from.fourMomentum().y(),
+                  (float)from.fourMomentum().z()});
+}
+
+ActsFatras::Hit EDM4hepUtil::readSimHit(const edm4hep::SimTrackerHit& from,
+                                        MapParticleIdFrom particleMapper,
+                                        MapGeometryIdFrom geometryMapper) {
+  ActsFatras::Barcode particleId = particleMapper(from.getMCParticle());
+
+  // TODO what about the digitization?
```
paulgessinger, line 71: Can you elaborate?

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+void EDM4hepUtil::writeSimHit(const ActsFatras::Hit& from,
+                              edm4hep::MutableSimTrackerHit to,
+                              MapParticleIdTo particleMapper,
+                              MapGeometryIdTo geometryMapper) {
+  const Acts::Vector4& globalPos4 = from.fourPosition();
+  const Acts::Vector4& momentum4Before = from.momentum4Before();
+  const auto delta4 = from.momentum4After() - momentum4Before;
+
+  if (particleMapper != nullptr) {
```
paulgessinger, line 118: I **think** `operator bool()` is well defined for `std::function`. So maybe

```cpp
if (particleMapper) {}
```

is enough?

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,305 @@
...
+  auto pos = from.getPosition();
+  auto cov = from.getCovMatrix();
+
+  DigitizedParameters dParameters;
+
+  dParameters.indices.push_back(Acts::eBoundLoc0);
+  dParameters.values.push_back(pos.x);
+  dParameters.variances.push_back(cov[0]);
+
+  // TODO cut this out for 1D
```
paulgessinger, line 163: Can we identify the dimensionality without checking/enumerating the edm4hep hit types?

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,305 @@
...
+  auto to = createMeasurement(dParameters, sourceLink);
+
+  if (fromClusters != nullptr) {
+    for (auto objectId : from.getRawHits()) {
+      const auto& c = fromClusters->at(objectId.index);
+
+      // TODO get EDM4hep fixed
+      // misusing some fields to store ACTS specific information
+      // don't ask ...
+      ActsFatras::Channelizer::Bin2D bin{(unsigned int)c.getType(),
```
paulgessinger, line 181: Again, prefer `static_cast`.

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,305 @@
...
+        to.setType(EDM4hepUtil::EDM4HEP_ACTS_POSITION_TYPE);
+        // TODO set uv (which are in global spherical coordinates with r=1)
+        to.setPosition({parameters[Acts::eBoundLoc0],
+                        parameters[Acts::eBoundLoc1],
+                        parameters[Acts::eBoundTime]});
+
+        auto covariance =
+            (m.expander() * m.covariance() * m.expander().transpose()).eval();
+        to.setCovMatrix({
+            (float)covariance(Acts::eBoundLoc0, Acts::eBoundLoc0),
```
paulgessinger, line 221: Would prefer `static_cast`.

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+        auto parameters = (m.expander() * m.parameters()).eval();
+
+        to.setTime(parameters[Acts::eBoundTime] / Acts::UnitConstants::ns);
+
+        to.setType(EDM4hepUtil::EDM4HEP_ACTS_POSITION_TYPE);
+        // TODO set uv (which are in global spherical coordinates with r=1)
+        to.setPosition({parameters[Acts::eBoundLoc0],
+                        parameters[Acts::eBoundLoc1],
```
paulgessinger, line 215: This will just be zero in case of lower dimension (i.e. 1D) measurements, right?

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,305 @@
...
+        state.projector() * state.covariance() * state.projector().transpose();
+
+    edm4hep::TrackState trackState;
+
+    trackState.D0 = meas[Acts::eBoundLoc0];
+    trackState.Z0 = meas[Acts::eBoundLoc1];
+    trackState.phi = meas[Acts::eBoundPhi];
+    trackState.tanLambda = std::tan(M_PI_2 - meas[Acts::eBoundTheta]);
+    trackState.omega = meas[Acts::eBoundQOverP];  // TODO convert
+    trackState.referencePoint = {0, 0, 0};
```
paulgessinger, line 284: Oh I see this is probably what d0/z0 is relative too. Maybe we can convert to perigee at 0, 0, 0?

Examples/Io/Root/src/RootMaterialTrackReader.cpp
```diff
@@ -76,9 +76,14 @@ ActsExamples::RootMaterialTrackReader::RootMaterialTrackReader(
 }
 
 ActsExamples::RootMaterialTrackReader::~RootMaterialTrackReader() {
+  delete m_inputChain;
```
paulgessinger, line 79: Are these `delete`s not needed?

Examples/Python/src/DD4hepComponent.cpp
```diff
@@ -46,13 +46,17 @@ PYBIND11_MODULE(ActsPythonBindingsDD4hep, m) {
...
+        py::class_<DD4hep::DD4hepDetector,
+                   std::shared_ptr<DD4hep::DD4hepDetector>>(m, "DD4hepDetector")
+            .def(py::init<>())
+            .def("finalize",
+                 py::overload_cast<
+                     DD4hep::DD4hepGeometryService::Config,
+                     std::shared_ptr<const Acts::IMaterialDecorator>>(
+                     &DD4hep::DD4hepDetector::finalize));
+    ACTS_PYTHON_STRUCT_BEGIN(gd, DD4hep::DD4hepDetector);
+    ACTS_PYTHON_MEMBER(geometryService);
```
paulgessinger, line 59: That's why you did this, I see.
Your call whether to switch to a private member + getter.

Examples/Python/tests/helpers/__init__.py
```diff
@@ -37,9 +37,18 @@
     import acts.examples.hepmc3
 
     hepmc3Enabled = True
-except ImportError:
+except Exception as e:
+    print(e)
```
paulgessinger, line 41: Needed?

Examples/Python/tests/helpers/__init__.py
```diff
@@ -37,9 +37,18 @@
...
     hepmc3Enabled = False
 
+try:
+    import acts.examples.edm4hep
+
+    edm4hepEnabled = True
+except Exception as e:
+    print(e)
```
paulgessinger, line 49: Needed?

Examples/Python/tests/test_detectors.py
```diff
@@ -32,15 +36,16 @@ def test_generic_geometry():
...
         rConfig=config,
-        jFileName="thirdparty/OpenDataDetector/config/odd-material-mapping-config.json",
+        jFileName=str(
+            getOpenDataDetectorDirectory() / "config/odd-material-mapping-config.json"
+        ),
         level=acts.logging.WARNING,
     )
 
-    detector, geo, _ = getOpenDataDetector(matDeco)
+    _, geo, _ = getOpenDataDetector(matDeco)
```
paulgessinger, line 48: Same issue as before.

Examples/Io/Root/src/RootMaterialTrackReader.cpp
paulgessinger: Haha nevermind, I read the diff the wrong way around.

Examples/Io/EDM4hep/include/ActsExamples/Io/EDM4hep/EDM4hepMeasurementWriter.hpp
```diff
@@ -0,0 +1,77 @@
...
+                     const MeasurementContainer& measurements) final;
+
+ private:
+  Config m_cfg;
+
+  podio::ROOTWriter m_writer;
+  podio::EventStore m_store;
+
+  edm4hep::TrackerHitPlaneCollection* m_trackerHitPlaneCollection;
+  edm4hep::TrackerHitCollection* m_trackerHitRawCollection;
```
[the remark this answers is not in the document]
andiwand: yeah exactly. it was just nicer than using references

.github/workflows/builds.yml
```diff
@@ -130,6 +128,7 @@ jobs:
           -DACTS_BUILD_EVERYTHING=ON
           -DACTS_BUILD_ODD=ON
           -DACTS_BUILD_EXAMPLES_PYTHON_BINDINGS=ON
+          -DACTS_BUILD_EXAMPLES_EDM4HEP=ON
```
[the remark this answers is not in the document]
paulgessinger: I see...

Examples/Io/EDM4hep/include/ActsExamples/Io/EDM4hep/EDM4hepSimHitReader.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+ private:
+  Config m_cfg;
+  std::pair<size_t, size_t> m_eventsRange;
+  std::unique_ptr<const Acts::Logger> m_logger;
+
+  podio::ROOTReader m_reader;
+  podio::EventStore m_store;
+
+  std::vector<std::string> m_collections;
```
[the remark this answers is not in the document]
andiwand: actually this is what is currently implemented. I just cache the names of the collections so I don't have to query EDM4hep at each event

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
andiwand: according to their doc the vertex should be in `mm` https://github.com/key4hep/EDM4hep/blob/master/edm4hep.yaml#L133

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
andiwand: yeah exactly they don't offer a field for particle ID. it is only identified by an index in their collection

Examples/Io/EDM4hep/include/ActsExamples/Io/EDM4hep/EDM4hepMeasurementWriter.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include "edm4hep/TrackerHitCollection.h"
+#include "edm4hep/TrackerHitPlaneCollection.h"
+#include "podio/EventStore.h"
+#include "podio/ROOTWriter.h"
+
+namespace ActsExamples {
+
+/// Write out a measurement cluster collection to EDM4hep.
+///
+/// Inpersistent information:
```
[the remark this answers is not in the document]
andiwand: yes currently I could not preserve the listed information in the EDM4hep so if you read in again this will be lost

Core/include/Acts/EventData/MultiTrajectory.ipp
```diff
@@ -60,7 +60,7 @@ inline auto TrackStateProxy<M, ReadOnly>::parameters() const -> Parameters {
     idx = data().ipredicted;
   }
 
-  return Parameters(m_traj->m_params.data.col(idx).data());
+  return Parameters(m_traj->m_params.col(idx).data());
```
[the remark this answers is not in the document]
andiwand: this is where I use it https://github.com/acts-project/acts/blob/f22cef73c4485044be37922b64237d8dca08da17/Examples/Io/EDM4hep/src/EDM4hepUtil.cpp#L274

looked like to me that `data` is not a public member and the other functions also don't rely on it so I thought this is most likely a bug?

Examples/Io/EDM4hep/src/EDM4hepMeasurementReader.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+std::pair<size_t, size_t> EDM4hepMeasurementReader::availableEvents() const {
+  return m_eventsRange;
+}
+
+ProcessCode EDM4hepMeasurementReader::read(const AlgorithmContext& ctx) {
+  MeasurementContainer measurements;
+  ClusterContainer clusters;
+  // TODO what about those?
```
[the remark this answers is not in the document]
andiwand: currently I leave these collections empty but I guess they should be filled up with something. at least CSV/Root IO does so

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
andiwand: I mixed things up here. simhits are pre-digitization right?

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+Measurement EDM4hepUtil::readMeasurement(
+    edm4hep::TrackerHitPlane from,
+    const edm4hep::TrackerHitCollection* fromClusters, Cluster* toCluster,
+    MapGeometryIdFrom geometryMapper) {
+  // no need for digitization as we only want to identify the sensor
+  Acts::GeometryIdentifier geometryId = geometryMapper(from.getCellID());
+
+  // TODO what about the hit index?
+  IndexSourceLink sourceLink{geometryId, 0};
```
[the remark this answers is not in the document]
andiwand: okay I will use the `id()` for now. I was not sure if the index has to be meaningful in terms of being part of a trajectory

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
andiwand: I fear we can only go through the type or do something like setting U/V to zero

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,310 @@
...
+  dParameters.variances.push_back(cov[2]);
+
+  dParameters.indices.push_back(Acts::eBoundTime);
+  dParameters.values.push_back(pos.z);
+  dParameters.variances.push_back(cov[5]);
+
+  auto to = createMeasurement(dParameters, sourceLink);
+
+  if (fromClusters) {
+    for (auto objectId : from.getRawHits()) {
```
niermann999, line 175: Is ```const``` possible?

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,310 @@
...
+            static_cast<float>(covariance(Acts::eBoundLoc0, Acts::eBoundLoc0)),
+            static_cast<float>(covariance(Acts::eBoundLoc1, Acts::eBoundLoc0)),
+            static_cast<float>(covariance(Acts::eBoundLoc1, Acts::eBoundLoc1)),
+            0,
+            0,
+            0,
+        });
+
+        if (fromCluster) {
+          for (auto& c : fromCluster->channels) {
```
niermann999, line 231: Could we maybe make ```const``` explicit?

Examples/Io/EDM4hep/include/ActsExamples/Io/EDM4hep/EDM4hepUtil.hpp
```diff
@@ -0,0 +1,126 @@
...
+    std::function<Acts::GeometryIdentifier(std::uint64_t cellId)>;
+using MapGeometryIdTo =
+    std::function<std::uint64_t(Acts::GeometryIdentifier geometryId)>;
+
+/// Reads a Fatras particle from EDM4hep.
+///
+/// Inpersistent information:
+/// - particle ID
+/// - process
+ActsFatras::Particle readParticle(edm4hep::MCParticle from,
```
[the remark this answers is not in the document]
andiwand: podio seems to use pointer-like wrapper for each of their objects and collections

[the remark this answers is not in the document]
andiwand: *pointer-like wrapper with value semantics :smile:

Core/include/Acts/EventData/MultiTrajectory.ipp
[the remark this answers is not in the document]
paulgessinger: This was changed in #1289 I believe. Can you check when you rebase / merge?

Examples/Io/EDM4hep/include/ActsExamples/Io/EDM4hep/EDM4hepSimHitReader.hpp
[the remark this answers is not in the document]
paulgessinger: Oh I see, I thought this was an input parameter. Nevermind then.

Examples/Io/EDM4hep/src/EDM4hepMeasurementReader.cpp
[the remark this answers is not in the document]
paulgessinger: Ok I see. I think actualy downstream jobs will struggle / fail without these connections being set. For example I don't think the fitting will work if the source links are not populated.

Examples/Io/EDM4hep/src/EDM4hepParticleReader.cpp
```diff
@@ -0,0 +1,71 @@
...
+  m_store.clear();
+  m_reader.goToEvent(ctx.eventNumber);
+
+  SimParticleContainer::sequence_type unordered;
+
+  for (const auto& mcParticle : *m_mcParticleCollection) {
+    auto particle =
+        EDM4hepUtil::readParticle(mcParticle, [](edm4hep::MCParticle p) {
+          ActsFatras::Barcode result;
+          // TODO dont use podio internal id
```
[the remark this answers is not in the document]
paulgessinger: I don't know for sure if we don't expect the contents of barcode anywhere. I believe it encodes vertex id and so on, so can be used to determine if two particles came from the same vertex.

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
paulgessinger: Ok this is something we should clarify with the edm4hep people. They store the vertex location, so during reading we might have to reconstruct this.

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
paulgessinger: They are

Examples/Io/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+#include "edm4hep/TrackState.h"
+
+namespace ActsExamples {
+
+ActsFatras::Particle EDM4hepUtil::readParticle(
+    edm4hep::MCParticle from, MapParticleIdFrom particleMapper) {
+  ActsFatras::Barcode particleId = particleMapper(from);
+
+  ActsFatras::Particle to(particleId, (Acts::PdgParticle)from.getPDG(),
```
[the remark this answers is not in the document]
paulgessinger: Oh didn't see this. Can you switch this to `static_cast` as well?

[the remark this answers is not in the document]
andiwand: yeah nice catch @niermann999 - I changed it right away. for numerics I usually use c style casts because I think it is more readable. but otherwise e.g. enums I prefer `static_cast` too.

andiwand:
maybe we can store the barcode in `MCParticle::*Status` and make it persistent

.github/workflows/builds.yml
```diff
@@ -285,11 +283,9 @@ jobs:
           restore-keys: |
             ${{ runner.os  }}-ccache-linux-nodeps_${{ env.CCACHE_KEY_SUFFIX }}_
 
-
       - name: Configure
-        run: >
+        run: > # disabling ninja because of race condition
```
paulgessinger, line 287: Why is this needed?

.github/workflows/builds.yml
```diff
@@ -206,11 +204,11 @@ jobs:
           && source build/python/setup.sh
           && export LD_LIBRARY_PATH=$PWD/build/thirdparty/OpenDataDetector/factory:$LD_LIBRARY_PATH
           && pip3 install -r Examples/Python/tests/requirements.txt
-          && pytest -rFs
+          && pytest -v
```
paulgessinger, line 207: I think `-rFs` is a little cleaner because it doesn't report everything. Can we do both at least?

.github/workflows/builds.yml
andiwand: I don't remember exactly but I think I was facing flapping here because some downloads were too slow/fast

.github/workflows/builds.yml
andiwand: sure - I think with `-rFs` I was not able to see which unit test failed only which file. that made it hard for me to cross check with my local testing

.github/workflows/builds.yml
andiwand: let me try to revert the CI changes that are not necessary since master seems to be stable

paulgessinger (approved):
LGTM!

PR #1263, opened 2022-05-21
ci: Use restore-key to continually update ccache storage

paulgessinger:
We observed a brief period of speedup in  the CI after introducing ccache, which degraded again after a few weeks:

citime_Analysis_0
citime_Builds_0
citime_Builds_1
citime_Builds_2
citime_Builds_3

This is due to the fact that GitHub Actions caches do not get updated at the end of their run. This PR introduces a workaround. The CI will

1. Look for a key like `cache_key_$OS_$JOB_$COMMIT_SHA` and use that if found. If not it will
2. Look for the newest key starting with `cache_key_$OS_$JOB_`, and use that if found. Otherwise it will perform a build from scratch
3. At the end of the job, if a cache using the restore key (i.e. 2. above) was used, it will create a new cache with the key from 1 (`cache_key_$OS_$JOB_$COMMIT_SHA`). The next CI run will use that cache as a starting point.

Clearly, this means that the cache size will continually increase. However, ccache has a maximum size setting, which I set to 2GB in this PR. We'll have to see if that results in a reasonably sized, reasonably up-to-date cache and corresponding fast build times.


andiwand (approved):
great :+1:

PR #1266, opened 2022-05-24
feat: Allow Separate Magnetic Field

LuisFelipeCoelho:
This PR does the same as #1227 but for the magnetic field. 
In Athena the field is slightly different (due to computational reasons) in the grid creation and in the finder.
This PR allows ACTS to reproduce the same behaviour if `allowSeparateBField = true` otherwise it will throw `invalid_argument` if the values are different


LuisFelipeCoelho:
This is actually not needed to be implemented in ACTS

PR #1269, opened 2022-05-24
fix: Add another compat fix for clang+libstdc++

paulgessinger:
This fixes an issue in compiling using clang + libstdc++ after #1232 was merged.


niermann999 (approved):
Interesting. Looks good to me

PR #1270, opened 2022-05-24
ci: Reduce ccache limit to 150MB

paulgessinger:
The hope is that this reduces individual cache sizes and allows older caches (i.e. main) to be retained for longer.


paulgessinger:
@andiwand what do you think?

andiwand:
lets give it a try @paulgessinger !

paulgessinger:
Sooo, can you approve it @andiwand? 😄

andiwand (approved):

Issue #1273, opened 2022-05-26
Add CPU Profiling Tool To ACTS


Scott-James-Hurley:
Adding a tool to profile CPU usage for the ACTS project will allow developers to identify parts of the code that are not as performant as expected, making it easier to find and resolve performance bottlenecks. 

After an investigation [gperftools](https://github.com/gperftools/gperftools) seems to be the most promising solution. It offers a [sampling-based CPU profiler](https://en.wikipedia.org/wiki/Profiling_(computer_programming)#Statistical_profilers) which means less effort on the developers' part to start measuring CPU performance. 

As a bonus it also includes a memory profiler and heap checker for measuring memory allocations and testing for memory leaks respectively.

This issue would involve:

- adding build options for enabling profiling
- adding documentation on how to install any dependencies, details on basic usage and resources for further reading on relevant topics
- anything else to make it easier for developers to profile the code.

PR #1274, opened 2022-05-27
feat: Add Profiling Support with gperftools

Scott-James-Hurley:
Closes: #1273

This PR adds two new build options: `ACTS_ENABLE_CPU_PROFILING` and `ACTS_ENABLE_MEMORY_PROFILING` that will link libraries and add compile options to enable the use of gperftools' CPU profiler and memory profiler/heap checker respectively. The libraries are linked in the `Core` package as the code here is always built and is the main point of interest for performance profiling.

This PR also adds documentation to `docs/howto` for installing, enabling and using gperftools with ACTS, including some information about basic commands that I found useful when profiling ACTS as well as relevant links.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,9 @@ option(ACTS_USE_SYSTEM_BOOST "Use a system-provided boost" ON)
 option(ACTS_SETUP_EIGEN3 "Explicitly set up Eigen3 for the project" ON)
 option(ACTS_USE_SYSTEM_EIGEN3 "Use a system-provided eigen3" ON)
 option(ACTS_BUILD_ODD "Build the OpenDataDetector" OFF)
+# profiling related optios
+option(ACTS_ENABLE_CPU_PROFILING "Enable CPU profiling using gperftools" OFF)
+option(ACTS_ENABLE_MEMORY_PROFILING "Enable memory profiling using gperftools" OFF)
 
 # handle option inter-dependencies and the everything flag
 # NOTE: ordering is important here. dependencies must come before dependees
@@ -337,6 +340,12 @@ add_subdirectory_if(Examples ACTS_BUILD_EXAMPLES)
 if(ACTS_BUILD_BENCHMARKS OR ACTS_BUILD_INTEGRATIONTESTS OR ACTS_BUILD_UNITTESTS)
   enable_testing() # must be set in the main CMakeLists.txt
   add_subdirectory(Tests)
+endif()
+
+#generate debug information
+if(ACTS_ENABLE_CPU_PROFILING)
+  message(STATUS "Added -g compile flag")
+  add_compile_options(-g)
 endif()
 
 # documentation
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -342,12 +342,6 @@ if(ACTS_BUILD_BENCHMARKS OR ACTS_BUILD_INTEGRATIONTESTS OR ACTS_BUILD_UNITTESTS)
   add_subdirectory(Tests)
 endif()
 
-#generate debug information
-if(ACTS_ENABLE_CPU_PROFILING)
-  message(STATUS "Added -g compile flag")
-  add_compile_options(-g)
-endif()
-
 # documentation
 add_subdirectory_if(docs ACTS_BUILD_DOCS)
 
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -57,6 +57,8 @@ For convenience, it is possible to build the required boost and eigen3 dependenc
 Other options are also
 available and are discussed in the [Building Acts](#building-acts) section.
 
+The prerequisites for profiling the ACTS project with gperftools are detailed in docs/howto/profiling.md
+
 ## Building Acts
 
 Acts uses [CMake](https://cmake.org) to configure, build, and install the
@@ -243,7 +245,7 @@ components.
 | ACTS_BUILD_EVERYTHING               | Build with most options enabled (except HepMC3 and documentation)                                     |
 | ACTS_BUILD_PLUGIN_CUDA              | Build CUDA plugin                                                                                     |
 | ACTS_BUILD_PLUGIN_DD4HEP            | Build DD4hep geometry plugin                                                                          |
-| ACTS_BUILD_PLUGIN_EXATRKX           | Build Exa.TrkX plugin                                                                             |
+| ACTS_BUILD_PLUGIN_EXATRKX           | Build Exa.TrkX plugin                                                                                 |
 | ACTS_BUILD_PLUGIN_IDENTIFICATION    | Build Identification plugin                                                                           |
 | ACTS_BUILD_PLUGIN_JSON              | Build Json plugin                                                                                     |
 | ACTS_BUILD_PLUGIN_LEGACY            | Build legacy plugin                                                                                   |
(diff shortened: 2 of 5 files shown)


docs/howto/profiling.md
```diff
@@ -0,0 +1,153 @@
...
+
+Alternatively, you can use the following commands to install it:
+
+```
+git clone https://github.com/gperftools/gperftools
+cd gperftools
+./autogen.sh
+./configure
+make
+make install
```
timadye, line 26: Will this install in a location that ACTS CMake can find?
I guess that is for systems with root access. Is there a way to set it up in another directory?

docs/getting_started.md
```diff
@@ -57,6 +57,8 @@ For convenience, it is possible to build the required boost and eigen3 dependenc
 Other options are also
 available and are discussed in the [Building Acts](#building-acts) section.
 
+The prerequisites for profiling the ACTS project with gperftools are detailed in docs/howto/profiling.md
```
timadye, line 60:
```suggestion
The prerequisites for profiling the ACTS project with gperftools are detailed [here](howto/profiling.md).
```
but I don't know if that works when converting to readthedocs.

timadye: I hope not: `getting_started.md` is Markdown after all. But I don't know what defines if something gets added to readthedocs. Maybe @robertlangenberg can advise.

docs/howto/profiling.md
Scott-James-Hurley: Added a build option for specifying the install directory of gperftools and instructions on how to specify a directory when installing gperftools.

docs/howto/profiling.md
```diff
@@ -0,0 +1,156 @@
...
+
+```
+git clone https://github.com/gperftools/gperftools
+cd gperftools
+git tag -l # checkout the latest release version
+git checkout <gperftools-X.x>
+./autogen.sh
+./configure
+make prefix=/usr libdir=<your/desired/install/dir>
+make prefix=/usr libdir=<your/desired/install/dir> install
```
timadye, line 28: doesn't `prefix=/usr` mean that files apart from library ones are written to `/usr`? That's not writable without root.

Would it work to specify just `prefix=<your/desired/install/dir>`?

docs/howto/profiling.md
```diff
@@ -0,0 +1,156 @@
...
+-DACTS_ENABLE_CPU_PROFILING=ON
+```
+
+Similarly, to enable the memory profiler the following build option should be used:
+
+```
+-DACTS_ENABLE_MEMORY_PROFILING=ON
+```
+
+Alternatively, you can avoid rebuiding the project by pointing the LD_PRELOAD environment variable to the profiler library for CPU profiling:
```
timadye, line 64: Since this is not the recommended method (but still good to know), can you put this in a sub-sub-section, eg.

### Alternatives to recompiling

so people who read too quickly know they don't have to do both.

docs/howto/profiling.md
```diff
@@ -0,0 +1,156 @@
...
+LD_PRELOAD="<path/to/libprofiler.so>" <other_options> <path/to/binary> <binary_flags>
+```
+
+You can do the same thing with the tcmalloc library for memory profiling:
+
+```
+LD_PRELOAD="<path/to/libtcmalloc.so>" <other_options> <path/to/binary> <binary_flags>
+```
+
+Using the LD_PRELOAD method is not recommended by the developers of gperftools so using the build options is prefereable. Both CPU and memory profiling can be enabled at the same time but note that turning on memory profiling (or the heap checker) will affect performance.
```
timadye, line 76: maybe codify `LD_PRELOAD`, here and elsewhere.

docs/howto/profiling.md
```diff
@@ -0,0 +1,156 @@
...
+
+```
+LD_PRELOAD="<path/to/first/library> <path/to/second/library>"
+```
+
+Note that these steps don't turn on profiling, they only enable it to work. The following section details how to turn it on.
+
+## Produce a CPU Profile
+
+To turn on CPU profiling when running an executable define the CPUPROFILE environment variable when executing the program:
```
timadye, line 87: Maybe codify `CPUPROFILE`

docs/howto/profiling.md
```diff
@@ -0,0 +1,156 @@
...
+
+### View Profile as a Graph
+
+A graphical representaion of a profile can be produced using:
+
+```
+pprof -pdf <path/to/binary> <path/to/profile> > <path/to/pdf>
+```
+
+Where path/to/binary is the binary is used to produce the profile in the first place.
```
timadye, line 125:
```suggestion
Where `path/to/binary` is the binary is used to produce the profile in the first place.
```
(a better way to make this type of comment)

docs/getting_started.md
```diff
@@ -243,7 +245,7 @@ components.
 | ACTS_BUILD_EVERYTHING               | Build with most options enabled (except HepMC3 and documentation)                                     |
 | ACTS_BUILD_PLUGIN_CUDA              | Build CUDA plugin                                                                                     |
 | ACTS_BUILD_PLUGIN_DD4HEP            | Build DD4hep geometry plugin                                                                          |
-| ACTS_BUILD_PLUGIN_EXATRKX           | Build Exa.TrkX plugin                                                                             |
+| ACTS_BUILD_PLUGIN_EXATRKX           | Build Exa.TrkX plugin                                                                                 |
```
timadye, line 248: Sorry, I'm really nit-picking, but what changed here? Better maybe to leave unchanged text without invisibe changes.

docs/getting_started.md
```diff
@@ -274,7 +276,10 @@ components.
 | ACTS_USE_SYSTEM_EIGEN3              | Use the system eigen3 libraries (defaults to ON)                                                      |
 | ACTS_USE_SYSTEM_VECMEM              | Use system provided vecmem installation                                                               |
 | ACTS_USE_SYSTEM_PYBIND11            | Use pybind11 installed in the system                                                                  |
-| ACTS_USE_SYSTEM_ACTSDD4HEP          | Use ActsDD4hep glue library externally (and don't include it in the build)                            |
+| ACTS_USE_SYSTEM_ACTSDD4HEP          | Use ActsDD4hep glue library externally (and don't include it in the build)                            |
```
timadye, line 279: another invisible change

timadye (changes requested):

docs/getting_started.md
Scott-James-Hurley: Comparing the two file versions in main and this branch shows no change to line `279`. Not sure why the line is highlighted as being changed.

docs/howto/profiling.md
Scott-James-Hurley: Tested this on Mercury and was able to profile ACTS.

docs/howto/profiling.md
Scott-James-Hurley: I've amended the documentation. Tested the changes on Mercury and was able to profile ACTS.

timadye (approved):
looks good to me

robertlangenberg:
Excellent, thanks for this and apologies for not replying earlier. Will merge this in as soon as the ci runs through

asalzburger:
Ok, just approved the workflows, we should be able to merge this today

asalzburger (approved):
This was approved by Tim, re-approving to move forward.

CMakeLists.txt
```diff
@@ -339,6 +342,12 @@ if(ACTS_BUILD_BENCHMARKS OR ACTS_BUILD_INTEGRATIONTESTS OR ACTS_BUILD_UNITTESTS)
   add_subdirectory(Tests)
 endif()
 
+#generate debug information
+if(ACTS_ENABLE_CPU_PROFILING)
+  message(STATUS "Added -g compile flag")
+  add_compile_options(-g)
```
paulgessinger, line 348: I'm a bit concerned about using this to set `-g` globally here. Could this be pushed to `ActsCompilerOptions.cmake` and use `ACTS_CXX_FLAGS` and `CMAKE_CXX_FLAGS` there?

Scott-James-Hurley:
> At least the memory profile should be fairly deterministic, right? Would it be possible / reasonable to run this on a CI machine (even though it might not be quiet)?

I've found that the memory profiler *greatly* increases the execution time of a program. It could be worth trying but it may end up being impractical running with the memory profiler regularly.

CMakeLists.txt
Scott-James-Hurley: I've found that the `-g` option is unnecessary to enable profiling so I've removed it. Thanks for the feedback.

Scott-James-Hurley: I originally thought it was necessary and could've sworn that I'd tried to run the profiler without `-g` to no avail. However, since this was raised I've successfully run the profiler on ACTS and a standalone program without it on two machines and it *seems* fine. I can't find any reference to the flag in gperf's documentation despite it being included in most example articles from other sites I've seen. More than happy to include `-g` again to ensure correct behaviour.

timadye: Sorry, I don't have time to check this right now, but I seem to remember that the ACTS default was to build with debug with optimisation. If so, then `-g` will be there even if you don't add it explicitly. You can check that by looking at the compiler command in the build log, or in the `ps` output when building.

Scott-James-Hurley: @timadye's right, thanks.

paulgessinger (approved):
Let's go.

PR #1276, opened 2022-05-31
fix: Add missing dependency in Exa.TrkX plugin

benjaminhuth:
The plugin depends on the boost-headers, which was not explicitly declared in the `CMakeLists.txt`. I guess this wasn't detected because in the CI environment the boost headers are installed in the standard include paths...


AJPfleger:
looks good and makes sense to me

PR #1278, opened 2022-06-02
feat: Update ITk seeding example

LuisFelipeCoelho:
This PR updates the configuration of the ITk_seeding.py
I will follow up this PR with the CI checks for ITk_seeding.py

@noemina


Core/include/Acts/Seeding/BinnedSPGroup.hpp
```diff
@@ -194,7 +194,7 @@ class BinnedSPGroupIterator {
   BinnedSPGroupIterator(const SpacePointGrid<external_spacepoint_t>* spgrid,
                         BinFinder<external_spacepoint_t>* botBinFinder,
                         BinFinder<external_spacepoint_t>* tBinFinder,
-                        boost::container::small_vector<size_t, 20> bins = {}) {
+                        std::vector<size_t> bins = {}) {
```
AJPfleger, line 197: Maybe I missed it, but why do change the type from `boost::container::small_vector` to `std::vector`?

LuisFelipeCoelho: I changed it to `std::vector` so `zBinsCustomLooping` can be easily configured in `itk_seeding.py`. Also, I don't think we need `boost::container::small_vector` since it won't change after being initialized.

paulgessinger: Yeah no strong need for small vector is it's not modified.

robertlangenberg:
Why did you change the boost::container::small_vector to std::vector?

LuisFelipeCoelho:
@robertlangenberg, I discussed with @paulgessinger and we decided that there is no strong need for small vector as it is not modified.

> I changed it to `std::vector` so `zBinsCustomLooping` can be easily configured in `itk_seeding.py`. Also, I don't think we need `boost::container::small_vector` since it won't change after being initialized.

robertlangenberg (approved):
Right, I hadn't seen the resolved conversation. Other changes look good, let's get this in.

PR #1282, opened 2022-06-08
refactor: Move propagator options to general fitter options

osbornjd:
This PR makes the propagator options a dedicated member of the `GeneralFitterOptions` struct so that it can be explicitly configured for both the KF and GSF.


paulgessinger:
CI Bridge jobs should run now.

osbornjd:
Thanks Paul, I'll take a look at the CI after it is finished and then just let me know if there is anything else I need to do.

osbornjd:
@paulgessinger as far as I can tell the CI passes except for the requirement of a reviewer - is my understanding correct?

paulgessinger (approved):
Seems fine to me!

paulgessinger:
I don't quite get why the CI jobs don't launch. @osbornjd can you push an (empty) commit to try to nudge the CI to do something?

osbornjd:
> I don't quite get why the CI jobs don't launch. @osbornjd can you push an (empty) commit to try to nudge the CI to do something?

I pushed an empty commit but then realized I needed to merge with main. Looks like the CI is running now, though.

paulgessinger (approved):
LGTM!

PR #1283, opened 2022-06-14
ci: Retry nodeps build if it fails the first time

paulgessinger:
For some reason the inclusion of ccache seems to have made the nodeps
build (specifically the boost build) fail sporadically. Since it works
most of the time and we want to keep using ccache for this build as
well, this PR attempts to just rertry the build step once if it fails.


andiwand (approved):
should we reference one of the failed workflows here?

PR #1285, opened 2022-06-22
ci: Bump ccache max size to 500MB

paulgessinger:
This apparently requires some trial & error. After #1270's reduction to 150MB we've seen build times go up again. So let's try 500MB and see.


paulgessinger:
citime_Builds_2
citime_Builds_0-2

andiwand (approved):

Issue #1284, opened 2022-06-21
[Bug] FATRAS has trouble simulating electrons with eta between -4 and -1


Maury-arty:
When given the output of Particle Gun for electrons (uniform distribution of eta from -4 to 4), FATRAS has trouble simulating electrons which have eta between -4 and -1. The output of Particle Gun was manually checked and has a uniform distribution in eta as asked.

Concretely, for some electrons, no hits are generated. Those electrons are all with eta between -4 and -1 (in their initial state).
Also, when plotting the eta distribution of all hits, one can see a depletion for electrons between -4 and -1, whereas this distribution remains uniform for muons.

As can be seen in the spatial plot of all hits, in the case of electrons, there is a lack of hits in the "rear" part of the detector while in the muon case, the whole detector has hits.

This issue also arises when using the ITk geometry, the ATLAS B field, different regions of pT, both signs for particles...

To reproduce the error (I used the (deprecated) executables), with the TrackML geometry and a constant B field:
`ActsExampleParticleGun --events=10000 --output-dir=1_particle_gun --output-csv --gen-vertex-xy-std-mm=0.012 --gen-vertex-z-std-mm=50 --gen-eta=-4:4 --gen-eta-uniform --gen-pdg=11 --gen-mom-gev=1:5 --gen-mom-transverse`
`ActsExampleFatrasGeneric --input-dir=1_particle_gun --output-dir=2_fatras --output-csv --remove-neutral --bf-constant-tesla=0:0:2`

Replace --gen-pdg option with 13 for muons (mu-).

![spatial_hits_electrons](https://user-images.githubusercontent.com/62028932/174828186-b218e9ee-fa91-4d50-83c8-ae650693e109.png)
![spatial_hits_muons](https://user-images.githubusercontent.com/62028932/174828192-f7c812da-509c-44d7-bfa1-3bb9bcb590cd.png)
![fatras_hits_muons_electrons_eta](https://user-images.githubusercontent.com/62028932/174828195-61d47245-325f-4207-93ca-adc86f8097f8.png)
![initial_state_electrons_eta](https://user-images.githubusercontent.com/62028932/174828201-111c7454-7483-4c36-919a-cfe116addcfa.png)

andiwand:
removing the material with `--mat-input-type=none` or turning off EM radiation with `--fatras-em-radiation=0` removes the artifacts

[...]

PR #1288, opened 2022-06-22
fix: SimulationActor produces negative MaterialSlab thickness

andiwand:
fix https://github.com/acts-project/acts/issues/1284

SimulationActor produces negative MaterialSlab thickness if `normal` and `before` produce an angle > 90°

apply `abs` to avoid negative scaling factors


Core/src/Material/MaterialSlab.cpp
```diff
@@ -22,6 +23,9 @@ Acts::MaterialSlab::MaterialSlab(const Material& material, float thickness)
       m_thickness(thickness),
       m_thicknessInX0((eps < material.X0()) ? (thickness / material.X0()) : 0),
       m_thicknessInL0((eps < material.L0()) ? (thickness / material.L0()) : 0) {
+  if (thickness < 0) {
```
andiwand, line 26: not sure if you usually do such runtime checks in Acts but they were very useful for debugging this problem. I am a fan of fail fast

asalzburger: Fine with me, consistency checking on construction is something I support, over-checking everything would certainly have a performance penalty.

asalzburger (approved):
Straight-forward bug-fix.

paulgessinger (approved):

PR #1292, opened 2022-06-27
test: Parametrize tests on geometry, add minimal GSF reproducibility test

paulgessinger:
Uses a fixture to more elegantly parametrize the CKF tracks and truth
tracking tests with the generic detector and the ODD. Add a truth
tracking GSF test with hash checks. Also adjusts `truth_tracking_gsf.py`
slightly to mirror `truth_tracking_kalman.py` a bit more closely.


benjaminhuth (approved):
Looks good to me! Thanks for adding the check to the GSF!

paulgessinger:
The GSF currently produces a large number of WARNINGs and ERRORs. I'm working on making the job fail on these (we've had this before but had to disable it for CKF+ODD). I'll do this in a follow-up PR, maybe you could then help me look into the ERRORs?

paulgessinger:
Hashes updated. Truth track kalman generic are all new, ODD changed a few because it's now with material, GSF + ODD & generic is all new.

@benjaminhuth can you reapprove?

paulgessinger:
@benjaminhuth can you reapprove this?

benjaminhuth (approved):

PR #1293, opened 2022-06-27
feat: TGeoLayerBuilder has configurable detector element factory

paulgessinger:
This allows customizing what concrete detector element is constructed for the detector geometry.


paulgessinger:
@osbornjd I created this quickly to give an idea of how I think this could work.
Let me know if this looks like it does what you need. Otherwise, we can close and replace this PR with one by you!

paulgessinger:
Hey @osbornjd. That's actually a good point. I didn't realize you were using `TGeoDetector` since it's part of the Examples code, and comes with all the caveats associated. On the other hand, it doesn't hurt in principle to add this to the config chain like you suggest.

I don't see however how this could be read from the command line for the executables though. Would you be able to set the sPHENIX-specific factory from code?

osbornjd:
> I don't see however how this could be read from the command line for the executables though. Would you be able to set the sPHENIX-specific factory from code?

I started thinking about this last night and realized it wouldn't make sense. You'd have to provide an option to specify the factory to build sPHENIX detector elements, and then there would have to be some flag in the layer builder like `if(cfg.sphenixoption) buildsPHENIXDetectors()` or whatever. That definitely doesn't scale. Which would imply we would have to change the `TGeoLayerBuilder` to suit our specific needs, but that isn't very nice either because it would then require each experiment to have their own version of the layer builder. Let me think on it some more...

osbornjd:
Just to point to what we are working with, we basically emulated the code [here](https://github.com/acts-project/acts/blob/a4a1b96f3fe26b2de458ee71b100b66677295c33/Examples/Run/Geometry/TGeoGeometryExample.cpp#L18) and e.g. [here](https://github.com/acts-project/acts/blob/a4a1b96f3fe26b2de458ee71b100b66677295c33/Examples/Run/Common/src/GeometryExampleBase.cpp#L62) to build our tracking geometry from the TGeo model. So in principle if the TGeoDetecotr object had its own config that could take a factory method which creates an experiment specific detector element, that would be sufficient. 

See our implementation [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/bf1e80f13e94eb452a8b5f89af076ac8c4f607ad/offline/packages/trackreco/MakeActsGeometry.cc#L540). 

Otherwise, the option would be to modify the LayerBuilder code in our fork/branch to explicitly build the sPHENIX detector element class. This would in principle be straightforward also, but maybe isn't the best idea as we are then modifying upstream code that may change in future version of Acts.

osbornjd:
> If we then add the factory function to TGeoDetector::Config you should be all set, right?

I think that is right.

paulgessinger:
Hey @osbornjd, I added the factory option to `TGeoDetector` and carry it through to `TGeoLayerBuilder`. Can you have another look at this?

osbornjd (commented):
This looks good to me, I think it should be fine to merge. 

I can test it out in our environment once we integrate the next Acts version release into our stack. But the overall configurable design makes sense and looks like it can be configured at the highest level of the geometry building.

paulgessinger:
@osbornjd if you're happy, can you approve the PR?

osbornjd (approved):
Sorry Paul, looks good to me.

Issue #881, opened 2021-07-15
JsonMaterialWriter performs it's writes in the destructor


paulgessinger:
I understand why this is done, but it is likely not a great situation. I think writing should occur at the end of `Sequencer::run` and not when the sequencer is deallocated.

I propose we either:

1. Refactor `JsonMaterialWriter` (and consequently `MaterialMapping`) so that it can become an `IWriter`, that gets an `endRun` call when the event loop finishes.
2. Add functionality to `Sequencer` such that algorithms also receive an `endRun` (or `finalize`?) call, so that `MaterialMapping` can call a write function on `JsonMaterialWriter`.

What do you think @Corentin-Allaire?

paulgessinger:
I think the actual `JsonMaterialWriter` actually doesn't do that anymore. I think there are other writers that still do, renaming the issue and keeping it open for now.

paulgessinger:
As discussed in #1296, refactoring for the purposes of this issue needs to be done carefully, so ROOT file closure happens in the destructor still, so that it also occurs when an exception is encountered.

PR #1296, opened 2022-06-28
fix: Output ROOT file closing on destruct

paulgessinger:
While I believe the writing should happen in `endRun`, I've seen
segfaults when ROOT's exit handler tries to close them and sees
corrupted memory. This PR should fix that for these two classes.


AJPfleger:
Does this problem affect only these two classes? I am asking, because we use the same approach for some/all other writer-classes.

paulgessinger:
We certainly could. I saw crashes for these two classes specifically, so those are the ones I touched for now.

AJPfleger (approved):
The changes look good to me, but I think we should do one of the following:
- Check and adapt the other writer-classes
- Open an Issue to not forget about ^ in the future

AJPfleger:
Yes, [#881](https://github.com/acts-project/acts/issues/881) seems well suited.

PR #1286, opened 2022-06-22
refact python examples vertex_fitting.py

andiwand:
in order to use `vertex_fitting.py` in ODD I need to modify the `trackParameters` to use a different name to match the one produced by [`seeding.py`](https://github.com/acts-project/acts/blob/main/Examples/Scripts/Python/seeding.py#L362)

PR #1299, opened 2022-06-29
fix: full chain vertexing

andiwand:
fix full chain odd vertexing added here https://github.com/acts-project/acts/pull/1286

some early discussion here https://mattermost.web.cern.ch/acts/pl/m6psyge3apgcucesrgojeaajny

- output track parameters from `TrackFindingAlgorithm`
- refactor error handling in `*VertexFinderAlgorithm`
- rewire default whiteboard names in python examples
- use IVF in `full_chain_odd.py` with 2 muons (which should give us an actual vertex)


andiwand:
there is actually an in-chain vertexing https://github.com/acts-project/acts/blob/main/Examples/Python/tests/test_examples.py#L949

Examples/Algorithms/Vertexing/src/AdaptiveMultiVertexFinderAlgorithm.cpp
```diff
@@ -129,11 +129,11 @@ ActsExamples::AdaptiveMultiVertexFinderAlgorithm::execute(
...
 
-    if (not result.ok()) {
+    if (result.ok()) {
+      vertices = std::move(result.value());
+    } else {
       ACTS_ERROR("Error in vertex finder: " << result.error().message());
-      return ProcessCode::ABORT;
     }
-    vertices = *result;
   }
```
AJPfleger, line 137: Thanks for cleaning this up!

AJPfleger (approved):
Looks good to me.

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -307,6 +307,7 @@ int ActsExamples::Sequencer::run() {
               StopWatch sw(localClocksAlgorithms[ialgo++]);
               ACTS_VERBOSE("Execute algorithm: " << alg->name());
               if (alg->execute(++context) != ProcessCode::SUCCESS) {
+                ACTS_ERROR("Failed to execute algorithm: " << alg->name());
```
paulgessinger, line 310: This could even be FATAL I guess.
```suggestion
                ACTS_FATAL("Failed to execute algorithm: " << alg->name());
```

Examples/Python/tests/test_examples.py
```diff
@@ -1021,5 +1021,7 @@ def test_full_chain_odd_example(tmp_path):
     assert script.exists()
     env = os.environ.copy()
     env["NEVENTS"] = "1"
-    env["ACTS_LOG_FAILURE_THRESHOLD"] = "WARNING"
+    # increased threshold because of vertexing
+    # see https://github.com/acts-project/acts/pull/1299
+    env["ACTS_LOG_FAILURE_THRESHOLD"] = "FATAL"
```
paulgessinger, line 1026: Right.  This is where it gets tricky then. I guess under "normal" operation, there should be no ERRORs, or conversely, ERRORs (or even WARNINGs) should signal some kind of problem with the code.

Now that the vertexing throws ERRORs but they're not intrinsically fatal anymore, I'm not entirely sure what the best approach here is.

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -67,11 +67,24 @@
...
+        absEtaMax=2.5,
+        loc0Max=4.0 * u.mm,  # rho max
+        ptMin=500 * u.MeV,
+    )
+)
 s = addVertexFitting(
     s,
     field,
-    vertexFinder=VertexFinder.Truth,
+    vertexFinder=VertexFinder.Iterative,
```
paulgessinger, line 85: Can we try the AMVF here? (This is what ATLAS uses now)

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -67,11 +67,24 @@
...
     outputDirRoot=outputDir,
+    logLevel=acts.logging.VERBOSE,
```
paulgessinger, line 87: Should this be verbose by default?

paulgessinger (commented):
We just need to figure out what to do with the new ERRORs.

@asalzburger do you have any thoughts on this?

Examples/Python/tests/test_examples.py
andiwand: I thought the situation is similar. we ultimately don't want to have errors under such nice conditions but for now we have some. I thought this is due to not having an ambiguity resolution? or do you think there is another problem upstream?

Examples/Scripts/Python/full_chain_odd.py
andiwand: sure we can have AMVF

Examples/Scripts/Python/full_chain_odd.py
andiwand: nope :building_construction:

Examples/Python/tests/test_examples.py
paulgessinger: I'd say the difference is this here right now isn't really a "bug" in the strict sense.

andiwand: true. I also don't like to push the threshold up again but if we want to add vertexing right now I don't see an alternative. or is there?

paulgessinger: We could make it configurable I guess...

paulgessinger:
Another thing I was just thinking of is if we should add vertexing performance plots to the output and consequently the monitoring.

Examples/Python/tests/test_examples.py
andiwand: you mean not run the vertexing in this case but somewhere else for example?

paulgessinger:
No I actually meant configuring whether the vertexing algorithm should log this as `ERROR`/`WARNING` or just as info.

Examples/Python/tests/test_examples.py
paulgessinger: I was just thinking of making it configurable whether the vertexing algorithm logs this as `WARNING`/`ERROR` or just as `INFO`.

paulgessinger:
Yeah right. Can you create an issue to add vertexing monitoring to `physmon.py` so we can remember it?

andiwand:
added an option now @paulgessinger

andiwand:
no the `EigenStepper` fails... https://github.com/acts-project/acts/runs/7282100728?check_suite_focus=true#step:8:377

I guess it does not make lots of sense to surround these errors anymore but fix it directly where it happens which is the input tracks I guess

paulgessinger:
Hm. `12:34:26    HelTrkLinPro   ERROR     Step failed with EigenStepperError:3: Step size adjustment exceeds maximum trials` means the stepper is unable to adjust the step size down to meet the accurracy target. This seems to be triggered from the linearization. 

@asalzburger do you have any ideas here?

andiwand:
https://github.com/acts-project/acts/runs/7285048520?check_suite_focus=true#step:8:581

```
15:22:42    HelTrkLinPro   ERROR     Step failed with EigenStepperError:3: Step size adjustment exceeds maximum trials
```

this is the fishy one like you said @paulgessinger right?

paulgessinger:
Exactly, and it seems like it's the actual step size adjustment that runs into the limit. This I believe is where it wants to reduce step size until it hits the accuracy target.

andiwand:
so I think this is where it comes from https://github.com/acts-project/acts/blob/8a2260b530a7b248ca38ed99be6ae7c081dcf48a/Core/include/Acts/Propagator/EigenStepper.ipp#L220-L225

I can see what the step size is doing but I feel like I would approach the problem from the wrong end

andiwand:
this brach includes now changes from https://github.com/acts-project/acts/pull/1351 and https://github.com/acts-project/acts/pull/1347 in order to test it. we should merge those other PRs first

andiwand:
sadly this still does not work. I tried to use more particles for each event. I suspected that missing seeds for one of the particles would put the vertex fit in a bad position. I think this is still true but since there are so many duplicates for each track the order in which they go into the vertexing could still result into the same problem

the good news is that using https://github.com/acts-project/acts/pull/1363 seems to eliminate all the problems

andiwand:
after merging #1363 the full chain vertexing works now

andiwand:
ready @paulgessinger ?

Examples/Algorithms/Vertexing/src/VertexFitterAlgorithm.cpp
```diff
@@ -118,15 +122,19 @@ ActsExamples::ProcessCode ActsExamples::VertexFitterAlgorithm::execute(
       if (fitRes.ok()) {
         fittedVertices.push_back(*fitRes);
       } else {
-        ACTS_ERROR("Error in vertex fit with constraint.");
-        ACTS_ERROR(fitRes.error().message());
+        ACTS_ERROR("Error in constrainted vertex fitter: "
```
paulgessinger, line 125:
```suggestion
        ACTS_ERROR("Error in constrained vertex fitter: "
```

paulgessinger:
Looks ok, let's go with IVF first and look into AMVF. Do you know if AMVF runs into errors in the current status?

andiwand:
> Looks ok, let's go with IVF first and look into AMVF. Do you know if AMVF runs into errors in the current status?

AMVF will produce an error in the current scenario. I guess it is because we loose one of the muons and then fail to fit the vertex. with 4 muons AMVF works but IVF even works with 2 but I am not sure why

paulgessinger (approved):

PR #1301, opened 2022-07-06
fix: json gcc runtime error

andiwand:
I was facing a runtime error at https://github.com/acts-project/acts/blob/main/Plugins/Json/include/Acts/Plugins/Json/GeometryHierarchyMapJsonConverter.hpp#L110

after applying the changes everything is back to normal

looks like a problem with gcc and nlohmann/json?

```
gcc (Ubuntu 11.2.0-19ubuntu1) 11.2.0
```


andiwand:
the `static constexpr const char* `change was actually not critical but I left it in because it think it is more correct?

first I was not even able to print `kHeaderKey` for example

the critical change was unrolling the initialization list but I really don't know why it caused a runtime error

HadrienG2:
Note that compiler runtime errors are always bugs and should be reported as such to the compiler developers : https://gcc.gnu.org/bugzilla/

andiwand:
@HadrienG2 I thought it could be due to a bug in the libraryl like UB that turns out okay with clang but not gcc. or they use a language "feature" that does not exist in the standard. not sure how likely those cases are compared to each other tho

andiwand:
@HadrienG2 sorry I think I messed things up with my poor PR title. there was not a runtime error while compiling with gcc but a runtime error unfolded when executing Acts compiling with gcc while it works fine with clang

HadrienG2:
Ah, sorry for misinterpreting on my side as well.

A simplest check that is quite effective at discriminating between UB and other issues is to rebuild without optimizations. If the runtime error goes away in this configuration, then it's likely to be a UB problem.

HadrienG2:
...and if the bug is still there at -O0, then you can build with debug symbols (already enabled by CMake Debug build), and use GDB to locate the precise point at which the program is crashing, which is often quite helpful.

PR #1302, opened 2022-07-06
test: remove kalman smoother related log thresholds

### LINKED ISSUE #1215: Gain Matrix Update fails sometimes for the ODD

andiwand:
after having #1298 in we should be able to remove these thresholds

Closes #1215

### DIFF
--- a/Examples/Python/tests/test_examples.py
+++ b/Examples/Python/tests/test_examples.py
@@ -344,6 +344,7 @@ def test_event_recording(tmp_path):
 
     env = os.environ.copy()
     env["NEVENTS"] = "1"
+    env["ACTS_LOG_FAILURE_THRESHOLD"] = "WARNING"
     subprocess.check_call([str(script)], cwd=tmp_path, env=env)
 
     from acts.examples.hepmc3 import HepMC3AsciiReader
@@ -878,8 +879,7 @@ def test_ckf_tracks_example(
         s=s,
     )
 
-    with failure_threshold(acts.logging.FATAL, enabled=detector_config.name == "odd"):
-        s.run()
+    s.run()
 
     del s  # files are closed in destructors, not great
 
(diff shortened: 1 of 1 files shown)


Examples/Python/tests/test_examples.py
```diff
@@ -344,9 +344,6 @@ def test_event_recording(tmp_path):
 
     env = os.environ.copy()
     env["NEVENTS"] = "1"
-    # @TODO: Fix failure in gain matrix smoothing
-    # See https://github.com/acts-project/acts/issues/1215
-    env["ACTS_LOG_FAILURE_THRESHOLD"] = "WARNING"
```
paulgessinger, line 349: Sorry I wasn't clear, this should stay I think.

Examples/Python/tests/test_examples.py
```diff
@@ -1026,7 +1021,4 @@ def test_full_chain_odd_example(tmp_path):
     assert script.exists()
     env = os.environ.copy()
     env["NEVENTS"] = "1"
-    # @TODO: Fix failure in gain matrix smoothing
-    # See https://github.com/acts-project/acts/issues/1215
-    env["ACTS_LOG_FAILURE_THRESHOLD"] = "FATAL"
```
paulgessinger, line 1031: I think this should just be changed to `WARNING`.

Examples/Python/tests/test_examples.py
```diff
@@ -881,8 +878,6 @@ def test_ckf_tracks_example(
         s=s,
     )
 
-    # @TODO: Fix failure in gain matrix smoothing
-    # See https://github.com/acts-project/acts/issues/1215
```
paulgessinger, line 885: This entire would need to go I believe.

paulgessinger (approved):

PR #1305, opened 2022-07-07
feat: Hough Transform first implementation

jahreda:
This is a first implementation of a Hough Transform into the ACTS code (also my first PR into ACTS). I think I have found some decent solutions for all the problems and questions I had before. The HT implementation is largely based on the ATLAS EF Tracing Hough Transform. The implementation is incomplete in that it only has example region selection and z sub-region slicing, but this can be implemented for specific geometries.

Currently the code exists as an Example, so I would like some feedback as to where best to put this (from previous discussions it wasn't fully clear). 

Just to show it's working, from 4 muon events in the barrel (narrow phi region):
22:36:33    SeedingPerfo   INFO      Efficiency (nMatchedParticles / nAllParticles) = 0.994784
22:36:33    SeedingPerfo   INFO      Fake rate (nUnMatchedSeeds / nAllSeeds) = 0.617954
22:36:33    SeedingPerfo   INFO      Duplication rate (nDuplicatedMatchedParticles / nMatchedParticles) = 0.997378
22:36:33    SeedingPerfo   INFO      Average number of duplicated seeds ((nMatchedSeeds - nMatchedParticles) / nMatchedParticles) = 26.7891
22:36:33    SeedingPerfo   INFO      Wrote performance plots to '/afs/cern.ch/work/j/jahreda/acts_april28/acts/data/output_four_muons_hough/performance_seeding_hists.root:/'
22:36:33    RootTrackPar   INFO      Write estimated parameters from seed to tree 'estimatedparams' in '/afs/cern.ch/work/j/jahreda/acts_april28/acts/data/output_four_muons_hough/estimatedparams.root'
22:36:34    Sequencer      INFO      Processed 10000 events in 79.477941 s (wall clock)
22:36:34    Sequencer      INFO      Average time per event: 77.954230 ms/event


robertlangenberg (commented):
Hi Jahred, thanks for this, I'd suggets to make this configurable (layers, histogram size etc.) and move it into Core as an algorithm, e.g. next to the Seeding. Then you could use this algorithm in this example you created, such that people can use the algorithm directly, using your example as guideline. Do be careful though, examples tend to be copy-pasted and I've been plagued by nonsensical defaults used by different experiments. In the same vein, it'd be really important to have as much in-place documentation of the code as possible.

As for performance, the performance improvement trick that Andi mentioned would be quite helpful here, having the histogram location of a measurement stored in both a measurement wrapper and in the histogram itself, such that only looping over the measurements is required: 
1) loop over measurement wrappers, 
2) check in histogram location retrieved from measurement, 
3) mark measurement as valid or invalid.

If I understood correctly, you're so far looping over the whole histogram (HoughTransformSeeder.cpp:255), with 1.5M bins that should take much longer than iterating over measurements per layer.

Let me know if you think this is reasonable.

jahreda:
Thanks, Robert. It's great to have someone taking a look.  A few questions:

1) Isn't the code configurable? For example:
* Number of layers: https://github.com/jahreda/acts/blob/jahredaHTDev/Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp#L133
* Binning (number of bins): https://github.com/jahreda/acts/blob/jahredaHTDev/Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp#L146
and range:
(https://github.com/jahreda/acts/blob/jahredaHTDev/Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp#L141)

Were you thinking this has to be done in some other ways? If so, can you point me to how you'd like me to set up the config?

2) Moving this to Core makes sense - but do you have examples (of the seeding or otherwise) for how you suggest doing that and also providing an "example"? Outside of the functions here:
https://github.com/jahreda/acts/blob/jahredaHTDev/Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/DefaultHoughFunctions.hpp

I think nothing is actually geometry dependent.

3) I'm also curious what you mean by "next to the Seeding"

4) Finally, I think I maybe misunderstood what Andi was suggesting for measurements or maybe I'm simply not remembering. You are right that the code does loop over all the measurements. I think no matter what one has to loop over all them at least once. What I do is loop over them and pull out just a handful of pieces of information: layer, phi, r, z and index. This is done here:
https://github.com/jahreda/acts/blob/jahredaHTDev/Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp#L210
When we loop over the bins in a layer we can then quickly check if we are in the right layer or not. And if not we can skip the measurement:
https://github.com/jahreda/acts/blob/jahredaHTDev/Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp#L309

In principle we could speed that up by storing the above separately by layer. Is that what you propose? Or something else?

Best
Jahred

jahreda:
Hi @robertlangenberg 

Just pinging this thread to see if you had a chance to look at my replies. I have been on vacation, traveling and then dealing with a sick family so I haven't made our Tuesday meeting recently. And this coming week I'm home with my son on Tuesday, but maybe best to iterate here, anyway.

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp
```diff
@@ -0,0 +1,267 @@
...
+    unsigned m_imageSize_x = 7000;  // i.e. number of bins in phi_track
+    unsigned m_imageSize_y = 216;   // i.e. number of bins in q/pT
+
+    std::vector<unsigned> m_hitExtend_x = {
+        1, 1, 0, 0, 0, 0,
+        0, 0, 0, 0};  // Hit lines will fill extra bins in x by this amount on
+                      // each side, size == nLayers
+
+    /// === Seeds for Hough ==
+    std::vector<int> m_threshold = {9};  // Minimum point value post-convolution
```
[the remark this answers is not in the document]
robertlangenberg: also add that the center element in the vector corresponds to the current phi and q_over_pT bin, while all other elements correspond to the neighbors (such that you'd usually want the values to be mirrored?), the closer to begin and end, the further away from the center bin.

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,529 @@
...
+        if (i == 0 && j == 0)
+          continue;
+        if (y + j < image.size(0) && x + i < image.size(1)) {
+          if (image(y + j, x + i).first > image(y, x).first)
+            return false;
+          if (image(y + j, x + i).first == image(y, x).first) {
+            if (image(y + j, x + i).second.size() > image(y, x).second.size())
+              return false;
+            if (image(y + j, x + i).second.size() ==
+                    image(y, x).second.size() &&
```
robertlangenberg, line 387: having y first and x second as parameters for image gets my heart rate up, I didn't check if it's correct, if so it'd confuse me even more 😅

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,529 @@
...
+
+  static thread_local ProtoTrackContainer protoTracks;
+  protoTracks.clear();
+
+  for (auto subregion : m_cfg.m_subRegions) {
+    ActsExamples::Image m_image =
+        createImage(spacePointPtrs, measurementStructs, subregion);
+
+    for (unsigned y = 0; y < m_cfg.m_imageSize_y; y++)
+      for (unsigned x = 0; x < m_cfg.m_imageSize_x; x++)
```
robertlangenberg, line 226: here you're looping over the whole 7000 x 216 bins, you can avoid that by looping over the MeasurementStructs (and SpacePoints, if you decide to keep them) only, which should be 1 or 2 orders of magnitude less. From each measurementStruct, you can jump into the Image. In the Image it would be enough to store a single bit for each layer that has been hit.

Examples/Run/Reconstruction/Common/HoughExample.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+  houghCfg.m_subRegions = {0, 1};
+  houghCfg.m_traceHits = true;
+
+  houghCfg.m_xMin = -0.05;  // minphi
+  houghCfg.m_xMax = 0.25;   // maxphi
+  houghCfg.m_yMin = -1.25;  // min q/pt, -1/0.8 GeV  for extension
+  houghCfg.m_yMax = 1.25;   // max q/pt, -1/0.8 GeV  for extension
+
+  houghCfg.m_imageSize_x = 216;
```
robertlangenberg, line 138: the default is 7000, is this on purpose the same as m_imageSize_y? This would mean ~40000 bins. With these values I can believe the 70ms, with the 7000 phiBins it'd wager it would go out of memory or take *very* long.

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,529 @@
...
+  for (unsigned index = 0; index < spacepoints.size(); index++) {
+    const SimSpacePoint* sp = spacepoints[index];
+    double r = sqrt(sp->x() * sp->x() + sp->y() * sp->y());
+    double z = sp->z();
+    unsigned hitlayer = (*(m_cfg.findLayerIDSP))(r);
+    if (hitlayer != layer)
+      continue;
+    if (!((*(m_cfg.inSliceSP))(z, hitlayer, subregion)))
+      continue;
+    float phi = atan2(sp->y(), sp->x());
```
robertlangenberg, line 290: use std::atan2

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,529 @@
...
+
+ActsExamples::Image ActsExamples::HoughTransformSeeder::createLayerImage(
+    unsigned layer, std::vector<const SimSpacePoint*>& spacepoints,
+    std::vector<std::shared_ptr<const MeasurementStruct>>& meass,
+    int subregion) const {
+  ActsExamples::Image image(m_cfg.m_imageSize_y, m_cfg.m_imageSize_x);
+
+  for (unsigned index = 0; index < spacepoints.size(); index++) {
+    const SimSpacePoint* sp = spacepoints[index];
+    double r = sqrt(sp->x() * sp->x() + sp->y() * sp->y());
```
robertlangenberg, line 283: use std::sqrt

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,529 @@
...
+    double z = sp->z();
+    unsigned hitlayer = (*(m_cfg.findLayerIDSP))(r);
+    if (hitlayer != layer)
+      continue;
+    if (!((*(m_cfg.inSliceSP))(z, hitlayer, subregion)))
+      continue;
+    float phi = atan2(sp->y(), sp->x());
+
+    // This scans over y (pT) because that is more efficient in memory, in C.
+    // Unknown if firmware will want to scan over x instead.
```
robertlangenberg, line 293: firmware?

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,529 @@
...
+    for (unsigned y_ = 0; y_ < m_cfg.m_imageSize_y; y_++) {
+      unsigned y_bin_min = y_;
+      unsigned y_bin_max = (y_ + 1);
+
+      // Find the min/max x bins
+      auto xBins = yToXBins(y_bin_min, y_bin_max, r, phi, hitlayer);
+      // Update the image
+      for (unsigned y = y_bin_min; y < y_bin_max; y++)
```
robertlangenberg, line 301: this loops over exactly one entry, and that's `y_`. the loop is not necessary if i'm not mistaken.

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,529 @@
...
+        for (unsigned x = xBins.first; x < xBins.second; x++) {
+          image(y, x).first++;
+          image(y, x).second.insert(index);
```
robertlangenberg, line 304: this... adds... each spacepoint of the current layer to EVERY x AND y bin of this layerImage... no?

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,529 @@
...
+      // Update the image
+      for (unsigned y = y_bin_min; y < y_bin_max; y++)
+        for (unsigned x = xBins.first; x < xBins.second; x++) {
+          image(y, x).first++;
+          image(y, x).second.insert(
+              index +
+              spacepoints
+                  .size());  // add spacepoints size to differentiate them
+        }
+    }
```
robertlangenberg, line 334: same comments as above

robertlangenberg (changes requested):
Hi @jahreda, sorry for the late review. I added some comments, it'd probably be helpful to have a voice chat over how to loop over the measurements/spacepoints instead of the whole histogram and how to reduce the histogram size.

Examples/Run/Reconstruction/Common/HoughExample.cpp
```diff
@@ -0,0 +1,214 @@
...
+#include "ActsExamples/Utilities/Paths.hpp"
+
+#include <memory>
+
+#include <boost/program_options.hpp>
+
+using namespace Acts::UnitLiterals;
+using namespace ActsExamples;
+
+int runHoughExample(int argc, char* argv[],
```
tboldagh, line 40: I wonder if, instead of c++ example we should enable python bindings and prepare python script for running this example.

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,529 @@
...
+                  cov.block<2, 2>(Acts::eBoundLoc0, Acts::eBoundLoc0);
+              return std::make_pair(lpar, lcov);
+            },
+            measurements[sourceLink.get().index()]);
+
+        // transform local position to global coordinates
+        Acts::Vector3 globalFakeMom(1, 1, 1);
+        Acts::Vector3 globalPos =
+            surface->localToGlobal(ctx.geoContext, localPos, globalFakeMom);
+        double r = sqrt(globalPos[Acts::ePos0] * globalPos[Acts::ePos0] +
```
tboldagh, line 205: that can be conveniently replaced by `hypot(globalPos[Acts::ePos0], globalPos[Acts::ePos1])`

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,529 @@
...
+          /* now we need to unpack the hits; there should be multiple track
+             candidates if we have multiple hits in a given layer So the first
+             thing is to unpack the indices (which is what we need) by layer */
+          std::vector<std::vector<Index>> hitIndicesAll(
+              m_cfg.m_nLayers);  // [layer,Index]
+          std::vector<size_t> nHitsPerLayer(m_cfg.m_nLayers);
+          for (auto index : m_image(y, x).second) {
+            if (index < spacePointPtrs.size()) {  // is a spacepoint
+              const SimSpacePoint* sp = spacePointPtrs[index];
+              double r = sqrt(sp->x() * sp->x() + sp->y() * sp->y());
```
tboldagh, line 237: `hypot`

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
tboldagh: Or better `std::hypot`

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/DefaultHoughFunctions.hpp
```diff
@@ -0,0 +1,150 @@
+// This file is part of the Acts project.
+//
+// Copyright (C) 2020-2021 CERN for the benefit of the Acts project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#pragma once
+
```
tboldagh, line 10: Would it be better if these functions go to some namespace?

Examples/Run/Reconstruction/Common/HoughExample.cpp
```diff
@@ -0,0 +1,214 @@
...
+  houghCfg.inputSourceLinks = digiCfg.outputSourceLinks;
+  houghCfg.geometrySelection = {Acts::GeometryIdentifier().setVolume(0)};
+
+  houghCfg.m_subRegions = {0, 1};
+  houghCfg.m_traceHits = true;
+
+  houghCfg.m_xMin = -0.05;  // minphi
+  houghCfg.m_xMax = 0.25;   // maxphi
+  houghCfg.m_yMin = -1.25;  // min q/pt, -1/0.8 GeV  for extension
+  houghCfg.m_yMax = 1.25;   // max q/pt, -1/0.8 GeV  for extension
```
robertlangenberg, line 136:
```suggestion
  houghCfg.m_yMax = 1.25;   // max q/pt, +1/0.8 GeV  for extension
```

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp
```diff
@@ -0,0 +1,267 @@
...
+    /// e.g. single-hit space points for pixel-like detectors or double-hit
+    /// space points for strip-like detectors.
+    /// Note that we don't *need* spacepoints (measurements can be used instead)
+    std::vector<std::string> inputSpacePoints;
+    /// Output track seed collection.
+    std::string outputSeeds;
+    /// Output hough track collection.
+    std::string outputProtoTracks;
+    /// Input source links collection.
+    std::string inputSourceLinks;
```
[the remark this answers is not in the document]
jahreda: I'm going to be honest here - I'm not certain how the SP formation works fully, I hope to learn this at the tutorial :)

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp
```diff
@@ -0,0 +1,267 @@
...
+    /// Trace each hit that goes in a bin
+    /// Disabling this will save memory/time since each bin doesn't have to
+    /// store all its hits but the roads created won't have hits from
+    /// convolution, etc.
+    bool m_traceHits = true;
+
+    float m_xMin = 0;            // minphi
+    float m_xMax = 2 * 3.14159;  // maxphi
+    float m_yMin = -1.0;         // min q/pt, -1/1 GeV
+    float m_yMax = 1.0;          // max q/pt, +1/1 GeV
```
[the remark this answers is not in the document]
jahreda: Was sort of hoping to try and enforce the idea that this could be implemented in other ways (we have an eta-z0 HT that has been worked on in EFTracking, for example). I'm not set on that if you think it really is important, of course

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp
```diff
@@ -0,0 +1,267 @@
...
+
+    unsigned m_imageSize_x = 7000;  // i.e. number of bins in phi_track
+    unsigned m_imageSize_y = 216;   // i.e. number of bins in q/pT
```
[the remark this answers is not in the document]
jahreda: I've added some comments, let me know if that is sufficient!

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
jahreda: I think the idea was to consider other options in the future that might be more complicated

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
jahreda: No, see our previous discussion

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
jahreda: No, see previous discussion

Examples/Run/Reconstruction/Common/HoughExample.cpp
jahreda: I agree, I think we discussed that we could have a follow-up PR if that is OK?

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/DefaultHoughFunctions.hpp
jahreda: Is there a recommendation for this? It's not clear if this should go into ActsExample? Otherwise?

jahreda:
I pushed a significant number of updates. The only *major* things I think left (aside from the python bindings) are understanding how best to provide the inputs to the algorithm. It's not clear to me how best to do that as part of ActsExamples::Sequencer, so for now I've left the SP and measurement duality and just moved the functions to create the internal structure

Examples/Run/Reconstruction/Common/HoughExample.cpp
tboldagh: We discussed .. and concluded that it would be better to have followup PR.

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/DefaultHoughFunctions.hpp
tboldagh: What about:
```
namespace ActsExamples {
  namespace DefaultHoughFunctions {
  }
}
```
A bit mouthful but will be clear where these functions came from.

jahreda: OK, I added that, fair

tboldagh:
It seems we have hit the issue with lack of "core" SpacePoint interface. 
To make a progress I suggest: 
1) merge HT as an example algorithm. 
2) work out what to do with the SP (SP collection presumably) - with high priority
3) move the HT to core (presumably after seeing some early results)

@paulgessinger @asalzburger - what do you think?

jahreda:
Thank you for the follow-up, @tboldagh. I think your plan makes sense if @paulgessinger and @asalzburger agree, and in that case hopefully they can start another review so we can proceed

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/DefaultHoughFunctions.hpp
```diff
@@ -0,0 +1,116 @@
...
+      using ResultBool = Acts::Result<bool>;
+      using ResultUnsigned = Acts::Result<unsigned>;
+            
+      ResultDouble fieldCorrectionDefault(unsigned region, double y, double r) {
+         if (region == 999)
+            return y + r;  // this should not be found, for now this is a dummy to show what one *could* do
+         return ResultDouble::success(0.0);
+      }
+      
+      ResultUnsigned findLayerIDDefault(double r) {
```
asalzburger, line 39: To make this configurable, wouldn't that be a good place to use the `Delegate<>` from the Utilities, that comes with practically zero cost overhead and you can make this customisable.

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/DefaultHoughFunctions.hpp
```diff
@@ -0,0 +1,116 @@
...
+         else if (r < 900)
+            return ResultUnsigned::success(8);  
+         else if (r < 1100)
+            return ResultUnsigned::success(9);
+         return ResultUnsigned::failure(HoughError::Failure);  /// shouldn't be here, this won't be used
+      }
+      
+// default with two slices, one for negative and one for positive z, counting
+// some small overlaps, and -1 means "just take everything"
+      ResultBool inSliceDefault(double z, unsigned layer, int slice) {
```
asalzburger, line 65: Same here.

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp
```diff
@@ -0,0 +1,267 @@
...
+// @author Riley Xu then modified to ACTS by Jahred Adelman
+// @brief Implements track-seeding using a Hough transform.
+//
+// Using the Lorentz force equation, one can relate the phi of a track and the
+// coordinate of a single hit:
+//
+//      A * q / pT = sin(phi_track - phi_hit) / r
+//
+// where
+//      A   : 3 * 10^-4 GeV / (c*mm*e) for the ATLAS B field (to be configured)
```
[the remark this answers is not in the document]
asalzburger: One could extract this into a `Projector` type struct that is then e.g. specialised for ATLAS as a ConstFieldProjector(2_Tesla).

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/DefaultHoughFunctions.hpp
asalzburger: Please ignore, I saw that these are actually used in Delegates, so great thinking ...

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/DefaultHoughFunctions.hpp
asalzburger: Ignore this one as well.

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp
```diff
@@ -0,0 +1,279 @@
...
+#include <string>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+using ResultDouble = Acts::Result<double>;
+using ResultBool = Acts::Result<bool>;
+using ResultUnsigned = Acts::Result<unsigned>;
+
+using FieldCorrector = Acts::Delegate<ResultDouble(unsigned, double, double)>; // (unsigned region, double y, double r)
```
asalzburger, line 89: Nice, this is really great.

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp
```diff
@@ -0,0 +1,279 @@
...
+// The unsigned is a counter that will point to a spacepoint or to a measurement
+// object
+
+/// An houghHist is a 2d array of points, where each point has a value.
+/// The value starts as the number of hit layers, but can change with effects
+/// like a convolution. Also stored are indices of all hits that contributed to
+/// each bin. Size m_houghHistSize_y * m_houghHistSize_x. (NOTE y is row coordinate) For
+/// now, what is stored is actually the index of the object in the vectors, so
+/// we can get the Index layer
+typedef vector2D<std::pair<int, std::unordered_set<unsigned>>> HoughHist;
```
asalzburger, line 112: usually we use (as a convention)

```
using HoughHist = vector2D<...>;
```

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp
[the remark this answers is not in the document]
asalzburger: SpacePoints are built from measurement(s) and thus the source link should create this.

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp
```diff
@@ -0,0 +1,267 @@
...
+        0, 0, 0, 0};  // Hit lines will fill extra bins in x by this amount on
+                      // each side, size == nLayers
+
+    /// === Seeds for Hough ==
+    std::vector<int> m_threshold = {9};  // Minimum point value post-convolution
+                                         // to accept as a road (inclusive)
+    int m_localMaxWindowSize =
+        0;  // Only create candidates from a local maximum, requires traceHits
+    double kA = 0.0003;  // Assume B = 2T constant. Can apply corrections to
+                         // this with fieldCorrection function
```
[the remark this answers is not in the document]
asalzburger: That's practically the same as above, probably we should add somewhere some utility code that generates this as a constexpr when a B-Field is given.

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughVectors.hpp
```diff
@@ -0,0 +1,139 @@
+// This file is part of the Acts project.
+//
+// Copyright (C) 2021 CERN for the benefit of the Acts project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
```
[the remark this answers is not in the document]
asalzburger: Indeed, the grid class would give you the convenience to have Bound and Closed axes (e.g. for phi wrapping) ...

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,479 @@
...
+          std::vector<size_t> nHitsPerLayer(m_cfg.nLayers);
+          for (auto measurementIndex : m_houghHist(y, x).second) {
+             HoughMeasurementStruct *meas = houghMeasurementStructs[measurementIndex].get();
+             hitIndicesAll[meas->layer].push_back(meas->index);
+             nHitsPerLayer[meas->layer]++;
+          }
+
+          std::vector<std::vector<int>> combs = getComboIndices(nHitsPerLayer);
+
+          for (size_t icomb = 0; icomb < combs.size();
```
asalzburger, line 154: We have a utility class `Acts/Utilities/Enumerate.hpp` which allows you to do 

```c++
for (auto [icomb, comb] : enumerate(combs)){
 
std::cout << "Index " << comb << " has value " << comb << std::end;
}
```

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,479 @@
...
+  unsigned width = m_cfg.threshold.size() / 2;
+  if (x < width || (houghHist.size(1) - x) < width)
+    return false;
+  for (unsigned i = 0; i < m_cfg.threshold.size(); i++) {
+    if (houghHist(y, x - width + i).first < m_cfg.threshold[i])
+      return false;
+  }
+
+  // Pass local-maximum check, if used
+  if (m_cfg.localMaxWindowSize)
```
asalzburger, line 239: In our naming conventions we have decided to always use brackets, even if not necessary, so instead of writing 
```c++
for (cond1)
  for(cond2)
    do_something();
```

We would advertise to do 
```c++
for (cond1){
  for (cond2){
   doItAgain();  
}
}
```

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,479 @@
...
+      combos[icomb][isize] = static_cast<int>(index / nCombs_prior[isize]);
+      index = index % nCombs_prior[isize];
+    }
+  }
+
+  return combos;
+}
+
+
+void ActsExamples::HoughTransformSeeder::AddSPs(const AlgorithmContext& ctx) const {
```
asalzburger, line 401: Naming convention: `addSpacePoints(...)`

asalzburger: Then, one could actually make this callable with a spacepoint container, like this, this could be used for full event processing or batch-wise processing.

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+  // construct the combined input container of space point pointers from all
+  // configured input sources.
+  for (const auto& isp : m_cfg.inputSpacePoints) {
+    for (auto& sp : ctx.eventStore.get<SimSpacePointContainer>(isp)) {
```
asalzburger, line 406: Here's what we discussed regarding the spacepoints, we really should sort out a single `SpacePoint` representation present in `Core`

asalzburger: This will get rid of SimSpacePoints and then will not stop this from moving the code into core.

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
```diff
@@ -0,0 +1,479 @@
...
+       float phi = std::atan2(sp.y(), sp.x());
+       ResultUnsigned hitlayer = m_cfg.layerIDFinder(r).value();
+       if (!(hitlayer.ok())) continue;
+       auto meas = std::shared_ptr<HoughMeasurementStruct>(new HoughMeasurementStruct(hitlayer.value(), phi, r, z, sp.measurementIndex(), HoughHitType::SP));
+       houghMeasurementStructs.push_back(meas);
+    }
+  }
+}
+
+void ActsExamples::HoughTransformSeeder::AddMeasurements(const AlgorithmContext& ctx) const {
```
asalzburger, line 419: Naming convention `addMeasurements(...)`

Examples/Run/Reconstruction/Common/HoughExample.cpp
asalzburger: Agreed.

asalzburger (changes requested):
I have one major comment, I *think* it would be great to see if one could use the `Acts::Grid` directly, I am not sure if it makes sense to introduce a new grid structure - at least we should agree that the `Grid` is a target before moving the code to main.

Most other comments are suggestions and naming convention nitpicks.

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp
jahreda: Can you expand a bit more what you mean about the B field and using a Projector type struct, ie point me to something specific? I think I'm not sure what you're implying. I tried looking at the code and I see (in MultiTrajectory.hpp):
```
 using Projector =
      Eigen::Matrix<typename Covariance::Scalar, M, eBoundSize, ProjectorFlags>;
```
but I don't think that's what you mean?

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
jahreda: Yes, I guess the question is if we should wait for that or merge to Examples. I guess that depends on the timeline?

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
jahreda: I changed the name, do we prefer for this to take containers? We do need a single container type then :)

Examples/Algorithms/TrackFinding/src/HoughTransformSeeder.cpp
jahreda: 1) For the B field automatic generation, that sounds nice. Is it necessarily something that works in all cases? For more complex geometries with B field corrections is it clear that it's easy to get the constant value? (Maybe, I have not thought it through), and is there an example of where I can access the constant B field (tried looking and not obvious)
2) I tried asking some months ago about the grid class. I see the readthedocs instructions, but I haven't seen an obvious, simple explanation of how this should be used (I was pointed to SpacePointGrid.ipp but... well, it isn't so helpful to me, maybe that's my fault)

jahreda:
Thanks, Andi! I implemented the simple ones. I did also have some questions, especially (but not only) on the grid implementation, since it's not obvious to me how best to do this. A simple example would help!

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp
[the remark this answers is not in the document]
asalzburger: Can be done later.

asalzburger (approved):
Ok, let's move that in as is, there are a few cleanups but we can do this when moving to the Core repo.

jahreda:
Hi Andi - before figuring out those conflicts, I'm trying to understand if you want to follow up on anything else. I think from what I saw the use of grid is the only real open item? I'm still not sure how best to do that, but it also sounded like you agreed to move ahead as long as we put that on a to-do list?

tboldagh:
> Hi Andi - before figuring out those conflicts, I'm trying to understand if you want to follow up on anything else. I think from what I saw the use of grid is the only real open item? I'm still not sure how best to do that, but it also sounded like you agreed to move ahead as long as we put that on a to-do list?

Actually before moving HT to the core we can fix theses issues in smaller PRs  while still in examples.

asalzburger:
Yeah, indeed, the conflicts look rather minor, so if we could resolve them, I would put the HT in. Then, following what we are doing lately, we can shift the code over into Core using the `Experimental` namespace to indicate that this is not yet production code and might change.

jahreda:
@asalzburger I merged my code (required a bit of tinkering) and resolved all conversations so that things can be merged (even if there's more to do as we discussed). I think the remaining blocker is that Robert has requested changes - I guess you have the power to remove them?

robertlangenberg (approved):
Don’t want to stand in the way of progress :)

asalzburger:
Updating and merging when pipeline succeeds.

asalzburger:
clang-tidy is failing, so what I suggest I will do tomorrow is:
- merge this in, followed by a PR that fixes the clang-tidy complaints

andiwand:
I can also fix the issues if you like @asalzburger but I fear I have not permissions to push on this branch

asalzburger:
That's why I suggested to merge it & create a hotfix PR that goes in right after.

andiwand:
> That's why I suggested to merge it & create a hotfix PR that goes in right after.

sounds good. lets do that tomorrow. I will create a branch based on this one and we can merge one after the other so nobody will be interrupted by failing ci

andiwand:
clang tidy fixes ready here https://github.com/acts-project/acts/pull/1691

PR #1309, opened 2022-07-09
refactor: Move add* functions to python modules

### LINKED ISSUE #1242: Python bindings: ship modules containing the `addXYZ` functions
This would allow us to conveniently configure the full chain example in e.g. the ODD.

I guess one option would be to include them under the `acts` package, maybe something like `acts.examples.fragments`?

/cc @timadye

tboldagh:
The add* (i.e. addFatras, addPythia8) were moved to modules
in Examples/python/acts/examples.
Two modules were made: simulation hosting generation, simulation and digization
and reconstruciton.

They can be used as follows:
from acts.examples.simulation import addFatras

Example scripts are refactored to use the new location of add* functions.

resolves #1242

FI: @timadye  @paulgessinger


Examples/Scripts/Python/common.py
```diff
@@ -80,9 +80,9 @@ def addPythia8(
 
     NB. this version is included here only for compatibility. Please use pythia8.addPythia8 instead.
```
timadye, line 81: Maybe change the comment to refer to `acts.examples.sumulation`

timadye:
* Can you move `itk.buildITkGeometry` to a new `acts.examples.itk`
* Can you move `common.getOpenDataDetector`/`getOpenDataDetectorDirectory` to a new `acts.examples.odd`?
    * @paulgessinger, or does the `odd` stuff belong under `thirdparty/OpenDataDetector`?
* Those are only the other cases I can think of, where we import from scripts in `Examples/Scripts/Python`.
    * Can we check whether there are any others?
* Maybe we can retire `common.py` including the `common.addPythia8` compatibility wrapper (and ignore my previous comments about this).

tboldagh:
> I think the ODD python functions / helpers should live in this repository.

By _this_ you mean Scripts/Python/... or stay where they are now?

Examples/Scripts/Python/common.py
```diff
@@ -1,98 +1,40 @@
...
+#     nhard=1,
+#     npileup=200,
+#     beam0=acts.PdgParticle.eProton,
+#     beam1=acts.PdgParticle.eProton,
+#     cmsEnergy=14 * acts.UnitConstants.TeV,
+#     vertexStddev: acts.Vector4 = acts.Vector4(0, 0, 0, 0),
+#     vertexMean: acts.Vector4 = acts.Vector4(0, 0, 0, 0),
+# ):
+#     """This function steers the particle generation using Pythia8
+
```
paulgessinger, line 22: Do we need this file at all?

Examples/Scripts/Python/propagation.py
```diff
@@ -4,9 +4,7 @@
 import acts
 import acts.examples
 
-from acts.examples import GenericDetector, AlignedDetector
-
-from common import getOpenDataDetectorDirectory
+from acts.examples import GenericDetector
```
paulgessinger, line 8: Same here, as an example script I have some commented out alternatives further down in this file to encourage changing the file. If we want to keep that, I would think we should keep these (unused) imports.

tboldagh: Thanks, ATLAS habit, there we get flake complaining about unused imports, will brig them back.

tboldagh:
> I think I'm happy with this. I added a few more comments, and we should make the CI go green, then we can merge this from my side.

Indeed. I am a but puzzled about the issues themselves. One is genuine segfault (in linux_exampes_test) and other is about missing ODD description. Is the location correct actually?

paulgessinger:
The SEGFAULT is also surprising to me.
For the ODD path, you'd have to check if the path resolves correctly. The implementation assumes the location of the ODD folder relative to the python file, so you might have to update this when moving the functions to another file.

Examples/Scripts/Python/full_chain_itk.py
```diff
@@ -10,11 +10,20 @@
 field = acts.examples.MagneticFieldMapXyz(str(geo_dir / "bfield/ATLAS-BField-xyz.root"))
```
timadye, line 10: I'm actually commenting about the previous line (line 9), but I can't get GitHub to select it.
This should be:
```
detector, trackingGeometry, decorators = acts.examples.itk.buildITkGeometry(geo_dir)
```
and the import should replace `itk` with `acts.examples.itk`.

timadye: and line 51, the commented-out import:
```
# from acts.examples.reconstruction import SeedingAlgorithm, ParticleSmearingSigmas
```
(this really is quite annoying that GitHub doesn't let you make suggestions for additional changes)

timadye:
I noticed that `vertex_fitting.py` hasn't been moved. Should we do that one too?

timadye:
The discussion about `getOpenDataDetectorDirectory` in #1334 is probably more relevant here, so I'll make a related comment here.

I've been testing out the new slimmed-down scripts from `Examples/Scripts/Python`. The GenericDetector and ITk examples all seem to work fine when run from that directory (`vertex_fitting.py` segfaults for me, but I don't think that's related to this PR which doesn't yet touch that script).

Now we can copy a script to another directory to test there, and that mostly works fine, except for accessing the GenericDetector digi and seeding config files. Those are relative to the script directory, but for this case the files are all passed from the source directory, so it's easy for the user to fix up.

So if my recent suggestions are included, I'm happy with all these changes, once we get the ODD sorted.

Examples/Scripts/Python/seeding.py
```diff
@@ -1,13 +1,11 @@
-#!/usr/bin/env python3
```
timadye, line 1: I guess the usual `#!/usr/bin/env python3` was accidentally removed. Can we have it back?

tboldagh: My (or black's) fault. Sorry. Will restore in the next push (keep forgetting about the black)

Examples/Scripts/Python/common.py
tboldagh: That content is gone form this file.

Examples/Scripts/Python/full_chain_itk.py
tboldagh: addressing in the next commit

tboldagh: Fixed

tboldagh:
@timadye  @paulgessinger - I believe this PR is good to go. Hopefully commits squashing works!

timadye:
Looks good. I haven't tested the scripts again, but if they now work for you, then I'm sure they are good.

If this gets merged now, then the same move will need to be done in #1331 and #1334, rather than here, but that's just as good as making the change here.

I don't have permission to approve, so I leave that to @paulgessinger.

timadye (approved):

PR #1310, opened 2022-07-10
refactor: Remove globalToLocal log and return actual error message

osbornjd:
This is a very simple PR to change this logging message from `FATAL` level to `WARNING`. This is leftover from when the `globalToLocal` function did not return an `Acts::Result`, so if the transformation failed it was actually fatal. This result can now be passed and caught, thus it isn't necessary to flag it as `FATAL`.


paulgessinger:
Should the function not return the error in this case? The way I read the code now, it will just try to continue after the `WARNING`.

osbornjd:
> Should the function not return the error in this case? The way I read the code now, it will just try to continue after the `WARNING`.

Ha, good point... this really must be leftover then from when this piece of code just threw a `FATAL` log and then crashed. I'll change it to not log anything and instead return the error message similar to the track parameters.

paulgessinger:
Sounds good. Can you maybe adjust the PR title to reflect the updated nature of the changes?

osbornjd:
> Sounds good. Can you maybe adjust the PR title and descriptions to reflect the updated nature of the changes? Aside from that lgtm.

Sure, done.

paulgessinger (approved):

PR #1314, opened 2022-07-11
chore: Update CVMFS/LCG setup scripts

paulgessinger:
Removes LCG97 and 98 setup scripts in favor of 100 & 101.
Also updates the scripts to run on CentOS Stream 8


Corentin-Allaire (approved):
Looks good ! If I understand correctly this mean that we won't be testing acts with gcc9 any more right ?

paulgessinger:
I'm not dropping any versions from CI in this PR. In the CI we we're already on LCG100 and LCG101. The ubuntu 20.04 build still runs gcc9 I think, though.

andiwand (approved):

PR #1315, opened 2022-07-11
fix: Make python test failure with compile-time log threshold more informative

paulgessinger:
If `ACTS_LOG_FAILURE_THRESHOLD` is set, a compile time threshold is
enabled and `setFailureThreshold` throws an exception. The python tests
assume this call to `setFailureThreshold` always works. This PR prints an informative error message about what to do about this, but reraises the exception.


HadrienG2:
I feel uneasy about this PR sort-of swallowing the error. This can result in a test that looks like it's testing something (e.g. absence of warnings) and passes from a CI perspective (returns exit code 0), but actually doesn't test what it's intended to test, which I think is dangerous.

It would probably be fine with the test just failing (as it currently does) with a clearer message which states that Python tests are incompatible with compile-time log failure thresholds and indicates how the build configuration should be changed.

paulgessinger:
Fair enough @HadrienG2, @andiwand. I changed this PR to raise a new `RuntimeError` with a more informative error message. If you think this isn't providing any benefit over the exception message from C++, we can also close this PR. 

Let me know what you think!

andiwand (approved):
I think it makes sense as people may encounter this when stating with acts and can help themselves with such a descriptive message

PR #1316, opened 2022-07-11
refactor: Improve Python import helpers

andiwand:
after my recent changes hepmc3 test will only run if the environment is sourced. this refactors the import helpers in order to fix that


Examples/Python/tests/helpers/__init__.py
```diff
@@ -33,19 +33,19 @@
...
 
-edm4hepEnabled = "EDM4HEP_DIR" in os.environ
-if edm4hepEnabled:
-    try:
-        import acts.examples.edm4hep
-    except ImportError:
-        edm4hepEnabled = False
+    hepmc3Enabled = True
+except ImportError:
+    hepmc3Enabled = False
```
AJPfleger, line 41: Would it make sense to also change the above `geant4Enabled` (l.8-13) and `dd4hepEnabled` (l.29-34) to this cleaner version? (as also already done with `rootEnabled`)

paulgessinger: The underlying issue is that `edm4hep` and `hepmc3` weren't importing additional modules in the helper python files. Geant4, DD4hep and ROOT are not functional without the runtime environment variables set, so this makes sense to have differently in my opinion.

andiwand:
waiting for https://gitlab.cern.ch/acts/OpenDataDetector/-/merge_requests/34

andiwand:
good to go @paulgessinger ?

paulgessinger (approved):
LGTM! Let's merge if the CI runs through now without skipped tests.

PR #1317, opened 2022-07-11
feat: Propagator accepts prepared result instance

paulgessinger:
This PR adds overloads to the `Propagator::propagate` method which allow passing in a prepared result object. I want to use this to have the CKF reuse a single `MultiTrajectory` object for all input seeds, and @benjaminhuth might be able to use this to simplify the GSF internals.


benjaminhuth (approved):
Looks good to me!

The only thing I thought about is if it would be better to use r-value references and `std::forward` for the `input_result` since it would allow perfect forwarding... However, this has probably no measurable impact.

paulgessinger:
Hm, I didn't expect this to change any outputs.

paulgessinger:
I updated the changes hashes now, @benjaminhuth.

benjaminhuth:
hmm it seems that only the `tracksummary`-hashes have changed, not the `trackstates`, right?
This is more weird than before, since that means that the propagation/fitting itself does not change on the bit-level, but the summary does... Maybe this is worth investigating @paulgessinger?

paulgessinger:
I guess so, yes. We don't have performance plots running in CI. Do you have an implementation for diagnostics plots other than what's directly in `tracksummary_gsf.root`?

benjaminhuth:
Hmm that should be possible with the same tools used for the `KalmanFitter`, since the result type and the writers are identical...
@paulgessinger is there a way to download these `tracksummary_gsf.root` from the CI? Or should I build acts in both configurations?

paulgessinger:
The GSF currently doesn't run in the physics perf monitoring, so it's not stored in the physmon artifact unfortunately.

andiwand:
but we could just add another artifact in the github actions yaml right?

is this a pure CI problem and not reproducible locally?

Core/include/Acts/Propagator/Propagator.ipp
```diff
@@ -10,10 +10,9 @@
 
 template <typename S, typename N>
 template <typename result_t, typename propagator_state_t>
-auto Acts::Propagator<S, N>::propagate_impl(propagator_state_t& state) const
-    -> Result<result_t> {
-  result_t result;
-
+auto Acts::Propagator<S, N>::propagate_impl(propagator_state_t& state,
+                                            result_t& result) const
```
andiwand, line 14: the only different I can spot from the changes (no clue what the code does) is that `result` was a newly instantiated object before and now passed in by reference which could bring in a dirty state. not sure if that applies and makes a difference here

benjaminhuth:
> is this a pure CI problem and not reproducible locally?

No I don't think so, it would be just convenience and time saving not to rebuild ACTS in 2 different versions...

benjaminhuth:
Okay, locally I see also differences in both files (states and summary). I see also "high-level differences" between the runs, in the main-branch there are two `MultiStepperError:4` which do not happen in your feature branch...
I'll investigate this.

Core/include/Acts/Propagator/Propagator.ipp
```diff
@@ -172,6 +192,31 @@ auto Acts::Propagator<S, N>::propagate(
   // Type of track parameters produced at the end of the propagation
   using return_parameter_type = BoundTrackParameters;
 
+  // Type of the full propagation result, including output from actions
+  using ResultType =
+      action_list_t_result_t<return_parameter_type,
+                             typename propagator_options_t::action_list_type>;
+
+  return propagate(start, target, options, ResultType{});
```
benjaminhuth, line 200:
```suggestion
  return propagate<parameters_t, propagator_options_t, target_aborter_t, path_aborter_t>(start, target, options, ResultType{});
```

Core/include/Acts/Propagator/Propagator.ipp
```diff
@@ -93,17 +92,39 @@ auto Acts::Propagator<S, N>::propagate(
...
 
+  static_assert(std::is_copy_constructible<ReturnParameterType>::value,
+                "return track parameter type must be copy-constructible");
+
   // Type of the full propagation result, including output from actions
   using ResultType =
       action_list_t_result_t<ReturnParameterType,
                              typename propagator_options_t::action_list_type>;
 
+  return propagate(start, options, ResultType{});
```
benjaminhuth, line 106:
```suggestion
  return propagate<parameters_t, propagator_options_t, path_aborter_t>(start, options, ResultType{});
```

Core/include/Acts/Propagator/Propagator.ipp
benjaminhuth, line 200:
```suggestion
  return propagate<parameters_t, propagator_options_t, target_aborter_t, path_aborter_t>(start, target, options, ResultType{});
```

Core/include/Acts/Propagator/Propagator.hpp
```diff
@@ -394,6 +428,38 @@ class Propagator final {
...
 
+  /// @brief Propagate track parameters - User method
+  ///
+  /// This function performs the propagation of the track parameters according
+  /// to the internal implementation object until at least one abort condition
+  /// is fulfilled, the destination surface is hit or the maximum number of
+  /// steps/path length as given in the propagation options is reached.
+  ///
+  /// @tparam parameters_t Type of initial track parameters to propagate
+  /// @tparam surface_t Type of target surface
```
benjaminhuth, line 439: This documentation is slightly incorrect I think, the template parameters `target_aborter_t` and `path_aborter_t` are not documented, but therefore nonexisting template parameters.

benjaminhuth (changes requested):

Core/include/Acts/Propagator/Propagator.hpp
paulgessinger: Ok, this applies to all overloads i think.

paulgessinger:
All green I think. Can you approve if you're happy @benjaminhuth ?

benjaminhuth (approved):

PR #1319, opened 2022-07-11
fix: Update the Material Mapping ploting script to remove json dependency

Corentin-Allaire:
This PR removes a dependency on json that was present in the plotting script for the material mapping. This stopped working a will ago and broke completely when the recent change to the way json is integrated in Acts.


robertlangenberg (approved):
looks good

PR #1320, opened 2022-07-11
ci: Pull more config into physmon script, extend PG eta range to +-4

paulgessinger:
I think for the physmon jobs, given the ODD, it makes little sense to only test particle up to $|\eta| < \pm2$. This PR adjusts `physmon.py` to include the particle gun configuration so I can customize it to produce particles up to $|\eta| < \pm 4$.


paulgessinger:
I had a look at the physmon outputs. They're basically as expected.

# CKF Seeded 
grafik
Not entirely sure why this looks so asymmetric.

grafik
grafik
Bit of an odd feature in the tracking efficiency in $\eta$. Since this doesn't show up in the other "seeding" modes, it might be some feature of the seeding?

# CKF truth estimated
grafik

# CKF truth smeared
grafik

andiwand (approved):
plots look mostly reasonable IMO

the on with the holes looks a bit weird but that might be due to statistical fluctuations? the error bars are quite big. we could check with a different eta distribution, a different seed or more samples. but I don't think it is necessary

paulgessinger:
I updated the references. I realized now that it's actually pretty hard to put in the correct reference commit SHA, because we squash and merge. No idea if that can be improved.

andiwand:
> I updated the references. I realized now that it's actually pretty hard to put in the correct reference commit SHA, because we squash and merge. No idea if that can be improved.

yeah I think it is almost impossible. shouldn't be the git history be enough?

paulgessinger:
Probably yeah.

paulgessinger:
I think this is good now, @andiwand.

PR #1321, opened 2022-07-12
ci: License checker script reports incorrect diff

paulgessinger:
Caused by an outdated regex that cleans the year incorrectly.


Corentin-Allaire (approved):
Looks good, thanks for finding this ! How is it that it didn't affect us before ? Does this only appear the CI artefacts and not in the check itself ?

PR #1323, opened 2022-07-12
fix: Energy loss function

beomki-yeo:
This PR fixes the Bethe energy loss function in `Interactions.cpp`

`Acts::computeEnergyLossBethe` and `Acts::computeEnergyLossLandau` is implemented from equation 33.5 and 33.11 of [Review in Particle Physics](https://journals.aps.org/prd/abstract/10.1103/PhysRevD.98.030001)

And their first term is computed with `computeEpsilon` function, following the notation of equation 33.11:

```c++
/// Compute epsilon energy pre-factor for RPP2018 eq. 33.11.
///
/// Defined as
///
///     (K/2) * (Z/A)*rho * x * (q²/beta²)
///
/// where (Z/A)*rho is the electron density in the material and x is the
/// traversed length (thickness) of the material.
inline float computeEpsilon(float molarElectronDensity, float thickness,
                            const RelativisticQuantities& rq) {
  return 0.5f * K * molarElectronDensity * thickness * rq.q2OverBeta2;
}
```

Problem is that the epsilon term of equation 33.5 for Bethe function doesnt have the `0.5f` term

Eq. 33.5
image

Eq. 33.11
image

If we are going to use the same epsilon term, we need to multiply 2 in the next term of Bethe energy loss function

After the change, the output of `Acts::computeEnergyLossBethe` gets consistent with Figure 33.2 of the reference.

### Update July 18th

Looks like there is a _literature_ error in Landau energy loss function (eq. 33.11).
It used the mass term of the incident particle (`m`), but I guess it should be the mass of electron (`m_e`) as equation 33.5.
**I need a cross-check from someone who can access [Straggling in thin silicon detectors](https://journals.aps.org/rmp/abstract/10.1103/RevModPhys.60.663), which I cannot** (Update: It is confirmed that it should be the rest mass of electron)

Finally, to get a consistent result of energy loss function, we need to fix the `computeDeltaHalf` term as well.
But this is already mentioned in the comment which says: "Should we use RPP2018 eq. 33.7 instead w/ tabulated constants?"

```c++
/// Compute the density correction factor delta/2.
///
/// Uses RPP2018 eq. 33.6 which is only valid for high energies.
///
/// @todo Should we use RPP2018 eq. 33.7 instead w/ tabulated constants?
inline float computeDeltaHalf(float meanExitationPotential,
                              float molarElectronDensity,
                              const RelativisticQuantities& rq) {
```

Yeah it does make big difference between 33.6 (valid for very high energy particles) and 33.7.
**I am not going to include delatHalf term correction in this PR**, but it is included in acts-project/detray#282

After fixing codes (including the deltaHalf term), the Bethe energy loss function matches to the value in ([PDG](https://pdg.lbl.gov/2022/AtomicNuclearProperties/index.html))

In the following figure, Bethe energy loss (dE/dx) in Silicon is compared for Acts main, this PR, acts-project/detray#282, and PDG value. The value from this PR is not accurate yet due to the incorrect deltaHalf term

For the most probable energy loss from Landau equation under the same condition of Figure 33.7 of [Review in Particle Physics](https://journals.aps.org/prd/abstract/10.1103/PhysRevD.98.030001)
, detray#282 got 0.526 MeV, which is very close to the value in the figure.
This PR got 0.739 MeV due to the incorrect delatHalf term


paulgessinger:
Looking at eq. 33.5 and 33.11 this seems indeed like an inconsistency to me. @asalzburger can you have a look at this?

Of course this changes the output of everything we run that runs the fast sim.

asalzburger:
Hi @beomki-yeo - great catch, I assigned that to me & will cross-check the details.

beomki-yeo:
OK. BTW I still see the inconsistency in energy loss function (from both Bethe and Landau) compared to the value in the figures of the same reference. I thought the discrepancies are from the approximated mean excitation energy, but it seems not the main reason apparently (the approximation is not that bad). I will provide the details later when I have time but do you guys have any idea what the reason could be

Core/src/Material/Interactions.cpp
```diff
@@ -163,7 +163,7 @@ float Acts::computeEnergyLossBethe(const MaterialSlab& slab, int /* unused */,
   // the required modification only change the prefactor which becomes
   // identical to the prefactor epsilon for the most probable value.
   const auto running =
-      0.5f * std::log(u / I) + 0.5f * std::log(wmax / I) - rq.beta2 - dhalf;
+      std::log(u / I) + std::log(wmax / I) - 2.0f * rq.beta2 - 2.0f * dhalf;
```
andiwand, line 166: looking at `Eq. 33.5` - are we missing a square for `I`? or is this already absorbed in `const auto I = slab.material().meanExcitationEnergy();`?

beomki-yeo: squres is not missing
`std::log(u/I) + std::log(wmax / I) = std::log(u * wmax/ I^2)` is eqaul to the first term in the paranthese

paulgessinger:
Re your update: this is indeed very interesting. If we're now sort of mitigating literature errors, I guess we should thoroughly document the actual equations being used. Do you think it makes sense to have this as a page on the documentation, which references the source publications, but also lists explicitly the combined equations being used now? (in a future PR)

andiwand:
If we want to track the documentation in this PR we should at least link it in the code since git history is a bit loose after a following change IMO

but it would not hurt to have an MD documenting the formulas used in the code and referencing where they come from. the code can then point to this doc

paulgessinger:
I agree with @andiwand. What I was suggesting was to have a document which explicitly writes out the equations as they are implemented after this PR, rather than having something like

```cpp
// like eq 33.5 from Ref 1 but with epsilon from eq 35.22 and adjusted m to account for ...
```

because I think that will make it much harder to get up to speed on what's happening.

andiwand:
this seems to have some impact on https://github.com/acts-project/acts/issues/1385#issuecomment-1217968896

do you think we can get this in in the near future? @beomki-yeo @paulgessinger 

it looks like some reference tests are failing because the output changed. but otherwise the fix is complete?

beomki-yeo:
The fix is not complete yet. I still need to modify the density correction effect.

beomki-yeo:
@asalzburger 

Here are the key changes:
- Multiply x2 to `computeEnergyLossBethe` explained above
- Add `densityEffectData` member variable to `Material`: Unless it is specified in the constructor (e.g. `fromMassDensity` or `fromMolarDensity`, it will be the default value which doesn't make any changes. I actually put the real data in `makeSilicon()` and `makeBeryllium()` defined in `PredefinedMaterials.hpp`

After the changes, Integration test of `PropagationDenseConstant` failed in covariance check due to the low tolerance value (0.05) so I had to increase it to 0.07. I am not also sure if the updates in the PR are meant to increase the covariance during the propagation :p  

Let me also ping @niermann999 as she might be interested in synchronizing detray and Acts material.

beomki-yeo:
Hmm truth tracking test fails :/...

asalzburger:
I saw - let's discuss this tomorrow, I have an idea how to validate this.

andiwand:
> Hmm truth tracking test fails :/... @paulgessinger Is there any possibility that physmon uses data simulated from the reference commit?

simulation is done on the fly but we compare to reference. so I guess a failure is expected if we fix something in the energy loss?

beomki-yeo:
> simulation is done on the fly but we compare to reference. 

"compare to reference" means "compare to the data simulated from reference commit"? If that is the case, the shift in qoverp pull value distribution can be explained. If not, I made a stupid mistake in the PR

andiwand:
yes - the reference is just an output generated by a previous run of physmon with a commit from the main branch. so if you change the physics this will reflected in your physmon run and will fail the comparison

if we are happy with the new results you can copy the new output files from the CI and put the into the reference folder before merging. that way we make sure nobody accidentally breaks things again

paulgessinger:
Should we discuss what to do with this PR at some point?

beomki-yeo:
Yes for sure, would you like to meet in person later today? - the qop pull distribution is distored with the change and I have no clue on it right now

beomki-yeo:
Hopefully CI will be OK after this run :pray:

paulgessinger:
Discussed in the meeting that this might have overlap with what was seen in: https://github.com/acts-project/acts/pull/1631

beomki-yeo:
Actually I was surprised by #1631. If I understood correctly, the simulation is not using the interactions but reconstruction is using them?

benjaminhuth:
> Actually I was surprised by #1631. If I understood correctly, the simulation is not using the interactions but reconstruction is using them in physmon?

I think exactely thats the case at the moment. I was surprised as well when I saw it.

beomki-yeo:
There have been a couple of updates...

1. The material interactions are turned on in `simulation.py`. This makes the pull value distributions better indeed. For example,  the current pull value distribution of qop is quite narrow where the sigma is much smaller than 1.

2. After turning on the interactions, the propagation failed sometimes because the momentum of a particle drops to zero. Technically qop is not calculated properly in  `update` function `EigenStepper.ipp`. I set the qop to the infinity in case of the zero momentum.

Meanwhile I also see many of physmon plots failing the test with the changes
You can see the physmon plots here (I don't know how to retrieve the plots generated from the CI directly) : https://drive.google.com/file/d/1T0EeINoVEvk5WEyzTVs-7S--l6gE6e-D/view?usp=share_link

Tests/IntegrationTests/Autodiff/PropagationAutodiffDenseConstant.cpp
```diff
@@ -45,7 +45,7 @@ constexpr auto epsPos = 10_um;
 constexpr auto epsDir = 1_mrad;
 constexpr auto epsMom = 5_MeV;
 // relative covariance tolerance
-constexpr auto epsCov = 0.05;
+constexpr auto epsCov = 0.07;
```
beomki-yeo, line 48: I increases the tolerance because test failed with (4,1) element. No clue tbh  ¯\_(ツ)_/¯ 
```
/mnt/nvme0n1/byeo/projects/acts/acts/Tests/IntegrationTests/PropagationTests.hpp(149): error: in "PropagationDenseConstant/CovarianceToStrawAlongZ/_971": check Acts::Test::checkCloseCovariance((cmp.covariance().value()), (ref.covariance().value()), (relativeTolerance)) has failed. The difference between the covariance matrix term -0.0172329 and its reference 0.00901893, at index (4, 1), is not within tolerance 0.0210812. The covariance matrix being tested was
     1.62512     0.455501    0.0253009  -0.00188958   -0.0546908     -412.164
     1.27222      14.1597    0.0172062   -0.0981616    -0.228108     -43.9163
   0.0243024  -0.00634497  0.000418648  6.33088e-05  -0.00129378     -11.8281
 -0.00764448   -0.0986385 -0.000102058  0.000685389   0.00157456    0.0680835
    -0.05116   -0.0172329  -0.00144978  9.07089e-05     0.013057      106.168
    -412.784     -44.6307     -11.8451    0.0727551      106.325  5.61715e+07
and the reference covariance matrix was
     1.62462     0.454629    0.0252862  -0.00188958   -0.0532582     -411.651
     1.27033      14.1588     0.017152   -0.0981606    -0.201052     -43.6726
   0.0242892  -0.00635785  0.000418194  6.33053e-05  -0.00129365     -11.8075
 -0.00763733   -0.0986342 -0.000101775  0.000685389   0.00138661    0.0680857
   -0.048235   0.00901893  -0.00140563 -9.07072e-05    0.0125552      105.683
    -412.271     -44.3859     -11.8245    0.0727532      105.838  5.61715e+07
```

beomki-yeo:
@andiwand @paulgessinger @asalzburger @niermann999  This PR is ready to be reviewed.

Let me summarize the changes:
1. Fix the Bethe energy loss function by multiplying 2  
2. Fix the Landau energy loss by replacing the incident particle mass with the electron mass
3. Add the density effect data into the material
4. Add unit tests which compare the PDG and ACTS energy loss

andiwand (approved):
looks good to me. I have to minor comments. otherwise it is good to go from my side

Core/include/Acts/Material/Interactions.hpp
```diff
@@ -13,6 +13,29 @@
...
+    const auto mOverP = mass * std::abs(qOverP / q);
+    const auto pOverM = 1.0f / mOverP;
+    // beta² = p²/E² = p²/(m² + p²) = 1/(1 + (m/p)²)
+    beta2 = 1.0f / (1.0f + mOverP * mOverP);
+    // beta*gamma = (p/sqrt(m² + p²))*(sqrt(m² + p²)/m) = p/m
+    betaGamma = pOverM;
+    // gamma = sqrt(m² + p²)/m = sqrt(1 + (p/m)²)
+    gamma = std::sqrt(1.0f + pOverM * pOverM);
+  }
+};
```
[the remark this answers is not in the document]
beomki-yeo: Yeah We can consider it in the future -  it is only used in `Interactions.hpp` and `Interactions.cpp` atm

andiwand (approved):

andiwand (approved):

andiwand:
good to go from my side. any last comments before I merge @asalzburger @paulgessinger

andiwand:
> Please don't tell me Athena has been using incident particle mass in Landau energy loss: https://gitlab.cern.ch/atlas/athena/-/blob/master/Tracking/TrkExtrapolation/TrkExTools/src/EnergyLossUpdator.cxx#L348
> 
> Ha ha

lets hope not :thinking: it would be pions in that case right?

beomki-yeo:
I think it was not electron;

https://gitlab.cern.ch/atlas/athena/-/blob/master/Tracking/TrkExtrapolation/TrkExTools/src/EnergyLossUpdator.cxx#L52

Yeah Pions or muons I guess

andiwand:
the `iPot * iPot` does also not align with the equation you posted no?

```cpp
      // calculate the most probable value of the Landau distribution
      double mpv = m_mpvScale * xi / (beta * beta) *
                   (std::log(m * eta2 * kaz / (iPot * iPot)) + 0.2 -
                    beta * beta - delta); //
                                          // 12.325);
```

beomki-yeo:
TBH, I didn't look into other terms seriously; I will crosscheck

beomki-yeo:
Nah the equation I posted is in the form of `log(A/I) + log(B/I) = log(AB/I^2)` so I think it is OK

beomki-yeo:
Just a remark on the pure impact of Landau function mass input

monitored(incident particle: muon I guess)
reference(electron)

andiwand:
> Nah the equation I posted is in the form of `log(A/I) + log(B/I) = log(AB/I^2)` so I think it is OK

haha I fell for the same trick twice - thanks :smile: 

lets bring this PR one more time into the round table on tuesday and then merge it

beomki-yeo:
Of course

Core/src/Material/Interactions.cpp
```diff
@@ -97,36 +75,63 @@ inline float logDeriveWMax(float mass, float qOverP,
...
+  const auto density = slab.material().densityEffectData();
+  /// Uses RPP2018 eq. 33.6 which is only valid for high energies.
+  if (density == Acts::detail::DensityEffectData{}) {
+    // only relevant for very high ernergies; use arbitrary cutoff
+    if (rq.betaGamma < 10.0f) {
+      return 0.0f;
+    }
+    // pre-factor according to RPP2019 table 33.1
+    const auto plasmaEnergy =
+        PlasmaEnergyScale * std::sqrt(1000.f * molarElectronDensity);
```
beomki-yeo, line 103: Wow.. The original Equation for density effect correction was NOT bad! It was just missing 1000 factor in calculating `plasmaEnergy`. It can reproduce the PDG value very well.

If people don't disagree I will delete density effect data in Material.

asalzburger: But don't we have to store the density of the material anyways?

```
  parameters[eMolarDensity] = m_molarRho;
```

beomki-yeo:
I also need to update the CI reference and root hash tag - I will do that promptly

Core/src/Material/Interactions.cpp
beomki-yeo: > I will delete density effect data in Material.

You are not seeing that anymore in the PR because I already deleted it.

beomki-yeo: There was sth like this in the PR but I removed it today:
https://github.com/acts-project/detray/blob/main/core/include/detray/materials/detail/density_effect_data.hpp

beomki-yeo:
It is good to go now

andiwand (approved):

PR #1325, opened 2022-07-14
fix: Fix track-picking threshold in TrackFittingAlgorithm

benjaminhuth:
The current `TrackFittingAlgorithm` does not allow to pick track with index 0, this fixes the threshold.


Examples/Algorithms/TrackFitting/src/TrackFittingAlgorithm.cpp
```diff
@@ -86,7 +86,7 @@ ActsExamples::ProcessCode ActsExamples::TrackFittingAlgorithm::execute(
   std::vector<const Acts::Surface*> surfSequence;
   for (std::size_t itrack = 0; itrack < protoTracks.size(); ++itrack) {
     // Check if you are not in picking mode
-    if (m_cfg.pickTrack > 0 and m_cfg.pickTrack != static_cast<int>(itrack)) {
+    if (m_cfg.pickTrack > -1 and m_cfg.pickTrack != static_cast<int>(itrack)) {
```
andiwand, line 89: personally I prefer `!= -1` in such a case becuase `-1` encodes "do not pick" but with `>` every negative int will do the same

andiwand (approved):
feel free to ignore my comment as I don't know what the default strategy is in acts to handle special cases

Examples/Algorithms/TrackFitting/src/TrackFittingAlgorithm.cpp
AJPfleger: As far as I have seen, we have not yet had a similar special case handling in Acts.
I think excluding all negative ints makes sense here.

PR #1326, opened 2022-07-14
fix: propagation bug in GSF

benjaminhuth:
This fixes a bug that could lead to a infinite-loop in the propagation (and thus to the "reached step limit" error). The problem was that this did reset the `currentSurface` to `nullptr` regardeless if it hat a valid value before.


paulgessinger (approved):

Core/include/Acts/Propagator/MultiStepperAborters.hpp
```diff
@@ -41,6 +41,8 @@ struct MultiStepperSurfaceReached {
   bool operator()(propagator_state_t& state, const stepper_t& stepper,
                   const Surface& targetSurface) const {
     bool reached = true;
+    const auto oldCurrentSurface = state.navigation.currentSurface;
+
     for (auto cmp : stepper.componentIterable(state.stepping)) {
       auto singleState = cmp.singleState(state);
       const auto& singleStepper = cmp.singleStepper(stepper);
```
andiwand, line 48: unrelated to your changes but is it important that we keep looking after the first `reached = false;`? if not we can just reset and return there and remove `reached`

andiwand (approved):

Core/include/Acts/Propagator/MultiStepperAborters.hpp
benjaminhuth: Hmm unfortunately not, we must call the `SurfaceReached`-aborter on every component, since it sets the stepSize accordingly if we havn't reached the target yet, and this is necessary to ensure they can still reach the target.

PR #1327, opened 2022-07-15
feat: easy surface debug print

benjaminhuth:
During debugging the GSF propagation I noticed that it is rather verbose to print the full surface information (you need to create a `std::stringstream` object, invoke the `surface.toStream(...)` method and then print the stream...).

Therefore I implemented a more easy-to-use `operator<<` which can be used like this:

```c++
std::cout << std::tie(surface, geometryContext) << "\n";
```

This also simplifies some logging statements.


paulgessinger:
Ha! Very nice solution! This is why we got rid of the operator overloads and moved to `toStream`. We didn't think of using a tuple for this! Thanks a lot!

andiwand (approved):
nice :+1:

PR #1334, opened 2022-07-21
test: geant4 python example

andiwand:
similar to [fatras.py](https://github.com/acts-project/acts/blob/main/Examples/Scripts/Python/fatras.py) I tried to add a Geant4 utility function to add it to the chain instead of fatras


andiwand:
any thoughts on this @paulgessinger @timadye ?

paulgessinger:
This is a very good idea, thanks!

timadye:
Looks good. Just a couple of suggestions:
* Can this be coordinated with #1309? Probably simplest is to make a similar move as done there, moving `addGeant4` to the new `Examples/Python/python/acts/examples/simulation.py`, but keep `runGeant4` where it is.
* Comparing with `addFatras`, I notice that `addGeant4` doesn't have any output writers. I guess they would work the same as for Fatras (since both output to `particles_final` etc), so the same code could be copied. That's potentially a lot of duplicated code, so maybe could be done with an internal routine (`addFratrasWriters`?) called by both `addGeant4` and `addFatras` in `acts.examples.simulation`.
What do you think?

andiwand:
those are very good suggestions @timadye thank you! I am going to implement this next week

andiwand:
I implemented your suggestions @timadye . do you know when https://github.com/acts-project/acts/pull/1309 will be ready? I guess we can merge #1309 or #1334 first either way and deal with the conflicts in the remaining PR. happy to do it in #1309 too

Examples/Scripts/Python/geant4.py
```diff
@@ -0,0 +1,121 @@
...
+    alg = Geant4Simulation(
+        level=s.config.logLevel,
+        config=g4conf,
+    )
+
+    # Sequencer
+    s.addAlgorithm(alg)
+
+    # Output
+    addFatrasWriters(s, None, None)
```
timadye, line 81: Shouldn't this pass the output dirs?

andiwand: uh nice catch. autocomplete got me here

timadye:
> do you know when #1309 will be ready? I guess we can merge #1309 or #1334 first either way and deal with the conflicts in the remaining PR. happy to do it in #1309 too

Hopefully @tboldagh and @paulgessinger can comment on the best way/order to do this.

timadye (approved):

tboldagh:
> > do you know when #1309 will be ready?
I actively look into is now. Struggling a bit with ACTS complex build system though. But earlier today I was able to reproduce the issue seen online. So far so good.

tboldagh:
Need an advice one one thing here. The `getOpenDataDetecotrDirectory` discovery mechanism for the ODD files path relied on the fact the the `__file__` was pointing in fact to the sources directory. However, now when above function is installed the `__file__` points to the `build` dir. I though that the way to go is to use `DD4hep_DIR ` env variable. But I am struggling with it actually.

paulgessinger:
That's a good point. I think these functions can't really work in the installed case, since they assume they know where ODD is located. 
I don't think we can assume this in the installed case. 

Do we need to support this autoconfig ODD mode if ACTS is installed?

andiwand:
in this case we might want to install the ODD as well as part of Acts. but we might overshoot the original goal

timadye:
Alternatively, could we treat the ODD like we treat the ITk and let the user specify the base directory? For the ITk, we need to do that because it's in a separate repo, but we could use a similar mechanism for the ODD.

paulgessinger:
We use the ODD all over the place in the python based tests and it's our performance baseline. I think it's reasonable to have ODD work "out-of-the-box" in the most common scenario.

tboldagh:
Maybe `odd.py` could only contain the `getOpenDataDetector` whereas the `getOpenDataDetectorDirectory` could go back to the sources dir and thus allow the discovery. Archeologists will wonder though.

timadye:
> Maybe `odd.py` could only contain the `getOpenDataDetector` whereas the `getOpenDataDetectorDirectory` could go back to the sources dir and thus allow the discovery. Archeologists will wonder though.

How do we refer to the source directory in the import? If we kept the current layout, it would mean ODD scripts would only work if run from the source directory, which was one of the things this change was supposed to fix.

tboldagh:
> > Maybe `odd.py` could only contain the `getOpenDataDetector` whereas the `getOpenDataDetectorDirectory` could go back to the sources dir and thus allow the discovery. Archeologists will wonder though.
> 
> How do we refer to the source directory in the import? If we kept the current layout, it would mean ODD scripts would only work if run from the source directory, which was one of the things this change was supposed to fix.

I would refactor it as follows:
The `getOpenDataDetector` would get the path explicitly in the argument. All current uses, (in `tests` would need to import explicitly the getOpenDataDetectorDirectory from where they imported it before.  That should work. I probably miss something.

timadye:
> We use the ODD all over the place in the python based tests and it's our performance baseline. I think it's reasonable to have ODD work "out-of-the-box" in the most common scenario.

Yes, emphatically. But maybe we could move the OddDir out to the top-level and have those all pick the default.

How about if `getOpenDataDetectorDirectory` uses the caller's location (either looking up the stack, or passing `__FILE__` as an argument)? The user scripts could also allow the user to specify it on the command-line, in case they want to run it elsewhere.

paulgessinger:
For the latter use case, honestly, I'd prefer the user to just use `getOpenDataDetector` directly rather than rely on the directory determined automatically.
I'm also a bit worried that using the CWD is a bit prone to errors, and remember that it's also one reason for us to move the importable `add*` functions into modules.

timadye:
> I would refactor it as follows: The `getOpenDataDetector` would get the path explicitly in the argument. All current uses, (in `tests` would need to import explicitly the getOpenDataDetectorDirectory from where they imported it before. That should work. I probably miss something.

I think this would work. My proposal to use the caller's location by default would allow you to keep `getOpenDataDetectorDirectory` with `getOpenDataDetector` in `odd.py`.

timadye:
@paulgessinger, that sounds reasonable.

Just to note, that I don't think we are proposing to base anything on the CWD. I think @tboldagh was proposing to have it based on `getOpenDataDetectorDirectory`'s location. I suggested it could be on `getOpenDataDetectorDirectory`'s caller's location.

andiwand:
updated after https://github.com/acts-project/acts/pull/1309

not tested yet

paulgessinger:
Should we actually run this from pytest @andiwand? Might have to go through `subprocess` again because G4.

Examples/Python/python/acts/examples/simulation.py
```diff
@@ -394,6 +394,16 @@ def addFatras(
     s.addAlgorithm(alg)
 
     # Output
+    addFatrasWriters(s, outputDirCsv, outputDirRoot)
+
+    return s
+
+
+def addFatrasWriters(
```
paulgessinger, line 402: Pedantic, but 
```suggestion
def addSimWriters(
```
?

timadye: Sorry, that's my fault for suggesting that name. `addSimWriters` is much better.

andiwand:
> Should we actually run this from pytest @andiwand? Might have to go through `subprocess` again because G4.

good idea will do

andiwand:
this should be ready now

Examples/Python/python/acts/examples/simulation.py
```diff
@@ -394,6 +394,17 @@ def addFatras(
...
 
     # Output
+    addSimWriters(s, alg.config.outputSimHits, outputDirCsv, outputDirRoot)
+
+    return s
+
+
+def addSimWriters(
+    s: acts.examples.Sequencer,
+    inputSimHits: str = None,
```
paulgessinger, line 404: Technically 
```suggestion
    inputSimHits: Optional[str] = None,
```

Examples/Python/tests/test_examples.py
```diff
@@ -135,6 +135,54 @@ def test_fatras(trk_geo, tmp_path, field, assert_root_hash):
         assert_root_hash(f, rfp)
 
 
+@pytest.mark.slow
+@pytest.mark.skipif(not geant4Enabled, reason="Geant4 not set up")
+@pytest.mark.skipif(not dd4hepEnabled, reason="DD4hep not set up")
+def test_geant4(tmp_path, field, assert_root_hash):
+    # This test literally only ensures that the full chain example can run without erroring out
```
paulgessinger, line 142:
```suggestion
    # This test literally only ensures that the geant 4 example can run without erroring out
```

Examples/Python/tests/test_examples.py
```diff
@@ -135,6 +135,54 @@ def test_fatras(trk_geo, tmp_path, field, assert_root_hash):
...
+        stderr=subprocess.STDOUT,
+    )
+
+    assert_csv_output(csv, "particles_final")
+    assert_csv_output(csv, "particles_initial")
+    assert_csv_output(csv, "hits")
+    for f in root_files:
+        rfp = tmp_path / f
+        assert rfp.exists()
+        assert rfp.stat().st_size > 2**10 * 10
```
paulgessinger, line 183: Could we add has checks on these? I *think* the G4 output should be deterministic here.

paulgessinger (commented):
Sorry for more comments. 😅

Examples/Python/tests/test_examples.py
andiwand: I tried to use the fatras checks but it failed at the `hits`. the CI produced different hits than I had locally :disappointed:

paulgessinger: That's ok though. We have this in other cases too, we just need to pull the reference from the CI in that case.

andiwand: you mean using `assert_root_hash` right?
in that case the unit tests will fail locally right? happy to add it if that is expected

paulgessinger: The hash checks are off by default, and there's a mechanism to supply a local hash file with references. See the [docs page](https://acts.readthedocs.io/en/latest/examples/python_bindings.html#running-the-hash-checks-locally-and-how-to-update-the-reference-hashes)

andiwand: ah nice will do. the other checks from fatras I had to remove because it produced different results locally and in the CI

andiwand:
I think I addressed all your comments @paulgessinger

paulgessinger (approved):

PR #1257, opened 2022-05-17
refactor!: Build DD4hep geometry from hierarchy, use VariantParameters

paulgessinger:
This PR changes the DD4hep integration model. Instead of the DD4hep factory code having to depend on ACTS or the ActsDD4hep glue lib, they now use a mechanism we merged into DD4hep (as of `1.21`): [`MapStringVariantStruct`](https://dd4hep.web.cern.ch/dd4hep/reference/structdd4hep_1_1rec_1_1MapStringVariantStruct.html), which is wrapped into a detector element *extension* as `VariantParameters`.

This extension is independent of ACTS and encodes key value pairs, with the keys being strings and the values being any of string, int, float or bool. ACTS will use information attached to detector elements via this mechanism to identify layers, surface axes and material properties (layer and boundary).

This PR also removes the need to active subdetectors to be marked up using a specific extension: using the DD4hep `DetType` enum, ACTS will look for detector elements flagged `TRACKER` and `BARREL`, `ENDCAP` or `BEAMPIPE` and handle them accordingly. Passive structures that cannot be flagged with the above use a *variant* parameter `passive_layer` to tell ACTS to translate them.

## Possible parameters ACTS will interpret

ACTS geometry translation uses parameters attached to DD4hep detector elements via the `VariantParameters` extension. You can set this extension from your factory code, or use DD4hep's plugin mechanism to set parameters on detector elements via name. Here's an excerpt from the Open Data Detector on how to use this to mark a passive structure as a *passive layer*:

```xml
<lccdd>
  <detectors>
    <detector id="ODD_Solenoid_ID" name="Solenoid" type="ODDCylinder" beampipe="false" vis="Aluminum">
      <type_flags type="DetType_TRACKER" />
      <boundary_material surface="inner" binning="binPhi,binZ" bins0="mat_sol_bPhi" bins1="mat_sol_bZ"/>
      <tubs name="Solenoid" rmin="sol_rmin" rmax="sol_rmax" dz="sol_hlength" material="Aluminum">
        <layer_material surface="representing" binning="binPhi,binZ" bins0="mat_sol_bPhi" bins1="mat_sol_bZ"/>
      </tubs>
    </detector>
  </detectors>

  <plugins>
    <plugin name="DD4hep_ParametersPlugin">
      <argument value="Solenoid"/>
      <argument value="passive_layer: bool = true"/>
    </plugin>	
  </plugins>
</lccdd>
```

* Layer

[...]

PR #1337, opened 2022-07-23
refactor: remove old DD4hep example detectors

paulgessinger:
In preparation for #1257, this PR removes the old DD4hep detector examples that we don't plan to maintain going forward.


andiwand (approved):
:broom:

PR #1338, opened 2022-07-25
docs: General update of some docs pages, mark some pages as outdated

paulgessinger:
This PR updates a number of docs pages. Some are moved to markdown from rST (possible now due to MyST). Some pages get banners saying they're outdated (we hope to have our docathon soon to update them).

I also add a script that parses `CMakeLists.txt` for options and generates a markdown table for our getting started file.


docs/core/utilities/logging.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+Then you can run your executable, which uses Acts tools and algorithms, in
+the following way (tested under Unix)
+
+```console
+$ LD_PRELOAD=<YOUR_SHARED_LIBRARY> path/to/your/exectuable
+```
+
+For an example have a look at CustomDefaultLogger.cpp which you can use as
```
AJPfleger, line 121: I think \`CustomDefaultLogger.cpp\` is missing the grave accents.

AJPfleger (changes requested):

docs/core/utilities/logging.md
paulgessinger:
```suggestion
For an example have a look at `CustomDefaultLogger.cpp` which you can use as
```

AJPfleger (approved):
lgtm

PR #1339, opened 2022-07-25
fix: Multi-Stepper stepping error in multiple components + refactoring

benjaminhuth:
This PR does refactor the `MultiEigenStepperLoop` and fixes a situation, where a few stepping-erros could break the whole propagation.
- **Fix:** If some components had stepping errors and all other components where on a surface, the propagation terminated
- **Refactor:**
  - remove internal logger from `MultiEigenStepperLoop` since it is not really configurable from outside
  - improve `MultiEigenStepperLoop::step` (only construct summary-string in `VERBOSE` mode or on error, remove unnecessary loops and vectors, move things to lambdas, ...)

**NOTE:** This does not fix the reason why stepping errors occur in some situations, this (hopefully) will be done in a subsequent PR)


Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
```diff
@@ -35,16 +35,14 @@ auto MultiEigenStepperLoop<E, R, A>::boundState(
...
-                 << failedBoundTransforms << " of " << numberComponents(state)
-                 << " transforms failed");
-    }
-
     if (states.size() == 0) {
       return MultiStepperError::AllComponentsConversionToBoundFailed;
     }
 
+    if (failedBoundTransforms > 0) {
+      return MultiStepperError::SomeComponentsConversionToBoundFailed;
```
paulgessinger, line 43: So this doesn't log ERROR anymore, but newly returns an error, right?
Should this be an additional INFO or lower log message still to give more info if needed for debugging?

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
```diff
@@ -100,8 +98,29 @@ template <typename E, typename R, typename A>
...
+  const auto& logger = state.options.logger;
   State& stepping = state.stepping;
 
+  // It is not possible to remove components from the vector, since the
+  // GSF actor relies on the fact that the ordering and number of
+  // components does not change
+  auto invalidateComponent = [](auto& cmp) {
+    cmp.status = Intersection3D::Status::missed;
+    cmp.weight = 0.0;
+    cmp.state.pars.template segment<3>(eFreeDir0) = Vector3::Zero();
```
paulgessinger, line 110: This could potentially be made explicit with `std::optional` I think.

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
```diff
@@ -146,15 +154,17 @@ Result<double> MultiEigenStepperLoop<E, R, A>::step(
 
   // Loop over all components and collect results in vector, write some
   // summary information to a stringstream
-  SmallVector<Result<double>> results;
-  std::stringstream ss;
+  SmallVector<std::pair<Result<double>, bool>> results;
```
paulgessinger, line 157: Another potential case for `std::optional`?

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
benjaminhuth: So the problem is that there is no access to a logger anymore (since I removed the internal logger of the `MultiEigenStepper` ... I'm not sure if it would be better to change the general stepper interface to pass a `LoggerWrapper` instance to this function @paulgessinger ?

paulgessinger: I see. The main point of `LoggerWrapper` was that it can be easily defaulted, which then doesn't log anything. In any case, if we change this, it should be a separate PR I would say.

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
paulgessinger: Up to you.

paulgessinger (approved):
ok let's go.

PR #1341, opened 2022-07-25
feat: improve Python named-type argument checks

timadye:
* `@NamedTypeArgs` decorator checks we don't have named-type `kwargs` following positional `args`.
* Improve exception message if we have repeated `kwargs`


paulgessinger (approved):
lgtm

PR #1342, opened 2022-07-26
chore: clang-tidy modernize-use-using

paulgessinger:
Reduce clang-tidy warnings


Fatras/src/Utilities/LandauDistribution.cpp
```diff
@@ -179,10 +179,12 @@ double ActsFatras::LandauDistribution::quantile(double z) {
...
       51.005773, 53.437996, 56.123356, 59.103894};
 
-  if (z <= 0)
+  if (z <= 0) {
     return -std::numeric_limits<double>::infinity();
-  if (z >= 1)
+  }
+  if (z >= 1) {
     return std::numeric_limits<double>::infinity();
+  }
```
AJPfleger, line 188: This shouldn't be in this PR. Probably a duplicate from PR [#1343](https://github.com/acts-project/acts/pull/1343). But if the other one gets merged first this should vanish, I think.

AJPfleger (approved):
lgtm

Fatras/src/Utilities/LandauDistribution.cpp
paulgessinger: Right, I think it's gone now.

paulgessinger:
This would need reapproval, @AJPfleger.

AJPfleger (approved):

Issue #1345, opened 2022-07-26
Vertexing produces propagation errors


paulgessinger:
During propagation to the vertex position, the propagation seems to get stuck.

@andiwand has observed 

```
Step failed with EigenStepperError:3: Step size adjustment exceeds maximum trials
```

which seems to be the accurracy correction flip flopping between two step sizes and failing to converge.

In addition to this error, ATLAS observes an error

```
18:35:26 18:35:26    HelTrkLinPro   WARNING   Pixel::Barrel | No further navigation action, proceed to target. This is very likely an error
```

in minbias events with low-pt in reprocessing ([JIRA ticket](https://its.cern.ch/jira/browse/ATLASRECTS-7148))

andiwand:
see https://github.com/acts-project/acts/pull/1299 if you want to reproduce this issue

andiwand:
somehow we end up with a perigee inside the pixel barrel

´´´
intersect Acts::PerigeeSurface:
     Center position  (x, y, z) = (20.3294226, -79.1081599, 169.4577863) from pos 0.002635 0.000678 0.116209 with direction  0.108395 -0.421310  0.900416 navDir backward
StandardAborters: setStepSize navDir backward distance -187.753649
ConstrainedStep ( -∞,  -∞,  -∞,  -∞ ) update value 187.753649 original direction backward
17:49:50    HelTrkLinPro   ERROR     Step failed with EigenStepperError:3: Step size adjustment exceeds maximum trials
´´´

to me it looks like every time the perigee is not inside the beam pip we crash

[debug.log](https://github.com/acts-project/acts/files/9191398/debug.log)

asalzburger:
That's exactly what I was referring to in the JIRA. The perigee is in fact allowed to be in the pixel barrel - we should make sure the shorter catches this and ends the propagation accordingly.

andiwand:
here an extreme case

```
intersect Acts::PerigeeSurface:
     Center position  (x, y, z) = (2998.3168793, 474.3935881, -3004.7428436) from pos -0.007314  0.046232 -0.015186 with direction  0.701984  0.111057 -0.703481 navDir backward
StandardAborters: setStepSize navDir backward distance -4271.215506
ConstrainedStep ( -∞,  -∞,  -∞,  -∞ ) update value 4271.215506 original direction backward
08:34:00    HelTrkLinPro   ERROR     Step failed with EigenStepperError:3: Step size adjustment exceeds maximum trials
```

[...]

PR #1346, opened 2022-07-27
test: assert ConstrainedStep size

andiwand:
while working on https://github.com/acts-project/acts/issues/1345 I stumbled across `ConstrainedStep` and the way how it converts to a `Scalar`. I believe there are cases where it will return `+/- max` which will break the propagation

in this PR I add an `assert` to catch those cases

### DIFF
--- a/Core/include/Acts/Propagator/ConstrainedStep.hpp
+++ b/Core/include/Acts/Propagator/ConstrainedStep.hpp
@@ -95,10 +95,14 @@ struct ConstrainedStep {
   /// Cast operator to double, returning the min/max value
   /// depending on the direction
   operator Scalar() const {
+    Scalar result;
     if (direction == NavigationDirection::Forward) {
-      return (*std::min_element(values.begin(), values.end()));
+      result = *(std::min_element(values.begin(), values.end()));
+    } else {
+      result = *(std::max_element(values.begin(), values.end()));
     }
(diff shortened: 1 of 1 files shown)


Core/include/Acts/Propagator/ConstrainedStep.hpp
```diff
@@ -95,10 +95,14 @@ struct ConstrainedStep {
...
   operator Scalar() const {
+    Scalar result;
     if (direction == NavigationDirection::Forward) {
-      return (*std::min_element(values.begin(), values.end()));
+      result = (*std::min_element(values.begin(), values.end()));
+    } else {
+      result = (*std::max_element(values.begin(), values.end()));
     }
-    return (*std::max_element(values.begin(), values.end()));
+    assert(std::abs(result) < max_scalar);
```
paulgessinger, line 104: Where does `max_scalar` come from?

andiwand: oh sorry partial changes. fixing...

paulgessinger (approved):
lgtm

andiwand:
pandora's box opened

andiwand:
I guess we can close this PR @paulgessinger ? will follow up with https://github.com/acts-project/acts/pull/1347

PR #1348, opened 2022-07-27
fix: Add missing import for exatrkx example

paulgessinger:
After #1309, there's an import missing in `exatrkx.py`. This job isn't in the *required* checks, so the PR was merged with this failure.

I think this should fix this.

/cc @tboldagh @benjaminhuth

### DIFF
--- a/Examples/Scripts/Python/exatrkx.py
+++ b/Examples/Scripts/Python/exatrkx.py
@@ -10,6 +10,7 @@ from acts import UnitConstants as u
 if "__main__" == __name__:
     import os
     from digitization import configureDigitization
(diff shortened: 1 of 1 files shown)


tboldagh:
Thanks @paulgessinger

Examples/Scripts/Python/exatrkx.py
```diff
@@ -10,6 +10,7 @@
 if "__main__" == __name__:
     import os
     from digitization import configureDigitization
+    from acts.examples.reconstruction import addExaTrkx
```
tboldagh, line 13: indeed

tboldagh (approved):

PR #1349, opened 2022-07-27
ci: Switch gitlab test job to needs instead of dependencies

paulgessinger:
This should speed up CI execution a bit.


paulgessinger:
Tiny change. @AJPfleger ?

AJPfleger (approved):

PR #1350, opened 2022-07-27
feat: addFatras preselectParticles config for ParticleSelector

timadye:
Allow preselection of particles before Fatras using `ParticleSelectorConfig()`, eg.
```
s = addFatras(
    s, trackingGeometry, field,
    ParticleSelectorConfig(
        eta=(-4.0, 4.0),
        pt=(1 * u.GeV, 10 * u.GeV),
        removeNeutral=True,
    ))
```
This is usually required to use Pythia8 with Fatras.

To this end, an example of using Pythia8 to generate ttbar + mu=200 pile-up is included (commented-out) in `full_chain_itk.py`.


paulgessinger (approved):
Seems reasonable to me.

PR #1356, opened 2022-07-29
fix: improve Python full-chain ITk example

timadye:
For ttbar + PU200 example code (commented out in `full_chain_itk.py`), change particle selector pT > 1 GeV to pT > 0.15 GeV.

This change allows lower momentum particles to be included in the simulation. They may not be reconstructed and certainly won't be included in the efficiency calculation, but do provide more realistic hits in the detector to test the pattern recognition. @asalzburger suggested Fatras should still be able to cope with 150 MeV.

Note, that this is in code that is commented out in the example script. The PR merely provides a better example for anyone wanting to test ITk with ttbar and pile-up.

### DIFF
--- a/Examples/Scripts/Python/full_chain_itk.py
+++ b/Examples/Scripts/Python/full_chain_itk.py
@@ -55,7 +55,7 @@ s = addFatras(
     s,
     trackingGeometry,
     field,
-    # ParticleSelectorConfig(eta=(-4.0, 4.0), pt=(1.0 * u.GeV, 10.0 * u.GeV), removeNeutral=True),
+    # ParticleSelectorConfig(eta=(-4.0, 4.0), pt=(150 * u.MeV, None), removeNeutral=True),
     outputDirRoot=outputDir,
     rnd=rnd,
 )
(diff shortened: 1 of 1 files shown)


AJPfleger (commented):
Seems OK to me. What is the reason for this change? (Maybe you could also add it to the intro-comment)

Examples/Scripts/Python/full_chain_itk.py
```diff
@@ -55,7 +55,7 @@
     s,
     trackingGeometry,
     field,
-    # ParticleSelectorConfig(eta=(-4.0, 4.0), pt=(1.0 * u.GeV, 10.0 * u.GeV), removeNeutral=True),
+    # ParticleSelectorConfig(eta=(-4.0, 4.0), pt=(150 * u.MeV, None), removeNeutral=True),
```
paulgessinger, line 58: This was and stays commented out?

timadye: Yes, it's just an example in case people want to run a ttbar example. We decided we didn't want command-line options, and I think a separate file is overkill, so I think this is a reasonable way to document this.
I tested it by uncommenting.

timadye:
> Seems OK to me. What is the reason for this change? (Maybe you could also add it to the intro-comment)

Good idea.

AJPfleger (approved):

PR #1357, opened 2022-08-02
docs: remove artifact from setup_and_run_vertexing.md

andiwand:


paulgessinger:
I thought we use the anchor somewhere. Do we not?

andiwand:
oh that can be I just viewed the file with github but we actually render them here https://acts.readthedocs.io/en/latest/examples/howto/setup_and_run_vertexing.html?highlight=vertexing which looks fine. sorry for the noise

PR #1358, opened 2022-08-02
chore: clang-tidy modernize-use-nullptr

paulgessinger:
This enables the `modernize-use-nullptr` check in clang-tidy. This prevents using `0` for a null pointer.
This PR also changes `0` to `nullptr` wherever already used as a null pointer.


AJPfleger (approved):
lgtm.

Maybe you could add in the description:
"This PR also changes `0` to `nullptr` wherever already used as a null pointer."

PR #1359, opened 2022-08-02
chore: clang-tidy performance-move-const-arg

paulgessinger:
This enables the `performance-move-const-arg` check in clang-tidy.
It suggests replacing/removing `std::move` where it won't actually do anything, e.g. because the called function takes a const reference anyway, or the underlying type is trivially copyable. This PR removes the `std::move` calls in these cases.


AJPfleger (commented):
Looks good to me in general.

Maybe you could add in the description, that you also changed all relevant `move()`-cases in this PR.

As far as I can remember, the `CylinderSurface.*` are very similar to the `ConeSurface.*`, and the `cbounds` are there static too. But it seems you haven't touched them in this PR.

paulgessinger:
> As far as I can remember, the `CylinderSurface.*` are very similar to the `ConeSurface.*`, and the `cbounds` are there static too. But it seems you haven't touched them in this PR.

In `CylinderSurface` I think the shared pointer is either moved from a copy (not a const ref) or the const ref is not moved, so the clang-tidy warning doesn't trigger there.

AJPfleger (approved):

PR #1363, opened 2022-08-03
fix: any direction for helical track linearizer

andiwand:
Previously the `HelicalTrackLinearizer` assumed that the vertex we are fitting has to be reachable by back propagation. This is not always the case. In this PR we try to estimate the direction in which we will find the vertex and use it for the propagation.

@asalzburger mentioned that it would be good to back propagate from the first measurement of the particle on the track as this will always be "ahead" of the vertex

cc @paulgessinger


Core/include/Acts/Vertexing/HelicalTrackLinearizer.ipp
```diff
@@ -19,17 +19,21 @@ Acts::Result<Acts::LinearizedTrack> Acts::
...
+  auto intersection = perigeeSurface->intersect(gctx, params.position(gctx),
+                                                params.unitDirection(), false);
+
   // Create propagator options
   auto logger = getDefaultLogger("HelTrkLinProp", Logging::INFO);
   propagator_options_t pOptions(gctx, mctx, LoggerWrapper{*logger});
-  pOptions.direction = NavigationDirection::Backward;
+  pOptions.direction = intersection.intersection.pathLength >= 0
+                           ? NavigationDirection::Forward
+                           : NavigationDirection::Backward;
```
paulgessinger, line 30: So we only to the one propagation in this case then? Or should we failsafe it by trying both but in this ordering?

andiwand: I tried to do both directions here https://github.com/acts-project/acts/pull/1363/commits/4cc26a2966ac381905bf204d6a598b5684368409 just without order

I think we can go with the single direction for now and if that does not work at some point we could think about having an any direction propagation

paulgessinger (approved):
lgtm

acts-project-service:
The backport to `develop/v9.1.x` failed:
```
```
```bash
git fetch
# Create a new working tree
git worktree add .worktrees/backport-develop/v9.1.x develop/v9.1.x
cd .worktrees/backport-develop/v9.1.x
# Create a new branch
git switch --create backport-1363-to-develop/v9.1.x
git cherry-pick -x --mainline 1 8639c37faff32a288e2417df0a7f099131132b35
# Push it to GitHub
git push --set-upstream origin backport-1363-to-develop/v9.1.x
cd ../..
# Delete the working tree
git worktree remove .worktrees/backport-develop/v9.1.x
```
Then, create a pull request where the `base` branch is `develop/v9.1.x` and the `compare`/`head` branch is `backport-1363-to-develop/v9.1.x`.

PR #1364, opened 2022-08-03
fix: Refactor and fix component merging for GSF

benjaminhuth:
Fixes a bug in the component merging where the covariance was computed without respecting cyclic coordinates
This did blow up the variance in some cases and did lead to change of direction for some components. 

Additional to this, now the surface-type dependent behavior (with the `AngleDescription`) should be applied in all places where the component merging is used.

This also completely redid the unit test for the component merging, that should do a better job in testing the functionality.


asalzburger (approved):
Makes sense to me.

PR #1371, opened 2022-08-03
fix: initialising world in GdmlDetectorConstruction

asalzburger:
This PR is a one-liner (a few characters actually), which fixes the initialised world value to `nullptr` hence the detector construction runs now after checking
```c++
if (m_world == nullptr){
construct()
}
```


plariono (approved):
All good for me.

paulgessinger:
@andiwand indeed I believe we can (https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-member-init.html).

PR #1374, opened 2022-08-03
ci: Disable macOS job in v9.1.x

paulgessinger:
It's too much work to get the macOS job to build correctly for that old release, so I'm removing it for that series.


andiwand (approved):

paulgessinger:
Sorry @andiwand, I actually turned around and switch the job off because it's getting compile errors that I'm not willing to fix.

andiwand (approved):
fair enough!

lets take out the trash :grin:

PR #1378, opened 2022-08-04
feat: Checks for configuration of delta R values in seed finder

LuisFelipeCoelho:
This changes the initialisation of `deltaRMinTopSP`, `deltaRMaxTopSP`, `deltaRMinBottomSP` and `deltaRMaxBottomSP` to quiet_NaN. If the user does not configure these values, they take the values of `deltaRMin` and `deltaRMax`.

I think this makes the configuration easier and avoids errors when the user wants to use the same values for top and bottom.

@robertlangenberg


stephenswat (commented):
Overall, I think this is a sensible change, although I would suggest adding a check of the form:

```c++
static_assert(std::numeric_limits<decltype(deltaRMinTopSP)>::has_quiet_NaN, "Value of delta R must support NaN values");
```

Just to protect against some hard to debug problems if you ever run into platforms with not-quite-compliant floating point numbers.

stephenswat (approved):
Looks good with these updates!

LuisFelipeCoelho:
I don't know why the format check is failing 🤔

andiwand:
> I don't know why the format check is failing thinking

the CI is running `clang-format` version 10 which might do slightly different formatting then your newer local version

any chance you can get version 10 locally? @LuisFelipeCoelho

LuisFelipeCoelho:
Thank you @andiwand

andiwand (approved):

PR #1381, opened 2022-08-04
fix: RootBFieldWriter incorrect output

timadye:
* Fix `RootBFieldWriter` to write out the correct binning. The file can now be read back in for identical result.
  * Don't write out phi bins for the `(R,z)` field map, since there is no place for the phi coordinate in the root file. Uses `phi=0`.
  * In principle, this code is supposed to allow writing 2D or 3D file format, whether the source is `(R,z)` or `(x,y,z)`. I don't know if this is useful, and I haven't tested it.
* Tidy up BField reading in `FieldMapRootIo.cpp` and `BFieldMapUtils.cpp`
* Add `pytest` `test_bfield_writing`. This uses an expanded `bfield_writing.py` example to:
  1. create a 2T `acts.SolenoidBField`
  2. write it out as an `(R,z)` field-map to `solenoid.root`
  3. read it back in from `solenoid.root`
  4. write it again to `solenoid2.root`
  5. check that both files have the correct (identical) number of entries and `root_file_hashes.txt`.
    * Note: The two files should be identical, but don't have the same `root_file_hashes.txt`.
    * I have checked that they have the same entries using `root` `solenoid->Scan("*")`.

The test doesn't check the `(x,y,z)` field-map I/O, but I have tested read/write gives identical results with the ITk field-map.

This PR fixes the problems I reported in #669. @osbornjd, please check if it fixes your issues too.


Examples/Io/Root/src/RootBFieldWriter.cpp
```diff
@@ -35,23 +35,14 @@ void RootBFieldWriter::run(const Config& config,
...
       TFile::Open(config.fileName.c_str(), config.fileMode.c_str());
   if (outputFile == nullptr) {
-    throw std::ios_base::failure("Could not open '" + config.fileName);
+    throw std::ios_base::failure("Could not open '" + config.fileName + "'");
   }
-  outputFile->cd();
-  TTree* outputTree =
-      new TTree(config.treeName.c_str(), config.treeName.c_str());
+  TTree* outputTree = new TTree(config.treeName.c_str(),
+                                config.treeName.c_str(), 99, outputFile);
```
andiwand, line 41: what does mean `99`?

Examples/Io/Root/src/RootBFieldWriter.cpp
```diff
@@ -140,27 +142,22 @@ void RootBFieldWriter::run(const Config& config,
...
+            "written out by this writer.");
       }
     }
 
-    stepX = fabs(minX - maxX) / nBinsX;
-    stepY = fabs(minY - maxY) / nBinsY;
-    stepZ = fabs(minZ - maxZ) / nBinsZ;
+    stepX = fabs(minX - maxX) / (nBinsX - 1);
+    stepY = fabs(minY - maxY) / (nBinsY - 1);
+    stepZ = fabs(minZ - maxZ) / (nBinsZ - 1);
```
andiwand, line 154: should we switch to `std::abs` here?

Examples/Io/Root/src/RootBFieldWriter.cpp
```diff
@@ -224,48 +225,37 @@ void RootBFieldWriter::run(const Config& config,
...
-          r = raw_r / Acts::UnitConstants::mm;
-          Bz = bField.z() / Acts::UnitConstants::T;
-          Br = VectorHelpers::perp(bField) / Acts::UnitConstants::T;
-          outputTree->Fill();
-        }
-      }
-    }  // for
+
+    stepR = fabs(minR - maxR) / (nBinsR - 1);
+    stepZ = fabs(minZ - maxZ) / (nBinsZ - 1);
```
andiwand, line 238: switch to `std::abs`?

Examples/Scripts/Python/bfield_writing.py
```diff
@@ -1,29 +1,50 @@
...
+    cfg.bField = field
+    cfg.gridType = acts.examples.RootBFieldWriter.GridType.rz
+    cfg.fileName = str(outputDir / "solenoid.root")
+    cfg.treeName = "solenoid"
 
-acts.examples.RootBFieldWriter.run(cfg, acts.logging.VERBOSE)
+    acts.examples.RootBFieldWriter.run(cfg, acts.logging.VERBOSE)
 
-print("Done")
+    for i in range(rewrites):
```
andiwand, line 31: do you think there is a use case for rewriting more than once?

andiwand (approved):
look good to me - thank you for fixing this!

Examples/Io/Root/src/RootBFieldWriter.cpp
timadye: That's just the default value of the intervening constructor argument. It's called `splitlevel`. but it doesn't really matter what it means. Unfortunately you can't specify C++ keyword arguments like you can with Python.

The main value of this change is to specify the `outputFile` explicitly, rather than relying on the global "current `TDirectory`".

Examples/Io/Root/src/RootBFieldWriter.cpp
timadye: Good question, which led me down a rabbit hole.
 * I don't know which is better in practice, but I prefer `std::abs`.
 * both `fabs` and `std::abs` should have `#include <cmath>`.
 * If we reversed `min` and `max`, we might not need `abs` at all. If the user specifies them the wrong way round, then the subsequent code wouldn't work with `fabs`, but might still work without. So maybe it's better to remove these `fabs` altogether.

```suggestion
    stepX = (maxX - minX) / (nBinsX - 1);
    stepY = (maxY - minY) / (nBinsY - 1);
    stepZ = (maxZ - minZ) / (nBinsZ - 1);
```

What do you think?

Examples/Io/Root/src/RootBFieldWriter.cpp
timadye: Same questions as above. I'll resolve this one.

Examples/Io/Root/src/RootBFieldWriter.cpp
andiwand: I agree we should be able to assume that it is ordered. maybe we can add an assertion for max > min or exception as you prefer!

timadye: Good idea! I changed it.

andiwand (approved):

Examples/Detectors/MagneticField/src/FieldMapRootIo.cpp
```diff
@@ -102,7 +114,7 @@ ActsExamples::makeMagneticFieldMapXyzFromRoot(
     zPos.push_back(z);
     bField.push_back(Acts::Vector3(Bx, By, Bz));
   }
-  inputFile->Close();
+  delete inputFile;
```
paulgessinger, line 117: Does the destructor close the file correctly?

andiwand:
If you want to merge already just attach the automerge label @timadye !

Examples/Detectors/MagneticField/src/FieldMapRootIo.cpp
timadye: Yes, it should. The `TTree` is owned by the `TFile`, so we don't delete the `TTree`.

As you know, there is a potential issue for `TFile`s owned by other objects if they are destroyed in the global destructor, since this can give a double delete if they are destroyed after ROOT's global destructor. That's not a problem here.

paulgessinger: Right. But I thought if we `TFile::Open` here, we must also `->Close()`. But ok if `delete` does the job I'm ok.

PR #1384, opened 2022-08-05
ci: Cleaned up the workflow files a little.

wermos:
There were many instances of trailing `--` in the CMake commands, followed by nothing, which I removed. Moreover, many of the targets were passed  to the underlying build system as a build-system-specific flag. I changed those to make it so that the `--target` CMake flag is used.


.github/workflows/builds.yml
```diff
@@ -398,13 +398,13 @@ jobs:
           -DACTS_USE_SYSTEM_EIGEN3=OFF
           -DACTS_BUILD_PLUGIN_JSON=ON
       - name: Build
-        run: ${SETUP} && ( cmake --build build -- || cmake --build build -- )
+        run: ${SETUP} && cmake --build build
```
andiwand, line 401: I believe that second try was actually intentional by @paulgessinger because we saw some random failures?

andiwand (commented):
we need to clarify on the comment but otherwise happy to put this in

.github/workflows/builds.yml
paulgessinger: Yes, this is to implement sort of retry mechanism because this build fails for technical reasons. Can you revert this bit?

wermos: I'll do so, and add a comment about why this is necessary too.

paulgessinger (approved):

PR #1393, opened 2022-08-08
docs: Add docs clean target to cmake

paulgessinger:
This PR adds a target `docs-clean` to CMake which will clean out all files generated by the docs build.


AJPfleger (approved):
lgtm. thanks

andiwand:
should we include a dependency from `docs` to `docs-clean` so we always get a fresh build?

paulgessinger:
> should we include a dependency from `docs` to `docs-clean` so we always get a fresh build?

Not sure. That would take longer to run every time.

PR #1395, opened 2022-08-08
docs: first brush over the geometry pages

asalzburger:
This PR combines the two geometry pages into one and adds some additional info and highlights.


docs/core/geometry.md
```diff
@@ -209,41 +223,57 @@ Possible extensions are:
...
 geometry,
 
-## Attaching a 3D detector geometry
+### Attaching a 3D detector geometry
 
 Usually, a 3D detector model geometry exists, which is either native to the full
 detector simulation (Geant4) or is translated into it. This model, however, is
 in general too detailed for track reconstruction: navigating through the
-detailed detector geometry
+detailed detector geometry is generally costy and one can profit greatly from a simplication mechanism.
```
paulgessinger, line 244:
```suggestion
detailed detector geometry is generally costly and one can profit greatly from a simplication mechanism.
```

docs/core/geometry.md
```diff
@@ -126,6 +131,10 @@ surfaces, a special `InfiniteBounds` class is available.
 ![DiscBounds](/figures/geometry/DiscBounds.png)
 ![PlaneBounds](/figures/geometry/PlaneBounds.png)
 
+```{tip}
+The coordinate systems define - in an ideal setup - also the readout measurement directions. In such a case, a track prediction from the propgation, e.g. will already be in the correct frame of the measurement and residual or compatibility checks will not need additional coordinate transformations.
```
AJPfleger, line 135:
```suggestion
The coordinate systems define - in an ideal setup - also the readout measurement directions. In such a case, a track prediction from the propagation, e.g. will already be in the correct frame of the measurement and residual or compatibility checks will not need additional coordinate transformations.
```

docs/core/geometry.md
```diff
@@ -184,7 +193,7 @@ at creation.
 ![VolumeBounds](/figures/geometry/VolumeBounds.png)
 ![CylinderVolumeBounds](/figures/geometry/CylinderVolumeBounds.png)
 
-## Material description
+## Detector Material description
```
AJPfleger, line 196:
```suggestion
## Detector material description
```

docs/core/geometry.md
```diff
@@ -197,6 +206,11 @@ The basic information for any material is:
 
 This information is confined together in the `Material` class.
 
+```{note}
+In track reconstruction, only an effective material description is needed, i.e. non-physical values in regards of atomic number, elementary charge or even density are allowed, as long as the effective relative radiation length and `A/Z x rho` ratio can be retrieved.
```
AJPfleger, line 210:
```suggestion
In track reconstruction, only an effective material description is needed, i.e. non-physical values in regards of the atomic number, the elementary charge or even the density are allowed, as long as the effective relative radiation length and `A/Z x rho` ratio can be retrieved.
```

docs/core/geometry.md
```diff
@@ -209,41 +223,57 @@ Possible extensions are:
  * `HomogeneousSurfaceMaterial`, homogeneous material description on a surface
  * `BinnedSurfaceMaterial`, an arbitrarily binned material description with a
     corresponding `BinUtility` object
+
+In addition, a dedicated extension exists to allow configuration of the material
+mapping process, that is in further descibed below.
```
AJPfleger, line 228:
```suggestion
mapping process, that is in further described below.
```

docs/core/geometry.md
AJPfleger: also "simplification" instead "simplication"?

AJPfleger (changes requested):

docs/core/geometry.md
```diff
@@ -209,41 +223,57 @@ Possible extensions are:
...
-### Existing plugins for 3D geometry libraries
+#### Existing plugins for 3D geometry libraries
 
 Very simple helper methods for 3D libraries exist, they are certainly not
 optimised, but used for templating:
 
-* `TGeoDetectorElement` connects a TGeo volume to a `Surface` (see [](TGeo plugin))
-* `DD4HepDetectorElement` connects a DD4hep volume (based on TGeo) to a `Surface` (see [](DD4hep plugin))
+* `TGeoDetectorElement` connects a TGeo volume to a `Surface`
+* `DD4HepDetectorElement` connects a DD4hep volume (based on TGeo) to a `Surface`
```
benjaminhuth, line 262: We could add the `Geant4DetectorElement` here as well, maybe also with a link to the documentation (via the `{class}` prefix and namespace)

benjaminhuth (approved):
I think we can also merge this in in the current state, right?

docs/core/geometry.md
asalzburger: This will have to be updated anyway for the `Detector` infrastructure, I will make a new PR once this is done.

asalzburger:
> I think we can also merge this in in the current state, right?

Yes, I did a little cleanup & added the `Geant4DetectorElement` as suggested.

benjaminhuth (approved):

PR #1397, opened 2022-08-08
docs: polish tgeo plugin doc

niermann999:
I did not find much to change, just a few typos and a mentioning of the surface splitting, I also made the figure captions visible.

TODO: Add link to examples, once TGeoDetector is described there


benjaminhuth (approved):
I think we could also merge this in the current state, what do you think @niermann999 ?

niermann999:
> I think we could also merge this in the current state, what do you think @niermann999 ?

I do believe so, I don't remember why I kept it WIP

benjaminhuth (approved):

PR #1402, opened 2022-08-08
docs: adding Fatras description

asalzburger:
This PR adds the Fatras documentation consisting of
* Fatras simulation
* Digitisation
* Configuration


AJPfleger (commented):
I think the new UnitTests should be in a separate PR.

benjaminhuth:
@asalzburger I would suggest to bring this in in this state now, what do you think? Should we separate the unit test from this PR as @AJPfleger suggested or merge this directly?

asalzburger:
Indeed, as suggested,  the new unit tests should be split off.

asalzburger:
Unit tests are partly superseded, closing this one for now.

PR #1404, opened 2022-08-08
refactor: no return for add* python helpers

andiwand:
after working on the full chain doc I wondered why we do `s = addSomething(s, ...)` all the time. to me it looks kind of confusing as the reader might think `s` is changing every time we add something.

maybe we want to remove the return value of the add functions completely? @paulgessinger @timadye


timadye:
Hi @andiwand,

This was originally proposed by @asalzburger in #1098, so we should get his feedback.

I think the idea is to emphasize that the `Sequencer` object is modified by the routine, even though it returns the same object. But maybe it is overkill to get the user/caller do it.

For myself, I don't have a strong opinion either way, so I wouldn't mind if it was changed - as long as it was changed consistently in all the examples and documentation.

Thanks,
Tim.

andiwand:
hmm makes sense. I just wonder if the return value emphasizes this enough as you can easily skip it without any loss of functionality. I guess what would be nice is something like `s.addSomething(...)`. we might be able to achieve this by wrapping the sequencer now that everything is in `simulation.py` or `reconstruction.py`?

timadye:
That could be easier to use, but could make the implementation harder to understand. The aim with this interface is to make it simple to get started, but not to discourage people playing inside the `add*()` functions' code. I'll leave it to others to comment on this case, because I am sometimes similarly tempted to improve the interface.

paulgessinger:
I would prefer not to wrap the sequencer. My original reason for having it as an argument an return value was that you could pass in `None` and the original `run*` function would create a default sequencer. I guess this is not relevant anymore with the `add*` functions. 

I think my preferred solution would be to drop the return value consistently.

andiwand:
I agree - I will draft these changes

timadye:
Cool. Can you change the PR title, since this doesn't any more just affect the OOD.

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -645,7 +639,7 @@ def addCKFTracks(
     outputDirCsv: Optional[Union[Path, str]] = None,
     outputDirRoot: Optional[Union[Path, str]] = None,
     selectedParticles: str = "truth_seeds_selected",
-) -> acts.examples.Sequencer:
+) -> None:
```
timadye, line 642: I think this indicates that the function returns `None`, not that it doesn't return a value. If it really doesn't return a value, then I think this should be `NoReturn`, but I wonder if just not specifying a return type is OK - and clearer.

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -536,16 +536,14 @@ def customLogLevel(custom: acts.logging.Level = acts.logging.INFO):
                 )
             )
 
-    return s
```
timadye, line 539: I think it might be better to return the value, even if we no longer recommend the caller *using* the value. Then people wouldn't have to update all their scripts immediately.

Better would be to add a comment in the docstring to say that the `Sequence` value is returned for compatibility reasons.

If this is adopted, then I don't know what is best for the function signature, which is mostly just for documentation purposes. (Do we ever check it?)

Examples/Python/python/acts/examples/simulation.py
```diff
@@ -305,7 +298,7 @@ def addPythia8(
             )
         )
 
-    return evGen if returnEvGen else s
```
timadye, line 308: As it says above, `evGen` was only returned for compatibility with `common.addPythia8`. Since that's been removed, we can drop this: `return s` (as I suggested before) or don't return anything.
Similarly, we can remove the docstring comments about `common.addPythia8`.

Examples/Python/python/acts/examples/reconstruction.py
timadye: Thanks for the correction. Sorry for the incorrect comment: I misread the stackoverflow.

Examples/Python/python/acts/examples/reconstruction.py
andiwand: I was also thinking about this but I fear if we do not make a clean cut no one will adopt. cc @paulgessinger

timadye: I think people will follow your documentation, the examples, and especially our workshop, all of which show the new syntax. Returning an unused value is a common idiom, so I don't think it's a problem even if this never changes - at least until we need something else, which would anyway change the signature.

andiwand: agreed :+1: will adopt this tomorrow

andiwand:
I made the changes @timadye. I left the `None` return type for now, with the hope that a linter might notify the consumer that a return value should not be expected

Examples/Scripts/Python/full_chain_itk.py
```diff
@@ -33,15 +33,15 @@
 from acts.examples.itk import itkSeedingAlgConfig
 
 s = acts.examples.Sequencer(events=100, numThreads=-1)
-s = addParticleGun(
+addParticleGun(
     s,
     MomentumConfig(1.0 * u.GeV, 10.0 * u.GeV, True),
     EtaConfig(-4.0, 4.0, True),
     ParticleConfig(1, acts.PdgParticle.eMuon, True),
```
timadye, line 40:
```suggestion
    MomentumConfig(1.0 * u.GeV, 10.0 * u.GeV, transverse=True),
    EtaConfig(-4.0, 4.0, uniform=True),
    ParticleConfig(1, acts.PdgParticle.eMuon, randomizeCharge=True),
```
can you add kwargs as for the odd example? There are also a few other cases which could be changed if you like, but these are the important ones I think.

timadye (approved):
all looks good to me

acts-project-service:
```
```
```bash
git fetch
# Create a new working tree
# Create a new branch
git switch --create backport-1404-to-develop/v19.x
git cherry-pick -x --mainline 1 afb91a583e6db641c9b39354c7494da5726b8681
# Push it to GitHub
git push --set-upstream origin backport-1404-to-develop/v19.x
cd ../..
# Delete the working tree
```
Then, create a pull request where the `base` branch is `develop/v19.x` and the `compare`/`head` branch is `backport-1404-to-develop/v19.x`.

PR #1409, opened 2022-08-09
fix: change typeid check to dynamic_cast

asalzburger:
This PR changes the way we check for compatibility in the `WhiteBoard` as sometimes the typeid check would fail, even if the cast would work.

The reason for the failure is not fully understood, but a `dynamic_cast` is ok in this case, as this is not a time-critical operation.

### DIFF
--- a/Examples/Framework/include/ActsExamples/Framework/WhiteBoard.hpp
+++ b/Examples/Framework/include/ActsExamples/Framework/WhiteBoard.hpp
@@ -100,11 +100,14 @@ inline const T& ActsExamples::WhiteBoard::get(const std::string& name) const {
     throw std::out_of_range("Object '" + name + "' does not exists");
   }
   const IHolder* holder = it->second.get();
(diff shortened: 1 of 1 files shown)


paulgessinger:
> Just a question out of curiosity: Is there a reason why the `WitheBoard` does not e.g. use `std::any` (that would bring these typechecks)? I could imagine that would be more robust than a custom logic, but I have not thought deeply about that.

This was written back on C++14 (not by me 😄) so that's why.

I think switching to `std::any` would also be a good idea at some point.

paulgessinger (approved):

andiwand:
I wonder if this is due to const or using base classes. but std::any would result in the same problem because it will not do any fancy casting right?

benjaminhuth:
> I wonder if this is due to const or using base classes. but std::any would result in the same problem because it will not do any fancy casting right?

Hmm as far as I know std::any is completely typesafe... Or what do you mean by fancy casting?

PR #1410, opened 2022-08-09
feat: Enable Delegates to conveniently use stateful lambdas

benjaminhuth:
This PR adds functionality to the `Acts::Delegate`, that enables to construct, assign and connect stateful lambdas like this:

```c++
int n = 0;
auto lambda = [&](){ ++n; };

Acts::Delegate<void()> d(lambda);
d = lambda;
d.connect(lambda);
```

The use of temporary stateful lambdas is forbidden due to deleted overloads with `Callable &&`:
```c++
Acts::Delegate<void()> d([&](){ ++n; }); // compile error
```

I havn't checked yet if this can simplify code somewhere in Core or Examples, but I ran in situations myself where this would be useful.

### DIFF
--- a/Core/include/Acts/Utilities/Delegate.hpp
+++ b/Core/include/Acts/Utilities/Delegate.hpp
@@ -35,13 +35,17 @@ class Delegate<R(Args...)> {
   /// Alias of the return type
   using return_type = R;
   /// Alias to the function pointer type this class will store
-  using function_type = return_type (*)(const void*, Args...);
+  using function_type = return_type (*)(const void *, Args...);
 
   using function_ptr_type = return_type (*)(Args...);
 
   template <typename T, typename C>
   using isSignatureCompatible =
-      decltype(std::declval<T&>() = std::declval<C>());
+      decltype(std::declval<T &>() = std::declval<C>());
+
+  template <typename T>
+  using isNoFunPtr = std::enable_if_t<
+      not std::is_convertible_v<std::decay_t<T>, function_type>>;
 
  public:
   Delegate() = default;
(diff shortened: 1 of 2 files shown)


paulgessinger (approved):
So this just explicitly tries to grab the `operator()` function pointer for a lambda? Should be ok I think!

asalzburger:
@paulgessinger is this still waiting for the sentinel backlog?

paulgessinger:
Still working on fixing that, sorry. I'll override the merge here.

timadye:
Can we document the additional functionality in this PR in the [comments at the top](https://github.com/acts-project/acts/blob/main/Core/include/Acts/Utilities/Delegate.hpp#L34). I saw the limitations listed there and used a [complicated workaround](https://gitlab.cern.ch/atlas/athena/-/merge_requests/71697/diffs?commit_id=0a721f48eb7f7a7dbf929a6dcb8bf1eef684e059), when I could have just used `connect(lambda)`. Fortunately, I eventually noticed the comments further down.

paulgessinger:
Thinking about this again, I'm wondering if we should make this more explicit somehow, with a tag type to clearly mark what's going on. 

I've seen some cases where this was (ab)used and it can be a bit of a footgun. We use the lvalue-ness to imply it's owned somewhere... but that's not clear to the user at all. 

Thoughts?

andiwand:
ping @benjaminhuth

timadye:
I assume the issue is that the delegate does not own the lambda - or is there something else I might be missing? My use has the lambda in the same scope as the extensions (and hence delegate), so this is good. It is made clear in the comments at the top of `Delegate.h`. I'm not sure what else one can do to protect the user.

benjaminhuth:
@paulgessinger I'm not 100% sure what you mean by "make it more explicit"?

Like `Delegate(my_stateful_lambda, Acts::Stateful{})` resp `my_delegate.connect(my_stateful_lambda, Acts::Stateful{})`? And what kind of abuses you refer to?

So to me it seems like a good proposal to improve the documentation to showcase the construction/connection with stateful lambdas...

paulgessinger:
I've seen it being used to set a lambda and then not being aware that it immediately dangles when the lambda itself goes out of scope. 

This can happen as well of course with member functions but it's less likely to do by mistake I'd say.

PR #1412, opened 2022-08-10
refactor: change WhiteBoard backend to std::any

benjaminhuth:
Just a test to see if it would work. Still WIP


Examples/Framework/include/ActsExamples/Framework/WhiteBoard.hpp
```diff
@@ -99,17 +84,16 @@ inline const T& ActsExamples::WhiteBoard::get(const std::string& name) const {
   if (it == m_store.end()) {
     throw std::out_of_range("Object '" + name + "' does not exists");
   }
-  const IHolder* holder = it->second.get();
 
-  const auto* castedHolder = dynamic_cast<const HolderT<T>*>(holder);
-  if (castedHolder == nullptr) {
-    throw std::out_of_range("Type mismatch for object '" + name + "'");
+  try {
+    const T& ret = *std::any_cast<T>(&it->second);
```
timadye, line 89: I think the pointer form of `std::any_cast<T>(const std::any*)` returns `nullptr` if the types don't match (case 4, [here](https://en.cppreference.com/w/cpp/utility/any/any_cast)). It doesn't throw `std::bad_any_cast` as the non-pointer form does.
The behaviour is like what `dynamic_cast` does, so this is closer to the original code.

benjaminhuth: Ah you're right!

timadye (approved):
lgtm

benjaminhuth:
So there still are these failures in the examples tests and physmon tests... 
It looks like something does not work here with the access of the stored objects I think, but currently fixing this is not my priority. I'll leave the PR open for now, so if someone has an idea whats going on here we can proceed on this.

andiwand:
what was the problem with these changes @benjaminhuth ? I might want to follow up on this later

PR #1430, opened 2022-08-12
docs: Fix sphinx-level warnings

paulgessinger:
This change makes sphinx fail on warnings (it will keep going to the end
though). Also, all warnings are either fixed or ignored with a
expression pattern matching. This is necessary because the way the API
doc works, sphinx will complain about duplicate anchors.


paulgessinger:
Hm, locally this works for me. I'll have to investigate.

docs/plugins/dd4hep.md
```diff
@@ -5,7 +5,7 @@ The DD4hepPlugin allows building of a {class}`Acts::TrackingGeometry` from
 
 :::{note}
 Previous versions of ACTS (prior to `v20.0.0`) used a different integration
-mechanism. Please refer to [](dd4hep_old) for more information.
+mechanism. Please refer to the documentation of prior versions for more information
```
andiwand, line 8:
```suggestion
mechanism. Please refer to the documentation of prior versions for more information.
```

docs/core/definitions/units.rst
```diff
@@ -1,118 +1,4 @@
 Unit definitions and conversions
```
andiwand, line 1: should we port this file to markdown?

docs/conf.py
```diff
@@ -17,20 +18,25 @@
...
 source_suffix = {
     ".rst": "restructuredtext",
     ".md": "markdown",
 }
 master_doc = "index"
 # ensure the in-source build directory is ignored
-exclude_patterns = [
-    "_build",
-]
+exclude_patterns = ["_build", "api/api_stub.rst", "api/api_index.rst"]
```
andiwand, line 39: is the `api/*` stuff used somewhere? if not maybe we can remove it entirely

docs/conf.py
```diff
@@ -17,20 +18,25 @@
...
+
 extensions = [
     "breathe",
     "myst_parser",
     "sphinx.ext.mathjax",
+    "warnings_filter",
 ]
+
+warnings_filter_config = str(Path(__file__).parent / "known-warnings.txt")
+warnings_filter_silent = True
```
andiwand, line 31: off-pr this is where I miss some kind of export keyword in python

Core/include/Acts/EventData/VectorMultiTrajectory.hpp
```diff
@@ -16,7 +16,9 @@ namespace Acts {
 
 class VectorMultiTrajectory final
     : public MultiTrajectory<VectorMultiTrajectory> {
+#ifndef DOXYGEN
   friend MultiTrajectory<VectorMultiTrajectory>;
+#endif
```
andiwand, line 21: out of curiosity: why does doxgen complain about `friend`?

andiwand (approved):

Core/include/Acts/EventData/VectorMultiTrajectory.hpp
paulgessinger: 🤷  Doxygen is in fact fine, but sphinx+breathe somehow get tripped up. I could actually remove this to check if this also goes away with Doxygen 1.9.4, I didn't try that exact combination.

docs/conf.py
paulgessinger: True. This mechanism of `conf.py` is a bit awkward.

docs/conf.py
paulgessinger: The `index.rst` links out to `api/api.rst`, so that shouldn't be ignored. At the same time, we don't want sphinx to build these extra `.rst` files.

docs/core/definitions/units.rst
paulgessinger: Should be easy enough.

paulgessinger: Done now. I unfortunately have to use `eval-rst` to have the rST inside the doxygen comment render correctly.

andiwand (approved):

PR #1431, opened 2022-08-13
docs: Use DEVNULL instead of PIPE in apidoc subprocess

paulgessinger:
I suspect this might have been failing our docs build for a while.


paulgessinger:
This seems to fix it indeed.

andiwand (approved):

Issue #1385, opened 2022-08-05
Failures during propagation


paulgessinger:
I've observed two more cases of propagation failures when running over Pythia8 HardQCD + SoftQCD events at mu200.

- [x] FATRAS: 
  `16:54:01    FatrasSimula   ERROR     event 571 particle id=(82|134|591|0|0)|pdg=0|q=-1|m=0|p=11.5181 failed to simulate with error PropagatorError:3: Propagation reached the configured maximum number of steps`. 
  This could be a navigation problem where the particle leaves the detector, or potentially a looper that isn't cought by the looper protection.
- [ ] Track finding
  ```
  16:51:54    TrackFinding   ERROR     Propapation failed: EigenStepperError:3 Step size adjustment 
  exceeds maximum trials with the initial parameters 33554 :
   1.37607
   2.05756
   0.0118902
   0.912868
   2.01847
  70.6078
  ```

/cc @andiwand @asalzburger

andiwand:
I can look into this next week

paulgessinger:
I should be able to supply an input file if you don't want to run Pythia locally.

andiwand:
I lately made the effort of compiling it so I would like to try running it as long as my CPU does not throttle :smile:

andiwand:
> FATRAS:
> 16:54:01 FatrasSimula ERROR event 571 particle id=(82|134|591|0|0)|pdg=0|q=-1|m=0|p=11.5181 failed to simulate with error PropagatorError:3: Propagation reached the configured maximum number of steps.
> This could be a navigation problem where the particle leaves the detector, or potentially a looper that isn't cought by the looper protection.

looking at the log output the particle seems to be invalid
- `pdg=0` means invalid https://github.com/acts-project/acts/blob/main/Core/include/Acts/Utilities/PdgParticle.hpp#L17
- it has `q=-1` but also `m=0`

I wonder if this is a bug in the pythia8 conversion or a bug in fatras

@paulgessinger

andiwand:
> 16:51:54    TrackFinding   ERROR     Propapation failed: EigenStepperError:3 Step size adjustment 
exceeds maximum trials with the initial parameters 33554 :

the "good" news is that all of them seem to have the same origin. the step size goes to zero and we run into the iteration maximum. as far as I know the step size would only get smaller and smaller if we cannot reach the desired accuracy in the runge-kutta integration. I will investigate further why this happens

andiwand:
[...]

PR #1433, opened 2022-08-15
fix: navigator priority with negative boundary distance

andiwand:
trying to fix a bug discovered with fatras here https://github.com/acts-project/acts/issues/1385

in some special cases the propagation will get stuck in a loop as the navigator is flip-flopping the target boundary of two volumes. one ahead and one behind.

these changes reorder the priority of the boundaries. first we will target everything in propagation direction and only if there are none left we will allow opposite direction


andiwand:
these changes seem to fix the fatras propagation errors seen here https://github.com/acts-project/acts/issues/1385 @paulgessinger 

I wonder if we could discard the wrong sign entirely because we would only end up stepping in the wrong direction?

paulgessinger:
Hold this PR for a potential unit test addition, or merge as is @andiwand?

andiwand:
> Hold this PR for a potential unit test addition, or merge as is @andiwand?

I will try to set one up today/tomorrow

paulgessinger (approved):
let's gamble, try merging.

PR #1440, opened 2022-08-16
chore: Clang-tidy check headers

paulgessinger:


CI/clang_tidy/merge.py
```diff
@@ -0,0 +1,39 @@
...
+    items = set()
+
+    for infile in args.inputfiles:
+        with open(infile, "r", encoding="utf-8") as f:
+            items = items | set(ItemCollection(__root__=json.load(f)).__root__)
+
+    items = list(items)
+
+    print("Write to", args.output)
+    with open(args.output, "w+") as jf:
```
andiwand, line 34: doesn't append result in an invalid json?

andiwand (commented):
142 files :smile: 

looks cool!

CI/clang_tidy/merge.py
paulgessinger: I guess you're right. Will change this.

Core/include/Acts/Digitization/PlanarModuleCluster.hpp
```diff
@@ -40,8 +40,7 @@ class PlanarModuleCluster : public Measurement<BoundIndices, 3> {
                       Base::CovarianceMatrix cov, double loc0, double loc1,
                       double t, std::vector<DigitizationCell> dCells,
                       const DigitizationModule* dModule = nullptr)
-      : Base(sourceLink, kIndices, Base::ParametersVector(loc0, loc1, t),
-             std::move(cov)),
+      : Base(sourceLink, kIndices, Base::ParametersVector(loc0, loc1, t), cov),
```
andiwand, line 43: why did it complain about the moves? not movable?

paulgessinger: I think "trivially copyable".

paulgessinger:
Hm, this now seems to have a number of false positives..

paulgessinger:
I bumped up clang-tidy to v16 to see if that fixes the false negatives (i.e. warnings that are incorrect), but no luck. Not sure how to proceed with this then.

I can circumvent these with comment markers telling clang-tidy to ignore them, but that seems a bit fragile.

paulgessinger:
I added an extra clang tidy config file that the header script applies. That's not necessarily great, but right now I don't see a great alternative. 

Thought? @andiwand @benjaminhuth @timadye

CI/clang_tidy/merge.py
andiwand: I think this is not resolved yet

paulgessinger, line 34:
```suggestion
    with open(args.output, "w") as jf:
```

paulgessinger: I switched this to `w` instead of `w+`. Should be changed now.

andiwand:
> I added an extra clang tidy config file that the header script applies. That's not necessarily great, but right now I don't see a great alternative.
> 
> Thought? @andiwand @benjaminhuth @timadye

fair enough. we can always improve on that but I think it is good to have a baseline merged

andiwand (approved):

.github/workflows/docs.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     container: ghcr.io/acts-project/ubuntu2004:v29
     env:
-        DOXYGEN_WARN_AS_ERROR: YES
+        DOXYGEN_WARN_AS_ERROR: FAIL_ON_WARNINGS
```
andiwand, line 16: shouldn't that disappear with https://github.com/acts-project/acts/pull/1497/files ? or is a merge from master missing?

benjaminhuth:
I have to admit I did not understood every detail of the clang-tidy integration. It looks quite complicated in total, but I would agree that its great to have a working version and maybe later improve on this...

.github/workflows/docs.yml
paulgessinger: Yes it should, as soon as this branch is updated it should be gone I believe.

paulgessinger:
> I have to admit I did not understood every detail of the clang-tidy integration. It looks quite complicated in total, but I would agree that its great to have a working version and maybe later improve on this...

Getting clang-tidy to run on headers is not straightforward. That's why it gets somewhat complicated unfortunately. But from the changeset, you can see that it does flag a number of notes that we can benefit from fixing.

In total we now have the integrated CMake clang-tidy run based off of the compilation database, and then this header one, where I just use `find` and `parallel` to execute clang-tidy on all the headers, dropping unavoidable compile errors, and merge everything together.

andiwand:
can you add a little wrap-up in the description @paulgessinger ? as far as I understood clang-tidy is a pain in the butt with header file implementations and you are using a work around here. but the warning/errors are not consistent with cpp clang-tidy results?

paulgessinger:
The warnings/errors should be consistent. I **think** that in case some of the includes are not resolved, clang-tidy sometimes produces wrong results. The implicit bool conversion warnings that popped up here looked to me like it was defaulting to the C behavior that unknown types are `int`s, and then complaining that `bool` was auto-converted to `int`. That's clearly wrong, so I filtered this by disabling that specific warning for the header mode.

I propose we merge this (with all the fixes), see if we get bogus warnings frequently, and revert the check part if that's the case.

benjaminhuth:
Just one question maybe:
There are these two files `.clang-tidy` and `.clang-tidy-headers`, and additional to that in the file `cmake/ActsStaticAnalysis.cmake` are also some options defined... 
Is some of this redundant or do I not understand correctly how this interacts?

benjaminhuth:
Okay, that sounds reansonable. Let's merge then!

andiwand:
should the cmake clang-tidy config and `.clang-tidy` always match? if so it might be a future improvement to read from / write to one of them

paulgessinger:
You mean to generate it in the CI before the clang-tidy job runs?

andiwand:
either generate `.clang-tidy` from by running cmake or reading `.clang-tidy` from cmake and use it for calling clang-tidy locally and in the ci

benjaminhuth:
Okay, if we change this I would argue against generating the file in cmake. This has no real advantage I think.

Running the cmake-triggered clang-tidy with the global `.clang-tidy` would be a good option if we don't want to steer the the checks using cmake options (which is not done currently I think)...

PR #1444, opened 2022-08-17
feat: don't use TBB in Sequence(numThreads=1)

timadye:
* Let's the `Sequencer` bypass the Intel TBB library if `numThreads=1`.
  * This could simplify debugging.
* Uses some wrapper functions for most of the TBB functions that we use in `Sequencer`.
  * Note that only a small subset of TBB functions are implemented, and `tbb::blocked_range` (which doesn't require any thread setup) is still taken from the TBB library.
* However, if `NO_TBB` is defined, then don't use TBB library at all (throws an exception unless `numThreads=1` or `-1`).
  * This should allow the ACTS Examples to be built without the TBB library (and removes one dependency on ROOT).

At the moment, this doesn't have any CMake support for disabling linking with the TBB library. If someone thinks this might be useful, please let me know - or add it to this PR! 😄

### DIFF
--- a/Examples/Framework/include/ActsExamples/Framework/Sequencer.hpp
+++ b/Examples/Framework/include/ActsExamples/Framework/Sequencer.hpp
@@ -13,6 +13,7 @@
 #include "ActsExamples/Framework/IReader.hpp"
 #include "ActsExamples/Framework/IService.hpp"
 #include "ActsExamples/Framework/IWriter.hpp"
+#include "ActsExamples/Utilities/tbbWrap.hpp"
 #include <Acts/Utilities/Logger.hpp>
 
 #include <cstddef>
@@ -21,8 +22,6 @@
 #include <string>
 #include <utility>
 #include <vector>
-
-#include <tbb/task_arena.h>
 
 namespace ActsExamples {
 
@@ -114,7 +113,7 @@ class Sequencer {
   std::pair<size_t, size_t> determineEventsRange() const;
 
   Config m_cfg;
-  tbb::task_arena m_taskArena;
+  tbbWrap::task_arena m_taskArena;
   std::vector<std::shared_ptr<IService>> m_services;
   std::vector<std::shared_ptr<IContextDecorator>> m_decorators;
   std::vector<std::shared_ptr<IReader>> m_readers;
--- a/Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
+++ b/Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
@@ -1,0 +1,162 @@
+// This file is part of the Acts project.
+//
+// Copyright (C) 2022 CERN for the benefit of the Acts project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+// uncomment to remove all use of tbb library.
+// #define NO_TBB
+
+// #include <iostream>
+
+#ifdef NO_TBB
+#include <stdexcept>
+#else
+#include <tbb/parallel_for.h>
+#include <tbb/queuing_mutex.h>
+#include <tbb/task_arena.h>
+#endif
+
+namespace ActsExamples {
+
+/// Wrapper for most of the tbb functions that we use in Sequencer.
+/// It disables the use of tbb if nthreads=1.
+/// Note that only a small subset of tbb functions are implemented, and
+/// tbb::blocked_range (which doesn't require any thread setup) is still taken
+/// from the tbb library.
+/// However, if NO_TBB is defined, then don't use tbb library at all
+/// (requires nthreads=1 or -1). This allows the ACTS Examples to be built
+/// without the tbb library (and reduces the dependency on ROOT).
+
+#ifdef NO_TBB
+namespace tbb {
+struct task_arena {
+  static const int automatic = -1;
+};
+
+template <typename Value>
+struct blocked_range {
+  blocked_range(Value begin_, Value end_) : my_end(end_), my_begin(begin_) {}
+  Value begin() const { return my_begin; }
+  Value end() const { return my_end; }
+
+ private:
+  Value my_end;
+  Value my_begin;
+};
+}  // namespace tbb
+#endif
+
+namespace tbbWrap {
+static bool enableTBB(int nthreads = -99) {
+  static bool setting = false;
+  if (nthreads != -99) {
+#ifdef NO_TBB
+    if (nthreads > 1) {
+      throw std::runtime_error(
+          "tbb is not supported, so can't do multi-threading.");
+    }
+#else
+    bool newSetting = (nthreads != 1);
+    if (newSetting != setting) {
+      if (newSetting) {
+        // std::cout << "Enable TBB" << std::endl;
+        setting = newSetting;
+      } else {
+        // std::cout << "Don't disable TBB, since it is already in use." <<
+        // std::endl;
+      }
+    }
+#endif
+  }
+  // std::cout << "TBB is " << (setting ? "enabled" : "disabled") << std::endl;
+  return setting;
+}
+
+class task_arena {
+#ifndef NO_TBB
+  std::unique_ptr<tbb::task_arena> tbb;
+#endif
+
+ public:
+  task_arena(int nthreads = tbb::task_arena::automatic,
+             [[maybe_unused]] unsigned dedicated = 1) {
+    if (enableTBB(nthreads)) {
+#ifndef NO_TBB
+      tbb.reset(new tbb::task_arena(nthreads, dedicated));
+#endif
+    }
+  }
+
+  template <typename F>
+  void execute(const F& f) {
+#ifndef NO_TBB
+    if (tbb) {
+      tbb->execute(f);
+    } else
+#endif
+    {
+      f();
+    }
+  }
+};
+
+class parallel_for {
+ public:
+  template <typename R, typename F>
+  parallel_for(const R& r, const F& f) {
+#ifndef NO_TBB
+    if (enableTBB()) {
+      tbb::parallel_for(r, f);
+    } else
+#endif
+    {
+      f(r);
+    }
+  }
+};
+
+class queuing_mutex {
+#ifndef NO_TBB
+  std::unique_ptr<tbb::queuing_mutex> tbb;
+#endif
+
+ public:
+  queuing_mutex() {
+#ifndef NO_TBB
+    if (enableTBB()) {
+      tbb.reset(new tbb::queuing_mutex());
+    }
+#endif
+  }
+
+  class scoped_lock {
+#ifndef NO_TBB
+    std::unique_ptr<tbb::queuing_mutex::scoped_lock> tbb;
+#endif
+
+   public:
+    scoped_lock() {
+#ifndef NO_TBB
+      if (enableTBB()) {
+        tbb.reset(new tbb::queuing_mutex::scoped_lock());
+      }
+#endif
+    }
+
+    scoped_lock([[maybe_unused]] queuing_mutex& m) {
+#ifndef NO_TBB
+      if (enableTBB()) {
+        tbb.reset(new tbb::queuing_mutex::scoped_lock(*m.tbb.get()));
+      }
+#endif
+    }
+  };
+};
+
+}  // namespace tbbWrap
+}  // namespace ActsExamples
--- a/Examples/Framework/src/Framework/Sequencer.cpp
+++ b/Examples/Framework/src/Framework/Sequencer.cpp
@@ -13,22 +13,30 @@
 #include "ActsExamples/Utilities/Paths.hpp"
 
 #include <algorithm>
+#include <atomic>
 #include <chrono>
 #include <exception>
 #include <numeric>
 
+#ifndef NO_TBB
 #include <TROOT.h>
+#endif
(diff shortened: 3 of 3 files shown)


Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
```diff
@@ -0,0 +1,169 @@
...
+/// However, if NO_TBB is defined, then don't use tbb library at all
+/// (requires nthreads=1 or -1). This allows the ACTS Examples to be built
+/// without the tbb library (and reduces the dependency on ROOT).
+/// Based on an idea from
+///   https://stackoverflow.com/questions/59736661/how-to-completely-switch-off-threading-in-tbb-code
+
+#ifdef NO_TBB
+namespace tbb {
+struct task_arena {
+  static const int automatic = -1;
```
HadrienG2, line 40: `constexpr`?

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
```diff
@@ -0,0 +1,169 @@
...
+/// Note that only a small subset of tbb functions are implemented, and
+/// tbb::blocked_range (which doesn't require any thread setup) is still taken
+/// from the tbb library.
+/// However, if NO_TBB is defined, then don't use tbb library at all
+/// (requires nthreads=1 or -1). This allows the ACTS Examples to be built
+/// without the tbb library (and reduces the dependency on ROOT).
+/// Based on an idea from
+///   https://stackoverflow.com/questions/59736661/how-to-completely-switch-off-threading-in-tbb-code
+
+#ifdef NO_TBB
```
HadrienG2, line 37: A way to set this from CMake would be desirable.

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
```diff
@@ -0,0 +1,169 @@
...
+#include <tbb/task_arena.h>
+#endif
+
+namespace ActsExamples {
+
+/// Wrapper for most of the tbb functions that we use in Sequencer.
+/// It disables the use of tbb if nthreads=1.
+/// Note that only a small subset of tbb functions are implemented, and
+/// tbb::blocked_range (which doesn't require any thread setup) is still taken
+/// from the tbb library.
```
HadrienG2, line 30: This seems outdated, the current code reimplements `tbb::blocked_range`.

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
```diff
@@ -0,0 +1,169 @@
...
+  Value my_end;
+  Value my_begin;
+};
+}  // namespace tbb
+#endif
+
+namespace tbbWrap {
+static bool enableTBB(int nthreads = -99) {
+  static bool setting = false;
+  if (nthreads != -99) {
```
HadrienG2, line 59: Maybe just check for >0, zero and negative numbers don't make sense.

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
```diff
@@ -0,0 +1,169 @@
...
+ private:
+  Value my_end;
+  Value my_begin;
+};
+}  // namespace tbb
+#endif
+
+namespace tbbWrap {
+static bool enableTBB(int nthreads = -99) {
+  static bool setting = false;
```
HadrienG2, line 58: This is not thread-safe. That's probably okay since I can't imagine a situation where the callee would be multi-threaded, but do document it in the docs of enableTBB and task_arena.

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
```diff
@@ -0,0 +1,169 @@
...
+        setting = newSetting;
+      } else {
+        // std::cout << "Don't disable TBB, since it is already in use." <<
+        // std::endl;
+      }
+    }
+#endif
+  }
+  // std::cout << "TBB is " << (setting ? "enabled" : "disabled") << std::endl;
+  return setting;
```
HadrienG2, line 79: Remove commented code.

If you agree with the proposal below of not wrapping TBB synchronization primitives, that will leave the `setting` static variable essentially unused, so consider removing this function entirely and replacing the enableTBB call of task_arena with a simple if (nthreads > 1).

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
```diff
@@ -0,0 +1,169 @@
...
+    scoped_lock(queuing_mutex&){
+#else
+    scoped_lock(queuing_mutex& m) {
+      if (enableTBB()) {
+        tbb.reset(new tbb::queuing_mutex::scoped_lock(*m.tbb.get()));
+      }
+#endif
+    }
+  };
+};
```
HadrienG2, line 166: I'm very uneasy about having lock classes that do nothing in some configurations.

To me, this raises the more philosophical question of whether we actually want to remove all uses of the TBB library, or just the parallelization part (task_arena and parallel_for). I would argue that the latter is sufficient to achieve your stated goal of simplifying debugging through direct calls and reduced thread spawning.

Also, if you want to stick with the current approach, consider using std::optional instead of std::unique_ptr.

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+#endif
+  }
+  // std::cout << "TBB is " << (setting ? "enabled" : "disabled") << std::endl;
+  return setting;
+}
+
+class task_arena {
+#ifndef NO_TBB
+  std::unique_ptr<tbb::task_arena> tbb;
```
HadrienG2, line 84: std::optional should be good enough here.

HadrienG2 (changes requested):

Examples/Framework/include/ActsExamples/Framework/Sequencer.hpp
```diff
@@ -13,6 +13,7 @@
 #include "ActsExamples/Framework/IReader.hpp"
 #include "ActsExamples/Framework/IService.hpp"
 #include "ActsExamples/Framework/IWriter.hpp"
+#include "ActsExamples/Utilities/tbbWrap.hpp"
```
paulgessinger, line 16: I think `TbbWrap.hpp` would be most aligned with our naming guidelines.

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
paulgessinger:
```suggestion
#ifdef ACTS_NO_TBB
```

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: I was mimicking what `tbb` does, but this could be `constexpr`, maybe better in a `namespace`, for us.

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: good idea

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: it only reimplements it with `NO_TBB`, which I mention next. I should clarify a bit further.

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: We use `-1` for `automatic` threads, and I think `-2` has a meaning too. I picked something unlikely to indicate the `enableTBB` default value.

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: good idea

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: Unfortunately it's also needed by `parallel_for`, which doesn't have direct access to the `task_arena`.

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: excellent idea

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: Sticking with `tbb::queuing_mutex` without wrapping it would certainly be cleaner. I wasn't sure whether it would work without initialising the rest of the `tbb` library, thread pool and everything.
Even if we can determine that's OK, would it play nice with ROOT where we don't call `ROOT::EnableThreadSafety()`? I really don't want to enable that in the unthreaded case, because that introduces even more confusions.

Examples/Framework/include/ActsExamples/Framework/Sequencer.hpp
timadye: I wanted to keep the lower-case `tbb` in the namespace name `tbbWrap` to make it easier to grep both, but if the naming convention is more important, I can change it.

Or were you suggesting to change just `TbbWrap.hpp`, and leaving `namespace tbbWrap`?

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: > A way to set this from CMake would be desirable.

I don't know how the library detection works, so I hope you guys can advise. I was thinking to have an `ACTS_USE_TBB` setting, which would default to `ON` if `ACTS_BUILD_EXAMPLES` is `ON` and the `tbb` library is found.

If `ACTS_USE_TBB=ON`, then it would link to the library. If it is `OFF`, then it would add `-DACTS_NO_TBB` to the compiler options.

Would this work? Is it overkill?

paulgessinger: Thinking about it, the option should in any case have `EXAMPLES` in the name, to make clear TBB is not used by the core library.

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: I just checked, and if I run with a single thread and don't do any ROOT I/O with the Generic detector, then ACTS doesn't need to initialise ROOT at all, so no longer forks anything. I don't think the ROOT stuff interferes much, but this is a nice side-effect.

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: done

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: added to comment

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: added comments about thread safety

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: removed commented code.

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: switched to `std::optional`

Examples/Framework/include/ActsExamples/Utilities/tbbWrap.hpp
timadye: CMake option `-DACTS_USE_EXAMPLES_TBB=off` will add `-DACTS_EXAMPLES_NO_TBB` to the compiler options and disable (direct) linking with the `tbb` library. I moved the CMake settings to `Examples/Framework`, since it's only directly used there. 

However I find that `libtbb.so.2` is still linked in, but I think that's because it's used in my versions of DD4hep and ROOT. I tested this works both with (v19.x) and without DD4hep.

HadrienG2 (approved):

paulgessinger (changes requested):
Can we add one python test that runs the Sequencer in explicitly in single thread mode? I think there should be a few that do it "by accident" but I'd like to make sure that we have one test case that explicitly covers this.

timadye:
> Can you add one python test that runs the Sequencer in explicitly in single thread mode? I think there should be a few that do it "by accident" but I'd like to make sure that we have one test case that explicitly covers this.

I added `test_sequencer_single_threaded` and `test_sequencer_multi_threaded`.

paulgessinger (approved):

Examples/Framework/include/ActsExamples/Framework/Sequencer.hpp
timadye: @paulgessinger, what do you think? Should I change `namespace TbbWrap` and/or `TbbWrap.hpp`, or leave it as it is?

PR #1448, opened 2022-08-19
refactor: python examples log level overwrite mechanism

andiwand:
I noticed that some python `add*` functions provide a mechanism to overwrite the `logLevel` but others don't which made me change `reconstruction.py` locally from time to time

I tried to generalize the existing functionality and extract it into a helper function. then I applied it to all the `add*` functions

after some [discussion](https://github.com/acts-project/acts/pull/1448#issuecomment-1225324331) I removed `dump_args_calls` from the `add*` functions as this can have side effects and break the CI


timadye (commented):
Thanks for this. It will be good to have a consistent interface. To that end, can you do the same for `itk.py` and `odd.py`, where relevant? Those have the potential to show a lot more detail, so we need to be sure to keep the default the same.

One problem is that the logic used to select between the different settings (`s.config.logLevel`, `add*(logLevel)`, and `customLogLevel(custom)`) is confusing and often doesn't do what we want. The original idea of the `custom` level was to allow different levels for some algorithms to match the original implementation (before we moved to `add*()`).

Note that `max(custom.value, logLevel.value)` chooses the *least* verbose of the two, so you can only make it less verbose than the default (usually `INFO`) with the `add*(logLevel)` argument.

@paulgessinger and I discussed this briefly on Tuesday, but didn't come to a conclusion. We need to think on this some more.

I don't think this PR changes the logic, so maybe we can accept it and leave improving the logic till later. Or should we try to fix it all now?

andiwand:
> Thanks for this. It will be good to have a consistent interface. To that end, can you do the same for itk.py and odd.py, where relevant? Those have the potential to show a lot more detail, so we need to be sure to keep the default the same.

I will take a look. the interesting thing there is that we do not have a sequencer which needs further generalization of the refactor

> One problem is that the logic used to select between the different settings (s.config.logLevel, add*(logLevel), and customLogLevel(custom)) is confusing and often doesn't do what we want. The original idea of the custom level was to allow different levels for some algorithms to match the original implementation (before we moved to add*()).

> Note that max(custom.value, logLevel.value) chooses the least verbose of the two, so you can only make it less verbose than the default (usually INFO) with the add*(logLevel) argument.

I feel like `custom` should always overrule if present - if not present we use `add*(logLevel)` - if not present we use `.config.logLevel`. Would that make sense?

> I don't think this PR changes the logic, so maybe we can accept it and leave improving the logic till later. Or should we try to fix it all now?

I am happy to include it if we can agree on something. otherwise I would go with a pure refactor that does not change the behavior first and later we can change the rest

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -8,6 +7,8 @@
 import acts
 import acts.examples
 
+from . import defaultLogging
```
paulgessinger, line 10: I think this is generally discouraged now.

andiwand:
@timadye I noticed that `acts.examples.dump_args_calls(locals())` will fail the CI somehow inside pytest. this didn't happen before because it was not called. do you know how this could happen and how to fix it? I will also try to look into this tomorrow

timadye:
> @timadye I noticed that `acts.examples.dump_args_calls(locals())` will fail the CI somehow when multiple pytests run in a row. this didn't happen before because it was not called. do you know how this could happen and how to fix it? I will also try to look into this tomorrow

Hi @andiwand, I never used that in the CI, just for debugging. I guess it's good to to also test `dump_args_calls` in the CI, so we should try to fix it. Where is it called?

andiwand:
> > @timadye I noticed that `acts.examples.dump_args_calls(locals())` will fail the CI somehow when multiple pytests run in a row. this didn't happen before because it was not called. do you know how this could happen and how to fix it? I will also try to look into this tomorrow
> 
> Hi @andiwand, I never used that in the CI, just for debugging. I guess it's good to to also test `dump_args_calls` in the CI, so we should try to fix it. Where is it called?

this is one of the CI jobs then fails https://github.com/acts-project/acts/runs/7975219427?check_suite_focus=true#step:8:4748. I tried to isolate this a little and it looks like removing the if [here](https://github.com/acts-project/acts/blob/main/Examples/Python/python/acts/examples/reconstruction.py#L194-L195) and just going with

```
acts.examples.dump_args_calls(locals())
```

will fail `pytest Examples/Python/tests/test_examples.py`

andiwand:
Hey @timadye ! I think I know what is going on: the problem is that after wrapping `isinstance` will not work anymore and this is used quite a lot in the python unit tests. before my changes the wrapping was not triggered in the unit test context but now they are and this is where it breaks.

Do you think it makes sense to move the wrapping out of `add*` as this has some side effects? Is it possible to have the call in the main script for example?

Another alternative might be to flag the call explicitly with a named argument in all the `add*` functions?

timadye:
> the problem is that after wrapping `isinstance` will not work anymore and this is used quite a lot in the python unit tests.

Yes, `isinstance` checks for wrapped classes won't work, unless you use `inspect.unwrap`, eg.
```
isinstance(mdecorator, inspect.unwrap(acts.IMaterialDecorator)
```
That's safe to do, whether or not the class is wrapped.

I only see a couple of `isinstance()` calls in `test_examples.py`, so it might be good to protect them. Does that help?

> Do you think it makes sense to move the wrapping out of `add*` as this has some side effects? Is it possible to have the call in the main script for example?

That's a good idea. Even if we can get pytest working with `dump_args_calls`, I think it would be still simpler not to use it during the testing. It's probably best to leave the `dump_args_calls` setting to the main program. The full_chain tests have them commented out.

andiwand:
great then this should be ready @timadye if you want to have another look at the changes!

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -185,14 +184,7 @@ def addSeeding(
...
     """
 
-    def customLogLevel(custom: acts.logging.Level = acts.logging.INFO):
-        """override logging level"""
-        if logLevel is None:
-            return s.config.logLevel
-        return acts.logging.Level(max(custom.value, logLevel.value))
-
-    if int(customLogLevel()) <= int(acts.logging.DEBUG):
-        acts.examples.dump_args_calls(locals())
```
timadye, line 195: As discussed previously, `dump_args_calls` is no longer called with `DEBUG` `logLevel`. I think this is the only change to existing functionality.
Please comment on this in the PR description.

Examples/Python/python/acts/examples/__init__.py
```diff
@@ -285,3 +286,39 @@ def _allmods(mod, base, found):
...
+
+
+class CustomLogLevel(Protocol):
+    def __call__(
+        self,
+        override: Optional[acts.logging.Level] = None,
+        min: acts.logging.Level = acts.logging.VERBOSE,
+        max: acts.logging.Level = acts.logging.FATAL,
+    ) -> acts.logging.Level:
+        ...
```
timadye, line 298: Sorry, I'm not familiar with the use of a `Protocol` class. Is this just to keep the argument checker happy for the `defaultLogging` definition?
If so, do the `__call__` argument names need to match the `customLogLevel` argument names?

andiwand: this is also new to me. the idea was to have a type for `customLogLevel` which is returned by the `defaultLogging` function. my hope was also that vscode would pick that up and give you hints about `customLogLevel`. at least for me that did not work but I kept it to define the function

Examples/Python/python/acts/examples/odd.py
```diff
@@ -47,7 +56,7 @@ def getOpenDataDetector(odd_dir, mdecorator=None):
         mdecorator = acts.JsonMaterialDecorator(
             rConfig=config,
             jFileName=str(odd_dir / "config/odd-material-mapping-config.json"),
-            level=acts.logging.WARNING,
+            level=customLogLevel(minLevel=acts.logging.WARNING),
```
tboldagh, line 59: Just a stupid question. Why sometimes the level is called maxLevel and sometimes minLevel ?

Examples/Python/python/acts/examples/__init__.py
```diff
@@ -285,3 +286,39 @@ def _allmods(mod, base, found):
...
+    s=None,
+    logLevel: Optional[acts.logging.Level] = None,
+) -> CustomLogLevel:
+    """
+    Establishes a default logging strategy for the python examples interface.
+
+    Returns a function that determines the log level in the following schema:
+    - if `overrideLevel` for `customLogLevel` is set return it
+    - if `logLevel` is set use it other wise use the log level of the sequencer `s.config.logLevel`
+    - the returned log level is bound between `min` and `max` provided to `customLogLevel`
```
tboldagh, line 311: Maybe a snippet of examples would be useful here:
i.e. if you want all components to produce only warnings or more severe errors do that:...
if you want switch all components to INFO do that ...

Examples/Python/python/acts/examples/odd.py
andiwand: I wanted to make it more explicit compared to the existing `customLogLevel` (e.g. [here](https://github.com/acts-project/acts/blob/main/Examples/Python/python/acts/examples/reconstruction.py#L875)) if you want to have a lower or upper bound on the log level. not sure how useful it is. if we think we can live without it we can also drop it

Examples/Python/python/acts/examples/__init__.py
andiwand: good point will do

Examples/Python/python/acts/examples/__init__.py
paulgessinger: I think you could use 
```python
Callable[[Optional[acts.logging.Level], acts.logging.Level, acts.logging.Level], acts.logging.Level]
```
as the type?

paulgessinger:
This seems like a good idea to me. Can we get this in @andiwand @timadye?

Examples/Python/python/acts/examples/__init__.py
andiwand: my problem with this was that I could not express what the different parameters mean. but I guess from the vscode point of view it does not make a difference because the `Protocol` is not inspected right now anyway

andiwand: an alternative would be to make a class out of `defaultLogging` and make it callable. maybe IDEs could handle that better. on the other hand a class might look like we are wrapping the logger

andiwand:
> This seems like a good idea to me. Can we get this in @andiwand @timadye?

if you could take another look @timadye that would be great! from my side it looks okay now

Examples/Python/python/acts/examples/__init__.py
```diff
@@ -285,3 +286,43 @@ def _allmods(mod, base, found):
...
         print("dump_args for module functions:", ", ".join(donemods))
     return alldone
+
+
+class CustomLogLevel(Protocol):
+    def __call__(
+        self,
+        override: Optional[acts.logging.Level] = None,
+        min: acts.logging.Level = acts.logging.VERBOSE,
+        max: acts.logging.Level = acts.logging.FATAL,
```
timadye, line 296:
```suggestion
        minLevel: acts.logging.Level = acts.logging.VERBOSE,
        maxLevel: acts.logging.Level = acts.logging.FATAL,
```
I don't know if the _self_ should be in there. Is that a feature of how this is defined?

andiwand: I am not sure about this but this is how I found it on the web. it kind of made sense because the callable might use some state in the general case

Examples/Python/python/acts/examples/__init__.py
```diff
@@ -285,3 +286,43 @@ def _allmods(mod, base, found):
...
+    - the returned log level is bound between `minLevel` and `maxLevel` provided to `customLogLevel`
+
+    Examples:
+    - `customLogLevel(minLevel=acts.logging.INFO)` to get a log level that is INFO or higher
+      (depending on the sequencer and `logLevel` param) which is useful to suppress a component which
+      produces a bunch of logs below INFO and you are actually more interested in another component
+    - `customLogLevel(maxLevel=acts.logging.INFO)` to get a log level that is INFO or lower
+      (depending on the sequencer and `logLevel` param) which is useful to get more details from a
+      component that will produce logs of interest below the default level
+    """
```
timadye, line 318: This is a good explanation, but let me raise a long-standing concern (not specific to your changes).

I've been wondering if we need to be clearer about the potential confusion from *lower* `logLevel` giving *more* output. That's not a problem, until we compare > or < to a `minLevel` or `maxLevel`.
Put another way, the `minLevel` defines the maximum amount of logging; the `maxLevel` defines the minimum amount of logging.
Unless someone can think of a better argument name, I think the best we can do is document it. Do you think it's worth adding a comment to this effect, or maybe the nice new examples are enough?

timadye:
Otherwise looks good to me. When we have resolved the question about `minLevel` and `maxLevel` (just docs, I think), I think we can approve. Should I press the button?

Examples/Python/python/acts/examples/__init__.py
andiwand: I agree this is quite confusing and I always need to think twice about it :smile: 

I don't see an easy way out of this except better doc as you said.

maybe instead of min max which introduces numbers for comparison with > or < we could rephrase this `minLevel` -> `ignoreBelowLevel` and `maxLevel` -> `requrieAboveLevel`

timadye: > we could rephrase this `minLevel` -> `ignoreBelowLevel` and `maxLevel` -> `requireAboveLevel`
I want to like this, but to me it is perhaps even more confusing. But don't take my word on it, perhaps it is clearer to others.

andiwand: in that case I would just default to the min/max for now and we can improve it at any point

timadye: yeah. Maybe add a sentence to the docs explaining this?

Examples/Python/python/acts/examples/__init__.py
```diff
@@ -285,3 +286,42 @@ def _allmods(mod, base, found):
...
```
andiwand, line 318:
```suggestion
    Examples:
    - `customLogLevel(minLevel=acts.logging.INFO)` to get a log level that is INFO or higher
      (depending on the sequencer and `logLevel` param) which is useful to suppress a component which
      produces a bunch of logs below INFO and you are actually more interested in another component
    - `customLogLevel(maxLevel=acts.logging.INFO)` to get a log level that is INFO or lower
      (depending on the sequencer and `logLevel` param) which is useful to get more details from a
      component that will produce logs of interest below the default level
    - in summary `minLevel` defines the maximum amount of logging and `maxLevel` defines the minimum amount of logging
    """
```

timadye (approved):
Looks good to me. Sorry for all the niggling comments.

acts-project-service:
```
```
```bash
git fetch
# Create a new working tree
# Create a new branch
git switch --create backport-1448-to-develop/v19.x
git cherry-pick -x --mainline 1 433a379bf53912a5ef3d515ebb23f2bf3d1af331
# Push it to GitHub
git push --set-upstream origin backport-1448-to-develop/v19.x
cd ../..
# Delete the working tree
```
Then, create a pull request where the `base` branch is `develop/v19.x` and the `compare`/`head` branch is `backport-1448-to-develop/v19.x`.

PR #1452, opened 2022-08-19
feat: Identifier can be customized for sensitive surfaces

paulgessinger:
This PR reduces the width of the sensitive part of the geometry identifier from 28bits to 20bits, to make room for an 8bit *extra* identifier part. The idea is that this can be used by experiments to label sensitive elements. The bit masks after the change look like:

```cpp
static constexpr Value kVolumeMask    = 0xff00000000000000; // (2^8)-1 = 255 volumes
static constexpr Value kBoundaryMask  = 0x00ff000000000000; // (2^8)-1 = 255 boundaries
static constexpr Value kLayerMask     = 0x0000fff000000000; // (2^12)-1 = 4096 layers
static constexpr Value kApproachMask  = 0x0000000ff0000000; // (2^8)-1 = 255 approach surfaces
static constexpr Value kSensitiveMask = 0x000000000fffff00; // (2^20)-1 sensitive surfaces
static constexpr Value kExtraMask     = 0x00000000000000ff; // (2^8)-1 extra values
```

In addition, it adds a hook function to the geometry closure of signature

```cpp
GeometryIdentifier(GeometryIdentifier orig, const Surface& surface);
```

to customize the geometry identifier. Technically, this hook can be used to also reset the volume and layer ids, but is primarily intended to set the *extra* components.


Core/include/Acts/Geometry/GeometryIdentifier.hpp
```diff
@@ -74,18 +78,20 @@ class GeometryIdentifier {
...
-  static constexpr Value kApproachMask = 0x0000000ff0000000;
-  // (2^28)-1 sensitive surfaces
-  static constexpr Value kSensitiveMask = 0x000000000fffffff;
+  // clang-format off
+  static constexpr Value kVolumeMask    = 0xff00000000000000; // (2^8)-1 = 255 volumes
+  static constexpr Value kBoundaryMask  = 0x00ff000000000000; // (2^8)-1 = 255 boundaries
+  static constexpr Value kLayerMask     = 0x0000fff000000000; // (2^12)-1 = 4096 layers
+  static constexpr Value kApproachMask  = 0x0000000ff0000000; // (2^8)-1 = 255 approach surfaces
+  static constexpr Value kSensitiveMask = 0x000000000fffff00; // (2^20)-1 sensitive surfaces
+  static constexpr Value kExtraMask     = 0x00000000000000ff; // (2^8)-1 extra values
```
andiwand, line 93: should we add the explicit number of combinations here too?

```suggestion
  static constexpr Value kSensitiveMask = 0x000000000fffff00; // (2^20)-1 = 1048575 sensitive surfaces
  static constexpr Value kExtraMask     = 0x00000000000000ff; // (2^8)-1 = 255 extra values
```

Core/include/Acts/Geometry/GeometryIdentifier.hpp
```diff
@@ -124,6 +130,9 @@ class GeometryIdentifier {
 
 std::ostream& operator<<(std::ostream& os, GeometryIdentifier id);
 
+using GeometryIdentifierHook =
+    std::function<GeometryIdentifier(GeometryIdentifier, const Surface&)>;
```
andiwand, line 134: I like it :+1: apart from changing the identifier it should also be easy to build a mapping between acts and experiment IDs

Core/include/Acts/Geometry/GeometryIdentifier.hpp
```diff
@@ -53,6 +55,8 @@ class GeometryIdentifier {
   constexpr Value approach() const { return getBits(kApproachMask); }
   /// Return the sensitive identifier.
   constexpr Value sensitive() const { return getBits(kSensitiveMask); }
+  /// Return the extra identifier
+  constexpr Value extra() const { return getBits(kExtraMask); }
```
andiwand, line 59: should we state possible use cases somewhere?

one thing that comes to my mind is tagging surfaces of interest e.g. something in the endcaps that can easily be filtered while construction

Tests/UnitTests/Core/Geometry/TrackingGeometryClosureTests.cpp
```diff
@@ -140,9 +151,113 @@ BOOST_AUTO_TEST_CASE(GeometryIdentifier_closeGeometry_test) {
...
+      GeometryIdentifier::Value layer_id = 0;
+      for (auto lay : vol.confinedLayers()->arrayObjects()) {
+        // check the layer volume id and layer layer id
+        auto lay_vol_id = lay->geometryId().volume();
+        auto lay_lay_id = lay->geometryId().layer();
+        BOOST_CHECK_EQUAL(++layer_id, lay_lay_id);
+        BOOST_CHECK_EQUAL(geoid, lay_vol_id);
+        // test the layer approach surfaces
+        if (lay->approachDescriptor() != nullptr) {
+          // approach surfacesare counted from 1 - n
```
andiwand, line 229:
```suggestion
          // approach surfaces are counted from 1 - n
```

andiwand (approved):
I like it :+1:

asalzburger:
So, if we had lets say 20 surfaces front size and 20 surfaces backside, would you do 0-20 and a side bit, but 0-40 and a side bit ?

Core/include/Acts/Geometry/GeometryIdentifier.hpp
paulgessinger: Sure, before it was hundreds of millions, so I didn't bother

Core/include/Acts/Geometry/GeometryIdentifier.hpp
asalzburger: > So, if we had lets say 20 surfaces front size and 20 surfaces backside, would you do 0-20 and a side bit, but 0-40 and a side bit ?

Answered in person: second one will keep the current behaviour ... Good.

asalzburger (approved):
Good thing.

paulgessinger:
I addressed @andiwand's comments. Can one of you approve again?

andiwand (approved):

paulgessinger:
Hmm. I didn't see hash failures locally...

paulgessinger:
The fast-test right after FATRAS only already fails. I'll have to investigate. I'm taking this WIP again.

Core/include/Acts/Geometry/GeometryIdentifier.hpp
```diff
@@ -74,18 +80,20 @@ class GeometryIdentifier {
...
-  // (2^12)-1 = 4096 layers
-  static constexpr Value kLayerMask = 0x0000fff000000000;
-  // (2^8)-1 = 255 approach surfaces
-  static constexpr Value kApproachMask = 0x0000000ff0000000;
-  // (2^28)-1 sensitive surfaces
-  static constexpr Value kSensitiveMask = 0x000000000fffffff;
+  // clang-format off
+  static constexpr Value kVolumeMask    = 0xff00000000000000; // (2^8)-1 = 255 volumes
+  static constexpr Value kBoundaryMask  = 0x00ff000000000000; // (2^8)-1 = 255 boundaries
+  static constexpr Value kLayerMask     = 0x0000fff000000000; // (2^12)-1 = 4096 layers
```
AJPfleger, line 92:
```suggestion
  static constexpr Value kLayerMask     = 0x0000fff000000000; // (2^12)-1 = 4095 layers
```

paulgessinger:
Ok I think the hash changes come from the fact that we write the actual geometry id 64bit integer value to the outputs, which clearly changes if we reduce the bit width of the sensitive id.

andiwand (approved):

PR #1454, opened 2022-08-20
fix: stop CKF branch if particle is stopped

andiwand:
in some special cases particles will be stopped while propagation in the CKF due to high energy losses. in this PR this is caught and the CKF branch will be dropped.

additionally the error `NoTrackFound` was removed and CKF will only warn about that from now on.

the abortlist was added to the CKF result because this is the only mutable state that is accessible. after a discussion with Paul we concluded that "result" is just a state and was designed for that anyways.

see https://github.com/acts-project/acts/issues/1385#issuecomment-1217968896


Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -254,6 +254,9 @@ struct CombinatorialKalmanFilterResult {
   bool finished = false;
 
   Result<void> result{Result<void>::success()};
+
+  // TODO place into options and make them accessible?
+  AbortList<PathLimitReached, EndOfWorldReached, ParticleStopped> abortList;
```
andiwand, line 259: do you see a way to put this into the options @paulgessinger ?

AJPfleger (approved):
lgtm

PR #1458, opened 2022-08-22
refactor: Improve material mapping speed

Corentin-Allaire:
This PR improve the speed of the material mapping : 
- Improve the root material track reader, root used to loop over all the entry to determine the number of track per events. Solving this improve the mapping speed by a factor 30 for the average map with enough input events.
- If the material need to be map multiple time in a row (to optimise the binning for example) it can now be done faster with the `mat-mapping-read-surfaces` option. The intersection of the material track and the surfaces are now stored in the input root file after the first mapping so that computing it can be skipped on subsequent mapping (up to 50% faster).
  
To allow this speed up the `SurfaceMaterialMapper` has been restructured : 
- Changing the structure to work with input file where the surface has already been associated with the material. 
- Avoid averaging bin multiple time (this was causing a bug in the variance computation)
- Add the information on the associated surface to the material interaction (so that it can be available in the next iteration)


Core/src/Material/SurfaceMaterialMapper.cpp
```diff
@@ -199,8 +199,33 @@ void Acts::SurfaceMaterialMapper::finalizeMaps(State& mState) const {
 
 void Acts::SurfaceMaterialMapper::mapMaterialTrack(
     State& mState, RecordedMaterialTrack& mTrack) const {
-  using VectorHelpers::makeVector4;
+  // Retrieve the recorded material from the recorded material track
+  auto& rMaterial = mTrack.second.materialInteractions;
+  ACTS_VERBOSE("Retrieved " << rMaterial.size()
+                            << " recorded material steps to map.")
 
+  // Check if the material interaction are associated with a surface. If yes we
```
AJPfleger, line 207:
```suggestion
  // Check if the material interactions are associated with a surface. If yes we
```

Core/include/Acts/Material/SurfaceMaterialMapper.hpp
```diff
@@ -170,6 +171,24 @@ class SurfaceMaterialMapper {
...
+  ///
+  void mapInteraction(State& mState, RecordedMaterialTrack& mTrack) const;
+
+  /// Loop through all the material interaction and add them to associated
+  /// surface
+  ///
+  /// @param mState The current state map
+  /// @param rMaterial Vector of all the material interaction that will be mapped
+  ///
+  /// @note The material interaction are assumed to have an associated surface ID
```
AJPfleger, line 188:
```suggestion
  /// Loop through all the material interactions and add them to the
  /// associated surface
  ///
  /// @param mState The current state map
  /// @param rMaterial Vector of all the material interactions that will be mapped
  ///
  /// @note The material interactions are assumed to have an associated surface ID
```

Core/include/Acts/Material/SurfaceMaterialMapper.hpp
```diff
@@ -170,6 +171,24 @@ class SurfaceMaterialMapper {
   /// to be ordered from the starting position along the starting direction
   void mapMaterialTrack(State& mState, RecordedMaterialTrack& mTrack) const;
 
+  /// Loop through all the material interaction and add them to associated
+  /// surface
```
AJPfleger, line 175:
```suggestion
  /// Loop through all the material interactions and add them to the
  /// associated surface
```

Core/src/Material/SurfaceMaterialMapper.cpp
AJPfleger, line 207:
```suggestion
  // Check if the material interactions are associated with a surface. If yes we
```

Core/src/Material/SurfaceMaterialMapper.cpp
```diff
@@ -199,8 +199,33 @@ void Acts::SurfaceMaterialMapper::finalizeMaps(State& mState) const {
...
+  // simply need to loop over them and accumulate the material
+  if (rMaterial.begin()->intersectionID != GeometryIdentifier()) {
+    ACTS_VERBOSE(
+        "Material surfaces are associated with the material interaction. The "
+        "association interaction/surfaces won't be performed again");
```
AJPfleger, line 212:
```suggestion
        "association interaction/surfaces won't be performed again.");
```

Core/src/Material/SurfaceMaterialMapper.cpp
```diff
@@ -199,8 +199,33 @@ void Acts::SurfaceMaterialMapper::finalizeMaps(State& mState) const {
...
+    mapSurfaceInteraction(mState, rMaterial);
+    return;
+  } else {
+    ACTS_VERBOSE(
+        "Material interaction need to be associated with surfaces. Collecting "
+        "all surface on the trajectory");
```
AJPfleger, line 218:
```suggestion
        "all surfaces on the trajectory.");
```

Core/src/Material/SurfaceMaterialMapper.cpp
```diff
@@ -415,10 +439,62 @@ void Acts::SurfaceMaterialMapper::mapMaterialTrack(
...
 }
+
+void Acts::SurfaceMaterialMapper::mapSurfaceInteraction(
+    State& mState, std::vector<MaterialInteraction>& rMaterial) const {
+  using MapBin = std::pair<AccumulatedSurfaceMaterial*, std::array<size_t, 3>>;
+  std::map<AccumulatedSurfaceMaterial*, std::array<size_t, 3>> touchedMapBins;
+  std::map<AccumulatedSurfaceMaterial*, std::shared_ptr<const ISurfaceMaterial>>
+      touchedMaterialBin;
+
+  // Looping over all the material interaction
```
AJPfleger, line 465:
```suggestion
  // Looping over all the material interactions
```

Core/src/Material/SurfaceMaterialMapper.cpp
```diff
@@ -415,10 +439,62 @@ void Acts::SurfaceMaterialMapper::mapMaterialTrack(
...
+  auto rmIter = rMaterial.begin();
+  while (rmIter != rMaterial.end()) {
+    // get the current interaction informations
```
AJPfleger, line 468:
```suggestion
    // get the current interaction information
```

Examples/Algorithms/MaterialMapping/include/ActsExamples/MaterialMapping/MaterialMappingOptions.hpp
```diff
@@ -40,7 +40,13 @@ void addMaterialMappingOptions(aopt_t& opt) {
       "mat-mapping-volume-stepsize",
       po::value<float>()->default_value(std::numeric_limits<float>::infinity()),
       "Step size of the sampling of volume material for the mapping "
-      "(should be smaller than the size of the bins in depth)");
+      "(should be smaller than the size of the bins in depth)")(
+      "mat-mapping-read-surfaces",
+      po::value<bool>()->default_value(po::value<bool>()->default_value(false)),
+      "Read the surface associated with each material hit, can be used to "
+      "speed up the mapping. "
+      "The mapping need to have been performed at least once for the surface "
```
AJPfleger, line 48:
```suggestion
      "The mapping needs to have been performed at least once for the surface "
```

Examples/Io/Root/src/RootMaterialTrackReader.cpp
```diff
@@ -63,11 +68,18 @@ ActsExamples::RootMaterialTrackReader::RootMaterialTrackReader(
...
 
   m_events = m_inputChain->GetMaximum("event_id") + 1;
-  ACTS_DEBUG("The full chain has " << m_events << " entries.");
+  size_t nentries = m_inputChain->GetEntries();
+  m_batch = nentries / m_events;
+  ACTS_DEBUG("The full chain has "
+             << nentries << " entries for " << m_events
+             << " events this correspond to a batch size of : " << m_batch);
+  std::cout << "The full chain has " << nentries << " entries for " << m_events
+            << " events this correspond to a batch size of : " << m_batch
```
AJPfleger, line 77:
```suggestion
             << " events this corresponds to a batch size of: " << m_batch);
  std::cout << "The full chain has " << nentries << " entries for " << m_events
            << " events this corresponds to a batch size of: " << m_batch
```

Examples/Io/Root/src/RootMaterialTrackReader.cpp
```diff
@@ -164,6 +171,17 @@ ActsExamples::ProcessCode ActsExamples::RootMaterialTrackReader::read(
             Acts::Material::fromMassDensity(mX0, mL0, (*m_step_A)[is],
                                             (*m_step_Z)[is], (*m_step_rho)[is]),
             s);
+        if (m_cfg.readSurface) {
+          // add the surface information to to the interaction this allows the
```
AJPfleger, line 175:
```suggestion
          // add the surface information to the interaction this allows the
```

docs/examples/howto/run_material_mapping.rst
```diff
@@ -132,6 +133,9 @@ In addition to root and JSON output, one can also output the material map to a C
 .. note::
   You can map onto surfaces and volumes separately (for example if you want to optimise first one then the other). In that case after mapping one of those you will need to use the resulting JSON material map as an input to the ``mat-input-file``.
 
+.. note::
+  You might need to run the material mapping multiple time in a row to optimise the binning for each surface, which can be time-consuming. To improve the speed of subsequent mapping one can turn the ``mat-mapping-read-surfaces`` option to true, with this option the intersection of the material track with the surfaces will be taken directly from the input root file (which would have been updated during the first mapping). This will save the time needed to compute them. This can save up to 50% of the mapping time. Please note that this can only be used if the mapping surfaces used didn't change between this execution and the last.
```
AJPfleger, line 137:
```suggestion
  You might need to run the material mapping multiple times in a row to optimise the binning for each surface, which can be time-consuming. To improve the speed of subsequent mapping one can turn the ``mat-mapping-read-surfaces`` option to true, with this option the intersection of the material track with the surfaces will be taken directly from the input root file (which would have been updated during the first mapping). This will save the time needed to compute them. This can save up to 50% of the mapping time. Please note that this can only be used if the mapping surfaces used didn't change between this execution and the last.
```

Corentin-Allaire:
@AJPfleger Thanks for the spelling check !

Core/src/Material/SurfaceMaterialMapper.cpp
```diff
@@ -415,10 +439,62 @@ void Acts::SurfaceMaterialMapper::mapMaterialTrack(
           missedMaterial->second.trackVariance(
               mSurface.position,
               mState.inputSurfaceMaterial[currentID]->materialSlab(
-                  mSurface.position));
+                  mSurface.position),
+              true);
         }
         missedMaterial->second.trackAverage(mSurface.position, true);
+
+        // Add an empty material hit for futur material mapping iteration
```
paulgessinger, line 447:
```suggestion
        // Add an empty material hit for future material mapping iteration
```

Core/src/Material/SurfaceMaterialMapper.cpp
```diff
@@ -372,14 +390,20 @@ void Acts::SurfaceMaterialMapper::mapMaterialTrack(
     // Now assign the material for the accumulation process
     auto tBin = currentAccMaterial->second.accumulate(
         currentPos, rmIter->materialSlab, currentPathCorrection);
-    auto itBin =
-        touchedMapBins.insert(MapBin(&(currentAccMaterial->second), tBin));
+    if (touchedMapBins.count(&(currentAccMaterial->second)) == 0) {
```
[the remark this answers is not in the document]
paulgessinger: So previously it was recording them every time even if already in `touchedMapBins`?

Core/src/Material/SurfaceMaterialMapper.cpp
```diff
@@ -415,10 +439,62 @@ void Acts::SurfaceMaterialMapper::mapMaterialTrack(
...
+  while (rmIter != rMaterial.end()) {
+    // get the current interaction information
+    GeometryIdentifier currentID = rmIter->intersectionID;
+    Vector3 currentPos = rmIter->intersection;
+    auto currentAccMaterial = mState.accumulatedMaterial.find(currentID);
+
+    // Now assign the material for the accumulation process
+    auto tBin = currentAccMaterial->second.accumulate(
+        currentPos, rmIter->materialSlab, rmIter->pathCorrection);
+    if (touchedMapBins.count(&(currentAccMaterial->second)) == 0) {
```
paulgessinger, line 476: Consider `find()`

Examples/Io/Root/include/ActsExamples/Io/Root/RootMaterialTrackReader.hpp
```diff
@@ -38,6 +38,8 @@ class RootMaterialTrackReader : public IReader {
 
     /// Whether the events are ordered or not
     bool orderedEvents = true;
+    // Read surface information for the root file
+    bool readSurface = false;
```
paulgessinger, line 42:
```suggestion
    bool readCachedSurfaceInformation = false;
```

Examples/Io/Root/include/ActsExamples/Io/Root/RootMaterialTrackReader.hpp
```diff
@@ -79,6 +81,9 @@ class RootMaterialTrackReader : public IReader {
   /// The number of events
   size_t m_events = 0;
 
+  /// The batch size (number of track per events)
+  size_t m_batch = 0;
```
paulgessinger, line 85:
```suggestion
  size_t m_batchSize = 0;
```

Examples/Scripts/Python/material_recording.py
```diff
@@ -87,4 +90,4 @@ def runMaterialRecording(g4geo, outputDir, tracksPerEvent=10000, s=None):
     )
     g4geo = acts.examples.geant4.dd4hep.DDG4DetectorConstruction(dd4hepSvc)
 
-    runMaterialRecording(g4geo=g4geo, tracksPerEvent=10, outputDir=os.getcwd()).run()
+    runMaterialRecording(g4geo=g4geo, tracksPerEvent=100, outputDir=os.getcwd()).run()
```
paulgessinger, line 93: I'm guessing with the changes from this PR, the runtime of the example doesn't increase drastically with this?

Examples/Python/tests/root_file_hashes.txt
```diff
@@ -42,11 +42,11 @@ test_propagation__propagation_steps.root: b5da12754f97b67dce6bf85a7345481ec5bebc
...
-test_material_mapping__material-map_tracks.root: 80beaaf4b9acee9eef6f2bb2094d1c819edb07f22e949833a0dc4143a697ad32
-test_material_mapping__propagation-material.root: 15d5dfd89583b965ec5a2625126a190b4af92f21b4f2f9ce2946e98bf8dff34a
-test_volume_material_mapping__material-map-volume_tracks.root: 14815e3f42a64c140450302f06a4381ede147ee4edbb8a88f3dfac7e7ab53fa7
-test_volume_material_mapping__propagation-volume-material.root: 47b488e258ca4c964ba9f68a65e4ea576bc096f2c8476b1363b9f98b5cc96820
+test_material_recording__geant4_material_tracks.root: 69d3b13e800615f38fcfef9bf2c6ac1ec799f2d3edba62954be668fa44282cf1
+test_material_mapping__material-map_tracks.root: efe7ce049e1b5257f37485f6788c3e67e83bb3ca987dd0e0bc34edb9cc2a6c0d
+test_material_mapping__propagation-material.root: e3677926877a57a87cb6db59cf6e3016146e5c7baf4b67fd65424a070502bc65
+test_volume_material_mapping__material-map-volume_tracks.root: ecd77adfe5421218df55d07842591f2721e9e130ad55e71bb0da94977a74dfe7
+test_volume_material_mapping__propagation-volume-material.root: e3a65793321cc2ec901bce83df068ef5e364bf5091beec9a3f80b5ab6f8c5d87
```
paulgessinger, line 49: I assume you checked that the changed outputs look reasonable?

paulgessinger (commented):
Can you elaborate a bit more what the changes in algorithm structure are? It's not clear to me how the changes to `SurfaceMaterialMapper` relate to the changes to mention in the PR description.

I see you remove averaging of empty bins, are there other optimizations?

Core/src/Material/SurfaceMaterialMapper.cpp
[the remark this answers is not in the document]
Corentin-Allaire: Yes and no we would add the same bin multiple time to touchedMapBins. This wasn't really an issue because it just meant we would run the averaging on that bin multiple time with one hit (effectively doing nothing except waist time). But it was causing some issue in the newly added variance computation and was a waste of time so I fixed it.

Examples/Python/tests/root_file_hashes.txt
Corentin-Allaire: Yes and I have been running the material mapping with the new change for a while. They give the same result as before. The difference here come from the extra information I added to the root file.

Examples/Scripts/Python/geometry.py
```diff
@@ -72,6 +72,7 @@ def runGeometry(
                 processRepresenting=True,
                 processBoundaries=True,
                 processVolumes=True,
+                processNonMaterial=True,
```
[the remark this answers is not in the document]
Corentin-Allaire: Yes that flag has always existed. If you want to use geometry.py for the material mapping this flag need to be here and it was use in the equivalent example (WriteTrackingGeometry). I thought it would be good to add this here in preparation of moving the mapping doc to the Python bindings

Examples/Scripts/Python/material_recording.py
Corentin-Allaire: The pytest for this example set those value so it shouldn't affect it. I changed this in preparation of moving the mapping doc to the python binding (those value are what you need for a good map). Since people often ask me how many event they need.

Core/src/Material/SurfaceMaterialMapper.cpp
```diff
@@ -199,8 +199,33 @@ void Acts::SurfaceMaterialMapper::finalizeMaps(State& mState) const {
...
     State& mState, RecordedMaterialTrack& mTrack) const {
-  using VectorHelpers::makeVector4;
+  // Retrieve the recorded material from the recorded material track
+  auto& rMaterial = mTrack.second.materialInteractions;
+  ACTS_VERBOSE("Retrieved " << rMaterial.size()
+                            << " recorded material steps to map.")
 
+  // Check if the material interactions are associated with a surface. If yes we
+  // simply need to loop over them and accumulate the material
+  if (rMaterial.begin()->intersectionID != GeometryIdentifier()) {
```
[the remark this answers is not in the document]
Corentin-Allaire: `intersectionID` would only different from `GeometryIdentifier()` if the root reader was used with the `readSurface` option. This mean that surface information has been associated with the material interaction and this would be the same across the entire root file. Basically I use this as a proxy to check if the root reader was used with `readSurface=true`.

Corentin-Allaire:
> Can you elaborate a bit more what the changes in algorithm structure are? It's not clear to me how the changes to `SurfaceMaterialMapper` relate to the changes to mention in the PR description.
> 
> I see you remove averaging of empty bins, are there other optimizations?

The main change to the `SurfaceMaterialMapper` are : 
- Changing the structure to work with input file where the surface has already been associated with the material. 
- Avoid averaging bin multiple time (as you have notice)
- Add the information on the associated surface to the material interaction (so that it can be available in the next iteration)

I will add those info to the description of the PR.

Core/include/Acts/Material/MaterialInteraction.hpp
```diff
@@ -34,6 +35,10 @@ struct MaterialInteraction {
   double sigmaTheta2 = 0.0;
   /// Expected q/p variance due to the interactions.
   double sigmaQoP2 = 0.0;
+  /// The position where the interaction occured.
+  Vector3 intersection = Vector3(0., 0., 0);
+  /// The ID where the interaction occured.
+  GeometryIdentifier intersectionID;
```
asalzburger, line 41: Don't you get the geoId from the surface?

asalzburger (approved):
Looks fine from my side - so the caching does the trick here.

paulgessinger (approved):
Let's go.

PR #1459, opened 2022-08-22
feat: Algorithm lifecycle methods: initialize & finalize

paulgessinger:
While working on #1417, I was missing a way for an algorithm to collect performance numbers while running (thread-safe) and printing them when finalizing. I believe this is going to be more and more important, so this PR is adding simple `initialize` and `finalize` methods to our algorithm interface. They are defaulted by `BareAlgorithm` so all other algorithms don't necessarily have to implement.

This PR uses this mechanism to add some output on the number of failed seeds to the end of the TrackFindingAlgorithm, but there should be more cases where this will be helpful.

### DIFF
--- a/Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/TrackFindingAlgorithm.hpp
+++ b/Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/TrackFindingAlgorithm.hpp
@@ -15,8 +15,10 @@
 #include "ActsExamples/EventData/Measurement.hpp"
 #include "ActsExamples/EventData/Track.hpp"
 #include "ActsExamples/Framework/BareAlgorithm.hpp"
+#include "ActsExamples/Framework/ProcessCode.hpp"
 #include "ActsExamples/MagneticField/MagneticField.hpp"
 
+#include <atomic>
 #include <functional>
 #include <vector>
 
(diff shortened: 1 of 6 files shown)


Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/TrackFindingAlgorithm.hpp
```diff
@@ -90,8 +92,13 @@ class TrackFindingAlgorithm final : public BareAlgorithm {
   void computeSharedHits(const source_link_accessor_container_t& sourcelinks,
                          TrackFinderResult&) const;
 
+  ActsExamples::ProcessCode finalize() const override;
+
  private:
   Config m_cfg;
+
+  mutable std::atomic<size_t> m_nTotalSeeds;
+  mutable std::atomic<size_t> m_nFailedSeeds;
```
paulgessinger, line 101: Before I merge this, let me see if @HadrienG2 kills me over this. 😅

HadrienG2: Nah, that makes total sense :)

If you really want me to nitpick, you don't really need seq_cst semantics here, so fetch_add(1, std::memory_order_relaxed) would be a more efficient alternative to ++. But I don't think the code path you're running those increments is anywhere near hot enough to justify this sort of fine tuning.

PR #1460, opened 2022-08-22
feat: Update RootVertexPerformanceWriter to work with track finding

paulgessinger:
Previously, the `RootVertexPerformanceWriter` assumed a 1:1 correspondence between truth particles and reconstructed tracks. This PR implements hit based truth matching to enable this to run.

### DIFF
--- a/Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
+++ b/Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
@@ -51,10 +51,16 @@ ActsExamples::RootVertexPerformanceWriter::RootVertexPerformanceWriter(
     throw std::invalid_argument(
         "Collection with selected truth particles missing");
   }
-  if (m_cfg.inputAssociatedTruthParticles.empty()) {
+  if (m_cfg.inputAssociatedTruthParticles.empty() &&
+      (m_cfg.inputFittedTracksIndices.empty() ||
+       m_cfg.inputAllFittedTracksTips.empty() ||
+       m_cfg.inputTrajectories.empty())) {
     throw std::invalid_argument(
-        "Collection with track-associated truth particles missing");
+        "You need to either provide collection of truth particles matching 1:1 "
+        "to tracks, or track indices and all-tips container to do truth "
+        "matching");
   }
+
   if (m_cfg.inputFittedTracks.empty()) {
     throw std::invalid_argument(
         "Collection with all fitted track parameters missing");
@@ -75,6 +81,20 @@ ActsExamples::RootVertexPerformanceWriter::RootVertexPerformanceWriter(
     m_outputTree->Branch("diffx", &m_diffx);
     m_outputTree->Branch("diffy", &m_diffy);
     m_outputTree->Branch("diffz", &m_diffz);
+
+    m_outputTree->Branch("recoX", &m_recoX);
+    m_outputTree->Branch("recoY", &m_recoY);
+    m_outputTree->Branch("recoZ", &m_recoZ);
+
+    m_outputTree->Branch("truthX", &m_truthX);
+    m_outputTree->Branch("truthY", &m_truthY);
+    m_outputTree->Branch("truthZ", &m_truthZ);
+
+    m_outputTree->Branch("covXX", &m_covXX);
+    m_outputTree->Branch("covYY", &m_covYY);
+    m_outputTree->Branch("covXY", &m_covXY);
+    m_outputTree->Branch("covYX", &m_covYX);
+    m_outputTree->Branch("trkVtxMatch", &m_trackVtxMatchFraction);
     m_outputTree->Branch("nRecoVtx", &m_nrecoVtx);
     m_outputTree->Branch("nTrueVtx", &m_ntrueVtx);
     m_outputTree->Branch("nVtxDetectorAcceptance", &m_nVtxDetAcceptance);
@@ -181,23 +201,6 @@ ActsExamples::ProcessCode ActsExamples::RootVertexPerformanceWriter::writeT(
   ACTS_INFO("Total number of detector-accepted truth primary vertices : "
             << m_nVtxDetAcceptance);
 
-  // Read track-associated truth particle input collection
-  const auto& associatedTruthParticles =
-      ctx.eventStore.get<SimParticleContainer>(
-          m_cfg.inputAssociatedTruthParticles);
-  // Get number of track-associated true primary vertices
-  m_nVtxReconstructable =
-      getNumberOfReconstructableVertices(associatedTruthParticles);
-
-  ACTS_INFO("Total number of reco track-associated truth particles in event : "
-            << associatedTruthParticles.size());
-  ACTS_INFO("Total number of reco track-associated truth primary vertices : "
-            << m_nVtxReconstructable);
-
-  /*****************  Start x,y,z resolution plots here *****************/
-  // Matching tracks at vertex to fitted tracks that are in turn matched
-  // to truth particles. Match reco and true vtx if >50% of tracks match
-
   const auto& inputFittedTracks =
       ctx.eventStore.get<std::vector<Acts::BoundTrackParameters>>(
           m_cfg.inputFittedTracks);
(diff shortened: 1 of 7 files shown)


Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -934,17 +936,26 @@ def customLogLevel(custom: acts.logging.Level = acts.logging.INFO):
...
             )
+        kwargs = {}
+        if trajectories is not None:
+            kwargs["inputTrajectories"] = "trajectories"
+        else:
+            kwargs["inputAssociatedTruthParticles"] = associatedParticles
         s.addWriter(
             RootVertexPerformanceWriter(
-                level=customLogLevel(),
+                level=acts.logging.VERBOSE,
```
andiwand, line 946: should we keep this verbose?

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -934,17 +936,26 @@ def customLogLevel(custom: acts.logging.Level = acts.logging.INFO):
                 "Using RootVertexPerformanceWriter with smeared particles is not necessarily supported. "
                 "Please get in touch with us"
             )
+        kwargs = {}
+        if trajectories is not None:
+            kwargs["inputTrajectories"] = "trajectories"
+        else:
+            kwargs["inputAssociatedTruthParticles"] = associatedParticles
```
andiwand, line 943: I wondered if this would be cleaner to pass in the `RootVertexPerformanceWriter` with `"trajectories" if trajectories is not None else ""`

Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
```diff
@@ -181,36 +201,189 @@ ActsExamples::ProcessCode ActsExamples::RootVertexPerformanceWriter::writeT(
...
+        << associatedTruthParticles.size());
+    ACTS_INFO("Total number of reco track-associated truth primary vertices : "
+              << m_nVtxReconstructable);
+
+    /*****************  Start x,y,z resolution plots here *****************/
+    // Matching tracks at vertex to fitted tracks that are in turn matched
+    // to truth particles. Match reco and true vtx if >50% of tracks match
+
+    if (associatedTruthParticles.size() != inputFittedTracks.size()) {
+      ACTS_INFO(
```
andiwand, line 231: would an error be more appropriate here?

Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
```diff
@@ -181,36 +201,189 @@ ActsExamples::ProcessCode ActsExamples::RootVertexPerformanceWriter::writeT(
...
+    std::vector<ParticleHitCount> particleHitCounts;
+
+    using HitParticlesMap = IndexMultimap<ActsFatras::Barcode>;
+    const auto& hitParticlesMap =
+        ctx.eventStore.get<HitParticlesMap>(m_cfg.inputMeasurementParticlesMap);
+
+    SimParticleContainer associatedTruthParticles;
+
+    for (size_t i = 0; i < inputFittedTracks.size(); i++) {
+      // ACTS_VERBOSE("i=" << i);
```
andiwand, line 343: remove or un-comment? (two more below)

Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
```diff
@@ -181,36 +201,189 @@ ActsExamples::ProcessCode ActsExamples::RootVertexPerformanceWriter::writeT(
...
+              // Next vertex now
+              break;
+            }
+          }
+        }
+      }  // end loop vertices
+    }
+  } else if (!m_cfg.inputFittedTracksIndices.empty() &&
+             !m_cfg.inputAllFittedTracksTips.empty() &&
+             !m_cfg.inputTrajectories.empty()) {
```
andiwand, line 320:
```suggestion
  } else {
```

as this should be guaranteed by the constructor?

Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
```diff
@@ -270,12 +445,29 @@ ActsExamples::ProcessCode ActsExamples::RootVertexPerformanceWriter::writeT(
...
             // Next vertex now
             break;
           }
         }
       }
     }  // end loop vertices
+
+  } else {
+    throw std::invalid_argument{"Invalid truth input configuration"};
   }
```
andiwand, line 471:
```suggestion
  }
```

if we remove the `else if` above

Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
```diff
@@ -181,36 +201,189 @@ ActsExamples::ProcessCode ActsExamples::RootVertexPerformanceWriter::writeT(
...
+        }  // end loop tracks
+
+        // Now find true vtx with most matching tracks at reco vtx
+        // and check if it contributes more than 50 of all tracks
+        std::map<int, int> fmap;
+        for (int priVtxId : contributingTruthVertices) {
+          fmap[priVtxId]++;
+        }
+        int maxOccurrenceId = -1;
+        int maxOccurence = -1;
```
andiwand, line 269:
```suggestion
        int maxOccurrence = -1;
```

Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
```diff
@@ -181,36 +201,189 @@ ActsExamples::ProcessCode ActsExamples::RootVertexPerformanceWriter::writeT(
...
+        for (auto it : fmap) {
+          if (it.second > maxOccurence) {
+            maxOccurence = it.second;
```
andiwand, line 272:
```suggestion
            maxOccurrence = it.second;
```

Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
```diff
@@ -181,36 +201,189 @@ ActsExamples::ProcessCode ActsExamples::RootVertexPerformanceWriter::writeT(
...
+
+              int priVtxId = particle.particleId().vertexPrimary();
+              contributingTruthVertices.push_back(priVtxId);
+            }
+            idx++;
+          }
+        }  // end loop tracks
+
+        // Now find true vtx with most matching tracks at reco vtx
+        // and check if it contributes more than 50 of all tracks
```
andiwand, line 263:
```suggestion
        // and check if it contributes more than 50% of all tracks
```

andiwand (approved):

Examples/Python/python/acts/examples/reconstruction.py
paulgessinger: Probably not.

Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
paulgessinger: Now that it supports the other case, maybe yeah.

Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
paulgessinger: Yeah true probably not necessary.

Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
paulgessinger: Ok I checked again, and this still triggers in the case of <1 track efficiency e.g. in truth tracking. I added conditional ERROR or INFO messages.

Examples/Python/python/acts/examples/reconstruction.py
paulgessinger: Changed.

Examples/Python/python/acts/examples/reconstruction.py
paulgessinger: Changed.

Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
paulgessinger: Removed

paulgessinger:
Ok I fixed one source of inconsistency resulting in output changes, and then checked that the remaining output changes we see here only come from the newly added branches.

I'll let the CI run and update the hashes, and then I think this is good to go.

andiwand (approved):

andiwand (approved):

PR #1465, opened 2022-08-24
feat: experimental geometry - surfaces, portals and volumes

asalzburger:
This PR is introducing an attempt of an alternative geometry, which is built only from 
 * surface objects
 * portals (containing surface objects)
 * detector volumes

It uses the `Delegate<>` mechanism to delegate the local navigation to user-definable functions. The code is so far all concentrated in an `Experimental` folder.  There's only a tiny modification needed in the `VolumeBounds` class in order to return `non-const` surfaces.

This geometry building also implements a strict `non-cost` at construction and `const` at runtime policy.


asalzburger:
This sits on top of the SVG-plugin update.

@paulgessinger @andiwand  FYI

I have not yet done anything with the `ObjectStore<T>` as we discussed (i.e. changing to iterator access).

asalzburger:
Closing, superseded.

Issue #1456, opened 2022-08-22
Bug: SpacePointGridCreator::createGrid() fails when minHelixRadius < config.rMax / 2


askamkin:
If `minHelixRadius < config.rMax / 2`, createGrid (Core/include/Acts/Seeding/SpacePointGrid.ipp) takes the square root of a negative number:
```
float maxR2 = config.rMax * config.rMax;
float xOuter = maxR2 / (2 * minHelixRadius);
float yOuter = std::sqrt(maxR2 - xOuter * xOuter);
```
It seems strange that the method does not work for detectors w/ some configurations. Perhaps the method functionality should be revisited.

I suggest adding a check for this situation w/ a proper error message.

paulgessinger:
Hey @askamkin, thanks for the issue.

@LuisFelipeCoelho, can you comment on this?

LuisFelipeCoelho:
I can add a check for this situation with an error message.

I am also adding @robertlangenberg here in case he has any suggestions.

robertlangenberg:
Not allowing smaller radii is sensible, as otherwise the particle would not escape the detector (i.e. reach rMax). Of course, it should have been checked for in the code. Thank you for notifying us!

PR #1466, opened 2022-08-24
feat: check if yOuter takes the square root of a negative number when minHelixRadius < rMax / 2 in SpacePointGrid.ipp

LuisFelipeCoelho:
This PR adds a check to `SpacePointGrid.ipp` if `minHelixRadius` is smaller than `config.rMax / 2` as suggested in #1456.

This happens if `rMax` of the grid in `mm` is greater than `2 x pT[GeV] / (3·10-4 × B[T])`. I am not sure if this combination of parameters makes sense in one specific detector configuration.

@robertlangenberg


robertlangenberg (approved):
This is sensible, as otherwise the particle would not escape the detector (i.e. reach rMax). Thank you for fixing this.

PR #825, opened 2021-06-02
refactor!: B field access returns Result

paulgessinger:
Previously, we didn't have a mechanism to communicate that a field query failed. This is problematic, because it's not clear what to do when for instance our interpolated field goes out of bounds (see #784). This PR makes `MagneticFieldProvider` require the `getField` and `getFieldGradient` methods to return `Result<Vector3>` instead of the bare vector. This changes the public interface slightly, and all clients need to be adjusted.

Additionally, the `InterpolatedBFieldMap` now checks each time a field cell is created or if the field is queried directly, whether the query position is out of bounds of the map, and will return a new `MagneticFieldError::OutOfBounds` if that happens.

BREAKING CHANGE:
- The signature of field query methods in `MagneticFieldProvider` changes from
   ```cpp
   virtual Vector3 getField(const Vector3& position, Cache& cache) const = 0;
   virtual Vector3 getFieldGradient(const Vector3& position, ActsMatrix<3, 3>& derivative, Cache& cache) const = 0;
   ```
   to
   ```cpp
   virtual Result<Vector3> getField(const Vector3& position, Cache& cache) const = 0;
   virtual Result<Vector3> getFieldGradient(const Vector3& position, ActsMatrix<3, 3>& derivative, Cache& cache) const = 0;
   ```
- `InterpolatedBFieldMap::getMin` and `InterpolatedBFieldMap::getMax` now return the extent of the valid **interpolation domain**, rather than the raw grid extent.

Issue #821, opened 2021-05-31
How do we do B field interpolation?


paulgessinger:
Right now, `Grid` is used to store and retrieve B field values to interpolate between. `Grid` can be queried for neighbor indices for a given position, and will only return indices that corresponds to cells that exist, i.e. no neighbors that would be behind the edge of the map. 

Both `Grid`'s internal `interpolate` method as well as `FieldCell`, however, assume that there will always be a full set of neighbor indices. This appears to have caused UB related to #784. 

The question is, what do we do here? I think we have the following options:

1. Out-of-bounds B field values are `(0,0,0)`. This is easy to implement, but leads to wrong B field values in the border region of a field map.
2. Clamp to the last valid value. `Grid` can be configured to return the last valid bin value for out of bounds query positions. This doesn't affect the neighbor calculation however, those will still not include over-the-edge bins. This can be implemented, but is more difficult as we have to change the neighbor algorithm.
3. We error out if a field map is queried in the border region. This is very explicit, and probably the most correct approach.

Do we care about the exact validity of the interpolation at the grid borders? I think we don't as it's not well defined what we should do anyway. This doesn't really come up often as our field maps are typically larger than the tracking geometries, but apparently happens in case of the solenoid field map (#784). 

I would probably go for either option 1 or 3, with a slight preference for option 3.

/cc @asalzburger @HadrienG2

HadrienG2:
I agree with your choices. If it doesn't introduce unsufferable overhead in this performance-critical part, proper error reporting is best. If it does introduce overhead, then returning nonsensical values is fine, since this code path should never be hit by non-buggy clients anyway.

paulgessinger:
I need to add a loop (very likely unrolled) that does one `<` comparison between the local indices and the maximum number of indices. So the overhead should be negligible I think. I need to change all the interfaces to return our `Result` object though, but that should be ok I believe.

HadrienG2:
Note that if Result is too costly, exceptions are another error reporting possibility that lies somewhere inbetween Result and silly output on the performance vs correctness continuum.

paulgessinger:
I would prefer `Result` because it forces clients to handle errors.

Issue #784, opened 2021-04-27
bug: solenoid magnetic field (defaults) in Examples looks corrupt


asalzburger:
When running with the new `bf-solenoid` option I get many errors like this:

```sh
12:49:19    FatrasAlgori   ERROR     event 94 particle id=(1|0|382|0|0)|pdg=mu-|q=-1|m=0.105658|p=368.2 failed to simulate with error EigenStepperError:3: Step size adjustment exceeds maximum trials
```
I am running with this option:
```sh
--bf-solenoid-mag-tesla 2
```

None of these warnings appear either with the constant field nor with the ATLAS magnetic field map.

paulgessinger:
This seems to only occur if the log level is not `DEBUG`. Suboptimal...

paulgessinger:
For reference, I think I can reproduce this with

```console
bin/ActsExampleFatrasGeneric --bf-solenoid-mag-tesla 2 --skip 36623 -n1 -j1
```

paulgessinger:
This is WEIRD as heck. When I run 

```console 
bin/ActsExampleFatrasGeneric --bf-solenoid-mag-tesla 2 --skip 36623 -n1 -j1 -l0
```

the final bit of output looks like this:

```
14:54:54    Simulation     VERBOSE   Step with size = 65.8949 performed
14:54:54    Simulation     VERBOSE   Target: 0 | Target stepSize (path limit) updated to (263.579,  +∞, 4166.31,  +∞ )
14:54:54    Simulation     VERBOSE   Step with size = 263.579 performed
14:54:54    Simulation     VERBOSE   Target: 0 | Target stepSize (path limit) updated to (1054.32,  +∞, 3902.73,  +∞ )
14:54:54    Simulation     VERBOSE   Step with size = 1054.32 performed
14:54:54    Simulation     VERBOSE   Target: 0 | Target stepSize (path limit) updated to (4217.27,  +∞, 2848.41,  +∞ )
14:54:54    Simulation     VERBOSE   Step with size = 2848.41 performed
14:54:54    Simulation     VERBOSE   Target: x | Path limit reached at distance 0
14:54:54    Simulation     VERBOSE   Stepping loop done.
14:54:54    FatrasAlgori   DEBUG     1 simulated particles (initial state)
14:54:54    FatrasAlgori   DEBUG     1 simulated particles (final state)
14:54:54    FatrasAlgori   DEBUG     13 simulated hits
[...]

PR #1467, opened 2022-08-25
fix: Remove incomplete initialisation from InterpolatedBFieldMap

timadye:
* Remove unneeded/misleading partial initialisation.
   * Note `std::array<Vector3, 8> neighbors{Vector3{0.0, 0.0, 0.0}};` only initialises the first element of the `std::array`.
* `assert` if `neighbors` `array` not fully filled.

Originally "**fix: B-field interpolation errors**", but this PR no longer provides a fix to the following problem.
* Remove `InterpolatedBFieldMap` boundary check. Instead, use the under/overflow bins to return the B-field outside the grid.
  * Was generating ~259 errors per event (`full_chain_itk.py` ttbar+PU200), with each error generating 2 error messages (one quite verbose) and 4 warnings.
  * We don't need to propagate these track candidates further, but we should leave that decision up to the caller and not flood the logfile.


timadye:
Label WiP because it may take some time for deal with the fallout from a unit test failure. The ATLAS B-field mostly works fine because the under/overflow bins all have B=(0,0,0).
The unit test expects the error, so can be fixed by expecting `ok`. But adding some tests for B-field outside the range hit issues with the `Grid` methods that mostly do not support the user/overflow bins. In particular, the neighboring bin finding doesn't work for underflow bins (can't return bin=-1 in an unsigned int, so doesn't initialise the array of neighbors). Also the interpolation doesn't do extrapolation.
I think I can work round these issues and improve the unit test for these cases. But maybe this isn't the right way to go about it. It seems to me that `InterpolatedBFieldMap` should be able to cope with requests outside its range, either by returning B=(0,0,0) or the edge values. The former would be simpler, but the latter fits the binned map model.

paulgessinger:
Thanks first of all.

I don't think the magnetic field interpolation should fail silently in this way, to be honest. If the particle propagation goes out of bounds of the interpolation domain, that's a serious issue. It either means that your field map is not sufficiently large or there's a navigation error for some reason.

I would personally prefer encouraging users to fix either the geometry (if that's the source of the navigation issues), fix a bug in the navigation, or explicitly extent the interpolation domain.

Thoughts?

timadye:
> I don't think the magnetic field interpolation should fail silently in this way, to be honest.

`InterpolatedBFieldMap` defines underflow and overflow bins, so there is a clear definition for values outside the range. So I think it is a matter of interpretation whether this is a failure or just a limited domain of implementation. Other field map types (constant, solenoid) don't impose a limited domain except where they make no sense (eg. R<0).

> If the particle propagation goes out of bounds of the interpolation domain, that's a serious issue.

For me, "if it goes out of the bounds of the detector, that's a serious issue". The magnetic field goes on forever! (even if it is 0). In this case, it *may* have identified an error, but that's using the field map to cross-check the propagation. Perhaps useful, but a bit ugly.

Sorry for the philosophy. For my bug (#1484), I'd be happy with anything that gets rid of the GBytes of logfile messages.

paulgessinger:
While the InterpolatedBField indeed has under and overflow values, but we had issues in the past where it went out of bounds and people didn't notice and only after non-negligible debugging did we figure out that was the case.
I'll try to dig up the issues.

paulgessinger:
Got it, it's here https://github.com/acts-project/acts/issues/784 and then I introduced this error mode in https://github.com/acts-project/acts/pull/825

andiwand:
makes sense @paulgessinger but if we do not intend to extrapolate the B field at all the overflow functionality sounds kind of useless to me?

if we want to have a defined B fields in case of overflow the more appropriate point to check for navigation failures would be in the navigator IMO. we could require an end of world bounds in the config and check at each iteration if we stepped into the void and error there which would be more explicit

paulgessinger:
We use the overflow functionality of Grid in the SurfaceArray. If we can make it optional, that would be a possibility.

To your other point: fully agree that ideally the navigation should catch this. Isn't that what the `EndOfWorld` aborter is supposed to do?

andiwand:
you are right `EndOfWorld` does the job usually only CKF sneaks in because you cannot use standard aborters there. but I guess that is a detail of CKF and has to be dealt with there specifically

timadye:
> Got it, it's here #784 and then I introduced this error mode in #825.
> I also had some discussion with @HadrienG2 about this in #821.

Sorry, I didn't know that this had already been widely discussed and the current behavior was (fairly) recently introduced. I might have argued for your options 1 or 2, but not sufficiently strongly to warrant reopening the issue.

I notice that #825 introduced the initialisation in [InterpolatedBFieldMap.hpp line 197](https://github.com/acts-project/acts/blob/main/Core/include/Acts/MagneticField/InterpolatedBFieldMap.hpp#L197):
```
    std::array<Vector3, nCorners> neighbors{Vector3{0.0, 0.0, 0.0}};
````
I think that's a mistake, which I tried to correct in this PR (not yet quite right there). It only initialises the first element of `neighbors`. The array will be filled later, as long as the `closestPointsIndices` iterator yields `nCorners` values. I think it should always do that, except in the `MagneticFieldError::OutOfBounds` case (actually, IIRC, I found only in the lower or left bounds, while the upper and right bounds still give the correct number of indices).

Anyway, I think it would be faster to remove the initialisation, or perhaps safer to use `std::array::fill` to fill all elements. Either would be better than a deeply misleading first-value initialisation. I checked on Godbolt, and `fill` (with `Vector3{0,0,0}`) after declaration is just as fast as initialising all elements, but much easier to do for `constexpr nCorners`.

While we are looking here, I see that there may be an optimisation opportunity to replace the `isInsideLocal`, which compares the coordinate position with the `double` bounds, with comparisons on the `indices`. I don't know if that's also safer in edge cases.

paulgessinger:
Hey @timadye, that sounds fair. I think I didn't realize back then that this only initializes the first value.

I guess we can just drop the initialization if it's basically guaranteed to be initialized later on.

andiwand:
how do we move forward with this @paulgessinger @timadye ?

andiwand (approved):

PR #1469, opened 2022-08-26
perf: Remove x87 elliptic integral in solenoid B-field

stephenswat:
As it stands, the solenoid B-field code uses the `boost::math::ellint_1` and `boost::math::ellint_2` methods. For reasons which I do not understand, these functions use 80-bit x87 floating point numbers, which leads to a lot of conversion between x87 and SSE floating points. It also has the downside that x87 operations are generally slower then their SSE counterparts on modern CPUs.

The interesting thing is that Boost's method for deciding on the floating point precision seems to be quite arbitrary; even when the
functions are explicitly templated to use 64-bit `double` types, the internals of these functions still use 80-bit floating points.

The best solution I have found is to use a compiler preprocessor flag, `BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS`, which tells Boost not to use x87 floating points at all. Enabling this flag only in the solenoid B-field translation unit should be safe and improve performance. Very unscientifically benchmarked on my Intel i5-7300U CPU, the solenoid B-field benchmark with the old code takes 615.7±7.3µs per iteration, while the new code takes 353.7±7.0µs, a speedup of about 75%.


AJPfleger (approved):
nice! lgtm

paulgessinger:
Seeing this only now, I'm guessing you checked that the solenoid field output is approximately compatible @stephenswat? (not identical of course)

acts-project-service:
```
```
```bash
git fetch
# Create a new working tree
# Create a new branch
git switch --create backport-1469-to-develop/v19.x
git cherry-pick -x --mainline 1 3795148974ec2682f84218b7f8c14d1f3be121c1
# Push it to GitHub
git push --set-upstream origin backport-1469-to-develop/v19.x
cd ../..
# Delete the working tree
```
Then, create a pull request where the `base` branch is `develop/v19.x` and the `compare`/`head` branch is `backport-1469-to-develop/v19.x`.

PR #1470, opened 2022-08-26
feat: Add a tool for writing B-fields to disk in CSV format

stephenswat:
This PR adds a new example tool, `CsvBFieldWriter`, that dumps a B-field to disk in CSV format. The tool supports both grid-based and non-grid-based vector fields, it supports Cartesian and symmetrical cylindrical coordinates, and it supports user-specified ranges and bin counts.

Comes with free Python bindings.

### DIFF
--- a/Examples/Io/Csv/include/ActsExamples/Io/Csv/CsvBFieldWriter.hpp
+++ b/Examples/Io/Csv/include/ActsExamples/Io/Csv/CsvBFieldWriter.hpp
@@ -1,0 +1,94 @@
+// This file is part of the Acts project.
+//
+// Copyright (C) 2022 CERN for the benefit of the Acts project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+#include "Acts/MagneticField/InterpolatedBFieldMap.hpp"
+#include "Acts/MagneticField/MagneticFieldProvider.hpp"
+#include "Acts/Utilities/Logger.hpp"
+
+#include <array>
+#include <optional>
+
+namespace ActsExamples {
+/// @brief Writer for B-fields that outputs field data in CSV format.
+///
+/// This tool allows users to dump their magnetic field data to disk in CSV
+/// format, which is (relatively) human-readable and widely supported.
+class CsvBFieldWriter {
+ public:
+  /// @brief Enumeration type for the coordinate system to use in the writing
+  /// process.
+  ///
+  /// This enumeration sets Cartesian coordinates in the XYZ case, and
+  /// symmetrical cylindrical coordinates in the RZ case.
+  enum class CoordinateType { XYZ, RZ };
+
+  /// @brief Configuration type for the magnetic field writer.
+  ///
+  /// @tparam Coord Sets the coordinate system to use: either Cartesian or
+  /// cylindrical.
+  /// @tparam Grid If true, assume that the magnetic field is grid-based and
+  /// that it has a built-in bin count and size. Otherwise, assume it is
+  /// analytical.
+  template <CoordinateType Coord, bool Grid>
+  struct Config {
+    /// Helper type constructor that makes values optional in grid-based
+    /// environments. The reason for this is that grid-based fields have
+    /// built-in bins and sizes, and as such it is not strictly necessary for
+    /// the user to supply them. For gridless fields this does not work, and the
+    /// data must become mandatory.
+    template <typename T>
+    using Functor = std::conditional_t<Grid, std::optional<T>, T>;
+
+    /// @brief Dimensionality of vectors.
+    ///
+    /// Set the number of scalars in our coordinates to 3 if we have 3D
+    /// Cartesian coordinates, or 2 for cylindrical.
+    static constexpr std::size_t NDims = Coord == CoordinateType::RZ ? 2 : 3;
+
+    /// @brief Output file to write the magnetic field to.
+    std::string fileName = "bfield.csv";
+
+    /// @brief The range over which to dump the vector field.
+    ///
+    /// For each dimension, contains two optional values: the minimum and
+    /// maximum in that dimension. If any value is non-extant, the gridded
+    /// field's internal contents are used as a default.
+    std::array<std::array<Functor<double>, 2>, NDims> range;
+
+    /// @brief Number of bins in the output vector field.
+    ///
+    /// The same logic as above holds for the ranges, use the user-supplied
+    /// value if available, otherwise use the value from the grid.
+    std::array<Functor<std::size_t>, NDims> bins;
+
+    /// @brief Magnetic field to read from.
+    ///
+    /// Note that the type of this field is different from grid-based and
+    /// non-grid-based fields: @class Acts::InterpolatedMagneticField and
+    /// @class Acts::MagneticFieldProvider, respectively.
+    std::shared_ptr<std::add_const_t<std::conditional_t<
+        Grid, Acts::InterpolatedMagneticField, Acts::MagneticFieldProvider>>>
+        bField;
+  };
+
+  ///@brief Write magnetic field data to a CSV file.
+  ///
+  ///@tparam Coord Specifies the coordinate system to use while writing; this is
+  /// either Cartesian (XYZ) or cylindrical (RZ).
+  ///@tparam Grid Boolean specifying whether the magnetic field is a grid and has
+  /// set size, or whether it is analytical.
+  ///
+  ///@param config Configuration of the writing job.
+  ///@param level Log level to use during execution.
+  template <CoordinateType Coord, bool Grid>
+  static void run(const Config<Coord, Grid>& config,
+                  Acts::Logging::Level level);
+};
+}  // namespace ActsExamples
--- a/Examples/Io/Csv/src/CsvBFieldWriter.cpp
+++ b/Examples/Io/Csv/src/CsvBFieldWriter.cpp
@@ -1,0 +1,240 @@
+// This file is part of the Acts project.
+//
+// Copyright (C) 2022 CERN for the benefit of the Acts project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#include "ActsExamples/Io/Csv/CsvBFieldWriter.hpp"
+
+#include "Acts/Definitions/Algebra.hpp"
+#include "Acts/MagneticField/BFieldMapUtils.hpp"
+#include "Acts/MagneticField/MagneticFieldContext.hpp"
+#include "Acts/Utilities/Helpers.hpp"
+#include "Acts/Utilities/Logger.hpp"
+#include "Acts/Utilities/Result.hpp"
+
+#include <dfe/dfe_io_dsv.hpp>
+
+namespace ActsExamples {
+template <CsvBFieldWriter::CoordinateType Coord, bool Grid>
+void CsvBFieldWriter::run(const Config<Coord, Grid>& config,
+                          Acts::Logging::Level level) {
+  ACTS_LOCAL_LOGGER(Acts::getDefaultLogger("CsvBFieldWriter", level));
+
+  // Some helper typedefs, which will make our life easier down the line.
+  using ConfigType = std::decay_t<decltype(config)>;
+  using FieldType = typename decltype(ConfigType::bField)::element_type;
+  using Vector = Acts::ActsVector<ConfigType::NDims>;
+
+  FieldType& field = *config.bField;
+
+  // Determine the columns to write to the CSV file. For Cartesian coordinates,
+  // we give the x, y, and z coordinates for both the position and the field
+  // vector. For cylindrical coordinates, we give r and z.
+  std::vector<std::string> fields;
+
+  if constexpr (Coord == CoordinateType::XYZ) {
+    fields = {"x", "y", "z", "Bx", "By", "Bz"};
+  } else {
+    fields = {"r", "z", "Br", "Bz"};
+  }
+
+  // Initialize a CSV writer to the specified filename using the specified
+  // column names.
+  dfe::io_dsv_impl::DsvWriter<','> writer(fields, config.fileName);
+
+  // We proceed by finding the number of bins, as well as the minimum and
+  // maximum coordinates. This process depends quite heavily on the structure
+  // of the magnetic field, so we need some compile-time conditionals.
+  std::array<std::size_t, ConfigType::NDims> bins;
+  Vector min, max;
+
+  if constexpr (Grid) {
+    // First, we should check whether the Bfield actually has a
+    // three-dimensional bin count and size.
+    if (bins.size() != ConfigType::NDims ||
+        field.getMin().size() != ConfigType::NDims ||
+        field.getMax().size() != ConfigType::NDims) {
+      throw std::invalid_argument("Incorrect input B-field dimensionality.");
+    }
+
+    // If our magnetic field is grid-like, we know that it has a built-in size
+    // and bin count. We can use these, but the user may want to override the
+    // values with custom values.
+    for (std::size_t i = 0; i < ConfigType::NDims; ++i) {
+      // For each dimension, we first get the corresponding value from the
+      // grid-based vector field...
+      bins[i] = field.getNBins()[i];
+      min[i] = field.getMin()[i];
+      max[i] = field.getMax()[i];
+
+      // ...and then we override them with the optional user-supplied values.
+      if (config.bins[i]) {
+        bins[i] = *config.bins[i];
+      }
+
+      if (config.range[i][0]) {
+        min[i] = *config.range[i][0];
+      }
+
+      if (config.range[i][1]) {
+        max[i] = *config.range[i][1];
+      }
+    }
+  } else {
+    // If the vector field is not grid based, then there is no side and we are
+    // forced to use the user-supplied data. Remember that in this case, the
+    // values are not optional.
+    for (std::size_t i = 0; i < ConfigType::NDims; ++i) {
+      bins[i] = config.bins[i];
+      min[i] = config.range[i][0];
+      max[i] = config.range[i][1];
+    }
+  }
+
+  // Next, we calculate the size (in physical space) of each bin.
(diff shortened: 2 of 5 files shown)


andiwand (approved):
looks great :+1:

Examples/Io/Csv/src/CsvBFieldWriter.cpp
```diff
@@ -0,0 +1,261 @@
...
+              << (bins[2] + 1) << " to file " << config.fileName << "...");
+
+    std::size_t total_items = (bins[0] + 1) * (bins[1] + 1) * (bins[2] + 1);
+
+    /*
+     * For Cartesian coordinates, iterate over bins in the x, y, and z
+     * directions. Note that we iterate one additional time because we are
+     * writing the _edges_ of the bins, and the final bin needs to be closed.
+     */
+    for (std::size_t x = 0; x <= bins[0]; ++x) {
```
andiwand, line 148: now that I see this I wonder if there was a similar reason for `+1` and `<=` in https://github.com/acts-project/acts/pull/1381/files#diff-8195d95c9e69fcf3f1f7fade7f05b8253792f151a35fa7fc47da3c8800fc90b6L159 @timadye ?

Examples/Io/Csv/src/CsvBFieldWriter.cpp
```diff
@@ -0,0 +1,261 @@
...
+      for (std::size_t y = 0; y <= bins[1]; ++y) {
+        for (std::size_t z = 0; z <= bins[2]; ++z) {
+          /*
+           * Compute the geometric position of this bin, then request the
+           * magnetic field vector at that position.
+           */
+          Acts::Vector3 pos = {x * delta[0] + min[0], y * delta[1] + min[1],
+                               z * delta[2] + min[2]};
```
AJPfleger, line 156: How happy are you with the performance? Would it make sense to rewrite this part to something like this?
```
Acts::Vector3 pos = { min_x-delta_x, min_y-delta_y, min_z-delta_z }
for (x-bins)
  pos = { +delta_x, ..., ... }
  for (y-bins)
    pos = {..., +delta_y, ... }
    for (z-bins)
      pos = { ..., ..., +delta_z }
```

(same for cylindrical)

AJPfleger (changes requested):

Examples/Io/Csv/src/CsvBFieldWriter.cpp
stephenswat: Hmm, this is an interesting suggestion, but my issue with this is floating point error. If we go for a loop structure like this, if we iterate 300 times, then we incur floating point precision loss 300 times instead of just once, which is probably not ideal. I don't know how big the impact of this would be in practice, but this tool is really a "use once" sort of tool, because you don't need to dump B-fields to disk every day!

Examples/Io/Csv/include/ActsExamples/Io/Csv/CsvBFieldWriter.hpp
```diff
@@ -0,0 +1,114 @@
+// This file is part of the Acts project.
+//
+// Copyright (C) 2017-2021 CERN for the benefit of the Acts project
```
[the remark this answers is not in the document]
stephenswat: Well spotted, will fix these after lunch. :)

Examples/Io/Csv/src/CsvBFieldWriter.cpp
andiwand: I believe sequential adds are less accurate than multiplying. `pos` could drift away because of stacked FP inaccuracies

Examples/Io/Csv/src/CsvBFieldWriter.cpp
stephenswat: Good question, to me it makes sense that a 300 × 300 × 300 _bin_ B-field would consist of 301 × 301 × 301 vectors, because you need to account for the additional edge points. But we will wait to see what Tim thinks.

Examples/Io/Csv/src/CsvBFieldWriter.cpp
```diff
@@ -0,0 +1,261 @@
...
+     * coordinates, because we assume rotational cylindrical symmetry. This
+     * makes the procedure quite a bit faster, too. Great!
+     */
+    for (std::size_t r = 0; r <= bins[0]; ++r) {
+      for (std::size_t z = 0; z <= bins[1]; ++z) {
+        /*
+         * Calculate the position (still in three dimensions), assuming that
+         * the phi coordinate is zero. Then grab the field.
+         */
+        Acts::Vector3 pos(min[0] + r * delta[0], 0.f, min[1] + z * delta[1]);
```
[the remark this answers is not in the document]
stephenswat: Yes, this would be a good idea.

asalzburger:
Is this good to go? There are still conversations unresolved.

stephenswat:
I think we're waiting for some comments from Tim.

Examples/Io/Csv/include/ActsExamples/Io/Csv/CsvBFieldWriter.hpp
```diff
@@ -0,0 +1,114 @@
...
+#include "Acts/MagneticField/InterpolatedBFieldMap.hpp"
+#include "Acts/MagneticField/MagneticFieldProvider.hpp"
+#include "Acts/Utilities/Logger.hpp"
+
+#include <array>
+#include <optional>
+
+namespace ActsExamples {
+/**
+ * @brief Writer for B-fields that outputs field data in CSV format.
```
paulgessinger, line 20: Would you mind converting these to `///`?

Examples/Python/src/Output.cpp
```diff
@@ -601,5 +602,40 @@ void addOutput(Context& ctx) {
...
+
+    py::enum_<Writer::CoordinateType>(w, "CoordinateType")
+        .value("rz", Writer::CoordinateType::RZ)
+        .value("xyz", Writer::CoordinateType::XYZ);
+
+    REGISTER_CSV_BFIELD_WRITER_BINDING("XyzGrid", XYZ, true);
+    REGISTER_CSV_BFIELD_WRITER_BINDING("XyzGridless", XYZ, false);
+    REGISTER_CSV_BFIELD_WRITER_BINDING("RzGrid", RZ, true);
+    REGISTER_CSV_BFIELD_WRITER_BINDING("RzGridless", RZ, false);
+  }
```
paulgessinger, line 639: Can you `#undef` the macro here, even though it's not strictly needed?

Examples/Python/src/Output.cpp
```diff
@@ -601,5 +602,40 @@ void addOutput(Context& ctx) {
     ACTS_PYTHON_MEMBER(nSimulatedEvents);
     ACTS_PYTHON_STRUCT_END();
   }
+
+#define REGISTER_CSV_BFIELD_WRITER_BINDING(suffix, coordinate, grid)      \
```
paulgessinger, line 606: I would probably have gone for a template function instead of the macro, but not critical.

Examples/Io/Csv/include/ActsExamples/Io/Csv/CsvBFieldWriter.hpp
stephenswat: Sure, do you also prefer `//` for non-Doxygen comments?

stephenswat: I've gone and converted it all to `///` and `//`.

Examples/Python/src/Output.cpp
stephenswat: Done.

Examples/Python/src/Output.cpp
stephenswat: Changed to use a template function.

Examples/Io/Csv/src/CsvBFieldWriter.cpp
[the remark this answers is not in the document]
stephenswat: Actually, this method of writing (where phi is not variable) assumes cylindrical symmetry anyway, so I am not sure this is too useful.

stephenswat:
Okay, unless Tim has any comments on the outstanding issue I think this is good to go from my side.

timadye:
Sorry, I was away. I'll take a look today.

Examples/Python/src/Output.cpp
andiwand: the macro is now unused and can be dropped no?

stephenswat: Wait why is this still here

Examples/Io/Csv/src/CsvBFieldWriter.cpp
```diff
@@ -0,0 +1,219 @@
...
+// times by just compiling these symbols once into the object file
+// corresponding to this TU.
+template void CsvBFieldWriter::run<CsvBFieldWriter::CoordinateType::XYZ, true>(
+    const Config<CoordinateType::XYZ, true>&, Acts::Logging::Level);
+template void CsvBFieldWriter::run<CsvBFieldWriter::CoordinateType::RZ, true>(
+    const Config<CoordinateType::RZ, true>&, Acts::Logging::Level);
+template void CsvBFieldWriter::run<CsvBFieldWriter::CoordinateType::XYZ, false>(
+    const Config<CoordinateType::XYZ, false>&, Acts::Logging::Level);
+template void CsvBFieldWriter::run<CsvBFieldWriter::CoordinateType::RZ, false>(
+    const Config<CoordinateType::RZ, false>&, Acts::Logging::Level);
```
paulgessinger, line 218: Doesn't hurt but did you measure if this has an effect? I believe the compiler will instantiate the template in other TU anyway, and just discard duplicates during linking (like with `inline`).

Examples/Io/Csv/src/CsvBFieldWriter.cpp
timadye: Yeah, I removed the final bin to match the existing ROOT format. That gives field points at the lower-left point of each bin, not including the overflow bin. For the ATLAS B-field that I tested it on, it has 201 x 201 x 301 points, but that translates to 201 x 201 x 301 bins (not counting underflow and overflow bins).

For your CSV format, you can write whatever format you like, but it might be useful to keep consistent if possible. If you do include the overflow bin, then you should probably also include the underflow bin. Also, there may be problems calling `getField()` for value >= max (this is something I am trying to address in #1467).

timadye:
Are you planning to include a CSV *reading* tool? Apart from providing another good format, that would be useful to test this by writing and reading back.
I see there's a `Examples/Detectors/MagneticField/src/FieldMapTextIo.cpp`, but that looks like it is space-separated values.

timadye:
Does this have a different interface compared to `Examples/Io/Root/include/ActsExamples/Io/Root/RootBFieldWriter.hpp`? It looks like this has the `CsvBFieldWriter::CoordinateType` coded as a template parameter, rather than an `enum Config::gridType`. That could be an improvement, but maybe needs considering. If it is better, we might consider changing `RootBFieldWriter` at a later stage. If so, do we need a more general `BFieldWriter::CoordinateType` or whatever?

Does this difference follow through to the Python API? It would also be nice to have a Python example and/or test to be sure it works.

Sorry for all the comments.

Examples/Io/Csv/src/CsvBFieldWriter.cpp
stephenswat: Right, but that still increases compile times in those other TUs, right? I haven't tested this at all, but it seemed to me like hiding them implementation from other translation units entirely would be the most sound way of doing this. But I am open to change it.

stephenswat:
Hi Tim, I do have some changes in the pipeline to the Python B-field writing examples, but I opted to keep them separate from this PR for the time being. I like the idea of having a CSV reader, and I can work on implementing one.

Regarding the coordinate type enum, I agree that it might be useful to merge them, but there is also something to be said for embedding the capabilities of each writer into the class itself.

Examples/Io/Csv/src/CsvBFieldWriter.cpp
paulgessinger: I think what I'm saying is I don't think this does anything.

stephenswat: Well in the end I don't feel strongly either way; if you prefer to not have the specializations here I can move the code to a header file and it should work. Let me know what you prefer.

Examples/Io/Csv/src/CsvBFieldWriter.cpp
stephenswat: Hmm, I configured the output to be the same as the CSV files at https://gitlab.cern.ch/acts/acts-data/-/tree/master/MagneticField/ATLAS. Perhaps we should try to converge on a preferred format for this.

timadye: Can you check how many bins are given in the `bins` array for the ATLAS B-field? When I last tested this, reading in `acts-itk/bfield/ATLAS-BField-xyz.root`, I got `bins = field.getNBins() = {201,201,301}`, so `x < bins[0]` should include all 201 elements in the CSV.

Sorry if I misunderstood.

paulgessinger:
@timadye @stephenswat can you advise if we can merge this?

timadye:
I still had one issue outstanding, [here](https://github.com/acts-project/acts/pull/1470#discussion_r963005185). I hope @stephenswat can check that.

The definitive check would be to read the ATLAS B-field, write with this PR, and check the files have the same binning. If you are not able to do that, I can give it a go.

Examples/Io/Csv/src/CsvBFieldWriter.cpp
stephenswat: Hi Tim, I think perhaps the core problem here is that we disagree on what the concept of bin is. I think this makes sense in part because the term "bin" is somewhat undescriptive. Let's take, for example, the following example:



We might imagine this being some two-dimensional vector field spanning the range [-30mm, 30mm]<sup>2</sup>. It is my understanding that in such a model, the vector field has 49 "bins"; 7 in the horizontal axis and 7 in the vertical axis. Each of the bins is positioned exactly on the cross points of the horizontal and vertical lines, much like if one was playing a game of baduk. The distance between the points, in this case, is 10mm; this meshes nicely with the idea that we need 7 points to span a range of 60mm (which is 6 times the "bin" resolution if 10mm): it's the number of squares ("texels") plus one.

The only other intuition I can muster is to imagine that the "bins" are positioned at the center of the white squares defined by this grid. Perhaps this is the way you are thinking about this, and perhaps this is why we disagree on the structure of the output file?

The [ATLAS magnetic field in the Acts data repository](https://gitlab.cern.ch/acts/acts-data/-/tree/master/MagneticField/ATLAS) has a resolution of 10cm and spans 20 meters in the _x_ and _y_ axes, and 30 meters in the _z_ axis. The only way I can make sense of this is by imagining the "bins" as I described earlier: as the intersections of the grid lines. This also meshes with the contents of the file, which has increments of 10cm starting from the origin 0, rather than having the same grid but offset by 5cm, as would be expected in the alternative interpretation. The ATLAS field also has 201×201×301 "bins", rather than 200×200×300; the former is what would be produced by the implementation in this PR.

timadye: Hi Stephen,

Thanks for the nice discussion. I always find this confusing, so each time have to go back to the source to understand it again. I'm pretty sure there is still an inconsistency, but I thought I should check my understanding by running the code.

For me, the best test would be to read in [`acts-itk/bfield/ATLAS-BField-xyz.root`](https://gitlab.cern.ch/atlas/acts-itk/-/tree/main/bfield) (or [`acts-data/MagneticField/ATLAS/ATLASBField_xyz.root`](https://gitlab.cern.ch/acts/acts-data/-/blob/master/MagneticField/ATLAS) if you prefer) and write it out as CSV. This should be possible with Python `acts.examples.MagneticFieldMapXyz` and your new `acts.examples.CsvBFieldWriter`. The CSV output should specify the same points as the ROOT input, or the `ATLASBField_xyz.txt` file that was used to create it.

Unfortunately, I can't get this to work. I modified a RootBFieldWriter example (adapted from [`bfield_writing.py`](https://github.com/acts-project/acts/blob/main/Examples/Scripts/Python/bfield_writing.py)) like this:

```
import acts.examples
field = acts.examples.MagneticFieldMapXyz("acts-itk/bfield/ATLAS-BField-xyz.root")
cfg = acts.examples.CsvBFieldWriter.ConfigXyzGrid()
cfg.bField = field
cfg.fileName = str("ATLAS-BField-xyz.csv")
acts.examples.CsvBFieldWriter.runXyzGrid(cfg, acts.logging.VERBOSE)
```

It's a bit unfortunate that the Csv interface is different from the Root interface. Can that be simplified?

Anyway, my attempt gives:

```
RuntimeError: B-field returned a non-extant value!
```

Do you know what's going on?

Thanks,
Tim.

timadye: I just noticed that it does print a message:

```
15:49:51    CsvBFieldWri   INFO      Writing XYZ field of size 202 x 202 x 302 to file ATLAS-BField-xyz.csv...
```
and created the CSV file, but only the first row with `(x,y,z) = (-10000,-10000,-15000...14900.33)`. That does indeed have the wrong z bins, but of course this isn't a complete test.

stephenswat:
I have incorporated Tim's patches, and marked him appropriately as a co-author of the commits.

andiwand:
this conflicts now with https://github.com/acts-project/acts/pull/1510

happy to re-approve after the conflict is resolved

timadye:
> I have incorporated Tim's patches, and marked him appropriately as a co-author of the commits.

Thanks @stephenswat! This looks good to me.

I guess you will make the other changes in a separate PR, namely:
1. harmonize the `runXyzGrid()` methods with the `RootBFieldWriter` and others.
2. replace my ugly `dynamic_cast` with a static test, eg. `if constexpr`.

We discussed this in the Tuesday dev meeting and were happy to go ahead with this PR as-is. I could still test the output is correct on the ATLAS B-field. Should I do that now, or wait for you to resolve the conflict with #1510?

asalzburger:
Hi @stephenswat - can you resolve the conflict here? Then we should be able to get this in.

stephenswat:
Conflict resolved.

timadye:
Let me test this now, and then I hope we can merge this sucker.

timadye (approved):
I tested with the ATLAS B-field map, and it looks the same as the root input.

andiwand (approved):

PR #1478, opened 2022-08-31
test: big bounding box vs frustum intersection

andiwand:
@paulgessinger mentioned that big bounding bounding boxed or near frustums might not intersect as no edge or vertex would be inside it. this unit test proves that it works and should protect us from regressions


Tests/UnitTests/Core/Utilities/BoundingBoxTest.cpp
```diff
@@ -1199,6 +1199,31 @@ BOOST_AUTO_TEST_CASE(frustum_intersect) {
...
+    Frustum fr(pos, dir, 0.972419);
+    fr.draw(ply, 10);
+
+    Box::Size size(vec3(100, 100, 2));
+    Box bb(&o, pos + dir * 10, size);
+    bb.draw(ply);
+
+    BOOST_CHECK(bb.intersect(fr));
+
+    std::ofstream os("frust3d-4s-bigbox.ply");
```
paulgessinger, line 1223: I guess the output shouldn't happen automatically?

andiwand: I actually copied it blindly from the other unit test. now that you mention it it might be good to have a toggle of that in general maybe?

paulgessinger: Oh, my bad then!

Tests/UnitTests/Core/Utilities/BoundingBoxTest.cpp
```diff
@@ -1199,6 +1199,31 @@ BOOST_AUTO_TEST_CASE(frustum_intersect) {
...
+
+    PlyVisualization3D<BoundingBoxScalar> ply;
+
+    vec3 pos = {0, 0, 0};
+    vec3 dir = {0, 0, 1};
+    Frustum fr(pos, dir, 0.972419);
+    fr.draw(ply, 10);
+
+    Box::Size size(vec3(100, 100, 2));
+    Box bb(&o, pos + dir * 10, size);
```
paulgessinger, line 1218: How about
```suggestion
    Box bb(&o, pos + dir * 7, size);
```

so it looks something like this:

image

paulgessinger (approved):

PR #1488, opened 2022-09-02
ci: Only run main ubuntu builds on WIP PRs

paulgessinger:


paulgessinger:
@benjaminhuth one problem with this approach is that removing the WIP label does not trigger the previously skipped workflows. What do you think?

benjaminhuth:
Hmm I'm not an expert on github actions, but could this not be done with this?

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#label

benjaminhuth:
Ah no sorry, that seems to be related to the global creation and deletion...

andiwand:
@benjaminhuth I fear that only runs when you add/remove/modify a label for the repo but not for the PR

@paulgessinger I think it is okay. we can always trigger by commit. manual trigger would be a nice solution but might not be worth the time implementing it right away

benjaminhuth:
https://stackoverflow.com/questions/62325286/run-github-actions-when-pull-requests-have-a-specific-label

This seems to adress this... 
If we add

```yaml
on:
  pull_request:
    types: [ unlabeled ]
```

would this work?

Maybe it would always trigger if we unlabel a PR with any label then, but that should be rare cases?

benjaminhuth:
however, its maybe a bit unconvenient to require a empty commit in some cases, but often there will be anyhow a sync with `main` necessary and that should trigger it then...

paulgessinger:
Triggering the CI on unlabel might put more load on the CI than running all jobs on WIP PRs.

benjaminhuth:
@paulgessinger I tried a bit on my fork to set this up correctely, and I think I found a way to express this:

On the requirements the `push` trigger must be removed, so that the `pull_request: synchronized` is triggered. Then we must specify all triggers for pull request:
```yaml
on:
  pull_request:
    branches:
      - main
      - 'release/**'
      - 'develop/**'
    paths-ignore:
      - "docs/**"
    types:
      - opened
      - reopened
      - synchronize 
      - unlabeled
```


The condition for the linux_ubuntu job then must be: 
```yaml
linux_ubuntu:
  if: "${{ (github.event.action != 'unlabeled') || (github.event.action == 'unlabeled' && github.event.label.name == ':construction: WIP') }}"
```

The one for the non-WIP-jobs is:
```yaml
only_if_not_wip:
  if: "${{ (github.event.action != 'unlabeled' && !contains(github.event.pull_request.labels.*.name, ':construction: WIP')) || (github.event.action == 'unlabeled' && github.event.label.name == ':construction: WIP') }}"
```

The only other difference is that due to the removement of the push label, the CI won't be triggered on push onto branches of the repo that not have a PR associated anymore. But since we only work with PRs on the main repo, this should be fine, right?

PR #1492, opened 2022-09-05
ci: RTD fix for API listing

paulgessinger:


paulgessinger:
@benjaminhuth this should bring it back.

docs/conf.py
```diff
@@ -112,7 +115,12 @@
         cwd=cwd,
         env=env,
     )
+    if not api_index_target.exists():
+        shutil.copyfile(cwd / "api/api_index.rst", cwd / "api/api.rst")
```
benjaminhuth, line 119:
```suggestion
        shutil.copyfile(cwd / "api/api_index.rst", api_index_target)
```

benjaminhuth (approved):
Ah so the problem was that on readthedocs the thing isn't run with cmake and thus did not copy the file...
Looks good!

paulgessinger:
I applied your suggestion, can you reapprove?

benjaminhuth (approved):

PR #1497, opened 2022-09-06
docs: Switch to doxygen FAIL_ON_WARNINGS

paulgessinger:
Instead of exiting at the first warning that's treated as error, doxygen
will continue processing and exit fith a non-zero status code. This has
the benefit that all WARNINGs are visible at the same time.


benjaminhuth (approved):
Sounds like a very reasonable change to me.

PR #1500, opened 2022-09-07
fix: Make JsonMaterialDecorator IO error more explicit

paulgessinger:
This has caused confusion because the error doesn't mention which file it's actually trying to open.


Corentin-Allaire (approved):
I saw the Jira ticket, it's great that you added this.

PR #1507, opened 2022-09-12
refactor: Reuse MultiTrajectory in (C)KF

paulgessinger:
This PR changes the CKF and KF to reuse the same MultiTrajectory. CKF does this internally, retaining a single MTJ outside the for loop over the seeds. KF's `fit()` method now accepts an optional `trajectory` argument, that it will append track states into. If it's not given, it will create a fresh MTJ.

This cuts down on the number of objects we have flying around.

In the post-processing of CKF outputs via `Trajectory` objects, all of the objects per event will now point to the same MTJ. I plan to refactor this when the Track EDM lands.


benjaminhuth (approved):
Looks good to me! Very straight forward changes, nothing I want to comment on...

Just out of curiosity: Do you have measured if this has a performance impact?

paulgessinger:
@benjaminhuth No I haven't but that's actually a good point.

paulgessinger:
I ran KF, CKF and GSF examples to get a rough idea of the performance:

| Example | main            | feat            | events |
| ------- | --------------- | --------------- | ------ |
| ckf     | 6.287 +- 0.037s | 6.047 +- 0.076s | 1250   |
| kf      | 5.834 +- 0.017s | 5.752 +- 0.037s | 15000  |
| gsf     | 5.978 +- 0.035s | 6.598 +- 0.020s | 100    |

It seems CKF and KF are *slightly* faster, while GSF gets substantially slower. This is average + stdev of 10 runs each.

@benjaminhuth Any idea why this change might be affecting GSF differently than (C)KF? Overhead from `shared_ptr` copies?

paulgessinger:
Ok I think I managed to bring down the GSF example runtime on this branch to something like 6.043 +- 0.018s by lifting where it resolves if an input trajectory is given. That might be acceptable, @benjaminhuth?

benjaminhuth:
Interesting that the position where the MT is constructed has such a big impact on the performance...

However, GSF performance is a topic that must be addressed in general at some point, so that is totally fine with the change.

paulgessinger:
I think this is working now, can you reapprove @benjaminhuth?

benjaminhuth (approved):

PR #1508, opened 2022-09-12
fix: ParticleSmearing options not setup in AMVF example exe

### LINKED ISSUE #1506: Error while i run ActsExampleAdaptiveMultiVertexFinder
Hi! 
My main goal is to create tracks and vertexes based on real particle information from Pythia and by using ACTS to reconstruct tracks and vertexes. when i try to use ActsExampleAdaptiveMultiVertexFinder  i get error : terminate called after throwing an instance of 'boost::wrapexcept<boost::bad_any_cast>'
  what():  boost::bad_any_cast: failed conversion using boost::any_cast

How can i fix this error? 
Acts

paulgessinger:
Missing call to `addParticleSmearingOptions`.

Fixes #1506


andiwand (approved):
thank you for fixing this!

PR #1510, opened 2022-09-12
refactor: Add macro to simplify algorithm binding

benjaminhuth:
While working with the python bindings, I thought its quite a lot of repetive boiler plate code to add an Algorithm to the python bindings. With the help of `Boost.Preprocessor` this can be simplified quite a lot:

```c++ 
ACTS_PYTHON_DECLARE_ALGORITHM(
    ActsExamples::FooAlgorithm,
    "FooAlgorithm", 
    inputA, inputB, output)
```

expands to 

```c++
{
  using Alg = ActsExamples::FooAlgorithm;
  using Config = Alg::Config;
  auto alg = py::class_<Alg, ActsExamples::BareAlgorithm, std::shared_ptr<Alg>>(
                 mex, "FooAlgorithm")
                 .def(py::init<const Config &, Acts::Logging::Level>(),
                      py::arg("config"), py::arg("level"))
                 .def_property_readonly("config", &Alg::config);
  auto c = py::class_<Config>(alg, "Config").def(py::init<>());
  ACTS_PYTHON_STRUCT_BEGIN(c, Config);
  ACTS_PYTHON_MEMBER(inputA);
  ACTS_PYTHON_MEMBER(inputB);
  ACTS_PYTHON_MEMBER(output);
  ACTS_PYTHON_STRUCT_END();
}
```

This supports any number of arguments due to the preprocessor loops of Boost. Since we use macros anyways here to simplify declaration, I think this could be an acceptable use case of another macro.

I have this already appied for the Exa.TrkX algorithm, but wanted to wait for feedback (@paulgessinger) before doing the rest of the algorithms.


Examples/Python/include/Acts/Plugins/Python/Utilities.hpp
```diff
@@ -64,5 +69,26 @@ void patchKwargsConstructor(T& c) {
...
+            .def(py::init<const Config&, Acts::Logging::Level>(),           \
+                 py::arg("config"), py::arg("level"))                       \
+            .def_property_readonly("config", &Alg::config);                 \
+                                                                            \
+    auto c = py::class_<Config>(alg, "Config").def(py::init<>());           \
+    ACTS_PYTHON_STRUCT_BEGIN(c, Config);                                    \
+    BOOST_PP_SEQ_FOR_EACH(ACTS_PYTHON_MEMBER_LOOP, _,                       \
+                          BOOST_PP_VARIADIC_TO_SEQ(__VA_ARGS__))            \
+    ACTS_PYTHON_STRUCT_END();                                               \
+  }
```
stephenswat, line 94: Can we wrap this loop body in a `do { } while (0)` construct? :slightly_smiling_face:

Examples/Python/include/Acts/Plugins/Python/Utilities.hpp
```diff
@@ -64,5 +69,26 @@ void patchKwargsConstructor(T& c) {
...
+        py::class_<Alg, ActsExamples::BareAlgorithm, std::shared_ptr<Alg>>( \
+            mex, name)                                                      \
+            .def(py::init<const Config&, Acts::Logging::Level>(),           \
+                 py::arg("config"), py::arg("level"))                       \
+            .def_property_readonly("config", &Alg::config);                 \
+                                                                            \
+    auto c = py::class_<Config>(alg, "Config").def(py::init<>());           \
+    ACTS_PYTHON_STRUCT_BEGIN(c, Config);                                    \
+    BOOST_PP_SEQ_FOR_EACH(ACTS_PYTHON_MEMBER_LOOP, _,                       \
+                          BOOST_PP_VARIADIC_TO_SEQ(__VA_ARGS__))            \
```
paulgessinger, line 92: I was originally trying to do something like this, but didn't want to use boost for the functionality. But I guess this is a good idea since it simplifies the code.

Examples/Python/include/Acts/Plugins/Python/Utilities.hpp
benjaminhuth: sure we can!

Examples/Python/include/Acts/Plugins/Python/Utilities.hpp
benjaminhuth: Yeah but I also quickly thought about doing this without boost, but after seeing what obscure constructs are needed for a preprocessor loop...^^

paulgessinger:
Existing python tests seem to fail with this change. Can you check if you can reproduce locally, @benjaminhuth?

benjaminhuth:
Yeah I already know what the problem is. Most writers/readers do not provide a `config()` accessor in C++, so I left the corresponding macro without defining that in python.

However, it seems like that the `config` property it is used in one place at least.

I see three options there
* remove that one use case
* Add config accessors to all readers and writers and defining that in the macro
* Don't use the macro in the one special case (not a good solution I think)

I would go with the second option, because I think it is useful to have this accessor... What do you think @paulgessinger ?

paulgessinger:
Yes I agree, adding the config accessor everywhere seems like the right thing to do.

benjaminhuth:
@paulgessinger This should be ready to go in now, could you have a look and approve?

Examples/Python/src/ExampleAlgorithms.cpp
```diff
@@ -32,92 +32,24 @@ void addExampleAlgorithms(Context& ctx) {
...
-        .def_readwrite("inputTrackParameters",
-                       &Alg::Config::inputTrackParameters);
-  }
+  ACTS_PYTHON_DECLARE_ALGORITHM(
+      ActsExamples::FatrasSimulation, mex, "FatrasSimulation", inputParticles,
+      outputParticlesInitial, outputParticlesFinal, outputSimHits,
+      imputParametrisationNuclearInteraction, randomNumbers, trackingGeometry,
+      magneticField, pMin, emScattering, emEnergyLossIonisation,
+      emEnergyLossRadiation, emPhotonConversion, generateHitsOnSensitive,
+      generateHitsOnMaterial, generateHitsOnPassive, averageHitsPerParticle);
```
paulgessinger, line 41: That's the only thing I'm skeptical about: this is hard to read, and might make diffs annoying if properties get added. I think the benefits outweigh this downside though.

Examples/Python/src/Json.cpp
```diff
@@ -116,4 +116,4 @@ void addJson(Context& ctx) {
     ACTS_PYTHON_STRUCT_END();
   }
 }
-}  // namespace Acts::Python
\ No newline at end of file
+}  // namespace Acts::Python
```
paulgessinger, line 119: Whitespace only change?

paulgessinger (commented):
Looks good I think. If the output doesn't change, I'm happy.

Examples/Python/src/ExampleAlgorithms.cpp
benjaminhuth: so in principle we *could* provide another `.clang-format` in the `Examples/Python/src` directory that breaks at each comma... I think clang-format does take always the nearest one, right?

paulgessinger (approved):

PR #1512, opened 2022-09-12
refactor!: MTJ stores measurement as jagged vector

paulgessinger:
Addresses #1516.

```
 | x x x | x | x x |  ...
   ^       ^     ^
 M1,D=3 M2,D=1 M3,D=2
```

BREAKING CHANGE: `Acts::MultiTrajectory` measurement access methods change:
```diff
- constexpr auto measurement(IndexType measIdx) const;
+ template <size_t measdim> 
+ constexpr auto measurement(IndexType measIdx) const;
```
and 
```diff
- constexpr auto measurementCovariance(IndexType covIdx)
+ template <size_t measdim>
+ constexpr auto measurementCovariance(IndexType covIdx)
```


paulgessinger:
This currently includes changes from #1507.

paulgessinger:
CI failures seem to be a GCC8 bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86594

I think this is ready for review.

paulgessinger:
I pushed a workaround.

paulgessinger:
@tboldagh thoughts?

Core/include/Acts/EventData/VectorMultiTrajectory.hpp
```diff
@@ -307,10 +333,12 @@ class VectorMultiTrajectory final
   std::vector<IndexType> m_previous;
   std::vector<typename detail_lt::Types<eBoundSize>::Coefficients> m_params;
   std::vector<typename detail_lt::Types<eBoundSize>::Covariance> m_cov;
-  std::vector<typename detail_lt::Types<MeasurementSizeMax>::Coefficients>
-      m_meas;
-  std::vector<typename detail_lt::Types<MeasurementSizeMax>::Covariance>
-      m_measCov;
+
+  std::vector<double> m_meas;
```
tboldagh, line 337: So this is now a flat vector which can only be used with offsets to get the measurements/covariances?

paulgessinger: In this implementation: yes. My thinking was that on the xAOD side it's `std::vector<std::vector<double>>` anyway, so the access remains trivial.

paulgessinger:
Ok I've marked this as backward breaking now.
Should we get this in @tboldagh?

Core/include/Acts/EventData/MultiTrajectory.hpp
```diff
@@ -578,24 +622,41 @@ class TrackStateProxy {
...
 
   /// Dynamic measurement vector with only the valid dimensions.
   /// @return The effective calibrated measurement vector
   auto effectiveCalibrated() const {
-    return calibrated().head(calibratedSize());
+    double* data{nullptr};
+
+    visit_measurement(*this, calibratedSize(), [&](auto meas, auto cov) {
+      (void)cov;
+      data = const_cast<double*>(meas.data());
```
benjaminhuth, line 641: Can you comment on the `const_cast` here? I quickly tried to look through this, but didn't really understood why this is necessary here. Is it because the `Eigen` types require non-const pointers? And are we sure that we don't invoke UB here?

Maybe i would be good to add an extra comment on why this is used...

Core/include/Acts/EventData/MeasurementHelpers.hpp
```diff
@@ -36,6 +36,15 @@ struct visit_measurement_callable {
...
 };
+
+template <size_t I>
+struct visit_measurement_callable_ts {
+  template <typename track_state_t, typename L>
+  auto static constexpr invoke(track_state_t ts, L&& lambda) {
+    return lambda(ts.template calibrated<I>(),
+                  ts.template calibratedCovariance<I>());
+  }
+};
```
paulgessinger, line 47:
```suggestion
```

Core/include/Acts/EventData/MultiTrajectory.hpp
paulgessinger: should be gone i think

paulgessinger:
Still need to update the memory stats calculation

tboldagh:
> Ok I think this is ready now.
> 
> @tboldagh would you be able to have another look?

Yes. Today evening.

Core/include/Acts/EventData/MultiTrajectory.hpp
```diff
@@ -223,8 +235,17 @@ class TrackStateProxy {
...
+      if (ACTS_CHECK_BIT(src, PM::Calibrated)) {
+        // on-demand allocate calibrated
+        dest |= PM::Calibrated;
+      }
+
+      if ((static_cast<std::underlying_type_t<TrackStatePropMask>>(
+               (src ^ dest) & src) != 0 ||
+           dest == TrackStatePropMask::None ||
+           src == TrackStatePropMask::None) &&
+          mask != TrackStatePropMask::None) {
```
tboldagh, line 248: Can any hint of what is wrong be part of exception message (the requirements are not quite nontrivial).

tboldagh: Also, what should happen if destination TS has smaller size allocated for measurement?

Core/include/Acts/TrackFitting/GainMatrixUpdater.hpp
```diff
@@ -81,8 +80,17 @@ class GainMatrixUpdater {
             trackState.predictedCovariance(),
             trackState.filtered(),
             trackState.filteredCovariance(),
-            trackState.calibrated(),
-            trackState.calibratedCovariance(),
+            // This abuses an incorrectly sized vector / matrix to access the
+            // data pointer! This works (don't use the matrix as is!), but be
+            // careful!
+            trackState
```
tboldagh, line 86: Would it possible to hide this c++ syntax mess somehow in the TracksState?

Core/include/Acts/Utilities/Helpers.hpp
```diff
@@ -387,6 +388,28 @@ auto template_switch(size_t v, Args&&... args) {
...
+auto template_switch_lambda(size_t v, Lambda&& func, Args&&... args) {
+  if (v == N) {
+    return func(std::integral_constant<size_t, N>{},
+                std::forward<Args>(args)...);
+  }
+  if constexpr (N < NMAX) {
+    return template_switch_lambda<N + 1, NMAX>(v, func,
+                                               std::forward<Args>(args)...);
+  }
+  std::cerr << "template_switch<Fn, " << N << ", " << NMAX << ">(v=" << v
```
tboldagh, line 408: I am afraid this will not tell anything someone in, say Alice. If this can not be handled via exception maybe a prefix message: "ACTS Helpers". Still not very helpful but at least hints the direction.

Core/src/TrackFitting/GainMatrixUpdater.cpp
```diff
@@ -16,71 +19,70 @@ std::tuple<double, std::error_code> GainMatrixUpdater::visitMeasurement(
...
+    //        currently take a +0.7GB build memory consumption hit in the
+    //        EventDataView unit tests. Revisit this once Measurement
+    //        overhead problems (Acts issue #350) are sorted out.
+    //
+    ParametersVector residual;
+    residual = calibrated - H * trackState.filtered;
+    ACTS_VERBOSE("Residual: " << residual.transpose());
+
+    chi2 = (residual.transpose() *
+            ((CovarianceMatrix::Identity() - H * K) * calibratedCovariance)
```
tboldagh, line 78: A curiosity. Since now CovMatrix can be of different size why there is no size specifier here? Not sure how it cold actually work.

tboldagh (approved):
I have no substantive comments. Just few small requests to make the diagnostics more useful. It at all possible.

Core/include/Acts/Utilities/Helpers.hpp
paulgessinger: I guess it could be an exception.

Core/src/TrackFitting/GainMatrixUpdater.cpp
paulgessinger: H is the projector and has the dimension `eBoundSize x N`, where N is the measurement dimension. This should give an `NxN` matrix, which can be multiplied by `calibratedCovariance` which is also `NxN`.

Core/include/Acts/TrackFitting/GainMatrixUpdater.hpp
paulgessinger: Not easily. We could explicitly expose the raw `const double*` / `double*` to allow construction of custom proxy types, but I wasn't sure that's a great idea either.

Core/include/Acts/EventData/MultiTrajectory.hpp
paulgessinger: I'll add some comments to the if to explain what it tries to check.

If the destination TS has a calibrated measurement allocated, it will reallocated **unless** the dimensions match up. So even if we copy from smaller to larger we will reallocate. My thinking was that in almost all cases we will be allocating from nothing, so it doesn't make a difference.

tboldagh (approved):

tboldagh (approved):
No comment. I made a mistake in using the gitlab UI.

PR #1513, opened 2022-09-12
refactor: improve full_chain_itk.py example

timadye:
* `full_chain_itk.py` can now run ttbar+PU200 selected by a flag (`ttbar_pu200`), rather than commented-out code.
* Remove truth selection for particle gun. This is not needed and can give problems if a lower generated pT range is selected.
* Remove previously commented-out `SeedingAlgorithm.TruthEstimated` and `SeedingAlgorithm.TruthSmeared` options.
* simplify imports


timadye:
This is a proposal for clearer code to explain at the ACTS workshop.

If this is good, we can do something similar for `full_chain_odd.py`. I could include in this PR. Do we want to add the ttbar+PU200 example there too?

andiwand (approved):
looks good to me :+1: 

I think it makes sense to add similar changes to the odd full chain. I can do that in a separate PR

timadye:
> I think it makes sense to add similar changes to the odd full chain. I can do that in a separate PR

Thanks! That would be great if you could add this to the odd full chain. If that's an sort of problem, I can do it.

PR #1520, opened 2022-09-14
docs: Update logging doc, add info on thresholds

paulgessinger:
This mainly adds information on the logging thresholds configuration.


benjaminhuth:
Very nice and understandable. After reading through, I found that there is one reference where I would expect a link or something to `CustomDefaultLogger.cpp` in line 128.

This is not in the part that you changed, but maybe you could add a link here @paulgessinger?

paulgessinger:
Added now, @benjaminhuth.

paulgessinger:
Minor change since @benjaminhuth's approval. I'll merge this.

PR #1521, opened 2022-09-14
fix: Added missing return to seedfinder::CreateSeedsForGroup

guilhermeAlmeida1:
Added missing return at the end of non-void function seedfinder::CreateSeedsForGroup.


LuisFelipeCoelho (approved):

paulgessinger:
Hmm. Codecov seems to be MIA.

paulgessinger:
There we go.

andiwand:
@paulgessinger is this something we can catch with clang tidy?

paulgessinger:
I guess `-Wreturn-type` in the compilation would already do it.

PR #1523, opened 2022-09-14
docs: Enable sphinx linkcheck build in CI

paulgessinger:


paulgessinger:
Let's see if this works.

paulgessinger:
I think it's more basic, the way I add the argument to the command in cmake escapes the space, so it doesn't parse it correctly.

paulgessinger:
I fixed the invocation I think, and also found an actual broken link! Can you check this again @benjaminhuth?

benjaminhuth (approved):
Very cool, the tool seems to work!

PR #1542, opened 2022-09-21
refactor: Add early stop mechanism for GSF forward pass

benjaminhuth:
Very simple PR that stops the GSF forward pass if all measurements are found similar to the mechanism in the KF. Should mainly reduce runtime since before the GSF always propagated until it reached end-of-world.

Also changes a related variable from `int` to `std::size_t` for consistency.


benjaminhuth:
As expected, this changes the root hashes for the `trackstates` file, but not for the `tracksummary` file.

asalzburger (approved):
Makes total sense, as the GSF will never run as a track finder, this is a good one.

asalzburger:
@benjaminhuth there are some conflicts in the hash files ...

asalzburger (approved):

PR #1543, opened 2022-09-21
fix: maybe-uninitialized warning in SpacepointBuilder

benjaminhuth:
My new GCC does emit a `-Wmaybe-uninitialized` warning in the `SpacePointBuilder.ipp`. Indeed, in the Error case (the `else` branch in the `buildSpacePoint` function) the values `gpos` and `gCov` seems to stay uninitialized.

I'm not sure if zero initialization is actually the best thing to do here (maybe @toyamaza can comment on that), maybe return a `Result<void>` could  also be an option?


toyamaza:
Thanks @benjaminhuth for fixing this. Looks good to me.

toyamaza (approved):

PR #1538, opened 2022-09-20
refactor: improve full_chain_odd.py example

andiwand:
- align odd full chain with itk full chain

see #1513

PR #1545, opened 2022-09-21
refactor: use TrackSelector in addVertexFitting

andiwand:
follow-up of https://github.com/acts-project/acts/pull/1538

@timadye noticed that we have to use `s.addAlgorithm(acts.examples.TrackSelector(...` in the full chain examples in order to have vertexing. this PR integrates the track selection into the `addVertexFitting` helper

### DIFF
--- a/Examples/Python/python/acts/examples/reconstruction.py
+++ b/Examples/Python/python/acts/examples/reconstruction.py
@@ -111,10 +111,10 @@ TrackParamsEstimationConfig = namedtuple(
     defaults=[(None, None)],
 )
 
-TrackSelectorConfig = namedtuple(
-    "TrackSelectorConfig",
-    ["absEtaMax", "ptMin"],
-    defaults=[2.5, 500 * u.MeV],
+TrackSelectorRanges = namedtuple(
+    "TrackSelectorRanges",
+    ["absEta", "pt"],
+    defaults=[(None, None)] * 2,
 )
 
 
--- a/Examples/Python/python/acts/examples/reconstruction.py
+++ b/Examples/Python/python/acts/examples/reconstruction.py
@@ -908,8 +908,9 @@ def addVertexFitting(
     associatedParticles: str = "particles_input",
     trajectories: Optional[str] = None,
     trackParameters: str = "fittedTrackParameters",
+    trackIndices: str = "outputTrackIndices",
     vertexFinder: VertexFinder = VertexFinder.Truth,
-    trackSelectorRanges: Optional[TrackSelectorRanges] = None,
+    trackSelectorRanges: Optional[TrackSelectorRanges] = TrackSelectorRanges(),
     logLevel: Optional[acts.logging.Level] = None,
 ) -> None:
     """This function steers the vertex fitting
@@ -943,7 +944,6 @@ def addVertexFitting(
 
     if trackSelectorRanges is not None:
         selectedTrackParameters = "trackparameters"
-        trackIndices = "outputTrackIndices"
 
         trackSelector = acts.examples.TrackSelector(
             level=customLogLevel(),
@@ -972,7 +972,6 @@ def addVertexFitting(
         s.addAlgorithm(trackSelector)
     else:
         selectedTrackParameters = trackParameters
-        trackIndices = ""
 
     inputParticles = "particles_input"
     outputVertices = "fittedVertices"
--- a/Examples/Python/python/acts/examples/reconstruction.py
+++ b/Examples/Python/python/acts/examples/reconstruction.py
@@ -941,10 +941,10 @@ def addVertexFitting(
 
     customLogLevel = acts.examples.defaultLogging(s, logLevel)
 
-    selectedTrackParameters = "trackparameters"
-    trackIndices = "outputTrackIndices"
-
     if trackSelectorRanges is not None:
+        selectedTrackParameters = "trackparameters"
+        trackIndices = "outputTrackIndices"
+
         trackSelector = acts.examples.TrackSelector(
             level=customLogLevel(),
             inputTrackParameters=trackParameters,
@@ -972,6 +972,7 @@ def addVertexFitting(
         s.addAlgorithm(trackSelector)
     else:
         selectedTrackParameters = trackParameters
+        trackIndices = ""
 
     inputParticles = "particles_input"
     outputVertices = "fittedVertices"
--- a/Examples/Python/python/acts/examples/reconstruction.py
+++ b/Examples/Python/python/acts/examples/reconstruction.py
@@ -942,8 +942,10 @@ def addVertexFitting(
             outputTrackIndices=trackIndices,
             **acts.examples.defaultKWArgs(
                 removeNeutral=trackSelectorRanges.removeNeutral,
+                absEtaMin=trackSelectorRanges.absEta[0],
                 absEtaMax=trackSelectorRanges.absEta[1],
                 ptMin=trackSelectorRanges.pt[0],
+                ptMax=trackSelectorRanges.pt[1],
             ),
         )
         s.addAlgorithm(trackSelector)
--- a/Examples/Python/python/acts/examples/reconstruction.py
+++ b/Examples/Python/python/acts/examples/reconstruction.py
@@ -111,6 +111,12 @@ TrackParamsEstimationConfig = namedtuple(
     defaults=[(None, None)],
 )
 
+TrackSelectorConfig = namedtuple(
+    "TrackSelectorConfig",
+    ["absEtaMax", "ptMin"],
+    defaults=[2.5, 500 * u.MeV],
+)
+
 
 @acts.examples.NamedTypeArgs(
     seedingAlgorithm=SeedingAlgorithm,
@@ -890,8 +896,9 @@ def addVertexFitting(
     outputDirRoot: Optional[Union[Path, str]] = None,
     associatedParticles: str = "particles_input",
     trajectories: Optional[str] = None,
-    trackParameters: str = "trackparameters",
+    trackParameters: str = "fittedTrackParameters",
     vertexFinder: VertexFinder = VertexFinder.Truth,
+    trackSelectorConfig: Optional[TrackSelectorConfig] = TrackSelectorConfig(),
     logLevel: Optional[acts.logging.Level] = None,
 ) -> None:
     """This function steers the vertex fitting
@@ -920,6 +927,23 @@ def addVertexFitting(
 
     customLogLevel = acts.examples.defaultLogging(s, logLevel)
 
+    selectedTrackParameters = "trackparameters"
+    trackIndices = "outputTrackIndices"
+
+    if trackSelectorConfig is not None:
+        trackSelector = acts.examples.TrackSelector(
+            level=customLogLevel(),
+            inputTrackParameters=trackParameters,
+            outputTrackParameters=selectedTrackParameters,
+            outputTrackIndices=trackIndices,
+            removeNeutral=True,
+            absEtaMax=trackSelectorConfig.absEtaMax,
+            ptMin=trackSelectorConfig.ptMin,
+        )
+        s.addAlgorithm(trackSelector)
+    else:
+        selectedTrackParameters = trackParameters
+
     inputParticles = "particles_input"
     outputVertices = "fittedVertices"
     selectedParticles = "particles_selected"
@@ -936,7 +960,7 @@ def addVertexFitting(
         fitVertices = VertexFitterAlgorithm(
             level=customLogLevel(),
             bField=field,
-            inputTrackParameters=trackParameters,
+            inputTrackParameters=selectedTrackParameters,
             inputProtoVertices=findVertices.config.outputProtoVertices,
             outputVertices=outputVertices,
         )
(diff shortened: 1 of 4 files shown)


Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -111,6 +111,12 @@
     defaults=[(None, None)],
 )
 
+TrackSelectorConfig = namedtuple(
+    "TrackSelectorConfig",
+    ["absEtaMax", "ptMin"],
+    defaults=[2.5, 500 * u.MeV],
+)
```
timadye, line 118:
```suggestion
TrackSelectorRanges = namedtuple(
    "TrackSelectorRanges",
    ["absEta", "pt"],
    defaults=[(None, None)] * 2,
)
```
except list all the parameters in `Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TrackSelector.hpp`. Except `removeCharged` and `removeNeutral`, they can all be specified as ranges, hence my suggested name change.
The defaults are given as `None` so we can keep the defaults from the `TrackSelector` C++. We handle the `None`s in the argument passing below...

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -920,6 +927,23 @@ def addVertexFitting(
...
+
+    if trackSelectorConfig is not None:
+        trackSelector = acts.examples.TrackSelector(
+            level=customLogLevel(),
+            inputTrackParameters=trackParameters,
+            outputTrackParameters=selectedTrackParameters,
+            outputTrackIndices=trackIndices,
+            removeNeutral=True,
+            absEtaMax=trackSelectorConfig.absEtaMax,
+            ptMin=trackSelectorConfig.ptMin,
```
timadye, line 941:
```suggestion
            **acts.examples.defaultKWArgs(
               removeNeutral=trackSelectorRanges.removeNeutral,
               absEtaMax=trackSelectorRanges.absEta[1],
               ptMin=trackSelectorRanges.pt[0],
            ),
```
and add all the other parameters.

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -890,8 +896,9 @@ def addVertexFitting(
     outputDirRoot: Optional[Union[Path, str]] = None,
     associatedParticles: str = "particles_input",
     trajectories: Optional[str] = None,
-    trackParameters: str = "trackparameters",
+    trackParameters: str = "fittedTrackParameters",
     vertexFinder: VertexFinder = VertexFinder.Truth,
+    trackSelectorConfig: Optional[TrackSelectorConfig] = TrackSelectorConfig(),
```
timadye, line 901:
```suggestion
    trackSelectorRanges: Optional[TrackSelectorRanges] = None,
```
also add a decorator to the `def addVertexFitting`:
```
@acts.examples.NamedTypeArgs(
trackSelectorRanges=TrackSelectorRanges
)
```
and a comment to the docstring:
```
    trackSelectorRanges : TrackSelectorRanges(absEta, pt)
        TrackSelector configuration. Each range is specified as a tuple of (min,max).
        Defaults of no cuts specified in Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TrackSelector.hpp
```

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -890,8 +896,9 @@ def addVertexFitting(
     outputDirRoot: Optional[Union[Path, str]] = None,
     associatedParticles: str = "particles_input",
     trajectories: Optional[str] = None,
-    trackParameters: str = "trackparameters",
+    trackParameters: str = "fittedTrackParameters",
```
timadye, line 899: if we need to change the default, then this change is not compatible with current clients. We should check all clients!

timadye (changes requested):
This is a good start, but we should probably support all the settings of `TrackSelector`. For examples of how to handle these `namedtuples`, see some of the other uses in `reconstruction.py`, eg. `TruthSeedRanges` and `truthSeedRanges` (just search for all uses).

timadye:
and of course we need to change `full_chain_odd.py` and `full_chain_itk.py` to use this. Let's hope #1538 is accepted by then.

Examples/Python/python/acts/examples/reconstruction.py
andiwand: thanks for the suggestions! I applied all of them but taking a second look the `addSeeding` helper is also using the default construction for the config/ranges. I thought that might be good as this will be the default usage

Examples/Python/python/acts/examples/reconstruction.py
andiwand: one problem with this is that we except more input than we handle in the algorithm with might be confusing to the user

Examples/Python/python/acts/examples/reconstruction.py
andiwand: I tried to look for them now and adopt the code. I think the rename makes sense because "per default" there will be a kalman pass before the vertexing

andiwand:
> and of course we need to change `full_chain_odd.py` and `full_chain_itk.py` to use this. Let's hope #1538 is accepted by then.

yes I wanted to get https://github.com/acts-project/acts/pull/1538 merged first otherwise we will face a bunch of conflicts

Examples/Python/python/acts/examples/reconstruction.py
timadye: > one problem with this is that we except more input than we handle in the algorithm with might be confusing to the user

Sorry, I don't understand this comment. We should only define parameters in `TrackSelectorRanges` that are defined in `TrackSelector`. Any of them *could* be used to select tracks as input for the vertexing.

Examples/Python/python/acts/examples/reconstruction.py
timadye: The advantage of having `None` as the default is that you can skip the `TrackSelector` altogether if it is not needed. This works because `TrackSelector` defaults are "don't cut any tracks".

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -920,6 +931,25 @@ def addVertexFitting(
...
+    if trackSelectorRanges is not None:
+        trackSelector = acts.examples.TrackSelector(
+            level=customLogLevel(),
+            inputTrackParameters=trackParameters,
+            outputTrackParameters=selectedTrackParameters,
+            outputTrackIndices=trackIndices,
+            **acts.examples.defaultKWArgs(
+                removeNeutral=trackSelectorRanges.removeNeutral,
+                absEtaMax=trackSelectorRanges.absEta[1],
+                ptMin=trackSelectorRanges.pt[0],
```
timadye, line 946:
```suggestion
                absEtaMin=trackSelectorRanges.absEta[0],
                absEtaMax=trackSelectorRanges.absEta[1],
                ptMin=trackSelectorRanges.pt[0],
                ptMax=trackSelectorRanges.pt[1],
```
and then add all the other selectors. Sorry, I didn't write them all out, but it's not so easy in the comments. If you are having trouble with this, I can make a commit of my own.

Examples/Scripts/Python/vertex_fitting.py
```diff
@@ -99,6 +99,7 @@ def runVertexFitting(
         outputDirRoot=outputDir if outputRoot else None,
         associatedParticles=associatedParticles,
         trackParameters=trackParameters,
+        trackSelectorRanges=None,
```
timadye, line 102: not needed if `None` is the default

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -77,21 +78,10 @@
...
-        removeNeutral=True,
-        absEtaMax=2.5,
-        loc0Max=4.0 * u.mm,  # rho max
-        ptMin=500 * u.MeV,
-    )
-)
 addVertexFitting(
     s,
     field,
+    TrackSelectorRanges(pt=(400.0 * u.MeV, None)),
```
timadye, line 84:
```suggestion
    TrackSelectorRanges(pt=(400.0 * u.MeV, None), absEta=(None, 2.5), removeNeutral=True),
```

timadye:
Sorry if I jumped in while you are still editing. I'll stop there. Let me know if you want me to take another look.

andiwand:
> Sorry if I jumped in while you are still editing. I'll stop there. Let me know if you want me to take another look.

thank you for the reviews! I will wait for the other branch to get in and resolve the conflicts. afterwards I can re-request a review if you are up for it

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -920,6 +941,39 @@ def addVertexFitting(
...
+                ptMax=trackSelectorRanges.pt[1],
+                removeCharged=trackSelectorRanges.removeCharged,
+                removeNeutral=trackSelectorRanges.removeNeutral,
+            ),
+        )
+        s.addAlgorithm(trackSelector)
+    else:
+        selectedTrackParameters = trackParameters
+        trackIndices = ""
```
andiwand, line 975: not sure if this makes sense. any suggestions @paulgessinger @timadye

timadye: If I call with `trackSelectorRanges=None`, so setting `trackIndices=""` I get:
```
You need to either provide collection of truth particles matching 1:1 to tracks, or track indices and all-tips container to do truth matching
```
Maybe the only way to do the truth matching is to use a `TrackSelector` with no cuts, and revert all the changes that allow for `trackSelectorRanges=None`. If so, apologies for suggesting that.

I'm still not clear how this is supposed to work. I assumed that `TrackSelector`'s `outputTrackIndices` gave the mapping between `inputTrackParameters` and `outputTrackParameters`, but `RootVertexPerformanceWriter`'s checks imply that `inputFittedTracksIndices` gives the mapping between `input(All|Selected)TruthParticles` and `inputFittedTracks`.

andiwand: In order to close this PR I would just default to an empty track selection for now. the callee can still overwrite `trackSelectorRanges` if necessary. afterwards I will take a deeper look into `RootVertexPerformanceWriter` and see if and what needs to be adopted

timadye: @paulgessinger, thanks for the explanation. We certainly don't have a one-to-one match (is that for truth tracking?), so we should keep the indices in all cases. We can just use the no-cuts `TrackSelector` always and make the default `TrackSelectorRanges()`.

ah, I see @andiwand just did that.

timadye: Just one concern about making the default `TrackSelectorRanges()`: this changes the default behavior to running `TrackSelector`, where before it wasn't run. Will that cause problems for the truth tracking? I don't know if we have any examples of that.

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -884,14 +900,17 @@ class VertexFinder(Enum):
...
     s,
     field,
     outputDirRoot: Optional[Union[Path, str]] = None,
     associatedParticles: str = "particles_input",
     trajectories: Optional[str] = None,
-    trackParameters: str = "trackparameters",
+    trackParameters: str = "fittedTrackParameters",
+    trackIndices: str = "outputTrackIndices",
     vertexFinder: VertexFinder = VertexFinder.Truth,
+    trackSelectorRanges: Optional[TrackSelectorRanges] = TrackSelectorRanges(),
```
timadye, line 913:
```suggestion
    trackSelectorRanges: TrackSelectorRanges = TrackSelectorRanges(),
```
I think it is only `Optional` if the default is `None`.

Examples/Python/python/acts/examples/reconstruction.py
timadye: ah, I see `addVertexFitting` is also used by `Examples/Scripts/Python/vertex_fitting.py`, which still uses a separate `TrackSelector` call. I guess that should be integrated, and you can put `trackSelectorRanges=None` for the case where it isn't applied.

timadye: sorry, I'll shut up now and let you finish the other changes and can make a final check when you are done.

andiwand: oh yeah good point about the `vertex_fitting.py` - I will adopt that. hopefully the CI results will recover then

andiwand:
looks like this is passing now @timadye

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -907,6 +926,9 @@ def addVertexFitting(
         RootVertexPerformanceWriter.inputAssociatedTruthParticles
     vertexFinder : VertexFinder, Truth
         vertexFinder algorithm: one of Truth, AMVF, Iterative
+    trackSelectorRanges : TrackSelectorRanges(absEta, pt)
```
timadye, line 929: can you document all the parameters here?

timadye (changes requested):
just a couple of small things...

Examples/Scripts/Python/vertex_fitting.py
```diff
@@ -65,37 +69,34 @@ def runVertexFitting(
         )
         s.addAlgorithm(ptclSmearing)
         associatedParticles = selectedParticles
+
+        trackSelectorRanges = TrackSelectorRanges()
```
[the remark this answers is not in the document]
andiwand: I thought this broke a test case somehow but now that I look at it again I don't see why

timadye (approved):
looks good to go!

PR #1558, opened 2022-09-27
refactor: time optimization for calculation of min and max r variable range

LuisFelipeCoelho:
This PR changes the position of rMiddleSPRange calculation and adds a break when SPs are sorted in r


paulgessinger:
Can you consider using a pair or separate argument for the min/max values, instead of a `Vector2`? I don't think using `Vector2` brings any tangible benefit, and it might be a bit confusing what it means.

benjaminhuth:
> Can you consider using a pair or separate argument for the min/max values, instead of a `Vector2`? I don't think using `Vector2` brings any tangible benefit, and it might be a bit confusing what it means.

Actually, as Andi mentioned last developer meeting, it looks like a good use case for `Acts::Range1D` from the utilities folder.

I also noticed that there is this `SeedConfirmationRangeConfig.hpp that also defines two ranges as just min and max values that could also be refactored, however not necessarily im this PR.

LuisFelipeCoelho:
@benjaminhuth I changed it to `Acts::Range1D` and will also do the same in `SeedConfirmationRangeConfig.hpp` in a different PR

andiwand (approved):

PR #1564, opened 2022-09-29
fix: Remove example spacepoint from seeding code

### LINKED ISSUE #1557: Bug: SeedfinderOrthogonal in Acts Core uses ActsExamples::SimSpacePoint
Looking at the code for the Orthogonal Seeding in Acts Core that was introduced with https://github.com/acts-project/acts/pull/904 by @stephenswat, it looks like it uses `ActsExamples::SimSpacePoint`: https://github.com/acts-project/acts/blob/main/Core/include/Acts/Seeding/SeedFinderOrthogonal.ipp#L509 

```cpp
/*
* Create a vector to contain protoseeds.
*/
std::vector<std::pair<
  float, std::unique_ptr<const InternalSeed<ActsExamples::SimSpacePoint>>>>
  protoseeds;
```

As far as I can see `ActsExamples::SimSpacePoint` is defined in the ActsExamples folder: https://github.com/acts-project/acts/blob/main/Examples/Framework/include/ActsExamples/EventData/SimSpacePoint.hpp 

Tagging @paulgessinger and @LuisFelipeCoelho as well

stephenswat:
As pointed out in #1557, the existing orthogonal seed finding code uses the `SimSpacePoint` from the ACTS examples namespace, which should not be used in the ACTS core. This commit gets rid of the uses of this type.

Fixes #1557.

### DIFF
--- a/Core/include/Acts/Seeding/SeedFinderOrthogonal.ipp
+++ b/Core/include/Acts/Seeding/SeedFinderOrthogonal.ipp
@@ -506,7 +506,7 @@ void SeedFinderOrthogonal<external_spacepoint_t>::processFromMiddleSP(
    * Create a vector to contain protoseeds.
    */
   std::vector<std::pair<
-      float, std::unique_ptr<const InternalSeed<ActsExamples::SimSpacePoint>>>>
+      float, std::unique_ptr<const InternalSeed<external_spacepoint_t>>>>
       protoseeds;
 
   int numQualitySeeds = 0;


andiwand (approved):

stephenswat:
@paulgessinger this CI is stuck on codecov not being run, but it seems like no codecov job was ever queued. Any idea how I can go about fixing this?

paulgessinger:
Codecov runs externally, it doesn't show up until they post results. Sometimes there's a glitch. I'll merge this manually.

PR #423, opened 2020-09-02
build: use Boost::filesystem

msmk0:
This switches uses of `std::filesystem` to `boost::filesystem`. Using `std::filesystem` is not consistently supported by all compilers that claim to support C++17. Using `boost::filesystem` instead should provide better compatibility across all systems.

This fixes compile issues seen by @krasznaa in the context of the SYCL developments.

PR #242, opened 2020-06-11
Use std::filesystem instead of boost::filesystem

msmk0:
Fixes #193.

PR #1569, opened 2022-10-03
refactor: replace boost filesystem with std

andiwand:
since `std::filesystem` is fairly established by now, this PR switches out `boost::filesystem` completely.

there is no `unique_path` in `std` because of security concerns and I replaced it by a known path in the systems temporary.

this also adds `cmake/FindFilesystem.cmake` because gcc8 will not link to `stdc++fs` or `c++fs` automatically. as soon as gcc8 is not relevant for us or downstream we can remove it.


stephenswat:
Would just like to draw your attention to #242 and #423; i.e. the change from `boost::filesystem` to `std::filesystem` was made before, and then reverted because of problems with `std::filesystem`. Not sure if those issues still exist, but be aware.

Note that these issues were encountered two years ago, so they may very well be fixed now.

andiwand:
thanks for the reference @stephenswat ! after adding `stdc++fs` to the link libraries this seems to work at least for the CI. not sure if it could result in any other downstream issues. cc @paulgessinger 

the only problem is the missing `unique_path` function. I tried to work around but macos complains about the work around. maybe a defined folder name is good enough

andiwand:
oh I fear macos still does not like it https://github.com/acts-project/acts/actions/runs/3174144070/jobs/5170577273#step:6:400

Examples/Detectors/MagneticField/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ target_link_libraries(
   # the ROOT libraries should be private, but if we do that then the linker
   # fails with some missing ROOT symbols.
   PUBLIC ActsCore ActsExamplesFramework ROOT::Core ROOT::Tree
-  PRIVATE Boost::filesystem Boost::program_options)
+  PRIVATE Boost::program_options stdc++fs)
```
paulgessinger, line 15: I don't think hard coding the libname and also specifically the one from libstdc++ is a good idea necessarily.

andiwand:
@krasznaa do you have an opinion on this? [sycl](https://github.com/acts-project/acts/actions/runs/3174652875/jobs/5171711285) is failing right now because [FindFilesystem.cmake](https://github.com/acts-project/acts/pull/1569/files#diff-1ef7efcd62777b6d2d048781d6f097fdd5be311cab1e5252477c8288f32c77e9) does not manage to discover `std::filesystem`

.github/workflows/builds.yml
```diff
@@ -421,7 +421,7 @@ jobs:
       - name: ccache stats
         run: ccache -s
       - name: Unit tests
-        run: ${SETUP} && ${PRELOAD} && cmake --build build --target test
```
andiwand, line 424: @gagnonlg do you remember why this was necessary? it segfaulted because the `std::filesystem` because of standard lib mismatch and removing solved the problem

andiwand:
@paulgessinger mentioned that it might be worth checking if that breaks the athena acts build downstream. I will try to check that tomorrow. otherwise this should be good to go

.github/workflows/builds.yml
andiwand: thanks! looks like it was added here https://github.com/acts-project/acts/pull/871

gagnonlg: I see! I don't quite remember, but I see no problem with the proposed chabge

Examples/CMakeLists.txt
```diff
@@ -4,4 +4,4 @@ add_subdirectory(Framework)
 add_subdirectory(Io)
 add_subdirectory(Run)
 add_subdirectory_if(Python ACTS_BUILD_EXAMPLES_PYTHON_BINDINGS)
-add_subdirectory_if(Scripts ACTS_BUILD_ANALYSIS_APPS)
\ No newline at end of file
+add_subdirectory_if(Scripts ACTS_BUILD_ANALYSIS_APPS)
```
paulgessinger, line 7: Whitespace change only?

andiwand:
athena build worked :+1: ready to merge I think

andiwand:
macos is failing because the python version changed again and root does not like that https://github.com/acts-project/acts/actions/runs/3189311751/jobs/5202975240#step:5:37

will follow up as soon as we have newly built dependencies for macos

gagnonlg:
> macos is failing because the python version changed again and root does not like that https://github.com/acts-project/acts/actions/runs/3189311751/jobs/5202975240#step:5:37

Is this error from the DD4hep cmake file? @paulgessinger I thought we had fixed that already?

paulgessinger:
Yeah I thought so too, but somehow the interplay isn't working the way I thought it would. Recompiling will work now, I think long term we'll just have to include python in out dependencies build so we always build the same version.

acts-project-service:
```
```
```bash
git fetch
# Create a new working tree
# Create a new branch
git switch --create backport-1569-to-develop/v19.x
git cherry-pick -x --mainline 1 f7c5dcedd6271644d32d5e3a492ca4e19cb50cf6
# Push it to GitHub
git push --set-upstream origin backport-1569-to-develop/v19.x
cd ../..
# Delete the working tree
```
Then, create a pull request where the `base` branch is `develop/v19.x` and the `compare`/`head` branch is `backport-1569-to-develop/v19.x`.

PR #1566, opened 2022-09-30
feat: Allow building with system libs by default

stephenswat:
As it stands, the Acts build system will pull quite a bit of software from the internet, and turning this behaviour off entirely currently involves setting quite a number of flags. This commit adds a new CMake option, `ACTS_USE_SYSTEM_LIBS`, which makes using a system installation the default for all dependencies.

PR #1573, opened 2022-10-04
ci: Fix options + docs, add CI job to enforce match

### LINKED ISSUE #1571: ACTS_SETUP_ACTSVG does not seem to do anything
PR #1360 added a CMake option called ACTS_SETUP_ACTSVG that serves no apparent purpose, in the sense that the only hit when searching for it in the upstream acts and actsvg repos is its declaration.

Judging by the differing naming conventions in the two repos, I suspect this might be a leftover from a past development stage. If so, it might be safe to just delete it, but I wanted to check out before just in case...

paulgessinger:
This PR removes obsolete options mentioned in #1571. It also adds a new CI job to check if the CMake options match the documentation in `docs/getting_started.md`.

The `docs/parse_cmake_options.py` has been improved so it can update only the options table (using comment markers), and has a mode that will check for a correct table and fail otherwise. It also learns to resolve defaults through other variables (only one level), as in #1566.

This should help us keep the actual options and documentation in sync.

Fixes #1571


HadrienG2 (approved):

HadrienG2:
You may want to also remove the related _acts_actsdd4hep_version variable while you're at it.

HadrienG2 (approved):

PR #1576, opened 2022-10-06
feat: break loop over triplets in seed finder

LuisFelipeCoelho:
break loop over triplets if reached the last top SP in seed finder to avoid unnecessary calculations


AJPfleger (approved):
lgtm
@LuisFelipeCoelho,  could you update the branch?

PR #1579, opened 2022-10-06
feat: poor man's ambiguity resolution example

andiwand:
my first attempt of a poor man's ambiguity resolution. iteratively evict tracks that do not match our criteria. worst tracks first. when there is nothing left to remove we are done. I fear this will be kinda slow for bigger events.

I am not happy with the python interface. somehow we end up with a lot of different whiteboard entries for tracks which makes it hard to plug and play. following up in https://github.com/acts-project/acts/pull/1588

I also refactored the `TrackSelector` and removed the `trackIndices`


benjaminhuth:
Thanks, great to get something like this now in a first version!

However, one question regarding the algorithm interface... I'm not totally sure if I miss something important here, but the `inputTrajectories` normally contain already track parameters, right? Thus I do not really understand why they come also as an extra input...

Wouldn't be an algoriithm interface possible that is just:

```c++
class AmbiguityResolutionAlgorithm final : public BareAlgorithm {
 public:
  struct Config {
    /// Input source links collection.
    std::string inputSourceLinks;
    /// Input trajectories collection.
    std::string inputTrajectories;
    /// Output trajectories
    std::string outputTrajectories;

    ///
    std::uint32_t maximumSharedHits = 1;
  };
```

That than internally makes copies of the `Acts::MultiTrajectory` and changes the indices & parameters. But maybe I also understand something completely wrong here...

andiwand:
@paulgessinger I guess the interface can be simplified as soon as we have a track entity? do you see simplification opportunities right now?

paulgessinger:
Yes, the Track class will combine nominal fitted track parameters + track states (= called trajectory right now)

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
```diff
@@ -0,0 +1,175 @@
...
+      relativeSharedHits[i] = 1.0f * sharedHits[i] / hitCount[indexTrack];
+    }
+
+    auto it = std::max_element(std::begin(relativeSharedHits),
+                               std::end(relativeSharedHits));
+    if (it == std::end(relativeSharedHits)) {
+      break;
+    }
+
+    const auto index = it - std::begin(relativeSharedHits);
```
paulgessinger, line 154: I guess this should be `std::distance`?

andiwand:
truth tracking did not change as expected

vertexing looks different I think?

image

CKF output can be ignored and passes anyway

IVF output is very different. even the truth IVF which should not be the case?

will look into this tomorrow but I think we are getting close

CI/physmon/physmon.py
```diff
@@ -175,21 +177,27 @@
             trackingGeometry,
             field,
             CKFPerformanceConfig(ptMin=400.0 * u.MeV, nMeasurementsMin=6),
+            TrackSelectorRanges(
+                removeNeutral=True,
+                absEta=(None, 2.5),
+                loc0=(None, 4.0 * u.mm),
+                pt=(500 * u.MeV, None),
+            ),
```
paulgessinger, line 185: What's the default here? Why don't we go higher and select like before?

CI/physmon/physmon.py
```diff
@@ -298,21 +306,27 @@
                 trackingGeometry,
                 field,
                 CKFPerformanceConfig(ptMin=400.0 * u.MeV, nMeasurementsMin=6),
+                TrackSelectorRanges(
+                    removeNeutral=True,
+                    absEta=(None, 2.5),
+                    loc0=(None, 4.0 * u.mm),
+                    pt=(500 * u.MeV, None),
+                ),
```
paulgessinger, line 314: Does this select tracks or particles?

CI/physmon/physmon.py
```diff
@@ -175,21 +177,27 @@
...
             s,
             field,
-            TrackSelectorRanges(
-                removeNeutral=True,
-                absEta=(None, 2.5),
-                loc0=(None, 4.0 * u.mm),
-                pt=(500 * u.MeV, None),
-            ),
             vertexFinder=VertexFinder.Iterative,
-            trajectories="trajectories",
```
paulgessinger, line 192: Does it still run the correct truth matching in that case?

andiwand: hmm looking at this now I am not sure it ever did because I though it would only do that if `trajectories` are not present. the param is defaulted so the code should be doing the same thing

CI/physmon/physmon.py
andiwand: I just moved the track selection from the vertexing to CKF because it made more sense to me. similar to seeding where we also filter in the `addSeeding` command

CI/physmon/physmon.py
andiwand: I just moved this from the vertexing to the CFK so it should do the same as before

andiwand:
well that looks weird :thinking: 

ckf duplications

image

ambi duplications

image

andiwand:
that issue above is fixed now. the only problem now is the difference for the vertexing with truth tracking. I just tried this local and it looks like we do not use the 1:1 matching and this is where the difference comes from. @paulgessinger I guess this should be fixed?

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TrackSelector.cpp
```diff
@@ -43,7 +43,7 @@ ActsExamples::ProcessCode ActsExamples::TrackSelector::execute(
     const bool validCharge = validNeutral or validCharged;
     return validCharge and
            within(trk.transverseMomentum(), m_cfg.ptMin, m_cfg.ptMax) and
-           within(std::abs(theta), m_cfg.absEtaMin, m_cfg.absEtaMax) and
```
paulgessinger, line 48: Arguably, this should be a separate PR.

andiwand: yeah I fear I bundled a few things together here to make this work :smile:

andiwand: sourced this out to https://github.com/acts-project/acts/pull/1615

benjaminhuth (changes requested):
Looks good! Most of the comments are really minor things and should be easy to fix :)

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
```diff
@@ -0,0 +1,203 @@
...
+  }
+  if (m_cfg.outputTrackParametersTips.empty()) {
+    throw std::invalid_argument(
+        "Missing track parameters tips output collection");
+  }
+}
+
+namespace {
+
+std::vector<std::size_t> computeSharedHits(
```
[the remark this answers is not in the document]
andiwand: the problem with the one from the track finding was that it has a different interface than I require. I guess we can generalize this in the core at some point. but for now it seems easier to have a duplicate

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
```diff
@@ -0,0 +1,203 @@
...
+        std::begin(relativeSharedHits), std::end(relativeSharedHits));
+    const auto index =
+        std::distance(std::begin(relativeSharedHits), maxRelativeSharedHits);
+    trackIndices.erase(std::begin(trackIndices) + index);
+  }
+
+  if (trackIndices.size() == trackParameters.size()) {
+    const auto sharedHits =
+        computeSharedHits(sourceLinks, trajectories, trackIndices, trackTips);
+    std::cout << "ambi: ";
```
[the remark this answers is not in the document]
andiwand: upsi those are debug remains. removing them

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TrackSelector.cpp
```diff
@@ -57,31 +55,51 @@ ActsExamples::ProcessCode ActsExamples::TrackSelector::execute(
                   m_cfg.timeMax);
   };
 
+  const bool doTips = !m_cfg.inputTrackParametersTips.empty();
+
   // prepare input and output containers
   const auto& inputTrackParameters =
       ctx.eventStore.get<TrackParametersContainer>(m_cfg.inputTrackParameters);
+  const std::vector<std::pair<size_t, size_t>>* inputTrackParametersTips =
```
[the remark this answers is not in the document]
andiwand: the reason I used a pointer here is that we do not make a copy in that case. but that might not be critical anyways

Examples/Python/src/TrackFinding.cpp
```diff
@@ -296,6 +297,29 @@ void addTrackFinding(Context& ctx) {
                      std::pair<GeometryIdentifier, MeasurementSelectorCuts>>>())
             .def(py::init(constructor));
   }
+
+  {
```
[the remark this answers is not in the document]
andiwand: oh yeah good point. it wasn't merged when I started working on this

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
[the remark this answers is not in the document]
benjaminhuth: Okay, makes sense! then maybe just add a comment with a `TODO` so that it does not get forgotten then?

benjaminhuth (approved):
Looks good, let's get this in!

paulgessinger (approved):

PR #1581, opened 2022-10-06
refactor: Improve verbose logging in Kalman Filter

benjaminhuth:
This is a super small PR, that additionally prints some information in the KF verbose logger, I believe the current position of the propagation is a missing information there. I need this very often, e.g. during the debugging of the GSF and have it there in all my local branches.

I would like it to add this to the main branch, since it shouldn't cost performance and then I do not need to add this always manually.


Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -325,7 +325,10 @@ class KalmanFitter {
         return;
       }
 
-      ACTS_VERBOSE("KalmanFitter step");
+      ACTS_VERBOSE("KalmanFitter step at pos: "
+                   << stepper.position(state.stepping).transpose()
```
andiwand, line 329: thanks for transposing them. this saves a lot of log lines

andiwand (approved):

Core/include/Acts/TrackFitting/KalmanFitter.hpp
benjaminhuth: I wonder if anyone ever needed the non-transposed print-out of an `Eigen::Vector` :D

PR #1582, opened 2022-10-07
feat: move rRangeSPExtent to BinnedSPGroup

LuisFelipeCoelho:
This would avoid unnecessary iterations over the SPs in **Athena** `SeedingTool`


LuisFelipeCoelho:
FYI: @paulgessinger

PR #1588, opened 2022-10-07
feat: add whiteboard aliases

andiwand:
I noticed that a lot of information on the whiteboard rolls from one algorithm to the next but requires a different name in the process. that makes it hard to use the correct name downstream.

this PR adds an alias mechanism to the sequencer. the alias can be changes from one entry to the next. when queried the sequencer will return the entry behind the alias.


Examples/Framework/include/ActsExamples/Framework/WhiteBoard.hpp
```diff
@@ -95,21 +100,42 @@ inline void ActsExamples::WhiteBoard::add(const std::string& name, T&& object) {
...
+  ACTS_VERBOSE("Retrieved object '" << resolvedName << "'");
   return castedHolder->value;
 }
 
 inline bool ActsExamples::WhiteBoard::exists(const std::string& name) const {
-  return m_store.find(name) != m_store.end();
-}
\ No newline at end of file
+  return (m_store.find(name) != m_store.end()) ||
+         (m_alias.find(name) != m_alias.end());
```
paulgessinger, line 120: This would also have to do the resolution to be correct no? I.e. if an alias exists, but the value the alias points at does not, this function should return false.

paulgessinger: Ah never mind, when you set an alias it checks if the underlying key exists.

paulgessinger:
I'm wondering if it might be easier if you could set the aliases statically for all events from python. That way, you wouldn't check if the alias target exists when creating the alias, but rather when trying to resolve it.

Examples/Python/python/acts/examples/simulation.py
```diff
@@ -12,6 +12,7 @@
     ParticlesPrinter,
     RootParticleWriter,
 )
+from .utility import addAlias
```
paulgessinger, line 15: I would prefer absolute imports

andiwand:
> I'm wondering if it might be easier if you could set the aliases statically for all events from python. That way, you wouldn't check if the alias target exists when creating the alias, but rather when trying to resolve it.
> 
> I guess this would be tricky though because the whiteboard is created event-by-event...

yeah I am also not really happy with the overhead this introduced but I don't really see an alternative. if we would do it statically we would loose the evolution of the alias e.g. `particles -> particles_initial` and then `particles -> particles_selected`. we would need a mechanism to remember the ordering.

paulgessinger:
Another option would be to add a convenience function to `Sequencer` which internally attaches this algorithm but improves on the interface. This could even be at the python binding level I guess. I'm thinking something like

```python
s.addAlias(particles="particles_initial")
```

andiwand:
hmm that sounds interesting. so I guess if you need the link to the algorithm/reader/writer it would look like this?

```python
s.addAlias(algo, particles="particles_initial")
```
but `addAlias` would need to be a C++ implementation no? or how would you imagine an implementation at the python binding level?

andiwand:
I think it would be good to put https://github.com/acts-project/acts/pull/1579 in first and then catch up here

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -825,13 +820,18 @@ def addCKFTracks(
 def addTrackSelection(
     s: acts.examples.Sequencer,
     trackSelectorRanges: TrackSelectorRanges,
-    inputTrackParameters: str,
-    inputTrackParametersTips: str,
-    outputTrackParameters: str,
-    outputTrackParametersTips: str,
+    inputTrackParameters: Optional[str] = None,
```
benjaminhuth, line 823: Hmm I'm a bit worried about this interface. It is not entirely straight forward that not providing a parameter causes the function to use a hardcoded string as alias... The minimum solution would be to document this in the docstring I would say.

However, is there a reason why not calling `resolveAlias` in the "main" script? Then the add functions would still be just dealing with reagular keys, what also would make debugging easier I guess.

This would also allow a kind of "don't use alias, so don't bother with them" style. In this case it could happen that a user is confronted with the alias system, even without using it.

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -862,6 +862,9 @@ def addTrackSelection(
 
     s.addAlgorithm(trackSelector)
 
+    s.addAlias("trackParameters", trackSelector.config.outputTrackParameters)
```
benjaminhuth, line 865: also, this behaviour should be documented I guess

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -1049,11 +1058,19 @@ def addVertexFitting(
         RootVertexPerformanceWriter,
     )
 
+    if trackParameters is None:
+        trackParameters = s.resolveAlias("trackParameters")
```
benjaminhuth, line 1062: same as above

benjaminhuth (changes requested):

benjaminhuth:
Looks good in general :) I just think we should address this one point before merging...

Examples/Python/python/acts/examples/reconstruction.py
benjaminhuth: Okay actually with a second thought I totally agree with you, the add functions should try to connect to each other without need of interaction in the default case

I think it is really a matter of documentation to make the structure clear. E.g. somehow like this: "This function resolves and overwrites the aliases "trackParameters" and "trackParameterTips". To overwrite this behaviour, pass the inputTackParameters and foobar directely."

benjaminhuth: What would be also nice is maybe a checking that gives the user useful hints: "Alias 'foo' not found. Pass 'bar' explicitly if you don't want to use the alias system" or something like that

andiwand:
I refactored this again. the aliases are added to the sequencer which is adding it to the whiteboard. as soon as an aliased entry is added to the whiteboard it will create a second entry with the aliased name. (this also preserves your string distance mechanism @benjaminhuth ). I have to use a shared pointer instead of a unique pointer in the whiteboard but that should be fine.

let me know what you think of this version @paulgessinger @timadye @benjaminhuth

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -83,6 +84,11 @@ void ActsExamples::Sequencer::addWriter(std::shared_ptr<IWriter> writer) {
   ACTS_INFO("Added writer '" << m_writers.back()->name() << "'");
 }
 
+void ActsExamples::Sequencer::addAlias(const std::string& from,
+                                       const std::string& to) {
+  m_aliases[to] = from;
```
benjaminhuth, line 89: This allows "overwriting" von aliasses. This is fine in principle, but it might lead to unexpected behaviour. From the function signature it could be assumed, that it is possible to have multiple aliases to a specific whiteboard entry, but that is not the case (because of the unique keys in the map).
Baybe it would be better to use something like 

```suggestion
  auto [it, success] = m_aliases.insert({to, from});
  if( not success ) {
    throw std::runtime_error("Alias to '" + to + "' already set");
  }
```

benjaminhuth (approved):
One small thing I think could be improved to reduce user errors, but you can also ignore it since its just a minor thing.
Anyways, looks good so lets get this in asap :)

Examples/Framework/src/Framework/Sequencer.cpp
timadye: Here `from` is the alias name and `to` is the name of the object it refers to, right? Would `aliasName` and `objectName` be clearer argument names?

I'm a bit confused about the `map[to]=from` (rather than how I'd naively expect `map[from]=to`, but I guess that's just how it is needed in the subsequent implementation (I didn't look at that yet).

If that's the case, couldn't we reasonably need multiple aliases pointing to the same object? Eg. different algorithms using the same input, but separately using aliases. I guess this bit could be handled with a `std::multimap`, but don't know if the rest of the implementation could cope with this situation.

andiwand: that is a good point. for now I would limit it explicitly that an object name can only be aliased once. whenever we see a need for enhancing this it should be easy to enable this, for example with a multimap

benjaminhuth: I would also go with unique solution right now, to keep complexity as low as possible.

benjaminhuth (approved):
Lets get this in!

PR #1595, opened 2022-10-12
feat: Vertexing mu scan and CI job that creates it

paulgessinger:
This PR adds a configuration + plotting script for this kind of plot produced by the CI:

vertexing

This currently runs on 25 events, but we might have to tune this to keep the runtime reasonable.


AJPfleger (approved):
lgtm

paulgessinger:
I made mistake with where I put the output file. Need to fix that before merging.

paulgessinger:
Ready now, can you reapprove @AJPfleger?

AJPfleger (approved):
lgtm
thanks for restricting the `$ rm` in the last commit.

PR #1597, opened 2022-10-13
feat: Make whiteboard suggest names on mismatch

benjaminhuth:
This little feature addresses something which annoys me quite often: I misspell the name of whiteboard entries in scripts. This PR let the whiteboard suggest alternatives on mismatch based on the *Levenshtein distance* for words:

Screenshot_20221013_153039

Might not be very important, but was fun to implement and also could be useful for developers :)


paulgessinger (approved):
🤣  nice

PR #1601, opened 2022-10-17
feat: deltaZ cut in SeedFinderOrthogonal

LuisFelipeCoelho:
same cut included in default seeding #1209


AJPfleger (approved):
good that the code coverage did not drop this time :)

PR #1605, opened 2022-10-19
fix: orthogonal seeding configuration with Python bindings

LuisFelipeCoelho:
The `SeedingOrthogonalAlgorithm` configuration is overwriting the configuration from the Python modules in `reconstruction.py`. I do not think the configuration needs to be defined in `SeedingOrthogonalAlgorithm.hpp` as well, since it is already defined in `SeedFinderOrthogonalConfig.hpp`. 

Also, I removed from `reconstruction.py` some features that are not yet available in orthogonal seeding or that will not be inplemented (for example, features related to the grid).


andiwand (approved):
looks good to me

PR #1606, opened 2022-10-20
ci: update macos deps

andiwand:
with https://github.com/acts-project/ci-dependencies/pull/12 I added a specific python version to the macos ci dependencies which hopefully resolves the root/dd4hep issue for now


andiwand:
looks like this works @paulgessinger :tada:

paulgessinger (approved):
lgtm

PR #1609, opened 2022-10-20
refactor!: Put GSF in namespace Acts::Experimental

benjaminhuth:
Since the GSF is not yet fully functional, it is put in the namespace `Acts::Experimental`.

Also adds some hints in the documentation, that components in this namespace are not yet production read, and also not part of the public API (I think that is reasonable, since clients should not rely on experimental code).

However, since this would be a new rule, this change is breaking.


andiwand (approved):
looks good to me. not sure if anybody else wants to take a look as this is a breaking change. cc @asalzburger @paulgessinger

benjaminhuth:
Maybe we should also quickly discuss this in tomorrows developer meeting, since it proposes changes to the definition of the public API / the versioning...

andiwand (approved):

asalzburger (approved):
Reapproving, after AStefl had looked over it

PR #1612, opened 2022-10-21
ci: Add reporting to physmon

paulgessinger:
This attempts to add a workflow that comments details of the physmon result on the relevant PR.


andiwand (approved):
:+1: nice

PR #1614, opened 2022-10-24
chore: Update physmon template

paulgessinger:
Adds IVF plots, a bit of restructuring


andiwand (approved):

andiwand (approved):

paulgessinger:
@andiwand It's strange, this bump didn't change the numpy version requirements. I don't understand why this fails.

andiwand:
hmm maybe github is messing with the installed versions of numpy and so on. I just check a different run and it was fine. I am sure I saw an error like that today already but it was able to continue. @paulgessinger

PR #1617, opened 2022-10-24
refactor: Restructure track fitting functions

benjaminhuth:
This does
* Decouple the track fitting functions from the `TrackFittingAlgorithm`, so no static member functions anymore, rather free functions. I think that is better, because it is possible to add new fitters without touching the algorithm file etc...
* Reduce number of track fitter functions from 2*N to N (no seperate function anymore for direct fitter). The selection is now done by checking a `std::optional<std::vector<const Surface *>>` in the function arguments. I think this reduces code duplication etc.

*Should* not increase build memory consumption, since everything stays separated as before in the different compilation units.

**Note:**
Also I think the direct track fitting is broken anyways in this algorithm, but that should be addressed seperately. A small problem I think i fixed in this PR, but in general this only works properly if there are no material surfaces that are non-measurement surfaces (because only these are added to the surface sequence). Maybe we should discuss this somehow. We do not test direct fitting right now, right?


Examples/Algorithms/TrackFitting/include/ActsExamples/TrackFitting/TrackFittingAlgorithm.hpp
```diff
@@ -52,27 +50,14 @@ class TrackFittingAlgorithm final : public BareAlgorithm {
...
-  /// @note This is separated into a virtual interface to keep compilation units
-  /// small
-  class DirectedTrackFitterFunction {
-   public:
-    virtual ~DirectedTrackFitterFunction() = default;
-    virtual TrackFitterResult operator()(
-        const std::vector<std::reference_wrapper<const IndexSourceLink>>&,
-        const TrackParameters&, const GeneralFitterOptions&,
-        const std::vector<const Acts::Surface*>&,
+        const std::optional<std::vector<const Acts::Surface*>>&,
```
paulgessinger, line 60: This could also be a separate overload, instead of an `optional` argument. Might be more obvious?

benjaminhuth: Yeah indeed, might be better!

Examples/Algorithms/TrackFitting/include/ActsExamples/TrackFitting/TrackFittingAlgorithm.hpp
```diff
@@ -30,11 +30,9 @@ namespace ActsExamples {
 
 class TrackFittingAlgorithm final : public BareAlgorithm {
  public:
-  /// Track fitter function that takes input measurements, initial trackstate
-  /// and fitter options and returns some track-fitter-specific result.
-  using TrackFitterOptions =
-      Acts::KalmanFitterOptions<Acts::VectorMultiTrajectory>;
-
+  // All track fitter function must return the same type. For now this is the
```
AJPfleger, line 33:
```suggestion
  // All track fitter functions must return the same type. For now this is the
```

andiwand (approved):

andiwand (approved):

PR #1619, opened 2022-10-25
chore: auto cancel

andiwand:
kill previous CI runs when pushing a new commit


.github/workflows/analysis.yml
```diff
@@ -12,6 +12,10 @@ on:
     paths-ignore:
       - "docs/**"
 
+concurrency: 
+  group: ${{ github.ref }}
```
andiwand, line 16: only scoped for the ref. should be enough right @paulgessinger ?

paulgessinger: This kills all workflows for the ref, right?

andiwand: yeah this is how I understand it

paulgessinger (approved):

PR #1620, opened 2022-10-25
feat: port RootNTupleReader

andiwand:
port of Thomas' RootNTupleReader
- https://github.com/ttlindkvist/ttlindkvistACTS/blob/main/ttlindkvistACTS/include/ttlindkvistACTS/RootNTupleReader.hpp
- https://github.com/ttlindkvist/ttlindkvistACTS/blob/main/ttlindkvistACTS/src/RootNTupleReader.cpp

compiles but is untested


Examples/Io/Root/src/RootNTupleReader.cpp
```diff
@@ -0,0 +1,417 @@
...
+
+  const unsigned int nTracks = m_branches.m_track_d0->size();
+  const unsigned int nTruthVtx = m_branches.m_truthvertex_z->size();
+  const unsigned int nRecoVtx = m_branches.m_recovertex_z->size();
+
+  ACTS_DEBUG("nTracks = " << nTracks);
+  ACTS_DEBUG("nTruthVtx = " << nTruthVtx);
+  ACTS_DEBUG("nRecoVtx = " << nRecoVtx);
+
+  std::vector<Acts::BoundTrackParameters> trackContainer;
```
benjaminhuth, line 299: I wonder if we should use the typedefs like `ActsExamples::TrackParametersContainer` etc. to stick to the convention.
Same below for the other vectors if possible?

Examples/Io/Root/src/RootNTupleReader.cpp
```diff
@@ -0,0 +1,417 @@
...
+// truthvertex_t        | std::vector<float>       | AsJagged(AsDtype('>f4'), he...
+// truthvertex_isPU     | std::vector<int32_t>     | AsJagged(AsDtype('>i4'), he...
+// truthvertex_isHS     | std::vector<int32_t>     | AsJagged(AsDtype('>i4'), he...
+// jet_tracks_idx       | std::vector<std::vect... | AsObjects(AsVector(True, As...
+// recovertex_tracks... | std::vector<std::vect... | AsObjects(AsVector(True, As...
+// truthvertex_track... | std::vector<std::vect... | AsObjects(AsVector(True, As...
+// clang-format on
+
+ActsExamples::RootNTupleReader::BranchPointerWrapper::BranchPointerWrapper() {
+  m_track_d0 = new std::vector<float>;
```
benjaminhuth, line 109: Is there a reason why we need to allocate these on the heap? If so, I would strongly argue to use `std::unique_ptr`.
However, I don't see a reason why these vectors cannot live on the stack, if this is possible, that would be the preferable solution.

Examples/Io/Root/src/RootNTupleReader.cpp
```diff
@@ -0,0 +1,417 @@
...
+  beamspotConstraint.setPosition(beamspotPos);
+  beamspotConstraint.setCovariance(beamspotCov);
+
+  context.eventStore.add(m_cfg.outputTrackParameters,
+                         std::move(trackContainer));
+  context.eventStore.add(m_cfg.outputTruthVtxParameters,
+                         std::move(truthVertexContainer));
+  context.eventStore.add(m_cfg.outputRecoVtxParameters,
+                         std::move(recoVertexContainer));
+  context.eventStore.add(m_cfg.outputBranchPointerWrapper, &m_branches);
```
benjaminhuth, line 409: We seem to add a pointer here to the whiteboard... this is problematic from a livetime perspective, right?

Actually I do not understand why we write this to the whiteborad at all, since we take the things from `m_branches` before and add them to the other containers...

Examples/Io/Root/include/ActsExamples/Io/Root/RootNTupleReader.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+    std::string outputTrackParameters = "nTupleTrackParameters";
+    std::string outputTruthVtxParameters = "nTupleTruthVtxParameters";
+    std::string outputRecoVtxParameters = "nTupleRecoVtxParameters";
+    std::string outputBranchPointerWrapper = "nTupleBranchPointerWrapper";
+    std::string outputBeamspotConstraint = "beamspotConstraint";
+  };
+
+  struct BranchPointerWrapper {
+    std::vector<float> *m_track_d0;
```
benjaminhuth, line 42: As mentioned in the other comment, I think these vectors can simply live on the stack.

Examples/Io/Root/include/ActsExamples/Io/Root/RootNTupleReader.hpp
```diff
@@ -0,0 +1,142 @@
...
+#include "ActsExamples/Framework/ProcessCode.hpp"
+
+#include <mutex>
+#include <vector>
+
+class TChain;
+
+namespace ActsExamples {
+
+class RootNTupleReader : public ActsExamples::IReader {
```
benjaminhuth, line 25: Actually now that I look at it I'm a bit unhappy with the name, as it does not express the content of the N tuple... Or is in our context RootNTuple a common term that differs from the [ROOT type](https://root.cern.ch/doc/master/classTNtuple.html)?

Examples/Io/Root/include/ActsExamples/Io/Root/RootNTupleReader.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+  /// Constructor
+  /// @param config The Configuration struct
+  RootNTupleReader(const Config &config, Acts::Logging::Level level);
+
+  /// Destructor
+  ~RootNTupleReader();
+
+  /// Framework name() method
+  std::string name() const final override { return "RootNTupleReader"; }
```
benjaminhuth, line 97: If there is a better name, also adjust here

Examples/Io/Root/include/ActsExamples/Io/Root/RootNTupleReader.hpp
```diff
@@ -0,0 +1,142 @@
...
+  std::mutex m_read_mutex;
+
+  /// The number of events
+  std::size_t m_events = 0;
+
+  /// The input tree name
+  TChain *m_inputChain = nullptr;
+
+  /// Event identifier.
+  std::int32_t m_eventNumber;
```
benjaminhuth, line 133: It seems to me this is not used at all. However, if we want to write to it, I would argue to put it locally in the read function.

Examples/Io/Root/include/ActsExamples/Io/Root/RootNTupleReader.hpp
```diff
@@ -0,0 +1,142 @@
...
+  /// The entry numbers for accessing events in increased order (there could be
+  /// multiple entries corresponding to one event number)
+  std::vector<long long> m_entryNumbers = {};
```
benjaminhuth, line 136: This seems to be only used in the Constructor, so I would argue to move it there as a local variable.

benjaminhuth (changes requested):
There are a few things in the code I would like to have changed, however, if its to much work, reach out to me and I approve anyways, as this is not used at the moment.

Examples/Io/Root/include/ActsExamples/Io/Root/RootNTupleReader.hpp
andiwand: I guess it should be a track param reader so maybe something like `RootTrackParameterReader` ?

Examples/Io/Root/include/ActsExamples/Io/Root/RootNTupleReader.hpp
andiwand: hmmm not sure but I think I saw this a few times with root. the pointer is later passed by reference to root so it can be redirected to the data in the event.

I think we also have this with other root code in acts :thinking:

Examples/Io/Root/src/RootNTupleReader.cpp
andiwand: I think this code is useless. even if we use pointers we do not need to initialize them with an empty vector

Examples/Io/Root/src/RootNTupleReader.cpp
andiwand: makes sense :+1: 

the other vector would be `std::vector<Acts::Vector4>`. do we have a common typedef for these?

Examples/Io/Root/src/RootNTupleReader.cpp
andiwand: true not sure what this was used for. I removed it

Examples/Io/Root/src/RootNTupleReader.cpp
```diff
@@ -0,0 +1,272 @@
...
+ActsExamples::ProcessCode ActsExamples::RootNTupleReader::read(
+    const ActsExamples::AlgorithmContext& context) {
+  ACTS_DEBUG("Trying to read track parameters from ntuple.");
+
+  Acts::Vector3 pos(0, 0, 0);
+  std::shared_ptr<Acts::PerigeeSurface> surface =
+      Acts::Surface::makeShared<Acts::PerigeeSurface>(pos);
+
+  if (context.eventNumber >= m_events) {
+    ACTS_ERROR("event out of bounds");
```
benjaminhuth, line 137: One last point: could we add a `return ActsExamples::ProcessCode::ABORT` here? (or whatever the error exit code is?)

Examples/Io/Root/src/RootNTupleReader.cpp
benjaminhuth: I don't know of one, I would just leave it then.

benjaminhuth (approved):
Okay, let's get it in!

andiwand:
@benjaminhuth I just added a python binding

ci seems to crash for whatever reason

benjaminhuth (approved):
nice

benjaminhuth:
yeah because of this I added & removed the `automerge` label to ensure that the order is correct... Does this `automerge priority` label work actually?

paulgessinger:
Not entirely sure.

PR #1621, opened 2022-10-25
feat: variable r range in orthogonal seeding

LuisFelipeCoelho:
This was included in the orthodox seeding in #1084


Core/include/Acts/Seeding/SeedFinderOrthogonal.ipp
```diff
@@ -680,6 +682,29 @@ void SeedFinderOrthogonal<external_spacepoint_t>::createSeeds(
...
+      if (rM > rMiddleMaxSPRange) {
+        continue;
+      }
+    } else {
+      if (rM > m_config.rMaxMiddle) {
+        continue;
+      }
+      if (rM < m_config.rMinMiddle) {
+        continue;
+      }
```
AJPfleger, line 705:
```suggestion
      if (rM < rMiddleMinSPRange || rM > rMiddleMaxSPRange) {
        continue;
      }
    } else {
      if (rM < m_config.rMinMiddle || rM > m_config.rMaxMiddle) {
        continue;
      }
```

Do you think it could easier to read like that?

Anyhow, I would suggest to switch the `if`s in the `else` to ask first for the min-condition and then the max-condition.

Core/include/Acts/Seeding/SeedFinderOrthogonalConfig.hpp
```diff
@@ -64,6 +64,14 @@ struct SeedFinderOrthogonalConfig {
...
   float rMin = 33 * Acts::UnitConstants::mm;
 
+  // radial range for middle SP
+  // variable range based on SP radius
+  bool useVariableMiddleSPRange = true;
+  float deltaRMiddleMinSPRange = 10. * Acts::UnitConstants::mm;
+  float deltaRMiddleMaxSPRange = 10. * Acts::UnitConstants::mm;
+  // range defined in vector for each z region
+  std::vector<std::vector<float>> rRangeMiddleSP;
+  // range deined by rMinMiddle and rMaxMiddle
```
AJPfleger, line 74:
```suggestion
  // range defined by rMinMiddle and rMaxMiddle
```

Core/include/Acts/Seeding/SeedFinderOrthogonal.ipp
AJPfleger: Maybe you want also have a look into the Range functions in `Core/include/Acts/Utilities/Range1D.hpp` as @benjaminhuth just told me.

LuisFelipeCoelho: Thanks for the comments. I changed the rMiddleMinSPRange to range1D in the default seeding recently but forgot to change it here 😵‍💫. Thanks for noticing it. 

Regarding the `if (rM < rMiddleMinSPRange || rM > rMiddleMaxSPRange) {` this might have to change in the future with the sorting of SPs but for now I agree it is easier to read.

LuisFelipeCoelho: I changed 'rMiddleSPRange' to range1D to keep it consistent with the default seeding. But I think since there are several variables in the SeedFinder configuration that could be replaced by a range1D, including `rMinMiddle` and `rMaxMiddle` I will follow up this PR with this specific change

AJPfleger (approved):

PR #1622, opened 2022-10-25
refactor!: std::unique_ptr to std::optional in Acts::PropagatorResult

benjaminhuth:
This changes the type that stores the parameters and covariance matrix in the `Acts::PropagatorResult`. 

I'm not sure why we have `std::unique_ptr` here, maybe this is from pre-C++17 where `std::optional` was not available? However, both from usibility (not copy-able) and from performance perspective (heap allocation) this does not make sense to me, but maybe I forget something...


Core/include/Acts/Propagator/Propagator.hpp
```diff
@@ -46,11 +46,11 @@ struct PropagatorResult : private detail::Extendable<result_list...> {
   /// Accessor to additional propagation quantities
   using detail::Extendable<result_list...>::get;
 
-  /// Final track parameters - initialized to null pointer
-  std::unique_ptr<parameters_t> endParameters = nullptr;
+  /// Final track parameters
+  std::optional<parameters_t> endParameters = std::nullopt;
```
andiwand, line 50: I think you could drop the `std::nullopt` since this is the default constructed value. but it may be more consistent across the repo to have it

andiwand (approved):
I guess the only advantage of `std::unique_ptr` is that the struct is more densely packed. but that doesn't look like a critical property here

Core/include/Acts/Propagator/Propagator.hpp
benjaminhuth: Yeah I left it because the same argument in principle works for the `std::unique_ptr`, and I think it is there for explcitness (if thats a word^^)... But you're right, maybe I drop it

benjaminhuth:
@andiwand After fixing the unittests etc, could you reapprove
@paulgessinger This should be good to go in now, so would be cool to wait with the next major release until this is in.

andiwand (approved):

PR #1627, opened 2022-10-26
refactor: Bethe-Heitler approximation and configuration

benjaminhuth:
This PR refactors a bit the Bethe Heitler Approximation
* moving it out of namespace `detail`, since its in fact part of the API of the GSF
* allow loading parameterizations from the python examples

Already contains #1609


Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
```diff
@@ -0,0 +1,322 @@
...
+  constexpr static double noChangeLimit = 0.0001;
+  constexpr static double singleGaussianLimit = 0.002;
+  constexpr static double lowerLimit = 0.10;
+  constexpr static double higherLimit = 0.20;
+
+ private:
+  Data m_low_data;
+  Data m_high_data;
+  bool m_low_transform;
+  bool m_high_transform;
```
andiwand, line 114: should we default those?

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
```diff
@@ -0,0 +1,322 @@
...
+
+    // Return no change
+    if (x < noChangeLimit) {
+      std::array<detail::GaussianComponent, NComponents> ret{};
+
+      ret[0].weight = 1.0;
+      ret[0].mean = 1.0;  // p_initial = p_final
+      ret[0].var = 0.0;
+
+      return ret;
```
andiwand, line 189: what happens to the other components?

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
```diff
@@ -0,0 +1,322 @@
...
+      std::array<detail::GaussianComponent, NComponents> ret{};
+      ret[0] = BetheHeitlerApproxSingleCmp::mixture(x)[0];
+      return ret;
+    }
+    // Return a component representation for lower x0
+    if (x < lowerLimit) {
+      return make_mixture(m_low_data, x, m_low_transform);
+    }
+    // Return a component representation for higher x0
+    else {
```
andiwand, line 202: else could be dropped

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
```diff
@@ -0,0 +1,322 @@
...
+
+      if (PolyDegree != degree) {
+        throw std::invalid_argument("Wrong wrong polynom order in '" +
+                                    filepath + "'");
+      }
+
+      if (!transform_code) {
+        throw std::invalid_argument("Transform-code is required in '" +
+                                    filepath + "'");
+      }
```
andiwand, line 243: does this mean we only support `transform_code = false`?

Examples/Algorithms/TrackFitting/src/GsfFitterFunction.cpp
```diff
@@ -106,23 +111,38 @@ std::shared_ptr<TrackFittingAlgorithm::TrackFitterFunction>
...
+      return Acts::Experimental::makeDefaultBetheHeitlerApprox();
+    } else if (std::filesystem::exists(lowBetheHeitlerPath) &&
+               std::filesystem::exists(highBetheHeitlerPath)) {
+      return Acts::Experimental::AtlasBetheHeitlerApprox<6, 5>::loadFromFile(
+          lowBetheHeitlerPath, highBetheHeitlerPath);
+    } else {
+      throw std::invalid_argument(
+          "Paths to bethe heitler parameterization do not exist. Pass empty "
+          "strings to load a default parameterization");
+    }
```
andiwand, line 130: wouldn't it be enough to just have the return statement as this would bubble up an exception anyway? if you want to be more descriptive of the error you can catch and throw again

Examples/Algorithms/TrackFitting/src/GsfFitterFunction.cpp
```diff
@@ -106,23 +111,38 @@ std::shared_ptr<TrackFittingAlgorithm::TrackFitterFunction>
 ActsExamples::makeGsfFitterFunction(
     std::shared_ptr<const Acts::TrackingGeometry> trackingGeometry,
     std::shared_ptr<const Acts::MagneticFieldProvider> magneticField,
+    std::string lowBetheHeitlerPath, std::string highBetheHeitlerPath,
```
andiwand, line 114: should we box these string in an optional to be more explicit?

andiwand (commented):
looks good :+1: I left a few comments. ping me when you are ready for a second round

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -15,10 +15,10 @@
 #include "Acts/Material/ISurfaceMaterial.hpp"
 #include "Acts/Surfaces/CylinderSurface.hpp"
 #include "Acts/Surfaces/Surface.hpp"
+#include "Acts/TrackFitting/BetheHeitlerApprox.hpp"
 #include "Acts/TrackFitting/GsfError.hpp"
 #include "Acts/TrackFitting/GsfOptions.hpp"
 #include "Acts/TrackFitting/KalmanFitter.hpp"
-#include "Acts/TrackFitting/detail/BetheHeitlerApprox.hpp"
```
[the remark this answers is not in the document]
benjaminhuth: Because you may require this to construct the `Acts::GaussianSumFitter` class, so it is not an implementation detail.

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
```diff
@@ -0,0 +1,322 @@
...
+
+#include <boost/math/quadrature/trapezoidal.hpp>
+#include <boost/math/special_functions/gamma.hpp>
+
+namespace Acts {
+
+namespace detail {
+
+struct GaussianComponent {
+  ActsScalar weight, mean, var;
```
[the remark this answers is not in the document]
benjaminhuth: good points! I'll adress this!

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
```diff
@@ -0,0 +1,322 @@
...
+  constexpr bool validXOverX0(ActsScalar x) const { return x < higherLimit; }
+
+  /// Generates the mixture from the polynomials and reweights them, so
+  /// that the sum of all weights is 1
+  ///
+  /// @param x The input in terms of x/x0 (pathlength in terms of radiation length)
+  auto mixture(ActsScalar x) const {
+    // Build a polynom
+    auto poly = [](ActsScalar xx,
+                   const std::array<ActsScalar, PolyDegree + 1> &coeffs) {
```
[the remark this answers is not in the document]
benjaminhuth: good idea

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
benjaminhuth: Hmm in this case I don't think, because there is no default constructor, and e.g., for the data it is not possible to specify defaults.
But I could explicitly delete the default constructor do be more explicit here.

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
benjaminhuth: Good point, at the moment they are ignored, but I always wanted to change to `boost::static_vector`. That would be more clear

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
benjaminhuth: yeah at the moment yes. The Atlas "file specification"  (which is the worst file specification I've ever seen) requires this, but there is not file present that has this...
But should be not very difficult to implement, I think about it.

Examples/Algorithms/TrackFitting/src/GsfFitterFunction.cpp
benjaminhuth: Hmm yeah I wanted to document the behavior in this situation. However, this is not optimal, because it is somehow counter-intuitive that empty string is valid and wrong path not... Maybe std::optional<std::string>? But how would this be on the python side? I could also expose the Bethe heitler object to python... that would be the clearest solution I guess... What is your opinion on this @andiwand ?

Examples/Algorithms/TrackFitting/src/GsfFitterFunction.cpp
benjaminhuth: ah yeah :D see the comment above

Examples/Algorithms/TrackFitting/src/GsfFitterFunction.cpp
```diff
@@ -111,38 +111,25 @@ std::shared_ptr<TrackFittingAlgorithm::TrackFitterFunction>
 ActsExamples::makeGsfFitterFunction(
     std::shared_ptr<const Acts::TrackingGeometry> trackingGeometry,
     std::shared_ptr<const Acts::MagneticFieldProvider> magneticField,
-    std::string lowBetheHeitlerPath, std::string highBetheHeitlerPath,
+    Acts::Experimental::AtlasBetheHeitlerApprox<6, 5> betheHeitlerApprox,
```
andiwand, line 114: would it make sense to use the typedef here?
```suggestion
    BHApprox betheHeitlerApprox,
```

Examples/Algorithms/TrackFitting/src/GsfFitterFunction.cpp
```diff
@@ -111,38 +111,25 @@ std::shared_ptr<TrackFittingAlgorithm::TrackFitterFunction>
...
-
   // Standard fitter
   Acts::Navigator::Config cfg{trackingGeometry};
   cfg.resolvePassive = false;
   cfg.resolveMaterial = true;
   cfg.resolveSensitive = true;
   Acts::Navigator navigator(cfg);
   Propagator propagator(std::move(stepper), std::move(navigator));
-  Fitter trackFitter(std::move(propagator), BHApprox(bhapp));
+  Fitter trackFitter(std::move(propagator), BHApprox(betheHeitlerApprox));
```
andiwand, line 126: should we use `move` even tho it might be a copy? or make the copy implicit?

```suggestion
  Fitter trackFitter(std::move(propagator), std::move(betheHeitlerApprox));
```

Examples/Python/src/TrackFitting.cpp
```diff
@@ -72,16 +72,24 @@ void addTrackFitting(Context& ctx) {
         py::arg("reverseFilteringMomThreshold"),
         py::arg("freeToBoundCorrection"));
 
+    using BetheHeitlerApprox =
```
andiwand, line 75: consistency with the other typedef i.e. use `BHApprox` or change the other one?

andiwand (approved):
looks great :+1: as always feel free to ignore my comments if you disagree

Examples/Algorithms/TrackFitting/src/GsfFitterFunction.cpp
benjaminhuth: Hmm the problem with the move is that we want to instaniate two fitters. If we do `std::move`, it might be that the second fitter gets an invalid object

benjaminhuth: I somehow want to make the copy explicit, if its ok for you

andiwand (approved):
:rocket:

andiwand:
oh no doxygen is down https://www.doxygen.nl/files/doxygen-1.9.4.linux.bin.tar.gz cc @paulgessinger

andiwand (approved):

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
```diff
@@ -0,0 +1,324 @@
...
+namespace Experimental {
+
+/// This class approximates the Bethe-Heitler with only one component. This is
+/// mainly inside @ref AtlasBetheHeitlerApprox, but can also be used as the
+/// only component approximation (then probably for debugging)
+struct BetheHeitlerApproxSingleCmp {
+  /// Returns the number of components the returned mixture will have
+  constexpr auto numComponents() const { return 1; }
+
+  /// Checks if an input is valid for the parameterization. The threshold for
```
AJPfleger, line 67:
```suggestion
  /// Checks if an input is valid for the parameterisation. The threshold for
```

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
```diff
@@ -0,0 +1,324 @@
...
+
+ private:
+  Data m_low_data;
+  Data m_high_data;
+  bool m_low_transform;
+  bool m_high_transform;
+
+ public:
+  /// Construct the Bethe-Heitler approximation description. Additional to the
+  /// coefficients of the polynomials, the information wether these values need
```
AJPfleger, line 122:
```suggestion
  /// coefficients of the polynomials, the information whether these values need
```

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
```diff
@@ -0,0 +1,324 @@
...
+    return make_mixture(m_high_data, high_x, m_high_transform);
+  }
+
+  /// Loads a parameterization from a file according to the Atlas file
+  /// description
+  ///
+  /// @param low_parameters_path Path to the foo.par file that stores
+  /// the parameterization for low x/x0
+  /// @param high_parameters_path Path to the foo.par file that stores
+  /// the parameterization for high x/x0
```
AJPfleger, line 222:
```suggestion
  /// Loads a parameterisation from a file according to the Atlas file
  /// description
  ///
  /// @param low_parameters_path Path to the foo.par file that stores
  /// the parameterisation for low x/x0
  /// @param high_parameters_path Path to the foo.par file that stores
  /// the parameterisation for high x/x0
```

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
```diff
@@ -0,0 +1,324 @@
...
+    const auto [high_data, high_transform] = read_file(high_parameters_path);
+
+    return AtlasBetheHeitlerApprox(low_data, high_data, low_transform,
+                                   high_transform);
+  }
+};
+
+/// Creates a @ref AtlasBetheHeitlerApprox object based on a ATLAS
+/// configuration, that are stored as static data in the source code.
+/// This may not be an optimal configuration, but should allow running
```
AJPfleger, line 274:
```suggestion
/// Creates a @ref AtlasBetheHeitlerApprox object based on an ATLAS
/// configuration, that are stored as static data in the source code.
/// This may not be an optimal configuration, but should allow to run
```

docs/core/track_fitting.md
```diff
@@ -69,15 +69,20 @@ A GSF example can be found in the Acts Examples Framework [here](https://github.
...
 
-This approximation of the Bethe-Heitler distribution is described in {class}`Acts::detail::BetheHeitlerApprox`. The class is templated on the number of components and the degree of the polynomial, and is designed to be used with the [parameterization files from ATLAS](https://gitlab.cern.ch/atlas/athena/-/tree/master/Tracking/TrkFitter/TrkGaussianSumFilter/Data). However, in principle the GSF could be constructed with custom classes with the same interface as {class}`Acts::detail::BetheHeitlerApprox`.
+However, ACTS ships with the class {class}`Acts::Experimental::AtlasBetheHeitlerApprox` that implements the ATLAS strategy for this task: To be able to evaluate the approximation of the Bethe-Heitler distribution for different materials and thicknesses, the individual Gaussian components (weight, mean, variance of the ratio $E_f/E_i$) are parametrised as polynomials in $x/x_0$. This class can load files in the ATLAS format that can be found [here](https://gitlab.cern.ch/atlas/athena/-/tree/master/Tracking/TrkFitter/TrkGaussianSumFilter/Data). A default parameterization can be created with {func}`Acts::Experimental::makeDefaultBetheHeitlerApprox`.
```
AJPfleger, line 74:
```suggestion
However, ACTS ships with the class {class}`Acts::Experimental::AtlasBetheHeitlerApprox` that implements the ATLAS strategy for this task: To be able to evaluate the approximation of the Bethe-Heitler distribution for different materials and thicknesses, the individual Gaussian components (weight, mean, variance of the ratio $E_f/E_i$) are parametrised as polynomials in $x/x_0$. This class can load files in the ATLAS format that can be found [here](https://gitlab.cern.ch/atlas/athena/-/tree/master/Tracking/TrkFitter/TrkGaussianSumFilter/Data). A default parameterisation can be created with {func}`Acts::Experimental::makeDefaultBetheHeitlerApprox`.
```

docs/core/track_fitting.md
```diff
@@ -69,15 +69,20 @@ A GSF example can be found in the Acts Examples Framework [here](https://github.
...
 
-For small $x/x_0$ the {class}`Acts::detail::BetheHeitlerApprox` only returns a one-component mixture or no change at all. When loading a custom parametrization, it is possible to specify different parameterizations for high and for low $x/x_0$. The thresholds are currently not configurable.
+The {class}`Acts::Experimental::AtlasBetheHeitlerApprox` is constructed with two parameterizations, allowing to use different parameterizations for different $x/x_0$. In particular, it has this behaviour:
```
AJPfleger, line 76:
```suggestion
The {class}`Acts::Experimental::AtlasBetheHeitlerApprox` is constructed with two parameterisations, allowing to use different parameterisations for different $x/x_0$. In particular, it has this behaviour:
```

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
benjaminhuth: I would actually stick to the AE spelling here...

andiwand (approved):

Issue #1522, opened 2022-09-14
Separate context and config information in SeedFinderConfig


tboldagh:
In the SeedFinder integration in Athena 
https://acode-browser1.usatlas.bnl.gov/lxr/source/athena/Tracking/Acts/ActsTrkTools/ActsTrkSeedingTool/src/SeedingTool.cxx
we need to copy construct the config object over and over because the config mixes actual configuration and more of context information.

Separating the two would allow us to reuse SeedFinderConfig object and since it will be made only once (faster) and maybe even think of some validation?

It would be breaking change though.

Tagging: @LuisFelipeCoelho @asalzburger 

BTW. Why the capitalisation Seedfinder is used instead of SeedFinder (like one would expect)?

tboldagh:
Hi Experts, I have some spare cycles and can do the required changes. How does the idea sound?

LuisFelipeCoelho:
The idea sounds good to me.

FYI: @robertlangenberg

tboldagh:
So I had a look. Whereas the beam spot position update is kind of expected update, how realistic is to change magnetic field when processing events? Outside of LHC experiments say.

tboldagh:
See the set of proposed changes:
https://github.com/acts-project/acts/compare/main...tboldagh:acts:refactor-seed-finder-config-and-context?expand=1

LuisFelipeCoelho:
I agree with the changes

LuisFelipeCoelho:
@robertlangenberg, do you know why Seedfinder is not capitalised?

robertlangenberg:
That was a simple oversight, I have no objection against renaming :)

asalzburger:
Yes, let us rename - I always disliked the `Seedfinder` spelling.

tboldagh:
The PR finally addressed this issue entirely. There are small optimisation that I still planned to do but otherwise this ticket can be closed.

PR #1630, opened 2022-10-26
refactor!: Move beam spot position and mag field value from SeedFinderConfig to SeedFinderOptions

tboldagh:
BREAKING CHANGE:
This PR changes API of the SeedFinder. Now, in addition to the Config object obtains also Options object. 
This way the Config object does not need to change in the client code (e.g. ATLAS), even if the beam spot moves. 

This PR addresses half of the issue https://github.com/acts-project/acts/issues/1522 but not yet closes it.
A similar move would need to happen for magnetic field.


Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -40,6 +41,7 @@ template <typename external_spacepoint_t, typename platform_t>
 template <template <typename...> typename container_t, typename sp_range_t>
 void SeedFinder<external_spacepoint_t, platform_t>::createSeedsForGroup(
     SeedingState& state,
+    const SeedFinderOptions&,
```
tboldagh, line 44: A question here. The SeedfinderConfig does not seem to be using beam spot position. 
Is there an intention to use it in the future?

LuisFelipeCoelho: As far as I know, there is no intention of using the beam spot position here, only in BinnedSPGroup

tboldagh: ok, thanks. So this will only be used for the B field.

LuisFelipeCoelho: If this will only be used for the B filed, do we need to pass it in `SeedFinder::createSeedsForGroup` or `SeedFinder::SeedFinder` is enough?

tboldagh: I have refactored it further, to also include bField. But instead of passing it each time to these functions the options are used at construction. 
Ideally another refactor would be useful where the SeedFinder is instantiated once in clients code and then createSeedsForGroups called for every events. For another PR I guess.

LuisFelipeCoelho (approved):
Looks good to me

paulgessinger:
CI has failures. The docs failure is known and should go away after an update.

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -500,7 +502,8 @@ def addSeeding(
                     useDeltaRorTopRadius=seedFilterConfigArg.useDeltaRorTopRadius,
                 )
             )
-
+            # print(help(acts.examples.SeedingOrthogonalAlgorithm))
```
LuisFelipeCoelho, line 505: do we want this comments here?

LuisFelipeCoelho:
I just left one quick question but after that I think it is good to go

CI/physmon/physmon.py
```diff
@@ -298,9 +299,11 @@
                     sigmaScattering=5,
                     radLengthPerSeed=0.1,
                     minPt=500 * u.MeV,
-                    bFieldInZ=1.99724 * u.T,
                     impactMax=3 * u.mm,
                 ),
+                SeedFinderOptionsArg(
+                    bFieldInZ=1.99724 * u.T,
+                ),
```
andiwand, line 306:
```suggestion
                SeedFinderOptionsArg(bFieldInZ=1.99724 * u.T),
```

I think the formatter is just unfolding this because of the `,`

andiwand (approved):
looks good I also just have one small cosmetic comment

CI/physmon/physmon.py
tboldagh: Thanks, this is addressed

Examples/Python/python/acts/examples/reconstruction.py
tboldagh: Well spot, Thanks. Fixed

andiwand (approved):
lets get this in

PR #1631, opened 2022-10-27
refactor: Enable Fatras interactions by default / make configurable in addFatras(...)

benjaminhuth:
It turned out that all interactions in Fatras where turned off by default... This is not optimal I think. This PR enables these by default both on the C++ site and on the python side, to be explicit.

As a consequence, all root hashes relying on FATRAs should change...


andiwand (approved):

benjaminhuth:
If I see it correctely in the physics performance monitor, this has significant impact with respect to some plots, e.g. qop for truth-tracking, this looks not really good...
I'm not sure how to proceed from here actually. Is this expected? Is this a problem of Fatras or of the Algorithms? What do you think, @asalzburger @paulgessinger ?

andiwand:
> Running with Geant could also be good to check. But acutally I do not have the capacity right now to investigate this in-depth...

I agree running Geant4 would be interesting. we can let this rest for now until someone runs into this as well or you want to follow up when you have more time

paulgessinger:
Thoughts @asalzburger?

benjaminhuth:
Hmm how about we go like this: We introduce a flag in the `addFatras(...)` method to let this be enabled easily (e.g. for running the GSF or other testing), but is off by default to let the physmon pass.
For these physics issues we open an issue, and discuss it there.
What do you think @andiwand @paulgessinger @asalzburger ?

andiwand:
I agree :+1: ping me if you need a review

benjaminhuth:
@andiwand I think this is ready to review (mainly the title has changed to be more appropriate^^)

andiwand (approved):

andiwand (approved):

beomki-yeo:
@paulgessinger @benjaminhuth In physmon truth tracking kalman, I got this:

1. Eta fixed to 0, all physics disabled in simulation
image

2. Eta fixed to 0, all physics enabled in simulation
image

Both don't look right. Let me think about it more...

beomki-yeo:
> nce does not look so bad but truth tracking q/p looks really bad

@andiwand Actually the monitored graph looks better than reference one (the sigma is closer to 1)

PR #1633, opened 2022-10-27
feat: Python documentation

benjaminhuth:
This could be a first version of a python documentation. It uses sphinx + autodoc + autosummary. Warning: It is not very pretty, and in some cases also not very functional.
But at least it is possible to view the signatures of e.g. the `addFoo` functions, or see the members of the config structs of the algorithms...
Probably there is much room for improvement though...

Note: I just modified the docstrings of the python functions so that they do not cause errors during gerneration.


benjaminhuth:
Ah this cannot work since we do not build the python plugin on readthedocs, right @paulgessinger? Do you think this would be possible, or to use an build artifact or something like that?

paulgessinger:
Hm good point. That might be a bit tricky since we'd have to restructure the way RTD builds this to pull the required info from a previous job. We'd have to trigger the RTD builds ourselves  rather than using their github integration.

PR #1634, opened 2022-10-28
feat: Make Clusterization.hpp aware of grid dimensionality 

gagnonlg:
This PR makes Clusterization.hpp aware of the grid dimensionality. Since the problem is much simpler on a 1-D grid, this enables a few optimizations. Right now, only support for 2D and 1D grid is implemented.

To aid in testing, unit tests for both cases have also been added.

Note, the changes are supposed to be backwards compatible. Previously it was assumed that the grid was 2-dimensional, thus the GridDim template parameter defaults to 2 where relevant.


gagnonlg:
Running an integration test on the athena side before un-WIP-ing

noemina (approved):
Nothing to mention, all looks good to me.

PR #1636, opened 2022-10-31
fix: Adjust full_chain_itk configuration for updated vertex fitting algori…

goetzgaycken:
The vertex fitting algorithm does not perform a track selection anymore (since #1579) but e.g. expects preselected track parameters as input.

### DIFF
--- a/Examples/Scripts/Python/full_chain_itk.py
+++ b/Examples/Scripts/Python/full_chain_itk.py
@@ -10,6 +10,7 @@ from acts.examples.simulation import (
     ParticleSelectorConfig,
     addDigitization,
 )
+from acts.examples import TrackSelector
 from acts.examples.reconstruction import (
     addSeeding,
     TruthSeedRanges,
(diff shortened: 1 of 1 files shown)


andiwand:
hey @goetzgaycken ! sorry I forgot to follow up with the itk full chain. you can attach the track selection to the track finding now. an example is given in the odd full chain [here](https://github.com/acts-project/acts/blob/main/Examples/Scripts/Python/full_chain_odd.py#L108)

goetzgaycken:
> you can attach the track selection to the track finding now. 
- Sure. But in run2/3 the track selection for vertex finding was tighter than for track finding. Is that different for run4 ? 
- Would that do something different than the track selector algorithm concerning the vertex finding ?

paulgessinger:
Hm, at least we could experiment with a tighter track selection before the vertexing. Thoughts @timadye ?

andiwand:
I moved the track selection into CKF because I thought it aligns better with the seeding interface for example. I did not consider any procedure from run2/3/4. I can add the track selection back into the vertexing if that is a requirement. But I guess the track finding should keep the track selection?

timadye:
I think it would be good to add the `TrackSelectorRanges` option back into `addVertexFitting`. Sorry I hadn't spotted that it was removed by #1579. We should be able to specify the selection in `addCKFTracks` or `addVertexFitting`, or both (as @goetzgaycken suggests), even if our examples plump for one or the other.

Can we include `addAmbiguityResolution` in the `full_chain_itk.py` example?

andiwand:
makes sense. sorry for the back and forth. having two track selectors makes the wiring a bit complicated so I would wait for https://github.com/acts-project/acts/pull/1588 to clean this up

goetzgaycken:
closing this, because problem fixed by #1642 (which adds ambiguity resolution which provides track selection)

PR #1637, opened 2022-10-31
refactor: DD4hep detector access

andiwand:
noticed that we pass around the geometry service in python which seems confusing and can be solved more directly by exposing the `dd4hep::Detector` object


paulgessinger (approved):
lgtm

Examples/Python/python/acts/examples/simulation.py
```diff
@@ -555,10 +553,12 @@ def addGeant4(
     else:
         particles_selected = "particles_input"
 
-    g4detector = DDG4DetectorConstruction(geometryService)
+    if g4detectorConstruction is None:
```
benjaminhuth, line 556: Hmm is this redundant? There is no default for `g4detectorConstruction`, so someone would need to pass explicitly `None` for the `g4detectorConstruction` to the function, which is not very likely.

benjaminhuth (approved):
Very nice, improves the interface quite a bit. Just a minor comment, feel free to ignore it!

Examples/Python/python/acts/examples/simulation.py
andiwand: you are right - I will make use of this `if` here https://github.com/acts-project/acts/blob/83298cbe8bc08761113c4b132f43acd89c9b5edf/Examples/Python/python/acts/examples/simulation.py#L589-L593 in https://github.com/acts-project/acts/pull/1638

so I think its fine to merge this temporarily even tho is has no effect

PR #1638, opened 2022-10-31
feat: Geant4 telescope detector examples

andiwand:
- Geant4 telescope detector construction
- cleanup telescope code
- auto Geant4 detector construction discovery


Examples/Algorithms/Geant4/src/TelescopeDetectorConstruction.cpp
```diff
@@ -0,0 +1,122 @@
...
+  //
+  G4double worldSizeX = 2 * m_cfg.bounds[0];
+  G4double worldSizeY = 2 * m_cfg.bounds[1];
+  G4double worldSizeZ = length;
+
+  // Construct the rotation
+  // This assumes the binValue is binX, binY or binZ. No reset is necessary in
+  // case of binZ
+  G4RotationMatrix* rotation = nullptr;
+  if (m_cfg.binValue == Acts::BinningValue::binX) {
```
benjaminhuth, line 59: I'm actually not entirely happy with comparing an `int` with an `Acts::BinningValue` (which is an enum)...
But maybe it would make more sense to change the `Acts::BinningValue` to a scoped enum?

But for the start it could also be enough to check if the value is 0, 1 or 2 and trow otherwise?

andiwand: same :smile: I think the reason is that this config value comes from python? I will change it to an enum class and use static cast

benjaminhuth: Yeah probably. The cleaner way would be to bind the `enum` properly to python, so that also in python we could pass something like `acts.BinningValue.binX` to the config...
But not sure if that must happen in that PR

Examples/Python/python/acts/examples/simulation.py
```diff
@@ -555,13 +559,19 @@ def addGeant4(
...
+
     g4conf = geant4SimulationConfig(
         level=customLogLevel(),
-        detector=g4detector,
+        detector=g4detectorConstruction,
         inputParticles="particles_input",
         trackingGeometry=trackingGeometry,
         magneticField=field,
+        volumeMappings=[],
+        materialMappings=["G4_Si"],
```
benjaminhuth, line 574: In principle I'm fine with this, but this means that this material is hardcoded now... Do you have checked if the ODD still works with this?
The best solution would be to couple the material mapping information the detectorconstruction object, but it would be maybe also sufficient to add more entries to the default, e.g. `["G4_Si", "Silicon", "Si"]` I guess (`"Silicon"` is the default entry on the C++ side, which gets overwritten by this).

Examples/Python/src/Detector.cpp
```diff
@@ -34,6 +34,11 @@ void addDetector(Context& ctx) {
         .def("name", &IContextDecorator::name);
   }
 
+  {
+    py::class_<IBaseDetector, std::shared_ptr<IBaseDetector>>(mex,
```
benjaminhuth, line 38: Do we actually need a binding for IBaseDetector? Since python is a dynamically typed language, I would have believed that we are fine without with, but maybe I'm wrong.

Examples/Scripts/Python/telescope_simulation.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+
+    field = acts.ConstantBField(acts.Vector3(0, 0, 2 * u.T))
+
+    outputDir = Path.cwd() / "telescope_simulation"
+    if not outputDir.exists():
+        outputDir.mkdir()
+
+    for geant, postfix in [(False, "fatras"), (True, "geant4")]:
+        rnd = acts.examples.RandomNumbers()
```
benjaminhuth, line 30: Maybe create the rnd object outside of the for-loop? And initialize with static seed like 42 to have reproducibility (or just make clear we have a static seed here, I'm not sure how default initialization is actually...)?

benjaminhuth (commented):
Looks good, just a few minor comments, but nothing critical

Examples/Python/python/acts/examples/simulation.py
andiwand: oh yeah that should not be merged. I think it would be good to expose the list as you did. but I will use "normal" silicon in the G4 construction

Examples/Algorithms/Geant4/src/DetectorConstructionHelper.cpp
```diff
@@ -0,0 +1,37 @@
...
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#include "ActsExamples/Geant4/DetectorConstructionHelper.hpp"
+
+#include "ActsExamples/TelescopeDetector/TelescopeDetector.hpp"
+#include "ActsExamples/TelescopeDetector/TelescopeG4DetectorConstruction.hpp"
+#ifdef ACTS_GEANT4_DD4HEP
```
andiwand, line 13: I am not too happy about those. any other ideas @benjaminhuth @paulgessinger ?

paulgessinger: Might be cleaner to determine this at CMake level only and add different `.cpp` files to the target? That's ~ what I did for the python bindings.

benjaminhuth: One alternative that comes to my mind would be to implement this in a pure python function, that can dynamically check if a type is available or not... Might also be an option, because I don't think we need to call this ever directly from C++, right?

Besides this I'm not sure if handling in CMake is useful here, because thant we could not do it in a single function, right?

Examples/Algorithms/Geant4/src/DetectorConstructionHelper.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "ActsExamples/DDG4/DDG4DetectorConstruction.hpp"
+#endif
+
+#include <stdexcept>
+
+G4VUserDetectorConstruction *ActsExamples::getG4DetectorContruction(
+    const ActsExamples::IBaseDetector &detector) {
+  if (auto telescope =
+          dynamic_cast<const ActsExamples::Telescope::TelescopeDetector *>(
+              &detector)) {
```
AJPfleger, line 24: I am not sure, if I understood this correctly, but isn't an assignment as a conditional always `true`? same for the `else if` below.

benjaminhuth: Hmm this is indeed correct code, if you do a declaration in an `if` statement, then the compiler tries to convert the declared variable contextually to `bool` (though this is not documented well e.g., at cppreference, but there you can find the bullet point "simple declaration" under as a valid condition).

andiwand: `dynamic_cast` will give you a `nullptr` if the cast fails for pointers. for references this would throw

paulgessinger: I would argue 
```suggestion
  if (auto telescope =
          dynamic_cast<const ActsExamples::Telescope::TelescopeDetector *>(
              &detector); telescope != nullptr) {
```
is better.

benjaminhuth: yeah this would be more clear I agree, though it is semantically equivalent.

andiwand:
this one contains now #1646 which should be merged first

PR #1640, opened 2022-10-31
feat: Basic GSF physics performance monitoring

benjaminhuth:
This sets up some basic performance monitoring for the GSF, it may need to be tuned in the future


andiwand (approved):
looks good to me :+1:

paulgessinger:
:fire:

PR #1642, opened 2022-11-01
fix: ITk full chain

andiwand:
- align ITk full chain with ODD full chain
  - add ambiguity resolution
  - catch up with interface changes of `addCKFTracks` and `addVertexFitting`

### DIFF
--- a/Examples/Scripts/Python/full_chain_itk.py
+++ b/Examples/Scripts/Python/full_chain_itk.py
@@ -15,9 +15,11 @@ from acts.examples.reconstruction import (
     TruthSeedRanges,
     addCKFTracks,
     CKFPerformanceConfig,
+    TrackSelectorRanges,
+    addAmbiguityResolution,
+    AmbiguityResolutionConfig,
     addVertexFitting,
     VertexFinder,
-    TrackSelectorRanges,
 )
 
 ttbar_pu200 = False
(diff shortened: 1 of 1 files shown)


Examples/Scripts/Python/full_chain_itk.py
```diff
@@ -90,16 +92,24 @@
...
     CKFPerformanceConfig(ptMin=1.0 * u.GeV if ttbar_pu200 else 0.0, nMeasurementsMin=6),
+    TrackSelectorRanges(pt=(1.0 * u.GeV, None), absEta=(None, 4.0), removeNeutral=True),
+    outputDirRoot=outputDir,
+)
+
+addAmbiguityResolution(
+    s,
+    AmbiguityResolutionConfig(
+        maximumSharedHits=3,
+    ),
```
timadye, line 103:
```suggestion
    AmbiguityResolutionConfig(maximumSharedHits=3),
```
This seems to be overzealous `black` formatting. I've noticed how it can put things on separate lines, but then doesn't go back when the option is simplified - possibly because of the extra `,` it adds.

timadye (approved):
Looks good to me.
I tested this with `ttbar_pu200=False` and `True` and it seems to work nicely.
I see there's now a `performance_ambi.root` (just running `CKFPerformanceWriter` again). Nice!

Examples/Scripts/Python/full_chain_itk.py
andiwand: oh good point will do the same for odd

andiwand:
> I see there's now a `performance_ambi.root` (just running `CKFPerformanceWriter` again). Nice!

yes @paulgessinger suggested to have an output to monitor this. I thought it might be easiest to reuse the `CKFPerformanceWriter` as it operates on the same data :smile:

timadye (approved):
lgtm

PR #1644, opened 2022-11-01
feat: examples restore vertexing track selector

andiwand:
https://github.com/acts-project/acts/pull/1579/ remove the track selector from `addVertexing` which is still required. see https://github.com/acts-project/acts/pull/1636

https://github.com/acts-project/acts/pull/1588 should clean this up a little

### DIFF
--- a/Examples/Python/python/acts/examples/reconstruction.py
+++ b/Examples/Python/python/acts/examples/reconstruction.py
@@ -679,7 +679,7 @@ def addCKFTracks(
     outputDirCsv: Optional[Union[Path, str]] = None,
     outputDirRoot: Optional[Union[Path, str]] = None,
     selectedParticles: str = "truth_seeds_selected",
-    trackSelectorRanges: Optional[TrackSelectorRanges] = TrackSelectorRanges(),
+    trackSelectorRanges: Optional[TrackSelectorRanges] = None,
     writeTrajectories: bool = True,
     logLevel: Optional[acts.logging.Level] = None,
 ) -> None:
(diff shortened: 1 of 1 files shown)


Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -679,7 +679,7 @@ def addCKFTracks(
     outputDirCsv: Optional[Union[Path, str]] = None,
     outputDirRoot: Optional[Union[Path, str]] = None,
     selectedParticles: str = "truth_seeds_selected",
-    trackSelectorRanges: Optional[TrackSelectorRanges] = TrackSelectorRanges(),
+    trackSelectorRanges: Optional[TrackSelectorRanges] = None,
```
timadye, line 682: This is fine and maybe a bit more efficient, but is the change necessary? `TrackSelectorRanges()` does nothing, I think.

The problem with removing the selector altogether is that you have different inputs/outputs. Have you checked the full chain works with and without `trackSelectorRanges`? I'm wary because this has bitten me several times in the past.

I guess this will be easier to handle once we have #1588. Is it worth delaying the change until then?

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -1023,6 +1021,7 @@ def addVertexFitting(
     trackParameters: str = "filteredTrackParameters",
     trackParametersTips: Optional[str] = "filteredTrackParametersTips",
     vertexFinder: VertexFinder = VertexFinder.Truth,
+    trackSelectorRanges: Optional[TrackSelectorRanges] = None,
```
timadye, line 1024: Similarly here: might want to put the default as `TrackSelectorRanges()`.

Examples/Python/python/acts/examples/reconstruction.py
andiwand: I just tried this locally and it works in all combinations. I would prefer having `None` if we want keep this parameter `Optional`.

depending on how crucial this issue is we can put this in now or war for https://github.com/acts-project/acts/pull/1588. it might take a few more days to get it ready

timadye: No problem. If it's tested, I'm happy with this change.

timadye (approved):

PR #1645, opened 2022-11-03
feat: adding portal, detector volume & detector

asalzburger:
This PR adds the new `Portal`, `DetectorVolume` and `Detector` geometry which is meant to eventually replace the `Layer`, `TrackingVolume`, `TrackingGeometry` infrastructure.

The new classes are currently put under the `Experimental` namespace.

Only the new classes, the necessary definitions and unit tests are added with this PR, which will be followed up by the visualisation, the detector building and the navigation code.

A description of the concepts and classes is added to the 
`docs/experimental_geometry.md` to this PR.


Core/include/Acts/Geometry/NavigationState.hpp
```diff
@@ -0,0 +1,141 @@
...
+    /// can differ for sensitive surfaces and portals
+    BoundaryCheck bCheck = true;
+  };
+
+  /// Surface candidate vector alias, this allows to use e.g. boost_small vector
+  /// or other stl like containers
+  using SurfaceCandidates = std::vector<SurfaceCandidate>;
+
+  /// The current position
+  Vector3 position = Vector3(0., 0., 0.);
```
benjaminhuth, line 56: Hmm just to mention this again: I think for this we already have a stepper interface that exposes this in a defined way without needing to know the details of the stepper-state implementation. This interface is also enforced by a one of these C++17 `static_assert`-concepts, so this should be reliable.

Wouldn't it be an option to also pass stepper and stepper state to the delegate? Since this in principle redundant copy of these information must be done every step?

And e.g. for the `MultiEigenStepper` this is not so cheap, because there we potentially perform a weighted average of multiple components to compute these values, and this would be computed regardless if it is needed...

Core/include/Acts/Geometry/Portal.hpp
```diff
@@ -0,0 +1,166 @@
...
+  /// Retrieve a @c std::shared_ptr for this surface (non-const version)
+  ///
+  /// @note Will error if this was not created through the @c makeShared factory
+  ///       since it needs access to the original reference. In C++14 this is
+  ///       undefined behavior (but most likely implemented as a @c bad_weak_ptr
+  ///       exception), in C++17 it is defined as that exception.
+  /// @note Only call this if you need shared ownership of this object.
+  ///
+  /// @return The shared pointer
+  std::shared_ptr<Portal> getSharedPtr();
```
benjaminhuth, line 74: In principle I think two `surface` accessors returning references should be enough, since we have have the `getSharedPtr` method for the surfaces as well, right?
I would prefer such an interface because it immediately tells the caller that the surface cannot be `nullptr` via interface.

Core/include/Acts/Definitions/Common.hpp
```diff
@@ -36,6 +36,11 @@ static constexpr ActsScalar s_curvilinearProjTolerance = 0.999995;
 /// respect to a given momentum or direction
 enum class NavigationDirection : int { Backward = -1, Forward = 1 };
 
+/// Convert navigation dir to index
```
benjaminhuth, line 39: Maybe add some documentation what this index refers to? since at least for me this is not immediately clear...

Core/include/Acts/Geometry/Portal.hpp
asalzburger: Agreed, changed.

Core/include/Acts/Definitions/Common.hpp
asalzburger: Fixed.

Core/include/Acts/Geometry/detail/DetectorVolumeFinders.hpp
```diff
@@ -0,0 +1,99 @@
...
+                            NavigationState& nState) noexcept(false) {
+    if (nState.currentDetector == nullptr) {
+      throw std::runtime_error(
+          "DetectorVolumeFinders: no detector set to navigation state.");
+    }
+
+    auto volumes = nState.currentDetector->volumes();
+    for (const auto v : volumes) {
+      if (v->inside(gctx, nState.position)) {
+        nState.currentVolume = v;
```
andiwand, line 45:
```suggestion
        nState.currentVolume = v;
        return;
```

otherwise we end up with `nullptr` at the end

Core/include/Acts/Geometry/detail/SurfaceCandidatesUpdators.hpp
```diff
@@ -0,0 +1,195 @@
...
+                   std::abs(pathToB) < s_onSurfaceTolerance) {
+          return true;
+        }
+        return pathToA < pathToB;
+      });
+  // Set the surface candidate
+  nState.surfaceCandidate = nCandidates.begin();
+}
+
+/// A ordered portal provider
```
andiwand, line 76:
```suggestion
/// An ordered portal list provider
```

Core/include/Acts/Geometry/detail/SurfaceCandidatesUpdators.hpp
```diff
@@ -0,0 +1,195 @@
...
+/// @return an ordered list of portal candidates
+inline static void portalAndSurfaceCandidates(const GeometryContext& gctx,
+                                              NavigationState& nState) {
+  if (nState.currentDetector == nullptr) {
+    throw std::runtime_error(
+        "PortalAndSurfaceCandidates: no detector volume set to navigation "
+        "state.");
+  }
+  // A volu
+  // A volume switch has happened, update list of portals & surfaces
```
andiwand, line 123:
```suggestion
  // A volume switch has happened, update list of portals & surfaces
```

Core/include/Acts/Geometry/detail/SurfaceCandidatesUpdators.hpp
```diff
@@ -0,0 +1,195 @@
...
+    }
+    // Filling the new portal candidates
+    const auto& portals = nState.currentVolume->portals();
+    PortalsFiller::fill(nState, portals);
+  }
+  // Sort and update
+  updateCandidates(gctx, nState);
+}
+
+/// A ordered portal and surface provider
```
andiwand, line 106:
```suggestion
/// An ordered portal and surface list provider
```

Core/src/Geometry/DetectorVolume.cpp
```diff
@@ -0,0 +1,228 @@
...
+    const std::vector<std::shared_ptr<DetectorVolume>>& volumes) {
+  m_SurfaceCandidatesUpdator = std::move(navStateUpdator);
+  m_surfaces = ObjectStore<std::shared_ptr<Surface>>(surfaces);
+  m_volumes = ObjectStore<std::shared_ptr<DetectorVolume>>(volumes);
+
+  // This can take a while for complicated geometries
+  // it also needs a geometry context, needs to be figured out
+  // if (not checkContainment(GeometryContext)) {
+  //  throw std::runtime_error("Objects are not contained by volume.");
+  //}
```
andiwand, line 143: dead?

Core/src/Geometry/DetectorVolume.cpp
```diff
@@ -0,0 +1,228 @@
...
+  // Assign the boundary Identifier
+  GeometryIdentifier portalId = geometryId;
+  for (auto [i, p] : enumerate(m_portals.internal)) {
+    portalId.setBoundary(i + 1);
+    p->assignGeometryId(portalId);
+  }
+
+  // Assign the sensitive/surface Identifier
+  GeometryIdentifier sensitiveId = geometryId;
+  /// @todo add passive count
```
andiwand, line 191: important todo?

Core/src/Geometry/Portal.cpp
```diff
@@ -0,0 +1,101 @@
...
+          : NavigationDirection::Backward;
+
+  if (m_volumeUpdators[indexFromDirection(tDir)].implementation == nullptr) {
+    throw std::invalid_argument(
+        "Portal: trying to fuse portal (keep) with no links.");
+  }
+  // And now check other direction
+  NavigationDirection oDir = (tDir == NavigationDirection::Forward)
+                                 ? NavigationDirection::Backward
+                                 : NavigationDirection::Forward;
```
andiwand, line 46: this might be worth a utility function

Core/src/Geometry/Portal.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+void Acts::Experimental::Portal::updateDetectorVolume(
+    const GeometryContext& gctx, NavigationState& nState) const {
+  const auto& position = nState.position;
+  const auto& direction = nState.direction;
+  const Vector3 normal = surface().normal(gctx, position);
+  NavigationDirection nDir = (normal.dot(direction) < 0.)
+                                 ? NavigationDirection::Backward
+                                 : NavigationDirection::Forward;
```
andiwand, line 94: use case for https://github.com/acts-project/acts/blob/0c76d370479a4511bd399800100b18984bae00cc/Core/include/Acts/Definitions/Common.hpp#L39 ?

andiwand (commented):
took a first look but left out the test code for now

Core/include/Acts/Geometry/Detector.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+ public:
+  /// Factory for producing memory managed instances of Detector.
+  /// Will forward all parameters and will attempt to find a suitable
+  /// constructor.
+  ///
+  /// @tparam Args the arguments that will be forwarded
+  template <typename... Args>
+  static std::shared_ptr<Detector> makeShared(Args&&... args) {
+    return std::shared_ptr<Detector>(new Detector(std::forward<Args>(args)...));
```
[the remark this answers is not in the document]
asalzburger: Na, for Detector and DetectorVolume we do not need to worry for the overhead at all, it's the Surfaces in the TrackParameters which we need to time/understand

Core/include/Acts/Geometry/DetectorVolume.hpp
```diff
@@ -0,0 +1,582 @@
...
+class Portal;
+class Detector;
+
+/// The Portal genertor definition
+///
+/// @param gctx the geometry context
+/// @param bounds the volume bounds
+/// @param volume the detector volume for which this generator is called
+using PortalGenerator = Delegate<std::vector<std::shared_ptr<Portal>>(
+    const Transform3& gctx, const VolumeBounds& bounds,
```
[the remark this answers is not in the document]
asalzburger: The transform covers everything, because that's what the VolumeBounds need to create the portal surfaces, but the naming is wrong.
Fixing this now.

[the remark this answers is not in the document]
asalzburger: Wait, it could actually be the geometry context as well, which would make it a cleaner interface as this is used everywhere else ...

Core/include/Acts/Geometry/DetectorVolume.hpp
```diff
@@ -0,0 +1,582 @@
...
+class DetectorVolumeFactory {
+ public:
+  /// Create a detector volume - from factory
+  /// @param portalGenerator the volume portal generator
+  /// @param gctx the geometry context for construction and potential contextual store
+  /// @param args the arguments forwarded to the detector volume contructor
+  ///
+  /// @return a shared object DetectorVolume
+  template <typename... Args>
+  static std::shared_ptr<DetectorVolume> construct(
```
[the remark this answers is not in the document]
asalzburger: Not for this. For the extractors, they are used in a context where

extractor.extract(...) is called, so either for a static function or some other extractor with internal payload.

Core/include/Acts/Geometry/NavigationDelegates.hpp
```diff
@@ -0,0 +1,85 @@
...
+class Portal;
+class DetectorVolume;
+class Detector;
+
+/// Base class for navigaiton delegate implementations
+/// that allows to then do some sort of type erasure,
+/// e.g. for plotting or I/O.
+///
+/// The interface is never touched in the navigation stream
+class IDelegateImpl {
```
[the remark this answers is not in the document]
asalzburger: I had some problems with that, and I eventually wanted to require a `name()` method in order to give at least some screen info ... ?

Core/src/Geometry/DetectorVolume.cpp
asalzburger: Not dead, but a fundamental problem:

I wanted to do that in an `assert()`, but for the containment I need the `GeometryContext`, so if I give a context object name in the constructor I get an `unused variable` WARNING, if I add [[maybe_unused]] in the header, the CUDA build fails ...

Core/src/Geometry/DetectorVolume.cpp
asalzburger: I actually think I will delegate the `GeometryIdentifier` setting entirely, so we can customise it.

Core/include/Acts/Geometry/detail/SurfaceCandidatesUpdators.hpp
asalzburger: Updated.

Core/src/Geometry/Portal.cpp
asalzburger: Right you are.

Core/include/Acts/Geometry/Portal.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+#include <map>
+#include <memory>
+#include <optional>
+
+namespace Acts {
+
+class ISurfaceMaterial;
+
+namespace Experimental {
```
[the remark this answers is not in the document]
asalzburger: We discussed this and decided to leave it in the main directory. The advantage is that we don't have to physically move files and change include paths once we promote it out of `Experimental`.

Core/include/Acts/Geometry/detail/DetectorVolumeFinders.hpp
asalzburger: Ha! I didn't read this comment - and, yes, this fixes the `Detector` unit test.

Core/include/Acts/Geometry/detail/SurfaceCandidatesUpdators.hpp
asalzburger: Fixed.

Core/src/Geometry/DetectorVolume.cpp
asalzburger: This will come in a follow-up PR though.

Core/include/Acts/Geometry/NavigationDelegates.hpp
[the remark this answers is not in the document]
benjaminhuth: Hmm not sure what the reasoning there is, but I would be very carefully with using void pointers anywhere, because then we completely loose type safety...

Core/include/Acts/Geometry/NavigationState.hpp
benjaminhuth: I agree, let's visit this again then.

Core/src/Geometry/DetectorVolume.cpp
benjaminhuth: Have you tried a workaround like this? Maybe the CUDA compiler has a problem with the attribute in the function argument list?

```c++
DetectorVolume::DetectorVolume(const GeometryContext &gctx, ...) {
    // ...
    [[maybe_unused]] const auto &gctx_ref = gctx;
    assert(checkContainment(gctx) && "Objects are not contained by volume.");
    // ...
}
```

Core/include/Acts/Geometry/NavigationDelegates.hpp
```diff
@@ -0,0 +1,81 @@
...
+ public:
+  virtual ~IDelegateImpl() {}
+};
+
+/// Memory managed delegate to guarantee the lifetime
+/// of eventual unterlying delegate memory and the
+/// delegate function
+///
+template <typename deletage_type>
+struct ManagedDelegate {
```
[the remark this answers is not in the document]
asalzburger: Just to clarify, you would were thinking of:

```c++
template <typename function_type> 
struct ManagedDelegate {
  Delegate<function_type> delegate;
}
```

right?

Core/include/Acts/Geometry/DetectorVolume.hpp
```diff
@@ -0,0 +1,445 @@
...
+  friend class DetectorVolumeFactory;
+
+  /// Nested object store that holds the internal (non-const),
+  /// reference counted objects and provides an external
+  /// (const raw pointer) access
+  ///
+  /// @tparam internal_t is the internal storage representation,
+  /// has to comply with std::shared_ptr semantics
+  ///
+  template <typename internal_t>
```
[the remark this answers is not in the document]
asalzburger: I wasn't sure if we needed another type of ref counting object here, I will leave it for the moment, but we can clean that up in a later PR.

Core/include/Acts/Geometry/NavigationDelegates.hpp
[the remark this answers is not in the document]
andiwand: yeah I agree that restricts it to the `Delegate` class. when ever we see a need to generalize this we can make it more broad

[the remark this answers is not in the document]
asalzburger: Ok, will change it.

Core/include/Acts/Geometry/NavigationDelegates.hpp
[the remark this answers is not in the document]
asalzburger: Ok, let's keep as it.

Core/src/Geometry/DetectorVolume.cpp
asalzburger: I will try this, that'd be a good workaround.

Core/include/Acts/Geometry/detail/DetectorVolumeFinders.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+namespace Acts {
+namespace Experimental {
+
+namespace detail {
+
+/// @brief The end of world sets the volume pointer of the
+/// navigation state to nullptr, usually indicates the end of
+/// the known world, hence the name
+struct TryAndErrorImpl : public IDelegateImpl {
```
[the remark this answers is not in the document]
asalzburger: Oh, this was the name before I re-implemented it  ... I will switch back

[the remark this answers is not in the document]
asalzburger: I also saw that the unit test was missing for this, which I added now.

Core/include/Acts/Geometry/DetectorVolume.hpp
```diff
@@ -0,0 +1,577 @@
...
+    if (nState.currentVolume == nullptr) {
+      throw std::runtime_error(
+          "AllSurfacesExtractor: no detector volume given.");
+    }
+    return nState.currentVolume->surfaces();
+  }
+};
+
+/// Helper extractors: indexed surfaces
+struct IndexedSurfacesExtractor {
```
paulgessinger, line 502: Could this and the other `struct`s just be free functions?

paulgessinger (commented):
This is how far I got so far. Commenting this first batch now, but might have more later.

Core/include/Acts/Geometry/DetectorVolume.hpp
```diff
@@ -0,0 +1,577 @@
...
+  ///
+  /// @note throws exception if misconfigured: no bounds
+  /// @note throws exception if ghe portal general or navigation
+  ///       state updator delegates are not connected
+  DetectorVolume(
+      const GeometryContext& gctx, const std::string& name,
+      const Transform3& transform, std::unique_ptr<VolumeBounds> bounds,
+      const std::vector<std::shared_ptr<Surface>>& surfaces,
+      const std::vector<std::shared_ptr<DetectorVolume>>& volumes,
+      ManagedSurfaceCandidatesUpdator&& navStateUpdator) noexcept(false);
```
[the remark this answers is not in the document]
asalzburger: Ok, I will call it `surfaceCandidateUpdator` then.

Core/include/Acts/Geometry/Detector.hpp
```diff
@@ -0,0 +1,161 @@
...
+  /// @param name the detecor name
+  /// @param volumes the objets contained by this detector
+  /// @param volumeFinder is a managed Delegate to find the assocaited volume
+  ///
+  /// @note will throw an exception if volumes vector is empty
+  /// @note will throw an exception if duplicate volume names exist
+  /// @note will throw an exception if the delegate is not connected
+  Detector(const std::string& name,
+           const std::vector<std::shared_ptr<DetectorVolume>>& volumes,
+           ManagedDetectorVolumeUpdator&& volumeFinder) noexcept(false);
```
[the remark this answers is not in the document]
asalzburger: I thought it would give the code-reader a clear idea that this may throw an exception.

Core/include/Acts/Geometry/Detector.hpp
```diff
@@ -0,0 +1,161 @@
...
+  std::string m_name = "Unnamed";
+
+  /// Volume store (internal/external)
+  DetectorVolume::ObjectStore<std::shared_ptr<DetectorVolume>> m_volumes;
+
+  /// A volume finder delegate
+  ManagedDetectorVolumeUpdator m_volumeFinder;
+
+  /// Name/index map
+  std::map<std::string, size_t> m_volumeNameIndex;
```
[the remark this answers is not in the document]
asalzburger: Right.

Core/include/Acts/Geometry/Detector.hpp
```diff
@@ -0,0 +1,161 @@
...
+};
+
+inline std::vector<std::shared_ptr<DetectorVolume>>& Detector::volumePtrs() {
```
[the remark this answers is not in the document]
asalzburger: I have this done consistently (with this naming) throughout the PR, I do not care much, but I would like to have it consistently done for volumes, detector volumes, detector.

Core/include/Acts/Geometry/DetectorVolume.hpp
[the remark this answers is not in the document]
asalzburger: Done.

Core/include/Acts/Geometry/DetectorVolume.hpp
```diff
@@ -0,0 +1,577 @@
...
+
+  /// Assign Detector to this volume (for back navigation issues)
+  /// @param detector the parenting detector class
+  void assignDetector(const Detector& detector);
+
+  /// Const access to the detector
+  const Detector* detector() const;
+
+ private:
+  /// Internal construction method that calls the poral genenerator
```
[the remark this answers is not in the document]
asalzburger: Fixed.

Core/include/Acts/Geometry/detail/DetectorVolumeUpdators.hpp
```diff
@@ -0,0 +1,148 @@
...
+  std::vector<const DetectorVolume*> dVolumes = {};
+
+  /// Extract the sub volumes from the volume
+  ///
+  /// @param gctx the geometry contextfor this extraction call (ignored)
+  /// @param nState is the navigation state (ingored)
+  /// @param indices are access indices into the surfaces store
+  inline const DetectorVolume* extract(
+      [[maybe_unused]] const GeometryContext& gctx,
+      [[maybe_unused]] const NavigationState& nState,
```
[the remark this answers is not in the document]
asalzburger: That's right, it's not `maybe_unsused` - it's certainly unused.

Core/include/Acts/Geometry/NavigationDelegates.hpp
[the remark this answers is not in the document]
benjaminhuth: I take back the point, I have misunderstood what was meant by void pointer in this case...

Core/include/Acts/Geometry/Detector.hpp
```diff
@@ -0,0 +1,161 @@
...
+  return m_volumes.internal;
+}
+
+inline const std::vector<const DetectorVolume*>& Detector::volumes() const {
+  return m_volumes.external;
```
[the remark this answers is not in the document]
paulgessinger: Ok I see. You can sort of do that anyway by using an offset from the begin iterator, which I think should compile down to the same thing the way that the vector iterator is implemented. But ok, we can keep it like this and we (maybe I) can see if this can be achieved reasonably with iterator pairs.

Core/include/Acts/Geometry/DetectorVolume.hpp
paulgessinger: I don't think I get it. It's pushed into a Delegate, isn't it? So this specific extractor (and the other structs) do not in fact have any members. I think in that case, I'd prefer these specific functions to just be free, even if others might have members and have to be structs.

Core/include/Acts/Geometry/NavigationDelegates.hpp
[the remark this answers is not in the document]
asalzburger: To answer Paul:
- we do/will do downcasts into the concrete types (the code is not part of this PR), e.g. when displaying the relationship of surfaces inside a grid, or when writing the DetectorVolume to Json (will be part of a follow-up PR).

- however, we never need to cast to `IDelegateImpl`, we can directly do the cast sequence into the possible implementations

Core/include/Acts/Geometry/DetectorVolume.hpp
asalzburger: The point is that this implementation allows for both implementations of an extractor: one with a non-member structure where you have a `static` access method, another one where you extract from a member directly. 
As said, I used both implementations, because they might be useful.

Let me give two examples:
- An extractor of a volume from the detector based on a e.g. a grid or a BoundingBox hierarchy or whatsoever may use an extractor that has access to all volumes in the detector
- An extractor of e.g. a boundary surface of a container-type volume that is built from three volumes (before the entire detector is built and the final list of volumes known) may just keep memory of the three volumes it needs in a private member and thus needs to be a non-static member function

In `detray` we do not allow this second type of access, but that brings the consequence that every volume, when constructed needs to be provided by the detector in order to guarantee fixed indices at construction time of an impartially built detector, such as e.g.

```c++
auto& volume = detector.new_volume("DiamondSensorDetector");
```

I thought this would be too restrictive for the `Acts` core code.

Core/include/Acts/Geometry/NavigationDelegates.hpp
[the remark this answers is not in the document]
asalzburger: Mmmm, maybe it's not possible ?

```sh
/Users/salzburg/Documents/work/dev/acts-experimental/Tests/UnitTests/Core/Utilities/DelegateTests.cpp:230:19: error: 'void' is not a class type
  auto pImplRaw = dynamic_cast<const ProviderImpl*>(mpDel.implementation.get());
                  ^                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

@benjaminhuth @paulgessinger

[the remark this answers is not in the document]
asalzburger: I get the same error with `std::dynamic_pointer_cast`

```c++
auto pImplRaw = std::dynamic_pointer_cast<ProviderImpl>(mpDel.implementation);
```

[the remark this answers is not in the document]
benjaminhuth: Hmm yeah so the problem is with `dynamic_cast` we need to cast from a polymorphic type to a related polymorphic type. But void is not polymorphic, so it is not possible to dynamic_cast from a `void *` to anything I would say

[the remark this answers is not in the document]
benjaminhuth: Maybe we need a reinterpret cast here? but that is kind of dangerous terrain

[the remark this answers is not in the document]
asalzburger: Ok, thanks - I need this down-casting possibility as I said e.g. for writing the delegates out and displaying them. So I will reintroduce a `IManagedDelegateImpl` base class, because a static_cast would need some sort of identification which if mismatches due to a bug does not help us.

Core/include/Acts/Definitions/Common.hpp
```diff
@@ -36,12 +36,37 @@ static constexpr ActsScalar s_curvilinearProjTolerance = 0.999995;
...
-inline constexpr NavigationDirection directionFromStepSize(double value) {
+/// Convert navigation dir to index [0,1] which allows to
+/// store direction dependent objects in std::array<T,2u>
+///
+/// @param nDir is the navigation direction at input
+///
+/// returns either 0 or 1
+inline constexpr size_t indexFromDirection(NavigationDirection nDir) {
+  return static_cast<size_t>((static_cast<int>(nDir) + 1) / 2);
+}
```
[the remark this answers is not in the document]
asalzburger: Ok, done. Also added UnitTests for the `Common` utility functions.

Core/include/Acts/Geometry/Detector.hpp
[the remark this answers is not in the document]
asalzburger: This is coupled to the `iterator` question, at the moment I internally store the `shared_ptr<Object>` in the geometry in order to make sure that the objects are held somewhere and kept in memory. One has access to these internal storage, as long as the parent object `Detector` or `DetectorVolume` are not yet const. 

Having them available in one single vector, is e.g. necessary for volume/surface material assignment which is/can be done after the detector construction.

Core/include/Acts/Geometry/detail/SurfaceCandidatesUpdators.hpp
```diff
@@ -0,0 +1,194 @@
...
+/// Generate a provider for all portals
+///
+/// @return a connected navigationstate updator
+inline static ManagedSurfaceCandidatesUpdator allPortals() {
+  ManagedSurfaceCandidatesUpdator managedUpdator;
+  SurfaceCandidatesUpdator nStateUpdator;
+  nStateUpdator.connect<&portalCandidates>();
+  managedUpdator.delegate = std::move(nStateUpdator);
+  managedUpdator.implementation = nullptr;
+  return managedUpdator;
```
andiwand, line 148: constructor?

Core/include/Acts/Geometry/detail/SurfaceCandidatesUpdators.hpp
```diff
@@ -0,0 +1,194 @@
...
+/// setup with many surfaces
+///
+/// @return a connected navigationstate updator
+inline static ManagedSurfaceCandidatesUpdator allPortalsAndSurfaces() {
+  ManagedSurfaceCandidatesUpdator managedUpdator;
+  SurfaceCandidatesUpdator nStateUpdator;
+  nStateUpdator.connect<&portalAndSurfaceCandidates>();
+  managedUpdator.delegate = std::move(nStateUpdator);
+  managedUpdator.implementation = nullptr;
+  return managedUpdator;
```
andiwand, line 163: constructor?

Core/src/Geometry/DetectorVolume.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include "Acts/Geometry/DetectorVolume.hpp"
+
+#include "Acts/Geometry/NavigationState.hpp"
+#include "Acts/Geometry/Portal.hpp"
+#include "Acts/Geometry/VolumeBounds.hpp"
+#include "Acts/Surfaces/Surface.hpp"
+#include "Acts/Utilities/Enumerate.hpp"
+#include "Acts/Utilities/Helpers.hpp"
+
+#include <assert.h>
```
andiwand, line 18:
```suggestion
#include <cassert>
```

Core/include/Acts/Geometry/Portal.hpp
```diff
@@ -0,0 +1,178 @@
...
+  /// @note Will error if this was not created through the @c makeShared factory
+  ///       since it needs access to the original reference. In C++14 this is
+  ///       undefined behavior, but most likely implemented as a @c bad_weak_ptr
+  ///       exception, in C++17 it is defined as that exception.
+  /// @note Only call this if you need shared ownership of this object.
+  ///
+  /// @return The shared pointer
+  std::shared_ptr<const Portal> getSharedPtr() const;
+
+  Portal() = delete;
```
andiwand, line 87: could be dropped as it is implicitly deleted

Core/include/Acts/Geometry/detail/DetectorVolumeUpdators.hpp
```diff
@@ -0,0 +1,148 @@
...
+  /// @param sVolume the volume to which it points
+  SingleDetectorVolumeImpl(const DetectorVolume* sVolume) noexcept(false)
+      : dVolume(sVolume) {
+    if (sVolume == nullptr) {
+      throw std::invalid_argument(
+          "DetectorVolumeUpdators: nullptr provided, use EndOfWorld instead.");
+    }
+  }
+
+  SingleDetectorVolumeImpl() = delete;
```
andiwand, line 56: could be dropped as it is implicitly deleted

Core/include/Acts/Geometry/detail/DetectorVolumeUpdators.hpp
```diff
@@ -0,0 +1,148 @@
...
+      throw std::invalid_argument(
+          "DetectorVolumeUpdators: mismatching boundaries and volume numbers");
+    }
+    // Initialize the grid entries
+    for (std::size_t ib = 1u; ib < gBoundaries.size(); ++ib) {
+      indexedUpdator.grid.at(ib) = ib - 1;
+    }
+  }
+  // Deleted default constructor
+  BoundVolumesGrid1Impl() = delete;
```
andiwand, line 134: could be dropped as it is implicitly deleted

andiwand (commented):
I can't really provide any more feedback right now. for me it would be good to bring it in and then test it with some real examples

Core/include/Acts/Geometry/detail/SurfaceCandidatesUpdators.hpp
asalzburger: Also done.

Core/src/Geometry/DetectorVolume.cpp
asalzburger: Point taken.

Core/include/Acts/Geometry/Portal.hpp
asalzburger: Yes, but I wanted to make it very clear to the code reader.

Core/include/Acts/Geometry/detail/DetectorVolumeUpdators.hpp
asalzburger: Same.

Core/include/Acts/Geometry/detail/DetectorVolumeUpdators.hpp
asalzburger: And ones more.

asalzburger:
There's one discussion point open, i.e. whether we should give the access to the const object to the iterators.
My suggestion is:
- we bring it in like this, and I try in a separate PR how an iterator access would look like and if it's feasible.

Core/include/Acts/Geometry/DetectorVolume.hpp
paulgessinger: But isn't the ability to have members completely independent from this specific struct?

andiwand (approved):

Core/src/Geometry/Portal.cpp
```diff
@@ -0,0 +1,96 @@
...
+  const auto& position = nState.position;
+  const auto& direction = nState.direction;
+  const Vector3 normal = surface().normal(gctx, position);
+  NavigationDirection nDir = directionFromStepSize(normal.dot(direction));
+  const auto& vUpdator = m_volumeUpdators[indexFromDirection(nDir)].delegate;
+  if (vUpdator.connected()) {
+    vUpdator(gctx, nState);
+  } else {
+    nState.currentVolume = nullptr;
+  }
```
paulgessinger, line 95: Could we not by default set the updator to a free function that sets `nState.currentVolume`? Then you could remove this extra check.

Core/include/Acts/Geometry/detail/NavigationStateUpdators.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+  /// @brief updates the navigation state with a single object that is filled in
+  ///
+  /// @param gctx is the Geometry context of this call
+  /// @param nState the navigation state to which the surfaces are attached
+  ///
+  /// @note this is attaching objects without intersecting nor checking
+  void update([[maybe_unused]] const GeometryContext& gctx,
+              NavigationState& nState) const {
+    filler_type::fill(nState, object);
```
paulgessinger, line 45: Ah, that's why you need the struct. This is an implicit concept here? Do I get this right, @asalzburger?

Core/include/Acts/Geometry/detail/NavigationStateUpdators.hpp
```diff
@@ -0,0 +1,173 @@
...
+class IndexedUpdatorImpl : public IManagedDelegateImpl {
+ public:
+  /// The grid where the indices are stored
+  grid_type grid;
+  /// An extractor helper to get the object(s) from the volume
+  extractor_type extractor;
+  /// These are the cast parameters
+  std::array<BinningValue, grid_type::DIM> casts;
+  /// A transform to be applied to the position
+  Transform3 transform = Transform3::Identity();
```
paulgessinger, line 91: Do these have to be public?

Core/include/Acts/Geometry/detail/NavigationStateUpdators.hpp
```diff
@@ -0,0 +1,173 @@
...
+      : grid(std::move(igrid)), casts(icasts), transform(itr) {}
+
+  /// @brief updates the navigation state with objects from the grid according
+  /// to the filling type AFTER applying `p3loc = transform * p3`
+  ///
+  /// @param gctx is the Geometry context of this call
+  /// @param nState the navigation state to which the surfaces are attached
+  ///
+  /// @note this is attaching objects without intersecting nor checking
+  void update([[maybe_unused]] const GeometryContext& gctx,
```
paulgessinger, line 109: What is `[[maybe_unused]]` needed for here?

Core/include/Acts/Geometry/detail/NavigationStateUpdators.hpp
```diff
@@ -0,0 +1,173 @@
...
+/// This is a chained extractor/filler implementation
+/// Since there is no control whether it is a static or
+/// payload extractor, these have to be provided by a tuple
+///
+/// @tparam updators_t the updators that will be called in sequence
+template <typename... updators_t>
+class ChainedUpdatorImpl {
+ public:
+  // The stored updators
+  std::tuple<updators_t...> updators;
```
paulgessinger, line 150: Also here: does this need to be public?

Core/include/Acts/Geometry/detail/PortalGenerators.hpp
```diff
@@ -0,0 +1,97 @@
...
+    NavigationDirection outsideDir =
+        (insideDir == Acts::NavigationDirection::Forward)
+            ? Acts::NavigationDirection::Backward
+            : Acts::NavigationDirection::Forward;
+    ManagedDetectorVolumeUpdator managedNullLink{std::move(nullLink), nullptr};
+    portal->assignDetectorVolumeUpdator(outsideDir, std::move(managedNullLink),
+                                        {});
+
+    // Portal is prepared
+    portals.push_back(portal);
```
paulgessinger, line 75:
```suggestion
    portals.push_back(std::move(portal));
```

Core/include/Acts/Geometry/detail/PortalGenerators.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+/// @brief Generator function for creation of portal surfaces
+/// for a cylindrical volume
+///
+/// @param dTransform a contextually resolved transform
+/// @param dBounds the detecor volume bounds
+/// @param dVolume the reference to the detector volume which generates this volume
+///
+/// @return a vector of newly created portals with registered inside volume
+inline static std::vector<std::shared_ptr<Portal>> portals(
```
paulgessinger, line 35: That's a fairly generic function name. Can this be more explicit?

Core/include/Acts/Geometry/detail/PortalHelper.hpp
```diff
@@ -0,0 +1,79 @@
...
+    std::shared_ptr<DetectorVolume> dVolume) noexcept(false) {
+  if (dVolume == nullptr) {
+    throw std::runtime_error(
+        "PortalsAndSubPortals: no detector volume provided.");
+  }
+
+  // Setting link to the mother volume to all sub volumes of this volume
+  for (auto vPtr : dVolume->volumePtrs()) {
+    for (auto pPtr : vPtr->portalPtrs()) {
+      // Creatint a link to the mother
```
paulgessinger, line 47:
```suggestion
      // Creating a link to the mother
```

Core/include/Acts/Utilities/Delegate.hpp
```diff
@@ -189,4 +190,34 @@ class Delegate<R(Args...)> {
...
+  Delegate<R(Args...)> delegate;
+  std::shared_ptr<IManagedDelegateImpl> implementation = nullptr;
+
+  /// The call operator that forwards to the @c Delegate type.
+  /// @param args The arguments to call the contained function with
+  /// @return Return value of the contained function
+  return_type operator()(Args... args) const {
+    return delegate(std::forward<Args>(args)...);
+  }
+};
```
paulgessinger, line 221: Do you mind if I break this out and make a separate PR for this @asalzburger?
I think we can get model the interface much closer to the existing `Delegate` type, and remove a lot of boilerplate around creating these.

Core/src/Geometry/Portal.cpp
asalzburger: Good idea!

Core/include/Acts/Geometry/detail/NavigationStateUpdators.hpp
asalzburger: As you can see, this is one way of implementing the `NavigationStateUpdator` delegate, that's why it's also in `detail`, within here I use some concepts, but it is internally to that implementation.

Core/include/Acts/Utilities/Delegate.hpp
asalzburger: Please, go ahead.

Core/include/Acts/Geometry/detail/NavigationStateUpdators.hpp
asalzburger: No, we can move them to private.

Core/include/Acts/Geometry/detail/PortalGenerators.hpp
asalzburger: Updated.

Core/include/Acts/Geometry/detail/NavigationStateUpdators.hpp
asalzburger: Done.

Core/include/Acts/Geometry/detail/NavigationStateUpdators.hpp
asalzburger: Partly, I moved to private where possible.

Core/include/Acts/Geometry/detail/NavigationStateUpdators.hpp
asalzburger: We will do this, once you have a new `ManagedDelegate<>` submitted.

Core/src/Geometry/Portal.cpp
asalzburger: I will put that on top of your new `ManagedDelegate` proposal.

asalzburger:
I've updated this to the new `OwningDelegate` infrastructure, it made the things quite easier, in fact.

asalzburger:
The debug unit test seems to fail, checking it now.

asalzburger:
Hey @andiwand - I think now all the clang-tidy fixes are also through, and the PR is reworked after @paulgessinger 's introduction of the `OwningDelegate`.

There will be some ironing out of a few things certainly along the way, but since this is under `Experimental` namespace, I hope we can move that in, in order not to block other development.

asalzburger:
@andiwand - if you're happy, please re-approve

asalzburger:
Since @andiwand approved and gave a thumbs up, I will merge that in.

PR #1646, opened 2022-11-03
feat: handle geant4 rotations

andiwand:
until now the geant4 surface mapper only supports translations. this PR adds rotation support as needed in https://github.com/acts-project/acts/pull/1638/


benjaminhuth (approved):
Looks good!

PR #1650, opened 2022-11-03
fix: Fixed Grid Creation

guilhermeAlmeida1:
The current implementation relies on a division rounded to int which is then used to calculate the positions. This leads to a significant propagation of the rounding which causes a significant part of the selected Z range to not be in any bin.

E.g.
zMin = -1186 ; zMax = 1186;
number of z bins =  5;

-> z in [1035, 1186] is not part of any bin


LuisFelipeCoelho (commented):
Thanks for fixing this, you probably just need to change the root hashes for the linux_examples_test

AJPfleger (approved):

LuisFelipeCoelho (approved):

PR #1651, opened 2022-11-04
fix: duplicated configuration in itk.py

LuisFelipeCoelho:
I noticed that `zBinEdges` is defined twice in `itk.py`
This PR removes the unnecessary definition
but I am not sure if this can be considered a bug


AJPfleger (approved):
good catch

PR #1653, opened 2022-11-04
refactor: required for new ITk geometry

timadye:
New element names in `buildITkGeometry` for updated ATLAS ITk and HGTD geometry, `ATLAS-P2-RUN4-01-00-00`.
This supports the new geometry TGeo root file just added to the [`acts-itk`](https://gitlab.cern.ch/atlas/acts-itk) repo.

* Note that without this PR, the latest version of `acts-itk` (tag [`ATLAS-P2-RUN4-01-00-00`](https://gitlab.cern.ch/atlas/acts-itk/-/tree/ATLAS-P2-RUN4-01-00-00)) will not work. `acts-itk` tag [`ATLAS-P2-23`](https://gitlab.cern.ch/atlas/acts-itk/-/tree/ATLAS-P2-23) can be used instead.

* With this PR, `acts-itk` [`master`](https://gitlab.cern.ch/atlas/acts-itk) (or tag [`ATLAS-P2-RUN4-01-00-00`](https://gitlab.cern.ch/atlas/acts-itk/-/tree/ATLAS-P2-RUN4-01-00-00)) should be used.

* The error message from `itk.py` indicating a version mismatch is:
```
ValueError: Incorrect binning configuration found: Number of configurations does not match number of protolayers in subvolume Pixel::Pixel. Quitting.
```

### DIFF
--- a/Examples/Python/python/acts/examples/itk.py
+++ b/Examples/Python/python/acts/examples/itk.py
@@ -53,6 +53,22 @@ def buildITkGeometry(
     equidistant = TGeoDetector.Config.BinningType.equidistant
     arbitrary = TGeoDetector.Config.BinningType.arbitrary
 
+    # ## Create TGeo geometry from `tgeo_fileName = itk-hgtd/ATLAS-ITk-HGTD.tgeo.root`.
+    # The `subVolumeName` and `sensitiveNames` specified below may change with new geometry versions
+    # in the root file (it changed ATLAS-P2-23 -> ATLAS-P2-RUN4-01-00-00).
+    # `TGeoParser` searches the tree below `subVolumeName` for all elements that match any of the
+    # list of `sensitiveNames` wildcards and also fall inside the `rRange`/`zRange` selections.
+    # If no `TGeoDetectorElements`` are found for an ACTS `Volume()`, then `TGeoDetector.create()`
+    # raises an exception along the lines of:
+    # 1. Missing tracking geometry - or
+    # 2. Incorrect binning configuration found: Number of configurations does not match number of protolayers
+    # Unless you know in advance, working out what names to change may not be trivial.
+    # I (@timadye) used a combination of
+    # * adding `printf`s in `Acts::TGeoParser::select()` (useful to find what it found with the old version),
+    # * printing object descendants from root (good for making long lists, but navigation cumbersome), and
+    # * browsing `TGeoManager` with ROOT's `TBrowser` (easy to navigate, but have to scan through long lists by eye).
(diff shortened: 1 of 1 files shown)


Examples/Python/python/acts/examples/itk.py
```diff
@@ -71,8 +71,8 @@ def buildITkGeometry(
                 binToleranceZ=(5 * u.mm, 5 * u.mm),
                 binTolerancePhi=(0.025 * u.mm, 0.025 * u.mm),
                 layers=LayerTriplet(True),
-                subVolumeName=LayerTriplet("Pixel::Pixel"),
-                sensitiveNames=LayerTriplet(["Pixel::siLog"]),
+                subVolumeName=LayerTriplet("ITkPixel__ITkPixelDetector"),
```
tboldagh, line 74: There is realy nothing to comment on here. The names are as they are. 
But I have two suggestions for this file style.
1) maybe a comment (next to this place could be given how this names can be found actually? I.e. It is not any name, and was not guessed. 

2) This file has several very long functions. Can this be partitioned in a more human/screen size chunks? Looking at this very place it seems we could have function: getInnerPixelsVolume, getOuterPixelsVolume, .... 

Maybe this is for another PR.

tboldagh (changes requested):

Examples/Python/python/acts/examples/itk.py
timadye: Both good suggestions, but may be tricky to do.
1. Getting the names was not easy, and I still might not have done it quite right - except that I get very similar performance at the end. I used a combination of adding `printf`s in the loading code (useful to find what it found before), printing object descendants from root (good for making long lists, but not everything easy to get at), and browsing with the `TGeoManager` with the new JSROOT TBrowser (easy to navigate, but have to scan through long lists by eye). Hopefully we won't have another major renaming, so maybe this is the last time.
2. This case replaces the even more verbose JSON 😄 Thinking about this, I don't know if it is clearer having it altogether, or split up.
Can we save this for another PR? It would be good to get this in before we have people hitting an incompatibility with `acts-itk` repo. We'll probably get clashes when people update one or the other, but then we'll be able to suggest to update to the latest for both.

tboldagh: ad 2). Sure. It makes sense to leave it. 
ad 1) "we won't have another major renaming" - I would not count on that, the ITk will be around for a while. 
Some info - along what you just wrote should be somehow available in the code imho. Maybe in the preamble in this file?

tboldagh (approved):
Excellent. Thanks. This should help next generations.

PR #1655, opened 2022-11-04
fix: Skip seeds with negative cotThetaAvg2

paulgessinger:


Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -394,6 +394,10 @@ void SeedFinder<external_spacepoint_t, platform_t>::createSeedsForGroup(
         if (m_config.arithmeticAverageCotTheta) {
           // use arithmetic average
           cotThetaAvg2 = std::pow((cotThetaB + cotThetaT) / 2, 2);
+        } else {
+          if (cotThetaAvg2 < 0) {
+            continue;
+          }
         }
```
AJPfleger, line 401:
```suggestion
        } else if (cotThetaAvg2 < 0) {
          continue;
        }
```

AJPfleger (changes requested):

Core/include/Acts/Seeding/SeedFinder.ipp
paulgessinger: ?

andiwand: to me it would be more readable to have the `cotThetaAvg2 = cotThetaB * cotThetaT;` in the `else` and leave it uninitialized up there.

paulgessinger: Does that even work without a warning?

AJPfleger: Just readability (sry, should have noted that)

I think we can't leave it uninitialized, since we use it later ...

paulgessinger: Let's see.

andiwand: yeah not sure how hard clang-tidy will pick on us

paulgessinger: Ok, same result as before, small output changes.

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -390,10 +390,12 @@ void SeedFinder<external_spacepoint_t, platform_t>::createSeedsForGroup(
...
 
         // use geometric average
-        float cotThetaAvg2 = cotThetaB * cotThetaT;
+        float cotThetaAvg2;
         if (m_config.arithmeticAverageCotTheta) {
           // use arithmetic average
           cotThetaAvg2 = std::pow((cotThetaB + cotThetaT) / 2, 2);
+        } else if (cotThetaAvg2 = cotThetaB * cotThetaT; cotThetaAvg2 <= 0) {
+          continue;
```
andiwand, line 398: in this case I would make the calculation explicit like above and then check for the error case even tho it looks much more verbose

```suggestion
        } else {
          cotThetaAvg2 = cotThetaB * cotThetaT;
          if (cotThetaAvg2 <= 0) {
            continue;
          }
```

andiwand (approved):
cool :+1: will see if clang-tidy header checks will call us out about it soon

Core/include/Acts/Seeding/SeedFinder.ipp
AJPfleger: and also move the comment `// use geometric average` down

LuisFelipeCoelho:
Shouldn't this be changed in the orthogonal seeding as well?

paulgessinger:
Probably, yeah

LuisFelipeCoelho (approved):
Sorry, we don't have cotThetaAvg2 in the orthogonal seeding. Probably I will add it later for the seeding performance writer

paulgessinger:
Let's wait until #1665 is merged.

Core/include/Acts/Seeding/SeedFinder.ipp
paulgessinger: Reverting this as the simplest way to fix the clang-tidy warning about the uninit variable.

paulgessinger:
Seeding reference changes are expected I think. Here's a subset of plots from the residuals and pulls, which shows a few failures: 
[res.html.zip](https://github.com/acts-project/acts/files/10111377/res.html.zip)

I'm guessing this is mostly due to bin changes as the ranges are still not fixed. One thing that I find strange though is the change in `all_res_t_prt`:

image

paulgessinger:
CI passes now with updated reference.

@LuisFelipeCoelho ?

LuisFelipeCoelho (approved):

LuisFelipeCoelho:
looks good to me

PR #1658, opened 2022-11-07
feat: track modifier for examples

andiwand:
while dealing with vertexing issues I thought adding a track parameter modification algorithm to the examples framework might be useful


andiwand:
this is not so useful right now. will close for now

PR #1662, opened 2022-11-09
chore: clang tidy headers

andiwand:
fix clang tidy for header files. also includes fixes for clang-tidy warnings from @paulgessinger 

replacement for https://github.com/acts-project/acts/pull/1440


andiwand:
@paulgessinger looks like cmake refuses to compile header files

paulgessinger:
Did you check if the headers end up in `compile_commands.json`?

paulgessinger:
Maybe we need something like: https://stackoverflow.com/questions/59716373/compile-headers-for-errors-using-cmake ?

andiwand:
> Maybe we need something like: https://stackoverflow.com/questions/59716373/compile-headers-for-errors-using-cmake ?

hmm but don't we include our headers in some of our source files anyway?

andiwand:
> Maybe some we don't? I don't actually fully understand this tbh.

so the source file for a header file does not seem to do the trick either... I need to check how this is supposed to work with header files again

cmake/ActsStaticAnalysis.cmake
```diff
@@ -52,29 +33,13 @@ if(ACTS_RUN_CLANG_TIDY)
...
     list(JOIN _errs "," CLANG_TIDY_ERRORS)
 
     message(STATUS "Enabled errors:")
     foreach(_err ${_errs})
       message(STATUS "|-> ${_err}")
     endforeach()
 
-    
-    set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_COMMAND};-checks=${CLANG_TIDY_CHECKS};-header-filter='.*';-warnings-as-errors=${CLANG_TIDY_ERRORS}")
+    set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_COMMAND};-checks=${CLANG_TIDY_CHECKS};-header-filter=.*;-warnings-as-errors=${CLANG_TIDY_ERRORS}")
```
andiwand, line 43: bash expansion is pure evil :imp:

andiwand:
I think this is ready @paulgessinger @benjaminhuth

Core/include/Acts/EventData/MultiTrajectory.hpp
```diff
@@ -379,7 +379,9 @@ class TrackStateProxy {
   /// @param srf Shared pointer to the surface to set
   /// @note This overload is only present in case @c ReadOnly is false.
   template <bool RO = ReadOnly, typename = std::enable_if_t<!RO>>
-  void setReferenceSurface(std::shared_ptr<const Surface> srf) {
+  void setReferenceSurface(
+      std::shared_ptr<const Surface>
+          srf) {  // NOLINT(performance-unnecessary-value-param)
```
andiwand, line 384: not sure why this is necessary... clang believes the `std::move(srf)` below does not do anything

Alignment/include/ActsAlignment/Kernel/detail/AlignmentEngine.hpp
```diff
@@ -118,7 +118,7 @@ TrackAlignmentState trackAlignmentState(
   std::vector<std::pair<size_t, bool>> measurementStates;
   measurementStates.reserve(15);
   // Number of smoothed states on the track
-  size_t nSmoothedStates = 0;
+  size_t nSmoothedStates = 0;  // NOLINT(-Werror,-Wunused-but-set-variable)
```
andiwand, line 121: intended unused?

paulgessinger:
Should we remove the HTML report until I get around to fix codereport?

Alignment/include/ActsAlignment/Kernel/detail/AlignmentEngine.hpp
benjaminhuth: hmm does clang-tidy ignore the `[[maybe_unused]]` label?

Core/include/Acts/EventData/MultiTrajectory.hpp
benjaminhuth: I think we have a implicit move here, see https://stackoverflow.com/questions/9779079/why-does-c11-have-implicit-moves-for-value-parameters-but-not-for-rvalue-para
So clang-tidy might be correct and we can remove the `std::move`...

cmake/ActsStaticAnalysis.cmake
benjaminhuth: indeed :D

andiwand:
attaching a full log of a local clang tidy run with examples and unit tests enabled

[clang-tidy.log](https://github.com/acts-project/acts/files/10002897/clang-tidy.log)

andiwand:
looks like this works now. https://gitlab.cern.ch/acts/ci-bridge/-/jobs/25840196 is failing because of the html report generation

benjaminhuth (approved):
Let's get this in!

benjaminhuth (approved):

PR #1669, opened 2022-11-10
ci: Change physmon PR determination

paulgessinger:
Tries to use an action to make determining the PR more robust. Kodiak updates the PR with a push, not a synchronize, so the report workflow did not run correctly.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -22,226 +22,226 @@ env:
   CCACHE_KEY_SUFFIX: r1
 
 jobs:
-  lcg:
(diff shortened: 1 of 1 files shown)


benjaminhuth (approved):
Let's hope this solves this issue reliable!

benjaminhuth:
Ah but the PR number job is failing @paulgessinger

paulgessinger:
Ok @benjaminhuth, hopefully now it'll work.

benjaminhuth:
Unfortunately the physmon job it still fails @paulgessinger

PR #1670, opened 2022-11-10
refactor: replace track params with trajectories in examples

andiwand:
somehow we ended up passing through track parameters and trajectory tips through our algorithm chain which can be replaced by trajectories

the only caveat is particle smearing which will produce track parameters which should be passed down the chain. so track selector and vertexing will still except track parameters as an input


Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TrackSelector.cpp
```diff
@@ -55,49 +54,64 @@ ActsExamples::ProcessCode ActsExamples::TrackSelector::execute(
...
+      const auto& trk = inputTrackParameters[i];
+      if (isValidTrack(trk)) {
+        outputTrackParameters.push_back(trk);
+      }
+    }
+    outputTrackParameters.shrink_to_fit();
+
+    ACTS_DEBUG("event " << ctx.eventNumber << " selected "
+                        << outputTrackParameters.size() << " from "
+                        << inputTrackParameters.size() << " tracks");
```
benjaminhuth, line 75: maybe specify in the debug log that we are in the parameters-branch

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TrackSelector.cpp
```diff
@@ -55,49 +54,64 @@ ActsExamples::ProcessCode ActsExamples::TrackSelector::execute(
...
+        parameters.emplace(tip, trajectories.trackParameters(tip));
+        ++outputCount;
       }
+
+      outputTrajectories.emplace_back(trajectories.multiTrajectoryPtr(), tips,
+                                      parameters);
     }
+
+    ACTS_DEBUG("event " << ctx.eventNumber << " selected " << outputCount
+                        << " from " << inputCount << " tracks");
```
benjaminhuth, line 111: maybe indicate in the debug log that we are in the trajectories branch

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TrackSelector.hpp
```diff
@@ -19,14 +19,14 @@ namespace ActsExamples {
 class TrackSelector final : public BareAlgorithm {
  public:
   struct Config {
-    /// Input track parameters collection.
+    /// Optional. Input track parameters collection.
```
benjaminhuth, line 22: maybe add in the comment that parameters and trajectories are mutually exclusive (same below I guess)

Examples/Algorithms/Vertexing/src/IterativeVertexFinderAlgorithm.cpp
```diff
@@ -56,10 +58,31 @@ ActsExamples::IterativeVertexFinderAlgorithm::IterativeVertexFinderAlgorithm(
...
+  std::vector<const Acts::BoundTrackParameters*> inputTrackPointers;
+
+  if (!m_cfg.inputTrackParameters.empty()) {
+    const auto& tmp =
+        ctx.eventStore.get<std::vector<Acts::BoundTrackParameters>>(
+            m_cfg.inputTrackParameters);
+    inputTrackParameters = tmp;
+    inputTrackPointers = makeTrackParametersPointerContainer(tmp);
+  } else {
+    const auto& inputTrajectories =
```
benjaminhuth, line 72: I would argue to put parts of this in a utility function in a private header file in `Examples/Algorithms/Vertexing/src`, e.g. like this

```c++
template<typename config_t>
auto makeParameterContainers(const config_t &config, const Acts::Examples::AlgorithmContext &ctx) {
  // ...
  return std::make_tuple(vec, vecPointer);
}

because the string identifier variables are named equally in the three algorithms, right?

**Review by benjaminhuth — CHANGES_REQUESTED** [review]
Very nice, I only would like this common utility function to remove at least a bit code duplication.

File: Examples/Io/Root/src/RootVertexPerformanceWriter.cpp [hunk 5]
```diff
@@ -238,50 +238,56 @@ ActsExamples::ProcessCode ActsExamples::RootVertexPerformanceWriter::writeT(
...
-      size_t nMajorityHits = particleHitCounts.front().hitCount;
-
-      auto trajState = Acts::MultiTrajectoryHelpers::trajectoryState(
-          traj.multiTrajectory(), tip);
-
-      if (nMajorityHits * 1. / trajState.nMeasurements <
-          m_cfg.truthMatchProbMin) {
-        continue;
-      }
+    for (const auto& trajectories : inputTrajectories) {
```
paulgessinger, line 250: Maybe `trajectory` (singular) for clarity?

paulgessinger (commented):
looks good i think

Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
andiwand: this is generally very confusing to me... `inputTrajectories` is a trajector**ies** container. so even the elements are multiple trajector**ies**. this is not well empathized in the examples framework so I would stick with the current "convention"

benjaminhuth (approved):

Examples/Io/Root/src/RootVertexPerformanceWriter.cpp
benjaminhuth: I would agree that this is not optimal, and there is no obvious better solution than this, so lets keep it like this.

PR #1672, opened 2022-11-11
feat: skip unnecessary iterations in seed filter if SPs are sorted by curvature

LuisFelipeCoelho:
This is a small change to avoid unnecessary iteration in the seed filter based on the sorting of the outer SPs.
The seeding performance is the same but slightly faster.


Core/include/Acts/Seeding/SeedFilter.ipp
```diff
@@ -65,34 +65,44 @@ void SeedFilter<external_spacepoint_t>::filterSeeds_2SpFixed(
               });
   }
 
-  for (auto& i : idx) {
```
tboldagh, line 68: Just reading this code as is, it may be better to give variable `idx` a more verbose name: i.e. `topSpSize`
same for the `i`, maybe topSpIndex?

Core/include/Acts/Seeding/SeedFilter.ipp
```diff
@@ -65,34 +65,44 @@ void SeedFilter<external_spacepoint_t>::filterSeeds_2SpFixed(
...
-                                                    : topSpVec[j]->radius();
+      float otherTop_r = m_cfg.useDeltaRorTopRadius
+                             ? topSpVec[compatibleTopSPIdx]->deltaR()
+                             : topSpVec[compatibleTopSPIdx]->radius();
 
       // curvature difference within limits?
-      if (invHelixDiameterVec[j] < lowerLimitCurv) {
+      if (invHelixDiameterVec[compatibleTopSPIdx] < lowerLimitCurv) {
+        if (m_cfg.curvatureSortingInFilter) {
+          startCompTopLoopIdx = j;
```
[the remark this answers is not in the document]
tboldagh: Another suggestion about the name. The `j` is good if the loop is small. For anything longer than 10 lines it is better us use some more meaningful name.

Core/include/Acts/Seeding/SeedFilter.ipp
```diff
@@ -65,34 +65,44 @@ void SeedFilter<external_spacepoint_t>::filterSeeds_2SpFixed(
...
+
+    for (size_t j = startCompTopLoopIdx; j < idx.size(); j++) {
+      size_t compatibleTopSPIdx = idx[j];
+      if (compatibleTopSPIdx == topSPIdx) {
         continue;
       }
 
-      float otherTop_r = m_cfg.useDeltaRorTopRadius ? topSpVec[j]->deltaR()
-                                                    : topSpVec[j]->radius();
+      float otherTop_r = m_cfg.useDeltaRorTopRadius
```
tboldagh, line 93: Variable naming comment, camelCase style is used in this code, so maybe `otherTopR` ?

Core/include/Acts/Seeding/SeedFilter.ipp
```diff
@@ -65,34 +65,44 @@ void SeedFilter<external_spacepoint_t>::filterSeeds_2SpFixed(
               });
   }
 
-  for (auto& i : idx) {
+  size_t startCompTopLoopIdx = 0;
+  // loop over top SPs and other compatible top SP candidates
```
tboldagh, line 69: Is there an important assumption about the SPs being sorted by phi used here? May be worth mentioning in the comment.

tboldagh (changes requested):
The change looks alright.
Just had few suggestions that could improve understanding the code by others.  
I would like to make a general comment. As of now the function has 200+ lines of quite involved logic. 
It may help to make it easier to follow if it could be partitioned in a few smaller functions with meaningful names.

LuisFelipeCoelho:
I tested this with 500 ttbar events with PU=200 in standalone and it seems to be slightly faster FYI: @paulgessinger 
ITkSeeding_ttbar_mu200_PPPtime

tboldagh (approved):
These changes look good to me.

PR #1673, opened 2022-11-11
fix: remove double to bool conversion in vertexing

andiwand:
dicovered in https://github.com/acts-project/acts/pull/1662


Core/include/Acts/Vertexing/AdaptiveGridTrackDensity.ipp
```diff
@@ -87,7 +87,7 @@ Acts::AdaptiveGridTrackDensity<trkGridSize>::addTrack(
 
   // Check if current track does affect grid density
   // in central bins at z-axis
-  if ((std::abs(dOffset) > trkGridSize - 1) / 2.) {
+  if (std::abs(dOffset) > (trkGridSize - 1) / 2.) {
```
andiwand, line 90: @baschlag do you know how this is supposed to work?

same below

paulgessinger:
@baschlag #1662

baschlag:
However, I think it should not have an impact on the physics performance of the Grid Seeder (which we are not monitoring anyway, are we?).
I implemented this check to only use tracks in the seeder that will also contribute with their track densities to the overall density. For that I'm checking if the track d0 is not too far away from the beam axis. In the current/previous version (without the fix) we're passing too many tracks to the seeder, but they will not contribute as their track density along the beam axis is negligible and is not calculated in the seeder.
So the fix definitely makes sense, but I supposed it will not have an impact on the physics performance. Only the CPU performance should be slightly enhanced (which is nice)

andiwand:
okay thanks for the insight @baschlag ! I will try to work around the unit tests and check the physics performance. if this checks out the unit tests have to be adopted anyways

andiwand:
updated the unit tests now. not sure how much sense this makes tho @baschlag @paulgessinger

benjaminhuth (approved):
Okay, let's get this in finally.

PR #1674, opened 2022-11-14
feat: Owning delegate

paulgessinger:
This PR changes `Acts::Delegate` so it can **optionally** own it's payload (i.e. isn't copyable and destroys it correctly). The default remains to not own it. The memory footprint of a non-owning stays as before, the owning delegate gets an extra pointer to a deleter function. I added an alias `OwningDelegate`.


asalzburger:
Did you test whether you can `dynamic_cast` the payload implementation back to the original type?
If so, can we add that to the UnitTests ?

This is, as I wrote, functionality that we need to support.

paulgessinger:
Is that still something we need @asalzburger? I thought we discussed that we could get rid of that pattern.

paulgessinger:
It does not support this. I have an alternative that I want to test.

paulgessinger:
If you're happy, @asalzburger, can we merge this one first?

PR #1675, opened 2022-11-15
feat: Allow delegate to hold non-void instance 

paulgessinger:
In case the holder type is not void, the delegate exposes the instance
as a pointer so you can downcast it if desired.

This includes the changes from #1674


paulgessinger:
fyi @asalzburger. What do you think?

Tests/UnitTests/Core/Utilities/DelegateTests.cpp
```diff
@@ -199,4 +201,212 @@ BOOST_AUTO_TEST_CASE(StatefullLambdas) {
...
+    d.connect<&ConcreteDelegate::func>(&c);
+    BOOST_CHECK_EQUAL(d(), "derived");
+
+    const auto* instance = d.instance();
+    static_assert(
+        std::is_same_v<
+            std::remove_const_t<std::remove_pointer_t<decltype(instance)>>,
+            ConcreteDelegate>,
+        "Did not get correct instance pointer");
+    BOOST_CHECK_NE(dynamic_cast<const DelegateInterface*>(instance), nullptr);
```
asalzburger, line 398: That's perfect, actually - I will be able to define dedicated interfaces for the navigation delegates.

Core/include/Acts/Utilities/Delegate.hpp
```diff
@@ -12,49 +12,72 @@
...
 
 namespace Acts {
 
-template <typename>
+/// Ownership enum for @c Delegate
+enum class DelegateType { Owning, NonOwning };
+
+// Specialization needed for defaulting ownership and for R(Args...) syntax
+template <typename, typename H = void, DelegateType = DelegateType::NonOwning>
 class Delegate;
```
asalzburger, line 26: I see, the introduction of `H` does the trick here.

asalzburger (approved):

asalzburger:
That should do the trick for me - I think that's a good compromise now.

asalzburger (approved):
Reapproving now.

PR #1676, opened 2022-11-15
build: Add -Wfloat-conversion to clang warnings

benjaminhuth:
this seems at least in clang catch implicit float-to-bool conversion.


andiwand (approved):
lets see if that catches https://github.com/acts-project/acts/pull/1673 too

paulgessinger:
This is a good idea. Just needs the code fixes now 😄

benjaminhuth:
I would propose a Fix-a-thon for this :D

paulgessinger:
I fixed all the conversion warnings with `static_cast` or more clever replacements.

cmake/ActsCompilerOptions.cmake
```diff
@@ -4,12 +4,16 @@ if (NOT CMAKE_BUILD_TYPE)
...
+set(ACTS_CXX_FLAGS "-Wall -Wextra -Wpedantic -Wshadow -Wfloat-conversion -Wno-unused-local-typedefs")
 set(ACTS_CXX_FLAGS_DEBUG "--coverage")
 set(ACTS_CXX_FLAGS_MINSIZEREL "")
 set(ACTS_CXX_FLAGS_RELEASE "")
 set(ACTS_CXX_FLAGS_RELWITHDEBINFO "")
 
+#if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang") 
+#  set(ACTS_CXX_FLAGS "${ACTS_CXX_FLAGS} -Wfloat-conversion")
+#endif()
+
```
paulgessinger, line 16: What was this about?

benjaminhuth: Hmm I think clang and gcc behave differently on this (see also the GCC builds failing)
I think we wanted to test that back then...

benjaminhuth: I think gcc does not allow an that fine-grained warning configuration, so from my point of view it would be an option to just enable the warnings on clang builds

benjaminhuth:
Would it be okay to merge this with the warning only enabled for clang @paulgessinger? Then we have at least these changes in for now.

Core/include/Acts/TrackFitting/detail/KLMixtureReduction.hpp
```diff
@@ -129,10 +129,6 @@ class SymmetricKLDistanceMatrix {
     setAssociated(n, m_mask, [&](std::size_t, std::size_t) { return false; });
   }
 
-  void setAssociatedDistances(std::size_t n, ActsScalar value) {
-    setAssociated(n, m_mask, [&](std::size_t, std::size_t) { return value; });
-  }
-
```
paulgessinger, line 135: This change is unrelated right?

benjaminhuth: No, this actually triggers the `-Wfloat-conversion`. I actually introduces this function/buga few weeks ago during a refactoring, but it was unused so the bug was never visible in the results...

PR #1678, opened 2022-11-16
feat: CSV track param reader

andiwand:
adds a CSV track parameter reader to the examples


AJPfleger (commented):
Just marginally a part of this PR, but is there I reason, why we have the header file `Examples/Io/Csv/src/CsvOutputData.hpp` in the `src`-directory instead of `include`?

Also I think we need to update the (c)-dates in the files, we have modified.

andiwand:
> is there I reason, why we have the header file `Examples/Io/Csv/src/CsvOutputData.hpp` in the `src`-directory instead of `include`?

I can only guess but it looks like the csv structs are not used anywhere outside and are therefore more like an implementation detail.

AJPfleger:
Ups, I see, I was very vague. With the year-update I meant `2019-2022`

Examples/Io/Csv/src/CsvOutputData.hpp
```diff
@@ -1,6 +1,6 @@
 // This file is part of the Acts project.
 //
-// Copyright (C) 2019 CERN for the benefit of the Acts project
+// Copyright (C) 2022 CERN for the benefit of the Acts project
```
andiwand, line 3:
```suggestion
// Copyright (C) 2019-2022 CERN for the benefit of the Acts project
```

AJPfleger:
> > ... is there I reason, why we have the header file `Examples/Io/Csv/src/CsvOutputData.hpp` in the `src`-directory instead of `include`?
> 
> I can only guess but it looks like the csv structs are not used anywhere outside and are therefore more like an implementation detail.

@paulgessinger ?

asalzburger (approved):
Looks good,  let's merge.

PR #1679, opened 2022-11-17
fix: OwningDelegate holder type

paulgessinger:
Missed a combination in the test and one typedef mismatch

### DIFF
--- a/Core/include/Acts/Utilities/Delegate.hpp
+++ b/Core/include/Acts/Utilities/Delegate.hpp
@@ -50,7 +50,7 @@ class Delegate<R(Args...), H, O> {
 
   using function_ptr_type = return_type (*)(Args...);
 
-  using deleter_type = void (*)(const void *);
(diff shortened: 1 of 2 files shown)


Core/include/Acts/Utilities/Delegate.hpp
```diff
@@ -50,7 +50,7 @@ class Delegate<R(Args...), H, O> {
 
   using function_ptr_type = return_type (*)(Args...);
 
-  using deleter_type = void (*)(const void *);
+  using deleter_type = void (*)(const holder_type *);
```
asalzburger, line 53: Ah right, that makes sense!

asalzburger (approved):

asalzburger:
Thanks @paulgessinger for the quick fix, that makes totally sense now. 
Let's wait for the CI to finish and then merge it in if things are fine.

andiwand (approved):

PR #1680, opened 2022-11-21
chore: ttbar vertexing for physmon

andiwand:
change event generation for vertexing performance to ttbar


andiwand:
this seems to be overkill

Issue #1657, opened 2022-11-04
Analysis apps return different output with different number of threads while reconstruction


andiwand:
In https://github.com/acts-project/acts/pull/1654 we observed that the analysis apps will return different outputs if the reconstruction ran on multiple threads. Our suspicion is that the event order will modify the output of the program.

I attached the observed and reference root files below.

[root_files.zip](https://github.com/acts-project/acts/files/9939321/root_files.zip)

[8_Physics performance checks.txt](https://github.com/acts-project/acts/files/9939426/8_Physics.performance.checks.txt)

andiwand:
it looks like root will produce different histograms when your data is shuffled. our output seems to be equivalent except for the order, which comes from multi threading.

not sure what the solution to this is.
- we could sort the root file before creating the histograms. I had no success sorting the root file in memory tho
- otherwise we have to refactor the analysis scripts which might be a ton of work
  - or we move them to uproot and rewrite them

any opinions @asalzburger @paulgessinger @benjaminhuth

HadrienG2:
> it looks like root will produce different histograms when your data is shuffled. our output seems to be equivalent except for the order, which comes from multi threading.

That makes sense, floating point summation is not associative so if you do summation in different order, you get slightly different results.

The solution IMO is to only test for bitwise histogram reproducibility when running in sequential mode, and resort to some tolerant approximate comparison in the parallel case (may be a simple matter of checking that bins are within N ulps of the expected result, or a fancier histogram statistical test).

benjaminhuth:
I agree.

I think the purpose of physmon is to check that the general performance of our algorithms is constant (e.g., I would also say it would be good if all that works with a random seed instead of a fixed one). However, if e.g. reordering let these tests fail, we are again close to a bitwise-test, and for that we have the root-hashes.

So I would argue instead of refactoring our tools we should think about how we can adjust the tests that they are not so sensitive...

andiwand:
> That could actually make sense. Making a (weighted) histogram is basically computing the sum of a bunch of floats, and floating point summation is not associative, so if you do summation in different order, you get slightly different results.

I thought in order to decide on the bins it has to do some min/max or sorting of the data but it looks like it doesn't. not sure how this works

so we could also use static binning but not sure how to come up with good values

> So I would argue instead of refactoring our tools we should think about how we can adjust the tests that they are not so sensitive...

[...]

PR #1654, opened 2022-11-04
ci: Run physmon jobs in parallel

benjaminhuth:
Some of the physmon jobs had `numThreads=1`, I think it would speed up the physmon to run all things in parallel (`numThreads=-1`).

PR #1681, opened 2022-11-21
chore: run physmon in parallel

andiwand:
followup to https://github.com/acts-project/acts/issues/1657 and https://github.com/acts-project/acts/pull/1654

let see if this works


andiwand:
sadly this is still crashing

benjaminhuth (approved):
Seems reasonable to me, we can think about putting this back in if we know that it works reliable!

benjaminhuth (approved):

PR #1682, opened 2022-11-22
feat: HelicalLinearizer accepts logger

paulgessinger:


Examples/Algorithms/Vertexing/src/IterativeVertexFinderAlgorithm.cpp
```diff
@@ -93,8 +93,8 @@ ActsExamples::ProcessCode ActsExamples::IterativeVertexFinderAlgorithm::execute(
   VertexSeeder::Config seederCfg(ipEst);
   VertexSeeder seeder(seederCfg);
   // Set up the actual vertex finder
-  VertexFinder::Config finderCfg(vertexFitter, linearizer, std::move(seeder),
-                                 ipEst);
+  VertexFinder::Config finderCfg(vertexFitter, std::move(linearizer),
```
andiwand, line 96: should we pass the logger down to the linearizer in the algorithm?

andiwand (approved):
looks good just one question about the default behavior in the examples framework

Examples/Algorithms/Vertexing/src/IterativeVertexFinderAlgorithm.cpp
andiwand: yeah exactly. I think there is a logger available in the algorithm which could be passed down?

paulgessinger: Not really easy. You can't pass in the algorithm logger because it's a unique ptr. There's also no good way to get the log level from the logger. I can work around this if we think it's needed.

andiwand: no thats fine. I just thought it might be nice to pass the logger down

PR #1692, opened 2022-11-23
fix: root hash only for last cycle

andiwand:
calculate root hash only for the last cycle

related to https://github.com/acts-project/acts/issues/1657 and https://github.com/acts-project/acts/pull/1654

cc @benjaminhuth


benjaminhuth:
Ah nice dass uproot das kann!

benjaminhuth (approved):

Issue #1690, opened 2022-11-23
Refactor Seeding to avoid repeated Finders instantiations


tboldagh:
At ACTS dev. weekly 22.11.22 we discussed and agreed that SeedFinder(s) should be refactored so that clients (example algorithm in ACTS or ATLAS Athena or ... ) do not need to reinstantiate it for every event. 

In practice this means that seed finding method should get part Options object that changes from event to event in the methods that process the data rather than at construction. 
Suitable Option objects were introduced in PR #1630.

The changes can happen in steps, one for each specific Seeding algorithm.
FI @LuisFelipeCoelho @stephenswat

PR #1693, opened 2022-11-24
refactor!: Refactored SeedFinder to avoid the need to recreate it for every event

tboldagh:
This PR is a technical refactor of the SeedFinder so that it does not need to be reinstantiated in every event. 
Instead it would be made only once, properly configured and then only its 
`createSeedsForGroup` called for events. 

BREAKING CHANGE
The arguments to SeedFinder constructor have changed and the arguments needed when calling it's main method. 

On the way, the SeedFinderConfig object was slimmed from those quantities that can't be filled w/o knowledge about the event (i.e. B field). Python binding for those properties that are overwritten anayways: `highland` and `maxScatteringAngle2`, `pTPerHelixRadius`, `minHelixDiameter2` and  `pT2perRadius` are removed.  @LuisFelipeCoelho - please check me on this.

Tested on standalone ODD full chain. It is marginally faster now. 

The OrthogonalSeedFinder should be likely refactored in the same way? For later PR. 
It addresses issue #1690 (not closing it)


Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -16,35 +16,36 @@ namespace Acts {
...
+  // helix radius in homogeneous magnetic field. Units are Kilotesla, MeV and
+  // millimeter
+  // TODO: change using ACTS units
+  const double pTPerHelixRadius = 300. * options.bFieldInZ;
+  const double minHelixDiameter2 =
+      std::pow(m_config.minPt * 2 / pTPerHelixRadius, 2);
+  const double pT2perRadius = std::pow(m_config.highland / pTPerHelixRadius, 2);
+  const double sigmapT2perRadius =
+      pT2perRadius * std::pow(2 * m_config.sigmaScattering, 2);
+
```
LuisFelipeCoelho, line 48: Since createSeedsForGroup is called for every middle SP do you think we could perform some of these calculations only once?

LuisFelipeCoelho (changes requested):
Thank you for this PR! 
From the monitoring plots it appears that it leads to a loss of efficiency. Probably a small error that we want to understand.

I agree that the OrthogonalSeedFinder should be refactored in the same way (in another PR if you prefer)

LuisFelipeCoelho:
I also agree that we should remove from the python binding the parameters that are overwritten: highland and maxScatteringAngle2, pTPerHelixRadius, minHelixDiameter2 and pT2perRadius

Core/include/Acts/Seeding/SeedFinder.ipp
tboldagh: This is done. Please have a look.

LuisFelipeCoelho (approved):
Thank you for these really nice changes.
clang_tidy is complaining about something, but apart from that I think this can go in.

tboldagh:
> Thank you for these really nice changes.
> clang_tidy is complaining about something, but apart from that I think this can go in.

Thanks. There is one thing I am unsure about. That is the protections agains units conversion are aimplemented following ACTS sprit. I.e. there are exceptions now, but in principle they can be avoided i.e. toInternalUnits could become no-op method if config is in internal units already.

LuisFelipeCoelho:
@tboldagh I am not sure why we have two different units. The units from the config are mostly the same as in Athena configuration, if we change it it could be a bit confusing, but I don't have a strong opinion about that. Maybe we could discuss that in the meeting

tboldagh:
@LuisFelipeCoelho - I looked at the plots and there seem to be difference  in the seeding. It is a bit strange to see it in pt put but not eta but maybe it is a plot resolution so something like this.  My question is therefore, we should aim at exact match, right?

tboldagh:
> Thank you for these really nice changes.
> clang_tidy is complaining about something, but apart from that I think this can go in.

It seems that all stuff is sorted out. Can you approve this PR. I think this is needed to move forward.

LuisFelipeCoelho (approved):

PR #1695, opened 2022-11-24
feat: Any type with configurable small buffer optimization

paulgessinger:


paulgessinger:
I ran valgrind on the unit tests which indicates I'm not leaking memory:

```console
$ valgrind --tool=memcheck bin/ActsUnitTestAny
==3104117== Memcheck, a memory error detector
==3104117== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==3104117== Using Valgrind-3.15.0 and LibVEX; rerun with -h for copyright info
==3104117== Command: bin/ActsUnitTestAny
==3104117==
Running 8 test cases...

*** No errors detected
==3104117==
==3104117== HEAP SUMMARY:
==3104117==     in use at exit: 0 bytes in 0 blocks
==3104117==   total heap usage: 13,141 allocs, 13,141 frees, 5,494,543 bytes allocated
==3104117==
==3104117== All heap blocks were freed -- no leaks are possible
==3104117==
==3104117== For lists of detected and suppressed errors, rerun with: -s
==3104117== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

For some reason, heaptrack thinks there is a memory leak, however:

```console
$ heaptrack bin/ActsUnitTestAny
heaptrack output will be written to "/home/pagessin/dev/acts/build/heaptrack.ActsUnitTestAny.3104211.gz"
starting application, this might take some time...
Running 8 test cases...

*** No errors detected
free(): invalid pointer
Aborted (core dumped)
heaptrack stats:
        allocations:            13141
        leaked allocations:     376
        temporary allocations:  10044
Heaptrack finished! Now run the following to investigate the data:
```

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,427 @@
...
+      // construct into local buffer
+      /*U* ptr =*/new (m_data.data()) U(std::forward<Args>(args)...);
+    } else {
+      // too large, heap allocate
+      U* heap = new U(std::forward<Args>(args)...);
+      *reinterpret_cast<U**>(m_data.data()) = heap;
+    }
+  }
+
+#if defined(_ENABLE_ACTS_ANY_DEBUG)
```
benjaminhuth, line 92: Is this really necessary, the `_ACTS_ANY_DEBUG` macro expands to nothing if `_ENABLE_ACTS_ANY_DEBUG` is not defined, right?

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,427 @@
...
+      *reinterpret_cast<U**>(m_data.data()) = heap;
+      _ACTS_ANY_DEBUG_BUFFER("-> buffer after", m_data);
+    }
+  }
+
+  template <typename T>
+  T& as() {
+    static_assert(std::is_same_v<T, std::decay_t<T>>,
+                  "Please pass the raw type, no const or ref");
+    assert(makeHandler<T>() == m_handler && "Bad type access");
```
benjaminhuth, line 129: Are we sure we want the check only in an assert, and not e.g. with an exception? Seems quite dangerous to me...
Additionally that would also allow this behaviour to be unit-testable...

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,427 @@
...
+  ~AnyBase() { destroy(); }
+
+  AnyBase(const AnyBase& other) {
+    if (m_handler == nullptr) {  // this object is empty
+      m_handler = other.m_handler;
+      assert(m_handler != nullptr && "Handler is null");
+      copyConstruct(other);
+    } else {
+      // @TODO: Support assigning between different types
+      assert(m_handler == other.m_handler);
```
benjaminhuth, line 168: Same as above, I would actually be in favour of an exception

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,427 @@
...
+    assert(from != nullptr && "Source is null");
+    assert(to != nullptr && "Target is null");
+
+    const T* _from = static_cast<const T*>(from);
+    T* _to = static_cast<T*>(to);
+
+    (*_to) = *_from;
+  }
+
+  alignas(std::max_align_t) std::array<char, SIZE> m_data{};
```
benjaminhuth, line 417: Maybe use `std::byte` instead of `char`?

benjaminhuth (approved):
Quite impressive work I would say! Just some minor comments, nothing critical...

Core/include/Acts/Utilities/Any.hpp
HadrienG2, line 417: Note that max_align_t is not enough for Eigen types, which may require AVX (32B) or AVX-512 (64B) SIMD alignment. It's only meant to provide the largest possible alignement for a scalar type :

> std::max_align_t is a [trivial](https://en.cppreference.com/w/cpp/named_req/TrivialType) [standard-layout](https://en.cppreference.com/w/cpp/named_req/StandardLayoutType) type whose [alignment requirement](https://en.cppreference.com/w/cpp/language/object#Alignment) is at least as strict (as large) as that of every scalar type.

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,427 @@
...
+      // stored on heap: just copy the pointer
+      *reinterpret_cast<void**>(m_data.data()) =
+          *reinterpret_cast<void**>(fromAny.m_data.data());
+      // do not delete in moved-from any
+      fromAny.m_handler = nullptr;
+      return;
+    }
+
+    if (m_handler->move == nullptr) {
+      // trivially move constructible -> trivially movable
```
HadrienG2, line 311: How about just "trivially movable" ?

HadrienG2 (commented):
I went through this and it's not immediately obvious to me why an invalid free is observed in the heaptrack case, but have a few comments anyway.

HadrienG2:
For context, I spotted what's going on in the heaptrack test above. Notice how the test output changes from...

```
Running 8 test cases...

*** No errors detected
```

...in the valgrind case to...

```
Running 8 test cases...

*** No errors detected
free(): invalid pointer
Aborted (core dumped)
```

...in the heaptrack case. What this means is that the memory allocator is bombing on us in the heaptrack case, hinting at some possible misuse of it.

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,427 @@
...
+      return *reinterpret_cast<const T*>(ptr);
+    }
+  }
+
+  ~AnyBase() { destroy(); }
+
+  AnyBase(const AnyBase& other) {
+    if (m_handler == nullptr) {  // this object is empty
+      m_handler = other.m_handler;
+      assert(m_handler != nullptr && "Handler is null");
```
HadrienG2, line 164: Why is it considered bad to assign from a null object?

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,427 @@
...
+      const void* ptr = *reinterpret_cast<void* const*>(m_data.data());
+      _ACTS_ANY_DEBUG("As heap: " << ptr);
+      return *reinterpret_cast<const T*>(ptr);
+    }
+  }
+
+  ~AnyBase() { destroy(); }
+
+  AnyBase(const AnyBase& other) {
+    if (m_handler == nullptr) {  // this object is empty
```
HadrienG2, line 162: This check is always true in a constructor (same for move constructor below).

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,427 @@
...
+      void* ptr = m_data.data();
+      if (m_handler->typeSize > SIZE) {
+        // stored on heap: interpret buffer as pointer
+        _ACTS_ANY_DEBUG("Pre-destroy data: " << (void*)m_data.data());
+
+        _ACTS_ANY_DEBUG_BUFFER("-> buffer pre-destroy", m_data);
+
+        ptr = *reinterpret_cast<void**>(m_data.data());
+        _ACTS_ANY_DEBUG("Pre-destroy: " << ptr);
+      }
```
HadrienG2, line 264: I think you have too many copies of this logic around. How about just adding a `void* data()` method that returns `(void*)m_data.data()` in the stack-allocated case and `*(void**)m_data.data()` in the heap-allocated case?

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,427 @@
...
+    void* to = m_data.data();
+    const void* from = fromAny.m_data.data();
+    if (m_handler->typeSize > SIZE) {
+      // stored on heap: interpret buffer as pointer
+      to = *reinterpret_cast<void**>(m_data.data());
+      from = *reinterpret_cast<void* const*>(fromAny.m_data.data());
+    }
+
+    if (m_handler->copyConstruct == nullptr) {
+      // trivially copyable
```
HadrienG2, line 332: "trivially copy constructible"?

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,427 @@
...
+  operator bool() const { return m_handler != nullptr; }
+
+ private:
+  struct Handler {
+    void (*destroy)(void* ptr) = nullptr;
+    void (*moveConstruct)(void* from, void* to) = nullptr;
+    void (*move)(void* from, void* to) = nullptr;
+    void (*copyConstruct)(const void* from, void* to) = nullptr;
+    void (*copy)(const void* from, void* to) = nullptr;
+    std::size_t typeSize{0};
```
HadrienG2, line 220: You're only ever using this to test if you're stored on the stack or the heap, so you might as well use a boolean that says just that to reduce runtime comparisons and hopefully slightly improve performance.

Core/include/Acts/Utilities/Any.hpp
paulgessinger: There's some extra looping etc that I want to avoid if possible.

Core/include/Acts/Utilities/Any.hpp
paulgessinger: Can switch to an exception

Core/include/Acts/Utilities/Any.hpp
paulgessinger: Can do, but shouldn't make a difference, no?

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    return cache;
+  }
+
+  template <typename T, typename... Args>
+  explicit AnyBase(std::in_place_type_t<T> /*unused*/, Args&&... args) {
+    using U = std::decay_t<T>;
+    static_assert(std::is_same_v<T, std::decay_t<T>>,
+                  "Please pass the raw type, no const or ref");
```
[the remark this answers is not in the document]
paulgessinger: Ok, removed where not needed.

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,427 @@
...
+          "Construct local (this=" << this << ") at: " << (void*)m_data.data());
+    } else {
+      // too large, heap allocate
+      U* heap = new U(std::move(value));
+      _ACTS_ANY_DEBUG("Construct heap (this=" << this << ") at: " << heap);
+      _ACTS_ANY_DEBUG_BUFFER("-> buffer before", m_data);
+      *reinterpret_cast<U**>(m_data.data()) = heap;
+      _ACTS_ANY_DEBUG_BUFFER("-> buffer after", m_data);
+    }
+  }
```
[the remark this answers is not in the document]
paulgessinger: You might be right.

Core/include/Acts/Utilities/Any.hpp
paulgessinger: Is there a good alternative?

Core/include/Acts/Utilities/Any.hpp
paulgessinger: This is after assignment, so we assigned an empty to an empty. Doesn't have to be an error, I guess, just a noop.

Core/include/Acts/Utilities/Any.hpp
paulgessinger: Fair enough.

Core/include/Acts/Utilities/Any.hpp
paulgessinger: That only solves it half of the time, since sometimes i need to be able to assign it too. But I can try to centralize the casting anyway.

HadrienG2: Indeed, for assignment you'd need the by-reference equivalent.

Core/include/Acts/Utilities/Any.hpp
HadrienG2: Not in standard C/++ as far as I know. The C11 standard says...

> It is implementation-defined if over-aligned types are supported, and their support may be different in each kind of [storage duration](https://en.cppreference.com/w/c/language/storage_duration).

As a non-standard extension with a portable fallback, you could do...

```c++
static constexpr size_t alignment = 
#ifdef __x86_64__
    64  // Good enough for AVX-512
#else
    alignof(std::max_align_t)
#endif
;
```

Of course, this idea can be further refined by testing for presence of AVX and AVX-512:

```c++
static constexpr size_t alignment = 
#ifdef __AVX512F__
    64
#elif defined(__AVX__)
    32
#elif defined(__SSE__)
    16
#else
    alignof(std::max_align_t)
#endif
;
```

HadrienG2: Main major difference is losing the "`char*` is interpreted as a string by `std::cout`" footgun, which is nice when debug-printing around.

paulgessinger: Ok

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,427 @@
...
+
+    cache.m_handler = makeHandler<U>();
+    if constexpr (sizeof(U) <= SIZE) {
+      // construct into local buffer
+      /*U* ptr =*/new (cache.m_data.data()) U(std::forward<Args>(args)...);
+    } else {
+      // too large, heap allocate
+      U* heap = new U(std::forward<Args>(args)...);
+      *reinterpret_cast<U**>(cache.m_data.data()) = heap;
+    }
```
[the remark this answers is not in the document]
paulgessinger: I got rid of this static factory altogether.

Core/include/Acts/Utilities/Any.hpp
paulgessinger: Ok replaced.

Core/include/Acts/Utilities/Any.hpp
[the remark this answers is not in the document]
paulgessinger: Ok, replaced with a minimal constructor fowarding to the constructor with `std::in_place_type`.

Core/include/Acts/Utilities/Any.hpp
paulgessinger: Ok, changed to the fancy version from your comment.

paulgessinger:
Ok I think I addressed all comments. heaptrack is still unhappy, valgrind is.
I added optional intrusive tracking of heap allocations and deallocations, which also indicates there's no heap memory leak.

With this caveat, should we move forward with this?
If everyone is happy, I'll put in doxygen comments.

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,495 @@
...
+    if (m_handler == nullptr && other.m_handler == nullptr) {
+      // both are empty, noop
+      return;
+    }
+
+    _ACTS_ANY_VERBOSE(
+        "Copy construct (this=" << this << ") at: " << (void*)m_data.data());
+
+    m_handler = other.m_handler;
+    assert(m_handler != nullptr && "Handler is null");
```
HadrienG2, line 169: Probably not needed anymore with above check? (same consideration applies to most other asserts in public interface.

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,495 @@
...
+    } else {
+      // too large, heap allocate
+      U* heap = new U(std::forward<Args>(args)...);
+      _ACTS_ANY_TRACK_ALLOCATION(T, heap);
+      setDataPtr(heap);
+    }
+  }
+
+  AnyBase() {
+    m_data.fill(std::byte{0});
```
HadrienG2, line 120: Why?

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,495 @@
...
+      }
+      if constexpr (!std::is_trivially_copy_constructible_v<T> ||
+                    sizeof(T) > SIZE) {
+        h.copyConstruct = &copyConstructImpl<T>;
+      }
+      if constexpr (!std::is_trivially_copy_assignable_v<T> ||
+                    sizeof(T) > SIZE) {
+        h.copy = &copyImpl<T>;
+      }
+      h.heap = sizeof(T) > SIZE;
```
HadrienG2, line 285: Consider setting `h.heap` in the beginning and replacing the repetitive `sizeof(T) > SIZE` conditional with `h.heap`. Also, maybe `heapAllocated` would be a clearer name, but naming is subjective, so up to you.

HadrienG2: Also, you may be able to finish removing this recurring conditional from the rest of the code by adding one tiny utility function to replace it everywhere:

```c++
template <typename T>
static bool heapAllocated() {
    return sizeof(T) > SIZE;
}
```

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,495 @@
...
+    T* _to = static_cast<T*>(to);
+
+    (*_to) = std::move(*_from);
+  }
+
+  template <typename T>
+  static void* copyConstructImpl(const void* from, void* to) {
+    _ACTS_ANY_VERBOSE("copy const: " << from << " -> " << to);
+    assert(from != nullptr && "Source is null");
+    // assert(to != nullptr && "Target is null");
```
HadrienG2, line 446: Drop the commented code (also, that new design is quite clever :smile: ).

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,495 @@
...
+        std::lock_guard guard{_s_any_mutex};
+        auto it =
+            _s_any_allocations.find(std::pair{std::type_index(typeid(T)), obj});
+        if (it == _s_any_allocations.end()) {
+          throw std::runtime_error{
+              "Trying to deallocate heap address that we didn't allocate"};
+        }
+        _s_any_allocations.erase(it);
+      }
+#endif
```
HadrienG2, line 416: It is strange that your extracted `_ACTS_ANY_TRACK_ALLOCATION` above, but left this conditional code block inline, even though both are used only once now. Consider being consistent by either extracting both or neither (I'd be in favor of both, it would consolidate all the dirty ifdefs in a single place).

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,495 @@
...
+    void (*destroy)(void* ptr) = nullptr;
+    void (*moveConstruct)(void* from, void* to) = nullptr;
+    void (*move)(void* from, void* to) = nullptr;
+    void* (*copyConstruct)(const void* from, void* to) = nullptr;
+    void (*copy)(const void* from, void* to) = nullptr;
+    bool heap{false};
+  };
+
+  template <typename T>
+  static Handler* makeHandler() {
```
HadrienG2, line 262: Consider constifying this pointer's target, and the inner static as well while you're at it.

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -0,0 +1,495 @@
...
+    if (!_s_any_allocations.empty()) {
+      std::cout << "Not all allocations have been released" << std::endl;
+      for (const auto& [idx, addr] : _s_any_allocations) {
+        std::cout << "- " << idx.name() << ": " << addr << std::endl;
+      }
+      std::abort();
+    }
+  }
+};
+static _AnyAllocationReporter s_reporter;
```
HadrienG2, line 82: This destructor will not run if the application aborts due to an allocator failure, as is currently happening... so I'm not sure if this is the right solution, although it is certainly an elegant one.

HadrienG2 (changes requested):
I think this is now in a pretty good shape. Please consider this last round of suggestions, and then ping me back once the doxygen is done for final approval.

paulgessinger:
When I run heaptrack on a standalone executable, it reports this `free` error, even when I remove the `Any` part completely. I suspect more and more that this error doesn't have anything to do with the implementation here.

HadrienG2:
My previous advice to throw a debugger at this still stands :)

Core/include/Acts/Utilities/Any.hpp
paulgessinger: Probably not, I like to do too many asserts rather than too few.

Core/include/Acts/Utilities/Any.hpp
paulgessinger: I guess so it's a nullptr by default. Might not be necessary

Core/include/Acts/Utilities/Any.hpp
paulgessinger: To the wirst point, I can't use `h.heapAllocated` in the ifs because it's not constexpr.

Core/include/Acts/Utilities/Any.hpp
HadrienG2: Well, your null state is really indicated by `m_handler` being `nullptr`, and it has to be because with SBO, you need to check `m_handler` in order to correctly interpret the value of `m_data` (it could be a perfectly valid array of zero bytes after all). So I would argue that this is unnecessary.

Core/include/Acts/Utilities/Any.hpp
HadrienG2: You can work around that limitation by using an intermediary constexpr value:

```c++
static constexpr bool heapAllocated = sizeof(T) > SIZE;
h.heapAllocated = heapAllocated;
if constexpr (!std::is_trivially_destructible_v<T> || heapAllocated) {
    /* ...and so on ... */
```

HadrienG2: Then again, the current code that runs heapAllocated<T>() multiple times isn't that bad.

Core/include/Acts/Utilities/Any.hpp
paulgessinger: That's true, but I don't have a better idea how to add a end of job hook without using an exit handler. Do you?

Core/include/Acts/Utilities/Any.hpp
paulgessinger: It's only used once, that's why. But I moved it to a macro nonetheless.

Core/include/Acts/Utilities/Any.hpp
HadrienG2: I would borrow an idea from the Linux RCU synchronization primitive and have "checkpoints" throughout your test code that are invoked at strategic points of the test code where you assert that no allocation should be live.

paulgessinger: Hm, but then it's not automatic, and the idea was that this is transparent and global.

paulgessinger:
I ran with a debugger, the stack trace looks to me like the abort comes from inside heaptrack code. There's a non-zero chance that this is because of some memory corruption, but the fact that if I run heaptrack on any binary I get the same abort, and also valgrind is happy gives me some confidence that there is no memory corruption happening.

Core/include/Acts/Utilities/Any.hpp
HadrienG2: Note that you can keep the automatic report, it does no harm. But with a manual report, you get more targeted and abort-immune info.

HadrienG2: Looks good to me overall, why do you have both _ACTS_ANY_CHECK_ALLOCATIONS (in the Any header) and CHECK_ANY_ALLOCATIONS (in the tests) doing the same thing ?

paulgessinger: I just messed up 😅. Removes the ones from `Any.hpp` now.

HadrienG2 (approved):

Core/include/Acts/Utilities/Any.hpp
HadrienG2: Alright, let's merge this :)

PR #1696, opened 2022-11-25
feat: Add Geometry ID hook for the endcap strip of the ODD

Corentin-Allaire:
This add a geometry ID hook that use the extra GeoID field to separate the bottom and top row of the ODD endcap. This will be useful to properly implement the ODD digitisation.

This also add the extra id field to the json writing of the tracking geometry and to the << overwrite for the Geometry ID.

When writing the geometry to json the extra id is now properly present : 
```
            {
                "extra": 1,
                "layer": 2,
                "sensitive": 1,
                "value": {
                    "bounds": {
                        "type": "TrapezoidBounds",
                        "values": [
                            53.2,
                            58.6,
                            78.0
                        ]
                    },
                    "geo_id": 2017612770500935937,
                    "material": {
                        "binUtility": {
                            "binningdata": [
                                {
                                    "bins": 1,
                                    "max": 58.599998474121094,
                                    "min": -58.599998474121094,
                                    "option": "open",
                                    "type": "equidistant",
                                    "value": "binX"
                                },
                                {
                                    "bins": 1,
                                    "max": 78.0,
                                    "min": -78.0,
                                    "option": "open",
                                    "type": "equidistant",
                                    "value": "binY"
                                }
                            ]
                        },
                        "mapMaterial": false,
                        "mappingType": "Default",
                        "type": "proto"
                    },
                    "transform": {
                        "rotation": [
                            0.01999866669333311,
                            0.9998000066665776,
                            0.0,
                            0.9998000066665776,
                            -0.01999866669333311,
                            0.0,
                            0.0,
                            0.0,
                            -1.0
                        ],
                        "translation": [
                            820.0,
                            1.5308084989341916e-16,
                            -2974.5
                        ]
                    },
                    "type": "PlaneSurface"
                },
                "volume": 28
            },
```


asalzburger:
Oh cool, that's a great feature.

Core/src/Geometry/GeometryIdentifier.cpp
```diff
@@ -24,7 +24,7 @@ std::ostream& Acts::operator<<(std::ostream& os, Acts::GeometryIdentifier id) {
                                               id.sensitive(), id.extra()};
 
   bool writeSeparator = false;
-  for (auto i = 0u; i < 5u; ++i) {
+  for (auto i = 0u; i < 6u; ++i) {
```
andiwand, line 27: could we do something like `sizeof(levels)`?

might be worth to use `std::array` up there and enforce same size

Examples/Detectors/DD4hepDetector/src/GeoIDHooks.cpp
```diff
@@ -0,0 +1,30 @@
...
+                                        const Acts::Surface& surface) {
+  if (identifier.volume() == 28 || identifier.volume() == 30) {
+    Acts::Vector3 center = surface.center(Acts::GeometryContext());
+    double radius = sqrt(center[0] * center[0] + center[1] * center[1]);
+    if (radius < 850) {
+      identifier.setExtra(1);
+    } else {
+      identifier.setExtra(2);
+    }
+    std::cout << identifier << std::endl;
```
andiwand, line 24: kill cout?

Examples/Detectors/DD4hepDetector/src/GeoIDHooks.cpp
```diff
@@ -0,0 +1,30 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#include "ActsExamples/DD4hepDetector/GeoIDHooks.h"
+
+namespace det {
+namespace geoIDHook {
+
+Acts::GeometryIdentifier stripEndcapODD(Acts::GeometryIdentifier identifier,
+                                        const Acts::Surface& surface) {
+  if (identifier.volume() == 28 || identifier.volume() == 30) {
```
andiwand, line 16: would it be possible to decide on this based on dd4hep information?

same with the radius below

Examples/Detectors/DD4hepDetector/include/ActsExamples/DD4hepDetector/GeoIDHooks.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "Acts/Geometry/GeometryIdentifier.hpp"
+#include "Acts/Surfaces/Surface.hpp"
+
+/// Geometry identifiers hooks to be used with DD4Hep detectors 
+/// to had some extra identifier to sensitives surfaces.
+namespace det {
+namespace geoIDHook {
```
paulgessinger, line 9: I think we should expand the names here to `GeometryIdentifierHooks`, and then also for the namespaces.

Examples/Detectors/DD4hepDetector/src/GeoIDHooks.cpp
benjaminhuth: Hmm you mean e.g. by name? I'm not actually sure, and even if this is somehow structured in the DD4hep description I don't think we need to be overly generic here. This is a geometry specific problem, and I think a geometry specific solution is ok for that...

paulgessinger: I think theoretically, one would be able to unpack the dd4hep specific detector element from the surface instance.

andiwand: sure I would also not go too far with this but if the IDs ever change this might be a funny thing to debug

benjaminhuth: Hmm in any case I would write some kind of python test that checks that and that should also fail if the geometry IDs change...
But if its easy to do in a more robust way, then even better.

Corentin-Allaire: To be fair if the ID changes all the Digitisation+MaterialMap+... would break so this won't be the biggest issue. On the other hand if there is a more elegant solution I am all for it !

Examples/Detectors/DD4hepDetector/include/ActsExamples/DD4hepDetector/GeoIDHooks.h
Corentin-Allaire: Good idea.

Examples/Detectors/DD4hepDetector/include/ActsExamples/DD4hepDetector/GeometryIdentifierHooks.hpp
```diff
@@ -0,0 +1,25 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+#include "Acts/Geometry/GeometryIdentifier.hpp"
+#include "Acts/Surfaces/Surface.hpp"
+
+/// Geometry identifiers hooks to be used with DD4Hep detectors
+/// to had some extra identifier to sensitives surfaces.
+namespace det {
```
asalzburger, line 16: Is this a new namespace?

asalzburger: Should probably not be 'set' then, I suppose, if it is really specific to the ODD, I would even call it `ODD::` ...

Examples/Detectors/DD4hepDetector/src/GeometryIdentifierHooks.cpp
```diff
@@ -0,0 +1,29 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#include "ActsExamples/DD4hepDetector/GeometryIdentifierHooks.hpp"
+
+namespace det {
+namespace GeometryIdentifierHooks {
+
+Acts::GeometryIdentifier stripEndcapODD(Acts::GeometryIdentifier identifier,
+                                        const Acts::Surface& surface) {
+  if (identifier.volume() == 28 || identifier.volume() == 30) {
```
asalzburger, line 16: A bit dangerous if the volume numbering changes, can't you change that to a callable struct and then make it such that you can add these parameters at struct construction (including the radius below).

Examples/Detectors/DD4hepDetector/src/GeometryIdentifierHooks.cpp
```diff
@@ -0,0 +1,29 @@
...
+    Acts::Vector3 center = surface.center(Acts::GeometryContext());
+    double radius = sqrt(center[0] * center[0] + center[1] * center[1]);
+    if (radius < 850) {
```
asalzburger, line 19: Same here.

asalzburger (changes requested):

asalzburger:
Hi Corentin, I *think* you can make a generic version out of that, which can do more than just ODD:

You can make a struct that swallows parameters and volume ids, and make that callable. So one can even adapt this things w/o rewriting the code ... even if e.g. volume numbers change.

Examples/Detectors/DD4hepDetector/include/ActsExamples/DD4hepDetector/GeometryIdentifierHooks.hpp
Corentin-Allaire: It is not a new namespace, it is also used by DetUtils. I could change it but if I rewrite this to be a generic layerspliter, I guess it would be fine keeping it as is.

Examples/Detectors/DD4hepDetector/src/GeometryIdentifierHooks.cpp
Corentin-Allaire: I can try to write some thing more generic. The issue is how to make it work with the bindings but maybe I can transform this into a function that take the volumes and radius as parameters and that return an std::function. I will have a try

Examples/Python/src/Geometry.cpp
```diff
@@ -22,9 +22,21 @@ using namespace pybind11::literals;
 namespace Acts::Python {
 void addGeometry(Context& ctx) {
   auto m = ctx.get("main");
-
-  py::class_<Acts::Surface, std::shared_ptr<Acts::Surface>>(m, "Surfaces");
-
+  {
+    py::class_<Acts::Surface, std::shared_ptr<Acts::Surface>>(m, "Surfaces")
```
paulgessinger, line 26: I guess if we're already doing this, I'd add the extra level and make bindings for `GeometryIdentifier` and then return that from this the `geometryId` method also in python.

Corentin-Allaire:
So, as discuss today during the meeting, implementing a more generic implementation seems quite complicated. So even if the current implementation is not conceptually satisfying, we will stay with this. 
I realised that we are failing the example (the root file store the ID which have changed) I will update the hash then we can go forward.

Corentin-Allaire:
I am not sure the physmon fail for this PR does anybody have an idea ?

andiwand:
> I am not sure the physmon fail for this PR does anybody have an idea ?

looks like the hashes changed too so I guess there is also some coupling to the geo IDs in the physmon too?

paulgessinger:
If you look at the plots from physmon, it actually shows some large differences. I'm guessing that maybe the digitization or similar don't apply the same way as they did before and that changes the output?

Corentin-Allaire:
> If you look at the plots from physmon, it actually shows some large differences. I'm guessing that maybe the digitization or similar don't apply the same way as they did before and that changes the output?
> 
> E.g. the number of measurements drops quite a bit, so maybe we're literally missing hits?
> 
> image

Found the issue, the GeometryHierarchyMap ID mask where not accounting for the extra bit which caused the digitisation to ignore the surfaces with extra ID. I am fixing the issue

paulgessinger:
Physmon saves the day again! 😄

Corentin-Allaire:
Seems like we are now passing all the test ! If there no objection I will close the conversation tomorrow and we can then move forward.

Corentin-Allaire:
@asalzburger do you think you could have another look ? I think you would need to approve since you have requested changes.

PR #1697, opened 2022-11-28
refactor: Remove dependency on Boost.program_options from detectors

benjaminhuth:
This removes the dependencies of the Examples detectors from Boost.program_options. This is done by moving the class `IDetectorBase` to the `Run` folder and creating new classes called `GenericDetectorWithOptions` etc., that inherit from it.

The next step would be to make the binaries compilation optional, but I wanted to seperate these things from each other.


andiwand (approved):
nice separation!

it may be worth to rename the detector base class into something that also contains "options"

benjaminhuth:
I did this, but then I went back because keeping this name allows me to leave a lot of files untouched, and I thought thought I want to do the separation with minimal impact.
But of course could be changed as well...

andiwand (approved):
lets get this in

PR #1698, opened 2022-11-28
feat: Improve const correctness of MultiTrajectory

paulgessinger:
Now `const TrackStateProxy&` should behave like `ConstTrackStateProxy`.


Core/include/Acts/EventData/MultiTrajectory.hpp
```diff
@@ -55,6 +55,34 @@ namespace detail_lt {
 template <typename T, bool select>
 using ConstIf = std::conditional_t<select, const T, T>;
 
+/// Helper type to make a member pointers constness transitive.
+template <typename T>
+struct TransitiveConstPointer {
```
benjaminhuth, line 60: Maybe put this in `Core/Utilities`?

benjaminhuth, line 60: I would actually prefer to put this into a `class` with private member data, since the possibility do directly access the `ptr` kind of defeats its purpose...

Would maybe also be good to put this into `Utilities/Core`

benjaminhuth (changes requested):
Very nice, only the thing with the `class` would be good in my opinion

Core/include/Acts/EventData/MultiTrajectory.hpp
paulgessinger: I'll make it a class and make the member private, but I think I'd keep this in the detail namespace in this file for now.

benjaminhuth (approved):
Okay, let's go!

PR #1701, opened 2022-11-29
feat: MultiTrajectory track state ranges

paulgessinger:
Add the ability to iterate over track states in for loops.

Aside from doing

```cpp
auto function = [](auto ts){
  // ...
};
t.visitBackwards(index, function);
```
you can now do

```cpp
for(auto ts : t.trackStateRange(index) {
  // ...
}

### CODE REVIEW

File: Core/include/Acts/EventData/MultiTrajectory.hpp [hunk 1]
```diff
@@ -1112,6 +1172,23 @@ class MultiTrajectory {
...
+  auto trackStateRange(IndexType iendpoint) const {
+    return detail_lt::TrackStateRange{getTrackState(iendpoint)};
+  }
+
+  /// Range for the track states from @p iendpoint to the trajectory start
+  /// @param iendpoint Trajectory entry point to start from
+  /// @return Iterator pair to iterate over
+  /// @note Mutable version
+  template <bool RO = ReadOnly, typename = std::enable_if_t<!RO>>
+  auto trackStateRange(IndexType iendpoint) {
```
benjaminhuth, line 1188: Is it possible to get a const `TrackStateRange` from a mutable trajectory? I think there is always the mutable overload called.
I would actually prefer two different function names `trackStateRange` and `constTrackStateRange`.

benjaminhuth (changes requested):
Very nice feature, I think that will makes the trajectory related code more readable and flexible. Only the thing with the const range for mutable trajectories I'd like to clarify.

PS: Would it be maybe even be worth deprecating the `applyBackwars`/`visitBackwards` functions or do I overlook a use-case? However, not really relevant for this PR...

Core/include/Acts/EventData/MultiTrajectory.hpp
paulgessinger: Well I was thinking in that case you would write something like

```cpp
for(const auto ts : mtj.trackStateRange()) {}
```

Do you think that's not enough?

paulgessinger:
`applyBackwards` and `visitBackwards` are technically a bit cheaper than the iterator range I think, so I'd probably keep them?

paulgessinger:
Ok, so the clang-tidy warning here is a false-positive. It thinks that the proxies are only used as const refs, but semantically that is incorrect.

andiwand:
> Ok, so the clang-tidy warning here is a false-positive. It thinks that the proxies are only used as const refs, but semantically that is incorrect.
> 
> Thoughts @andiwand?

but you only use them in a const way right? I think clang tidy just wants to get rid of the copy of the proxy

Core/include/Acts/EventData/MultiTrajectory.hpp
benjaminhuth: For the for loop, this works, you're right. But I also thought that users might want to write or use algorithms in terms of ranges (e.g., C++20 ranges algorithms...). 
Therefore I would found it nice to have this options...

paulgessinger:
CI is green can we go ahead?

Core/include/Acts/EventData/MultiTrajectory.hpp
```diff
@@ -980,6 +985,12 @@ class TrackStateProxy {
     return component<TrackStateType, hashString("typeFlags")>();
   }
 
+  template <bool RO = ReadOnly, typename = std::enable_if_t<!RO>>
+  MultiTrajectory<Trajectory>& trajectory() {
+    return *m_traj;
+  }
+  const MultiTrajectory<Trajectory>& trajectory() const { return *m_traj; }
```
andiwand, line 992:
```suggestion

  const MultiTrajectory<Trajectory>& trajectory() const { return *m_traj; }
```

andiwand (approved):
looks good to me

andiwand (approved):

benjaminhuth (approved):

PR #1705, opened 2022-12-02
refactor!: CKF interface change to once per seed

paulgessinger:
Previously, the CKF accepts a vector of seeds and produces a vector of outputs. In preparation for the Track container, it's preferrable to have the loop over seeds outside this function, i.e. the CKF is invoked once per seed, and the vector accumulation occurs outside of it


paulgessinger:
This is in preparation for #1729. Should we merge this ahead of that PR or skip this PR entirely and go with a single change? @andiwand @timadye @asalzburger ?

timadye (approved):
LGTM

PR #1706, opened 2022-12-02
fix: Change theta and eta calls on only 3D Vectors

pbutti:
The theta and eta calls could be done on 4D vectors. While they are well defined on a LorentzVector, for example, this was not the intended behaviour of the function, which supposedly should run only on 3D vectors. In particular eta on a 4-momentum was returning the wrong results. Changed the checks on rows to be exactly 3. In the future the eta/theta could be properly defined on 4d vectors as well by choosing a convention for the indices of the 4D vector


paulgessinger (approved):
LGTM!

benjaminhuth:
Ah you might need to add @pbutti to your CI-bridge-allow-list @paulgessinger ?

PR #1709, opened 2022-12-05
fix: fp abs compare

andiwand:
fixes potential over/underflow cases

### DIFF
--- a/Core/include/Acts/Vertexing/IterativeVertexFinder.ipp
+++ b/Core/include/Acts/Vertexing/IterativeVertexFinder.ipp
@@ -11,6 +11,8 @@ auto Acts::IterativeVertexFinder<vfitter_t, sfinder_t>::find(
     const std::vector<const InputTrack_t*>& trackVector,
     const VertexingOptions<InputTrack_t>& vertexingOptions, State& state) const
     -> Result<std::vector<Vertex<InputTrack_t>>> {
+  std::cout << "tracks " << trackVector.size() << std::endl;
+
   // Original tracks
   const std::vector<const InputTrack_t*>& origTracks = trackVector;
   // Tracks for seeding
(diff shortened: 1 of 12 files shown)


Core/include/Acts/Vertexing/IterativeVertexFinder.ipp
```diff
@@ -11,6 +11,8 @@ auto Acts::IterativeVertexFinder<vfitter_t, sfinder_t>::find(
     const std::vector<const InputTrack_t*>& trackVector,
     const VertexingOptions<InputTrack_t>& vertexingOptions, State& state) const
     -> Result<std::vector<Vertex<InputTrack_t>>> {
+  std::cout << "tracks " << trackVector.size() << std::endl;
```
paulgessinger, line 14: Extra cout?

Core/include/Acts/Vertexing/IterativeVertexFinder.ipp
```diff
@@ -31,6 +33,9 @@ auto Acts::IterativeVertexFinder<vfitter_t, sfinder_t>::find(
     // the seed vertexCollection
     Vertex<InputTrack_t>& seedVertex = *seedRes;
 
+    std::cout << "vertex seed " << seedVertex.fullPosition().transpose()
+              << std::endl;
+
```
paulgessinger, line 38: Extra cout?

Core/src/Geometry/CylinderVolumeHelper.cpp
```diff
@@ -499,7 +499,7 @@ bool Acts::CylinderVolumeHelper::estimateAndCheckDimension(
   double zEstFromLayerEnv = 0.5 * ((layerZmax) + (layerZmin));
   double halflengthFromLayer = 0.5 * std::abs((layerZmax) - (layerZmin));
 
-  bool concentric = (zEstFromLayerEnv * zEstFromLayerEnv < 0.001);
+  bool concentric = std::abs(zEstFromLayerEnv) < 0.031622777;
```
paulgessinger, line 502: Where's this number from?

andiwand: this is `sqrt(0.001)`. I could change it to that actually. what do you think?

paulgessinger (approved):

PR #1715, opened 2022-12-06
fix: Make `Range1D` a proper half-open interval

stephenswat:
Previously, the `Range1D` class checked for degeneracy as if it were a closed interval, but it is designed to be half-open. This commit changes the logic to represent this.


paulgessinger:
Unit test(s) seem to check for this (ostensibly) incorrect check. Can you fix it/them?

stephenswat:
This should now be ready to go.

andiwand (approved):

PR #1721, opened 2022-12-08
feat: Add python bindings to HoughTransform algorithm

tboldagh:
This PR adds python binding to the the HoughTransform algorithm in examples.
So far, it does not add python helper to set it up in test job. However I tested these bindings in the itk_full_chain job up to the level that I could see this algorithm digests some data.

Tagging @jahreda for a review.

### DIFF
--- a/Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp
+++ b/Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/HoughTransformSeeder.hpp
@@ -94,7 +94,8 @@ using LayerIDFinder = Acts::Delegate<ResultUnsigned(
     double)>;  // (double r) this function will map the r of a measurement to a
                // layer.
 using SliceTester = Acts::Delegate<ResultBool(
-    double, unsigned, int)>;  // (double z,unsigned layer, int slice)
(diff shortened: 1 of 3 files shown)


jahreda:
It looks good to me; we could check that we still find track candidates? Have you done that, Tomasz? (Just from the log or monitoring output?) Or do you need me to check? We should indeed make a Hough test job at some point!

timadye:
Hi @tboldagh,

That sounds like a very good idea. I think it still a good idea to have a main `addSeeding` function that calls the others based on the `seedingAlgorithm` enum, both for compatibility and for ease of comparison. I guess the `seedfinderConfig` will need to be different, but I think it's OK to have a merged `namedtuple` or separate arguments depending on how close they are.

Thanks,
Tim.

tboldagh:
Hi, after playing with the refactor in reconstruction.py I think it should be separate PR. 
The changes that are required to reduce the addSeeding to a digestible size are  quite extensive. 
I propose to go ahead with this PR and then refactor & introduce HT as seeding option in another PR.

tboldagh:
Sorry, mistakenly clicked close PR while I thought I close comment.

paulgessinger (approved):

paulgessinger:
Don't see any issues with this. It's not used right now, so would be good to add this to the pytest suite so it's covered.

Issue #1720, opened 2022-12-08
Rare navigator failure causing overflow FPE


noemina:
I am facing a rare navigation failure while investigating overflow FPEs. It manifests as follows:

I am propagating, using the `Acts::StraightLineStepper` in the ITk geometry within athena, in order to map the material on the tracking geometry layers. 
In rare case, I end up hitting a boundary layer too close to the intersection with another boundary (see attached picture). No boundary surfaces are therefore returned in [compatibleBoundaries](https://github.com/acts-project/acts/blob/85a67758f79f3b01aed4025592ab6a5e5ffbd323/Core/src/Geometry/TrackingVolume.cpp#L462) as the path length to the intersection to the next "correct" boundary to hit is [below the navigator (and stepper) tolerance](https://github.com/acts-project/acts/blob/3898a7fb639b841dd44fa4ee9a28d1fd88db36ff/Core/include/Acts/Utilities/Intersection.hpp#L199). 
Boundary surfaces to hit are checked twice, and after the second negative outcome the navigation proceed to target with `Target stepSize (path limit) updated to ( +∞,  +∞,  +∞,  +∞)` causing the overflow FPE. Beside the detected FPE, the use of +∞ as `stepSize` makes the next step being directly outside the geometry...

Reducing the tolerance when these case are detected, e.g. when [findBoundaries](https://github.com/acts-project/acts/blob/232e43dfa66f19823cb8548f876b33faa279791d/Core/include/Acts/Propagator/Navigator.hpp#L929) is called twice, makes the issue being more rare without solving it.

One possibility would be to force `compatibleBoundaries` to return the closest boundary layer along direction when needed. This is based on the fair (imo) assumption that if no layer or surface is hit, you want always to leave the volume through a boundary.

Are there situations where this assumption is wrong? 

@asalzburger and @paulgessinger are added to this issue to start the discussion.

![image](https://user-images.githubusercontent.com/58694235/206430179-18d1797f-b6eb-4f7b-acd9-f6ee558b54cd.png)

PR #1722, opened 2022-12-08
fix: Fixing navigation failures on volume boundaries edges

noemina:
Implements a way to hit boundaries when this is the only possible option. Closes https://github.com/acts-project/acts/issues/1720


paulgessinger:
image

👎

noemina:
> image
> 
> -1

Now fixed!

Core/src/Geometry/TrackingVolume.cpp
```diff
@@ -484,6 +484,27 @@ Acts::TrackingVolume::compatibleBoundaries(
       return BoundaryIntersection();
     }
 
+    if (options.forceIntersectBoundaries and
+        sIntersection.intersection.pathLength * options.navDir > 0) {
```
andiwand, line 488: does this do the same as `std::abs(sIntersection.intersection.pathLength) >0`? in that case we could also check for `sIntersection.intersection.pathLength != 0` if you think that is preferable

andiwand (approved):

Core/src/Geometry/TrackingVolume.cpp
noemina: This is to check that the boundary surface you want to intersect is along the direction, hence `sIntersection.intersection.pathLength` is required to have the same sign of `options.navDir`.

asalzburger (approved):

asalzburger:
If this is actually needed, we might consider the default to be set to `true` ?

noemina:
> If this is actually needed, we might consider the default to be set to `true` ?

@asalzburger , this can be set to `true` only if the previous checks fail.

PR #1723, opened 2022-12-08
feat: align amvf config with athena

andiwand:
config take from https://gitlab.cern.ch/atlas/athena/-/blob/master/Tracking/Acts/ActsTrkTools/ActsTrkPriVtxFinderTool/src/AdaptiveMultiPriVtxFinderTool.h


andiwand:
I guess this is bad because the unit tests couple to the default values. might be better to overwrite the options in the examples?

Core/include/Acts/Utilities/AnnealingUtility.hpp
```diff
@@ -30,8 +30,7 @@ class AnnealingUtility {
   /// @brief The configuration struct
   struct Config {
     // Config constructor with default temperature list: {64.,16.,4.,2.,1.5,1.}
```
AJPfleger, line 32: Should we update the comment too?
```suggestion
    // Config constructor with default temperature list: {1.}
```

AJPfleger (commented):
lgtm. I also rechecked the values (y)

andiwand:
@AJPfleger could you take another look?

PR #1724, opened 2022-12-08
refactor!: Add ability for cloning loggers with new name / level

paulgessinger:
We have a lot of defaulted logger instances to `getDefaultLogger("name", INFO)` in the code, where we never pass through loggers from top components. This means in e.g. Athena it's impossible to increase verbosity for these components.

This PR adds new facilities to the logger print and filter policies:

Print policies require:
```cpp
  virtual const std::string& name() const = 0;
  virtual std::unique_ptr<OutputPrintPolicy> clone(const std::string& name) const = 0;
```
and filter policies:

```cpp
virtual Level level() const = 0;
virtual std::unique_ptr<OutputFilterPolicy> clone(Level level) const = 0;
```

The output decorator mechanism is preferred by defaulting the behavior of `name()` to ask the wrappee for it. Our `NamedOutputDecorator` will terminate this chain and return it's name. The `DefaultPrintPolicy` which prints to `std::cout` should never get this call and will throw an exception.

With this in place, `Logger` now gets a `clone` method, which accepts optionally a new name and a new level, with some additional convenience helpers. This allows creating subloggers easily like:

```cpp
logger(); // main logger from e.g. the algorithm
auto subLogger = logger().clone("newname"); // inherits log level from parent logger.
```

BREAKING CHANGE: This change breaks custom logger print and filter policies due to changed interfaces. If you use `getDefaultLogger` you will not be affected.


andiwand (approved):
looks good!

paulgessinger:
Added some fixes, can you check and reapprove @andiwand?

andiwand (approved):
lets get this in

PR #1728, opened 2022-12-09
chore: Script to merge hash txt files

paulgessinger:
Small helper script to merge hash text files


CI/merge_hashes.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python3
```
andiwand, line 1: will this be used by the CI?

andiwand (approved):

CI/merge_hashes.py
paulgessinger: Not really, it's just convenient for updating I think, that's why I wanted to ship it with the repo. The hashes are FOR the CI though, so that's why it's there. Do you have another place in mind?

andiwand: no I think that makes perfect sense

AJPfleger:
@paulgessinger I think that script is still untested. Do you know if it still works or is used?

andiwand:
> @paulgessinger I think that script is still untested. Do you know if it still works or is used?

I use this quite often actually and it works fine 👍

PR #1730, opened 2022-12-09
fix: MultiTrajectory specialize DynamicColumn for bool

paulgessinger:
As vector is specialized on bool, which prevents getting references to the underlying bool mechanisms, we have to specialize `DynamiColumn` to break this specialization of vector.

I'm also rolling in another small const-correctness fix to dynamic columns into this PR.


paulgessinger:
There are no references to the `bool`s in `std::vector<bool>`.

benjaminhuth (approved):
Looks good!

PR #1732, opened 2022-12-12
feat: Optionally decorate geometry dumps to json.

goetzgaycken:
Added the possibility to pass a decorator to the json converter to add extra attributes to the json dump of certain objects. Decorator interfaces are defined for adding attributes to surface and volume or their material dumps.

One use case is to extend the json dumps of the ATLAS tracking geometry with the ATLAS detector element IDs.

The PR changes interfaces in Plugins/Json.


Plugins/Json/include/Acts/Plugins/Json/ITrackingGeometryJsonDecorator.hpp
```diff
@@ -0,0 +1,40 @@
+// This file is part of the Acts project.
+//
+// Copyright (C) 2017-2021 CERN for the benefit of the Acts project
```
niermann999, line 3: Very pedantic comment, but in the new files, let's keep the copyright up to date

Plugins/Json/include/Acts/Plugins/Json/GeometryHierarchyMapJsonConverter.hpp
```diff
@@ -114,15 +142,19 @@ nlohmann::json GeometryHierarchyMapJsonConverter<value_t>::toJson(
   nlohmann::json entries = nlohmann::json::array();
   for (std::size_t i = 0; i < container.size(); ++i) {
     auto entry = encodeIdentifier(container.idAt(i));
-    entry["value"] = nlohmann::json(container.valueAt(i));
+    auto value_json = nlohmann::json(container.valueAt(i));
+    if constexpr (not std::is_same<decorator_t, void>::value) {
```
niermann999, line 146: Just an idea, but if you want to get rid of the ```if constexpr```, then it might be possible to add an overload to ```decorateJson``` for ```decorator_t = void``` that does nothing? Or would that break the ```missing_specialization``` pattern?

Plugins/Json/include/Acts/Plugins/Json/GeometryHierarchyMapJsonConverter.hpp
```diff
@@ -102,9 +104,35 @@ class GeometryHierarchyMapJsonConverter {
...
+
+// methods to adapt type decorations for the given decorator
+template <class T, class decorator_t>
+void decorateJson([[maybe_unused]] const decorator_t* decorator, [[maybe_unused]] const T& src,
+                  [[maybe_unused]] nlohmann::json& dest) {
+  // this needs to be specialised
+  static_assert(
+      missing_specialization<T, decorator_t>::value,
+      "Explicit specialization needed for each decorator_t and src T");
+  src.compilationFailure();
```
niermann999, line 122: Just out of curiosity: What does this line do and why is it only present in this overload and not the other?

Plugins/Json/src/MaterialMapJsonConverter.cpp
```diff
@@ -28,6 +28,44 @@
...
 #include <map>
 
+namespace Acts {
+// specialisations of decoration helper function
+// to pick correct objects from the contaier object
+template <>
+inline void decorateJson<Acts::SurfaceAndMaterialWithContext>(
+    const ITrackingGeometryJsonDecorator* decorator,
+    const Acts::SurfaceAndMaterialWithContext& src, nlohmann::json& dest) {
+  if (decorator != nullptr && std::get<0>(src)) {
```
niermann999, line 38: Despite the line below, I don't think it's immediately obvious that ```std::get<0>(src)``` returns a pointer, so for readability I think it would help if it is explicitly checked against ```nullptr```, too

niermann999 (commented):
Looks good in general

Plugins/Json/include/Acts/Plugins/Json/GeometryHierarchyMapJsonConverter.hpp
```diff
@@ -51,7 +52,8 @@ class GeometryHierarchyMapJsonConverter {
   ///
   /// @param container Geometry hierarchy map that should be encoded
   /// @return Encoded Json object
-  nlohmann::json toJson(const Container& container) const;
+  nlohmann::json toJson(const Container& container,
+                        const decorator_t* decorator) const;
```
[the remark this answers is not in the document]
goetzgaycken: The reason behind the omission of a default argument is that in this way there are compiler errors if the decorator is not passed on.

Plugins/Json/include/Acts/Plugins/Json/ITrackingGeometryJsonDecorator.hpp
goetzgaycken: fixed. Thanks !

Plugins/Json/include/Acts/Plugins/Json/GeometryHierarchyMapJsonConverter.hpp
goetzgaycken: I do not really have a strong opinion on this. 
In both cases the compiler is needed to optimise away the code which does nothing. I thought that in this way it is more explicit when decorations are applied and when not. A e.g. "do_nothing" decorator still could do something. So its implementation would have to be checked. On the other hand a "void" decorator can absolutely not do anything.

Plugins/Json/include/Acts/Plugins/Json/GeometryHierarchyMapJsonConverter.hpp
goetzgaycken: In principle it could be reduced to static_assert( false, "...error message..."), but then gcc and clang do not mention the template parameters in the error message. So, this construct is there to produce hopefully understandable error messages during compilation, if there is no specialization of decorateJson for the given template parameters T and src.

Plugins/Json/src/MaterialMapJsonConverter.cpp
goetzgaycken: done. Thanks!

Plugins/Json/include/Acts/Plugins/Json/GeometryHierarchyMapJsonConverter.hpp
```diff
@@ -102,9 +104,35 @@ class GeometryHierarchyMapJsonConverter {
 
 // implementations
 
-template <typename value_t>
-nlohmann::json GeometryHierarchyMapJsonConverter<value_t>::toJson(
-    const Container& container) const {
+// auxiliary struct to indicate a missing specialization of a template which
+// requires specialisation
+template <typename T, class decorator_t>
+struct missing_specialization {
```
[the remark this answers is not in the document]
goetzgaycken: Makes sense to use standard constructs. I tend to reimplement simple constructs, since it requires less knowledge when reading the code, but I guess it leaves more room for inconsistent or confusing behaviour. Anyway done.

goetzgaycken:
changes from #8352a7a to #d318bc2 :
- rebase to main of 2023-01-17 8:05
- explicit nullptr comparison for get<0>(src)
- use std::false_type instead of "reinventing" the construct
- updated copyright message of changed files to -2023

goetzgaycken:
Changes from  d318bc2 to e3ca98d 
-  Fix compiler warning concerning potentially unused argument (if constrexpr(...) might disable some code) 
-  Fix doxygen warning about undocumented parameter

Plugins/Json/include/Acts/Plugins/Json/GeometryHierarchyMapJsonConverter.hpp
niermann999: Thanks for explaining! Should ```src.compilationFailure();``` not also be added in l. 131 ?

Plugins/Json/include/Acts/Plugins/Json/GeometryHierarchyMapJsonConverter.hpp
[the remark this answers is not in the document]
niermann999: Ok, I see

Plugins/Json/include/Acts/Plugins/Json/GeometryHierarchyMapJsonConverter.hpp
niermann999: Absolutely fine with me, I leave it to you

goetzgaycken:
changes from  e3ca98d to 1bc9e7a:
- implement clang-format differences

Plugins/Json/include/Acts/Plugins/Json/GeometryHierarchyMapJsonConverter.hpp
goetzgaycken: Sorry, I did not read Your first comment correctly. This was the initial idea to provoke a compilation error. This is a remnant I should delete. Thanks !

goetzgaycken:
changes from  1bc9e7a to c007e1d :

goetzgaycken:
changes from  1bc9e7a to c007e1d :
- removed remnant from an early idea.

Plugins/Json/include/Acts/Plugins/Json/ITrackingGeometryJsonDecorator.hpp
```diff
@@ -0,0 +1,40 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+#include "Acts/Plugins/Json/ActsJson.hpp"
+
+namespace Acts {
+class TrackingVolume;
+class Surface;
+class ISurfaceMaterial;
```
niermann999, line 16: Are these forward declarations ```ISurfaceMaterial``` and ```IVolumeMaterial``` leftovers? (And vice versa in ```IVolumeMaterialJsonDecorator```)

Plugins/Json/include/Acts/Plugins/Json/GeometryHierarchyMapJsonConverter.hpp
niermann999: Oh, sorry, I should have been more clear on what I meant :)

goetzgaycken: removed remnant.

goetzgaycken:
changes from  4b72276 to b2bcd7a :
  - removed unused forward declarations.

goetzgaycken:
concerning the test coverage:
- there are some new lines which cannot be tested by a unit test because they will provoke compilation failures. Not sure how the lines are counted this would be something between 2 and 18 lines.
- then there are some input adapters (between 12 and 34 lines), which could be tested  by writing a unit test for the tracking geometry to json conversion. The latter would however require a dummy tracking geometry. The resulting "unit" test would be more an integration test than a unit test.
- alternatively I could just call the adapters, but that would not test much.

paulgessinger:
Seems to me like the coverage decrease is somehow still from the recent change to KD-tree, maybe we should override the coverage here.

goetzgaycken:
> Seems to me like the coverage decrease is somehow still from the recent change to KD-tree, maybe we should override the coverage here.

can I ran the coverage analysis locally ?

niermann999 (approved):

paulgessinger:
> > Seems to me like the coverage decrease is somehow still from the recent change to KD-tree, maybe we should override the coverage here.
> 
> can I ran the coverage analysis locally ?

This full analysis suite run by codecov, no. You can run the coverage itself locally (I think `cmake --build --target coverage`) which gives you coverage output that can be processed and visualized separately. I haven't run this locally in quite a while either.

PR #1733, opened 2022-12-12
feat: Add BoundingBox to DetectorVolume

Matthewharri:
This PR adds in bounding boxes for the experimental DetectorVolume class. The way it is set up is such that when the portals for the volumes are created, there will also be a BoundingBox created for each volume, and will be saved to that volume so it will live for the lifetime of the volumes.


Core/include/Acts/Geometry/DetectorVolume.hpp
```diff
@@ -310,6 +315,8 @@ class DetectorVolume : public std::enable_shared_from_this<DetectorVolume> {
   /// Const access to the detector
   const Detector* detector() const;
 
+  const std::shared_ptr<BoundingBox> getBoundingBox() const;
```
paulgessinger, line 318: Do you need to expose this as a `shared_ptr`? In this form, the `const` on the left also doesn't do anything.
Would the following work?
```suggestion
  const BoundingBox& getBoundingBox() const;
```

Core/include/Acts/Geometry/DetectorVolume.hpp
```diff
@@ -40,6 +41,10 @@ using PortalGenerator = Delegate<std::vector<std::shared_ptr<Portal>>(
     const Transform3&, const VolumeBounds&,
     const std::shared_ptr<DetectorVolume>&)>;
 
+using BoundingBox =
```
paulgessinger, line 44: Should this typedef be scoped to the `DetectorVolume` class?

Core/include/Acts/Geometry/DetectorVolume.hpp
```diff
@@ -328,6 +335,14 @@ class DetectorVolume : public std::enable_shared_from_this<DetectorVolume> {
   /// @return a boolean indicating if the objects are properly contained
   bool checkContainment(const GeometryContext& gctx, size_t nseg = 1) const;
 
+  /// build the bounding box
+  ///
+  void boundingBox(const GeometryContext& gctx);
+
+  /// Retrieve bounding box
+  ///
+  void setBoundingBox(const GeometryContext& gctx);
```
paulgessinger, line 344: I don't understand the difference between these two functions. One just calls the other? Can we get rid of one of them? Also I'd prefer if the name was more concrete, i.e. something like `createBoundingBox` or so.

Core/include/Acts/Geometry/DetectorVolume.hpp
```diff
@@ -346,6 +361,9 @@ class DetectorVolume : public std::enable_shared_from_this<DetectorVolume> {
   /// Volume store (internal/external)
   ObjectStore<std::shared_ptr<DetectorVolume>> m_volumes;
 
+  /// BoundingBox
+  std::shared_ptr<BoundingBox> m_boundingBox;
```
paulgessinger, line 365: Maybe 
```suggestion
  std::shared_ptr<const BoundingBox> m_boundingBox;
```
if we don't intend to modify it after creation.

Core/src/Geometry/DetectorVolume.cpp
```diff
@@ -13,11 +13,16 @@
...
 #include "Acts/Surfaces/Surface.hpp"
+#include "Acts/Utilities/BoundingBox.hpp"
 #include "Acts/Utilities/Enumerate.hpp"
 #include "Acts/Utilities/Helpers.hpp"
 
 #include <cassert>
 
+using BoundingBox =
+    Acts::AxisAlignedBoundingBox<Acts::Experimental::DetectorVolume,
+                                 Acts::ActsScalar, 3>;
```
paulgessinger, line 24: I don't think you need to repeat the typedef here.

paulgessinger:
The docs CI job should go green once the branch is updated, but please take a look at the clang-tidy warnings.

Core/include/Acts/Geometry/DetectorVolume.hpp
Matthewharri: Done

Core/include/Acts/Geometry/DetectorVolume.hpp
Matthewharri: Yep! Done.

Core/include/Acts/Geometry/DetectorVolume.hpp
Matthewharri: Ah, yeah no difference. I got rid of setBoundingBox and changed the name of boundingBox to createBoundingBox.

Core/include/Acts/Geometry/DetectorVolume.hpp
Matthewharri: Done.

Core/src/Geometry/DetectorVolume.cpp
Matthewharri: Done.

paulgessinger (approved):

Core/include/Acts/Geometry/DetectorVolume.hpp
```diff
@@ -346,6 +357,9 @@ class DetectorVolume : public std::enable_shared_from_this<DetectorVolume> {
   /// Volume store (internal/external)
   ObjectStore<std::shared_ptr<DetectorVolume>> m_volumes;
 
+  /// BoundingBox
+  std::shared_ptr<const BoundingBox> m_boundingBox;
```
EdwardMoyse, line 361: Just wondering why this is a `shared_ptr`, given it is returned as a reference in `getBoundingBox`? Obviously it's not a huge overhead, so this is mostly curiosity.

cc @paulgessinger

paulgessinger: It needs to be owned somewhere, and we try to avoid owning by raw pointer. I suppose this could be a member variable if we're sure it will always have one.

Matthewharri: Should I make a change or am I good to mark this as resolved?

EdwardMoyse: @paulgessinger I definitely meant member variable.

If we don't make this change, what will be the behaviour of `getBoundaryBox()`?
```cpp
const Acts::Experimental::DetectorVolume::BoundingBox&
Acts::Experimental::DetectorVolume::getBoundingBox() const {
  return *m_boundingBox;
}
```
Isn't this returning a reference to a null pointer? 

At least, when I tried this locally:
```cpp
struct Base
{
    unsigned int a;
}

    std::shared_ptr<Base> p;
    Base& test = *p;
    std::cout<<"p.get() "<<p.get()<<std::endl;
```

I get:
```
p.get() 0
```

paulgessinger: If the `shared_ptr` is empty you get a `nullptr` for sure. I suppose the idea with `shared_ptr` would be to ensure it's non-null by calling the constructor function from all the constructors. 

I think I agree that just making it a member variable and making sure to initialize it would probably be preferable.

I believe you can do something like

```cpp
class DetectorVolume {
  static BoundingBox createBoundingBox(Args...);
  BoundingBox m_bb;

  DetectorVolum() : m_bb{createBoundingBox(...)} {}
};
```

if you want to centralize the code to construct the BB. I believe here the function would have to be static, i.e. you'd have to pass in the required info by hand, because `this` is not available yet (double check me on that).

Matthewharri: Hi @paulgessinger, maybe I'm just not understanding but I'm not entirely sure I can build a BB at construction because the portals aren't yet defined which are needed in making the BB.

asalzburger: We discussed this, and I think @Matthewharri is right here.

PR #1737, opened 2022-12-12
feat: root event diff script

andiwand:
small python script to check in which event a diff in a root file occurs 

can be invoked like the following

```
python CI/root_event_diff_np.py tracksummary_gsf_a.root:tracksummary tracksummary_gsf_b.root:tracksummary
```


andiwand:
@paulgessinger cc

paulgessinger (approved):

benjaminhuth (approved):

AJPfleger:
@andiwand Should we integrate the script in the CI or move it to scripts? does it still work?

andiwand:
> @andiwand Should we integrate the script in the CI or move it to scripts? does it still work?

I used it the other day and it seems to work fine. But I don't have a strong opinion where it should live

AJPfleger:
I guess, we can then leave it her as well for now :)

PR #1739, opened 2022-12-13
test: Only run pythia8 tests if enabled

paulgessinger:


andiwand (approved):

paulgessinger:
This still had a bug where it would always skip the pythia tests. Let's check if I managed to fix that now before merging.

PR #1740, opened 2022-12-13
refactor: SourceLink uses custom Acts::Any

paulgessinger:
This should be a transparent change in terms of the interface.

### DIFF
--- a/Core/include/Acts/EventData/SourceLink.hpp
+++ b/Core/include/Acts/EventData/SourceLink.hpp
@@ -9,7 +9,6 @@
 #pragma once
 
 #include "Acts/Geometry/GeometryIdentifier.hpp"
-#include "Acts/MagneticField/detail/SmallObjectCache.hpp"
 #include "Acts/Utilities/Any.hpp"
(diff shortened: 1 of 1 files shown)


Core/include/Acts/EventData/SourceLink.hpp
```diff
@@ -26,7 +25,7 @@ using geometry_id_t = decltype(std::declval<T>().geometryId());
 }  // namespace detail_sl
 
 class SourceLink final {
-  using any_type = std::any;
+  using any_type = AnyBase<16>;
```
andiwand, line 28: do you think it makes sense to typedef this to `ActsAny` or something like that? so we use a single byte width across the codebase

andiwand (approved):

Core/include/Acts/EventData/SourceLink.hpp
paulgessinger: That's already the case: https://github.com/acts-project/acts/blob/main/Core/include/Acts/Utilities/Any.hpp#L499

```cpp
using Any = AnyBase<sizeof(void*)>;
```

PR #1741, opened 2022-12-13
feat: track 2d/z and track 3d signed lifetimes

pbutti:
This PR is to add couple of methods from the Athena TrackToVertexIPEstimator. In particular the 3d, 2d and Z signed lifetimes with respect to a given direction, i.e. a jet direction, are provided. 
The computation follows the methods largely used by atlas here:
https://acode-browser1.usatlas.bnl.gov/lxr/source/athena/Tracking/TrkVertexFitter/TrkVertexFitterUtils/src/TrackToVertexIPEstimator.cxx#0230

The interface is kept as simple as possible, without using the intermediate VertexFinder::State.


pbutti:
I will fix the clang-format and the docs. Marked as WIP

Core/include/Acts/Vertexing/ImpactPointEstimator.hpp
```diff
@@ -143,6 +143,39 @@ class ImpactPointEstimator {
...
+  /// @brief Estimates the sign of the 3D lifetime of a given track
+  /// w.r.t. a vertex and a direction (e.g. a jet direction)
+  ///
+  /// @param track Track to estimate the IP from
+  /// @param vtx   Vertex the track belongs to
+  /// @param direction   The direction 
+  /// @param gctx  The geometry context
+  /// @param mctx  The magnetic field context
+  ///
+  /// @return A pair holding the sign for the 2D an Z lifetimes
```
paulgessinger, line 172: This only returns a single double.

paulgessinger (commented):
Hey @pbutti, can you fix the format check. Also would you consider adding two tiny unit tests to `ImpactPointEstimatoeTests.cpp` just so we have this functionality covered.

Core/include/Acts/Vertexing/ImpactPointEstimator.hpp
pbutti: Fixed. However I added a single test unit that checks both 2d and 3d signed lifetimes

paulgessinger (approved):

PR #1744, opened 2022-12-14
fix: Seed filter should make no use of external sp methods

CarloVarni:
As the name says...

This would prevent us from using the seed on an input collection that does not implement the `y()` and `z()` method. We do not know what the costumer EDM looks like and these same values can already be retrieved without using the external sp methods


CarloVarni:
tagging @paulgessinger 

These changes were started from `v21.1.0`, this is why there are so many commits... but the changes are just 4 lines

Core/include/Acts/Seeding/SeedFilter.ipp
```diff
@@ -268,10 +268,10 @@ void SeedFilter<external_spacepoint_t>::filterSeeds_1SpFixed(
                 float seed1_sum = 0;
                 float seed2_sum = 0;
                 for (int i = 0; i < 3; i++) {
-                  seed1_sum += pow(i1.second->sp[i]->sp().y(), 2) +
-                               pow(i1.second->sp[i]->sp().z(), 2);
-                  seed2_sum += pow(i2.second->sp[i]->sp().y(), 2) +
-                               pow(i2.second->sp[i]->sp().z(), 2);
+                  seed1_sum += pow(i1.second->sp[i]->y(), 2) +
```
tboldagh, line 271: The change looks allright technically.  
The thing that I have not understood was why the "x" coordinate is not used in here.

tboldagh (approved):

Core/include/Acts/Seeding/SeedFilter.ipp
CarloVarni: Resolving this conversation

Issue #1736, opened 2022-12-12
Drop `boost progress` from examples where it is not used


tboldagh:
In this two files:
https://github.com/acts-project/acts/blob/main/Examples/Scripts/MaterialMapping/MaterialComposition.cpp
https://github.com/acts-project/acts/blob/main/Examples/Scripts/TrackingPerformance/TrackSummary.cpp
The boost progress display is defined (which generates deprecation warnings when compiling in docker image) but seems not to be used. Should it be still kept around?
```
[199/1062] Building CXX object Examples/Scripts/MaterialMapping/CMakeFiles/ActsAnalysisMaterialComposition.dir/MaterialComposition.cpp.o
In file included from /usr/include/boost/config/header_deprecated.hpp:18,
                 from /usr/include/boost/progress.hpp:22,
                 from /ATLAS/tbold/acts/ht-py-bindings/acts/Examples/Scripts/MaterialMapping/MaterialComposition.cpp:21:
/usr/include/boost/progress.hpp:23:1: note: #pragma message: This header is deprecated. Use the facilities in <boost/timer/timer.hpp> instead.
   23 | BOOST_HEADER_DEPRECATED( "the facilities in <boost/timer/timer.hpp>" )
```

tboldagh:
Indeed, it was just a slightly annoing warning when compiling.

PR #1746, opened 2022-12-14
fix: Remove progress_display where it is not used

tboldagh:
Solves part of the issue #1736 by removing inclusion of the headers related to progress_display as it was unused.

### DIFF
--- a/Examples/Scripts/MaterialMapping/MaterialComposition.cpp
+++ b/Examples/Scripts/MaterialMapping/MaterialComposition.cpp
@@ -6,6 +6,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+#include "materialComposition.C"
+
 #include "ActsExamples/Utilities/Options.hpp"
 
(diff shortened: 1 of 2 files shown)


Examples/Scripts/MaterialMapping/MaterialComposition.cpp
```diff
@@ -15,19 +17,6 @@
...
-// Boost <=1.71 and lower do not have progress_display.hpp as a replacement yet
-#include <boost/progress.hpp>
-using progress_display = boost::progress_display;
-#else
-// Boost >=1.72 can use this as a replacement
-#include <boost/timer/progress_display.hpp>
-using progress_display = boost::timer::progress_display;
-#endif
-
-#include "materialComposition.C"
```
paulgessinger, line 29: This include conditionally uses `progress_display`. What's the intention of removing this? If we indeed want to remove it, will you remove it from `materialComposition.C` as well?

paulgessinger (commented):
I believe both of these cases are not actually unused. The `cpp` files include `.C` files that do in fact use the `progress_display` if they are not run from ROOT standalone.

PR #1749, opened 2022-12-14
refactor!: Set calibratedSize only through allocateCalibrated

### LINKED ISSUE #1743: Update of calibrate measurement size should become internal to the MTJ backend
In the MTJ implementation we have such line:
https://github.com/acts-project/acts/blob/5cab50d75dc892b31fb76a45e230c41f66e8429d/Core/include/Acts/EventData/MultiTrajectory.hpp#L931

This sets the size of the calibrated measurement. However with different backend (xAOD in ATLAS) we do not need this extra size as the data is kept in vector<vector<scalar>> and the size is known. We can add a fictional method to update the size but 
it seems that we can eliminate this method:
https://github.com/acts-project/acts/blob/5cab50d75dc892b31fb76a45e230c41f66e8429d/Core/include/Acts/EventData/MultiTrajectory.hpp#L888
and let the allocateCallibrated to handle this update. 

Tagging @paulgessinger 
and for info @CarloVarni Marcin Wolter

paulgessinger:
Closes #1743


tboldagh (approved):
Excellent. Thanks Paul!

PR #1752, opened 2022-12-15
build: Install Examples headers

osbornjd:
This PR adds some header installs for several Examples directories so that users can take advantage of some of the Examples infrastructure, for example the EDM objects.


osbornjd:
Going through our code, it looks like we take advantage of functionality in the following four packages in Examples
```
Examples/Detectors/TGeoDetector
Examples/Detectors/MagneticField
Examples/Framework
Examples/Run/Common
```
The TGeoDetector and MagneticField are probably clear (we use some of the functionality to help build and/or read in our geometry and field map), the framework is to take advantage of some of the EDM objects, and Common is to take advantage of some of the option features for (again) reading in some of the geometry/field/material etc. information.

I'll try to check that this is really true by compiling our code against Acts with only these package headers installed, but that appears to be it at first look @paulgessinger

osbornjd:
And one other comment, in order to use `PUBLIC_HEADER` it looks like you need to set a target property for CMake to know all of the headers that need to be installed:
https://stackoverflow.com/questions/46713143/cmake-install-library-and-development-files-as-separate-components-using-public
Before I go implement that, I just wanted to double check that this is what is desired and if so if the idea was to only add the header files that we explicitly are using rather than installing the entire package header list.

paulgessinger:
Ok let's see.

- `Examples/Detectors/MagneticField/include/ActsExamples/MagneticField/MagneticField.hpp` is just 2 typedefs, which I would suggest you just copy-past into your code.
- `TGeoDetector.hpp` is non-trivial functionality, so makes sense. Medium term I guess we should migrate this boilerplate to the TGeo plugin so it can be used from just the Core lib.
- Framework & Common: that's the framework bit. Your choice to use that, but as you're aware that's not something we commit to maintaining fully.

I won't insist on `PUBLIC_HEADER`. I'll leave it up to you if you want to only install the headers that you actually use (- `MagneticField.hpp` maybe, or keep the PR as is installing more headers. Both are ok from my side.

osbornjd:
> Ok let's see.
> 
>     * `Examples/Detectors/MagneticField/include/ActsExamples/MagneticField/MagneticField.hpp` is just 2 typedefs, which I would suggest you just copy-past into your code.

Okay, that's fine, we can do that. I suppose any changes to these typedefs in Acts/Core will then also throw errors in our code when we update, then. The reason we use them is that, as far as I can tell, the typedefs themselves are the only objects actually used outside of Acts/core whenever handling the field.

>     * `TGeoDetector.hpp` is non-trivial functionality, so makes sense. Medium term I guess we should migrate this boilerplate to the TGeo plugin so it can be used from just the Core lib.

That makes sense to me.

>     * Framework & Common: that's the framework bit. Your choice to use that, but as you're aware that's not something we commit to maintaining fully.
> 

We are aware, but again there are useful headers in the framework area such as (for example) the `Trajectories` class [here](https://github.com/acts-project/acts/blob/main/Examples/Framework/include/ActsExamples/EventData/Trajectories.hpp), which you can use to package up the final fit result. Some of them are similar to the MagneticField and we can just define the typedefs in our own code.

> I won't insist on `PUBLIC_HEADER`. I'll leave it up to you if you want to only install the headers that you actually use (- `MagneticField.hpp` maybe, or keep the PR as is installing more headers. Both are ok from my side.

It's clear from our discussion that there is a mix of some functionality that would be useful to be in core and some cleanup/simplification we can do in our code to reduce the dependence on some classes in the `Examples`. Let me go through our code, clean some things up, and see what remains (e.g. I know that TGeoDetector will remain, for example).

PR #1754, opened 2022-12-16
refactor: Make Examples binaries optional

benjaminhuth:
This adds the CMake option `ACTS_BUILD_EXAMPLES_BINARIES` that is `OFF` by default.

Also adds a bit of logic so that `Boost::progarm_options` and `Boost::unit_test_framework` are only imported if they are required. This way, a Acts build without tests, binaries or analysis scripts does only require boost headers.

This has also *some* impact on build times, I save ~6 minutes for `make -j1`...


andiwand (approved):
lets get this in

andiwand (approved):

PR #1762, opened 2022-12-19
feat: Visualization for DetectorVolume and Portal

Matthewharri:
This PR adds in visualization for DetectorVolume and Portals. This uses the existing method of `drawSurface` for the visualization. When drawing, the portals are drawn such that if the Portal contains two links it is drawn as green (as Portals should have two links), otherwise will be drawn red, indicating a link is missing.


asalzburger (changes requested):
Looks all good, I would suggest one change though:
- the `drawX(...)` should get through the signature the `ViewConfig connected`, `ViewConfig unconnected` with the green/red as two default options.

Matthewharri:
@asalzburger this should be all set now with the change you suggested.

paulgessinger (approved):

Core/include/Acts/Visualization/GeometryView3D.hpp
```diff
@@ -81,6 +82,38 @@ struct GeometryView3D {
...
+  /// @param transform An option additional transform
+  /// @param connected The config for connected portals
+  /// @param unconnected The config for unconnected portals
+  static void drawDetectorVolume(
+      IVisualization3D& helper,
+      const Acts::Experimental::DetectorVolume& volume,
+      const GeometryContext& gctx,
+      const Transform3& transform = Transform3::Identity(),
+      const ViewConfig& connected = ViewConfig({0, 255, 0}),
+      const ViewConfig& unconnected = ViewConfig({255, 0, 0}));
```
paulgessinger, line 115: Fo be consistent with the `cpp` and make clang-tidy happy:

```suggestion
  /// @param disconnected The config for unconnected portals
  static void drawDetectorVolume(
      IVisualization3D& helper,
      const Acts::Experimental::DetectorVolume& volume,
      const GeometryContext& gctx,
      const Transform3& transform = Transform3::Identity(),
      const ViewConfig& connected = ViewConfig({0, 255, 0}),
      const ViewConfig& disconnected = ViewConfig({255, 0, 0}));
```

paulgessinger:
@Matthewharri sorry I've missed that the parameter name mismatch is also there in the other files.

benjaminhuth (approved):

paulgessinger:
Wow, I managed to mess that up 😅. I pushed another commit, hopefully that's it.

benjaminhuth (approved):

PR #104, opened 2020-04-13
Implentation of CUDA for seed finding

beomki-yeo:
This PR is for implementing cuda plugin for seedfinding

### Major changes (Edited at 2020 Apr 22)

**1.  CMakeLists.txt for conditional compilation**

- Most of contents has moved to the new plugin: `/home/bomki/Projects/ACTS/Plugins/Cuda` after refactoring

- User can build it with following option in CMakeLists.txt:
option(ACTS_BUILD_CUDA_PLUGIN "Build Cuda plugin" OFF)

**2. Plugins/Cuda/include/Acts/Plugins/Cuda/Utilities**

`CudaScalar` `CudaVector` `CudaMatrix` are cuda memory wrapper classes. `CudaMatrix` is written with column-major style which is compatible with cuBlas (for future usage).

`CpuScalar` `CpuVector` `CpuMatrix` are for cpu counterpart.  User can choose whether to use pinned memory for fast memory transaction.

**3. Plugins/Cuda/include/Acts/Plugins/Cuda/Utilities/Seeding**

- For Cuda Seedfinder,  a partial template class specialization was used.

```
template <typename external_spacepoint_t >
class Seedfinder<external_spacepoint_t, Acts::Cuda > {
//...
}
```

Users can call seedfinder of each platform with the following:
```
Acts::Seedfinder<SpacePoint > seedfinder_cpu(config);
Acts::Seedfinder<SpacePoint, Acts::Cuda> seedfinder_cuda(config);
```

### Validation result (Edited at 2020 Apr 22)

The found seeds were compared between CPU and CUDA algorithm. Several event sizes (10k, 20k, 50k, 100k, and 200k spacepoints) was validated.

-l option is for matrix size and should be given enough big

```
./SeedfinderCudaTest -f sample_10k.txt  -a -l 1 -q
./SeedfinderCudaTest -f sample_20k.txt  -a -l 1 -q
./SeedfinderCudaTest -f sample_50k.txt  -a -l 2 -q
./SeedfinderCudaTest -f sample_100k.txt -a -l 2 -q
./SeedfinderCudaTest -f sample_200k.txt -a -l 4 -q
```

image

 **If you compile the plugin with `-DCMAKE_CUDA_FLAGS="-fmad=false"`, which disable the fast math mode, you can 100% seed matching ratio for all cases above.**

### Speedup in Walltime (Edited at 2020 May 19)

∙ CPU model: i7-5820K CPU @ 3.3 GHz
∙ GPU model: GTX 1070

[...]

PR #1764, opened 2022-12-20
refactor: Managing Seed Candidates with ad-hoc container

CarloVarni:
Rewriting some part of the seeding codes, in particular how we collect the triplets candidates given a specific middle space point. 

As of now these candidates are stored (and passed throughout the code) as `std::pair<float, std::unique_ptr<InternalSeed<external_spacepoint_t>>>`. Replacing the lower quality candidate with a new higher-quality candidate is not trivial if this has to happen multiple times during execution.

This is now handled by a specific class (`CandidatesForSpM<InternalSpacePoint<external_spacepoint_t>>`), which stores triplets candidates in a vector sorted as min heap tree. The first elements of this vector is always guaranteed to be the one with the lowest weight (i.e. quality) even after removals and/or additions of other candidates.

Inserting and removing elements from the stored collection is a log(N) operation.

Also taken a look around and made a few optimization changes:
- Do not create Internal Seeds anymore (do we actually need them at all in Acts??? And why they have their own class if they are nothing more then a `Seed<InternalSpacePoint<external_spacepoint_t>>`?)
- Avoid redundant or unnecessary sorting of vectors
- Modularization of SeedFinder methods to make code cleaner and more readable by humans

Both default and orthogonal seedings have been adapted.
No physics performance changes are expected from this.

Still have to investigate on timings w.r.t. default code

/cc @noemina @LuisFelipeCoelho @paulgessinger @tboldagh @gagnonlg


Core/include/Acts/Seeding/CandidatesForSpM.hpp
```diff
@@ -9,141 +9,136 @@
...
+
+  void bubbleup(std::vector<value_type>&, std::size_t);
+  void bubbledw(std::vector<value_type>&, std::size_t, std::size_t);
+
+  void addToCollection(std::vector<value_type>&, sp_type& SpB, sp_type& SpT,
+                       float weight, float zOrigin, bool isQuality);
+  void insertToCollection(std::vector<value_type>&, sp_type& SpB, sp_type& SpT,
+                          float weight, float zOrigin, bool isQuality);
+
+ public:
```
tboldagh, line 65: Public or private?

Core/include/Acts/Seeding/CandidatesForSpM.hpp
```diff
@@ -9,141 +9,136 @@
...
+  sp_type m_SpM;
+
+  // storage
+  // These vectors are sorted as a min heap tree
+  // Each node is lower then its childs
+  // Thus, it is guaranteed that the lower elements is at the front
+  // Sorting criteria is the seed quality
+
+  // storage for candidates with high quality
+  std::vector<value_type> m_storage_high;
```
tboldagh, line 83: I wonder if the std::set with ordering by "quality" would not be optimal here.

Core/include/Acts/Seeding/CandidatesForSpM.hpp
```diff
@@ -9,141 +9,136 @@
...
-    m_n_low = 0;
-    // clean fixed space points
-    m_SpB = default_value;
-    m_SpM = default_value;
-    // clean storage
-    m_storage_high.clear();
-    m_storage_low.clear();
-  }
-  
+template <typename external_space_point_t>
```
tboldagh, line 18: This class would benefit from more documentation.

Core/include/Acts/Seeding/CandidatesForSpM.hpp
```diff
@@ -9,141 +9,136 @@
...
+class CandidatesForSpM {
+  // Internal collection, we store only the components that are different and
```
tboldagh, line 20: But that is not a collection?

Core/include/Acts/Seeding/CandidatesForSpM.hpp
```diff
@@ -9,141 +9,136 @@
...
+ public:
+  // more complete collection of variables, used by external seeding code
+  enum Components : int { BSP = 0, MSP, TSP, WEIGHT, ZORIGIN, QUALITY };
+
+  using sp_type = external_space_point_t*;
+  using value_type = std::tuple<sp_type, sp_type, float, float>;
+  using output_type = std::tuple<sp_type, sp_type, sp_type, float, float, bool>;
+  static constexpr sp_type default_value = nullptr;
+
+  CandidatesForSpM();
```
tboldagh, line 33: CandidatesForMiddleSp would sound clearer. What do you think?

Core/include/Acts/Seeding/CandidatesForSpM.hpp
```diff
@@ -9,141 +9,136 @@
...
+  ~CandidatesForSpM() = default;
+
+  void setMaxElements(std::size_t n_low, std::size_t n_high);
+  void setMediumSp(sp_type);
```
tboldagh: Medium or Middle

Core/include/Acts/Seeding/CandidatesForSpM.hpp
```diff
@@ -9,141 +9,136 @@
...
+class CandidatesForSpM {
+  // Internal collection, we store only the components that are different and
+  // useful
+  enum InternalComponents : int { I_BSP = 0, I_TSP, I_WEIGHT, I_ZORIGIN };
+
+ public:
+  // more complete collection of variables, used by external seeding code
+  enum Components : int { BSP = 0, MSP, TSP, WEIGHT, ZORIGIN, QUALITY };
+
+  using sp_type = external_space_point_t*;
```
tboldagh, line 28: Somehow I find this alias confusing when used later. I.e. it is reminiscent of some sort of enum. Just a feeling maybe.

Core/include/Acts/Seeding/CandidatesForSpM.hpp
```diff
@@ -9,141 +9,136 @@
...
+  void setMediumSp(sp_type);
+  void setBottomSp(sp_type);
+  std::vector<output_type> storage() const;
+  const sp_type& spM() const;
+
+  void push(sp_type& SpT, float weight, float zOrigin, bool isQuality);
+  void clear();
+
+ private:
+  bool exists(std::size_t, std::size_t) const;
```
tboldagh: unclear what the meaning of the arguments is

tboldagh (commented):
Just a few comments concerning the readability of the code (naming). I tis clearly inherited but maybe it could be improved.  
One technical suggestion about use of quality ordered set. It may be that I have misunderstood things though.

Core/include/Acts/Seeding/CandidatesForSpM.hpp
CarloVarni: Hi @tboldagh, that is something I can try as well and compare with my current implementation. Indeed I was hoping to receive feedbacks exactly on using more efficient data structures. Thanks!

However I believe that `std::set` are implemented as balanced trees with the minimum value being the further left node. Insertion and deletion are log(N) operations (as in my implementation) as well as lookup for the minimum (while in my current implementation minimum lookup is a constant operation since it's always the first element in the vector)

Core/include/Acts/Seeding/CandidatesForSpM.hpp
CarloVarni: indeed it is supposed to be `private`

Core/include/Acts/Seeding/CandidatesForSpM.hpp
CarloVarni: agree, I'll change it

Core/include/Acts/Seeding/CandidatesForSpM.hpp
CarloVarni: I'll take a look and try to make the code cleaner

Core/include/Acts/Seeding/CandidatesForSpM.hpp
CarloVarni: will add

Core/include/Acts/Seeding/CandidatesForSpM.hpp
CarloVarni: I'll improve the documentation

Core/include/Acts/Seeding/CandidatesForSpM.hpp
CarloVarni: I'll change to middle

Core/include/Acts/Seeding/CandidatesForSpM.hpp
CarloVarni: It verifies the storage has that element available. That is used to determine if a father node has a valuable child stored in the collection. The first argument is the node number we are asking to retrieve, the second argument is the number of "valid" elements (i.e. `m_n`).

In the current implementation the `storage` has a maximum size of `m_max_size`, while the number of collected elements in the `storage` is given by `m_n`. When we have added `m_max_size` elements in the collection (and `storage.size()` returns `m_max_size`) and we pop one of them we simply reduce `m_n` by one to save time: the popped element is still in the collection but is considered as "non valid". That means that `storage.size()` will keep returning `m_max_size` but only a few of them are "valid" (i.e. all those with an index lower then `m_n`).

The `exists` function checks we are not trying to access a non valid element.

Core/include/Acts/Seeding/CandidatesForSpM.hpp
tboldagh: These log(N) are realy relevant at large N. Mostly irrelevant for small collections. Maybe still be worth trying.

CarloVarni: A quick test of mine with an ad-hoc code suggests that `std::set` may be slower. I believe this is due to the fact that in a `std::set` elements are often allocated dynamically as they are added and there is no requirement for items to be contiguous in memory, as for a `std::vector`

tboldagh: OK If the point is empirically verified then it is settled. Maybe in the doc of this class it would be worth saying something along these lines:
"This is in effect faster sorted container - implementation with std::set was tried and found slower." 
so that people do not waste time in future as I managed you to do! Apologies Carlo!

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
```diff
@@ -0,0 +1,187 @@
...
+  /// @param storage The collection into which the candidate should be stored
+  /// @param n The current number of stored elements in the container
+  /// @param n_max The maximum number of elements that can be stored in the container
+  /// @param SpB The bottom space point
+  /// @param SpM The middle space point
+  /// @param SpT The top space point
+  /// @param weight The quality of the triplet candidate
+  /// @param zOrigin The z-coordinate of the origin
+  /// @param isQuality Whether the triplet candidate is high or low quality
+  void push(std::vector<std::size_t>& storage, std::size_t& n,
```
tboldagh, line 103: I wonder if may be worth returning true for added and false otherwise.

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
```diff
@@ -0,0 +1,187 @@
...
+  /// @param storage The collection
+  /// @param current_size The current number of element stored in the collection. The function will
+  /// diminish this value by 1
+  void pop(std::vector<std::size_t>& storage, std::size_t& current_size);
+
+  /// @brief Return the weight for a candidate
+  /// @param storage The collection in which the element is stored
+  /// @param n Index of the element in the collection
+  /// @returns The weight of the candidate
+  float weight(const std::vector<std::size_t>& storage, std::size_t n) const;
```
tboldagh, line 128: It is not clear why the storage needs to be passed. Isn't it available as a member of this class?

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
```diff
@@ -0,0 +1,187 @@
...
+  /// weight is lower of it's child's weights. If so, it swaps the element with
+  /// the child with the lowest weight. Reiterate the process untill the element
+  /// is in the correct position on the tree
+  /// @param storage The collection
+  /// @param n The index of the element to place in the correct position
+  /// @param actual_size The current number of elements stored in the collection
+  void bubbledw(std::vector<std::size_t>& storage, std::size_t n,
+                std::size_t actual_size);
+
+  /// @brief Sdding a new triplet candidate to the collection. The function is called after the candidate has satisfied
```
tboldagh, line 147: typo

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
```diff
@@ -0,0 +1,187 @@
...
+  /// all the selection criteria
+  /// @param storage The collection
+  /// @param n Current number of stored elements in the collection
```
tboldagh, line 150: Isn't it known? I mean the current number?

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+  /// @brief constructor
+  CandidatesForMiddleSp() = default;
+  /// @brief Destructor
+  ~CandidatesForMiddleSp() = default;
+
+  /// @brief Setting maximum number of candidates to keep
+  /// @param n_low Maximum number of candidates in the low-quality collection
+  /// @param n_high Maximum number of candidates in the high-quality collection
+  void setMaxElements(std::size_t n_low, std::size_t n_high);
```
tboldagh, line 63: Why there is no constructor of this class taking these max values?

Core/include/Acts/Seeding/CandidatesForMiddleSp.ipp
```diff
@@ -0,0 +1,277 @@
...
+inline void CandidatesForMiddleSp<external_space_point_t>::pop(
+    std::vector<std::size_t>& storage, std::size_t& current_size) {
+  // Remove the candidate with the lowest weight in the collection
+  // By construction, this element is always the first element in the
+  // collection.
+  // The removal works this way: the first element of the collection
+  // is overwritten with the last element of the collection.
+  // The number of stored elements (current_size) is lowered by 1
+  // The new first element is moved down in the tree to its correct position
+  std::swap(storage[0], storage[current_size - 1]);
```
tboldagh, line 40: With c++20 we would use here shift_left I guess.

tboldagh: Anyways, it ma be clearer to write `std::swap(storage.front(), storage.back());`

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
```diff
@@ -0,0 +1,187 @@
...
+  // They are sorted as a min heap tree, in which
+  // Each node is lower then its childs
+  // Thus, it is guaranteed that the lower elements is at the front
+  // Sorting criteria is the seed quality
+  //
+  // This is in effect faster sorted container - implementation with std::set
+  // and std::priority_queue  were tried and found slower.
+
+  // storage for candidates with high quality
+  std::vector<std::size_t> m_storage_high{};
```
tboldagh, line 180: So these are indices in fact, not the storage, right? What do you think about naming these variables accordingly?

Core/include/Acts/Seeding/CandidatesForMiddleSp.ipp
```diff
@@ -0,0 +1,277 @@
...
+  return push(m_storage_low, m_n_low, m_max_size_low, SpB, SpM, SpT, weight,
+              zOrigin, isQuality);
+}
+
+template <typename external_space_point_t>
+void CandidatesForMiddleSp<external_space_point_t>::push(
+    std::vector<std::size_t>& storage, std::size_t& n, const std::size_t& n_max,
+    external_space_point_t& SpB, external_space_point_t& SpM,
+    external_space_point_t& SpT, float weight, float zOrigin, bool isQuality) {
+  // If we do not want to store candidates, returns
```
tboldagh, line 90: Any feeling which of the paths is more likely? May the other of these checks be relevant for performance, again in c++20 we'd mark some of these branches with `[[likely]]`, for now maybe ordering them is useful (so in future we can easily find the likely path and mark it. Just a suggestion.

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+  /// @brief Setting maximum number of candidates to keep
+  /// @param n_low Maximum number of candidates in the low-quality collection
+  /// @param n_high Maximum number of candidates in the high-quality collection
+  void setMaxElements(std::size_t n_low, std::size_t n_high);
+
+  /// @brief Retrieve the triplet candidates, the resulting vector is already sorted,
+  /// elements with higher quality first
+  /// @returns Vector of triplet candidates
+  std::vector<value_type> storage();
```
tboldagh, line 68: It is a bit strange that getter have a chance to modify the object. I.e. is not const method. I guess it has to be so in this case.

Core/include/Acts/Seeding/SeedFilter.hpp
```diff
@@ -52,46 +54,39 @@ class SeedFilter {
...
-      std::vector<std::pair<
-          float, std::unique_ptr<const InternalSeed<external_spacepoint_t>>>>&
-          outCont) const;
+  /// Filter seeds once all seeds for one middle space point have been created
+  /// @param candidates collection of seed candidates
+  /// @param numQualitySeeds number of high quality seeds in seed confirmation
+  /// @param outIt Output iterator for the seeds
+  /// for all seeds with the same middle space point
+  virtual void filterSeeds_1SpFixed(
+      std::vector<typename CandidatesForMiddleSp<
```
tboldagh, line 85: Just commenting that this is realy nice refactor.

Core/include/Acts/Seeding/SeedFilter.ipp
```diff
@@ -212,162 +208,95 @@ void SeedFilter<external_spacepoint_t>::filterSeeds_2SpFixed(
...
-        seedsPerSpM,
-    int& numQualitySeeds,
+    std::vector<typename CandidatesForMiddleSp<
+        InternalSpacePoint<external_spacepoint_t>>::value_type>& candidates,
+    std::size_t& numQualitySeeds,
     std::back_insert_iterator<std::vector<Seed<external_spacepoint_t>>> outIt)
     const {
-  // sort by weight and iterate only up to configured max number of seeds per
-  // middle SP
-  std::sort((seedsPerSpM.begin()), (seedsPerSpM.end()),
```
tboldagh, line 262: Thanks!

Core/include/Acts/Seeding/SeedFinderUtils.hpp
```diff
@@ -56,16 +56,16 @@ LinCircle transformCoordinates(external_spacepoint_t& sp,
 /// @param[in] bottom Should be true if vec are bottom spacepoints.
 /// @param[out] linCircleVec The output vector to write to.
 template <typename external_spacepoint_t>
-void transformCoordinates(
+std::vector<std::size_t> transformCoordinates(
```
tboldagh, line 59: There is now a returned object. Should documentation be updated?

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
```diff
@@ -0,0 +1,187 @@
...
+            bool isQuality);
+
+  /// @brief Clear the internal storage
+  void clear();
+
+  /// @brief A function for sorting the triplet candidates from higher to lower quality
+  /// @param i1 First triplet candidate
+  /// @param i2 Second triplet candidate
+  /// @returns The comparison result
+  static bool greaterSort(const value_type& i1, const value_type& i2);
```
tboldagh, line 89: Sine the name is arbitrary, can it be more meaningful and contain word "Quality" and hint about the actual order? My pick would be: "descendingByQuality"

tboldagh (changes requested):
Had a quick look. I realy have (again) some better naming suggestions and questions. Besides that it all looks good.

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
CarloVarni: done

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
CarloVarni: fixed

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
CarloVarni: the `storage` here is one of the two indices collections and `n` refers to the number of elements in that specific index collection. Since we have two different indices collections I preferred to pass them as inputs to the function.

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
CarloVarni: Same comment as above about the use of two different indices collections

Core/include/Acts/Seeding/CandidatesForMiddleSp.ipp
CarloVarni: I'm afraid not, it is not guaranteed that `storage.back()` corresponds to the element we want to swap. This happens if we pop more then once without adding (i.e. in the `storage()` method).

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
CarloVarni: changed to `m_indices_high` and `m_indices_low`

Core/include/Acts/Seeding/CandidatesForMiddleSp.ipp
CarloVarni: Hard to say since it depends on the max number of seeds the user wants to keep (default is 5 as of now) and how crowded the event is.

Core/include/Acts/Seeding/SeedFinderUtils.hpp
CarloVarni: updated

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
CarloVarni: Changed to `descendingByQuality`

Core/include/Acts/Seeding/CandidatesForMiddleSp.hpp
CarloVarni: This is only because I've added the `CandidateForMiddleSp` object in the `SeedingState` in order to create this object only once. In order to avoid misconfigurations (due to the user changing the max seed number) I've decided to introduce this method and set these values in the seed filter directly.

CarloVarni:
On a positive side, the effect of these changes is a slightly faster code. The plot is for the default seeding, but the same consideration applies to the orthogonal seeding. 

Sample: tt-bar PU200 (50 events)
Script: `full_chain_itk.py`

plot_default

paulgessinger:
@tboldagh, @noemina can one of you review this and hit "approve" if we can merge?

LuisFelipeCoelho (approved):

Core/include/Acts/Seeding/SeedFilter.ipp
```diff
@@ -189,17 +190,12 @@ void SeedFilter<external_spacepoint_t>::filterSeeds_2SpFixed(
       if (deltaSeedConf > 0) {
         // if we have not yet reached our max number of quality seeds we add the
         // new seed to outCont
+
+        candidates_collector.push(bottomSP, middleSP, *topSpVec[topSPIndex],
+                                  weight, zOrigin, true);
```
noemina, line 195: Can you extend the comment mentioning that `push` also checks the max number of quality seeds for a middle sp? If this is reached, we remove a seed with lower weight.

noemina (commented):
Great code! I really like it. I have left a few comments to correct typos and add lines to the documentation.
The major is to better explain the logic in the `bubbledw` method.

Core/include/Acts/Seeding/CandidatesForMiddleSp.ipp
```diff
@@ -0,0 +1,284 @@
...
+      bottom_l1->y() * bottom_l1->y() + bottom_l1->z() * bottom_l1->z();
+  seed1_sum +=
+      middle_l1->y() * middle_l1->y() + middle_l1->z() * middle_l1->z();
+  seed1_sum += top_l1->y() * top_l1->y() + top_l1->z() * top_l1->z();
+
+  seed2_sum +=
+      bottom_l2->y() * bottom_l2->y() + bottom_l2->z() * bottom_l2->z();
+  seed2_sum +=
+      middle_l2->y() * middle_l2->y() + middle_l2->z() * middle_l2->z();
+  seed2_sum += top_l2->y() * top_l2->y() + top_l2->z() * top_l2->z();
```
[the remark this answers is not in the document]
CarloVarni: This was introduced with https://github.com/acts-project/acts/pull/104 to make CUDA and CPU results the same. I'm not sure what the reasoning is here

CarloVarni:
Hi @noemina I should have addressed you comments. In particular I've extended the documentation on the `bubbledw` method. Hopefully it is now clearer what it does

Core/include/Acts/Seeding/CandidatesForMiddleSp.ipp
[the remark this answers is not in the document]
CarloVarni: @noemina if you are ok with it, I'll keep it as it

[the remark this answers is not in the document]
noemina: yes, I am ok with it.

CarloVarni:
thanks @noemina for the feedback. All threads are resolved. However the CI fails due to the codecov test, so I'm afraid that the automerge will not work

noemina:
Great! All good from my side.

paulgessinger (approved):
Assuming @noemina meant to approve this PR.

I suspect the codecov might resolve when this branch is updated to current main.

noemina (approved):
Of course!

CarloVarni:
@paulgessinger I'm afraid codecov is stubborn and doesn't want to succeed.

PR #1767, opened 2022-12-21
feat: new Acts::Detector building from Photodetector description (cylindrical)

asalzburger:
This PR introduced the infrastructure and classes for cylindrical detector building following a blue print proto detector.

This should allow maximal flexibility:

* The `Acts::Protodetector` description carries a blue print of what building blocks to build in which order
* This is then executed in a simple builder that follows this description
* It should allow telescope detector, cylindrical symmetric detectors and asymmetric detectors to be built with the new, same schema
*


asalzburger:
Superseded, closing.

PR #1769, opened 2022-12-22
feat: More robust navigation with bended tracks

### LINKED ISSUE #1596: Navigator: Resolving surfaces/layers/boundaries with bending tracks
**Problem:**
I recently had a problem with the navigation, when experimented with a toy detector with alot of surfaces. 

The problem was that resolving layers/surfaces/boundaries is done via straight line intersection. With a large geometry and/or very bending tracks this can become a problem (see e.g. the following sketch for my case with the forward-backward pass of a fitter)

image_2022_10_12T14_14_14_337Z

Now it turned out that this is also a problem e.g., in LDMX (@pbutti) with low momentum particles in a box volume, where sometimes boundaries are missed due to this issue.

**Solution**:
Since this is a problem also occuring in client code, I think that we shouldn't wait for the layerless geometry with that.

There are some possible solutions:
* Make the resolution stragegy configureable with the options `straight line` and `full`, which simply resolves all surfaces/layers/boundaries. We must check, if there could occur problems with the sorting of these surfaces by distance to the position.
* I think helix-extrapolation or something like that for the resolution would over-engineer the problem, but is also an option

@paulgessinger @asalzburger what do you think?

benjaminhuth:
This PR adds an option to the `Acts::Navigator` to disable the boundary check on layer resolval. It also adds an integration test that does check exactely this in a simple cuboid geometry. @pbutti could you check if this solves your problem?

should fix #1596


Tests/IntegrationTests/PropagationBendedTracks.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+BOOST_AUTO_TEST_CASE(with_boundary_check_no_bfield) {
+  auto logger = Acts::getDefaultLogger("test", Acts::Logging::FATAL);
+
+  auto navCfg = Acts::Navigator::Config{};
+  const auto xPositions =
+      xPositionsOfPassedSurfaces(navCfg, Acts::LoggerWrapper{*logger}, 0.0_T);
+
+  // without befield we exit at the side so we don't hit the surfaces at x ~
```
paulgessinger, line 80:
```suggestion
  // without bfield we exit at the side so we don't hit the surfaces at x ~
```

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -159,6 +159,10 @@ class Navigator {
 
     /// The tolerance used to defined "reached"
     double tolerance = s_onSurfaceTolerance;
+
+    /// Wether to perform boundary checks for layer resolving (improves
+    /// navigation for bended tracks)
```
paulgessinger, line 164: Can we make this a `BoundaryCheck` here already, rather than a book that's being Auto converted later on?

paulgessinger (commented):
I guess it's "bent" tracks.

Core/include/Acts/Propagator/Navigator.hpp
benjaminhuth: good point

Tests/IntegrationTests/PropagationBendedTracks.cpp
```diff
@@ -0,0 +1,139 @@
+// This file is part of the Acts project.
+//
+// Copyright (C) 2020 CERN for the benefit of the Acts project
```
AJPfleger, line 3:
```suggestion
// Copyright (C) 2022 CERN for the benefit of the Acts project
```

Core/include/Acts/Propagator/Navigator.hpp
paulgessinger:
```suggestion
    /// navigation for bent tracks)
```

paulgessinger (commented):
Few more typos, otherwise good to go!

PR #1773, opened 2023-01-03
feat: Add INFO printouts about the layers & volumes

tboldagh:
This PR adds permanent (to be discussed) messages about the geometry volumes & layers numbering. 
It seems to be the only robust way to get this information. 
The implementation is a variant of what @gagnonlg suggested. 

I can imagine an additional parameter steering if this messages occur. 
Tagging @asalzburger @gagnonlg


Core/include/Acts/Geometry/Layer.hpp
```diff
@@ -281,9 +282,12 @@ class Layer : public virtual GeometryObject {
   /// @param layerID is the geometry id of the volume
   ///                as calculated by the TrackingGeometry
   /// @param hook Identifier hook to be applied to surfaces
+  /// @param logger A @c LoggerWrapper instance
+  ///
   void closeGeometry(const IMaterialDecorator* materialDecorator,
                      const GeometryIdentifier& layerID,
-                     const GeometryIdentifierHook& hook);
+                     const GeometryIdentifierHook& hook,
+                     LoggerWrapper logger = getDummyLogger());
```
paulgessinger, line 290: Could this switch to using `const Logger&` with the same `getDummyLogger()` default? I'm trying to phase out `LoggerWrapper`.

tboldagh: Sure. Will do.

Core/src/Geometry/TrackingGeometry.cpp
```diff
@@ -12,17 +12,20 @@
...
+#include "Acts/Utilities/Logger.hpp"
 
 Acts::TrackingGeometry::TrackingGeometry(
     const MutableTrackingVolumePtr& highestVolume,
     const IMaterialDecorator* materialDecorator,
     const GeometryIdentifierHook& hook)
     : m_world(highestVolume),
       m_beam(Surface::makeShared<PerigeeSurface>(Vector3::Zero())) {
+  std::unique_ptr<const Logger> logger =
+      getDefaultLogger("TrackingGeometry", Logging::DEBUG);
```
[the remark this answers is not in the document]
tboldagh: Actually I think we would want to steer this printout externally (by logging level I guess). But I was not able to walk through the objects/classes so see where is the object that we interact with in python. It seems ODD geometry!?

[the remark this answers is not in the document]
tboldagh: I think we need a bit help in navigating where & what. Would be right ting to do to add logger as an argument to the TrackingGeometry c'tor. Then actually what? I would need to add configuring of it somewhere here for the ITk:
https://github.com/acts-project/acts/blob/main/Examples/Detectors/TGeoDetector/src/TGeoDetector.cpp 
?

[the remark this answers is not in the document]
tboldagh: ok. I think I have found the way up to the level when the logging level is settable from python script. 
+ fixes requested + INFO demoted to DEBUG actually. Please have a look.

Core/src/Geometry/TrackingGeometry.cpp
```diff
@@ -16,13 +16,15 @@
 Acts::TrackingGeometry::TrackingGeometry(
     const MutableTrackingVolumePtr& highestVolume,
     const IMaterialDecorator* materialDecorator,
-    const GeometryIdentifierHook& hook)
+    const GeometryIdentifierHook& hook, const Logger& logger)
     : m_world(highestVolume),
       m_beam(Surface::makeShared<PerigeeSurface>(Vector3::Zero())) {
+  std::unique_ptr<const Logger> mylogger = logger.clone("TrackingGeometry");
```
paulgessinger, line 22: This works, but since `logger` is never used for anything else, why not simply pass through directly (i.e. why the `clone`)?

Core/src/Geometry/TrackingGeometryBuilder.cpp
```diff
@@ -52,8 +52,9 @@ Acts::TrackingGeometryBuilder::trackingGeometry(
     const IMaterialDecorator* materialDecorator =
         m_cfg.materialDecorator ? m_cfg.materialDecorator.get() : nullptr;
     // build and set the TrackingGeometry
-    trackingGeometry.reset(new TrackingGeometry(
-        highestVolume, materialDecorator, m_cfg.geometryIdentifierHook));
+    trackingGeometry.reset(
+        new TrackingGeometry(highestVolume, materialDecorator,
+                             m_cfg.geometryIdentifierHook, logger()));
```
paulgessinger, line 57: I think this should be the same:
```suggestion
                             m_cfg.geometryIdentifierHook, logger));
```

Core/src/Geometry/TrackingGeometry.cpp
```diff
@@ -16,13 +16,15 @@
...
+
   // Close the geometry: assign geometryID and successively the material
   size_t volumeID = 0;
-  highestVolume->closeGeometry(materialDecorator, m_volumesById, volumeID,
-                               hook);
+  highestVolume->closeGeometry(materialDecorator, m_volumesById, volumeID, hook,
+                               *(mylogger.get()));
```
paulgessinger, line 27: Depending on if you changed `mylogger` or not, you might not need the `.get()`:
```suggestion
                               *(mylogger));
```

paulgessinger:
The pattern seems fine to me. If you can get rid of the `LoggerWrapper`s, I'm happy!

Core/src/Geometry/TrackingGeometry.cpp
tboldagh: On this. It is used here to rename "source". Maybe bad idea. If I could just clone it w/o capturing I think we would leak memory. No?

Core/src/Geometry/TrackingGeometryBuilder.cpp
tboldagh: To my understanding not realy. We need Logger& here. m_logger is unique ptr and we could call `*m_logger.get()` but having the logger() method returning already ref it seems better to just use it.

Core/src/Geometry/TrackingGeometry.cpp
tboldagh: Yes. My intention was to update the name. So there is a clone.

Core/src/Geometry/TrackingVolume.cpp
```diff
@@ -507,7 +509,7 @@ Acts::TrackingVolume::compatibleBoundaries(
 
     ACTS_VERBOSE("Check intersection with surface "
                  << bSurface->surfaceRepresentation().geometryId());
-    if (detail::checkIntersection(sIntersection.intersection, pLimit, oLimit,
+    if (detail::checkIntersection<decltype(sIntersection.intersection), decltype(logger)>(sIntersection.intersection, pLimit, oLimit,
```
tboldagh, line 512: @paulgessinger - this (and few lines below) call got a bit uglier. The compiles is not able to deduce that we want to pass by ref w.o these explicit parametrisation. It assumes we want to pass longer by copy and that is disallowed. 
What go you think? Does the loger parameter in the checkIntersection need to be templated parameter?

Core/src/Geometry/TrackingGeometry.cpp
tboldagh: Indeed. Given that the logger can have any name I thought that this way we can make the geometry printout kind of stable. But if this is no tin line of logging policy, surely can change.

Core/src/Geometry/TrackingVolume.cpp
paulgessinger: Is it not able to deduct the logger type?

paulgessinger: Also, from looking at it, no I don't think it needs to be templated, the whole mechanism checking the type of the logger can probably go away. Do you want to make this change in this PR, or should this be a follow up PR?

tboldagh: Do you mean that the logger can be a concrete type? Sure I can do it.

tboldagh: Actually that change cause avalanche of changes that should be done for clients, and client of clients .... 
I think it is better to do it as a followup.

paulgessinger: Ok!

Core/src/Geometry/TrackingGeometry.cpp
tboldagh: Removed the loden, just passing down the log now.

tboldagh:
The test seem to fail in quite obscure way on licence checks and include guards but no hint is given as to where and what is incorrect. Tagging @paulgessinger

paulgessinger:
I think this is a failure on GitHub's side

tboldagh:
Is there anything that I can do to make it work?

paulgessinger:
https://github.com/acts-project/acts/pull/1799 This should fix it I believe, @tboldagh.

tboldagh:
@paulgessinger 
> #1799 This should fix it I believe, @tboldagh.

There seems to be still some issue. I have made last push after the #1799 was merged

paulgessinger (approved):
Just the approval missing I think.

gagnonlg:
@tboldagh @paulgessinger how do you get this feature to work with the python-based workflow? I (perhaps naively) tried doing something like
```
    acts.logging.setLevel(acts.logging.DEBUG)
    detector, trackingGeometry, decorators = acts.examples.GenericDetector.create()
```

But nothing was printed

paulgessinger:
@gagnonlg the `setLevel` call does not set a global level for the other components. I believe `GenericDetector` has log levels in its config that you can set (you might be able to do this as `kwargs` to the `create()` call.

PR #1774, opened 2023-01-03
refactor!: Separate Config and Options for SpacePointGrid

CarloVarni:
As done for the SeedFinder, this PR separates Config and Options for the `SpacePointGrid`, thus having `SpacePointGridConfig` and `SpacePointGridOptions`. The `SpacePointGridOptions` contains only the b-field info. The value of this b-field is required to be equal to the value stored in `SeedFinderOptions`: https://github.com/acts-project/acts/blob/main/Examples/Algorithms/TrackFinding/src/SeedingAlgorithm.cpp#L121 

This requirement is not asked in Athena as of now and a fixed value from JO is currently used instead for the `SpacePointGrid` while it is updated on every event for the `SeedFinder` (this inconsistency will be fixed asap).

tagging @tbold @paulgessinger


tboldagh (approved):
As far as I can see these are only corrections of typos & inconsistencies? Approving.

CarloVarni:
@tboldagh can you re-approva this PR? I believe the CI needs an approval to merge it

tboldagh (approved):
Re-approving

PR #1775, opened 2023-01-07
fix: Orthogonal seeding beamspot and function

CarloVarni:
The orthogonal seed finder was creating internal space points imposing a beam spot position of `{0, 0}` which is hardly the case if we are not running on MC. 

Also, adding a function for extracting global position and variance from the external space point, instead on assuming methods. The function has to be user-provided.


Examples/Scripts/Python/full_chain_itk.py
```diff
@@ -12,6 +12,7 @@
 )
 from acts.examples.reconstruction import (
     addSeeding,
+    SeedingAlgorithm,
```
stephenswat, line 15: Is this a related change that needs to be included here?

stephenswat (changes requested):
Excellent idea, must absolutely be implemented. I would like to see the function objects passed as arbitrary template parameters.

Core/include/Acts/Seeding/SeedFinderOrthogonal.hpp
```diff
@@ -100,10 +101,14 @@ class SeedFinderOrthogonal {
...
+   * @param extract_coordinates User-defined function for extracting global position and
+   * covariance of the external space point
    */
   template <typename input_container_t, typename output_container_t>
-  void createSeeds(const input_container_t &spacePoints,
-                   output_container_t &out_cont) const;
+  void createSeeds(
+      const input_container_t &spacePoints, output_container_t &out_cont,
+      std::function<std::pair<Acts::Vector3, Acts::Vector2>(
+          typename input_container_t::value_type)> &extract_coordinates) const;
```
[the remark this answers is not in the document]
CarloVarni: Thanks for the comment. I should have implemented your suggestion. Take a look at the new diff to confirm you are are satisfied with the new changes.

Examples/Scripts/Python/full_chain_itk.py
CarloVarni: This is for selecting the seeding algorithm (e.g. Default vs Orthogonal) in the job option. It is not really connected to the PR title though. I'd keep it here without opening a different PR for this additional trivial change

stephenswat (approved):
Perfect, thank you!

PR #1777, opened 2023-01-09
fix: Linkage errors in BetheHeitlerApprox.hpp

benjaminhuth:


AJPfleger (approved):
seems ok

andiwand (approved):

PR #1779, opened 2023-01-10
feat: GeoID hook from python & ODD hook

benjaminhuth:
This PR adds a mechanism to add a geometry identifier hook directly from python, e.g. in `odd.py` (thanks @paulgessinger for the suggestion with the `py::object`). So there is no need to awkwardly define a ODD-specific hook somewhere in the C++ code.

I think this is **not breaking, even though** it changes `Acts::GeometryIdentifierHook` from a typedef to a `std::function` to a virtual baseclass. However, in `docs/versioning.rst` the Geometry module is not defined as "subject of the versioning".

The extra-bits in the ODD are now set like this:

image

Thanks also @Corentin-Allaire for the preparing work! I would suggest to close #1696 after this is merged.

cc @asalzburger


Examples/Python/src/Base.cpp
```diff
@@ -238,23 +238,29 @@ void addAlgebra(Acts::Python::Context& ctx) {
...
   py::class_<Acts::Vector4>(m, "Vector4")
       .def(py::init<double, double, double, double>())
       .def(py::init([](std::array<double, 4> a) {
         Acts::Vector4 v;
         v << a[0], a[1], a[2], a[3];
         return v;
-      }));
+      }))
+      .def("__getitem__",
+           [](const Acts::Vector4& self, Eigen::Index i) { return self[i]; });
```
Corentin-Allaire, line 263: I might have missed it but what is the context of those changes ?

Corentin-Allaire (approved):
Looks great to me ! Thanks for taking care of this.

Examples/Python/src/Base.cpp
benjaminhuth: Without the `__getitem__` functions it is not possible to access the content of an `Acts::Vector3` in python, and thus would not be possible to compute the radius to set the extra id.

benjaminhuth: I added them also to the other vectors for consistency

Corentin-Allaire: Make sense !

Core/src/Geometry/GeometryIdentifier.cpp
```diff
@@ -24,7 +24,7 @@ std::ostream& Acts::operator<<(std::ostream& os, Acts::GeometryIdentifier id) {
                                               id.sensitive(), id.extra()};
 
   bool writeSeparator = false;
-  for (auto i = 0u; i < 5u; ++i) {
+  for (auto i = 0u; i < (sizeof(levels) / sizeof(levels[0])); ++i) {
```
paulgessinger, line 27: That's not great. Can we use a `static const std::array` up top and use `size()` here?

Examples/Detectors/TGeoDetector/include/ActsExamples/TGeoDetector/TGeoDetector.hpp
```diff
@@ -61,8 +61,7 @@ struct TGeoDetector {
         Acts::TGeoLayerBuilder::defaultElementFactory;
 
     /// Optional geometry identfier hook to be used during closure
-    /// @note Will be @b copied when calling the geometry building
-    Acts::GeometryIdentifierHook geometryIdentifierHook;
+    std::shared_ptr<const Acts::GeometryIdentifierHook> geometryIdentifierHook;
```
paulgessinger, line 64: Default also here or not needed?

Examples/Python/src/Geometry.cpp
```diff
@@ -19,12 +20,40 @@
...
+             [](Acts::Surface& self) { return self.geometryId().approach(); })
+        .def("sensitiveId",
+             [](Acts::Surface& self) { return self.geometryId().sensitive(); })
+        .def("extraId",
+             [](Acts::Surface& self) { return self.geometryId().extra(); })
+        .def("center", [](Acts::Surface& self) {
+          return self.center(Acts::GeometryContext{});
+        });
+  }
   {
```
paulgessinger, line 57: Would it make sense to just add bindings to `GeometryIdentifier` itself, and add a binding for the `geometryId()` getter itself?

paulgessinger: Ah, `GeometryIdentifier` has bindings already, so do we need these extra methods on `Surface`?

paulgessinger:
I like this a lot, thanks @benjaminhuth !

Examples/Detectors/TGeoDetector/include/ActsExamples/TGeoDetector/TGeoDetector.hpp
benjaminhuth: Good catch, should be added to prevent crashes from default constructed configs.

Examples/Python/src/Geometry.cpp
benjaminhuth: actually not I would say, I'll remove them in favor of a geometry identifier accessor.

Core/src/Geometry/GeometryIdentifier.cpp
benjaminhuth: You're right, I overlooked this somehow. We should use `std::array` for both `names` and `levels`.

benjaminhuth:
I applied your suggestions @paulgessinger, If you approve again and the CI runs through, I think we can merge!

paulgessinger (approved):

paulgessinger (approved):

PR #1784, opened 2023-01-13
docs: Adding script and documentation for geometry and material handling

noemina:
Self-explanatory title.


docs/howto/run_material_mapping.rst
```diff
@@ -87,6 +87,58 @@ Then edit the config-map.json file
 
    $ python3 <source>/Examples/Scripts/MaterialMapping/configureMap.py geometry-map.json config-map.json
 
+Geometry visualisation and preparation of JSON files
```
tboldagh, line 90: Last time while writing the doc I was told that ACTS moved away from rst to md. Not sure it applies here.

Examples/Scripts/MaterialMapping/GeometryVisualisationAndMaterialHandling.py
```diff
@@ -0,0 +1,784 @@
...
+    interesting_volumes = []
+    v_index = 0
+    for elements in index_to_extends_layers_cylinders:
+        for coords in elements:
+            if v_index not in interesting_volumes:
+                interesting_volumes.append(v_index)
+            if index_to_names[v_index] not in steering_map:
+                steering_map[index_to_names[v_index]] = index_info(v_index+1)
+            steering_map[index_to_names[v_index]].layers_no_approach.append(coords[5])
+        v_index = v_index+1
```
tboldagh, line 146: It is realy a question. Does it mean that the volume ID are "generated" here and reflect exactly what is done doing geometry building in c++? My concern is if the could be desynchronised.

Examples/Scripts/MaterialMapping/GeometryVisualisationAndMaterialHandling.py
```diff
@@ -0,0 +1,784 @@
...
+
+    v_index = 0
+    for elements in index_to_extends_layers_discs:
+        for coords in elements:
+            if v_index not in interesting_volumes:
+                interesting_volumes.append(v_index)
+            if index_to_names[v_index] not in steering_map:
```
tboldagh, line 153: The logic here is straightforward. However, due to depth of the structure not super straightforward to capture. 
Could a helper function alike `appendIfMissing(dict, index)` help?

Examples/Scripts/MaterialMapping/GeometryVisualisationAndMaterialHandling.py
```diff
@@ -0,0 +1,784 @@
...
+        color = cm.rainbow(np.linspace(0, 1, len(index_to_extends_layers_cylinders)))
+
+        is_in_legend = []
+
+        plt.figure(figsize=(20,10))
+        # Plot all layers (w/o approach layers)
+        v_index = 0
+        for elements in index_to_extends_layers_cylinders:
+            for coords in elements:
+                A, B = [coords[3], coords[1]], [coords[2], coords[1]]
```
tboldagh, line 216: Does having A & B help to understand this better? Anyways, maybe a wrapping function with a clear name:
`x_values, y_values = fromCoordsToXY(coords)` could make it easier to follow.

docs/howto/run_material_mapping.rst
```diff
@@ -87,6 +87,58 @@ Then edit the config-map.json file
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To help you visualising the geometry and understand the hierarchical structure of volumes and layers, a script has been provided: ``Examples/scripts/MaterialMapping/GeometryVisualisationAndMaterialHandling.py``. The documentation of the module can be printed using:
+
+.. code-block:: console
+
+   $ python3 <source>/Examples/Scripts/MaterialMapping/GeometryVisualisationAndMaterialHandling.py --help
+
+   usage: GeometryVisualisationAndMaterialHandling.py [-h] [--geometry GEOMETRY] [--plot]
```
tboldagh, line 99: Maybe there is no need to quote help. It may get out of sync. Right?

docs/howto/run_material_mapping.rst
```diff
@@ -87,6 +87,58 @@ Then edit the config-map.json file
 
    $ python3 <source>/Examples/Scripts/MaterialMapping/configureMap.py geometry-map.json config-map.json
 
+Geometry visualisation and preparation of JSON files
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
tboldagh, line 91: In the Howtos section  there is this "[Howto produce geometry selection for SP making](https://acts.readthedocs.io/en/latest/howto/produce_geometry_selection.html)"
I guess this doc kind of makes the other obsolete. Should it be removed then in this PR?

tboldagh (commented):
Thanks, this seems like exactly the tool that I needed for dining which layers SPs should be made. I have only one concern (due to my limited understanding) if  the volume IDs assignment in C++ and here is guaranteed to be the same always.

Corentin-Allaire (changes requested):
Looks great I think it will be really useful to people doing material mapping in the future ! 
Maybe if you could add a line at the start of the how to material mapping mentioning this exist ?
I think adding some helper function to reduce a bit the depth of the code would be nice but if you don't have the time and since this is just a python script it is probably not mandatory.

docs/howto/run_material_mapping.rst
paulgessinger: At the very list they should be cross-linked, but yeah maybe it's best to remove that one.

Examples/Scripts/MaterialMapping/GeometryVisualisationAndMaterialHandling.py
noemina: The module reads the json file dumped from the tracking geometry. So, yes, I would say that this is reflecting what you generate with the c++ geometry building code.
I tested this within athena for the ITk and with the standalone ITk chain and all seems consistent.

noemina:
Thank you @tboldagh and @Corentin-Allaire for your useful comments. I am going to work on them asap.

Examples/Scripts/MaterialMapping/GeometryVisualisationAndMaterialHandling.py
tboldagh: But is this good enough? I think recently there is a functionality added to provide customised "labeler of volumes & layers" https://github.com/acts-project/acts/pull/1779. I understand the plan is to apply it also for the ITk. 

Anyway, what would we risk if the c++ and configs obtained though this script are out of sync? 
My thinking is following. If the issue due to inconsistency is evident and obvious. Fine, it will be spot, investigate and fixed quickly. 
If the issue may be subtle and visible only when doing careful validation maybe we should strive for perfect consistency?

Or maybe we should say in the doc, (take these volume & layer IDs with a grain of salt and please run GeoBuilding in DEBUG mode to re-verify them?

noemina: The indices dumped in the JSON used as input to this module are read from the geometry, whatever is the geometry ID hook mechanism. This means that they are the exact same ones as set in the geometry building.
There is no labelling of volumes and layers in this script.
I am pretty sure this script works even after https://github.com/acts-project/acts/pull/1779 is merged, but I can of course double check it.

tboldagh: oh. ok. I have missed that. Then this is non-issue. Sorry for the noise.

tboldagh:
What is prognosis to merging this one? Sorry for pushing but I wanted to make geo-selection file that is good for Hough Transform (i.e. last pixel layer and all strips) and this would be quite useful.

Corentin-Allaire:
> What is prognosis to merging this one? Sorry for pushing but I wanted to make geo-selection file that is good for Hough Transform (i.e. last pixel layer and all strips) and this would be quite useful.

Let me review the changes and I think it can go in afterward ?

Corentin-Allaire (approved):
Looks great to me !

PR #1757, opened 2022-12-16
refactor!: Changes to SeedFinderOrthogonal to allow reuse of the object 

tboldagh:
This PR applies similar changes to these in #1693 . As a result the SeedFinderOrthogonal does not need to be recreated for every event. 

Closes #1690

BREAKING CHANGE 
Method to crate OthogonalSeedFinder and createSeeds are modified. The former does not take SeedOptions while the later does now.

PR #1785, opened 2023-01-13
fix: SeedFinder(Orthogonal) handling of interna units for config and options object

CarloVarni:
Removing some derived variables from the `SeedFinderOrthogonalConfig`. These variables are derived from the magnetic filed and, as such, have to be recomputed for every event, while the other variables in the config are constant. 

These variables are already present in the `SeedFinderOptions` class, so I've added proper functions for their computation from the `SeedFinderOrthogonalConfig`. That's the same computation used for `SeedFinderConfig` but since they are two different classes I had to add a dedicated function that performs the same computation... not ideal imho

In addition (and probably main point in all of this PR): fixing issue computation of variables in the config and options objects related to a mis-handling of internal units

/cc @tbold @stephenswat

### DIFF
--- a/Core/include/Acts/Seeding/SeedFinderOrthogonalConfig.hpp
+++ b/Core/include/Acts/Seeding/SeedFinderOrthogonalConfig.hpp
@@ -128,7 +128,7 @@ struct SeedFinderOrthogonalConfig {
     // calculation of scattering using the highland formula
     // convert pT to p once theta angle is known
     config.highland = 13.6 * std::sqrt(radLengthPerSeed) *
-      (1 + 0.038 * std::log(radLengthPerSeed));
+                      (1 + 0.038 * std::log(radLengthPerSeed));
     const float maxScatteringAngle = config.highland / minPt;
     config.maxScatteringAngle2 = maxScatteringAngle * maxScatteringAngle;
     return config;
--- a/Core/include/Acts/Seeding/SeedFinderOrthogonalConfig.hpp
+++ b/Core/include/Acts/Seeding/SeedFinderOrthogonalConfig.hpp
@@ -104,9 +104,16 @@ struct SeedFinderOrthogonalConfig {
   float highland = 0;
   float maxScatteringAngle2 = 0;
 
+  bool isInInternalUnits = false;
   SeedFinderOrthogonalConfig toInternalUnits() const {
+    if (isInInternalUnits) {
+      throw std::runtime_error(
+          "Repeated conversion to internal units for "
+          "SeedFinderOrthogonalConfig");
+    }
     using namespace Acts::UnitLiterals;
     SeedFinderOrthogonalConfig config = *this;
+    config.isInInternalUnits = true;
     config.minPt /= 1_MeV;
     config.deltaRMinTopSP /= 1_mm;
     config.deltaRMaxTopSP /= 1_mm;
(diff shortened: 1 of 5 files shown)


CarloVarni:
I've been too fast in clicking the "ask review" button. There is still some work to do

Core/include/Acts/Seeding/SeedFinderOrthogonalConfig.hpp
```diff
@@ -127,6 +123,16 @@ struct SeedFinderOrthogonalConfig {
...
+  SeedFinderOrthogonalConfig calculateDerivedQuantities() const {
+    SeedFinderOrthogonalConfig config = *this;
+    // calculation of scattering using the highland formula
+    // convert pT to p once theta angle is known
+    config.highland = 13.6 * std::sqrt(radLengthPerSeed) *
+                      (1 + 0.038 * std::log(radLengthPerSeed));
+    const float maxScatteringAngle = config.highland / minPt;
+    config.maxScatteringAngle2 = maxScatteringAngle * maxScatteringAngle;
+    return config;
+  };
```
CarloVarni, line 135: I need a clarification here @tboldagh and @stephenswat 
Are these variables supposed to be computer with config in internal units? 

As of now (i.e. before my changes) the orthogonal seeding is computing these BEFORE converting to internal units. That means that minPt is 900 (entered as MeV in JO - but 1 MeV = 0.1 in Acts units).

My code computes these quantities AFTER converting to internal units, and this minPt is 0.9 (Correctly MeV). This produces the differences in the physics performances as posted by the bot.

Do you have comments here? I'd suppose my approach is the correct one, but it comes with a change in performance (and much less seeds btw)

stephenswat: Hi Carlo, when you say a change in performance do you mean a change in physics performance or a change in compute performance?

It seems to me your logic makes sense. This might also explain why this seeded sometimes produces many more seeds than the classic seeding algorithm.

CarloVarni: Concerning physics performance, the main effect is the duplicate rate: https://herald.dokku.paulgessinger.com/view/acts-project/acts/510272465/seeding_orthogonal.html 

The number of holes as well is effected: https://herald.dokku.paulgessinger.com/view/acts-project/acts/510272465/ckf_orthogonal.html 

See the physmon plots posted in the PR for all the details

CarloVarni: Concerning the number of produced seeds we go from 2616 to 1655 (63%)

Running on lxplus 
100 events
n threads: -1

CarloVarni: timing
Quickly-performed study on timing. Not surprisingly this is faster

stephenswat: That is an excellent result! We should go ahead with this!

tboldagh: So we found it twice, see there, comment 2:
https://github.com/acts-project/acts/pull/1757
It could have spared some @CarloVarni  time if the other PR wold have been merged already.

CarloVarni: Hi @tboldagh I see. I suppose then your PR fully superseeds this one?

paulgessinger: > So we found it twice, see there, comment 2:
> #1757
> It could have spared some @CarloVarni time if the other PR wold have been merged already.

Trying my best ✌️

tboldagh: Not sure. But maybe since mine is forgotten and this one got the wind we go ahead with this one? Either is fine.

paulgessinger:
We'll go with https://github.com/acts-project/acts/pull/1757 instead.

PR #1788, opened 2023-01-18
fix: Linkage errors in BetheHeitlerApprox.hpp [backport #1777 to develop/v20.x]

acts-project-service:
Backport a58b7985beb89cb20197aa311d16f45a7a5ef91f from #1777.
---


paulgessinger:
I bungled this, it's of course a backport v22 instead. I pushed an updated branch `develop/v22.x` that includes this PR.

PR #1791, opened 2023-01-20
refactor: Suppress large printout of Geant4

benjaminhuth:
This PR suppresses the large printout of physics parameters at the beginning of each Sequencer run by default.


Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
```diff
@@ -84,6 +87,13 @@ ActsExamples::Geant4Simulation::Geant4Simulation(
   G4EventManager::GetEventManager()->GetStackManager()->SetVerboseLevel(
       geantVerboseLevel);
 
+  // Suppress the printing of physics information.
+  if (m_cfg.printPhysicsParameters) {
+    G4HadronicParameters::Instance()->SetVerboseLevel(geantVerboseLevel);
+    G4HadronicProcessStore::Instance()->SetVerbose(geantVerboseLevel);
+    G4EmParameters::Instance()->SetIsPrintedFlag(true);
```
andiwand, line 94: looks like this does not compile. are you using a different G4 version?

andiwand (approved):
looks good but there seem to be some built issues

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
benjaminhuth: Probably, I mostly use very recent builds from G4... Hopefully there is some kind of `#ifdef` mechanism in G4 for different versions...

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
```diff
@@ -84,6 +88,15 @@ ActsExamples::Geant4Simulation::Geant4Simulation(
   G4EventManager::GetEventManager()->GetStackManager()->SetVerboseLevel(
       geantVerboseLevel);
 
+  // Suppress the printing of physics information.
+  if (m_cfg.printPhysicsParameters) {
+    G4HadronicParameters::Instance()->SetVerboseLevel(0);
```
andiwand, line 93: now clang tidy complains that this does not exist :disappointed: 

maybe we can just move everything into the version switch?

benjaminhuth: yeah lets do this

andiwand (approved):

PR #1792, opened 2023-01-23
fix: Let addGeant4 respect particle selection

Corentin-Allaire:
Fix the addGeant4 python function. With this fix the input of the geant4 simulation is changed to `particles_selected` which mean that the particle selector won't be ignored anymore.

### DIFF
--- a/Examples/Python/python/acts/examples/simulation.py
+++ b/Examples/Python/python/acts/examples/simulation.py
@@ -594,7 +594,7 @@ def addGeant4(
     g4conf = geant4SimulationConfig(
         level=customLogLevel(),
         detector=g4detectorConstruction,
-        inputParticles="particles_selected",
+        inputParticles=particles_selected,
         trackingGeometry=trackingGeometry,
         magneticField=field,
         volumeMappings=volumeMappings,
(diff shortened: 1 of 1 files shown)


Examples/Python/python/acts/examples/simulation.py
```diff
@@ -594,7 +594,7 @@ def addGeant4(
     g4conf = geant4SimulationConfig(
         level=customLogLevel(),
         detector=g4detectorConstruction,
-        inputParticles="particles_input",
+        inputParticles="particles_selected",
```
benjaminhuth, line 597: Actually, shouldn't it be the variable `particles_selected`, since this variable is populated correctly in the `if` clause in line 578?

benjaminhuth (changes requested):
This is really good that you caught this!

Examples/Python/python/acts/examples/simulation.py
benjaminhuth:
```suggestion
        inputParticles=particles_selected,
```

Corentin-Allaire: Good catch !

andiwand (approved):

benjaminhuth (approved):

Issue #1578, opened 2022-10-06
Segfault for ODD+Pythia8+Geant4


andiwand:
@benjaminhuth pointed out that ODD+Pythia8+Geant4 will segfault in full chain

I just verified this. See attached files for more information.

[segfault.txt](https://github.com/acts-project/acts/files/9723819/segfault.txt)
[full_chain.txt](https://github.com/acts-project/acts/files/9723821/full_chain.txt)

paulgessinger:
Hm, could be the two interfere?

asalzburger:
Is this executed in single thread ?

asalzburger:
No good ``numThreads=-1`` - this would need Geant4MT.

benjaminhuth:
For me this crashes with one thread as well

asalzburger:
Is that with the same error / segfault ?

benjaminhuth:
Hmm it looks a bit different to be honest, but couldn't check in detail for no.
I attache my gdb backtrace, maybe this gives a hint.

[backtrace.txt](https://github.com/acts-project/acts/files/9724203/backtrace.txt)

Corentin-Allaire:
As we discussed during today's meeting I tried to replace the ODD by the GDLM implementation of Alice_v3 and it ran through with just a few warning.
So the issue is either with the ODD itself or the ```DDG4DetectorConstruction```...

Corentin-Allaire:
I have had an other look at this and I just notice something. If I start removing the support from the ODD xml the segfault happen much later so maybe there is something bad with the support surface definition ?

paulgessinger:
For sure this is something that we need to fix.

benjaminhuth:
Okay, I have investigated this a bit and some new infos:

First of all, I enabled some logging facilities in Geant4, which gave me the result that this is caused by photons quite far away from the center in z direction (z is around 1e4):
![image](https://user-images.githubusercontent.com/37871400/213686049-a80ebfae-57ad-4931-a7aa-393a16dee381.png)

Then I also could reproduce the crash with the `ParticleGun`:

![image](https://user-images.githubusercontent.com/37871400/213686355-0a5115b7-8ecd-4f36-82f4-9c926d6e4564.png)

I'm not totally sure what to do with these information, but maybe someone has an idea :)

paulgessinger:
So it's G4 breaking in a specific region of the detector?

Corentin-Allaire:
[...]

PR #1794, opened 2023-01-24
feat: Add option to use odd with G4

Corentin-Allaire:
This add the option to use the G4 simulation with the Odd getting around the problem noticed in issue #1578.
Incidently it also fix a bug while using the odd full chain with only csv writting

### DIFF
--- a/Examples/Scripts/Python/full_chain_odd.py
+++ b/Examples/Scripts/Python/full_chain_odd.py
@@ -25,6 +25,7 @@ from common import getOpenDataDetectorDirectory
 from acts.examples.odd import getOpenDataDetector
 
 ttbar_pu200 = False
+g4_simulation = False
 u = acts.UnitConstants
 geoDir = getOpenDataDetectorDirectory()
 outputDir = pathlib.Path.cwd() / "odd_output"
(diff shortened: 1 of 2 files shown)


Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -73,19 +74,40 @@
...
+            ParticleSelectorConfig(
+                eta=(-3.0, 3.0), pt=(150 * u.MeV, None), removeNeutral=True
+            )
+            if ttbar_pu200
+            else ParticleSelectorConfig(),
+            outputDirRoot=outputDir,
+            rnd=rnd,
+        )
+    if g4_simulation:
+        if Sequencer.config.numThreads != 1:
```
andiwand, line 91:
```suggestion
        if s.config.numThreads != 1:
```

otherwise it would check on the default no?

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -73,19 +74,40 @@
...
+            field,
+            ParticleSelectorConfig(
+                eta=(-3.0, 3.0), pt=(150 * u.MeV, None), removeNeutral=True
+            )
+            if ttbar_pu200
+            else ParticleSelectorConfig(),
+            outputDirRoot=outputDir,
+            rnd=rnd,
+        )
+    if g4_simulation:
```
andiwand, line 90:
```suggestion
    else:
```

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -73,19 +74,40 @@
             rnd=rnd,
             outputDirRoot=outputDir,
         )
+    if not g4_simulation:
+        addFatras(
```
andiwand, line 78: I wonder if we should create a new helper like `addSimulation` and switch between fatras and g4 there because it is getting a little cumbersome. but that's something for a different PR

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -25,6 +25,7 @@
 from acts.examples.odd import getOpenDataDetector
 
 ttbar_pu200 = False
+g4_simulation = False
```
andiwand, line 28: should we also adopt the itk full chain?

Corentin-Allaire: I am pretty sure the ITk doesn't work with G4 (at least it didn't use to). But If I am wrong I would be more than happy to had it there

Examples/Scripts/Python/full_chain_odd.py
Corentin-Allaire: Fatras and G4 don't use the same argument at that point it would honestly be more cumbersome than helpful I think.

Examples/Scripts/Python/full_chain_odd.py
Corentin-Allaire: fixed

Examples/Scripts/Python/full_chain_odd.py
andiwand: but the only difference is the detector argument right now no?

Corentin-Allaire: One use a G4 detector,  the other use a tracking geometry. For a DD4Hep they look almost the same but for a different formt of detector (gdml, ...) it get more complex.

andiwand (approved):

PR #1796, opened 2023-01-24
fix: TGeo estimation of annulus bounds.

goetzgaycken:
fix: estimation of annulus bounds.
    
When estimating the annulus bounds from the TGeoArb8, also consider
a possible side flip in addition to the translation.


andiwand:
ready for more annulus bounds @niermann999 ? :smile:

niermann999:
> ready for more annulus bounds @niermann999 ? smile

Can't wait :P I must admit, I am not familiar with the geometry building from tGeo. Does this account for the flipped stereo angle in the backside modules?

paulgessinger (approved):

paulgessinger:
@niermann999 I think it's literally just applying the transform matrix on top of the vertices it extracted. The determination of the stereo angle and other defining parameters assumes the vertices are in the +xy plane, so I believe the rotation can be done on the output afterwards.

Plugins/TGeo/src/TGeoSurfaceConverter.cpp
```diff
@@ -147,14 +147,17 @@ Acts::TGeoSurfaceConverter::discComponents(const TGeoShape& tgShape,
...
+          // polyhedron, since there is a translation and
+          // also a side flip that needs to be applied.
+          // @TODO check that 3rd coordinate is not altered by
+          // the transformation ?
           std::vector<Vector2> vertices;
           for (unsigned int v = 0; v < 8; v += 2) {
-            Vector2 vtx = Vector2((polyTrl[0] + polyVrt[v + 0]) * scalor,
-                                  (polyTrl[1] + polyVrt[v + 1]) * scalor);
+            std::array<double, 3> local{polyVrt[v + 0], polyVrt[v + 1], 0.};
+            std::array<double, 3> global;
```
paulgessinger, line 158: clang-tidy would like you to initialize the array here:
```suggestion
            std::array<double, 3> global{};
```

goetzgaycken: Sure one can make the tools happy.

goetzgaycken:
> > ready for more annulus bounds @niermann999 ? smile
> 
> Can't wait :P I must admit, I am not familiar with the geometry building from tGeo. Does this account for the flipped stereo angle in the backside modules?

If I remember correctly, that's it.

niermann999 (approved):
Looks good to me

Plugins/TGeo/src/TGeoSurfaceConverter.cpp
```diff
@@ -147,14 +147,17 @@ Acts::TGeoSurfaceConverter::discComponents(const TGeoShape& tgShape,
...
+            maskTransform->LocalToMaster(local.data(), global.data());
+            Vector2 vtx = Vector2(global[0] * scalor, global[1] * scalor);
```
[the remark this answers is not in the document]
paulgessinger: yes

PR #1800, opened 2023-01-25
fix: Linkage errors in BetheHeitlerApprox.hpp [backport #1777 to develop/v21.1.x]

acts-project-service:
Backport a58b7985beb89cb20197aa311d16f45a7a5ef91f from #1777.
---


paulgessinger:
Failures are from intermediate fixes to the infra. I'm merging manually now.

PR #1801, opened 2023-01-25
refactor: Material mapping howto for python bindings 

Corentin-Allaire:
I have been procrastinating on this for way too long. In this PR, I update the material mapping howto to use the python bindings. The relevant bindings have been tested and updated to work as expected. I Also done some update of the plotting script (even if they should probably be phased out) to solve the remaining issue with them. 

In the near future I plan to update the material mapping autotuning, once it is done I will update the corresponding doc.

### DIFF
--- a/docs/howto/run_material_mapping.rst
+++ b/docs/howto/run_material_mapping.rst
@@ -30,16 +30,32 @@ Once Acts has been built we can start the mapping. The mapping is divided in two
 Mapping and configuration
 -------------------------
 
-First we need to extract the list of all the surfaces and volumes in our detector, to do so we will use the GeometryExample:
+First we need to extract the list of all the surfaces and volumes in our detector, to do so we will use the ``geometry.py`` script :
 
-.. code-block:: console
+.. code-block::
 
-   $ <build>/bin/ActsExampleGeometryDD4hep -n1 -j1 \
-       --mat-output-file geometry-map \
-       --dd4hep-input <source>/thirdparty/OpenDataDetector/xml/OpenDataDetector.xml \
-       --output-json \
-       --mat-output-allmaterial true \
-       --mat-output-sensitives false
+   $ python3 <source>/Examples/Scripts/Python/geometry.py 
+
+Ideally the following options should be use in the python file :
+
+.. code-block::
+
+   def runGeometry(
+      trackingGeometry,
+      decorators,
+      outputDir,
+      events=1,
+      outputObj=False,
+      outputCsv=False,
+      outputJson=True,
+      outputRoot=False,
+   ):
+
+For the following example we will be remapping the material of the ODD so we will get our detector via the following line :
+
+.. code-block::  console
+
+   detector, trackingGeometry, decorators = getOpenDataDetector(getOpenDataDetectorDirectory() )
 
 This algorithm is useful to obtain a visualisation of your detector using the different types of output available (``output-obj`` gives ``.obj`` with a 3D representation of the different subdetectors, for example). Here, we use ``output-json`` to obtain a map of all the surfaces and volumes in the detector with a ``ProtoSurfaceMaterial`` (or a ``ProtoVolumeMaterial``), ``mat-output-allmaterial`` ensure that a ``ProtoSurfaceMaterial`` (or a ``ProtoVolumeMaterial``) is associated to all the surfaces (or volumes), enforcing all of them to be written.
 Four types of surfaces exist:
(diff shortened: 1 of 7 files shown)


paulgessinger:
I think it'd be best if @noemina reviews this.

docs/howto/run_material_mapping.rst
```diff
@@ -30,16 +30,32 @@ Once Acts has been built we can start the mapping. The mapping is divided in two
 Mapping and configuration
 -------------------------
 
-First we need to extract the list of all the surfaces and volumes in our detector, to do so we will use the GeometryExample:
+First we need to extract the list of all the surfaces and volumes in our detector, to do so we will use the ``geometry.py`` script :
```
AJPfleger, line 33:
```suggestion
First we need to extract the list of all the surfaces and volumes in our detector. To do so we will use the ``geometry.py`` script:
```

docs/howto/run_material_mapping.rst
```diff
@@ -30,16 +30,32 @@ Once Acts has been built we can start the mapping. The mapping is divided in two
...
 
-.. code-block:: console
+.. code-block::
+
+   $ python3 <source>/Examples/Scripts/Python/geometry.py 
+
+Ideally the following options should be used in the python file :
```
AJPfleger, line 39:
```suggestion
Ideally the following options should be used in the python file:
```

docs/howto/run_material_mapping.rst
```diff
@@ -30,16 +30,32 @@ Once Acts has been built we can start the mapping. The mapping is divided in two
...
+      outputRoot=False,
+   ):
 
-   $ <build>/bin/ActsExampleGeometryDD4hep -n1 -j1 \
-       --mat-output-file geometry-map \
-       --dd4hep-input <source>/thirdparty/OpenDataDetector/xml/OpenDataDetector.xml \
-       --output-json \
-       --mat-output-allmaterial true \
-       --mat-output-sensitives false
+For the following example we will be remapping the material of the ODD, we will thus get our detector via the following line :
```
AJPfleger, line 54:
```suggestion
For the following example we will be remapping the material of the ODD, we will thus get our detector via the following line:
```

docs/howto/run_material_mapping.rst
```diff
@@ -239,4 +232,4 @@ Can be use with X,Y,Z is a list of volumes, this will plot the material ratio be
 Using a different detector
 --------------------------
 
-If you want to use a different type of detector, you will first need to ensure that the relevant packages were added during the compilation. After that, if your detector is a DD4hep detector you will just need to replace the path given to the ``--dd4hep-input`` option. In case it is another type of detector implementation, you can replace DD4hep in the name of the algorithm by what corresponds to your detector implementation. For more information on how to include your detector in that case you can refer to the documentation of the algorithm using the ``-h`` option.
+If you want to use a different type of detector, you will first need to ensure that the relevant packages were added during the compilation. After this you can just replace the detector initialisation in the different main function. For reference you can base your self on the Odd for DD4Hep detector and on the ITk for TGeo detector.
```
AJPfleger, line 235:
```suggestion
If you want to use a different type of detector, you will first need to ensure that the relevant packages were added during the compilation. After this you can just replace the detector initialisation in the different main function. For reference you can have a look on the Odd for DD4Hep detector and on the ITk for TGeo detector. 
```

AJPfleger (commented):
just a few points for readability in the docs :)

docs/howto/run_material_mapping.rst
```diff
@@ -126,48 +142,32 @@ The same script can be used to dump a steering file that can help you selecting
...
 Depending on what you want to do there are three options you can change:
 
-- ``mat-mapping-surfaces``: determine if material is mapped onto surfaces
-- ``mat-mapping-volumes``: determine if material is mapped onto volumes
-- ``mat-mapping-volume-stepsize``: determine the step size used in the sampling of the volume. This should be small compared to the bin size.
-
+- ``mapSurface``: determine if material is mapped onto surfaces
+- ``mapVolume``: determine if material is mapped onto volumes
+- ``mappingStep``: determine the step size used in the sampling of the volume. This should be small compared to the bin size.
+- ``readCachedSurfaceInformation`` if added the material-surface association will be taken from the input material track file (doesn't work with geantino file, you need to use the material track file obtained from running the material mapping).
```
noemina, line 170: Considering you are mentioning the options used in the material mapping, can you also document here what the `mappingStep` refers to?

noemina:
Hello, it seems pretty straightforward. I only have one comment on the documentation.

docs/howto/run_material_mapping.rst
Corentin-Allaire: Sure I can go into a bit more detail

Corentin-Allaire: Does this make sense to you : 

- ``mappingStep``: determine the step size used in the sampling of the volume. The material interaction point obtained though G4 are only at the first intersection between the track and the material, this means that the mapping will be incorrect if the material is deeper than the bin size. To avoid this, we create new additional material point spaced equally every ``mappingStep`` along the trajectory. This should be small compared to the bin size.

noemina: Does this parameter then make sense only for mapping onto volumes?
This is my suggestion, feel free to edit it further:
By default, the material interaction point obtained from G4 is accumulated at the intersection between the track and the volume material. The mapping will be therefore incorrect if the material extends through the bin. To avoid this, additional material points are created every mappingStep [mm] along the trajectory. The mapping step should be small compared to the bin size.

Corentin-Allaire: Exactly it is only used in the volume mapping. The surface mapper doesn't use this parameters. Thanks for the  suggestion !

Corentin-Allaire: I have updated the doc. It seems that now we have an issue in the CI related to the codding guideline page.

noemina (approved):
All good to me!

AJPfleger (approved):

PR #1803, opened 2023-01-27
fix: G4 material converter

andiwand:
nucleus charge VS effective mass

somehow there is no nice solution to get the relative atomic mass in G4 so I had to go through the element class

### DIFF
--- a/Plugins/Geant4/src/Geant4Converters.cpp
+++ b/Plugins/Geant4/src/Geant4Converters.cpp
@@ -22,6 +22,8 @@
 #include "Acts/Surfaces/TrapezoidBounds.hpp"
 #include "Acts/Utilities/Helpers.hpp"
 
+#include <cassert>
+
 #include "G4Box.hh"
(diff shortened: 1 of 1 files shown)


benjaminhuth (approved):

Plugins/Geant4/src/Geant4Converters.cpp
```diff
@@ -301,12 +303,14 @@ std::shared_ptr<Acts::HomogeneousSurfaceMaterial>
 Acts::Geant4MaterialConverter::surfaceMaterial(const G4Material& g4Material,
                                                ActsScalar original,
                                                ActsScalar compressed) {
+  assert(g4Material.GetNumberOfElements() == 1);
```
asalzburger, line 306: In fact, if is a compound, one would have to do the averaging ...

andiwand: followed up on this here https://github.com/acts-project/acts/pull/1805

PR #1804, opened 2023-01-27
refactor: move material composition code to the left

andiwand:
early exit and whitespace cleanup

picked up on this file because it crashed for me when `A >= 100`

changes are not critical so we can also drop it

### DIFF
--- a/Examples/Scripts/MaterialMapping/materialComposition.C
+++ b/Examples/Scripts/MaterialMapping/materialComposition.C
@@ -60,6 +60,13 @@ struct MaterialHistograms {
 
   MaterialHistograms() = default;
 
+  ~MaterialHistograms() {
+    delete x0_vs_eta;
+    delete l0_vs_eta;
+    delete x0_vs_phi;
+    delete l0_vs_phi;
+  }
+
   /// This fills the event into the histograms
   /// and clears the cache accordingly
   ///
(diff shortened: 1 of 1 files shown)


Examples/Scripts/MaterialMapping/materialComposition.C
```diff
@@ -108,128 +108,131 @@ void materialComposition(const std::string& inFile, const std::string& treeName,
...
+  // Draw all the atomic elements & get the histogram
+  inputTree->Draw("mat_A>>hA(100,0.5,100.5)");
+  TH1F* histA = dynamic_cast<TH1F*>(gDirectory->Get("hA"));
+  histA->Draw();
+
+  auto outputFile = TFile::Open(outFile.c_str(), "recreate");
+
+  float v_eta = 0;
+  float v_phi = 0;
+  std::vector<float>* stepLength = new std::vector<float>;
```
benjaminhuth, line 128: Are those vectors ever properly destructed? However, since this is not really part of the PR, maybe not that important...

andiwand: I fear not and there seem to be a few places in this code were we do not care about deallocation :skull: I would leave it as it is for now

benjaminhuth (approved):

PR #1805, opened 2023-01-27
feat: G4 plugin material mixing

andiwand:
copied from here https://github.com/acts-project/acts/blob/main/Examples/Algorithms/Geant4/src/MaterialSteppingAction.cpp#L61


asalzburger (approved):
That's how it should be.

PR #1809, opened 2023-01-30
feat: KalmanFitter marks missed surfaces as outliers

paulgessinger:
We already walk all states and check if they were passed again (to remove the smoothed parameters), we can use the same spot to mark as outliers


andiwand (approved):

paulgessinger:
@andiwand somehow the review got dismissed, can you try again?

andiwand (approved):

PR #1810, opened 2023-01-30
feat: mark state as outlier if reverse filtering missed the surface

CarloVarni:
As the title says, marking with `TrackStateFlag::Outlier` if the reverse filtering has missed the surface and there are no smoothed parameters


paulgessinger:
Ah, see #1809

CarloVarni:
> Ah, see #1809

closing this one then

PR #1812, opened 2023-01-30
refactor: Remove LoggerWrapper in favor of Logger refs

paulgessinger:
To harmonize / simplify the ways that loggers are used, this PR removes `LoggerWrapper`. To support the case of e.g. functions optionally accepting a logger, we can use `const Logger&` and default it to `getDummyLogger`. Since `getDummyLogger` returns a reference to a logger instance of static lifetime, all the arguments should be safe, even if they are retained (I think).


Alignment/include/ActsAlignment/Kernel/Alignment.hpp
```diff
@@ -142,7 +136,10 @@ struct Alignment {
   Alignment() = delete;
 
   /// Constructor from arguments
-  Alignment(fitter_t fitter) : m_fitter(std::move(fitter)) {}
+  Alignment(fitter_t fitter,
+            std::unique_ptr<const Acts::Logger> _logger =
+                Acts::getDefaultLogger("Alignment", Acts::Logging::INFO))
+      : m_fitter(std::move(fitter)), m_logger{std::move(_logger)} {}
```
andiwand, line 142:
```suggestion
            std::unique_ptr<const Acts::Logger> logger =
                Acts::getDefaultLogger("Alignment", Acts::Logging::INFO))
      : m_fitter(std::move(fitter)), m_logger{std::move(logger)} {}
```

Core/include/Acts/EventData/detail/CorrectedTransformationFreeToBound.hpp
```diff
@@ -91,7 +91,7 @@ struct CorrectedFreeToBoundTransformer {
       const FreeVector& freeParams, const FreeSymMatrix& freeCovariance,
       const Surface& surface, const GeometryContext& geoContext,
       NavigationDirection navDir = NavigationDirection::Forward,
-      LoggerWrapper logger = getDummyLogger()) const;
+      const Logger& logger = getDummyLogger()) const;
```
andiwand, line 94: I guess in the best case we never end up with a dummy logger but it gets always passed down from the user, right?

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -198,22 +198,17 @@ namespace detail {
...
 
   const bool coCriterion = cLimit > oLimit;
   const bool cpCriterion = std::abs(cLimit) < std::abs(pLimit) + tolerance;
 
   const bool accept = coCriterion and cpCriterion;
 
-  if constexpr (doLogging) {
     if (accept) {
       ACTS_VERBOSE("Intersection is WITHIN limit");
     } else {
```
andiwand, line 214: I guess this and the following should be moved to the left?

Examples/Algorithms/TrackFitting/src/TrackFittingAlgorithm.cpp
```diff
@@ -74,13 +74,9 @@ ActsExamples::ProcessCode ActsExamples::TrackFittingAlgorithm::execute(
...
-  GeneralFitterOptions options{ctx.geoContext,
-                               ctx.magFieldContext,
-                               ctx.calibContext,
-                               calibrator,
-                               &(*pSurface),
-                               Acts::LoggerWrapper{logger()},
-                               Acts::PropagatorPlainOptions()};
+  GeneralFitterOptions options{
+      ctx.geoContext, ctx.magFieldContext, ctx.calibContext,
+      calibrator,     &(*pSurface),        Acts::PropagatorPlainOptions()};
```
andiwand, line 79: I think this needs to be reformatted

Plugins/DD4hep/include/Acts/Plugins/DD4hep/ConvertDD4hepMaterial.hpp
```diff
@@ -27,10 +27,10 @@ class Layer;
 /// @param detElement the DD4hep detector element for which this material is
 ///                   assigned
 /// @param cylinderLayer is the target layer
-/// @param logger a @c LoggerWrapper for output
+/// @param logger a @c Logger  for output
```
andiwand, line 30:
```suggestion
/// @param logger a @c Logger for output
```

andiwand (approved):
looks good! thanks for this :+1:

Alignment/include/ActsAlignment/Kernel/Alignment.hpp
paulgessinger: I think this shadows the `const Logger& logger() const;` method, no?

Core/include/Acts/EventData/detail/CorrectedTransformationFreeToBound.hpp
paulgessinger: That's what I'm thinking yeah. The dummy logger is a tradeoff where we default to not logging at all. That's preferable over constructing a local logger in the function body (which wasn't actually the case in this specific instance before this PR, anyway)

Core/include/Acts/Utilities/Intersection.hpp
paulgessinger: Huh, clang-tidy doesn't complain?

paulgessinger: Nevermind, it does.

Examples/Algorithms/TrackFitting/src/TrackFittingAlgorithm.cpp
paulgessinger: Seems like clang-format wants it this way, for some reason.

andiwand (approved):

andiwand (approved):

PR #1815, opened 2023-01-31
fix: bug in calculation of cut for seed finder orthogonal

LuisFelipeCoelho:
Due to @tboldagh's recent changes (#1757) and the significant reduction in the number of seeds produced by the orthogonal seeding, I noticed an error in the calculation of interactionPointCut that resulted in a reduction in efficiency for |eta| > 3 in the ITk examples. This PR fixes the problem.

Before:
ITkSeeding_ttbar_mu0_PPP_trackeff_vs_eta copy
After:
ITkSeeding_ttbar_mu0_PPP_trackeff_vs_eta


stephenswat (approved):
Excellent, great job Luis! :smile:

tboldagh:
Very good finding @LuisFelipeCoelho 
With physics hat I am slightly worried about the overall logic. I.e. should the algorithm depend on absolute value of z at all?
I am eluding to a thing like reference frame independence.

LuisFelipeCoelho:
@tboldagh as far as I understand (@stephenswat please correct me if I am wrong) the orthogonal seeding is dependent on whether the track has monotonically increasing or decreasing z position. This factor accounts for that.

PR #1816, opened 2023-02-01
feat: collapse material tracks

andiwand:
- option to collapse material tracks
- fix and simplify material composition script
- fix typos


Examples/Io/Root/include/ActsExamples/Io/Root/RootMaterialTrackWriter.hpp
```diff
@@ -56,6 +56,9 @@ class RootMaterialTrackWriter
     bool storeSurface = false;
     /// Write the volume to which the material step correpond
     bool storeVolume = false;
+    /// Collapse consecutive interactions of a single surface into a sinle
+    /// interaction
```
AJPfleger, line 60:
```suggestion
    /// Collapse consecutive interactions of a single surface into a single
    /// interaction
```

Examples/Scripts/MaterialMapping/materialComposition.C
```diff
@@ -58,14 +60,16 @@ struct MaterialHistograms {
...
+  // yes we might leak memory here but this would break copy (and potentially
+  // move) and I dont want to implement them
+  /*
   ~MaterialHistograms() {
     delete x0_vs_eta;
     delete l0_vs_eta;
     delete x0_vs_phi;
     delete l0_vs_phi;
   }
+  */
```
AJPfleger, line 72: Could you give an estimate an estimate on how much memory we might be losing?

And maybe we should also add this as a warning to the PR-description (and open an issue) - or better: find someone who wants to implement them.

AJPfleger (changes requested):

Examples/Scripts/MaterialMapping/materialComposition.C
benjaminhuth: simple solution would be to make them to `shared_ptr` (or `unique_ptr` if copy is not needed). That gives you everything for free.

andiwand: but do you actually do that with root objects?

also: paul mentioned that the profiles might be owned by the canvas?

benjaminhuth: Actually its worth to read once through https://root.cern/manual/object_ownership/

And exactly, all histograms are owned by default by the current `gDirectory` object of root. Can be disabled, but in our case it should be just fine to remove the constructor.

benjaminhuth: I actually would remove the comment and the destructor completely, since there should be no memory leak.

benjaminhuth (approved):
Let's go!

PR #1820, opened 2023-02-02
refactor: Seeding does not use a grid of pointers

paulgessinger:


LuisFelipeCoelho:
It looks good to me, only need to change cuda and sycl

PR #1821, opened 2023-02-02
feat!: new cubic volume building

asalzburger:
This PR introduces a new `CubicVolumeHelper` and aligns the interface of `ITrackingVolumeHelper` to use `Extent` rather than specified `locZ, locR`, which does not make sense in an interface.

It also allows the KDTreeGeometryBuilder to build telescope like detectors if configured to do so.

Screenshot 2023-02-02 at 15 19 22


asalzburger:
This PR will come with a new test detector `NA60+`

PR #1823, opened 2023-02-03
refactor: GX2F uses TrackContainer + log message cleanup

paulgessinger:
Contrary to what I said before, the GX2F was not yet using `Acts::TrackContainer`. This PR switches it over. 
I also cleaned up some of the log messages in the Chi2 fitter and algorithm, the prefixes there are unnecessary because the loggers already print their own names.


AJPfleger (approved):

asalzburger:
This has been approved but is not up2date: adding automerge flag.

PR #1826, opened 2023-02-03
refactor: no optional covariance in GSF

benjaminhuth:
The GSF in principle worked with `std::optional<BoundSymMatrix>` for the covariance in almost all cases. However, it does not make any sense to run the GSF without covariance, and removing this does also make the code simpler.
Actually, it was assumed to have a value in almost all cases, so this should also make the code a bit saver.


Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
```diff
@@ -37,8 +37,10 @@ auto MultiEigenStepperLoop<E, R, A>::boundState(
                                           transportCov, freeToBoundCorrection);
 
       if (bs.ok()) {
-        states.push_back(
-            {state.components[i].weight, std::get<BoundTrackParameters>(*bs)});
+        const auto& btp = std::get<BoundTrackParameters>(*bs);
+        states.emplace_back(
+            state.components[i].weight, btp.parameters(),
+            btp.covariance().value_or(Acts::BoundSymMatrix::Zero()));
```
andiwand, line 43: will a zero covariance play nice later on?

andiwand (approved):
looks good. only left one question

andiwand:
@benjaminhuth I just saw this is WIP. let me know when I should re-review it

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
benjaminhuth: Hmm I thought if everything is consistent, `covTransport = true` should mean there is also a covariance there to get... However, I think this is not checked...
Maybe worth thinking about how to ensure this does not go to the outside.

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
```diff
@@ -98,14 +101,27 @@ auto MultiEigenStepperLoop<E, R, A>::curvilinearState(State& state,
       pos4 += state.components[i].weight * cp.fourPosition(state.geoContext);
       dir += state.components[i].weight * cp.unitDirection();
       qop += state.components[i].weight * (cp.charge() / cp.absoluteMomentum());
-      if (cp.covariance()) {
+      if( cp.covariance() ) {
```
andiwand, line 104: something went wrong with the formatting here

andiwand (approved):

andiwand (approved):
:+1:

Issue #1825, opened 2023-02-03
Wrong template order in Clusterization.ipp


gagnonlg:
In createClusters, the labelClusters function is called like so: 
```
labelClusters<CellCollection, Connect, GridDim>(cells, connect);
```
However, the labelClusters function is defined like so:
```
template <typename CellCollection, size_t GridDim = 2,
          typename Connect =
              DefaultConnect<typename CellCollection::value_type, GridDim>>
void labelClusters(CellCollection& cells, Connect connect = Connect());
```

Therefore the order is wrong in the createClusters wrapper. However, createClusters is called in the unit tests which compile and run, so I need to investigate what is happening, and if there's at all a bug lurking here.

gagnonlg:
As noted by @CarloVarni there was another ordering mistake in the .ipp where the "real" labelClusters is defined. The two mistakes aligned which is why this wasn't caught by the compiler.

paulgessinger:
Does this not get compiled anywhere? I thought we had a unit test cpp where the template is instantiated?

gagnonlg:
Yes it gets compiled in the tests. But, the `createClusters` wrapper *and* the definition of `labelClusters` in the .ipp both had the wrong ordering of template parameters, so it compiled anyways. I intended to have the GridDim parameter before the Connect type because the latter can be derived based on the former, which is what was fixed here.

gagnonlg:
Fixed by #1827

PR #1827, opened 2023-02-05
fix: change templates order for labelClusters

CarloVarni:
Connected to https://github.com/acts-project/acts/issues/1825 

Not sure if this closes that issue or it needs more investigation by @gagnonlg , but without this PR I cannot compile clustering with  `v23` in Athena

/cc @paulgessinger


gagnonlg:
Looks good to me, thanks @CarloVarni. So there was a mismatch between the declaration in the .hpp and .ipp. I would have thought it would have been caught by the compiler but seems like it saw it as an overload.

gagnonlg (approved):

gagnonlg:
Closes #1825

PR #1828, opened 2023-02-05
fix: Move GeoIdHook implementation to hpp

CarloVarni:
Related to https://github.com/acts-project/acts/issues/1829 

This is a quick bypass of the compilation issue in Athena, in order to move to tag `v23`

### DIFF
--- a/Core/include/Acts/Geometry/GeometryIdentifier.hpp
+++ b/Core/include/Acts/Geometry/GeometryIdentifier.hpp
@@ -137,7 +137,8 @@ std::ostream& operator<<(std::ostream& os, GeometryIdentifier id);
 struct GeometryIdentifierHook {
   virtual ~GeometryIdentifierHook() = default;
   virtual Acts::GeometryIdentifier decorateIdentifier(
-      Acts::GeometryIdentifier identifier, const Acts::Surface& surface) const;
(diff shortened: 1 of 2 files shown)


andiwand (approved):

Core/include/Acts/Geometry/GeometryIdentifier.hpp
```diff
@@ -137,7 +137,10 @@ std::ostream& operator<<(std::ostream& os, GeometryIdentifier id);
 struct GeometryIdentifierHook {
   virtual ~GeometryIdentifierHook() = default;
   virtual Acts::GeometryIdentifier decorateIdentifier(
-      Acts::GeometryIdentifier identifier, const Acts::Surface& surface) const;
+      Acts::GeometryIdentifier identifier,
+      const Acts::Surface& /*surface*/) const {
+    return identifier;
+  }
```
andiwand, line 143: it looks odd that we have to move this into the header. do you know why this is necessary? are we missing the compiled source file during linking?

andiwand (approved):

Core/include/Acts/Geometry/GeometryIdentifier.hpp
paulgessinger: This confuses me as well. It's unclear to me why the symbol would not be visible.

PR #1831, opened 2023-02-06
refactor: Change transformCoordinates to avoid recalculation of variables in Seedfinder

LuisFelipeCoelho:
This PR tries to optimise SeedFinder by removing unnecessary calculations


Core/include/Acts/Seeding/SeedFinderUtils.hpp
```diff
@@ -46,12 +46,32 @@ LinCircle transformCoordinates(external_spacepoint_t& sp,
...
+template <typename external_spacepoint_t>
+LinCircle transformCoordinates(InternalSpacePoint<external_spacepoint_t>& sp,
+                               InternalSpacePoint<external_spacepoint_t>& spM,
+                               std::array<float, 6> transformVariables);
+
+template <typename external_spacepoint_t, typename callable_t>
+LinCircle transformCoordinates(external_spacepoint_t& sp,
+                               external_spacepoint_t& spM,
+                               callable_t&& extractFunction,
+                               std::array<float, 6> transformVariables);
```
CarloVarni, line 67:
```suggestion
                               const std::array<float, 6>& transformVariables);
```

Core/include/Acts/Seeding/SeedFinderUtils.ipp
```diff
@@ -50,14 +50,73 @@ LinCircle transformCoordinates(external_spacepoint_t& sp,
...
           (cot_theta * cot_theta) * (varianceRM + varianceRSP)) *
          iDeltaR2;
   return l;
 }
 
+template <typename external_spacepoint_t>
+LinCircle transformCoordinates(
+    InternalSpacePoint<external_spacepoint_t>& sp,
+    const InternalSpacePoint<external_spacepoint_t>& spM,
+    const std::array<float, 6> transformVariables) {
```
CarloVarni, line 65:
```suggestion
    const std::array<float, 6>& transformVariables) {
```

Core/include/Acts/Seeding/SeedFinderUtils.ipp
```diff
@@ -50,14 +50,73 @@ LinCircle transformCoordinates(external_spacepoint_t& sp,
...
+  auto extractFunction =
+      [](const InternalSpacePoint<external_spacepoint_t>& obj)
+      -> std::array<float, 6> {
+    std::array<float, 6> output{obj.x(),      obj.y(),         obj.z(),
+                                obj.radius(), obj.varianceR(), obj.varianceZ()};
+    return output;
```
CarloVarni, line 71: maybe we can avoid lvalues here
```suggestion
    return {obj.x(),      obj.y(),         obj.z(),
                 obj.radius(), obj.varianceR(), obj.varianceZ()};
```

Core/include/Acts/Seeding/SeedFinderUtils.ipp
```diff
@@ -50,14 +50,73 @@ LinCircle transformCoordinates(external_spacepoint_t& sp,
...
+  };
+
+  return transformCoordinates<InternalSpacePoint<external_spacepoint_t>>(
+      sp, spM, extractFunction, transformVariables);
```
CarloVarni, line 75: would `sp, spM, extractFunction, std::move(transformVariables));` be faster?

Core/include/Acts/Seeding/SeedFinderUtils.ipp
```diff
@@ -50,14 +50,73 @@ LinCircle transformCoordinates(external_spacepoint_t& sp,
...
+}
+
+template <typename external_spacepoint_t, typename callable_t>
+LinCircle transformCoordinates(external_spacepoint_t& sp,
+                               const external_spacepoint_t& spM,
+                               callable_t&& extractFunction,
+                               const std::array<float, 6> transformVariables) {
```
CarloVarni, line 82:
```suggestion
                               const std::array<float, 6>& transformVariables) {
```

Core/include/Acts/Seeding/SeedFinderUtils.ipp
```diff
@@ -50,14 +50,73 @@ LinCircle transformCoordinates(external_spacepoint_t& sp,
...
+  // The computation inside this function is exactly identical to that in the
+  // vectorized version of this function, except that it operates on a single
+  // spacepoint. Please see the other version of this function for more
+  // detailed comments.
+
+  auto [xM, yM, zM, rM, varianceRM, varianceZM] = extractFunction(spM);
```
CarloVarni, line 88: I suppose this is the main issue... for every time we call this function we allocate this memory and then de-allocate. right?

Core/include/Acts/Seeding/SeedFinderUtils.ipp
```diff
@@ -120,24 +179,42 @@ std::vector<std::size_t> transformCoordinates(
...
-    sp->setDeltaR(std::sqrt((x * x) + (y * y) + (deltaZ * deltaZ)));
+    sp->setDeltaR(std::sqrt((xNewFrame * xNewFrame) + (yNewFrame * yNewFrame) +
+                            (deltaZ * deltaZ)));
+  }
+  //   sort the SP in order of cotTheta
+  std::sort(
+      indexes.begin(), indexes.end(),
+      [&linCircleVec](const std::size_t& a, const std::size_t& b) -> bool {
+        return linCircleVec[a].cotTheta < linCircleVec[b].cotTheta;
+      });
```
CarloVarni, line 205: what if we move this sorting outside of this function? I'd find it cleaner tbh

Core/include/Acts/Seeding/SeedFinderUtils.ipp
LuisFelipeCoelho: Thank you for your comments. I agree this is a problem. I could remove extractFunction entirely and only call for the variances which are the only values that we need, do you think this would be better?

LuisFelipeCoelho: yes, I agree, but these function calls are no different from what happens in Athena, so I am not sure if that explains the time difference

LuisFelipeCoelho:
To avoid reallocating memory at each iteration, I added a reserve statement to the sp vectors for the duplets: 

```
// reserve enough capacity
size_t dupletCapacity =
      isBottom ? m_config.bottomDupletCapacity : m_config.topDupletCapacity;
outVec.reserve(dupletCapacity); 
linCircleVec.reserve(dupletCapacity);
```

the dupletCapacity is a configurable variable that I set separately for bottom and top to 200 and 450, based on the maximum number of SPs after the cuts for ttbar <mu>=200 events:

Screenshot 2023-02-15 at 17 19 31

Screenshot 2023-02-15 at 17 19 23

The time difference per event with and without reserving the vector capacity does not seem to be big:

compare

@noemina @CarloVarni

LuisFelipeCoelho:
@Corentin-Allaire thank you for reviewing this. 
It should always be `doublet`, that is my mistake, I will fix it

Core/include/Acts/Seeding/SeedFinder.hpp
```diff
@@ -115,8 +113,8 @@ class SeedFinder {
   void getCompatibleDoublets(
```
Corentin-Allaire, line 113: Even if this is private I think it wouldn't be a bad idea to had comments to define what are the function input, output and what it does.

Core/include/Acts/Seeding/SeedFinderUtils.hpp
```diff
@@ -37,36 +35,60 @@ struct LinCircle {
...
-std::vector<std::size_t> transformCoordinates(
-    std::vector<external_spacepoint_t*>& vec, const external_spacepoint_t& spM,
-    bool bottom, std::vector<LinCircle>& linCircleVec,
-    callable_t&& extractFunction);
+void transformCoordinates(std::vector<external_spacepoint_t*>& vec,
+                          const external_spacepoint_t& spM, bool bottom,
+                          std::vector<LinCircle>& linCircleVec,
+                          callable_t&& extractFunction);
+
+/// @brief Returns a sorted vector of indeces based on cotangent of theta of a vector of spacepoints
```
Corentin-Allaire, line 81:
```suggestion
/// @brief Returns a sorted vector of indices based on cotangent of theta of a vector of spacepoints
```

Corentin-Allaire (changes requested):
Looks great to me ! I have added a few comments but the rest looks great. Did you measure what kind of impact this has on the seeding speed ?

Core/include/Acts/Seeding/SeedFinderUtils.ipp
```diff
@@ -9,7 +9,7 @@
 namespace Acts {
 template <typename external_spacepoint_t>
 inline LinCircle transformCoordinates(
-    const InternalSpacePoint<external_spacepoint_t>& sp,
+    InternalSpacePoint<external_spacepoint_t>& sp,
```
CarloVarni, line 12: what is the reason for this change?

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -143,18 +145,28 @@ template <typename sp_range_t, typename out_range_t>
...
   const int sign = isBottom ? -1 : 1;
 
   outVec.clear();
+  linCircleVec.clear();
+
+  // reserve enough capacity
+  size_t dubletCapacity =
+      isBottom ? m_config.bottomDubletCapacity : m_config.topDubletCapacity;
+  outVec.reserve(dubletCapacity);
+  linCircleVec.reserve(dubletCapacity);
```
CarloVarni, line 159: I would not go for this approach here... This is highly arbitrary since it depends on what kind of event you have in your hands. 

Also, we already know that the maximum capacity of these vectors is the size of `otherSPs`, right?

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -170,36 +182,47 @@ void SeedFinder<external_spacepoint_t, platform_t>::getCompatibleDoublets(
       continue;
     }
 
+    // if absolute value of z-distance is too big, try next SP in bin
     const float zO = otherSP->z();
-    float deltaZ = sign * (zO - zM);
+    float deltaZAbs = zO - zM;
```
CarloVarni, line 187: this can be negative. is it ok?

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -241,20 +269,11 @@ void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
   float varianceRM = spM.varianceR();
   float varianceZM = spM.varianceZ();
 
-  state.linCircleBottom.clear();
-  state.linCircleTop.clear();
+  auto sorted_bottoms =
+      cotThetaSortIndex(state.compatBottomSP, state.linCircleBottom);
+  auto sorted_tops = cotThetaSortIndex(state.compatTopSP, state.linCircleTop);
```
CarloVarni, line 274: from what I see (am I missing it?) the `cotThetaSortIndex` function makes no use of the first argument? Also, why not simply pass the `state` directly instead of having a list of variables?

Core/include/Acts/Seeding/SeedFinderUtils.hpp
```diff
@@ -38,35 +36,59 @@ struct LinCircle {
 /// @param[in] bottom Should be true if sp is a bottom SP.
 template <typename external_spacepoint_t>
 LinCircle transformCoordinates(
-    const InternalSpacePoint<external_spacepoint_t>& sp,
+    InternalSpacePoint<external_spacepoint_t>& sp,
     const InternalSpacePoint<external_spacepoint_t>& spM, bool bottom);
 
 template <typename external_spacepoint_t, typename callable_t>
-LinCircle transformCoordinates(const external_spacepoint_t& sp,
+LinCircle transformCoordinates(external_spacepoint_t& sp,
```
CarloVarni, line 43: why this change? we are actually trying to go towards constness

LuisFelipeCoelho:
@CarloVarni thank you for the comments. This improves around 15% for PPP but doesn't change much for SSS because m_config.interactionPointCut is disabled for SSS.

To avoid recalculation of some variables I moved the calculation in transform coordinates to getCompatibleDoublets, which is not the cleanest option. Do you have any suggestions?

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -306,15 +368,15 @@ inline void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
   float varianceRM = spM.varianceR();
   float varianceZM = spM.varianceZ();
 
-  state.linCircleBottom.clear();
-  state.linCircleTop.clear();
+  // state.linCircleBottom.clear();
+  // state.linCircleTop.clear();
```
CarloVarni, line 372: remove

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -306,15 +368,15 @@ inline void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
...
   std::size_t numTopSP = state.compatTopSP.size();
 
-  transformCoordinates(state.spacePointData, state.compatBottomSP, spM, true,
-                       state.linCircleBottom);
-  transformCoordinates(state.spacePointData, state.compatTopSP, spM, false,
-                       state.linCircleTop);
+  // transformCoordinates(state.spacePointData, state.compatBottomSP, spM,
+  //     true,state.linCircleBottom);
+  // transformCoordinates(state.spacePointData, state.compatTopSP, spM, false,
+  //                       state.linCircleTop);
```
CarloVarni, line 379: remove

CarloVarni: ok, we don't use it in the default seeding anymore

PR #1833, opened 2023-02-06
fix: GSF SourceLink reference

Maury98:
Building Athena (ATLAS' software platform) using ACTS v22.0.1 as externals and calling the GSF fitter lead to a fatal crash during compilation. This issue was fixed by explicitly using the SourceLink rather than its reference wrapper when storing the input measurements.

This PR should be ready to be merged.

Thanks to Hadrien Grasland for his help.

### DIFF
--- a/Core/include/Acts/TrackFitting/GaussianSumFitter.hpp
+++ b/Core/include/Acts/TrackFitting/GaussianSumFitter.hpp
@@ -236,7 +236,7 @@ struct GaussianSumFitter {
         inputMeasurements;
     for (auto it = begin; it != end; ++it) {
       const SourceLink& sl = *it;
-      inputMeasurements.emplace(sl.geometryId(), *it);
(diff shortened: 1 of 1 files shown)


paulgessinger:
Looks good in principle. I'm a bit confused why this didn't break in our ci.

andiwand:
maybe @benjaminhuth wants to take a final look. looks good from my side

benjaminhuth:
It confuses me as well that the compilation breaks. For me both versions look semantical identical, or do I miss something? However, if it fixes it, let's get it in!

@Maury98 Could you maybe provide the compiler error for more insight?

@paulgessinger Hmm as I understand the compilation fails, and a issues with the addresses would be a runtime-failure, right?

benjaminhuth (approved):

benjaminhuth:
Ah and @paulgessinger I think you need to add @Maury98 to the CI pipeline allow-list, right?

PR #1834, opened 2023-02-07
feat: configurable charge smearing in geometric digitization & exact digitization

benjaminhuth:
This makes path smearing configurable in the geometric digitization config like this:

```json
"charge-smearing": {
  "mean" : 0.0, "stddev" : 0.000, "type" : "Gauss"
}
```

For the charge smearing the `ActsFatras::SingleParameterSmearer<Rng>` is reused.

It also adds an `Exact` smearer, that does not smear, which is useful for debugging I think:

```json
"charge-smearing": {
  "type" : "Exact"
}
```

This can also be used in the smearing digitization to make a truth-digitization:
```json
{
    "volume" : 16,
    "value" : {
        "smearing" : [
          {"index" : 0, "type" : "Exact"},
          {"index" : 1, "type" : "Exact"}
        ]
    }
}
```

### DIFF
--- a/Examples/Algorithms/Digitization/include/ActsExamples/Digitization/DigitizationConfig.hpp
+++ b/Examples/Algorithms/Digitization/include/ActsExamples/Digitization/DigitizationConfig.hpp
@@ -56,10 +56,22 @@ struct GeometricConfig {
   };
   double thickness = 0.;
   /// Charge generation
-  ChargeGenerator charge = [](Acts::ActsScalar path, Acts::ActsScalar,
-                              RandomEngine &) -> Acts::ActsScalar {
-    return path;
-  };
+  ActsFatras::SingleParameterSmearFunction<ActsExamples::RandomEngine>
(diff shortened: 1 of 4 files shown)


benjaminhuth:
Actually I'm not sure about something in the digitization generally:

There is an `operator==` defined for Digitization-Config-objects, however, it does not check if the smearer function is equal. I *kind of* refactored this so it should work, but probably this is pretty slow.
As I think this is an optimization attempt to compare configurations and reuse them if they are equal, I wonder if doing a full comparision is worth it.
The other way around, only comparing some attributes (and thus having objects that compare equal but are not), is a bad solution I think and could be the source of very hard to find bugs.

I would actually vote for removing this whole part (event though I just refactored it^^), if there is no strong argument against it.

cc @asalzburger @paulgessinger

andiwand (approved):
looks good to me :+1:

andiwand (approved):

PR #1835, opened 2023-02-07
fix: enable compilation with ONNX 1.13

gagnonlg:
Simple PR to fix a compilation error against ONNX runtime 1.13.1. Basically, the `Ort::Session::Get{Input,Output}Name` functions got removed starting from 1.13, and we have the use the `Ort::Session::Get{Input,Output}NameAllocated` versions which were introduced in 1.12. 

One implication is that this PR introduces a lower bound of 1.12.0 on the ONNX runtime version.


paulgessinger:
Let me tag @benjaminhuth. Thoughts?

Corentin-Allaire:
I have done those exact same change (plus some other) in my own branch. I general I think we should discuss how Onnx is implemented in Acts right now (with the current implementation you need to create a new class that inherit from OnnxRuntimeBase each time you import a new network but I will probably make  PR for this once I open a PR for the ambiguity resolver

gagnonlg:
Good point about Exa.Trk, tagging @xju2 -- do you know if your pipeline works with OnnxRuntime >= 1.13?

gagnonlg:
>  I general I think we should discuss how Onnx is implemented in Acts right now (with the current implementation you need to create a new class that inherit from OnnxRuntimeBase each time you import a new network but we can discuss this once I open a PR for the ambiguity resolver

Do you mean that in your ambi. solver work, you rely on changing the current implementation? If so, maybe i'd be worth it to merge that part in (or start the discussion as you suggest) so as to minimize potential divergence? just my .02$

Corentin-Allaire:
> > I general I think we should discuss how Onnx is implemented in Acts right now (with the current implementation you need to create a new class that inherit from OnnxRuntimeBase each time you import a new network but we can discuss this once I open a PR for the ambiguity resolver
> 
> @Corentin-Allaire Do you mean that in your ambi. solver work, you rely on changing the current implementation? If so, maybe it'd be worth it to merge that part in sooner (or start the discussion as you suggest) so as to minimize potential divergence? just my .02$

My main change for now but it is not definitive was to switch `runONNXInference` to public that way your algorithm can have a Acts::OnnxRuntimeBase member that load the model of your choice instead of writing a new class for each model you import as it is done right now. 
What do you mean by `.02$` ? Is it a typo or something I don't understand ?

gagnonlg:
> @gagnonlg My main change for now but it is not definitive was to switch `runONNXInference` to public that way your algorithm can have a Acts::OnnxRuntimeBase member that load the model of your choice instead of writing a new class for each model you import as it is done right now.

I see, seems like that would be a backwards compatible change if that decision is made so there's no hurry, I guess

>  What do you mean by `.02$` ? Is it a typo or something I don't understand ?

Oh, it's literally "my 2 cents", an idiom that basically means "in my opinion", i.e. feel free to disagree

benjaminhuth:
We have a new image now, if you replace `ghcr.io/acts-project/ubuntu2004_exatrkx:v30` with `ghcr.io/acts-project/ubuntu2004_exatrkx:v31` in `.gitlab-ci.yaml`, then we should see if it acutally builds @gagnonlg

gagnonlg:
@benjaminhuth awesome, let's try it now

gagnonlg:
@benjaminhuth are you sure it's v31? Looks like OnnxRuntime 1.10 is still being picked up

gagnonlg:
> Ah I mixed the numbers. It is release 34, not 31. Sorry!

No worries, it's an easy fix!

benjaminhuth:
Hmm this time the error is 
```
section_end:1675953156:prepare_executor ERROR: Job failed: failed to pull image "ghcr.io/acts-project/ubuntu2004_exatrkx:v34" with specified policies
[always]: write /var/lib/docker/tmp/GetImageBlob814379636: no space left on device (manager.go:203:71s)
```

Do we need to reduce somehow the size of the image @paulgessinger ?

paulgessinger:
No I guess the disk is simply full on the VM. In principle I clear the image storage regularly but you can try rerunning which hopefully ends up on another VM.

xju2:
> 

Hi @gagnonlg , the ONNX models were not working because some operations were not supported by that time. Maybe the new version will fix those. ;-)

gagnonlg:
@benjaminhuth from the ci-bridge:

```
$ python3 src/Examples/Scripts/Python/exatrkx.py onnx
Traceback (most recent call last):
  File "src/Examples/Scripts/Python/exatrkx.py", line 5, in <module>
    import acts.examples
  File "/builds/acts/ci-bridge/build/python/acts/__init__.py", line 7, in <module>
    from .ActsPythonBindings import *
ImportError: libonnxruntime.so.1.13.1: cannot open shared object file: No such file or directory
```

benjaminhuth:
Nice that it builds without errors!

The test workflow still runs on the `v30` image, you would also need to update it to `v34` (also in `.gitlab-ci.yaml`)
But then I'm confident it runs through!

gagnonlg:
@benjaminhuth of course, I need to bump the test image as well :clown_face:

gagnonlg:
All fine now, I think this can go in. @benjaminhuth can you approve?

benjaminhuth (approved):

PR #1836, opened 2023-02-08
refactor: Move Chi2 example bindings to TrackFitting

paulgessinger:
I don't think there's a strong need for a separate cpp file for this purpose.


andiwand (approved):
simple and clean :+1:

PR #1840, opened 2023-02-08
ci: Mark ODD tests, move away from exact event count checks

paulgessinger:
The event count checks should be covered by the hash checks, and we have a much better way of obtaining local references with the hashes in any case.

### DIFF
--- a/Examples/Python/tests/test_examples.py
+++ b/Examples/Python/tests/test_examples.py
@@ -835,6 +835,7 @@ def test_volume_material_mapping(material_recording, tmp_path, assert_root_hash)
             marks=[
                 pytest.mark.skipif(not dd4hepEnabled, reason="DD4hep not set up"),
                 pytest.mark.slow,
+                pytest.mark.odd,
             ],
         ),
         (functools.partial(AlignedDetector.create, iovSize=1), 450),
--- a/Examples/Python/tests/test_examples.py
+++ b/Examples/Python/tests/test_examples.py
@@ -1082,6 +1082,7 @@ def test_ckf_tracks_example(
 @pytest.mark.skipif(not dd4hepEnabled, reason="DD4hep not set up")
 @pytest.mark.skipif(not pythia8Enabled, reason="Pythia8 not set up")
 @pytest.mark.slow
+@pytest.mark.odd
 @pytest.mark.filterwarnings("ignore::UserWarning")
 def test_vertex_fitting(tmp_path):
     detector, trackingGeometry, decorators = getOpenDataDetector(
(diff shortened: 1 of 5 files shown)


Examples/Python/tests/test_examples.py
```diff
@@ -62,7 +62,15 @@ def assert_entries(root_file, tree_name, exp):
...
     assert tree_name in keys
-    assert rf.Get(tree_name).GetEntries() == exp, f"{root_file}:{tree_name}"
+    if non_zero:
+        assert rf.Get(tree_name).GetEntries() > 0, f"{root_file}:{tree_name}"
+    if exp is not None:
+        assert rf.Get(tree_name).GetEntries() == exp, f"{root_file}:{tree_name}"
+
+
+def assert_has_entries(root_file, tree_name):
+    __tracebackhide__ = True
```
andiwand, line 72: huch never seen that python magic

Examples/Python/tests/test_examples.py
```diff
@@ -918,17 +916,17 @@ def test_digitization_example(trk_geo, tmp_path, assert_root_hash):
...
+    assert_entries(root_file, "vol9", 0)
+    for tn in (
+        8,
+        12,
+        13,
+        14,
+        16,
+        17,
+        18,
     ):
```
andiwand, line 928: nitpick should we collapse this on one line now?
```suggestion
    for tn in (8, 12, 13, 14, 16, 17, 18):
```

Examples/Python/tests/test_examples.py
```diff
@@ -973,18 +971,20 @@ def test_digitization_example_input(trk_geo, tmp_path, assert_root_hash):
...
+
+    for tn in (
+        8,
+        12,
+        13,
+        14,
+        16,
+        17,
+        18,
     ):
```
andiwand, line 986:
```suggestion
    for tn in (8, 12, 13, 14, 16, 17, 18):
```

andiwand (approved):
looks good :+1: left a few minor comments

Examples/Python/tests/test_examples.py
paulgessinger: I think this is pytest specific. It hides it from the trace in case an assertion failure happens. It's just cleaner to find out what's happening.

andiwand: that is actually really cool. I avoided helper assert functions because of that in the past

Examples/Python/tests/test_examples.py
paulgessinger: Sure

andiwand (approved):

paulgessinger:
Added one more test that I overlooked, can you approve again @andiwand ?

andiwand (approved):

andiwand (approved):

PR #1842, opened 2023-02-09
feat: persist ACTS barcodes in G4 sim

andiwand:
particle barcodes did not make sense after G4 sim because they were just counted upwards without paying attention to the semantics of the bits in the barcode.

this pr adds persistence to the barcodes by storing a mapping in the G4 specific event store. also the particle generation counter is increased with each secondary process


benjaminhuth (approved):
Very nicely done! I think we're getting closer to make Geant4 a drop-in replacement of FATRAS!

pbutti:
Just to confirm: I suppose this fixes these errors:
```
10:08:01    RootTrajecto   WARNING   Truth particle with barcode = 37 not found!
10:08:01    RootTrajecto   WARNING   Truth particle with barcode = 2 not found!
10:08:01    RootTrajecto   WARNING   Truth particle with barcode = 2 not found!
10:08:01    RootTrajecto   WARNING   Truth particle for mj 17 subtraj 0 not found!
10:08:01    RootTrajecto   WARNING   Truth particle with barcode = 73 not found in the input collection!
10:08:01    RootVertexPe   INFO      Total number of reco track-associated truth particles in event : 0
10:08:01    RootVertexPe   INFO      Total number of reco track-associated truth primary vertices : 0
```

andiwand:
yeah exactly @pbutti. the particle IDs haven't been persistent until now which confuses all the performance writers

we may also need a particle selector after the simulation because most of the secondary particles will not be reconstructible and will currently ruin all the performance results

asalzburger:
Hi, looks like this needs an update of the reference for the Geant4 example - not surprising.

asalzburger (approved):

PR #1843, opened 2023-02-09
feat: wire geoid hooks in g4 detector

andiwand:
- G4 detector was not able to handle geoid hooks
- sneaked in some cleanups - we can break this out if necessary


benjaminhuth (approved):
LGTM! I think when you format the python code, we get this in.

PR #1844, opened 2023-02-09
fix: make full_chain_odd.py really work with pythia8+Geant4, add test

benjaminhuth:
- Fixes some minor issues in the `full_chain_odd.py` when run it with pyhia8+geant4.
   - currently there are FPEs, but I think it is okay to disable this for now and fix that later
- Adds `argeparse` command line arguments to steer number of events, simulation and ttbar
- Adds test function for 1 event to the `pytest` (caveat: 1 event takes **10 minutes** on my local machine...)

cc @andiwand @paulgessinger @pbutti


Examples/Python/tests/test_examples.py
```diff
@@ -1225,7 +1253,7 @@ def test_full_chain_odd_example(tmp_path):
     env["NEVENTS"] = "1"
     env["ACTS_LOG_FAILURE_THRESHOLD"] = "WARNING"
```
andiwand, line 1254: should we kill those?

andiwand (commented):
looks good. lets see how long the ci takes

pbutti:
Hi @benjaminhuth,

thank you for these updates. Is there a way to also add the download of the G4 data files to the geant4.sh setup script (or to another setup script) ?

Examples/Python/tests/test_examples.py
benjaminhuth: yes

benjaminhuth:
I think we could try to add a check if the downloaded data of Geant4 are present. 
But not sure if we should try to download them inside this script.

pbutti:
In principle the check is there (as it will return errors for data missing). What would be good to have is the script that at least tells you to issue ```download_geant4_data.sh``` if the tables are missing

paulgessinger:
> In principle the check is there (as it will return errors for data missing). What would be good to have is the script that at least tells you to issue `download_geant4_data.sh` (when running in the docker) if the tables are missing

Btw I recently learned G4 actually ships with this functionality: `/usr/local/bin/geant4-config --install-datasets`.
So we don't need to use this extra script I need anymore.

Examples/Python/tests/test_examples.py
```diff
@@ -1224,10 +1224,37 @@ def test_full_chain_odd_example(tmp_path):
...
+    script = (
+        Path(__file__).parent.parent.parent.parent
+        / "Examples"
+        / "Scripts"
+        / "Python"
+        / "full_chain_odd.py"
+    )
+    assert script.exists()
+    env = os.environ.copy()
+    env["ACTS_LOG_FAILURE_THRESHOLD"] = "WARNING"
```
benjaminhuth, line 1255:
```suggestion
    env["ACTS_LOG_FAILURE_THRESHOLD"] = "ERROR"
```

andiwand (approved):
:rocket:

PR #1845, opened 2023-02-09
docs: Ignore https://cernvm.cern.ch linkcheck

benjaminhuth:
This let the docs build fail due to certificate errors currently

PR #1846, opened 2023-02-09
fix: format docs/conf.py

benjaminhuth:
This sneaked in with https://github.com/acts-project/acts/pull/1845


docs/conf.py
```diff
@@ -48,10 +48,7 @@
 myst_heading_anchors = 3
 
 linkcheck_retries = 5
-linkcheck_ignore = [
-    r"https://doi.org/.*",
-    r"https://cernvm.cern.ch/.*"
```
andiwand, line 53: I think if you put a `,` at the end it would keep the formatting

but both looks fine for me

andiwand (approved):

PR #1850, opened 2023-02-09
fix: Pass By Reference function and Extend object to binned group

CarloVarni:
self-explanatory


Core/include/Acts/Seeding/BinnedSPGroup.hpp
```diff
@@ -273,11 +273,11 @@ class BinnedSPGroup {
   template <typename spacepoint_iterator_t>
   BinnedSPGroup<external_spacepoint_t>(
       spacepoint_iterator_t spBegin, spacepoint_iterator_t spEnd,
-      GlobalPositionFunctor toGlobal,
+      GlobalPositionFunctor& toGlobal,
       std::shared_ptr<Acts::BinFinder<external_spacepoint_t>> botBinFinder,
       std::shared_ptr<Acts::BinFinder<external_spacepoint_t>> tBinFinder,
       std::unique_ptr<SpacePointGrid<external_spacepoint_t>> grid,
-      Acts::Extent rRangeSPExtent,
+      Acts::Extent& rRangeSPExtent,
```
asalzburger, line 280: will that be changed inside, otherwise it should be `const` right?

Core/include/Acts/Seeding/BinnedSPGroup.ipp
```diff
@@ -9,11 +9,11 @@ template <typename external_spacepoint_t>
 template <typename spacepoint_iterator_t>
 Acts::BinnedSPGroup<external_spacepoint_t>::BinnedSPGroup(
     spacepoint_iterator_t spBegin, spacepoint_iterator_t spEnd,
-    GlobalPositionFunctor toGlobal,
+    const GlobalPositionFunctor& toGlobal,
     std::shared_ptr<Acts::BinFinder<external_spacepoint_t>> botBinFinder,
     std::shared_ptr<Acts::BinFinder<external_spacepoint_t>> tBinFinder,
     std::unique_ptr<SpacePointGrid<external_spacepoint_t>> grid,
-    Acts::Extent rRangeSPExtent,
+    Acts::Extent& rRangeSPExtent,
```
asalzburger, line 16: same here.

asalzburger (approved):
Conditionally approved, but still waiting for clarification.

Core/include/Acts/Seeding/BinnedSPGroup.hpp
CarloVarni: It gets modifies:
```cpp
    // store x,y,z values in extent
    rRangeSPExtent.extend({spX, spY, spZ});
```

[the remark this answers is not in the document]
CarloVarni: I see now.... min and max are used later in the seeding with the Range1D object. However we were making a copy here (which confused me) so I suppose that in the seeding Algorithm we are actually using the default values of the Extent and whatever happens inside the binned sp group had no effect on it...

I'm now curious to see the effect of this change on performance and timings

CarloVarni:
[timings.pdf](https://github.com/acts-project/acts/files/10700514/timings.pdf)

Unfortunately this didn't make a miracle, but it's an improvement at least.

/cc @noemina

Core/include/Acts/Seeding/BinnedSPGroup.hpp
```diff
@@ -266,18 +266,15 @@ class BinnedSPGroup {
...
-      std::function<std::pair<Acts::Vector3, Acts::Vector2>(
-          const external_spacepoint_t&, float, float, float)>;
-
-  template <typename spacepoint_iterator_t>
+  template <typename spacepoint_iterator_t,
+	    typename callable_t>
   BinnedSPGroup<external_spacepoint_t>(
       spacepoint_iterator_t spBegin, spacepoint_iterator_t spEnd,
-      GlobalPositionFunctor toGlobal,
+      callable_t&& toGlobal,
```
andiwand, line 273: whats the reason to change this to an rvalue reference?

CarloVarni: It's a universal reference, so lvalue or rvalue according to what the end user passes

LuisFelipeCoelho:
Thank you @CarloVarni for finding this bug. It was included recently and probably does not affect the physics performance that much, as I did not notice it and neither did the CI.
But I think the time difference is significant!
This is the last comparison with the changes from PR #1764 (not yet with this PR):
Screenshot 2023-02-09 at 20 29 39

I will test with this PR as well, maybe it will be enough to reach Athena

LuisFelipeCoelho (changes requested):
Since this is used by the itk configuration only and it is not necessary to the default seeding, I guess `useVariableMiddleSPRange` needs to be set to false as default in seedFinderConfig so this does not affect the efficiency of the default seeding

I also think we don't actually need the extent, we just need to calculate the max and min deltaR from the number of R bins times the position of the first or last filled radial Bin times a configurable factor that in acts we call deltaRMiddleMaxSPRange and deltaRMiddleMinSPRange, as it is done in here: https://acode-browser1.usatlas.bnl.gov/lxr/source/athena/InnerDetector/InDetRecTools/SiSpacePointsSeedTool_xk/src/ITkSiSpacePointsSeedMaker.cxx#1499

LuisFelipeCoelho:
In case you don't want to change the extent I would only set useVariableMiddleSPRange to false so this does not affect the other seeding examples

CarloVarni:
Hi @LuisFelipeCoelho I have set the flag to `false` as you suggested. 

Concerning the Extent: I'd have this PR added to ACTS so that we can now pass the proper min and max values to the seeding. The question on whether to keep/change/remove the Extent object can be addressed in another PR eventually

LuisFelipeCoelho (approved):

andiwand:
@CarloVarni there seem to be a few new conflicts in the hashes. can you update those based on the new CI output to get this in?

Core/include/Acts/Seeding/SeedFinderConfig.hpp
```diff
@@ -51,7 +51,7 @@ struct SeedFinderConfig {
 
   // radial range for middle SP
   // variable range based on SP radius
-  bool useVariableMiddleSPRange = true;
+  bool useVariableMiddleSPRange = false;
```
andiwand, line 54: what is the reason for this change?

andiwand (commented):
looks like a bunch of reference files changed. maybe you want to take another look @paulgessinger

paulgessinger:
I haven't looked at the comparison plots, but I agree it would be good to check that the changes are indeed within expectations. Maybe we should wait for @CarloVarni to be back to work?

Core/include/Acts/Seeding/SeedFinderConfig.hpp
asalzburger: This has been commented by @LuisFelipeCoelho above, I suppose.

CarloVarni:
Comparison plots are now available: Seeding (https://herald.dokku.paulgessinger.com/view/acts-project/acts/565610827/seeding_seeded.html), CKF (https://herald.dokku.paulgessinger.com/view/acts-project/acts/565610827/ckf_seeded.html) and AMBI (https://herald.dokku.paulgessinger.com/view/acts-project/acts/565610827/ambi_seeded.html) 

Since this PR is simply fixing a bug (and passing now correct values to the seeding algorithm) I'd say these changes are expected. /cc @LuisFelipeCoelho 

If these are ok for you, I'll put the ref files back and proceed with the PR.

Core/include/Acts/Seeding/BinnedSPGroup.hpp
```diff
@@ -266,18 +266,14 @@ class BinnedSPGroup {
...
-  using GlobalPositionFunctor =
-      std::function<std::pair<Acts::Vector3, Acts::Vector2>(
-          const external_spacepoint_t&, float, float, float)>;
-
-  template <typename spacepoint_iterator_t>
+  template <typename spacepoint_iterator_t, typename callable_t>
   BinnedSPGroup<external_spacepoint_t>(
       spacepoint_iterator_t spBegin, spacepoint_iterator_t spEnd,
-      GlobalPositionFunctor toGlobal,
+      callable_t&& toGlobal,
```
goetzgaycken, line 272: Is there a reason why the functor has to be generalised ?

CarloVarni: It's more a small optimization here that adds some additional flexibility. Being a universal reference we can simply pass an lvalue or rvalue and we do not need to create a `std::function`. 

Said so, the functionality is the same

LuisFelipeCoelho (approved):

PR #1851, opened 2023-02-09
feat: post simulation particle filter

andiwand:
as discovered in https://github.com/acts-project/acts/pull/1842 we should filter particles before and after simulation especially with geant4 as we get a lot of secondaries


benjaminhuth (approved):
Let's get this in!

PR #1853, opened 2023-02-10
fix: FPE in RootTrajectorySummaryWriter

andiwand:
in case we cannot find the truth particle we try to cast `NaN` to `int` which will cause an FPE


Examples/Io/Root/src/RootTrajectorySummaryWriter.cpp
```diff
@@ -213,7 +213,7 @@ ActsExamples::ProcessCode ActsExamples::RootTrajectorySummaryWriter::writeT(
       // Initialize the truth particle info
       uint64_t majorityParticleId = std::numeric_limits<size_t>::max();
       unsigned int nMajorityHits = std::numeric_limits<unsigned int>::max();
-      float t_charge = NaNfloat;
+      int t_charge = 99;
```
paulgessinger, line 216: I think I'd prefer a check below when attempting the cast.
NaN is imo a better sentinel value than 99, which is arbitrary.

andiwand: I agree but the branch actually wants an int so I did't know what to put

in general it feels kind of stupid to put values into the particle branches if we dont know them. but I didn't want to refactor the whole thing

benjaminhuth: why not `std::numeric_limits<int>::max()`?

paulgessinger (approved):

PR #1854, opened 2023-02-10
feat: Add TrajectoryPrinter helper algorithm

paulgessinger:
This PR adds a helper algorithm that prints out Trajectories. I'm using this to debug. It might not be useful for a long time, but I thought it might make sense to add it at this point in time anyway.

### DIFF
--- a/Examples/Python/python/acts/examples/reconstruction.py
+++ b/Examples/Python/python/acts/examples/reconstruction.py
@@ -881,10 +881,12 @@ def addCKFTracks(
 
     s.addWhiteboardAlias("trajectories", trackFinder.config.outputTrajectories)
 
-    s.addAlgorithm(acts.examples.TrajectoryPrinter(
-        level=customLogLevel(),
-        inputTrajectories=trackFinder.config.outputTrajectories
-    ))
+    s.addAlgorithm(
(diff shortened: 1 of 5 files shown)


Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -881,6 +881,13 @@ def addCKFTracks(
 
     s.addWhiteboardAlias("trajectories", trackFinder.config.outputTrajectories)
 
+    s.addAlgorithm(
+        acts.examples.TrajectoryPrinter(
+            level=customLogLevel(),
+            inputTrajectories=trackFinder.config.outputTrajectories,
+        )
+    )
```
andiwand, line 889: should we hide this behind a flag or just have it switching on with the correct debug level?

andiwand (approved):
looks good just left one question

Examples/Python/python/acts/examples/reconstruction.py
paulgessinger: Hm, could do I guess.

paulgessinger:
On second thought, maybe we don't actually need to put this in.

PR #1859, opened 2023-02-13
refactor: print numeric value for multi index

andiwand:
was useful for me for debugging as it gets easier to compare numeric particle barcode values

also aligns the way we print these values in the root writer


Core/include/Acts/Utilities/MultiIndex.hpp
```diff
@@ -148,6 +148,7 @@ class MultiIndex {
     for (std::size_t lvl = 1; lvl < NumLevels; ++lvl) {
       os << '|' << idx.level(lvl);
     }
+    os << "=" << idx.value();
```
benjaminhuth, line 151: Actually I'm not sure about that change, since then there is no possibility anymore to print only the leve-wise representation, and its also easy to write `os << idx << "=" << idx.value()`

andiwand: sure but I wondered if that is ever necessary. if so I can turn this around and just adopt the prints in the writers

benjaminhuth: Not sure where this `MultiIndex` class is used, but actually I don't find it totally unneccessary to get the decomposition into the levels.
I think you mostly need either the value representation or the decomposed one, but not both. Therefore I would decouple them, but feel free to ignore this :)

andiwand: I agree I think you need either one and since it is only for printing I though why not have both :smile: 

but fair enough I can just change it at the writer

andiwand: now that you mention it. interesting that geo id does not make use of the multi index. it is only used for the barcode

andiwand: now that you mention it. interesting that geo id does not make use of the multi index. it is only used for the barcode

benjaminhuth (approved):

Core/include/Acts/Utilities/MultiIndex.hpp
benjaminhuth: > now that you mention it. interesting that geo id does not make use of the multi index. it is only used for the barcode

Good question. Maybe it would be worth to unify this at some point

PR #1862, opened 2023-02-13
refactor: rename sim output files

andiwand:
change sim output names and remove fatras prefix as we also use geant 4

also simplify sim writer logic


benjaminhuth (approved):
Very reasonable change, let's get them in

PR #1863, opened 2023-02-14
fix: periodic difference for Phi in RootSummaryWriter

benjaminhuth:
The residual computation in the `RootTrajectorySummaryWriter` did not account for the fact that phi is a cyclic variable. This should fix this.


andiwand (approved):
cool :+1: lets get this in

andiwand (approved):

andiwand (approved):

PR #1865, opened 2023-02-14
refactor: update itk_seeding.py

LuisFelipeCoelho:
This PR updates the csv itk seeding example


LuisFelipeCoelho:
To avoid reallocating memory at each iteration, I added a reserve statement to the sp vectors for the duplets: 

```
// reserve enough capacity
size_t dupletCapacity =
      isBottom ? m_config.bottomDupletCapacity : m_config.topDupletCapacity;
outVec.reserve(dupletCapacity); 
linCircleVec.reserve(dupletCapacity);
```

the dupletCapacity is a configurable variable that I set separately for bottom and top to 200 and 450, based on the maximum number of SPs after the cuts for ttbar <mu>=200 events:

Screenshot 2023-02-15 at 17 19 31

Screenshot 2023-02-15 at 17 19 23

The time difference per event with and without reserving the vector capacity does not seem to be big:

compare

@noemina @CarloVarni

asalzburger (approved):
Good from my point of view.

PR #1866, opened 2023-02-15
ci: Add script checking for `#pragma once` to CI

benjaminhuth:
I noticed a missing `#pragma once` in a file in `Core`, and thought it would be nice to check this in the CI.

### DIFF
--- a/CI/check_pragma_once.sh
+++ b/CI/check_pragma_once.sh
@@ -3,7 +3,7 @@
 ec=0
 
 for file in $(find Core Examples Tests Plugins -name "*.hpp"); do
-    res=$(grep "#pragma once" $file)
+    res=$(grep -e "^[[:space:]]*#pragma once" $file)
     if [[ "$res" != "#pragma once" ]]; then
         ec=1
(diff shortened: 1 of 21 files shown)


paulgessinger (approved):
Seems like a good idea to me!

andiwand (approved):
cool!

looks like we are missing some includes now which cascaded through before

CI/check_pragma_once.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+ec=0
+
+for file in $(find Core Examples Tests Plugins -name "*.hpp"); do
+    res=$(grep "#pragma once" $file)
```
gagnonlg, line 6: maybe pedantic (feel free to ignore) but you make this more robust, e.g. `grep -e '^[[:space:]]*#pragma once' test.cxx` . This way you will catch instance where the pragma is commented-out for instance

benjaminhuth: nice, I changed the `grep` command!

andiwand (approved):

PR #1868, opened 2023-02-16
feat: Batch input for ONNX

Corentin-Allaire:
This PR change two aspects of the OnnxRuntimeBase class. 
- First, it moves the inference from protected to public, this means it is no longer necessary to create a new class for every single network we want to import, and we can now directly use the OnnxRuntimeBase one.
- Second, this Pr allows the use of batch input (instead of running the network once per entry we can directly use a vector of entry) this can speed up greatly the inference. It would require the model to have been implemented with dynamic axes to work, but the code still supports non-dynamic axes. This still support the old interface where only one entry was used is thus non-breaking.
This PR will be needed later on for the integration of the ML Ambiguity solver. (The use of batch input result there in a factor 7 speed up)


Plugins/Onnx/src/OnnxRuntimeBase.cpp
```diff
@@ -11,102 +11,126 @@
...
+  if (batchSize != 1 &&
+      (inputNodeDims[0] != batchSize || outputNodeDims[0] != batchSize)) {
+    throw std::runtime_error(
+        "runONNXInference: batch size doesn't match the input or output node "
+        "size");
+  }
+
+  // Format the input as a single vector
+  std::vector<float> tensorValues;
+  for (auto track : inputTensorValues) {
```
benjaminhuth, line 92: If we would pass the data not in form of a `std::vector<std::vector<float>>` but as an `Eigen::Array` with dynamic size NxM this would have multiple advantages:
* We could directly pass the data pointer to Onnx without copying them into another vector (if the row-major / column major thing is correct)
* We would save some memory allocations I guess
* It would also easy to check that the size of the input is correct, and it would not be so easy to pass invalid input to the function (e.g., the inner `std::vector`s could have different length).

Corentin-Allaire: That seems like an elegant solution, I usually try to avoid using Eigen if not needed but I will give it a try !

benjaminhuth: I think something like `boost::multi_array<float, 2>` would maybe be an even better solution, so you could try this as well...
But even though Eigen is not optimal, most people in our field have experience with it which is an advantage.

Corentin-Allaire: Done, for your information de default of the `Eigen::Array` are CollumnMajor but change it is quite easy !

asalzburger (approved):
From my limited understanding, this looks fine.

PR #1869, opened 2023-02-16
feat: Add parameter getters to TrackProxy

paulgessinger:
This allows accessing the raw parameter elements by named methods, and to derive properties like absolute and transverse momentum.

Anything else that comes to mind we should add at this point?

### DIFF
--- a/Core/include/Acts/EventData/Track.hpp
+++ b/Core/include/Acts/EventData/Track.hpp
@@ -8,6 +8,9 @@
 
 #pragma once
 
+#include "Acts/Definitions/Algebra.hpp"
+#include "Acts/Definitions/TrackParametrization.hpp"
+#include "Acts/Definitions/Units.hpp"
 #include "Acts/EventData/MultiTrajectory.hpp"
 #include "Acts/Utilities/HashedString.hpp"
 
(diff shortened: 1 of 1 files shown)


Core/include/Acts/EventData/Track.hpp
```diff
@@ -206,6 +209,46 @@ class TrackProxy {
...
+  /// Access the time parameter of the track at the reference surface
+  /// @return The time parameter
+  ActsScalar time() const { return parameters()[eBoundTime]; }
+
+  /// Get the absolute momentum of the tack
+  /// @return The absolute track momentum
+  ActsScalar absoluteMomentum() const {
+    return std::abs(static_cast<ActsScalar>(Acts::UnitConstants::e) /
+                    parameters()[eBoundQOverP]);
+  }
```
andiwand, line 244: do we have to worry about not singly charged particles at some point?

Core/include/Acts/EventData/Track.hpp
```diff
@@ -206,6 +209,46 @@ class TrackProxy {
...
+
+  /// Get the transverse momentum of the track
+  /// @return The track transverse momentum value
+  ActsScalar transverseMomentum() const {
+    return std::sin(parameters()[eBoundTheta]) * absoluteMomentum();
+  }
```
andiwand, line 250: it is a simple calculation but I saw this a few times in the code - do you think we should have an utility for that?

andiwand (approved):
looks good to me

only left a comment related to singly charged particles and some possible utility

Core/include/Acts/EventData/Track.hpp
paulgessinger: In principle we could, although you have to make an assumption about the charge value.

andiwand: I guess we could also use the `charege` function here and as soon as we have a different charge there this would automatically work

andiwand (approved):

Core/include/Acts/EventData/Track.hpp
paulgessinger: I'll reuse `SinglyCharged`.

andiwand (approved):
looks good :+1:

PR #1870, opened 2023-02-16
feat: Enable removing tracks from container

paulgessinger:
This also moves some method definitions to a cpp file for VectorTrackContainer, and adds the ability to copy a const track vector container to a mutable one (for modification).

### DIFF
--- a/Core/include/Acts/EventData/VectorTrackContainer.hpp
+++ b/Core/include/Acts/EventData/VectorTrackContainer.hpp
@@ -80,21 +80,29 @@ class VectorTrackContainerBase {
     }
   }
 
-  void checkConsistency() const {
+  bool checkConsistency() const {
     size_t size = m_tipIndex.size();
     (void)size;
 
-    assert(m_tipIndex.size() == size);
-    assert(m_params.size() == size);
-    assert(m_cov.size() == size);
-    assert(m_referenceSurfaces.size() == size);
-    assert(m_nMeasurements.size() == size);
-    assert(m_nHoles.size() == size);
+    bool result = true;
+    result = result && m_tipIndex.size() == size;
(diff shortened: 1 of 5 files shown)


Core/include/Acts/EventData/VectorTrackContainer.hpp
```diff
@@ -88,6 +80,23 @@ class VectorTrackContainerBase {
...
+    assert(m_cov.size() == size);
+    assert(m_referenceSurfaces.size() == size);
+    assert(m_nMeasurements.size() == size);
+    assert(m_nHoles.size() == size);
+
+    for (const auto& [key, col] : m_dynamic) {
+      (void)key;
+      assert(col->size() == size);
+    }
+  }
```
andiwand, line 98: is this temporary? I guess we can leave it as it should be optimized out in release build?

andiwand (approved):

Core/include/Acts/EventData/VectorTrackContainer.hpp
paulgessinger: So actually you're right that the `for` loop should probably go in case of a release build. Maybe I'll add function-level guards against this so it doesn't run at all.

andiwand: I guess the optimizer would remove it because otherwise the body is empty?

andiwand (approved):

Core/include/Acts/EventData/VectorTrackContainer.hpp
paulgessinger: Reworked this so the `assert` is (also) around the `checkConsistency` call, so suould be zero cost now.

PR #1871, opened 2023-02-16
refactor: Improved const-correctness for TrackContainer in Examples

paulgessinger:
Also includes a fix of a missing method in the ConstVectorTrackContainer class.

### DIFF
--- a/Core/include/Acts/EventData/VectorTrackContainer.hpp
+++ b/Core/include/Acts/EventData/VectorTrackContainer.hpp
@@ -209,6 +209,10 @@ class ConstVectorTrackContainer final
     return ConstParameters{m_params[itrack].data()};
   }
 
+  ConstCovariance covariance(IndexType itrack) const {
+    return ConstCovariance{m_cov[itrack].data()};
+  }
+
   // END INTERFACE
 };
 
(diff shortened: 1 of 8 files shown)


Core/include/Acts/EventData/VectorTrackContainer.hpp
```diff
@@ -209,6 +209,10 @@ class ConstVectorTrackContainer final
     return ConstParameters{m_params[itrack].data()};
   }
 
+  ConstCovariance covariance(IndexType itrack) const {
+    return ConstCovariance{m_cov[itrack].data()};
+  }
```
andiwand, line 214: unrelated: looking at the pattern `something(Index itrack)` I wondered if it would make sense to have some boxing around this? so it would look like `track(Index itrack).covariance()`

andiwand (approved):
looks good - just left one unrelated question

Core/include/Acts/EventData/VectorTrackContainer.hpp
paulgessinger: That's basically what the `TrackProxy` class does.

```cpp
auto track = trackContainer.getTrack(5);
track.covariance();
```

andiwand (approved):

PR #1875, opened 2023-02-17
refactor: remove `operator==` from digitization config objects

benjaminhuth:
There is an operator== defined for `GeometricConfig` and `SmearingConfig`, however, this operator does not really test equality, since it cannot check the involved `std::function` objects. This PR removes this operator, since `==` has a clear defined semantics and should not be used if equality is not checked actually.


asalzburger:
Still some build error ...

asalzburger:
Hi @benjaminhuth - there's still a remaining usage of the `compatcitfy` in one of the tests, this makes the CI fail.

asalzburger (approved):
I am happy with the removal, we can always compactify a potential configuration file by checking the persistified json format.

asalzburger:
Waiting for the CI to run through.

andiwand:
this breaks a python unit test that asserts on the number of digitization configs

```
=================================== FAILURES ===================================
_______________________ test_digitization_config_example _______________________

trk_geo = <acts.ActsPythonBindings.TrackingGeometry object at 0x7f866cab0ef0>
tmp_path = PosixPath('/tmp/pytest-of-root/pytest-0/test_digitization_config_examp0')

    def test_digitization_config_example(trk_geo, tmp_path):
        from digitization_config import runDigitizationConfig
    
        out_file = tmp_path / "output.json"
        assert not out_file.exists()
    
        input = (
            Path(__file__).parent
            / "../../../Examples/Algorithms/Digitization/share/default-smearing-config-generic.json"
        )
        assert input.exists(), input.resolve()
    
        runDigitizationConfig(trk_geo, input=input, output=out_file)
    
        assert out_file.exists()
    
        with out_file.open() as fh:
            data = json.load(fh)
        assert len(data.keys()) == 2
        assert data["acts-geometry-hierarchy-map"]["format-version"] == 0
        assert (
            data["acts-geometry-hierarchy-map"]["value-identifier"]
            == "digitization-configuration"
        )
>       assert len(data["entries"]) == 27
E       assert 18728 == 27
E         +18728
E         -27

Examples/Python/tests/test_examples.py:1001: AssertionError
```

benjaminhuth:
@asalzburger Okay I finally decided to keep the `compactify` option but to implement it with a free function instead of `operator==`, think this is a good way to keep the feature and not having a missleading operator.

If thats okay, I think we could get this into the merge pipeline after a re-approval.

asalzburger (approved):

PR #1877, opened 2023-02-17
feat: Machine learning based Ambiguity Solver

Corentin-Allaire:
This MR depends on #1868.

Implement a machine learning based ambiguity solver. If Acts is compiled with the `ACTS_PLUGIN_ONNX` option, it can be used in the odd full chain by simply adding the `--MLSolver` argument.

In addition to the implementation of the solver, this PR adds an option for the regular solver to only work on track with a certain amount of hits (to obtain fairer comparison). 

This also add some small change to the Csv multitrajectory writing, it now also write the list of measurements associated with each track.

Finally, the python script to train the network, test the network and compute its performances are made available with the onnx file of the model. 

I haven't written the documentation for this solver, yet it will be part of a future MR after this gets merged in.


asalzburger:
This is on top of the batch input for ONNX #1868 ?

Corentin-Allaire:
> This is on top of the batch input for ONNX #1868 ?

Yes, the batch input caused a factor 7 speed up so I thought it should go in first. But I guess it is close to be merged

asalzburger:
Alright, I have seen that you had addressed @benjaminhuth s comments, so I merged #1868

Corentin-Allaire:
The jobs that don't pass the CI are related to the fact I have change the default solver to only consider track with at least 7 hits. I can put the threshold to 0 but honestly I would be concern by the quality of a track with < 7 hits.
If anybody has an opinion on this...

asalzburger:
> e CI are rel

In that case, the PR should include the updated reference, what do you think? @paulgessinger

Examples/Python/src/TrackFinding.cpp
```diff
@@ -10,6 +10,9 @@
 #include "Acts/Seeding/SeedFinderOrthogonalConfig.hpp"
 #include "Acts/TrackFinding/MeasurementSelector.hpp"
 #include "ActsExamples/TrackFinding/AmbiguityResolutionAlgorithm.hpp"
+#ifdef ACTS_PLUGIN_ONNX
```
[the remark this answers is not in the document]
Corentin-Allaire: I am clearly not a cmake expert, would that be better ? If yes could we implement that in a later PR ? I don't really have the time to dive in the cmake doc right now...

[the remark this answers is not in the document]
andiwand: sure no worries I can do that after we got this in

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -157,25 +166,40 @@
         field,
         CKFPerformanceConfig(
             ptMin=1.0 * u.GeV if ttbar_pu200 else 0.0,
-            nMeasurementsMin=6,
+            nMeasurementsMin=7,
```
andiwand, line 169: I guess because of that @paulgessinger

Corentin-Allaire:
> Is it obvious why the non-ambi outputs also change @Corentin-Allaire?

I the odd-full-chain the vertexing uses the output of the Solver as an input. If you remove some tracks (the one with < 7 measurements) then it will affect the vertexing. What was a bit surprising is that it appear to improve the vertexing performance.

Examples/Scripts/Python/full_chain_odd.py
andiwand: oh yeah sorry. but I guess it is the one below where it also changed from 6 to 7 right?

andiwand: here https://github.com/acts-project/acts/pull/1877/files#diff-218e044dd033aa9356a7aa4785f709eef7f2a72f1b7cb7e3db5a95006efc3b9bL170-L178

andiwand: but I guess this is also only for performance writing

Corentin-Allaire: Yes, what is changing the physmon is this line : https://github.com/acts-project/acts/pull/1877/files#diff-311aa2ecef0d9d175b50ee63b8107c6ef3ef32fe6d3276d1ad462201f4b16891R127-R131

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/AmbiguityResolutionAlgorithm.hpp
```diff
@@ -37,6 +37,9 @@ class AmbiguityResolutionAlgorithm final : public BareAlgorithm {
 
     /// Maximum amount of shared hits per track.
     std::uint32_t maximumSharedHits = 1;
+
+    /// Minumum number of measurement to form a track.
+    size_t nMeasurementsMin = 7;
```
andiwand, line 42: this is the change which causes physmon to create a diff, right @Corentin-Allaire ? cc @paulgessinger

Corentin-Allaire: Yes ! As I mention before In my opinion with the geometry of the ODD a track with < 7 measurement is very unlikely to be a good track

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionMLAlgorithm.cpp
```diff
@@ -0,0 +1,193 @@
...
+    std::pair<int, int> outputTips = trackTips.at(iTrack);
+
+    std::vector<Acts::MultiTrajectoryTraits::IndexType> tips;
+    Trajectories::IndexedParameters parameters;
+    tips.push_back(outputTips.second);
+    parameters.emplace(outputTips.second, trackParameters[iTrack]);
+
+    outputTrajectories.emplace_back(
+        trajectories[outputTips.first].multiTrajectory(), tips, parameters);
+  }
```
asalzburger, line 187: I really enjoyed seeing this and particularly your talk in today's meeting.

I have a bit of a design question:
- suppose we wanted to use this ML solver in ATLAS or sPHENIX wants to use it.
- can we encapsulate this code into the `Core` or a new `Core/ML` in order to make this algorithm not more than an empty shell that does input/output, but then pushes the prepared data into a `Core/ML/DuplicateSolver` ?

benjaminhuth: I'm not sure if we should have ONNX as a dependency of the `Core` module? If we push this code to the Onnx-Plugin, wouldn't it also be usable in ATLAS / sPHENIX?

Corentin-Allaire: I don't have a strong opinion either way, in core we could have only `Core/ML` depend on Onnx. What I dislike a bit about the Plugin in that case is that if we start to have more ML base algorithm they are all going to end up in the Onnx plugin and if we start to combine multiple ML library this will also cause some some issue. We could also have a ML Plugin but at that point what is the difference between this and a `Core/ML` directory ?

Corentin-Allaire: But I would argue that maybe this should be a general discussion in one of our weekly meeting on how we want to integrate ML development in Acts. And maybe for now we should just add this as is in Example ?

paulgessinger: For other plugins we have a split between a plugin and an examples "plugin". You need both to run it. For the python bindings, we can optionally include the ONNX examples in the python module.

Corentin-Allaire: So at this point what should we do ? Do We keep the code as is for now ? Do I move it to the Onnx Plugin or another one ? Or even to Core ? 

I think it would be in everyone interest to have this merge so we can add this to the test/CI.

benjaminhuth: I would vote for moving the parts that do not depend on the examples EDM to the plugin for now. Then it can be used without the Examples framework. However, one problem I currently see the usage of the `ActsExamples::TrajectoriesContainer`. Not sure if this could be easily replaced by the `Acts::TrackContainer` component? @paulgessinger 

If that requires some more shuffling around I would just leave it as is and refactor this at a later point.

Corentin-Allaire: I am not too sure either how easy it would be to replace the `ActsExamples::TrajectoriesContainer` by `Acts::TrackContainer` ...

benjaminhuth: Okay then I would say let's refactor this later and concentrate on the CI for now

andiwand: I think that is a story for another day as it would also effect a lot of other components. to me it makes sense to put this in for now as is. otherwise I fear we just outdate and stale it

Corentin-Allaire: Works for me. As soon as we merge this I can start working on a PR to add it to the CI.
Now the question is can someone review it ?

Corentin-Allaire: I probably also need to update the physmon again...

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -157,25 +166,40 @@
...
         ),
         TrackSelectorRanges(
             pt=(1.0 * u.GeV, None),
             absEta=(None, 3.0),
             loc0=(-4.0 * u.mm, 4.0 * u.mm),
         ),
         outputDirRoot=outputDir,
+        outputDirCsv=outputDir,
```
andiwand, line 177: is this relevant for training? otherwise I would just drop it since we are writing out the root files as well

same fore the ambiguity below

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
```diff
@@ -191,8 +197,9 @@ ActsExamples::ProcessCode ActsExamples::AmbiguityResolutionAlgorithm::execute(
       tips.push_back(tip);
       parameters.emplace(tip, trackParameters[iTrack]);
     }
-
-    outputTrajectories.emplace_back(traj.multiTrajectory(), tips, parameters);
+    if (!tips.empty()) {
+      outputTrajectories.emplace_back(traj.multiTrajectory(), tips, parameters);
+    }
```
andiwand, line 202: I think the idea here was to keep empty trajectories /cc @paulgessinger

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionMLAlgorithm.cpp
```diff
@@ -0,0 +1,193 @@
...
+      if (matchedTrack != hitToTrack.end()) {
+        // Add the track to the cluster associated to the matched track
+        cluster.at(matchedTrack->second).push_back(track->second.first);
+        break;
+      }
+    }
+    // None of the hits have been matched to a track create a new cluster
+    if (matchedTrack == hitToTrack.end()) {
+      cluster.emplace(track->second.first,
+                      std::vector<int>(1, track->second.first));
```
andiwand, line 75:
```suggestion
                      {track->second.first});
```

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionMLAlgorithm.cpp
```diff
@@ -0,0 +1,193 @@
...
+      for (auto hit = hits.begin(); hit != hits.end(); hit++) {
+        // Add the hits of the new cluster to the hitToTrack
+        hitToTrack.emplace(*hit, track->second.first);
+      }
```
andiwand, line 79:
```suggestion
      for (const auto& hit : hits) {
        // Add the hits of the new cluster to the hitToTrack
        hitToTrack.emplace(hit, track->second.first);
      }
```

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionMLAlgorithm.cpp
```diff
@@ -0,0 +1,193 @@
...
+  }
+}
+
+namespace {
+
+/// Clusterise tracks based on shared hits
+///
+/// @param trackMap : Multimap storing pair of track ID and vector of measurement ID. The keys are the number of measurement and are just there to focilitate the ordering.
+/// @return an unordered map representing the clusters, the keys the ID of the primary track of each cluster and the store a vector of track IDs.
+std::unordered_map<int, std::vector<int>> ClusteriseTrack(
```
andiwand, line 49:
```suggestion
std::unordered_map<int, std::vector<int>> clusterTracks(
```

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionMLAlgorithm.cpp
```diff
@@ -0,0 +1,193 @@
...
+      trackID++;
+    }
+    iTraj++;
+  }
+  // Performe the share hit based clustering
+  auto clusters = ClusteriseTrack(trackMap);
+  Acts::NetworkBatchInput networkInput(trackID + 1, 8);
+  trackID = 0;
+  // Get the input feature of the network for all the tracks
+  for (auto const& [key, val] : clusters) {
```
andiwand, line 133:
```suggestion
  for (const auto& [key, val] : clusters) {
```

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionMLAlgorithm.cpp
```diff
@@ -0,0 +1,193 @@
...
+    for (auto const& track : val) {
```
andiwand, line 134:
```suggestion
    for (const auto& track : val) {
```

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionMLAlgorithm.cpp
```diff
@@ -0,0 +1,193 @@
...
+  }
+  // Use the network to compute a score for all the tracks.
+  std::vector<std::vector<float>> outputTensor =
+      m_duplicateClassifier.runONNXInference(networkInput);
+  std::vector<int> goodTracks;
+  int iOut = 0;
+
+  // Loop over all the cluster and only keep the track with the highest score in
+  // each cluster
+  for (auto const& [key, val] : clusters) {
```
andiwand, line 160:
```suggestion
  for (const auto & [key, val] : clusters) {
```

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionMLAlgorithm.cpp
```diff
@@ -0,0 +1,193 @@
...
+    int bestTrackID = 0;
+    float bestTrackScore = 0;
+    for (auto const& track : val) {
```
andiwand, line 163:
```suggestion
    for (const auto& track : val) {
```

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionMLAlgorithm.cpp
```diff
@@ -0,0 +1,193 @@
...
+      iOut++;
+    }
+    goodTracks.push_back(bestTrackID);
+  }
+
+  // Create an output multitrajectory based of the good tracks
+  TrajectoriesContainer outputTrajectories;
+  outputTrajectories.reserve(goodTracks.size());
+  for (std::size_t iGoodTraj = 0; iGoodTraj < goodTracks.size(); ++iGoodTraj) {
+    const auto& iTrack = goodTracks[iGoodTraj];
```
andiwand, line 177:
```suggestion
  for (auto &&iTrack : goodTracks) {
```

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionMLAlgorithm.cpp
```diff
@@ -0,0 +1,193 @@
...
+    std::pair<int, int> outputTips = trackTips.at(iTrack);
```
andiwand, line 178:
```suggestion
    const auto& outputTips = trackTips.at(iTrack);
```

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_perf.py
```diff
@@ -0,0 +1,138 @@
+import glob
+import os
+import math
+
+import pandas as pd
+import numpy as np
+
+from sklearn.preprocessing import LabelEncoder, StandardScaler, OrdinalEncoder
+from sklearn.cluster import DBSCAN
+
```
andiwand, line 10:
```suggestion
```

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_perf.py
```diff
@@ -0,0 +1,138 @@
...
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.utils
```
andiwand, line 13:
```suggestion
```

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_perf.py
```diff
@@ -0,0 +1,138 @@
...
+        datafile = datafile.set_index("particleId")
+        # Transform the hit list from a string to an actual list
+        hitsIds = []
+        mergedIds = []
+        for list in datafile["Hits_ID"].values:
+            hitsIds.append(ast.literal_eval(list))
+        datafile["Hits_ID"] = hitsIds
+        # Combine dataset
+        data.append(datafile)
+    return data
```
andiwand, line 54: looks like this function is duplicated in `train_ambiguity_solver.py`. would be cool if we could import it in one of the modules

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_full_chain.py
```diff
@@ -0,0 +1,338 @@
...
+
+def PrepareData(data: pd.DataFrame) -> tuple[np.ndarray, np.ndarray]:
+    """Prepare the data"""
+    """
+    @param[in] data: input DataFrame to be prepared
+    @return: array of the network input and the corresponding truth  
+    """
+    target_column = "good/duplicate/fake"
+    global avg_mean
+    global avg_sdv
```
andiwand, line 66: :smiling_imp:

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_full_chain.py
```diff
@@ -0,0 +1,338 @@
...
+        for list in datafile["Hits_ID"].values:
+            hitsIds.append(ast.literal_eval(list))
+        datafile["Hits_ID"] = hitsIds
+        # Combine dataset
+        data.append(datafile)
+    return data
+
+
+def PrepareData(data: pd.DataFrame) -> tuple[np.ndarray, np.ndarray]:
+    """Prepare the data"""
```
andiwand, line 59: what is the difference to `PrepareDataSet`?

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_full_chain.py
```diff
@@ -0,0 +1,338 @@
...
+            "cluster",
+        ]
+    )
+    # Normalise variables
+    x_cat = OrdinalEncoder().fit_transform(input.select_dtypes("object"))
+    x = np.concatenate((x_cat, input), axis=1)
+    return x, y
+
+
+def ClusterTracks(
```
andiwand, line 89: personally I am not a fan of functions/variables starting with a capital letter. sorry had to complain at least once :smile:

Examples/Scripts/Python/MLAmbiguityResolution/train_ambiguity_solver.py
```diff
@@ -0,0 +1,388 @@
...
+import torch.nn.functional as F
+import torch.utils
+
+from torch.utils.tensorboard import SummaryWriter
+
+# If we want to use apple GPU (Metal)
+device = torch.device("mps")
+
+
+def PrepareDataSet(CKS_files: list[str]) -> pd.DataFrame:
```
andiwand, line 19: duplicate?

Examples/Scripts/Python/MLAmbiguityResolution/train_ambiguity_solver.py
```diff
@@ -0,0 +1,388 @@
...
+        data = pd.concat([data, datafile])
+    return data
+
+
+avg_mean = [0, 0, 0, 0, 0, 0, 0, 0]
+avg_sdv = [0, 0, 0, 0, 0, 0, 0, 0]
+events = 0
+
+
+def PrepareData(data: pd.DataFrame) -> tuple[np.ndarray, np.ndarray]:
```
andiwand, line 61: duplicate?

Examples/Scripts/Python/MLAmbiguityResolution/train_ambiguity_solver.py
```diff
@@ -0,0 +1,388 @@
...
+        self.sigmoid = nn.Sigmoid()
+
+    def forward(self, z):
+        z = F.relu(self.linear1(z))
+        z = F.relu(self.linear2(z))
+        z = F.relu(self.linear3(z))
+        return self.sigmoid(self.output(z))
+
+
+class Normalise(nn.Module):
```
andiwand, line 118: duplicate?

Examples/Scripts/Python/MLAmbiguityResolution/train_ambiguity_solver.py
```diff
@@ -0,0 +1,388 @@
...
+    scale.fit(input.select_dtypes("number"))
+    avg_mean = avg_mean + scale.mean_
+    avg_sdv = avg_sdv + scale.var_
+    events = events + 1
+    x_cat = OrdinalEncoder().fit_transform(input.select_dtypes("object"))
+    x = np.concatenate((x_cat, input), axis=1)
+    return x, y
+
+
+class DuplicateClassifier(nn.Module):
```
andiwand, line 99: duplicate?

andiwand (commented):
in general the python scripts could share some common code

left a few other comments. nothing critical - as always feel free to ignore

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_full_chain.py
benjaminhuth: you have also the official ACTS coding standard on your side :D https://acts.readthedocs.io/en/latest/codeguide.html#n-3-functions-and-methods-use-mixedcase
But not sure how strict we are in this respective

Examples/Scripts/Python/full_chain_odd.py
Corentin-Allaire: The training uses the csv files as input so yes it is kind of needed. The script I provide for performance evaluation also use Csv so it is also useful to have it for the solver. (They should not be too big ~ 0.1 to 1 Mo per event)

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionMLAlgorithm.cpp
paulgessinger: I think we should look into porting it to `Acts::TrackContainer` but since I'm working on other parts of the chain still, let's not block this for that.

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
Corentin-Allaire: The reason why I added this is that if we don't and we add the cutoff on the number of hits per track they we have warning in the CI and it fail. I would be happy to change it back though

paulgessinger: Not sure I get what you're saying.

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_full_chain.py
Corentin-Allaire: That is the only way I found to pass something to a function by reference in python... But we don't need it anymore so I will remove it

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
Corentin-Allaire: Sorry, without `if (!tips.empty()) {` you try to write empty trajectory if `Nmeas<Nminmeas` for a track, which is true many track. Writing a empty track cause a Warning and the CI is configure to fail in case of warning of the full chain. So, now that I have added a cutoff on the minimum number of hit per track before the solver, we would get Warning in the CI, making it fail.

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_full_chain.py
Corentin-Allaire: It hurt my brain to not start a line with capital letter but you are 100% right I will go change this

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_perf.py
Corentin-Allaire: Good idea ! I will see what I can do

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_full_chain.py
Corentin-Allaire: So prepare `PrepareDataSet` does the reading and formating of the csv input while `PrepareData` prepare the input of the NN (normalisation...). I am really bad at naming stuff very bad name for those function...

Examples/Scripts/Python/full_chain_odd.py
andiwand: I think we should stick with root and for the training people can just trow in csv directories by hand. because the full chain is just meant for sim+reco I would argue

Corentin-Allaire: I don't care that much, but I am not sure I follow your reasoning, why would root be better than Csv for sim+reco ? They are many people that use the Csv files for performance analysis afterward. If the concern is disk space it doesn't really matter as for 100 ttbar event the Csv represent less than 1% of the output.
If you insist on removing it I would argue it should at least be commented out so people know they can change this option.

andiwand: I also don't really care but I would like to keep it consistent. so either write everything in root or csv or both.

the only benefit of using root is that you get a single file and the writers tend to be more complete. at least from my experience sometimes the csv output does not contain all the information the root files do

Corentin-Allaire: So what I propose is that for all algorithm the let you write to csv I add  `// outputDirCsv=outputDir,` commented out that way it is still consistent and if people try to do stuff that require the Csv they should realise what they need to do to get then csv files without having to ask us directly. Works for you ?

andiwand: sounds good to me :+1:

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_perf.py
Corentin-Allaire: I have done the best I could in that regard, it should be clearer with the new implementation

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_full_chain.py
Corentin-Allaire: Should be done now, I hope I haven't missed one

Examples/Scripts/Python/MLAmbiguityResolution/ambiguity_solver_full_chain.py
Corentin-Allaire: The name of the function should be more expicit now.

Corentin-Allaire:
All of the comment should have been addressed now ! I made one tiny change to the file name for the Csv multitrajectory writing, it use to always be `CKFtracks.csv` now the naming depend on the algorithm that called the writer (so it is easier to write the output for the CKF and the solver at the same time) .

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -112,6 +122,7 @@
                 pt=(150 * u.MeV, None),
                 removeNeutral=True,
             ),
+            # outputDirRoot=outputDir,
             outputDirCsv=outputDir,
```
andiwand, line 126:
```suggestion
            outputDirRoot=outputDir,
            # outputDirCsv=outputDir,
```

if you want to include this fix here. otherwise I can do it in a followup pr

andiwand (approved):
great! thank you!

Examples/Scripts/Python/full_chain_odd.py
Corentin-Allaire: So I didn't change this because I am not sure it won't break something in the test or not. If you don't mind doing it as a follow up it would speed up the process of getting this merged

andiwand: I think this is a debugging leftover from someone

sure no worries

andiwand (approved):

PR #1879, opened 2023-02-20
feat: Add script for a parallel Geant4 simulation based on Python's `multiprocessing`

benjaminhuth:
This adds a small example script that shows how to run a Geant4 simulation in parallel with Python's `multiprocessing` module.
This is not possible to do that with the examples framework internal parallelism because of limitations in Geant4.


andiwand (approved):

Examples/Scripts/Python/geant4_parallel.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python3
+from pathlib import Path
+from multiprocessing import Pool
+from functools import partial
+
+import numpy as np
```
paulgessinger, line 6: We don't typically rely on thirdparty libs in the scripts directory. I suppose this is doable without numpy.

Examples/Scripts/Python/geant4_parallel.py
```diff
@@ -0,0 +1,69 @@
...
+        rnd=rnd,
+        outputDirCsv=outputDir / "csv",
+        outputDirRoot=None,
+    )
+    addGeant4(
+        s,
+        detector,
+        trackingGeometry,
+        field,
+        outputDirCsv=outputDir / "csv",
```
paulgessinger, line 50: Do we have to do CSV here?

Examples/Scripts/Python/geant4_parallel.py
```diff
@@ -0,0 +1,69 @@
...
+# * This works only for csv outputs, since they write the results in one file per event
+#   (the naming of the csv-files should be equivalent to a sequential run)
+
+
+def runGeant4EventRange(events, outputDir):
+    import acts
+    import acts.examples
+    from acts.examples.simulation import addParticleGun, addGeant4, EtaConfig
+    from acts.examples.odd import getOpenDataDetector
+    from common import getOpenDataDetectorDirectory
```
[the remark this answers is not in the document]
benjaminhuth: Actually I wanted to make sure that if e.g., Geant4 creates some global state on import, this state is local to the function and thus the process.

Examples/Scripts/Python/geant4_parallel.py
benjaminhuth: what do you mean? we cannot do ROOT output, so if we want to get something out of this we need to do CSV?

Examples/Scripts/Python/geant4_parallel.py
```diff
@@ -0,0 +1,69 @@
...
+
+    s = acts.examples.Sequencer(events=len(events), skip=events[0], numThreads=1)
+
+    outputDir = Path(outputDir)
+    addParticleGun(
+        s,
+        EtaConfig(-2.0, 2.0),
+        rnd=rnd,
+        outputDirCsv=outputDir / "csv",
+        outputDirRoot=None,
```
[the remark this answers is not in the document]
benjaminhuth: Yeah I actually wanted to be explicit here (as mentioned in the comments above the script.

Examples/Scripts/Python/geant4_parallel.py
paulgessinger: Why does ROOT not work?

benjaminhuth: Because then we have multiple root writers in the different processes that open the same file and try to write to it. I'm pretty sure that does not work out well...

paulgessinger: But can't we do multiple files and `hadd` afterwards?

benjaminhuth: Yeah in principle yes, of course. But for my purpose I need the csv-files anyways and I actually don't want to implement this right now (and it's not possible without thirdparty libraries I guess).
The nice thing about the CSV output is that it gives equivalent output kind of "out of the box", so I wanted to share it...

I could add a line in the comments that indicates how one could achieve this with root output as well

Examples/Scripts/Python/geant4_parallel.py
andiwand: I guess we can get rid of numpy here because we only use it to hand out the event numbers. not sure how critical this is

benjaminhuth: Yeah it's just kind of handy that numpy provides a function that does exactely what I want here, but of course that can be done also quite easily by hand

Examples/Scripts/Python/geant4_parallel.py
paulgessinger: Ok. The only reason I bring this up is that the CSV files are ludicrously large. `hadd` is part of ROOT.

But ok to go ahead with your suggestion, just asking.

paulgessinger:
I'm removing the autolabel, as there's no valid review. Is this good to go? Please only assign the automerge label when there's an approval otherwise we block the merge queue.

andiwand:
oh sorry I thought my review is still valid and this is ready to go

is there a way to see which PR is blocking the queue?

andiwand (approved):
looks good to me - I think this can be a nice workaround if you just want to speed up the G4 sim

PR #1880, opened 2023-02-20
refactor: GSF counts backward-pass misses as outliers

benjaminhuth:
Now, the GSF does flag states that were missed at the backward pass as outliers. This comes with some shuffling of MultiTrajectories in the end, maybe this can be improved at some point.

Also refactors the use of `std::map` to `std::unordered_map`, as this should perform better in this case (though not benchmarked).


paulgessinger:
Does the copying of track states have performance implications?

benjaminhuth:
Probably a bit, though not benchmarked yet. From a "bottleneck" point of view this is probably minor compared to other time consuming things in the GSF.

Actually I thought about a `reverseTrack` function in the MTJ. This should be not difficult to do in principle (just need to change the `previous` value correctly), but of course this is dangerous for endpoints that belong to trajectories with branches... What do you think @paulgessinger ?

Issue #1328, opened 2022-07-19
KalmanFilter error: No measurement detected during the propagation


CarloVarni:
When running KF refitting on a tt-bar sample with pile-up (mu 200), ACTS reports the following error message:
```
ActsReFitterAlg.ActsKalmanFitter.Acts Kalman Refit  ERROR KalmanFilter failed: KalmanFitterError:5, No measurement detected during the propagation
```
This fit failure comes from https://github.com/acts-project/acts/blob/main/Core/include/Acts/TrackFitting/KalmanFitter.hpp#L1056 in which the KF tries to refit a fitted track and doesn’t find a single measurement. Despite being a fit failure, should this result in an error message in the KF itself?

This has been observed when running the `ActsReFitterAlg` algorithm in Athena, and does not happen for a single mu input sample. The presence of an `ERROR` message will mark the algorithm execution as a failure, thus preventing its use in a CI test.

paulgessinger:
This was recently observed again in https://gitlab.cern.ch/atlas/athena/-/merge_requests/59359

paulgessinger:
I investigated a bit, and this looks like another case of a navigation failure. From the RDO file that has 5 events, I believe it's these cases that error out:

![diag_00](https://user-images.githubusercontent.com/1058585/208451859-ddf66e47-427e-405d-8782-b82bac32d513.png)
![diag_01](https://user-images.githubusercontent.com/1058585/208451862-b6eee982-ce1f-4925-878c-c2ad62245d6f.png)
![diag_02](https://user-images.githubusercontent.com/1058585/208451864-7f020073-40b1-4b60-93ea-574058f7933f.png)
![diag_03](https://user-images.githubusercontent.com/1058585/208451866-37a4514d-a623-4bcd-821e-164b12c6c294.png)
![diag_04](https://user-images.githubusercontent.com/1058585/208451867-cfeabc7b-1a00-4879-8900-7b2f1d099c71.png)
![diag_05](https://user-images.githubusercontent.com/1058585/208451870-a0a141d6-5a0b-4ee3-8eb9-c9e323b847a9.png)

Seems to me like they start out very forward and then (somehow) bend in the longitudinal plane and fail to reach the beampipe cylinder boundary. The navigation then runs into infinity, which the KF interprets as not having found any measurements.

noemina:
Yes, it might be related to it. Have you already tried with including [this](https://github.com/acts-project/acts/pull/1722) in ACTS?

[...]

PR #1881, opened 2023-02-21
refactor: EigenStepper overstep limit is configurable

paulgessinger:
This was already a dynamic member, but hard-coded to 100um. This PR
makes this a constructor argument, but defaults it to the same value.


andiwand (approved):
looks good to me

paulgessinger:
/cc @noemina @CarloVarni

In my testing, this fixes #1328

PR #1883, opened 2023-02-22
fix: Really postselect particles in `addGeant4`

benjaminhuth:
The postselection used so far the configuration of the preselection.

Also, so far the post selection did not affect the particle writing. People who apply post-selection probably also want to export only the selected particles.

This PR also applies the post-selection to "particles_final", since it was only applied to "particles_initial" so far.


Examples/Python/python/acts/examples/simulation.py
```diff
@@ -625,16 +627,27 @@ def addGeant4(
...
         addParticleSelection(
             s,
-            preSelectParticles,
+            postSelectParticles,
             inputParticles=g4conf.outputParticlesInitial,
-            outputParticles=postSelectedParticles,
+            outputParticles=particlesInitial,
+        )
+
+        particlesFinal = "particles_final_selected"
```
andiwand, line 638: what is the use case for final particles selected?

Examples/Python/python/acts/examples/simulation.py
```diff
@@ -625,16 +627,27 @@ def addGeant4(
...
+            inputParticles=g4conf.outputParticlesFinal,
+            outputParticles=particlesFinal,
         )
-        s.addWhiteboardAlias("particles", postSelectedParticles)
     else:
-        s.addWhiteboardAlias("particles", alg.config.outputParticlesInitial)
+        particlesInitial = alg.config.outputParticlesInitial
+        particlesFinal = alg.config.outputParticlesFinal
+
+    # Only add alieas for 'particles_initial' as this is the one we use most
```
andiwand, line 649:
```suggestion
    # Only add alias for 'particles_initial' as this is the one we use most
```

Examples/Python/python/acts/examples/simulation.py
```diff
@@ -625,16 +627,27 @@ def addGeant4(
...
+    s.addWhiteboardAlias("particles", particlesInitial)
```
andiwand, line 650: I think we should reflect all these changes in `addFatras` too

Examples/Python/python/acts/examples/simulation.py
benjaminhuth: mainly that we write them out. In my case it makes quite a difference (over 100MB per event).

Examples/Python/python/acts/examples/simulation.py
benjaminhuth: yeah good point, I'll update it later today

Examples/Python/python/acts/examples/simulation.py
andiwand: that't a good point I didn't think of that :+1:

Examples/Python/python/acts/examples/simulation.py
asalzburger: Was this done?

andiwand (approved):
lets gooo

PR #1884, opened 2023-02-22
perf: Use inline functions

CarloVarni:
Move to inline functions


paulgessinger:
Does this make a difference? I was under the impression that `template`s get `inline` linkage automatically, and that the decision of whether or not to actually inline a function call is up to the compiler anyway.

LuisFelipeCoelho:
I tested in athena with 100 ttbar mu200 events. The average time per event before was 2.67s for PPP and 1.20s for SSS, and with inline I got 2.56s and 1.13s. @noemina 

ITkSeeding_ttbar_14TeV_mu200_barplot 2

ITkSeeding_ttbar_14TeV_mu200_barplot

LuisFelipeCoelho (approved):

asalzburger:
Do we understand the difference in the physics performance (and hence the update in the root files) if this is *simply* using code to inline - or did I miss something?

CarloVarni:
Hi @asalzburger we are not updating root files with this PR. Maybe you meant https://github.com/acts-project/acts/pull/1850 ?

asalzburger:
Ha, you are right, sorry!

asalzburger:
Merging this in, then.

PR #1885, opened 2023-02-22
refactor: align Kalman fitter function with GSF fitter function

andiwand:
Kalman fitter function was missing some features the GSF fitter already has


Examples/Algorithms/TrackFitting/src/KalmanFitterFunction.cpp
```diff
@@ -129,14 +130,16 @@ ActsExamples::makeKalmanFitterFunction(
   cfg.resolvePassive = false;
   cfg.resolveMaterial = true;
   cfg.resolveSensitive = true;
-  Acts::Navigator navigator(cfg);
-  Propagator propagator(stepper, std::move(navigator));
-  Fitter trackFitter(std::move(propagator));
+  Acts::Navigator navigator(cfg); // logger.cloneWithSuffix("Navigator")
```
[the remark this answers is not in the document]
andiwand: my problem with this was that setting verbose will flood your logs with navigator/propagator/stepper output and that was not useful in my case

happy to add it back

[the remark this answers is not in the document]
paulgessinger: Isn't that the point though? If you don't want the output, don't set it to VERBOSE?

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -368,7 +368,7 @@ def addTruthSmearing(
     rnd: acts.examples.RandomNumbers,
     selectedParticles: str,
     particleSmearingSigmas: ParticleSmearingSigmas,
-    initialVarInflation: float,
+    initialVarInflation: list[float],
```
paulgessinger, line 371:
```suggestion
    initialVarInflation: List[float],
```

Examples/Algorithms/TrackFitting/src/KalmanFitterFunction.cpp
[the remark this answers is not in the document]
andiwand: in my case I only wanted the kalman fitter to give verbose output. everything else was a ton of noise

Examples/Python/python/acts/examples/reconstruction.py
andiwand: hm looks like it actually is `list`?

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -368,7 +368,7 @@ def addTruthSmearing(
     rnd: acts.examples.RandomNumbers,
     selectedParticles: str,
     particleSmearingSigmas: ParticleSmearingSigmas,
-    initialVarInflation: float,
+    initialVarInflation: List[float],
```
andiwand, line 371:
```suggestion
    initialVarInflation: list[float],
```

Examples/Algorithms/TrackFitting/src/KalmanFitterFunction.cpp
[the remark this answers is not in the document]
benjaminhuth: Actually I would definitively propagate the VERBOSE through, because there are enough situations where the navigation log is the interesting one...
It's always possible to reduce down the output again with e.g. `./my_script.py | grep <logger name>`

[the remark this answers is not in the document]
andiwand: makes sense will do

Examples/Algorithms/TrackFitting/src/KalmanFitterFunction.cpp
```diff
@@ -129,14 +130,18 @@ ActsExamples::makeKalmanFitterFunction(
   cfg.resolvePassive = false;
   cfg.resolveMaterial = true;
   cfg.resolveSensitive = true;
-  Acts::Navigator navigator(cfg);
-  Propagator propagator(stepper, std::move(navigator));
-  Fitter trackFitter(std::move(propagator));
+  Acts::Navigator navigator(cfg, logger.cloneWithSuffix("Navigator"));
+  Propagator propagator(stepper, std::move(navigator),
+                        logger.cloneWithSuffix("Propagator"));
+  Fitter trackFitter(std::move(propagator), logger.cloneWithSuffix("Fitter"));
```
benjaminhuth, line 136: Maybe we could name the Fitter "KF" instead of "Fitter"? Same with "DirectKF" below. Might be clearer, and also is shorter.

andiwand: my idea here was that the parent logger already has the "Kalman" prefix so it will propagate through to the child loggers

andiwand: https://github.com/acts-project/acts/pull/1885/files#diff-a0346d9527c4778d50b91394c01430aad0f56675c5f533ca09503f85c7c827a1R77

benjaminhuth: Ah I see

benjaminhuth (approved):

PR #1887, opened 2023-02-22
fix: G4 detector for Fatras sim

andiwand:
Fatras did not see sensitive surfaces in a G4 detector because they were not wired correctly.

This PR adds a function on the core surface to allow attaching a detector element after construction which makes the surface sensitive. This also breaks the cyclic dependency in detector construction


asalzburger:
Still some build failures it seems.

andiwand:
fixed @asalzburger

asalzburger (approved):
Looks alright, let's go for it.

PR #1888, opened 2023-02-22
feat: enable geantino particle generation

andiwand:
Allows shooting geantino particles into the detector with our particle gun

Example gun:

```
addParticleGun(
    s,
    ParticleConfig(1, acts.PdgParticle.eInvalid, randomizeCharge=False, charge=1, mass=0),
    rnd=rnd,
)
```


andiwand:
in principle we could drop this completely now

```
    /// Force pdgCode & mass & charge in G4 units (this is needed for Geantino
    /// simulation)
    std::optional<G4int> forcedPdgCode;
    std::optional<G4double> forcedCharge;  // e.g. 1 for charged geantino
    std::optional<G4double> forcedMass;    // e.g. 0 for geantino
```

but we still depend on it in https://github.com/acts-project/acts/blob/main/Examples/Run/Geant4/Common/src/Geant4Common.cpp#L77

PR #1889, opened 2023-02-22
chore: improve coloring for residuals and pulls script

andiwand:
color of the fit used to be the same as the predicted which made the plots hard to read


benjaminhuth (approved):
looks good!

asalzburger (approved):

PR #1890, opened 2023-02-22
fix: RootMeasurementWriter std vs var

andiwand:
variance was filled with standard deviation in the RootMeasurementWriter


benjaminhuth (approved):
good catch!

asalzburger:
This will need an update of the reference histograms then, right ?

benjaminhuth (approved):

PR #1893, opened 2023-02-23
feat: added python bindings for SurfaceTypes

dimitra97:
This PR adds python enum bindings as `acts.SurfaceType`. Those will be needed in order to set the desired surface type for Geant4 translation, in case the auto-detection does not work (e.g. through the ambiguity if `G4Tubs` are `CylinderSurface` or `StrawSurface`.


benjaminhuth (approved):
Looks good!

andiwand:
@paulgessinger / @asalzburger I think we are missing some permissions for the CI here?

paulgessinger:
>     * set a `next` milestone, which means we consider it for the next release.

Annoyingly, GitHub's permission model only allows people with write access to set milestones, so one of us has to do it.

PR #1894, opened 2023-02-23
ci: Add hashcheck for Exa.TrkX

benjaminhuth:
self explaining


Examples/Python/tests/test_examples.py
```diff
@@ -1267,3 +1270,44 @@ def test_bfield_writing(tmp_path, seq, assert_root_hash):
...
+        url = "https://acts.web.cern.ch/ci/exatrkx/onnx_models_v01.tar"
+    else:
+        url = "https://acts.web.cern.ch/ci/exatrkx/torchscript_models_v01.tar"
+
+    tarfile_name = tmp_path / "models.tar"
+    urllib.request.urlretrieve(url, tarfile_name)
+    tarfile.open(tarfile_name).extractall(tmp_path)
+
+    print(os.listdir(tmp_path))
+
```
benjaminhuth, line 1291:
```suggestion
```

.gitlab-ci.yml
```diff
@@ -75,6 +75,7 @@ build:
       - src/Examples/Scripts/Python
       - src/Examples/Algorithms/TrackFinding/share
       - src/Examples/Algorithms/Digitization/share
+      - src/pytest.ini
```
paulgessinger, line 78: Why does this go to the artifact? So you don't have to clone again? Don't you also need the `conftest.py`?

paulgessinger (commented):
I like the idea, but I think copying around the source in the artifacts is a bit weird. Can't we just re-clone in the second stage and run the test suite normally?

I'm also torn between downloading the model files from python. I guess we can do it like this, but it does redownload them every local test run, this way. They're not that large though, right?

.gitlab-ci.yml
benjaminhuth: Yeah without this file there are quite a lot of warnings from pytest about unkown marks. But you're right, cloning again is the better way, so I'll do that instead.

benjaminhuth:
They are of size O(10MB)... Yeah the thing is in the current state the plugin makes assumptions about the directory structure of the `tmpdir`, and this was the easiest way to do that. 
In a future update we should also refactor the way the script works, but I didn't want to touch the script for this PR to make sure I just check the current state.

paulgessinger (approved):

PR #1897, opened 2023-02-24
refactor: and fix truth seeding

andiwand:
truth seeding seems to be quite broken at the moment

1. you could end up with less sees than particles which will crash the kalman filter downstream
2. the proto tracks given by the seeding seem to only contain the seeded hits which is no good for the kalman filter

with this PR I tried to improve things by splitting the truth seeding out of `TrackParamsEstimationAlgorithm` and restore its original purpose. I added `TruthSeedingAlgorithm` which now also does `TruthTrackFinder` in order to keep seeded particles and track params aligned


Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TruthSeedingAlgorithm.hpp
```diff
@@ -0,0 +1,78 @@
...
+    /// space points for strip-like detectors.
+    std::vector<std::string> inputSpacePoints;
+    /// Output successfully seeded truth particles.
+    std::string outputParticles;
+    /// Output full proto track collection.
+    std::string outputFullProtoTracks;
+    /// Output seed collection.
+    std::string outputSeeds;
+    /// Output proto track collection.
+    std::string outputProtoTracks;
```
benjaminhuth, line 52: Here I'm not sure about the naming... I would rather go for something like: 
* `outputFullProtoTracks` -> `outputProtoTracks`
* `outputProtoTracks` -> `output3HitProtoTrack` / `outputThreeHitProtoTrack`

My natural assumption would be that a standard proto-track is a whole track. This also makes more explicit what in the short one is.

benjaminhuth (commented):
Really nice changes, I also always had the impression that this`TrackParamEstimationAlgorithm` served way to many purposes! I just have one minor comment about naming.

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TruthSeedingAlgorithm.hpp
andiwand: I thought the same first but then I realized that our other seeders also use `prototracks` to communicate the 3 hits which went into the estimate

that said - not sure what the intention behind `prototracks`. maybe using it for truth tracking is the exception here?

/cc @paulgessinger @asalzburger @LuisFelipeCoelho

asalzburger: I think it's fine to use ProtoTracks for seeds as well if they can serve the needs, it's probably just a special case for the truth tracking / refitting, I would say

benjaminhuth:
I think we should try to make this refactoring in a way that the hashes stay the same, right? Currently quite a lot of hash checks still fail.

andiwand:
will try to do that. physmon looks also kind of sketchy. but I wonder if we also did 3 hit fitting in the tests

benjaminhuth:
If we do 3 hit fitting, maybe change that in a follow-up PR? that way we can be sure that we have no change in behaviour in this refactoring...

andiwand:
this should be ready now @benjaminhuth @asalzburger

benjaminhuth (approved):
Very nice refactoring, let's get it in!

PR #1901, opened 2023-02-28
refactor: Unify readers, writers and algorithms

paulgessinger:
There's no strong reason to have these as separate objects. This PR will change outputs because of the random number seed generation, see #1905.

Blocked by: 
- #1905


Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -67,8 +67,9 @@ void ActsExamples::Sequencer::addReader(std::shared_ptr<IReader> reader) {
   if (not reader) {
     throw std::invalid_argument("Can not add empty/NULL reader");
   }
-  m_readers.push_back(std::move(reader));
+  m_readers.push_back(reader);
```
andiwand, line 70:
```suggestion
```

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -273,7 +270,6 @@ int ActsExamples::Sequencer::run() {
   ACTS_INFO("  " << m_decorators.size() << " context decorators");
   ACTS_INFO("  " << m_readers.size() << " readers");
```
andiwand, line 271:
```suggestion
```

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -67,8 +67,9 @@ void ActsExamples::Sequencer::addReader(std::shared_ptr<IReader> reader) {
   if (not reader) {
     throw std::invalid_argument("Can not add empty/NULL reader");
   }
-  m_readers.push_back(std::move(reader));
+  m_readers.push_back(reader);
   ACTS_INFO("Added reader '" << m_readers.back()->name() << "'");
```
andiwand, line 71:
```suggestion
```

Examples/Framework/include/ActsExamples/Framework/Sequencer.hpp
```diff
@@ -123,7 +123,6 @@ class Sequencer {
   std::vector<std::shared_ptr<IContextDecorator>> m_decorators;
   std::vector<std::shared_ptr<IReader>> m_readers;
```
andiwand, line 124:
```suggestion
```

Examples/Framework/src/Framework/Sequencer.cpp
andiwand: oh I seed I though this is a leftover. ignore my comments then :smile:

andiwand:
looks like only the truth tracking was affected by the change in randomness. I wonder if it would make sense to change it to truth estimated because that gives the optimal seeding results.

I also though a little about the RNG and I think I would prefer removing the algorithm id from the seed because it seems to do more harm then good? I would not expect funny correlations between the simulation algorithms. but that would definitely be something for a different PR

Examples/Framework/include/ActsExamples/Framework/IWriter.hpp
```diff
@@ -24,18 +25,19 @@ namespace ActsExamples {
...
-  virtual std::string name() const = 0;
-
   /// Write data from one event.
   virtual ProcessCode write(const AlgorithmContext& context) = 0;
 
-  /// End the run (e.g. aggregate statistics, write down output, close files).
-  virtual ProcessCode endRun() = 0;
+  /// Internal execute method forwards to the write method as mutable
+  /// @param context The algorithm context
+  ProcessCode internalExecute(const AlgorithmContext& context) override {
```
benjaminhuth, line 35: Maybe mark this `final`?

Examples/Framework/include/ActsExamples/Framework/IReader.hpp
```diff
@@ -39,6 +35,18 @@ class IReader {
   /// will most likely not be called in order. Implementations must use the
   /// event number provided to select the proper data to be read.
   virtual ProcessCode read(const AlgorithmContext& context) = 0;
+
+  /// Internal execute method forwards to the read method as mutable
+  /// @param context The algorithm context
+  ProcessCode internalExecute(const AlgorithmContext& context) override {
```
benjaminhuth, line 41: Maybe mark this `final`?

benjaminhuth (commented):
We could make the naming a bit more consistent maybe:
`IAlgorithm` -> `SequencerModule` (?)
`BareAlgorithm` -> `IAlgorithm`

Otherwise, I find it a bit confusing and also ambiguous if we talk about "algorithms", do we mean and implementation of `BareAlgorithm` or the a general thing plugged into the sequencer.

paulgessinger:
> We could make the naming a bit more consistent maybe: `IAlgorithm` -> `SequencerModule` (?) `BareAlgorithm` -> `IAlgorithm`
> 
> Otherwise, I find it a bit confusing and also ambiguous if we talk about "algorithms", do we mean and implementation of `BareAlgorithm` or the a general thing plugged into the sequencer.

Yeah we could do that I suppose.

paulgessinger:
Ok:

- `IAlgorithm` becomes `SequenceElement`
- `BareAlgorithm` becomes `IAlgorithm`
- `Service` goes away

What do you think @benjaminhuth @andiwand?

benjaminhuth:
Some pretty strange errors with DD4hep in the examples test. Looks like some multi-threading problem? But no clue how this could be triggered by the changes in the PR to the `DD4hepService`...

andiwand:
the beauty of dd4hep :heart: 

this might be the cause?

```
IndexError: Object 'mapped_material_tracks' does not exists, similar ones are: [ 'material-tracks' ]
```

but I didnt see how this can happen with the code changes

paulgessinger:
I'm guessing something changes in the order and it's now not in the whiteboard when requested.

paulgessinger:
Ok found it. Works locally now, let's see what the machine says.

paulgessinger:
Ok, all green now, can you approve again @benjaminhuth or @andiwand ?

andiwand (approved):
lets go

andiwand (approved):

PR #1904, opened 2023-02-28
feat: GSF backward-pass misses as outliers

benjaminhuth:
Now, the GSF does flag states as outliers that were missed at the backward pass.


andiwand (approved):
looks good to me :+1:

benjaminhuth:
Changes in physmon as expected

image

PR #1905, opened 2023-03-01
refactor: make RNG independent of algorithm order

andiwand:
after discussion with paul I though it would be beneficial to remove the dependence of the algorithm order of the RNG in the examples framework. changing the order of reader/writer/algorithms used to result in different outputs which might be unexpected and hard to debug

I believe the impact on the randomness in the simulation is negligible and should not result in funny correlations


paulgessinger:
/cc @HadrienG2 

Do we think it's a problem if all algorithms get the same random number sequence?

HadrienG2:
That would result in correlations, which would be directly visible if two copies of the same (or similar) algorithms are run, but I don't think we currently do that often. On different algorithms, I expect the effect to be more subtle.

andiwand:
will do :+1: at the same time we should then get a 1:1 in your branch if we merge this first

paulgessinger:
Docker error?

andiwand:
something is off with the CI and I just got more conflicts with main :smile:

paulgessinger:
Ok I think the CI is back now.

paulgessinger:
`test_full_chain_odd_example_pythia_geant4` still fails.

benjaminhuth:
Just one thought: Would it be useful to make this optional and off by default?
In case someone really wants to run actually two simulations for some reason...

andiwand:
> Just one thought: Would it be useful to make this optional and off by default? In case someone really wants to run actually two simulations for some reason...

I think it would be easier to just pass in two different RNG to the algorithms in this case

andiwand:
> `test_full_chain_odd_example_pythia_geant4` still fails.

oh no... we have another propagation error with the changed seed

```
10:21:12    Propagator     ERROR     Propagation reached the step count limit of 10000 (did 10000 steps)
10:21:12    CKF            ERROR     Propapation failed: PropagatorError:3 Propagation reached the configured maximum number of steps with the initial parameters: 
 8.31541
-26.5624
-2.46935
0.751848
-1.74987
  18.233
10:21:12    TrackFinding   WARNING   Track finding failed for seed 10638 with errorPropagatorError:3
```

andiwand:
increasing the iteration limit again seems to fix the problem. lets see if this passes this time

paulgessinger (approved):
lgtm

PR #1907, opened 2023-03-01
fix: remove deleted track selector params in python

andiwand:
after https://github.com/acts-project/acts/pull/1874 some track selector params were not wired correctly


tboldagh (approved):
It looks ok Andreas. Thanks

paulgessinger (approved):

PR #1908, opened 2023-03-01
refactor: Examples sequence elements gain Read/WriteDataHandles

paulgessinger:
This locks in the type at algorithm construction, registers the handles and allows checking the data flow more thoroughly before running the event sequence. 

This PR only changes the components used in the ODD chain example, but we should probably extend this to the rest of the Examples components as well.

Blocked by:
- #1901


paulgessinger:
This should be ready for review now, @benjaminhuth @andiwand

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -52,6 +61,22 @@ size_t saturatedAdd(size_t a, size_t b) {
...
+
+  if (status != 0) {
+    return sym;
+  }
+
+  return realname;
+#else
+  return sym;
+#endif
+}
```
andiwand, line 79: should we put this in a central utility header? might be useful in other places

Examples/Framework/include/ActsExamples/Framework/IAlgorithm.hpp
```diff
@@ -20,10 +20,11 @@
 
 #include <memory>
 #include <string>
+#include <vector>
```
andiwand, line 23: unused?

Examples/Framework/include/ActsExamples/Framework/WhiteBoard.hpp
```diff
@@ -8,6 +8,7 @@
 
 #pragma once
 
+#include "ActsExamples/Framework/IAlgorithm.hpp"
```
andiwand, line 11: unused?

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -106,17 +132,108 @@ void Sequencer::addElement(std::shared_ptr<SequenceElement> element) {
   if (not element) {
     throw std::invalid_argument("Can not add empty/NULL element");
   }
+
+  bool valid = true;
+  std::string elementType{getAlgorithmType(*element)};
```
andiwand, line 137:
```suggestion
  std::string elementType = getAlgorithmType(*element);
```
for consistency

Examples/Io/Root/include/ActsExamples/Io/Root/RootVertexPerformanceWriter.hpp
```diff
@@ -125,6 +128,26 @@ class RootVertexPerformanceWriter final
...
   int getNumberOfTruePriVertices(const SimParticleContainer& collection) const;
+
+  ReadDataHandle<SimParticleContainer> m_inputAllTruthParticles{
+      this, "InputAllTruthParticles"};
+
+  ReadDataHandle<SimParticleContainer> m_inputSelectedTruthParticles{
+      this, "InputSelectedTruthParticles"};
+
+  ReadDataHandle<std::vector<Acts::BoundTrackParameters>>
+      m_inputTrackParameters{this, "InputTrackParameters"};
```
andiwand, line 139: would it be possible to make this handle optional and decide in the constructor if it should be initialized or not?

i think it would be nice to carry the optional inputs/outputs through like this so they don't even end up in the listed ones for the sequencer element

andiwand (commented):
gave it a quick look. makes a lot of sense to me!

Examples/Framework/src/Framework/Sequencer.cpp
paulgessinger: Doesn't actually work because `getAlgorithmType` returns `string_view`. Do you think I should change that too?

Examples/Io/Root/include/ActsExamples/Io/Root/RootVertexPerformanceWriter.hpp
paulgessinger: That's exactly how it's implemented is actually.

Examples/Framework/src/Framework/Sequencer.cpp
paulgessinger: I'm removing this entirely in favor of `boost::core::demangle`.

Examples/Framework/src/Framework/Sequencer.cpp
andiwand: oh okay no I think it is fine then

Examples/Framework/src/Framework/WhiteBoard.cpp
```diff
@@ -12,6 +12,7 @@
 #include <string_view>
 
 #include <Eigen/Core>
+#include <boost/core/demangle.hpp>
```
andiwand, line 15: unused?

andiwand (approved):
nice! looks good to me :+1:

paulgessinger:
Ok, I've added a config flag to the sequencer that defaults to False for the checking functionality. I'll default it to true with the other PR I was mentioning. Thoughts?

andiwand:
sounds good to me :+1:

PR #1909, opened 2023-03-01
refactor!: MTJ get/set SourceLink by value

paulgessinger:
This changes the return type and argument for the getter and setter to source links by value. This decouples the interface one more step from the backend.

I also renamed `uncalibratedSourceLink` to getUncalibratedSourceLink` to make this a bit more clear.

I also fully remove the `calibratedSourceLink` component of MTJ. We didn't use this, and I think we won't.

BREAKING CHANGE: The `SourceLink` returned from `uncalibratedSourceLink` and `getUncalibratedSourceLink` is now a value. If you have a construct / access pattern like
```cpp
const auto& sl = ts.getUncalibratedSourceLink().template get<MySourceLink>();
```
you will now have a dangling reference. You'll need to change `const auto&` to `auto` to keep the `SourceLink` valid.


Core/include/Acts/EventData/MultiTrajectory.ipp
```diff
@@ -77,20 +77,15 @@ inline auto TrackStateProxy<D, M, ReadOnly>::projector() const -> Projector {
...
-inline auto TrackStateProxy<D, M, ReadOnly>::calibratedSourceLink() const
-    -> const SourceLink& {
-  assert(has<hashString("calibratedSourceLink")>());
-  return component<std::optional<SourceLink>,
-                   hashString("calibratedSourceLink")>()
-      .value();
+inline auto TrackStateProxy<D, M, ReadOnly>::getUncalibratedSourceLink() const
+    -> SourceLink {
+  assert(has<hashString("uncalibratedSourceLink")>());
+  return m_traj->getUncalibratedSourceLink(m_istate);
```
tboldagh, line 88: Misisng `_impl`?

Core/include/Acts/EventData/VectorMultiTrajectory.hpp
```diff
@@ -438,6 +435,10 @@ class VectorMultiTrajectory final
     m_measCov.resize(m_measCov.size() + measdim * measdim);
   }
 
+  void setUncalibratedSourceLink_impl(IndexType istate, SourceLink sourceLink) {
```
tboldagh, line 438: Actually now, when I look at it, shouldn't it be `const SourceLink&`?

tboldagh (approved):
Just two small comments / questions realy.

Core/include/Acts/EventData/VectorMultiTrajectory.hpp
paulgessinger: I move it into the other function call. I suppose we could keep it a const ref and then we copy once in the backend, but I thought it might be a good idea to support moving all the way. 

What do you prefer?

Core/include/Acts/EventData/MultiTrajectory.ipp
paulgessinger: This is still in the MTJ, there's a function in the `.hpp` that in turn calls the `_impl()` method.

Core/include/Acts/EventData/VectorMultiTrajectory.hpp
tboldagh: Ah, it ha been years, but I still forget about moving. All good as is.

Core/include/Acts/EventData/MultiTrajectory.ipp
tboldagh: A, yes, this is in proxy. This header is quite long ...

tboldagh (approved):

paulgessinger:
@tboldagh can you reapprove if you're happy?

CarloVarni (commented):
The changes look good. I only have one comment on the code

Core/include/Acts/EventData/MultiTrajectory.hpp
```diff
@@ -732,14 +715,18 @@ class TrackStateProxy {
 
   /// Uncalibrated measurement in the form of a source link. Const version
   /// @return The uncalibrated measurement source link
-  const SourceLink& uncalibratedSourceLink() const;
+  [[deprecated("Use getUncalibratedSourceLink() instead")]] SourceLink
+  uncalibratedSourceLink() const;
```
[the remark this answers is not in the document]
asalzburger: Good question, indeed. We've never really discussed a policy on that. 

Particularly in this case, if you keep the legacy version you increase the chance of someone ending up with a dangling reference. By forcing to change, the return type could be updated accordingly perhaps?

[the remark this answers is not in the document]
paulgessinger: I was debating whether to do this. Generally I think we'll have to transition to preserving backward compatibility more, and indeed this still compiles, but you're right it might be a problem. In theory, `const&` would prolong lifetime of a temporary returned here, but I've seen it crash, so yeah, might be better to remove the overload altogether.

tboldagh (approved):
I am happy with these changes.

Core/include/Acts/EventData/MultiTrajectory.hpp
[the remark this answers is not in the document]
paulgessinger: Ok, I dropped this function completely now.

paulgessinger:
I pushed a new commit as a response to @CarloVarni's feedback. Can one of you click approve one more time?

paulgessinger:
Oh, seems I didn't replace some `uncalibratedSourceLink` calls. I thought they'd have shown up as warnings and failed the CI. I'll fix this now.

paulgessinger:
They entered from the main branch. Should be fixed now.

PR #1910, opened 2023-03-02
refactor: seeding always sorts SPs

LuisFelipeCoelho:
Sorting has been shown to be less expensive then looping over all elements to select bottom and top sps for seeds. By sorting the SPs, we can avoid a considerable number of iterations, especially for high pile-up.
This PR removes the option to not sort the SPs during the seeding algorithm. @noemina


Core/include/Acts/Seeding/BinnedSPGroup.ipp
```diff
@@ -81,17 +81,14 @@ Acts::BinnedSPGroup<external_spacepoint_t>::BinnedSPGroup(
...
-          });
-    }
+  // sort SPs in R for each (z, phi) grid bin
+  for (auto& rbin : rBins) {
+    std::sort(
+        rbin.begin(), rbin.end(),
+        [](std::unique_ptr<InternalSpacePoint<external_spacepoint_t>>& a,
+           std::unique_ptr<InternalSpacePoint<external_spacepoint_t>>& b) {
+          return a->radius() < b->radius();
+        });
```
CarloVarni, line 91: Good idea. Indeed the same point I was thinking. 

Since we want to sort the SPs anyway I'd suggest the following:
- avoid the creation of a temporary r-binned collection of internal space points (that is then moved to the grid) and directly fill the grid with the internal space points on the fly
- keep track of the global bins in the grid that we have touched (a simple vector of `std::size_t` is ok)
- run on these indexes and sort ONLY these bins

On top of this idea, another (possibly better) addition would be:
- pass a reference of the external space point collection to BinnedSPGroup instead of the pointers to the beginning and end of it
- make a vector of indices from that and sort it according to the SP radius
- Run on these indexes, get the corresponding space point and create the internal space point 

With this second addition, you will not need anymore of keep track of the global bins since, by definition, the space points will be automatically sorted

LuisFelipeCoelho: Hi @CarloVarni  I tried to implement your first suggestion, but I used a set of `size_t` instead of a vector to avoid duplicates. I am not sure if this is the best option, let me know what you think.

With your second sugestion, I did not understand how the SPs in each r-bin will be sorted

CarloVarni: the set will use heap allocations. What about initializing a `std::vector<bool> changed_bins(nBins, false)` instead?
Every time you modify bin N, then you set `changed_bins[N] = true`. Once you are done, you can run on this vector to find out what are the bins you have modified

andiwand: is `boost::flat_set` an option?

andiwand: otherwise depending on the size of the collection you could use still use a vector and just check if the item is there already before inserting. sounds slow but should be faster than other data structures given the collections is sufficiently small

CarloVarni: Concerning your second question. If you sort the input vector then the sps you insert in the grid will keep the ordering. So that in bin N the sps will automatically be in the correct ordering.

Something like:
```cpp
BinnesSPgroup(const space_point_container_t& sps, <other inputs>) {
   std::vector<std::size_t> indexes(sps.size());
  for (std::size_t i(0); i<indexes.size(); ++i) indexes[i] = i;

// Sort the space points
  std::sort(indexes.begin(), indexes.end(), 
       [&sps] (const std::size_t& a, const std::size_t& b) 
           { return sps[a].radius() < sps[b].radius(); });

for (const auto& idx : indexes) {
const auto& sp = sps[idx];
// create the internal space point from it
auto internal_sp = <create internal space point>;
// fill the grid
grid->Fill(internal_sp);
}
}
```

CarloVarni: on this last point, we'd need to check what is faster: sorting the entire collection once, or sort the modified bins in the grid... 

I suspect the latter ... but better check

LuisFelipeCoelho: Thank you for your suggestions! I tested both strategies and apparently the first one is faster, I think this is because in the first strategy we remove some spacepoints that are outside z and phi region before sorting. 
I could also try to loop over the SPs and remove the unwanted ones, then sort all the indices at once, then loop agains and fill the grid. @CarloVarni 

This is the execution time for binnedSPgroups.ipp
image

CarloVarni: I would directly go with the current implementation. Looping on SPs and remove them and then sort sounds quite expensive

CarloVarni (approved):

PR #1913, opened 2023-03-02
refactor: Seeding algo level bin finders

paulgessinger:
Currently the SeedingAlgorithm creates the bin finders on every event. I don't think that's needed, but for const-reasons I have to change their argument types to be const everywhere.


paulgessinger:
Any objections @LuisFelipeCoelho?

LuisFelipeCoelho (approved):
looks good to me

PR #1914, opened 2023-03-02
refactor: Use pointer-to-impl pattern to improve algorithms

paulgessinger:
Currently, the vertexing algorithms recreate their tools every event. In order not having to put all the types in the header, using the pointer-to-impl pattern, we can keep them in the cpp file.

Not sure if this is a good pattern overall, but I wanted to try it out. Thoughts?


paulgessinger:
I was missing some changes to the AMVF config that were introduced while this branch was open.

paulgessinger:
Actually, after thinking more about this, maybe this isn't the best idea after all.

PR #1915, opened 2023-03-02
test: Run pytest with xdist on multiple procs

paulgessinger:
Install pytest-xdist in the CI and run pytest with `-n auto`. This will give us 2 jobs on GitHub runners, which is not a lot, but should allow long tasks to block execution a bit less.


andiwand (approved):
:+1:

paulgessinger:
This worked fine over on my fork...

PR #1917, opened 2023-03-03
docs: Machine Learning based Ambiguity Solver

Corentin-Allaire:
This add some doc for the onnx pluging and the ML ambiguity solver.


docs/plugins/MLAlgorithms.md
```diff
@@ -0,0 +1,38 @@
...
+
+The goal of the ambiguity solver is to remove duplicated and fake tracks that remains after the CKF. To perform this cleaning, this algorithm works in three steps :
+
+- Clustering : track are clustered together, one cluster ~ one truth particle
+- Ranking : tracks in each cluster are ranked, the best one is kept 
+- Cleaning : last pass over all the remain track to remove duplicate and fake (not implemented yet)
+
+### Clustering
+
+The clustering is implemented with the `clusterTracks` function. Its input is a multimap of pair of track ID and vector of measurement ID. The multimap uses the number of measurement associated with the tracks as a key, this is only a trick to sort efficiency the tracks by number of measurements. Then for each track, starting with the one with the most measurements, we check if a cluster share a hits with the track. If no, we create a new cluster and associate all the hits of the current track with the cluster. If yes, then the track is added to that cluster (note that the hits associated to the cluster doesn't change here). After looping over all the tracks, each of them should have been associated with a cluster.
```
andiwand, line 24:
```suggestion
The clustering is implemented with the `clusterTracks` function. Its input is a multimap of pair of track ID and vector of measurement ID. The multimap uses the number of measurement associated with the tracks as a key, this is only a trick to sort the tracks by the number of measurements efficiently. Then for each track, starting with the one which has the most measurements, we check if a cluster shares a hits with the track. If not, we create a new cluster and associate all the hits of the current track with the cluster. If yes, the track is added to that cluster (note that the hits associated to the cluster doesn't change here). After looping over all the tracks, each of them should have been associated with a cluster.   
```

docs/plugins/MLAlgorithms.md
```diff
@@ -0,0 +1,38 @@
...
+- Cleaning : last pass over all the remain track to remove duplicate and fake (not implemented yet)
+
+### Clustering
+
+The clustering is implemented with the `clusterTracks` function. Its input is a multimap of pair of track ID and vector of measurement ID. The multimap uses the number of measurement associated with the tracks as a key, this is only a trick to sort efficiency the tracks by number of measurements. Then for each track, starting with the one with the most measurements, we check if a cluster share a hits with the track. If no, we create a new cluster and associate all the hits of the current track with the cluster. If yes, then the track is added to that cluster (note that the hits associated to the cluster doesn't change here). After looping over all the tracks, each of them should have been associated with a cluster.   
+
+### Ranking
+
+At this step we have multiple cluster of tracks. We use a NN to compute a score for each track, the closest the score to 1, the more confident we are the track is the good track (the best one associated with the truth particle). Then for each cluster, we select the track with the highest score.
```
andiwand, line 28:
```suggestion
At this step we have multiple clusters of tracks. We use a NN to compute a score for each track, the closer the score is to 1, the more confident we are that the track is the best track (the best one associated with the truth particle). Then for each cluster, we select the track with the highest score.
```

docs/plugins/MLAlgorithms.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Cleaning
+
+Finally, in the remaining tracks, there might be some fakes and duplicate remaining. With the first tests it doesn't seem to be the case, so the current implementation stops here. But in the future if some configuration encounter too many fake/duplicate after the ranking as simple classification NN could be used to separate the good tracks from the others.
```
andiwand, line 32:
```suggestion
Finally, for the remaining tracks, there might be some fakes and duplicates. With the first tests it doesn't seem to be the case, so the current implementation stops here. But in the future if some configuration encounters too many fake/duplicate after the ranking a simple classification NN could be used to separate the good tracks from the others. 
```

docs/plugins/MLAlgorithms.md
```diff
@@ -0,0 +1,38 @@
...
+
+### How to use 
+
+The ML based Ambiguity solver comes with a pre-trained model to be used with the ODD. To use it, it is extremely simple, just call the `full_chain_odd.py` with the `--MLSolver` option. If you want to try this solver with another detector, a few more step are needed. First you need to train a model, as an input you will need the multitrajectory from the CKF (run a full chain up to the CKF then use the `CSVWriter`). At least 100 ttbar events with 200 PU are needed (1000 would be ideal). You can then run `Examples/Scripts/Python/MLAmbiguityResolution/train_ambiguity_solver.py ` on our dataset, 2/3 of the data should be used for the training and 1/3 for the validation. You will receive a `duplicateClassifier.onnx` file in output you can now use this file as an input for the AmbiguityResolutionMLAlgorithm to use it with your detector.
```
andiwand, line 36:
```suggestion
The ML based Ambiguity solver comes with a pre-trained model to be used with the ODD. Using it is extremely simple, just call the `full_chain_odd.py` with the `--MLSolver` option. If you want to try this solver with another detector, a few more step are needed. First you need to train a model. As an input you will need the multitrajectory output from the CKF (run a full chain up to the CKF then use the `CSVWriter`). At least 100 ttbar events with 200 PU are needed (1000 would be ideal). You can then run `Examples/Scripts/Python/MLAmbiguityResolution/train_ambiguity_solver.py ` on our dataset, 2/3 of the data should be used for the training and 1/3 for the validation. You will receive a `duplicateClassifier.onnx` file as an output which now can be used as an input for the `AmbiguityResolutionMLAlgorithm` with your detector.
```

andiwand (approved):
thanks for the doc! left some commends - as always feel free to ignore them - happy to merge this in

Corentin-Allaire:
> thanks for the doc! left some commends - as always feel free to ignore them - happy to merge this in

Thanks a lot @andiwand !

PR #1919, opened 2023-03-04
refactor!: Rewrite Binned SP Group

CarloVarni:
All credits go to @LuisFelipeCoelho 

I have integrated Luis' code changes to the seeding algorithm. The result should be a considerable speed up of the seeding execution. I've added a few changes w.r.t. Luis' implementation

Summary of the main changes:
- Neighbourhood (and its iterator) class has been removed
- BinnedSPGroup (and its iterator) class has been changed
- Minor additional changes for code optimization (effect is very small, or even null)

The code now runs on the grid bins. When it finds a not empty bin it stops and returns three vectors. These vectors contain indexes to grid bins and they represent the bottom, middle and top space point candidates. Seeding is run on these candidates (you run on the indexes and then ask the grid to retrieve the elements in that specific bin). Once done, we move to the next not-empty bin.

Note: input collections to `createSeedsForGroup` are now `boost::container::small_vector<size_t, 9>` instead of Neightbourhood objects. Much simpler, cleaner and faster.

/cc @noemina


CarloVarni:
@LuisFelipeCoelho can you check how this implementation compares to yours in terms of timings?

CarloVarni:
> @CarloVarni Thank you very much for this PR! The time improvement seems to agree with my implementation, but this one is definitely much nicer. One comment is that it is affecting the efficiency and maybe we want to understand why. I believe we are not supposed to change the physics performance with this, right?

indeed no performance change should be expected. I need to understand what is going on

CarloVarni:
@LuisFelipeCoelho the performance difference were o bug in the navigation on the local bins of the grid. It has been solved and this PR is ready to go. 

@LuisFelipeCoelho @paulgessinger can you take a look at this? 

I see other PR on the seeding already in place and I'd like to prefer avoiding merge conflicts on this one

LuisFelipeCoelho (approved):

PR #1922, opened 2023-03-06
feat: Move sort of vector outside of transformCoordinates

CarloVarni:


LuisFelipeCoelho (approved):
It looks good to me. Do you think it is worth measuring if the resize change improves CPU? Also, in `filterSeeds_2SpFixed` we do something similar with `compatibleSeedR` vector and maybe could be changed by resize + operator[] as well (not necessarily in this PR)

CarloVarni:
Not sure we gain anything from the resize operation. Or at least, not enough to be considered an improvement. I'm afraid the size of the vector is just to small to make a difference.

Concerning `filterSeeds_2SpFixed` I would argue that the function takes a very small amount of CPU time already compared to the other big CPU consumers in the seeding. That would make no difference I'm afraid

LuisFelipeCoelho (approved):

PR #1923, opened 2023-03-07
feat: Adding come const

CarloVarni:
Trying to move to `const` in the seeding... this is the first step but cannot yet go further than this

### DIFF
--- a/Core/include/Acts/Seeding/SeedFilter.ipp
+++ b/Core/include/Acts/Seeding/SeedFilter.ipp
@@ -1,3 +1,4 @@
+// -*- C++ -*-
 // This file is part of the Acts project.
 //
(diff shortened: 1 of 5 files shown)


Core/include/Acts/Seeding/SeedFilter.ipp
```diff
@@ -72,7 +73,7 @@ void SeedFilter<external_spacepoint_t>::filterSeeds_2SpFixed(
 
   size_t beginCompTopIndex = 0;
   // loop over top SPs and other compatible top SP candidates
-  for (auto& topSPIndex : topSPIndexVec) {
+  for (const std::size_t& topSPIndex : topSPIndexVec) {
```
paulgessinger, line 76: Does that even make sense? `size_t` is the same size as a pointer, so maybe just take a copy?

PR #1926, opened 2023-03-08
feat: Iterate only on what we need while looking for compatible bottoms and tops

CarloVarni:
This is build on top of https://github.com/acts-project/acts/pull/1919 and also assumes the space points in the grid bins are already sorted in radius. Thus https://github.com/acts-project/acts/pull/1910 is required as well

Idea here is that in a tt-bar sample the grid bins may be very populated. However, the grid collects space points according to their z-phi values, with no consideration for the r-coordinate.

While looking for compatible bottoms and tops candidates we iterate on all the space points in the selected grid bins and this is a huge waste of time since just a few percentage of the space points are to be considerate. Of course, we check the distance of the candidate w.r.t. the middle space point... yet it is a O(N) operation.

With this change I'm introducing here (excluding what's already in https://github.com/acts-project/acts/pull/1919) I first performs a binary search on the space points candidates to identify those in the proper radius range. This should be a O(logN) operation. And only then I run on them. 

Since `getCompatibleDoublets` is one of the main contributors in the seeding time, this may improve the CPU performance


CarloVarni:
From my local tests this should introduce a ~5% increase in the CPU timing

@paulgessinger @LuisFelipeCoelho @noemina let me know if you have further ideas on this?

LuisFelipeCoelho:
@CarloVarni that is interesting, I was expecting it to decrease the time 🙁. But I believe we need to continue working on these iterations since this seems to be the part of the seeding where we have the biggest time difference from Athena

In Athena, for some reason, they loop over the SPs once to remove the ones outside the radius region of interest, then they loop again to apply the other cuts, not sure if that would be any different from what we do
https://acode-browser1.usatlas.bnl.gov/lxr/source/athena/InnerDetector/InDetRecTools/SiSpacePointsSeedTool_xk/src/ITkSiSpacePointsSeedMaker.cxx#2076

https://acode-browser1.usatlas.bnl.gov/lxr/source/athena/InnerDetector/InDetRecTools/SiSpacePointsSeedTool_xk/src/ITkSiSpacePointsSeedMaker.cxx#2135

CarloVarni:
Damn me... I wanted to write decrease but I wrote increase ... 

To be clear: this DECREASES the CPU time by ~4-5%

LuisFelipeCoelho:
Ah! right! that is good then 😂

CarloVarni:
Ok, no surprise this change buys us so little time. With the current configuration:
```console
deltaRMinBottomSP: 6
deltaRMaxBottomSP: 120
deltaRMinTopSP: 6
deltaRMaxTopSP: 280
```
we are basically considering the entire pixel detector range.

The  good news is that we still gains something in this configuration. More strict values for these parameters will make us gain more (the smaller the allowed range, the greater the gain w.r.t. previous approach)

LuisFelipeCoelho (approved):

PR #1927, opened 2023-03-08
fix: Compute complete covariance for max weight reduction

benjaminhuth:
In case of returning the parameters of the component with the largest weight, it is of course not correct to return also the covariance matrix of this component, since it does not reflect the uncertainty of the whole distribution. Rather, return the the covariance computed from all components.


benjaminhuth:
This should be again a improvement:

image

andiwand (approved):

PR #1928, opened 2023-03-08
feat: GSF add final multi-component state to track container

benjaminhuth:


Core/include/Acts/TrackFitting/GaussianSumFitter.hpp
```diff
@@ -440,6 +440,15 @@ struct GaussianSumFitter {
       track.parameters() = params.parameters();
       track.covariance() = params.covariance().value();
       track.setReferenceSurface(params.referenceSurface().getSharedPtr());
+      
+      using MultiPars = std::optional<Acts::MultiComponentBoundTrackParameters<SinglyCharged>>;
+      const std::string key = "multiComponentState";
+      if(not trackContainer.hasColumn(key)) {
+        trackContainer.template addColumn<MultiPars>(key);
+      }
```
paulgessinger, line 448: Now here's a big question: should the tools / fitters add columns, or should they only react to the columns being there, and skip writing the data if they're not?

benjaminhuth: Good question. Probably the "reaction" way is better if we think of application in Athena or something else, because they might want to opt out of this.

paulgessinger: Makes sense to me.

paulgessinger:
Is this useful for debugging only, or also for primary outputs?

benjaminhuth:
Hmm so in principle this is not only for debugging, e.g. it could be useful in LDMX, but also I have some plans myself to try some things with this.

However, before that I will need to add a proper writer... But I would do this seperately, since I'm not yet sure how to approach this.

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -794,5 +794,44 @@ struct GsfActor {
   }
 };
 
+/// A actor that collects the final multi component state once the propagation
+/// finished
```
andiwand, line 798:
```suggestion
/// An actor that collects the final multi component state once the propagation
/// finished
```

docs/core/track_fitting.md
```diff
@@ -93,7 +93,9 @@ outline:
 
 The fit can be customized with several options, e.g., the maximum number of components. All options can be found in the {struct}`Acts::GsfOptions`.
 
-To simplify integration, the GSF returns a {class}`Acts::KalmanFitterResult` object, the same as the {class}`Acts::KalmanFitter`. This allows to use the same analysis tools for both fitters. Currently, the states of the individual components are not returned by the fitter.
+To simplify integration, the GSF returns a {class}`Acts::KalmanFitterResult` object, the same as the {class}`Acts::KalmanFitter`. This allows to use the same analysis tools for both fitters.
```
andiwand, line 96:
```suggestion
To simplify integration, the GSF returns an {class}`Acts::KalmanFitterResult` object, the same as the {class}`Acts::KalmanFitter`. This allows to use the same analysis tools for both fitters.
```

andiwand (approved):

PR #1929, opened 2023-03-09
refactor: avoid unnecessary calculations in triplet compatibility 

LuisFelipeCoelho:
This PR moves the calculation of `iSinTheta2 * options.sigmapT2perRadius` to outside the loop over the top triplet component to avoid recalculating this variable for every iteration.
It also avoids the calculation of the seed pT if not necessary.
This is not supposed to change the physics performance.


Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -578,13 +581,6 @@ void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
       // positive/negative in phi
       state.curvatures.push_back(B / std::sqrt(S2));
       state.impactParameters.push_back(Im);
-
-      // evaluate eta and pT of the seed
-      float cotThetaAvg = std::sqrt(cotThetaAvg2);
-      float theta = std::atan(1. / cotThetaAvg);
-      float eta = -std::log(std::tan(0.5 * theta));
-      state.etaVec.push_back(eta);
-      state.ptVec.push_back(pT);
```
CarloVarni, line 587: what were these values used for?

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -538,19 +539,21 @@ void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
...
+        // if pT > maxPtScattering, calculate allowed scattering angle using
+        // maxPtScattering instead of pt.
+        float pT = options.pTPerHelixRadius * std::sqrt(S2 / B2) / 2.;
+        if (pT > m_config.maxPtScattering) {
+          float pTscatterSigma =
+              (m_config.highland / m_config.maxPtScattering) *
+              m_config.sigmaScattering;
+          p2scatterSigma = pTscatterSigma * pTscatterSigma * iSinTheta2;
+        }
+      }
```
CarloVarni, line 555: it looks like the computation is actually different? In particular `sigmaSquaredSPtDependent` was not computed before and  `p2scatterSigma` at line 544 has an extra `iSinTheta2` then before?

If so, can you clarify (in the PR description is ok) what has changed?

CarloVarni (commented):
My understanding is that these changes will only affect the strip detector (if we think only about ATLAS). Are these changes actually run by the CI?

Core/include/Acts/Seeding/SeedFinder.ipp
LuisFelipeCoelho: I believe the calculation is the same:
previously we were re-evaluating pT2scatterSigma within the if statement and then multipling it by iSinTheta2.
Now we also have only one iSinTheta2 factor because sigmaSquaredSPtDependent is not used when re-evaluating pT2scatterSigma within the if statement

so in both cases pT2scatterSigma = [(highland / maxPtScattering) * sigmaScattering ] ^2 * iSinTheta2

for ITk we don't get inside the if statement so this avoids some calculation

Core/include/Acts/Seeding/SeedFinder.ipp
LuisFelipeCoelho: These variables are not used anywhere. They were intended to be included in a performance file, but this should be done in the Performance Writer instead

CarloVarni (approved):

PR #1931, opened 2023-03-09
refactor: navigator interface

andiwand:
I started pulling the navigator into the actor/aborter calls so we can then call functions on it similar to the stepper


andiwand:
I started pulling the navigator into the actor/aborter calls so we can then call functions on it similar to the stepper

I also aligned the parameter order of actor and aborter and improved some comments

cc @paulgessinger @benjaminhuth @asalzburger

benjaminhuth (approved):
looks good!

PR #1932, opened 2023-03-09
feat: experimental navigator for experimental geometry

andiwand:
WIP


Core/include/Acts/Navigation/NavigationState.hpp
```diff
@@ -44,7 +44,7 @@ struct NavigationState {
     const Portal* portal = nullptr;
     /// The boundary check used for the candidate, boundary checks
     /// can differ for sensitive surfaces and portals
-    BoundaryCheck bCheck = true;
+    BoundaryCheck boundaryCheck = true;
```
asalzburger, line 47: I like, makes it explicit.

Core/include/Acts/Navigation/NavigationState.hpp
```diff
@@ -89,52 +92,5 @@ struct NavigationState {
...
-};
-
-/// Fillers and attachers for surfaces to act on the navigation state
-struct SurfacesFiller {
-  /// Helper struct that allows to fill surfaces into the candidate vector it
-  /// allows to use common navigation structs for volume, portal, surfaces
-  ///
-  /// @param nState the navigation state
-  /// @param surfaces the surfaces that are filled in
-  inline static void fill(NavigationState& nState,
```
asalzburger, line 113: I see, you moved them into a new file, I like.

Core/include/Acts/Navigation/NextNavigator.hpp
```diff
@@ -0,0 +1,357 @@
...
+#include <string>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/container/small_vector.hpp>
+
+namespace Acts {
+
+namespace Experimental {
+
+class NextNavigator {
```
asalzburger, line 36: Let's keep the name for the moment, in the future, I would probably like to move that back to Navigator.

Core/include/Acts/Navigation/NextNavigator.hpp
```diff
@@ -0,0 +1,357 @@
...
+ public:
+  struct Config {
+    /// Detector for this Navigation
+    const Detector* detector = nullptr;
+
+    /// Configuration for this Navigator
```
asalzburger, line 42: A question we should ask ourselves, is the a per-call option or a global config, we could move the `resolve*` into per-call probably.

Core/include/Acts/Navigation/NextNavigator.hpp
```diff
@@ -0,0 +1,357 @@
...
+    /// Navigation state - external state: the current surface
+    const Surface* currentSurface = nullptr;
+    /// Navigation state - external state: the target surface
+    const Surface* targetSurface = nullptr;
+    /// Indicator if the target is reached
+    bool targetReached = false;
+    /// Navigation state : a break has been detected
+    bool navigationBreak = false;
+
+    void reset(const GeometryContext& /*geoContext*/, const Vector3& /*pos*/,
```
asalzburger, line 71: In the KF, do we have use for a `flipDirection()` ?

Core/include/Acts/Navigation/NextNavigator.hpp
```diff
@@ -0,0 +1,357 @@
...
+
+    auto& nState = state.navigation;
+    fillNavigationState(state, stepper, nState);
+
+    if (inactive()) {
+      ACTS_VERBOSE("navigator inactive");
+      return;
+    }
+
+    if (nState.currentDetector == nullptr) {
```
asalzburger, line 110: This hands over the currentDetector from `config` to `state`, I suppose.

Core/include/Acts/Navigation/NextNavigator.hpp
```diff
@@ -0,0 +1,357 @@
...
+    bool boundaryCheck = nState.surfaceCandidate->boundaryCheck;
+
+    if (nState.surfaceCandidate->surface != nullptr) {
+      nextSurface = nState.surfaceCandidate->surface;
+    } else if (nState.surfaceCandidate->portal != nullptr) {
+      nextPortal = nState.surfaceCandidate->portal;
+      nextSurface = &nextPortal->surface();
+      isPortal = true;
+    } else {
+      ACTS_ERROR("panic: not a surface not a portal - what is it?");
```
asalzburger, line 132: A discussion we should have: how do we want to do error handling in the propagation ...

Core/include/Acts/Navigation/NextNavigator.hpp
```diff
@@ -0,0 +1,357 @@
...
+      return;
+    }
+
+    // TODO not sure about the boundary check
```
asalzburger, line 136: I think here wen need to understand:
- portals always have 'BoundaryCheck = true`, passive may at maximum have a tolerance, sensitives can be user defined, we need to sort of transport this information.

Core/include/Acts/Navigation/NextNavigator.hpp
```diff
@@ -0,0 +1,357 @@
...
+    auto surfaceStatus = stepper.updateSurfaceStatus(
+        state.stepping, *nextSurface, boundaryCheck, logger());
+
+    // Check if we are at a surface
+    if (surfaceStatus == Intersection3D::Status::onSurface) {
+      ACTS_VERBOSE("landed on surface");
```
asalzburger, line 142: Would be good to keep the `volinfo(), posInfo()` with 

```
ACTS_VERBOSE(volInfo() << posInfo() << (isPortal ? "Portal reached" : "Surface reached"));
```

Core/include/Acts/Navigation/NextNavigator.hpp
```diff
@@ -0,0 +1,357 @@
...
+      bool boundaryCheck = c.boundaryCheck;
+      auto surfaceStatus = stepper.updateSurfaceStatus(state.stepping, surface,
+                                                       boundaryCheck, logger());
+      if (surfaceStatus == Intersection3D::Status::reachable) {
+        ACTS_VERBOSE("Surface reachable, step size updated to "
+                     << stepper.outputStepSize(state.stepping));
+        break;
+      }
+    }
+
```
asalzburger, line 224: Add a comment saying that in the target, we invalidate both current surface and portal, because we do not want to fall into loops.

Core/include/Acts/Navigation/NextNavigator.hpp
```diff
@@ -0,0 +1,357 @@
...
+    nState.charge = stepper.charge(state.stepping);
+    auto fieldResult = stepper.getField(state.stepping, nState.position);
+    if (!fieldResult.ok()) {
+      ACTS_ERROR("could not read from the magnetic field");
+    }
+    nState.magneticField = *fieldResult;
+  }
+};
+
+}  // namespace Experimental
```
asalzburger, line 355: I am pretty happy with this first draft, pretty much along the lines what I thought it should do.

Core/src/Detector/DetectorVolume.cpp
```diff
@@ -114,23 +114,9 @@ bool Acts::Experimental::DetectorVolume::inside(const GeometryContext& gctx,
...
-  //
-  // This can be avoided by a precise volume finder
-  if (not m_volumes.external.empty()) {
-    for (const auto v : volumes()) {
-      if (v->inside(gctx, nState.position)) {
-        v->updateNavigationState(gctx, nState);
-        return;
-      }
-    }
-  }
```
asalzburger, line 128: Great, this was a leftover.

andiwand (commented):
apparently I forgot to press the submit button again :smile: 

here are my comments from yesterday

andiwand:
so I think two things are missing now: the start and the target surface. I think I will add it to stay compatible with the existing navigator and we will need it for the fitters anyway.

the other thing would be to make the unit test actually assert on something

IMO we could merge as is because it is just experimental and it will make prototyping for muons easier

Core/include/Acts/Navigation/NextNavigator.hpp
asalzburger: Ok, let's leave it then.

Core/include/Acts/Navigation/NextNavigator.hpp
asalzburger: Makes sense.

Core/include/Acts/Navigation/NextNavigator.hpp
asalzburger: Indeed, for the KalmanFilter, if you provide a measurement, you need to do 'BoundaryCheck = false', because it's the updater that should tell you if the hit is taken or not, not the surface bounds. We should keep this in mind and refine.

asalzburger (approved):
Let's move this in & we can refine if needed in a follow-up.

Core/include/Acts/Navigation/NextNavigator.hpp
asalzburger: ok.

asalzburger:
@andiwand - ok, let's move this in, such that we can establish first testbeds, if we need refining/updating we will anyway learn this when using the new navigation interface.

asalzburger:
Great job, by the way!

PR #1936, opened 2023-03-10
refactor: Change Variables Name in order not to start with _

CarloVarni:
This generates a huge amount of compilation warnings during Acts Tag bumps in Athena (related only to the nomenclature) that may hide real warnings we should pay attention to.

Absolutely no change in the code behaviour


paulgessinger (approved):
While I don't think we really should have to adhere to the ATLAS style guide here, I agree it's the path of least resistance to not have a lot of warnings.

PR #1938, opened 2023-03-11
refactor: optimize examples ambiguity resolution

andiwand:
my first implementation was quite wasteful in terms of CPU resources because I just wanted to get it done to test downstream performance for the vertexing. I optimized the code in a way that it pre computes the shared hits and modifies these numbers when removing a track.

performance difference for the current `full_chain_odd.py` without Pythia8 and without Geant4

new

```
identifier	time_total_s	time_perevent_s
Algorithm:AmbiguityResolutionAlgorithm	0.5349	0.005349
```

vs old

```
identifier	time_total_s	time_perevent_s
Algorithm:AmbiguityResolutionAlgorithm	12.52	0.1252
```


andiwand:
@Corentin-Allaire I optimized my code now and added a chi2 comparison

Corentin-Allaire:
> @Corentin-Allaire I optimized my code now and added a chi2 comparison

@andiwand Wow ! That is impressive. Is it still the same performances ? If you want to can do some comparison with the ML solution once this gets merged

Corentin-Allaire:
I can review this on Monday if you want

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
```diff
@@ -38,148 +38,145 @@ ActsExamples::AmbiguityResolutionAlgorithm::AmbiguityResolutionAlgorithm(
...
-    const Acts::ConstVectorMultiTrajectory& multiTrajectory,
-    const std::size_t tip) {
-  std::size_t result = 0;
+void removeTrack(State& state, std::size_t iTrack) {
+  for (auto iMeasurement : state.measurementsPerTrack[iTrack]) {
+    state.tracksPerMeasurement[iMeasurement].erase(iTrack);
 
-  multiTrajectory.visitBackwards(tip, [&](const auto&) { ++result; });
+    if (state.tracksPerMeasurement[iMeasurement].size() == 1) {
+      auto jTrack = *std::begin(state.tracksPerMeasurement[iMeasurement]);
```
paulgessinger, line 126:
```suggestion
      auto jTrack = state.tracksPerMeasurement[iMeasurement].front();
```
?

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
```diff
@@ -38,148 +38,145 @@ ActsExamples::AmbiguityResolutionAlgorithm::AmbiguityResolutionAlgorithm(
...
-    std::vector<float> relativeSharedHits(trackIndices.size(), 0);
-    for (std::size_t i = 0; i < trackIndices.size(); ++i) {
-      const auto indexTrack = trackIndices[i];
-      relativeSharedHits[i] = 1.0f * sharedHits[i] / hitCount[indexTrack];
+    return state.trackChi2[a] < state.trackChi2[b];
+  };
+
+  for (std::size_t i = 0; i < m_cfg.maximumIterations; ++i) {
+    auto maximumSharedMeasurements = *std::max_element(
+        std::begin(state.selectedTracks), std::end(state.selectedTracks),
```
paulgessinger, line 161:
```suggestion
        state.selectedTracks.begin(), state.selectedTracks.end(),
```
but not important.

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
```diff
@@ -38,148 +38,145 @@ ActsExamples::AmbiguityResolutionAlgorithm::AmbiguityResolutionAlgorithm(
...
-        computeSharedHits(sourceLinks, trajectories, trackIndices, trackTips);
-
-    std::vector<float> relativeSharedHits(trackIndices.size(), 0);
-    for (std::size_t i = 0; i < trackIndices.size(); ++i) {
-      const auto indexTrack = trackIndices[i];
-      relativeSharedHits[i] = 1.0f * sharedHits[i] / hitCount[indexTrack];
-    }
+    auto badTrack =
+        *std::max_element(std::begin(state.selectedTracks),
+                          std::end(state.selectedTracks), badTrackComperator);
```
paulgessinger, line 173:
```suggestion
        *std::max_element(state.selectedTracks.begin(),
                          state.selectedTracks.end(), badTrackComperator);
```

paulgessinger (commented):
Looks pretty good.
I see a slight *reduction* in holes in the physmon output, anything else?

Could you maybe elaborate on the changes in the PR description? I think I got it, but would be nice for you to explain a little.

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
andiwand: this does not work on the set unfortunately

paulgessinger: Ok then nevermind.

andiwand:
> I see a slight _reduction_ in holes in the physmon output, anything else?

I am not sure where that comes from. I thought it might be due to some reordering which causes different tracks to be removed. I also added a different ordering based on the chi2 so not having 1:1 would be expected by now. but I definitely saw some difference in the output before that already

andiwand:
@Corentin-Allaire how hard would it be to move your performance evaluation into an root performance writer inside of acts? I thought that might be nice to monitor the results. I can also take this over if you could point me to your eval script

Corentin-Allaire:
> @Corentin-Allaire how hard would it be to move your performance evaluation into an root performance writer inside of acts? I thought that might be nice to monitor the results. I can also take this over if you could point me to your eval script

The issue for that is that I am pretty sure there is no root writer for the multitrajectory, so you have to go through the csv to get them... Porting the script itself (here : https://github.com/acts-project/acts/tree/main/Examples/Scripts/Python/MLAmbiguityResolution) is probably not too hard but getting the right input is probably more complicated

Corentin-Allaire:
@andiwand if you wan to have a look at how to implemented the eval in root I will let you. Right now my priority is to have the ML Solver in the CI. (Since it is not the case I notice last week that some of the suggestion on its MR actually broke the solver)

paulgessinger:
I think we should try to move most of these performance tools to ROOT outputs. I hope I'll have some time soon to start to centralize track-level outputs to TrackContainer.

andiwand:
@stephenswat tagging you after discussion

Corentin-Allaire:
> I also thought about having an intermediate output of the performance eval so either an separate algorithm which just measures performance and another one which writes it. because if you want to do parameter optimization you could do that without writing+reading root files which might be a big overhead

The definition of good/duplicate/fake is stillin the csv writer in the end so you will need to port it no matter what you decide

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
```diff
@@ -20,14 +20,14 @@
 #include <numeric>
 #include <stdexcept>
 
+#include <boost/container/flat_map.hpp>
+#include <boost/container/flat_set.hpp>
```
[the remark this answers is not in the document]
paulgessinger: Containers like this are fine. We specifically avoid compiled Boost components, and very heavy ones like MPL etc.

Corentin-Allaire (approved):
I think this can go in now. Once it is merge I will redo the performance comparison with the ML solver

Corentin-Allaire (approved):

PR #1939, opened 2023-03-13
refactor: Switch all of the Examples to read/write handles

paulgessinger:
Follow up to #1908. Switches over the rest of the Examples components. This also makes `WhiteBoard::{get,add}` private and adds the data handles as friends to restrict unregistered white board access.


paulgessinger:
CI is finally happy. @andiwand or @benjaminhuth maybe you could have a look?

andiwand (approved):
nice! looks good to me :+1:

PR #1940, opened 2023-03-13
feat: Refitting algorithm

benjaminhuth:
So far, real refitting of tracks was not possible in the examples framework. The direct-fitter option of the `ActsExamples::TrackFittingAlgorithm` was broken due to the fact, that only measurement surfaces were added to the surface sequence, and thus e.g. the traversed material could differ.

This adds `ActsExamples::RefittingAlgorithm`, that uses the new `Acts::TrackContainer` infrastructure to enable a very simple interface: Just a input container and a output container.

Furthermore, the `TrackFittingAlgorithm` looses the `directFit` option. To make this work, also the `TrackFitterFunction` had to be refactored and was pulled out of the `TrackFittingAlgorithm`.


Examples/Algorithms/TrackFitting/include/ActsExamples/TrackFitting/TrackFitterFunction.hpp
```diff
@@ -0,0 +1,98 @@
...
+#include "ActsExamples/TrackFitting/AlreadyCalibratedCalibrator.hpp"
+
+namespace ActsExamples {
+
+/// Fit function that takes the above parameters and runs a fit
+/// @note This is separated into a virtual interface to keep compilation units
+/// small.
+class TrackFitterFunction {
+ public:
+  using TrackContainer =
```
paulgessinger, line 34: Can this use the central typedef?

paulgessinger (commented):
Didn't look through all of it yet. Looks good as far as I can tell, a few remarks here.

Also, can we hold this until after #1939, and move to the DataHandles immediately?

andiwand (approved):
looks good to me

PR #1941, opened 2023-03-14
fix: Sequencer use-after-move in sequencer data flow check

paulgessinger:
If Sequencer data flow checks are turned on, there's a use-after-move bug that this PR fixes.


andiwand (approved):

benjaminhuth:
clang-tidy now wants a reference at Examples/Framework/src/Framework/Sequencer.cpp:112 @paulgessinger

benjaminhuth (approved):

PR #1945, opened 2023-03-14
ci: Add workflow to trigger Athena build

paulgessinger:
Only runs on the main branch, uses a secret trigger token to build it


.github/workflows/trigger_athena.yml
```diff
@@ -0,0 +1,15 @@
+name: Trigger Athena build
```
CarloVarni, line 1: I'd suggest not to use `trigger` in the name of the test and file. That may suggest this is related to the Athena trigger.

CarloVarni (commented):
To understand... this will add a test to the CI that checks the Athena build is not broken to the PR?

.github/workflows/trigger_athena.yml
paulgessinger: I see. It's literally called a "pipeline trigger" in GitLab 😅

paulgessinger:
> To understand... this will add a test to the CI that checks the Athena build is not broken to the PR?

Unfortunately no. This only makes sure that one CI job on GitLab runs for every commit to the main branch here.

.github/workflows/trigger_athena.yml
paulgessinger: Can we keep this? I think it's standard terminology.

CarloVarni (approved):

PR #1947, opened 2023-03-15
fix: itk_seeding script runs again

CarloVarni:
As a result of recent PRs and a Typo, the `itk_seeding.py` script was not running. 

Here are the changes:
- using an enum to specify the type of the input space points instead of relying on strings (typo-prones)
- scheduling of `SeedsToPrototracks` since now the seeding algorithm does not produce anymore prototracks

/cc @LuisFelipeCoelho


LuisFelipeCoelho (approved):
Thank you for changing that. It looks good to me.
Maybe we need a CI check for this file as well so it won't get outdated constantly

PR #1950, opened 2023-03-15
refactor: remove dead actor code

andiwand:
"observer" functions are not called if there is a proper actor function


benjaminhuth (approved):
It's always nice to have less code!

PR #1952, opened 2023-03-15
chore: rename unused params

andiwand:
- rename `/*unused*/` params into something more understandable
- improve doc


Examples/Algorithms/Digitization/include/ActsExamples/Digitization/DigitizationConfig.hpp
```diff
@@ -55,8 +55,7 @@ struct GeometricConfig {
   bool digital = false;
 
   /// Charge generation (configurable via the chargeSmearer)
-  Acts::ActsScalar charge(Acts::ActsScalar path, Acts::ActsScalar /*unused*/,
-                          RandomEngine &rng) const {
+  Acts::ActsScalar charge(Acts::ActsScalar path, RandomEngine &rng) const {
```
paulgessinger, line 58: The argument goes away completely and we never need it?

Examples/Algorithms/Digitization/src/DigitizationAlgorithm.cpp
```diff
@@ -304,7 +304,7 @@ ActsExamples::DigitizationAlgorithm::localParameters(
   for (const auto& ch : channels) {
     auto bin = ch.bin;
     Acts::ActsScalar charge =
-        geoCfg.digital ? 1. : geoCfg.charge(ch.activation, ch.activation, rng);
+        geoCfg.digital ? 1. : geoCfg.charge(ch.activation, rng);
```
paulgessinger, line 307: Oh so this was just duplicated? Or is this maybe a mistake at the call-site here?

paulgessinger (approved):

Examples/Algorithms/Digitization/src/DigitizationAlgorithm.cpp
andiwand: I was not sure about that. but as far as I could see this params was never used so I dropped it

andiwand:
CI is green now. could you take another look @paulgessinger ?

benjaminhuth (approved):
looks good to me

Examples/Algorithms/Digitization/src/DigitizationAlgorithm.cpp
benjaminhuth: I recently discussed this with @asalzburger, but unfortunately don't remember the meaning of this right now. However, since it is not used at the moment, I think we can remove it and eventually introduce it again if needed.

PR #1954, opened 2023-03-16
feat: add (phi, z) cuts to middle SP in orthogonal seeding

LuisFelipeCoelho:
in the orthogonal seeding we currently apply these cuts only to the bottom and top SPs


stephenswat (approved):
Looks excellent to me! :smile:

PR #1951, opened 2023-03-15
refactor: state observer functions for navigator

andiwand:
actor code used to reach through the propagator state directly into the navigator state which makes it hard to add alternative navigators

this PR adds getter functions to the navigator which access the state of the navigator similar to the stepper

blocked by
- https://github.com/acts-project/acts/pull/1979

PR #1955, opened 2023-03-16
ci: Switch main CI job to Ubuntu 22.04 

paulgessinger:
I'm adding the old Ubuntu 20.04 job as an "extra" job, basically switching it out for the prior 22.04 one.


paulgessinger:
Physmon and python level tests still fail, likely due to various version changes.

paulgessinger:
IVF seeded:
diffz
image

IVF truth estimated:
diffz
image
recoz
image

IVF orthogonal:
recoz
image

GSF truth tracking:
resmean t vs eta
image


---

I think all of these changes are tolerable.

.github/workflows/builds.yml
```diff
@@ -279,7 +279,7 @@ jobs:
           && git config --global safe.directory "$GITHUB_WORKSPACE"
           && pip3 install histcmp==0.5.2
           && pip3 install -r Examples/Scripts/requirements.txt
-          && /usr/local/bin/download_geant4_data.sh
+          && /usr/local/bin/geant4-config --install-datasets
```
paulgessinger, line 282: Turns out G4 actually ships with a script that does exactly what we need here. 😅

benjaminhuth:
I don't understand that GSF change. Seems also to happen in #1951

benjaminhuth:
> I don't understand that GSF change. Seems also to happen in #1951

Okay, I see the same locally on `main`... not sure what could be the reason, I'll have a look at it.

benjaminhuth (approved):
looks good to me!

PR #1957, opened 2023-03-17
ci: Add builds using C++20

paulgessinger:
The main workflow continues to run on C++17, but this adds variants to the `extra` build matrix that builds in C++20, too.

So we should now cover:
- Ubuntu 20.04 + C++17
- Ubuntu 22.04 + C++17 and C++20
- Ubuntu 22.04 clang + C++17 and C++20


andiwand (approved):
looks good to me :+1:

paulgessinger:
Nothing concrete yet, but I thought it doesn't hurt to check that we're ready in principle.

PR #1958, opened 2023-03-17
refactor!: Store unsigned 64bit integer for projector in MTJ backend

paulgessinger:
Is this breaking?


tboldagh (approved):
Definitely ok from perspective of other ATLAS backend.
Side comment. I realised that probably calling these backend classes: *Backend would relay help in understanding the code.

PR #1960, opened 2023-03-17
refactor: restructure onnx examples build

andiwand:
align ONNX examples build with existing infrastructure


andiwand:
@Corentin-Allaire this is how I imagined the build for the ML ambi solver. I now face the same build error you did in https://github.com/acts-project/acts/pull/1959 tho

Examples/Python/src/MLTrackFindingStub.cpp
```diff
@@ -0,0 +1,18 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#include "Acts/Plugins/Python/Utilities.hpp"
+
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+namespace Acts::Python {
+void addMLTrackFinding(Context& /*unused*/) {
```
Corentin-Allaire, line 15: Since this does nothing maybe it should return an error if called ?

Corentin-Allaire (commented):
Looks great. I have one final comment but except from that this good to go.

Examples/Python/src/MLTrackFindingStub.cpp
andiwand: this will actually still be called but only if `ML` is not present. in this case nothing will be added to the python bindings

andiwand:
last fix hopefully :crossed_fingers:

Examples/Python/src/MLTrackFindingStub.cpp
Corentin-Allaire: That was what I understood. But if you are calling `addMLTrackFinding` without having onnx installed you are probably making a mistake and there should probably be a message warning you.

andiwand: currently it is actually intended that it is called and does nothing because here https://github.com/acts-project/acts/blob/main/Examples/Python/src/ModuleEntry.cpp#L270 we use that to our advantage. we call the add function and either the module is present and will go through the binding stuff or it is not present and will not do anything

Corentin-Allaire: I didn't know that. Make sense !

PR #1961, opened 2023-03-17
fix: bug in ITk config

LuisFelipeCoelho:
Recently I changed deltaZMax in ITk configuration, but I should have changed only for PPP and not for SSS as well.
@CarloVarni I believe this is the reason for the higher number of SSS seeds that we were seeing. Sorry for that

### DIFF
--- a/Examples/Python/python/acts/examples/itk.py
+++ b/Examples/Python/python/acts/examples/itk.py
@@ -308,7 +308,6 @@ def itkSeedingAlgConfig(inputSpacePointsType: InputSpacePointsType):
     bFieldInZ = 2 * u.T
     deltaRMin = 20 * u.mm
     maxPtScattering = float("inf") * u.GeV
-    deltaZMax = float("inf") * u.mm
     zBinEdges = [
         -3000.0,
         -2500.0,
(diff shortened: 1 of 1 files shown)


CarloVarni:
Good this bug was found. However looking at what we have in Athena it looks like the two configurations (Acts standalone and Athena) diverged: https://gitlab.cern.ch/atlas/athena/-/blob/master/Tracking/Acts/ActsTrkTools/ActsTrkSeedingTool/python/ActsTrkSeedingToolConfig.py

You can see here for instance that the deltaZmax is still 600 for pixel and 900 for strips. 

We should find the time to compare the two configurations and sync them properly

CarloVarni:
Good this bug was found. However looking at what we have in Athena it looks like the two configurations (Acts standalone and Athena) diverged: https://gitlab.cern.ch/atlas/athena/-/blob/master/Tracking/Acts/ActsTrkTools/ActsTrkSeedingTool/python/ActsTrkSeedingToolConfig.py

You can see here for instance that the deltaZmax is still 600 for pixel and 900 for strips. 

We should find the time to compare the two configurations and sync them properly

PR #1962, opened 2023-03-18
refactor: simplify volume drawing and lower case variables

andiwand:


Core/src/Visualization/GeometryView3D.cpp
```diff
@@ -191,42 +192,23 @@ void Acts::GeometryView3D::drawPortal(IVisualization3D& helper,
 
 void Acts::GeometryView3D::drawDetectorVolume(
     IVisualization3D& helper, const Experimental::DetectorVolume& volume,
-    const GeometryContext& gctx, bool drawSurfaces, const Transform3& transform,
```
benjaminhuth, line 194: So this boolean switch just has no use case and is therefore removed?

andiwand: yeah exactly. I think the surfaces should always be a part of the volume. whenever we want to have that optional we can add this back.

andiwand:
@dimitra97 I adopted the code a little bit and removed the flag for now to simplify things

benjaminhuth (approved):
looks good to me

dimitra97:
> @dimitra97 I adopted the code a little bit and removed the flag for now to simplify things

Looks good! More simplified indeed.

dimitra97 (approved):

PR #1964, opened 2023-03-19
feat: Add logger to multi stepper

benjaminhuth:
Adds a logger to the `Acts::MultiEigenStepperLoop` in order to see again the log messages from the `step(...)` function.


Core/include/Acts/Propagator/MultiEigenStepperLoop.hpp
paulgessinger, line 1: Could also be done via `shared_ptr`. Not sure if copying that is cheaper than calling `clone()`. I guess probably?

PR #1965, opened 2023-03-19
feat: Sequencer prints every event start and finish in debug mode

benjaminhuth:
I found this useful for finding events that crash when running a large amounts of events in parallel, as otherwise is is difficult to find out which event crashed exactely.


paulgessinger (approved):
lgtm

PR #1967, opened 2023-03-20
feat: Adding ITk scripts for material mapping and validation 

noemina:
A few scripts are added to allow material mapping and validation for ITk in the ACTS standalone configuration.

@timadye @Corentin-Allaire @paulgessinger


Corentin-Allaire (approved):
Looks good to me !

Corentin-Allaire (approved):

PR #1968, opened 2023-03-20
fix: Skipping un-physical steps in material mapping (#1882)

### LINKED ISSUE #1882: Bug: last material interaction missing in validation output?
While mapping the material on the ITk tracking geometry (using the standalone workflow), I have seen that the material interaction for the outer boundary seems not to be collected in the root file. 
Things I have already tested:
- I checked that the mapping procedure hit the boundary layers I am selecting in the json file (I see the material indeed appearing in the mapping file)
- I also run the navigation in verbose mode to check that the same boundaries I map the material onto are hit 

I start believing that the last material interaction is simply not saved in the output file of the material validation job?

@Corentin-Allaire, could you help me with this?

noemina:
This PR fixes #1882.

The reason why the outer boundary was not seen is due to the accumulation of invalid material on it. Skipping un-physical steps (step length = 0) restores a sane behaviour of the mapping procedure.

I have also added a check on existence of the `event_id` branch in the `TChain`. If the branch doesn't exist, the batch size is evaluated to be 1.

@Corentin-Allaire @goetzgaycken


Corentin-Allaire (approved):
Looks great, thanks a lot for solving this issue !

noemina:
@Corentin-Allaire , do you understand why the tests fail?

Corentin-Allaire:
> @Corentin-Allaire , do you understand why the tests fail?

Yes now that you have fixed the issue in the material mapping the map created in the test are different from the on in the test.
You need to update the `/Examples/Python/tests/root_file_hashes.txt` with the new hashes that are in the CI loh

noemina:
> > @Corentin-Allaire , do you understand why the tests fail?
> 
> Yes now that you have fixed the issue in the material mapping the map created in the test are different from the on in the reference. You need to update the `/Examples/Python/tests/root_file_hashes.txt` with the new hashes that are in the CI log.
> 

Yes, Thanks. My question was if we expect the tests to fail. This PR skips empty steps, so we did have empty steps in the input file used in the tests, which were messing up the results there as well?

Corentin-Allaire:
> > > @Corentin-Allaire , do you understand why the tests fail?
> > 
> > 
> > Yes now that you have fixed the issue in the material mapping the map created in the test are different from the on in the reference. You need to update the `/Examples/Python/tests/root_file_hashes.txt` with the new hashes that are in the CI log.
> 
> Yes, Thanks. My question was if we expect the tests to fail. This PR skips empty steps, so we did have empty steps in the input file used in the tests, which were messing up the results there as well?

If I understood you MR correctly those empty step appear a the end of he detector so we probably have some in all of our files.

Corentin-Allaire (approved):

PR #1969, opened 2023-03-20
fix: precision of radLengthPerSeed in ITk configuration 

LuisFelipeCoelho:
Apparently radLengthPerSeed needs to be more precise in the ITk configuration because it can lead to differences for high pile-up. @CarloVarni @noemina


Examples/Python/tests/root_file_hashes.txt
```diff
@@ -15,7 +15,7 @@ test_seeding_orthogonal__performance_seeding.root: 60fbedcf5cb2b37cd8e5262519405
 test_seeding_orthogonal__particles.root: 74e08ee12bdaf9f7d273369c71c742df345475c8b73bb7ce223619d0dd1f87ee
 test_seeding_orthogonal__particles_final.root: a011dda7edbdc8b8a0e275f5940d46b8e0fab99e10b7df152be380bddfc25192
 test_seeding_orthogonal__particles_initial.root: 74e08ee12bdaf9f7d273369c71c742df345475c8b73bb7ce223619d0dd1f87ee
-test_itk_seeding__estimatedparams.root: 6840539a46e3056aa4916dd4b0e06e0d426b8cff52d06b23a80d921e6b39bdaf
+test_itk_seeding__estimatedparams.root: 0625fb5c34cbf66830051ccd2d504e42cfbb336fe38ad06002fb4a0dd187e0fc
```
CarloVarni, line 18: you change the hash but you don't change the root file?

LuisFelipeCoelho: I guess the file is the same but the hash is different?

CarloVarni: no idea. it sounds weird... maybe @paulgessinger knows

CarloVarni: is the CI failing without this hash update @LuisFelipeCoelho ?

LuisFelipeCoelho: yes it only complains about the hash

paulgessinger: The physmon does histogram comparisons. I suppose it accepts the change, even though the file content (slightly?) changes. 

Have you looked at the physmon histograms?

paulgessinger: Ah I don't think that root file is even covered by the physmon checks. So the hash is the only thing that changes, and we don't have a reference file commited to the repository for this.

CarloVarni (approved):

CarloVarni:
@LuisFelipeCoelho the CI should be ok now. But we have to make a new commit in order to trigger the CI again

PR #1970, opened 2023-03-20
fix: GSF handles another case of empty components correctly

benjaminhuth:
If the weight cutoff is very high, it could happen that components are empty after applying the energy loss, and an assertion triggers. This fixes this and prints out a warning.

### DIFF
--- a/Core/include/Acts/TrackFitting/detail/GsfActor.hpp
+++ b/Core/include/Acts/TrackFitting/detail/GsfActor.hpp
@@ -300,7 +300,7 @@ struct GsfActor {
         ACTS_WARNING(
             "No components left after applying energy loss. "
             "Is the weight cutoff too high?");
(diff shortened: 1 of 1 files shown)


Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -296,6 +296,14 @@ struct GsfActor {
       std::vector<ComponentCache> componentCache;
       convoluteComponents(state, stepper, tmpStates, componentCache);
 
+      if (componentCache.empty()) {
+        ACTS_WARNING(
+            "No components left after applying energy loss. "
+            "Is the weight cutoff too high?");
```
AJPfleger, line 302:
```suggestion
            "Is the weight cutoff = {...} too high?");
```
could we easily print out the value of the weight cutoff here?

PR #1972, opened 2023-03-21
feat: remove middle SPs in last layer

LuisFelipeCoelho:
This reduces iterations by removing the middle SPs on the last layer since there would be no outer SPs to complete a seed


Core/include/Acts/Seeding/SeedFinderConfig.hpp
```diff
@@ -119,6 +119,10 @@ struct SeedFinderConfig {
   // which will make seeding very slow!
   float rMin = 33 * Acts::UnitConstants::mm;
 
+  // z of last layers to avoid iterations
+  float zMinLastLayer = -2800 * Acts::UnitConstants::mm;
+  float zMaxLastLayer = 2800 * Acts::UnitConstants::mm;
```
CarloVarni, line 124: Indeed this is very experiment dependent since it requires information on the geometry of the detector. So it is good to have it as a configurable parameter. 

I would though prefer to have clearer names for these parameters. 'zMinLastLayer' and `zMaxLastLayer` are quite puzzling

Examples/Python/python/acts/examples/itk.py
```diff
@@ -297,6 +297,8 @@ def itkSeedingAlgConfig(inputSpacePointsType: InputSpacePointsType):
     # variables that do not change for pixel and strip SPs:
     zMax = 3000 * u.mm
     zMin = -3000 * u.mm
+    zMinLastLayer =  2700 * u.mm
+    zMaxLastLayer =  -2700 * u.mm
```
CarloVarni, line 301: why do we have these values as +/- 2800 in the default configuration, but then we have them +/- 2700 for the ITk case? 

Can't they directly be +/- 2700 in the configs as well?

Core/include/Acts/Seeding/SeedFinderConfig.hpp
```diff
@@ -119,6 +119,10 @@ struct SeedFinderConfig {
   // which will make seeding very slow!
   float rMin = 33 * Acts::UnitConstants::mm;
 
+  // z of last layers to avoid iterations
+  float zMinLastLayer = -2800 * Acts::UnitConstants::mm;
```
CarloVarni, line 123: as a general comment from my side: I find the seeding configuration to already have a worryingly high number of parameters to set (some of which look redundant or are not clear what they do). This PR will add two more to them.
Should we found a way to compress these configurations?

What I mean is: use a `std::pair<float, float>` instead of two `float` for defining a range for instance

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -659,6 +662,8 @@ def addOrthogonalSeeding(
             collisionRegionMax=seedFinderConfigArg.collisionRegion[1],
             zMin=seedFinderConfigArg.z[0],
             zMax=seedFinderConfigArg.z[1],
+						zMinLastLayer=(seedFinderConfigArg.zLastLayer[0] if seedFinderConfigArg.zLastLayer[0] not None else seedFinderConfigArg.z[0]),
```
CarloVarni, line 665: this is a weird indentation

CarloVarni:
I wonder... if we know the maximum z for the middle sp... would it be easier to do the following:

From the global bin position of the grid (input to the createSeedForGroup) you can get the local jobs position of the grid. If the z bin is at the edge, then skip any computation.

Just throwing ideas...

Also, there is a check that the size of neightbour bins are not zero (number of bins, not number of sps)... so I'm not sure if that check already covers this

CarloVarni:
> I wonder... if we know the maximum z for the middle sp... would it be easier to do the following:
> 
> From the global bin position of the grid (input to the createSeedForGroup) you can get the local jobs position of the grid. If the z bin is at the edge, then skip any computation.
> 
> Just throwing ideas...
> 
> Also, there is a check that the size of neightbour bins are not zero (number of bins, not number of sps)... so I'm not sure if that check already covers this

Change of plan ... just do the z check only if the middle sp is at the grid edge. That saves us on the number of checks. 

I realized that the grid bin edges are more generous then I was expected

CarloVarni (approved):

PR #1973, opened 2023-03-21
perf: Reduce memory consumption

CarloVarni:
We have three vectors (of `std::size_t`) that call a reserve of these values. Reduce the number to 100 since it makes no sense we have 1000000000 seed candidates for a single middle SP ... not even for ITk at PU200

/cc @LuisFelipeCoelho


LuisFelipeCoelho (approved):

CarloVarni:
@LuisFelipeCoelho you have to approve again since there have been commit after the last approval

LuisFelipeCoelho (approved):

PR #1979, opened 2023-03-22
refactor: make navigator state copyable

andiwand:
navigator state used to be not copyable because of the usage of iterators. this PR replaces the iterators by indices to enable safe copies

should fix https://github.com/acts-project/acts/pull/1951


andiwand:
this fixes the problem with https://github.com/acts-project/acts/pull/1951/ locally

benjaminhuth (approved):
Looks good to me!

PR #1981, opened 2023-03-22
build: Take CMAKE_CXX_STANDARD into account

paulgessinger:
We're currently setting cxx_std_17 target by target.
This PR changes this to 17 or higher, and it only sets it on ActsCore
and lets CMake propagate this to dependent targets.

Blocked by:
- #2002


cmake/ActsCompilerOptions.cmake
```diff
@@ -10,6 +10,11 @@ set(ACTS_CXX_FLAGS_MINSIZEREL "")
 set(ACTS_CXX_FLAGS_RELEASE "")
 set(ACTS_CXX_FLAGS_RELWITHDEBINFO "")
 
+set(ACTS_CXX_STANDARD_FEATURE cxx_std_17)
+if(${CMAKE_CXX_STANDARD} GREATER 17)
+  set(ACTS_CXX_STANDARD_FEATURE "cxx_std_${CMAKE_CXX_STANDARD}")
```
benjaminhuth, line 15: Actually I would send a `message(ERROR "...")` in case the user provides `CMAKE_CXX_STANDARD=14` or less and not silently go to C++17

paulgessinger: Done.

asalzburger:
This has been approved quite a while ago, adding automatic merge option.

andiwand (approved):

asalzburger:
@paulgessinger after the automerge pipeline went through in the night, we now get a build failure on the examples for this one, do you understand why?

asalzburger:
> I got the same error (ML ambi solver failure) on an unrelated PR, in that case just restarting resolved this.

I was afraid that this is a more general hiccup on all CI jobs now.

paulgessinger:
I don't see a build failure? One of the unit tests seems to infinite loop or hang otherwise, which is why the job gets killed after 6h.

asalzburger:
> I don't see a build failure? One of the unit tests seems to infinite loop or hang otherwise, which is why the job gets killed after 6h.

Sorry, CI failure ... a stuck job - can we trace which PR caused that first?

Tests/UnitTests/Core/EventData/MultiTrajectoryTests.cpp
```diff
@@ -1139,7 +1139,8 @@ BOOST_AUTO_TEST_CASE(MemoryStats) {
 
   for (int t = 0; t < type_axis.size(); t++) {
     for (int c = 0; c < column_axis.size(); c++) {
-      BOOST_CHECK_EQUAL(h.at(c, t), 0);
+      int v = h.at(c, t);
+      BOOST_CHECK_EQUAL(v, 0);
```
paulgessinger, line 1143: Somehow, this fixes the infinite hang.

andiwand (approved):

andiwand:
:rocket:

PR #1982, opened 2023-03-23
feat: Make layer envelope dimensions configurable

plariono:
The idea of this PR is to make the radial envelope parameters of the layer/volume builder configurable. This is important for the geometries with closely located layers: the standard (and hardcoded) dimensions of the envelopes result in volume overlapping.
I also realized that there are two envelopes built around each layer and I'm not sure I understand the idea behind it. 
@asalzburger, @paulgessinger could you comment on that?
I introduced two parameters, `beampipeEnvelopeR` and `layerEnvelopeR` that configure the radial envelope dimensions.


asalzburger:
All good, ready to go in when pipeline succeeds.

Examples/Detectors/TGeoDetector/include/ActsExamples/TGeoDetector/TGeoDetector.hpp
```diff
@@ -54,6 +54,8 @@ struct TGeoDetector {
     double beamPipeRadius{0};
     double beamPipeHalflengthZ{0};
     double beamPipeLayerThickness{0};
+    double beampipeEnvelopeR{1.0};
```
andiwand, line 57:
```suggestion
    double beamPipeEnvelopeR{1.0};
```

consistency

asalzburger:
Seems was not a complete patch :-) ... I will fix it as I caused it by in-situ merging.

plariono:
@asalzburger @andiwand thank you for the corrections and fixes. Now the letter case should be corrected everywhere.

PR #1984, opened 2023-03-23
refactor: explicit init, pre step, post step for navigator interface

andiwand:
my attempt to make the navigator code more readable. list of changes:

- explicit init instead of check and call in `status`
- `status` -> `postStep` because that's what is actually does and why it is called
- `target` -> `preStep` same as above


andiwand:
could one of you have a look at this @asalzburger @paulgessinger - would also okay for me to close if you do not think it is relevant enough

paulgessinger:
Could still be good for clarity, but I see the added complexity. @asalzburger what do you think?

benjaminhuth:
I would actually like this to go in, because everytime I need to do something in the navigator, I need to figure out myself again, what the `status` and `target` calls actually do...

andiwand:
This one actually only renames the functions and reorders the code in the navigator. That's why there is a lot of diff

The other PR https://github.com/acts-project/acts/pull/1987 was a bit more involved but I closed it now

benjaminhuth (approved):
Very nice. Only a minor code readablility suggestion, feel free to ignore it!

benjaminhuth (approved):

PR #1985, opened 2023-03-24
refactor!: split filterCandidates for ITk SSS seeding

LuisFelipeCoelho:
This PR proposes some calculation optimisation for `filterCandidates` by splitting tis function into `filterCandidates` and `filterCandidatesDetailedMeasurements`. This also allows us to remove two parameters in the seeding configuration.

@noemina @CarloVarni


CarloVarni:
If this PR removes config parameters then it is a breaking change. /cc @paulgessinger

Core/include/Acts/Seeding/SeedFinderUtils.ipp
```diff
@@ -188,19 +188,25 @@ inline bool xyzCoordinateCheck(
   const Acts::Vector3& stripCenterDistance =
       spacePointData.getStripCenterDistance(index);
 
+  // prepare variables
+  double xTopStripVector = topHalfStripLength * topStripDirection[0];
+  double yTopStripVector = topHalfStripLength * topStripDirection[1];
+  double zTopStripVector = topHalfStripLength * topStripDirection[2];
+  double xBottomStripVector = bottomHalfStripLength * bottomStripDirection[0];
+  double yBottomStripVector = bottomHalfStripLength * bottomStripDirection[1];
+  double zBottomStripVector = bottomHalfStripLength * bottomStripDirection[2];
```
CarloVarni, line 197: good point! thanks

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -691,6 +566,268 @@ inline void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
   }  // loop on bottoms
 }
 
+template <typename external_spacepoint_t, typename platform_t>
+inline void SeedFinder<external_spacepoint_t, platform_t>::
+    filterCandidatesDetailedMeasurements(
```
CarloVarni, line 571: how much of this is a copy-paste of the `filterCandidates` method?

CarloVarni (commented):
Few comments from my side. 
I like the idea of having separate functions for strips and pixel, since that would make the code cleaner and possibly faster. However, I'm hesitant in having copy-pasted code, which would be quite bug-prone. 

Is there a way to reduce even further the duplication of code?

Core/include/Acts/Seeding/SeedFinderUtils.ipp
CarloVarni: Can you disentagle this change from the rest of this PR? This will improve performance for SSS and is not controversial. I'd suggest opening a PR only with this change so that it can be accepted quickly

Core/include/Acts/Seeding/SeedFinder.ipp
LuisFelipeCoelho: Thank you for the comments! I think there is some duplicated code that can be avoided, I will work on that

paulgessinger:
What's the status of this? @LuisFelipeCoelho  @CarloVarni

LuisFelipeCoelho:
@paulgessinger I plan to completely change this PR and avoid the splitting of SSS and PPP with compile time if statements, same approach used by @CarloVarni in #2013

LuisFelipeCoelho:
I will close this PR, which will be replaced by #2108

PR #1986, opened 2023-03-24
refactor: SSS xyzCoordinateCheck calculations

LuisFelipeCoelho:
This PR reduces duplicated calculations in `xyzCoordinateCheck` function.

### DIFF
--- a/Core/include/Acts/Seeding/SeedFinderUtils.ipp
+++ b/Core/include/Acts/Seeding/SeedFinderUtils.ipp
@@ -188,19 +188,25 @@ inline bool xyzCoordinateCheck(
   const Acts::Vector3& stripCenterDistance =
       spacePointData.getStripCenterDistance(index);
 
+  // prepare variables
+  double xTopStripVector = topHalfStripLength * topStripDirection[0];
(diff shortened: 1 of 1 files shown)


CarloVarni (approved):

LuisFelipeCoelho:
I believe that this could be further optimised. Maybe `topHalfStripLength * topStripDirection` and `bottomHalfStripLength * bottomStripDirection` could be calculated only once per event

CarloVarni:
> I believe that this could be further optimised. Maybe `topHalfStripLength * topStripDirection` and `bottomHalfStripLength * bottomStripDirection` could be calculated only once per event (but for every SP)

Agree! These values can be store in the 'spacePointData' object. In my tests I noticed that these values are calculated anyway for ~95% of the input space points.. so no worries about computing them for all space points. 

Do you wanna do it in another PR or in this one?

CarloVarni:
If they are added to the spacePointData then this Pr will become a breaking one

LuisFelipeCoelho:
I see, if they are calculated for ~95% of the input space points we should do that. I can make a follow up PR to this one

PR #1988, opened 2023-03-24
refactor: remove/rearrange includes in propagator

andiwand:
in hope of reducing compilation time


Core/include/Acts/Geometry/GeometryObjectSorter.hpp
```diff
@@ -8,10 +8,8 @@
 
 #pragma once
 
-// clang-format off
 // Workaround for building on clang+libstdc++. Must always be first
 #include "Acts/Utilities/detail/ReferenceWrapperAnyCompat.hpp"
-// clang-format on
```
paulgessinger, line 14: I put these there so this never gets reordered by clang-format. Are you sure it won't reorder?

andiwand: I can revert this. just noticed that we have a couple of occasions without that format guard

benjaminhuth (approved):

PR #1989, opened 2023-03-24
refactor: move reference wrapper any workaroud to where it is needed

andiwand:
wondered what happens if I do that :crossed_fingers:


andiwand:
> Not sure, but this only broke in very special circumstances so I would probably keep this just to be safe?

do you remember if this broke downstream or directly in acts?

I just noticed that that workaround will bring in `<any>` and `<type_traits>` in basically all compilation units no matter it is used or not. but maybe that is not that much of an overhead?

alternatively I thought it could be included more targeted in placed where that workaround is actually necessary

benjaminhuth:
> I just noticed that that workaround will bring in `<any>` and `<type_traits>` in basically all compilation units no matter it is used or not. but maybe that is not that much of an overhead?

You could put the `#include`s into the `#ifdef` block. However, I think these headers are not too expensive...

benjaminhuth:
> I just noticed that that workaround will bring in `<any>` and `<type_traits>` in basically all compilation units no matter it is used or not. but maybe that is not that much of an overhead?

You could put the `#include`s into the `#ifdef` block. However, I think these headers are not too expensive...

andiwand:
closing for now because the impact might be to small compared to the effort

looks like the problem arises from having `std::reference_wrapper<const *Context>` all over the place which enables generated copy assign. an alternative would be having a pointer to the context.

because of the user extension to the context the current workaround seems to be more universal then only adding it to `ContextType.hpp`

PR #1991, opened 2023-03-25
perf: Space Point Formation requires functions that accepts const reference to source links

CarloVarni:


tboldagh (approved):
This is supper trivial change. Even if I am not the designated approver let me click approve too.

toyamaza:
looks good to me as well. thanks.

PR #1996, opened 2023-03-27
refactor: Ambiguity solvers now use Track containers

Corentin-Allaire:
This PR change both Ambiguity solvers to use the Track container instead of the Trajectory ones.

### DIFF
--- a/Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
+++ b/Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
@@ -43,9 +43,8 @@ namespace {
 struct State {
   std::size_t numberOfTracks{};
 
-  std::vector<std::pair<std::size_t, std::size_t>> trackTips;
+  std::vector<int> trackTips;
   std::vector<float> trackChi2;
-  std::vector<ActsExamples::TrackParameters> trackParameters;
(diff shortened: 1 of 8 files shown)


andiwand (approved):
lgtm! :+1:

Examples/Algorithms/TrackFinding/src/AmbiguityResolutionAlgorithm.cpp
```diff
@@ -28,25 +26,23 @@ ActsExamples::AmbiguityResolutionAlgorithm::AmbiguityResolutionAlgorithm(
...
+  m_outputTracks.initialize(m_cfg.outputTracks);
 }
 
 namespace {
 
 struct State {
   std::size_t numberOfTracks{};
 
-  std::vector<std::pair<std::size_t, std::size_t>> trackTips;
+  std::vector<int> trackTips;
```
paulgessinger, line 44: This is only internal to the algorithm, right?

Corentin-Allaire: Yes, I think so. @andiwand can probably comment more on this

andiwand: yes this is algorithm internal. I use this state to operate on with free functions

as far as I can tell the `IndexType` will end up as `std::uint32_t`. so `int` should be fine

PR #2006, opened 2023-03-31
refactor: LinCircle constructor and no use of fill function

CarloVarni:
Defining a constructor makes the fillLineCircle function redundant


LuisFelipeCoelho:
Do you know if this has any effect on CPU performance?

PR #2007, opened 2023-03-31
refactor: Split Track.hpp into TrackProxy.hpp and TrackContainer.hpp

paulgessinger:


Core/include/Acts/EventData/Track.hpp
```diff
@@ -8,794 +8,6 @@
...
-  }
-};
-
-template <typename T>
-using TrackAccessor = TrackAccessorBase<T, false>;
-template <typename T>
-using ConstTrackAccessor = TrackAccessorBase<T, true>;
-
-}  // namespace Acts
+#include "Acts/EventData/TrackContainer.hpp"
```
paulgessinger, line 13: Keeping this header around for now, but printing a warning.

andiwand: is this to avoid breaking changes? do we have a followup for the next version on this?

andiwand (approved):
looks good to me

paulgessinger:
Yeah I wanted to keep the other header around for another version or so @andiwand.

Core/include/Acts/EventData/Track.hpp
paulgessinger: Yeah I wanted to keep the other header around for another version or so @andiwand.

PR #2008, opened 2023-03-31
feat: Mockup builder class with a cpp script for test

dimitra97:
This PR creates the mockup  geometry with the chambers from a gdml file and extends the sector with the three chambers at phi.  The picture `barrel_chambers000.png` shows the geometry in an obj file obtained by running the `TestMockupBuilder.cpp` script, which is also in the PR in order to reproduce the geometry.

There is also some work in progress (e.g. python bindings)

@noemina  @asalzburger


Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
```diff
@@ -0,0 +1,344 @@
...
+  // Create the bounds of the detector volumes
+  float radius = straw_surfaces.front()->bounds().values()[0];
+  auto minY = *min_element(positionY.begin(), positionY.end());
+  auto maxY = *max_element(positionY.begin(), positionY.end());
+
+  auto minZ = *min_element(positionZ.begin(), positionZ.end());
+  auto maxZ = *max_element(positionZ.begin(), positionZ.end());
+
+  auto minX = *min_element(positionX.begin(), positionX.end());
+  auto maxX = *max_element(positionX.begin(), positionX.end());
```
noemina, line 125: You can set these parameters while looping on the straws above. Other thing you can use the `minmax_element` to find the smallest and greatest element in a range.

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
```diff
@@ -0,0 +1,344 @@
...
+  auto rmin_middle = det_volumes[1]->center().y() -
+                     2 * det_volumes[1]->volumeBounds().values()[1];
+  auto rmax_middle = det_volumes[1]->center().y() +
+                     3 * det_volumes[1]->volumeBounds().values()[1];
+
+  // radius of the outer cylinder
+  auto rmin_outer = det_volumes.back()->center().y() -
+                    2 * det_volumes.back()->volumeBounds().values()[1];
+  auto rmax_outer = det_volumes.back()->center().y() +
+                    3 * det_volumes.back()->volumeBounds().values()[1];
```
noemina, line 184: The code here, and below, relies on 3 layers (inner, middle and outer) being built. I think we need to generalise this such that adding/ removing one chamber would be possible.

noemina (commented):
Great to see this code appearing! I have added some comments.

Examples/Detectors/Geant4Detector/include/ActsExamples/Geant4Detector/MockupSectorBuilder.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+ private:
+  Config m_cfg;
+
+  G4VPhysicalVolume* g4World = nullptr;
+
+  int maxNumberOfSectors = 8;
+};
+
+}  // namespace ActsExamples
```
andiwand, line 87:
```suggestion
}  // namespace ActsExamples

```

usually we keep an empty line at the end

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+  // creation of the mother volume
+  auto detectorVolume = Acts::Experimental::DetectorVolumeFactory::construct(
+      Acts::Experimental::defaultPortalGenerator(), gctx, "sector", transform,
+      std::move(detectorVolumeBounds),
+      std::vector<std::shared_ptr<Acts::Surface>>{}, detectorVolumesOfSector,
+      Acts::Experimental::allPortalsAndSurfaces());
+
+  return detectorVolume;
+}
```
andiwand, line 344:
```suggestion
}

```

Examples/Detectors/Geant4Detector/include/ActsExamples/Geant4Detector/MockupSectorBuilder.hpp
```diff
@@ -0,0 +1,87 @@
...
+  void setConfiguration(const Config& config);
+
+  /// Set the g4World from the gdml file
+  void setWorld();
+
+  /// Build chamber
+  ///@param gctx The current geometry context object
+  ///@param chamber_config The configuration chamber struct
+  std::shared_ptr<Acts::Experimental::DetectorVolume> BuildChamber(
+      const Acts::GeometryContext& gctx, const ChamberConfig& chamber_config);
```
andiwand, line 69:
```suggestion
  /// Build chamber
  /// @param gctx The current geometry context object
  /// @param chamber_config The configuration chamber struct
  std::shared_ptr<Acts::Experimental::DetectorVolume> buildChamber(
      const Acts::GeometryContext& gctx, const ChamberConfig& chamber_config);
```

Examples/Detectors/Geant4Detector/include/ActsExamples/Geant4Detector/MockupSectorBuilder.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+  /// Build Sector
+  ///@param det_volumes The vector that contains the detector volumes of the Sector
+  ///@param gctx The current geometry context object
+  std::shared_ptr<Acts::Experimental::DetectorVolume> BuildSector(
+      std::vector<std::shared_ptr<Acts::Experimental::DetectorVolume>>
+          det_volumes,
+      const Acts::GeometryContext& gctx);
```
andiwand, line 77:
```suggestion
  /// Build Sector
  /// @param det_volumes The vector that contains the detector volumes of the Sector
  /// @param gctx The current geometry context object
  std::shared_ptr<Acts::Experimental::DetectorVolume> buildSector(
      std::vector<std::shared_ptr<Acts::Experimental::DetectorVolume>>
          det_volumes,
      const Acts::GeometryContext& gctx);
```

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
```diff
@@ -0,0 +1,344 @@
...
+#include <vector>
+
+// a function for sorting the detector volumes from the innermost to the
+// outermost
+
+bool sortByRadialDistance(
+    std::shared_ptr<Acts::Experimental::DetectorVolume>& det_vol1,
+    std::shared_ptr<Acts::Experimental::DetectorVolume>& det_vol2) {
+  return det_vol1->center().y() < det_vol2->center().y();
+}
```
andiwand, line 38: I think this is simple enough to go into a lambda

otherwise it should go into an anonymous namespace

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
```diff
@@ -0,0 +1,344 @@
...
+ActsExamples::MockupSectorBuilder::BuildSector(
+    std::vector<std::shared_ptr<Acts::Experimental::DetectorVolume>>
+        det_volumes,
+    const Acts::GeometryContext& gctx) {
+  if (m_cfg.NumberOfSectors > maxNumberOfSectors) {
+    throw std::invalid_argument("MockupSector:Number of max sectors exceeded");
+    return nullptr;
+  }
+
+  // sort the dtector volumes by their radial distance (from
```
andiwand, line 158:
```suggestion
  // sort the detector volumes by their radial distance (from
```

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
```diff
@@ -0,0 +1,344 @@
...
+      inner_detectorVolumesOfSector = {};
+  std::vector<std::shared_ptr<Acts::Experimental::DetectorVolume>>
+      middle_detectorVolumesOfSector = {};
+  std::vector<std::shared_ptr<Acts::Experimental::DetectorVolume>>
+      outer_detectorVolumesOfSector = {};
+
+  std::vector<std::shared_ptr<Acts::Surface>> shifted_surfaces = {};
+
+  int itr = 0;
+
```
andiwand, line 247:
```suggestion
```

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
```diff
@@ -0,0 +1,344 @@
...
+              Acts::Experimental::allPortalsAndSurfaces());
+
+      // detectorVolumesOfSector.push_back(detectorVolume_sec);
+
+      if (itr == 0)
+        inner_detectorVolumesOfSector.push_back(detectorVolume_sec);
+      if (itr == 1)
+        middle_detectorVolumesOfSector.push_back(detectorVolume_sec);
+      if (itr == 2)
+        outer_detectorVolumesOfSector.push_back(detectorVolume_sec);
```
andiwand, line 293: I think it would be good have braces `{}` around the if body

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
dimitra97: Yes i agree, but my question is if we want to insert the cylinder volumes for each of the layers (inner, middle, outer) regardless of whether there are not all the chambers there . because in that case, I need to define somehow the parameters of the cylinder bounds. Also, according to what I have done so far, all the sectors are identical, should I leave it like that? because if all sectors are identical, i think it is better not to have all the cylinder volumes for the layers (inner,middle,outer) (e. g if the inner chamber is absent, the cylinder volume should also absent, too)

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
```diff
@@ -87,15 +80,20 @@ ActsExamples::MockupSectorBuilder::BuildChamber(
...
+      g4detector.constructDetector(g4World_config, Acts::getDummyLogger());
 
   // The vector that holds the converted sensitive surfaces of the chamber
   std::vector<std::shared_ptr<Acts::Surface>> straw_surfaces = {};
-  // The y and z coordinates of the straws' centers
-  std::vector<float> positionY = {};
-  std::vector<float> positionZ = {};
-  std::vector<float> positionX = {};
+
+   //auto min_max = std::pair<float, float>(std::numeric_limits<float>::max(), -std::numeric_limits<float>::max());
```
noemina, line 80: You can remove this line here.

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
```diff
@@ -87,15 +80,20 @@ ActsExamples::MockupSectorBuilder::BuildChamber(
...
+  auto max_limit = std::numeric_limits<float>::max();
+  auto min_limit = -std::numeric_limits<float>::max();
+
+   std::array<std::pair <float,float>,3> min_max;
+   std::fill(min_max.begin(), min_max.end(), 
+            std::make_pair<float, float>(std::move(max_limit), std::move(min_limit)));
```
noemina, line 86:
```suggestion
            std::make_pair<float, float>(std::move(std::numeric_limits<float>::max()), std::move(-std::numeric_limits<float>::max())));
```

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
```diff
@@ -108,31 +106,34 @@ ActsExamples::MockupSectorBuilder::BuildChamber(
...
 
-  auto minX = *min_element(positionX.begin(), positionX.end());
-  auto maxX = *max_element(positionX.begin(), positionX.end());
 
-  Acts::ActsScalar hx = straw_surfaces.front()->bounds().values()[1] + 10;
-  Acts::ActsScalar hy = 0.5 * (std::abs(minY - maxY) + 2 * radius + 10.);
-  Acts::ActsScalar hz = 0.5 * (std::abs(minZ - maxZ) + 2 * radius + 10.);
+  Acts::ActsScalar hx = straw_surfaces.front()->bounds().values()[1] + m_cfg.toleranceOverlap;
+  Acts::ActsScalar hy = 0.5 * (std::abs(minValues.y() - maxValues.y()) + 2 * radius + m_cfg.toleranceOverlap);
+  Acts::ActsScalar hz = 0.5 * (std::abs(minValues.z() - maxValues.z()) + 2 * radius + m_cfg.toleranceOverlap);
```
noemina, line 121: Why do you have a factor 2 here?
Another comment is on the tolerance parameter. Do you think it is better to have one for the cuboid volume and one for the cylinder?

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
```diff
@@ -146,44 +147,22 @@ ActsExamples::MockupSectorBuilder::BuildChamber(
 }
 
 std::shared_ptr<Acts::Experimental::DetectorVolume>
-ActsExamples::MockupSectorBuilder::BuildSector(
+ActsExamples::MockupSectorBuilder::buildSector(
     std::vector<std::shared_ptr<Acts::Experimental::DetectorVolume>>
         det_volumes,
```
noemina, line 143: A lot of the naming conventions are not followed. 
`det_volumes` should be `detVolumes`, as so on for others...

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
```diff
@@ -204,24 +183,32 @@ ActsExamples::MockupSectorBuilder::BuildSector(
...
+  std::vector<float> halfZ(det_volumes.size());
+  std::vector<std::unique_ptr<Acts::CylinderVolumeBounds>>cylinderVolumesBounds(det_volumes.size());
+
+  for(std::size_t i=0; i<det_volumes.size(); i++){
+    const auto &det_vol = det_volumes[i];
+    rmins[i] = det_vol->center().y() - 
+               det_vol->volumeBounds().values()[1] - m_cfg.toleranceOverlap;
+    rmaxs[i] = std::sqrt(std::pow(det_vol->volumeBounds().values()[0],2) + 
+               std::pow(det_vol->center().y() + 
+               det_vol->volumeBounds().values()[1],2)) + m_cfg.toleranceOverlap; //bla bla from pythagoreum
```
noemina, line 196: :)

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
```diff
@@ -230,40 +217,42 @@ ActsExamples::MockupSectorBuilder::BuildSector(
...
-      middle_detectorVolumesOfSector = {};
-  std::vector<std::shared_ptr<Acts::Experimental::DetectorVolume>>
-      outer_detectorVolumesOfSector = {};
-
+  //create a vector for the shifted surfaces of each chamber
   std::vector<std::shared_ptr<Acts::Surface>> shifted_surfaces = {};
 
+  //creare an array of vectors that holds all the chambers of each sector
+
+  const int array_size =  det_volumes.size();
```
noemina, line 216: You can move this line at l193 and use `array_size` (should be `arraySize`).

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
dimitra97: Should I also change the names of m_cfg to mCfg , gml_path etc?

andiwand: for member variable we have the convention to start with `m_`. other that that it should be camelCase I think. but you will see snake_case from time to time which is also fine I think but camelCase should be preferred for new code

Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp
dimitra97: actually i think we said to do that directly:
  ```
 std::vector<std::unique_ptr<Acts::CylinderVolumeBounds>>
      cylinderVolumesBounds(detVolumes.size());
``` 
?? is it better?

File: Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp (hunk 13 above)
Comment (dimitra97) [reply Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp:121]: I have the factor 2 because we have one radius from the innermost surface plus one from the outermost,, since minY and maxY give me the centers of the surfaces. 
I am not sure about the the tolerance. I  think it is ok to keep it general and use it for both cases? But you can tell me what you think is better

Comment (noemina) [reply Examples/Detectors/Geant4Detector/src/MockupSectorBuilder.cpp:121]:
```suggestion
  Acts::ActsScalar hy = 0.5 * ((maxValues.y()+radius)-(minValues.y()-radius) + m_cfg.toleranceOverlap);
  Acts::ActsScalar hz = 0.5 * ((maxValues.z()+radius)-(minValues.z()-radius) + m_cfg.toleranceOverlap);
```

noemina:
@dimitra97, the configuration of the build fails because the your package requires actsvg.

noemina (approved):

andiwand:
:tada:

PR #2009, opened 2023-04-03
fix: MultiTrajectory parameters and covariance constness

paulgessinger:
I believe these functions were not fully usable before if used through a const `MultiTrajectory`.


benjaminhuth (approved):
looks good!

PR #2010, opened 2023-04-03
refactor: move multipleScattering2 to seeding options

LuisFelipeCoelho:
This PR moves the calculation of `m_config.maxScatteringAngle2 * m_config.sigmaScattering * m_config.sigmaScattering` to the seeding option since it can be done only once


CarloVarni (approved):

paulgessinger:
Is this a breaking change?

CarloVarni:
@paulgessinger I do not think so. This adds a variable to the options but it is internally computed, so it's not exposed to the user

PR #2011, opened 2023-04-03
ci: Patch geant4 config script to download datasets

paulgessinger:


.github/workflows/builds.yml
```diff
@@ -236,7 +236,8 @@ jobs:
       - name: Python level tests
         shell: bash
         run: >
-          /usr/local/bin/geant4-config --install-datasets
+          patch /usr/local/bin/geant4-config CI/geant4-config.patch
```
andiwand, line 239: `CI/geant4-config.patch` is missing not pushed yet right?

paulgessinger: dang :D

paulgessinger: It's there now.

andiwand (approved):

PR #2012, opened 2023-04-03
feat: Track level improvements

### LINKED ISSUE #1999: Issue with track.nMeasurements() in CKF tracks
When calling track.nMeasurements() on tracks reconstructed with the CKF, the return value is always 0. 
@paulgessinger checked and confirmed that CKF doesn't fill this correctly (While KF and other fitters seem like they do)

paulgessinger:
This PR does a few things:

- Add a method `appendTrackState` to `TrackProxy`, allowing convenient building of track state chains
- Add properties `nOutliers`, `nSharedHits`, `chi2` and `nDoF` to tracks
- Adds a helper function that will calculate these values from the track states and store them on the track. That is similar to the already existing function `trajectoryState` in `MultiTrajectoryHelpers`, but it stores the results on the tracks themselves.
- The helper function is called by the 3 fitters and the CKF, so tracks should always have these numbers.

Closes #1999

Blocked by:

- #2007
- #2011


Core/include/Acts/EventData/TrackHelpers.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+/// Helper function to calculate a number of track level quantities and store
+/// them on the track itself
+/// @tparam track_container_t the track container backend
+/// @tparam track_state_container_t the track state container backend
+/// @tparam holder_t the holder type for the track container backends
+/// @param track A track proxy to operate on
+template <typename track_container_t, typename track_state_container_t,
+          template <typename> class holder_t>
+void calculateTrackQuantities(
```
paulgessinger, line 23: @pbutti

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -1379,6 +1380,9 @@ class CombinatorialKalmanFilter {
       track.parameters() = parameters.parameters();
       track.covariance() = *parameters.covariance();
       track.setReferenceSurface(parameters.referenceSurface().getSharedPtr());
+
+      calculateTrackQuantities(track);
```
paulgessinger, line 1384: @pbutti

Core/include/Acts/EventData/TrackProxy.hpp
```diff
@@ -385,6 +386,19 @@ class TrackProxy {
     return m_container->trackStateRange(m_index);
   }
 
+  /// Append a track state to this track. This will modify the tip index to
+  /// point at the newly created track state, which will be directly after the
+  /// previous track state at tip index.
+  /// @param mask The allocation prop mask for the new track state
+  /// @return The newly added track state
+  template <bool RO = ReadOnly, typename = std::enable_if_t<!RO>>
+  auto appendTrackState(TrackStatePropMask mask = TrackStatePropMask::All) {
```
paulgessinger, line 395: @EdwardMoyse

Core/include/Acts/EventData/TrackHelpers.hpp
pbutti: LGTM. Only comment would be to add in the description some words about the "false" in the template arguments. What templated argument is that?

paulgessinger: They're unfortunately necessary to accept *any* kind of track container, not just a concrete one. I tried to clarify this with the comment above this:

```cpp
/// @tparam track_container_t the track container backend
/// @tparam track_state_container_t the track state container backend
/// @tparam holder_t the holder type for the track container backends
/// @param track A track proxy to operate on
```

EdwardMoyse: You have four template parameters here:
```cpp
Acts::TrackProxy<track_container_t, track_state_container_t, holder_t, false>
```
You explain what the first three are, but you don't explain why the final one is false. This is, I assume, what @pbutti is referring to?

I guess this is duplicated documentation from `TrackProxy` but it probably wouldn't hurt to say that since it will be modified, the passed `TrackProxy` must be writable, i.e. have `read_only` = false?

EdwardMoyse (approved):
Looks good to me. Thanks!

Core/include/Acts/EventData/TrackHelpers.hpp
paulgessinger: I see, and yes that's correct. I can add a note to the documentation of the `@param` (because it's not a template parameters to this functions).

EdwardMoyse: Yes, I meant add it to `@param track`

EdwardMoyse (approved):

PR #2021, opened 2023-04-08
refactor: change the definition of the rotation parameters

XiaocongAi:
This PR includes the following changes:
1) Change the definition of the rotation parameters to be around local axes
2) Extend the `AlignmentHelper` to get derivative of the surface local axes w.r..t rotation of a (mother) composite object


pbutti:
Hi Xiacong, 

In order to check a little bit more about the derivative of the residuals wrt local rotations could you point me to the relevant files to run some misalignment examples? Additionally, is there a support for unbiased residual monitoring in place?

I will be tied up a little bit until Friday, but I will be free Friday to check this PR in more detail.

XiaocongAi:
> Hi Xiacong,
> 
> In order to check a little bit more about the derivative of the residuals wrt local rotations could you point me to the relevant files to run some misalignment examples? Additionally, is there a support for unbiased residual monitoring in place?
> 
> I will be tied up a little bit until Friday, but I will be free Friday to check this PR in more detail.

Hi @pbutti , I'm not very sure if the available example in ACTS works well. @LaraCalic , is it possible for you to provide the command you used to run the misalignment example so that we could do some test for the changes made in this PR? Thanks!

XiaocongAi:
The Python bindings and the residual monitoring will be very helpful for future development. In the current example, there are two many alignment parameters as all the modules are decorated with misalignment. So maybe one thing to add in the example is configurable alignable detector elements so that we can test the alignment more subtly.

gagnonlg:
Still relevant, I think. @XiaocongAi are we still waiting for a way to properly test this change?

pbutti:
Hi @gagnonlg 
my bad, I should finish review this code and try to test it. If @LaraCalic could provide the command to run the misalignment example I can help.

Core/src/Surfaces/detail/AlignmentHelper.cpp
```diff
@@ -13,41 +13,51 @@
...
-  double cy = std::cos(rotAngles(1));
-  double sz = std::sin(rotAngles(0));
-  double cz = std::cos(rotAngles(0));
+  // const Vector3 rotAngles = rotation.eulerAngles(2, 1, 0);
+  // double sx = std::sin(rotAngles(2));
+  // double cx = std::cos(rotAngles(2));
+  // double sy = std::sin(rotAngles(1));
+  // double cy = std::cos(rotAngles(1));
+  // double sz = std::sin(rotAngles(0));
+  // double cz = std::cos(rotAngles(0));
```
paulgessinger, line 27:
```suggestion
```

Tests/UnitTests/Core/Surfaces/AlignmentHelperTests.cpp
```diff
@@ -54,25 +54,21 @@ BOOST_AUTO_TEST_CASE(alignment_helper_test) {
...
   // Calculate the expected derivative of local z axis to its rotation
   RotationMatrix3 expRotToZAxis = RotationMatrix3::Zero();
-  expRotToZAxis.col(0) =
-      Vector3(sz * cx - cz * sy * sx, -sz * sy * sx - cz * cx, -cy * sx);
-  expRotToZAxis.col(1) = Vector3(cz * cy * cx, sz * cy * cx, -sy * cx);
-  expRotToZAxis.col(2) =
-      Vector3(cz * sx - sz * sy * cx, cz * sy * cx + sz * sx, 0);
+  expRotToZAxis.col(0) = expRot * Vector3(0, -1, 0);
+  expRotToZAxis.col(1) = expRot * Vector3(1, 0, 0);
+  expRotToZAxis.col(2) = expRot * Vector3(0, 0, 0);
```
paulgessinger, line 71: Are the numbers being assigned still the same? Can you elaborate why / how this changes (it's not obvious to me)?

paulgessinger:
I've added two comments. The associated unit test is updated, and otherwise I don't think we have a good way to test this at the moment. I'd be in favor of merging this in it's current form and revisiting in the future.

Core/src/Surfaces/detail/AlignmentHelper.cpp
XiaocongAi: Done.

Tests/UnitTests/Core/Surfaces/AlignmentHelperTests.cpp
XiaocongAi: I have added some comments to briefly explain the mathematics behind this.

pbutti:
Hi @XiaocongAi
I am taking a deeper look to the alignment implementation and in particular to this request. I would like to start a discussion with some questions to get better understanding of the current implementation and bring this in as I'm starting to taking a look for alignment for LDMX as well.  
I would like to start by discussing the implementation of the rotation derivatives in the alignment helper. 
I must say I got a bit lost in the logic on how the composite structure is brought in and how the derivatives are derived (pun not intended)

1) Using R * R^T = 1, would expect d/dtheta R = S * R, where S is the skew-symmetric matrix. 
From line 
https://github.com/acts-project/acts/blob/e964f118105f8b7226bd8a13d5feee727430302f/Core/src/Surfaces/detail/AlignmentHelper.cpp#L28
You have d/dx R = R*S. What am I missing?

2) I'm confused a bit about the logic of the code. I must admit probably I should start running some tests, but I don't grasp how the hierarchy would be computed from the sensor derivatives to global structure derivatives. If I would like to obtain the derivatives of the first barrel layer rotation around Z, I would need to transform the derivatives of the sensors to a Rz rotation of the layer. Is this equivalent to what is done in 
https://github.com/acts-project/acts/blob/e964f118105f8b7226bd8a13d5feee727430302f/Core/src/Surfaces/detail/AlignmentHelper.cpp#L15 ?

Perhaps we could talk directly about the details?

XiaocongAi:
Hi @pbutti , I'm attaching a personal note
[AlignmentFormula.pdf](https://github.com/acts-project/acts/files/13301033/AlignmentFormula.pdf)
 (I'm planning to transform it to a white paper soon) where I derived all those equations. It will be great if you can have a look. And of course, we can have a chat about this.

andiwand:
how do we proceed here @pbutti @paulgessinger @XiaocongAi ?

paulgessinger (approved):
Let's get this in for now, we can still make changes if we learn more about this.

andiwand:
looks like there is a breaking unit test @XiaocongAi @paulgessinger @pbutti

paulgessinger:
Hmm. This seems new. It also is not a precision issue I don't think, so might need a fix.

XiaocongAi:
Hi @paulgessinger @andiwand @AJPfleger , sorry again for the delay. It should be ready for the merge.

paulgessinger (approved):
As far as I can judge, this looks good. Let's get it in.

acts-project-service:
## :red_circle: Athena integration test results [e7d6bf9f1766fdb4345186ed6976c630f3b275a8]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42257531) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7920686)!

PR #2022, opened 2023-04-11
feat: EDM4hep Track reader + writer

paulgessinger:
This PR adds a writer and a reader for `TrackContainer` to and from EDM4hep track objects and track states.

This is unfortunately not lossless:
- EDM4hep only accepts *perigee* parameters, so all track state hits get converted to bound parameters relative to a perigee centered at the hit position. When reading track states back in, this representation is kept (as there is no way to recover the original sensor). 
- The track parameters are converted to the LCIO parametrization, and the covariance is stored as a lower left triangle


Core/include/Acts/Utilities/Zip.hpp
```diff
@@ -21,16 +21,16 @@ namespace Acts {
...
   using ItB = decltype(rb.begin());
 
   struct It {
     ItA a;
     ItB b;
 
-    using reference =
-        std::tuple<typename ItA::reference, typename ItB::reference>;
+    using reference = std::tuple<decltype(*std::declval<ItA>()),
+                                 decltype(*std::declval<ItB>())>;
```
paulgessinger, line 33: I needed this to make it work with the proxy iterators. Do you see any issues @asalzburger?

asalzburger: Should be ok.

Plugins/EDM4hep/src/EDM4hepUtil.cpp
```diff
@@ -0,0 +1,22 @@
...
+#include "Acts/EventData/MultiTrajectoryHelpers.hpp"
+#include "Acts/Utilities/Helpers.hpp"
+
+#include "edm4hep/TrackState.h"
+
+namespace ActsExamples {
+
+// void EDM4hepUtil::writeTrack(TrackContainer::ConstTrackProxy track,
+// edm4hep::MutableTrack to) {}
+}  // namespace ActsExamples
```
andiwand, line 22:
```suggestion
 void EDM4hepUtil::writeTrack(TrackContainer::ConstTrackProxy track,
                              edm4hep::MutableTrack to) {}

}  // namespace ActsExamples
```

Plugins/EDM4hep/include/Acts/Plugins/EDM4hep/EDM4hepUtil.hpp
```diff
@@ -0,0 +1,420 @@
...
+}
+
+inline void unpackCovariance(const float* from, ActsSymMatrix<6>& to) {
+  auto k = [](size_t i, size_t j) { return (i + 1) * i / 2 + j; };
+  for (int i = 0; i < to.rows(); i++) {
+    for (int j = 0; j < to.cols(); j++) {
+      to(i, j) = from[j <= i ? k(i, j) : k(j, i)];
+    }
+  }
+}
```
andiwand, line 135: I wonder if eigen could do this for us?

andiwand (approved):
nice work! code looks very clear and clean to me. left two minor comments

Plugins/EDM4hep/include/Acts/Plugins/EDM4hep/EDM4hepUtil.hpp
paulgessinger: Not that I'm aware. If you find anything let me know!

paulgessinger:
I found some potential further issues locally on macOS. Taking this WIP until fixed.

andiwand (approved):

Issue #2023, opened 2023-04-12
ML Ambi python level test throws FPE *sometimes*


paulgessinger:
This unfortunately does not seem to be fully deterministic.

See: https://github.com/acts-project/acts/actions/runs/4673301747/jobs/8276745945

Any ideas @Corentin-Allaire?

benjaminhuth:
Just a rough idea without looking into the error or the code: Could it be that an rng inside ONNX is not seeded properly?

Corentin-Allaire:
Just to be sure FPE means that it is the job that failed not that the test didn't return the correct value. Right ?

Corentin-Allaire:
My guess is that it is somehow related to #2005 but since it passed the CI I am not sure how... What seem to be happening is that we overflow in Onnx, that as probably something to do with data preparation. Maybe we do not always load the model properly ? (I don't see why but that could be an issue)

benjaminhuth:
Another potential source of this problem could be uninitialized memory somewhere in the input data? Again, just guessing, not having tested anything.

paulgessinger:
> Just to be sure FPE means that it is the job that failed not that the test didn't return the correct value. Right ?

Exactly, @Corentin-Allaire

Corentin-Allaire:
Just to be sure could anybody have a look at the line in the test : https://github.com/acts-project/acts/blob/6e3571201557050fbe6938f4f1c2ebf2de93fb14/Examples/Python/tests/test_examples.py#L1234#L1267

Does this looks correct to you ? I am asking because it seems like the code work outside the test

paulgessinger:
I guess the problem is that the FPE is not reproducible. So for some reason, the CI runs into a an edge case where some FP operation is invalid.

Corentin-Allaire:
What surprise me is that even if it is not reproducible if we run this on a lot of events the error should appear again no ?

benjaminhuth:
Not sure if this would happen in any case... If it is a uninitialized memory bug for example, then I would expect it to appear independently of the number of events, but more dependent on the preconditions in memory...

Corentin-Allaire:
> Not sure if this would happen in any case... If it is a uninitialized memory bug for example, then I would expect it to appear independently of the number of events, but more dependent on the preconditions in memory...

In that case maybe it has nothing to do with the solver itself and it is just related to calling `full_chain_odd.py` twice in the test ?

paulgessinger:
[...]

PR #2026, opened 2023-04-12
fix: Solve the CI issue with the ML solver

Corentin-Allaire:
This ML will try to solve the issue #2023. Unfortunately I couldn't reproduce that issue on my machine so I have to test it directly on the CI.


Corentin-Allaire:
The issue should now be fixed. What was happening was that for the input, I was declaring an Eigen::Array with a size based on the number of tracks in the collection. But the Solver removes the tracks with less measurement than the minimum required, so the size of the Eigen::Array was too large, and the last few elements were just random memory that sometimes was interpreted as nan.

The good news is that we were only reading the entry corresponding to the tracks of interest afterwards, so the performance shouldn't be affected (actually, the code should be slightly faster for a larger number of required measurements).

andiwand (approved):
nice fix!

PR #2027, opened 2023-04-12
refactor: and fix new detector geometry

andiwand:
- detector volumes now also have a volume finder delegate
- detector volume inside method will not check against its subvolumes
- detector will not add subvolumes of its volumes to its list


Core/src/Detector/DetectorVolume.cpp
```diff
@@ -103,12 +107,6 @@ bool Acts::Experimental::DetectorVolume::inside(const GeometryContext& gctx,
   if (not excludeInserts or m_volumes.external.empty()) {
     return true;
   }
-  // Check exclusion through subvolume
-  for (const auto v : volumes()) {
```
paulgessinger, line 107: Not sure I understand this change, is this not what we want? Is it not needed anymore?

andiwand: I didn't see a reason why we would put a hole into our volume where there are subvolumes... @asalzburger did you have a specific use case in mind? because for me it broke the lookup mechanism for the volume

Core/include/Acts/Detector/PortalHelper.hpp
```diff
@@ -22,6 +22,11 @@
 namespace Acts {
 namespace Experimental {
 
+/// The Portal genertor definition
+using PortalGenerator = Delegate<std::vector<std::shared_ptr<Portal>>(
```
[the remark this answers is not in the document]
andiwand: true that trying to cleaning this up right now

but in principal repeated typedefs are okay no? the thing is we repeated the full type before and I thought this is basically as good as repeated typedefs

Core/src/Detector/DetectorVolume.cpp
andiwand: I see. I also refactored the lookup mechanism because it didn't give me the right volume if the hierarchy is too deep. it also looks to me like this `inside` method could break after 2 levels down the tree?

maybe we can discuss this tomorrow

asalzburger: Absolutely, I think we can achieve both - a hierarchy based version and a unique assignment. 

For me the single requirement would be that you can unique associate - and thus let the portals do the guiding through the detector.

My morning is fully booked, I will find you as soon as I am done.

Core/include/Acts/Detector/Detector.hpp
```diff
@@ -73,12 +66,22 @@ class Detector : public std::enable_shared_from_this<Detector> {
...
+  const DetectorVolume* rootVolume() const;
+
+  /// Non-const access to the root volume
   ///
-  /// @return the volumes shared pointer store
+  /// @return the root volume shared pointer
   std::vector<std::shared_ptr<DetectorVolume>>& volumePtrs();
 
-  /// Const access to sub volumes
+  /// Const access to the root volume
```
andiwand, line 84:
```suggestion
  /// Const access to sub volumes
```

Core/include/Acts/Detector/Detector.hpp
```diff
@@ -73,12 +66,22 @@ class Detector : public std::enable_shared_from_this<Detector> {
...
+
+  /// Const access to the root volume
+  ///
+  /// @return a vector to const DetectorVolume raw pointers
+  const DetectorVolume* rootVolume() const;
+
+  /// Non-const access to the root volume
   ///
-  /// @return the volumes shared pointer store
+  /// @return the root volume shared pointer
```
andiwand, line 81:
```suggestion
  /// @return the volumes shared pointer store
```

Core/include/Acts/Detector/Detector.hpp
```diff
@@ -124,37 +128,18 @@ class Detector : public std::enable_shared_from_this<Detector> {
...
 
   /// Name/index map to find volumes by name and detect duplicates
   std::unordered_map<std::string, size_t> m_volumeNameIndex;
 };
 
-inline std::vector<std::shared_ptr<DetectorVolume>>& Detector::volumePtrs() {
-  return m_volumes.internal;
-}
-
-inline const std::vector<const DetectorVolume*>& Detector::volumes() const {
```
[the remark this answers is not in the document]
andiwand: yes this function will still do that. previously it only handled sub and subsub volumes. this should be fixed now

Core/include/Acts/Detector/Portal.hpp
```diff
@@ -119,13 +114,13 @@ class Portal : public std::enable_shared_from_this<Portal> {
...
   ///
-  /// @param nDir the navigation direction for the link
+  /// @param dir the direction of the link
   /// @param dVolumeUpdator is the mangaged volume updator delegate
   /// @param attachedVolumes is the list of attached volumes for book keeping
   ///
   /// @note this overwrites the existing link
   void assignDetectorVolumeUpdator(
-      NavigationDirection nDir, DetectorVolumeUpdator&& dVolumeUpdator,
+      NavigationDirection dir, DetectorVolumeUpdator&& dVolumeUpdator,
```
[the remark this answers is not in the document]
andiwand: yeah exactly

Core/include/Acts/Detector/DetectorVolume.hpp
```diff
@@ -186,15 +191,23 @@ class DetectorVolume : public std::enable_shared_from_this<DetectorVolume> {
...
+
+  /// Check if a point is exclusively inside this volume i.e. this point is not
+  /// inside a subvolume.
+  ///
+  /// @param gctx the geometry context
+  /// @param position the position for the inside check
+  ///
+  /// @return a bool to indicate inside/outside
+  bool exclusivelyInside(const GeometryContext& gctx,
+                         const Vector3& position) const;
```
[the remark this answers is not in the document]
andiwand: exactly

Core/include/Acts/Detector/Portal.hpp
```diff
@@ -158,22 +153,5 @@ class Portal : public std::enable_shared_from_this<Portal> {
...
-}
-
-inline const Portal::DetectorVolumeUpdators& Portal::detectorVolumeUpdators()
-    const {
-  return m_volumeUpdators;
-}
-
-inline Portal::AttachedDetectorVolumes& Portal::attachedDetectorVolumes() {
-  return m_attachedVolumes;
-}
```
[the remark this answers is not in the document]
andiwand: yes! that solved some header dependency issues for me

andiwand:
handled your comments @asalzburger - thanks!

asalzburger (approved):
Alright then, let's move this in.

PR #2028, opened 2023-04-12
feat: Add method to return measurement indices

gagnonlg:
Add a method to return the measurement indices. This is useful to implement measurement calibration because to do this you basically need to 1) expand measurement parameters 2) calibrate 3) project back to the original subspace and 4) create a new measurement object. But, to create that object, we need to have the same ordering of the indices and it's cumbersome to obtain it without such a method. For instance, this is the best I could do:

```
	    std::array<Acts::BoundIndices, kSize> indices;
	    for (size_t i = 0; i < kSize; i++) {
		Eigen::Index _, c;
		double v = P.row(i).maxCoeff(&_, &c); // P is meas.projector()
		assert(v > 0 && "invalid indices");
		indices[i] = static_cast<Acts::BoundIndices>(c);
	    }
```

However, this probably needs a bit of discussion. For instance, currently the method will cast the subspace indices from uint8_t to the measurement indice type, and when the measurement is created they will just be cast back to uint8_t which seems wastefull. Not sure it's an issue but if we want to avoid this we could return the subspace itself and add a new constructor for Measurement. But that's more intrusive.

Thoughts, @asalzburger @paulgessinger ?


Core/include/Acts/EventData/Measurement.hpp
```diff
@@ -103,6 +103,16 @@ class Measurement {
...
+  /// The measurement indices
+  constexpr std::array<indices_t, kSize> indices() const {
+    std::array<uint8_t, kSize> subInds = m_subspace.indices();
+    std::array<indices_t, kSize> inds;
+    for (size_t i = 0; i < kSize; i++) {
+      inds[i] = static_cast<indices_t>(subInds[i]);
+    }
+    return inds;
+  }
+
```
paulgessinger, line 115: I think realistically, the bit width is going to be enough, so let's do it.

paulgessinger (approved):

Core/include/Acts/EventData/Measurement.hpp
gagnonlg: yeah thinking about it now, the roundtrip out and back in the subspace most likely gets optimized away.

PR #2030, opened 2023-04-13
refactor!: Generalize NavigationDirection to Direction

andiwand:
I realized that `NavigationDirection` was used in various different parts of our code in terms of a general direction. I saw two options to resolve this: create another enum to disambiguate them or generalize the `NavigationDirection` to `Direction`. Opted for the later because it seems to be the same concept and we do not use the same thing with two different meanings in a single place.

On top I added two new aliases `Positive` and `Negative` which made more sense to me in the context of surfaces and their normal. Also I added a few more utility functions and used them across the codebase.


paulgessinger (commented):
Seems reasonable. Is this a breaking change?

andiwand:
> Is this a breaking change?

I would think so because it is a very central type that changed name. But I don't know if it is actually used downstream

andiwand:
seeing all the free functions to handle the direction I wonder if it might be worth to box this in a class. any opinions @paulgessinger ?

paulgessinger:
Could be a class or a namespace if they're all free. If you think it makes sense grouping them logically.

andiwand:
I wrapped the direction into a class now which should prevent conversions further. What is not so nice is that `Forward` is not a `Direction` object but is auto converted into such. So we cannot do `Direction::Forward.index()` and so I added helper functions like `Direction::index(Direction::Forward)`

Core/include/Acts/TrackFitting/detail/GsfUtils.hpp
```diff
@@ -124,11 +124,8 @@ class ScopedGsfInfoPrinterAndChecker {
                  << stepper.direction(state.stepping).transpose()
                  << " and momentum " << stepper.momentum(state.stepping)
                  << " and charge " << stepper.charge(state.stepping));
-    ACTS_VERBOSE("Propagation is in "
-                 << (state.stepping.navDir == NavigationDirection::Forward
-                         ? "forward"
-                         : "backward")
-                 << " mode");
+    ACTS_VERBOSE("Propagation is in " << state.stepping.navDir.toString()
```
paulgessinger, line 127: The outstream operator should be defined here right?

Core/include/Acts/Definitions/Direction.hpp
```diff
@@ -0,0 +1,174 @@
...
+
+ private:
+  Value m_value = Value::Positive;
+};
+
+std::ostream& operator<<(std::ostream& os, Direction dir);
+
+// Direction * T
+
+inline constexpr auto operator*(Direction dir, int value) {
```
paulgessinger, line 118: I think in these cases I'd probably prefer explicit return types, rather than `auto`.

Core/include/Acts/TrackFitting/detail/GsfUtils.hpp
andiwand: true. will change that

Core/include/Acts/Definitions/Direction.hpp
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  /// Invert a direction
+  ///
+  /// @param dir is the direction at input
+  ///
+  /// @return an opposite direction
+  static inline constexpr Direction invert(Direction dir) {
+    return dir.invert();
+  }
```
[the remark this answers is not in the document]
andiwand: I prefer the object functions but there were some occasions where I could not directly call them because the literals like `Forward` are actually not of the `Direction` type but auto convertible. but maybe it is good enough explicitly convert in those cases

andiwand:
@paulgessinger the class implementation works now if you want to take another look

Core/include/Acts/Definitions/Direction.hpp
```diff
@@ -0,0 +1,161 @@
...
+  /// @param zeroAsPositive if set, 0 will be handled as "positive"
+  ///
+  /// @return a direciton enum
+  static inline constexpr Direction fromScalar(ActsScalar scalar,
+                                               bool zeroAsPositive = false) {
+    if (!zeroAsPositive) {
+      assert(scalar != 0);
+    }
+    return scalar >= 0 ? Value::Positive : Value::Negative;
+  }
```
paulgessinger, line 46: Would it be a good idea to have a separate function `fromScalarZeroAsPositive` to make it super extra clear, or is the argument enough?

Core/include/Acts/Definitions/Direction.hpp
```diff
@@ -0,0 +1,169 @@
...
+/// other general axes
+class Direction final {
+ private:
+  enum class Value : int {
+    Negative = -1,
+    Positive = 1,
+  };
+
+ public:
+  static constexpr auto Negative = Value::Negative;
```
benjaminhuth, line 28: Just a question: Wouldn't it make more sense to make static member functions like 

```c++
static constexpr auto Negative {
   return Direction(Value::Negative);
}
```

Because otherweise we either need to construct the Direction via `Direction(Direction::Negative)` or rely on implicit construction, right?

andiwand: yeah I also wondered if static functions would be a good alternative. to me it looked a bit odd because you need a function call to access the constants. as you said right now I rely on implicit conversions which is also not great

any opinion on that @paulgessinger ?

paulgessinger: Implicit conversion from a well-defined type should be ok though?

benjaminhuth: > Implicit conversion from a well-defined type should be ok though?

In principle yes, but this can also lead to unexpected errors, e.g. for template argument deduction (no implicit conversion done, but not sure if that is a realistic use case

Another advantage of the static function (as kind of a factory function) would be that we could make the constructor private and completely keep the enum internal.

I would like the static function approach more, but I'm fine with either of both. Probably not good investement of developer time to discuss this too lengthy^^

Core/src/Definitions/Direction.cpp
```diff
@@ -0,0 +1,28 @@
...
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#include "Acts/Definitions/Direction.hpp"
+
+#include <ostream>
+
+std::string Acts::Direction::toString() const {
```
benjaminhuth, line 13: Actually, in order to be a bit more efficient, we could return `std::string_view` here, right?

asalzburger:
@paulgessinger - can you check if this satisfies your comments?

andiwand:
resolved some conflicts with main and updated the unit tests. should be in a mergable state again

asalzburger (approved):
Impressive work, I will flag the PR as breaking change though.

PR #2032, opened 2023-04-14
ci: Fix skipped Geant4 tests, always enable all tests by default in CI

paulgessinger:
I'm manually skipping the exatrkx test in the default CI run via `-k` now.


paulgessinger:
There are some hash mismatches:

```
test_geant4__particles_final.root: ba00c21e6f0696316000f5ff7f0fb27df72d191157043bfa815844d8897056b7
test_geant4__particles_initial.root: 9eea306cd4ece5f509d73c0cc8de1bd321b2b40646bfcd5cc148a7ccf77980d8
test_geant4__hits.root: 6ab07fe4fe596c6948d7c4b821e51f6fae4ba69ac5cc62c07798c7de1cedfccb
```

I guess we didn't catch those when I bumped the G4 version because the jobs stopped being run. I'm probably going to just update them.

/cc @andiwand

Examples/Python/tests/helpers/__init__.py
```diff
@@ -74,6 +77,11 @@
 
 isCI = os.environ.get("CI", "false") == "true"
 
+if isCI:
+    for k, v in dict(locals()).items():
+        if k.endswith("Enabled"):
+            locals()[k] = True
```
andiwand, line 83: wow that looks evil :smile:

andiwand (approved):

Examples/Python/tests/helpers/__init__.py
paulgessinger: That way it turns on everything `*Enabled`, and we can't forget by accident.

andiwand: I guess one down side is that we will not know if the flags are corrupted again. do you think it makes sense to assert them to be turned instead of silently turning them on?

paulgessinger: That might be another alternative, yeah.

andiwand (approved):

PR #2034, opened 2023-04-17
feat: Minimal example for using C++20 concepts

paulgessinger:
This adds a very simple `SurfaceVisitorConcept` to test the conditional implementation.


paulgessinger:
The concept is intentionally wrong (takes a reference instead of a pointer) to showcase the error we (hopefully) get in the CI.

benjaminhuth (approved):
Very nice!

paulgessinger:
It did fail, and then I fixed the concept, so now we indeed expect success.

PR #2035, opened 2023-04-17
feat: Add ConstRefHolder type, add support to TrackContainer

paulgessinger:
This allows construction of `TrackContainer` from const references.


EdwardMoyse (approved):
Looks good to me.

paulgessinger:
Compiler segfault... nice.

paulgessinger:
Ok I had to drop the `explicit` from `ValueHolder` and go back to using implicit construction of the holders in the constructor arguments. This is not ideal, but at least the `ValueHolder` constructor is not as wide open as the reference or const reference one. Let's see if the CI agrees with me.

paulgessinger:
Segfault is gone, can you reapprove @EdwardMoyse ?

EdwardMoyse (commented):
Looks good

EdwardMoyse (approved):

PR #2036, opened 2023-04-17
build: Add tests for non-compilation of invalid behavior

paulgessinger:
This tests if certain constructs **do not** compile, i.e. to ensure compile time checks correctly remove something from overload resolution. Mainly aimed at `TrackContainer` and `MultiTrajectory`


Tests/UnitTests/CMakeLists.txt
```diff
@@ -25,6 +25,68 @@ macro(add_unittest _name)
   add_test(NAME ${_name} COMMAND ${_target})
 endmacro()
 
+function(add_non_compile_test name src)
```
benjaminhuth, line 28: Could you add a few lines of comment here to describe the high-level approach of these tests?
I think this is a quite complicated bit of `cmake` code, and I think that would help future developers quite a bit.

benjaminhuth (changes requested):
Really impressive work here to get this working! The only thing I think could be improved is the documentation of some parts (see comment)

benjaminhuth (approved):
Okay, let's get it in!

Issue #1778, opened 2023-01-09
Provide tools & setup for tracking of Heavy Ion on ITk evens in standalone ACTS environment


tboldagh:
The goal is to work out tools & setup to demonstrates ACTS tracking chain on for Heavy Ion events in ATLAS ITk detector.
In comparison to an existing chain the HI reconstruction is likely to need fast vertex z estimation and a modification to the seeding that would allow to contain the seeds to that vertex only. (Default chain should be obviously tried first to obtain reference.)

A concrete work items are:
 - review existing vertex Z finders
 - implement the vertex Z finder in ACTS 
 - consider implementations suitable for accelerators (not the 1st priority)
 - implement example algorithm using it
 - extend the functionality of the seed finder to also accept z vertex constraint (not only beam spot position & mag field as is now)
 -  optimise setup
 - consider adding plugin to hijing generator - only if particle gun will be deemed insufficient for testing - coordinate with @plariono who works on ALICE on it
 - integrate the solution with Athena

tboldagh:
ATLAS implementation in trigger can be found here:
https://gitlab.cern.ch/atlas/athena/-/blob/master/Trigger/TrigTools/IDScanZFinder/src/IDScanZFinderInternal.h

tboldagh:
And let me tag FI: @noemi @toyamaza @asalzburger

paulgessinger:
@tboldagh is there a plan for this?

pbalek:
Hi all,
I think PR #2038 is in a good shape. I tuned the parameters for Pb+Pb collisions, so it runs quite fast while it has a good resolution.

Now, we can skip some phi a z bins. That needs to be set as a parameter of the tool. Ideally, the tool could decide itself but let's keep that for the next version.
Also, I added some iterations, so some fraction of the worst planes/rays can be discarded each pass.

As of now, the tool is not used anywhere. However, I've seen #2195, so once that is merged, I think we can utilize the seed vertex finder there.

tboldagh:
There is an interesting evolution of this project. Updates to be seen soon.

PR #2038, opened 2023-04-17
feat: Seed Vertex Finder

pbalek:
This PR adds an algorithm for fast vertex estimation, `SingleSeedVertexFinder`. I can be used when a single vertex is present in the event. It uses spacepoints as its input and can thus run before track reconstruction.

The spacepoints are sorted based on their distance from Z-axis to `near`, `middle`, and `far` classes. All possible triplets are attempted to be created, one spacepoint from each class; only those that are reasonably straight and point nearby Z-axis are kept.

The vertex estimation has two modes:
1. Minimalization with respect to "planes"
   * The three spacepoints define a plane. A distance from the plane to a point (=vertex) is calculated. The sum of squares of the distances should be minimal. 
2. Minimalization with respect to "rays"
   * The three spacepoint are fitted with a straight line (=ray). A distance from the ray to a point (=vertex) is calculated. The sum of squares of the distances should be minimal. 

`SingleSeedVertexFinderAlgorithm` implements the `SingleSeedVertexFinder` and run it. Found vertices are saved by `VertexPerformanceWriter`.

Unit tests run in two cases for both "planes" and "rays". In one case, a small set of spacepoints is used whose output can be calculated by hand. In the second case, a larger set of spacepoints is generated using `std::mt19937` random number generator.

Related to issue #1778. Tagging @tboldagh


Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+
+  /// @brief Calculates parameters of the ray (starting point + direction), given the three points
+  /// @param triplet A single triplet (with 3 spacepoints)
+  /// @return A ray of {starting_point, direction}
+  Acts::Ray3D makeRayFromTriplet(const Triplet triplet) const;
+
+  /// @brief Find a point (=the vertex) that has minimum chi^2 with respect to all rays fitted through the triplets
+  /// @param triplets Vector of all valid triplets
+  /// @return Position {x,y,z} of the vertex
+  Acts::Vector3 findClosestPointFromRays(
```
andiwand, line 181: static?

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+  if (std::isnan(cfg.removeFraction)) {
+    ACTS_ERROR("value of removeFraction was not initialized");
+  }
+  if (cfg.removeFraction < 0. || cfg.removeFraction >= 1.) {
+    ACTS_ERROR("value of removeFraction is "
+               << cfg.removeFraction << ", allowed values are between 0 and 1");
+  }
+  if (std::isnan(cfg.minVtxShift)) {
+    ACTS_ERROR("value of minVtxShift was not initialized");
+  }
```
andiwand, line 100: I would argue that the `nan` checks can be removed as this is not the default value and it would most likely cause an FPE downstream anyways. also I think it is the users job to not give unreasonable configs

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+}
+
+template <typename spacepoint_t>
+Acts::Vector3 Acts::SeedVertexFinder<spacepoint_t>::findVertex(
+    const std::vector<spacepoint_t>& spacepoints) {
+  // sort spacepoints to different phi and z bins
+  std::vector<std::vector<std::vector<
+      std::vector<std::pair<spacepoint_t const*, Acts::ActsScalar>>>>>
+      sorted_spacepoints = sortSpacepoints(spacepoints);
```
andiwand, line 109: I see a few `snake_case` variable flying around here. please consider changing these to `camelCase` because of code guidelines

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+               << m_cfg.minimalizeWRT
+               << ", allowed values are \"planes\" or \"rays\" ");
+  }
+
+  return vtx;
+}
+
+template <typename spacepoint_t>
+std::vector<std::vector<
+    std::vector<std::vector<std::pair<spacepoint_t const*, Acts::ActsScalar>>>>>
```
andiwand, line 138: did not see east const in our codebase yet but also not sure if there is a guideline about it. anyways as long as it is consistent here I am fine with it

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+          middle_spacepoints.at(phislice).at(zslice).emplace_back(
+              (spacepoint_t const*)&sp, phi);
+        }
+      } else {
+        if (sp.r() < m_cfg.rMaxFar) {
+          far_spacepoints.at(phislice).at(zslice).emplace_back(
+              (spacepoint_t const*)&sp, phi);
+        }
+      }
+    }
```
andiwand, line 188: I wonder if that could move more to the left with proper `else if` cases

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+                      continue;
+                    }
+
+                    Acts::SeedVertexFinder<spacepoint_t>::Triplet tr(
+                        *near_sp.first, *middle_sp.first, *far_sp.first);
+
+                    if (isTripletValid(tr)) {
+                      triplets.push_back(tr);
+                    }
+                  }
```
andiwand, line 366: this look is kind of deep :fearful: I guess this is how the algorithm works... maybe lambdas could help out here but not sure. at the very least I would comment the `}` closing braces with what was iterated

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+
+  Acts::ActsScalar tanTheta =
+      std::sqrt(direction[0] * direction[0] + direction[1] * direction[1]) /
+      direction[2];
+  if (std::fabs(tanTheta) < std::tan(m_cfg.minTheta)) {
+    return false;
+  }
+
+  // norm to z-axis and to the ray
+  Acts::Vector3 norm{-1. * direction[1], 1. * direction[0], 0};
```
andiwand, line 422: is this normalized? if not maybe we should normalize it and then remove the `norm.dot(norm)` below

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+    const Acts::SeedVertexFinder<spacepoint_t>::Triplet triplet) const {
+  Acts::Vector3 a{triplet.a.x(), triplet.a.y(), triplet.a.z()};
+  Acts::Vector3 b{triplet.b.x(), triplet.b.y(), triplet.b.z()};
+  Acts::Vector3 c{triplet.c.x(), triplet.c.y(), triplet.c.z()};
+
+  Acts::Vector3 ba = b - a, ca = c - a;
+
+  Acts::Vector3 abg = ba.cross(ca);
+  // vector (alpha,beta,gamma) normalized to unity for convenience
+  abg /= std::sqrt(abg.dot(abg));
```
andiwand, line 458:
```suggestion
  Acts::Vector3 abg = ba.cross(ca).normalize();
```

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+  }
+
+  // elements of the linear equations to solve
+  Acts::SymMatrix3 A = Acts::SymMatrix3::Zero();
+  Acts::Vector3 B = Acts::Vector3::Zero();
+  for (const auto& triplet : triplets_with_planes) {
+    const auto& abg = triplet.first.first;
+    const auto& delta = triplet.first.second;
+
+    A += 2. * (abg * abg.transpose());
```
andiwand, line 497: is `abg` a vector? if so consider using `squaredNorm`

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+      unsigned int threshold = (unsigned int)(triplets_with_planes.size() *
+                                              (1. - m_cfg.removeFraction));
+
+      for (unsigned int tr = threshold + 1; tr < triplets_with_planes.size();
+           ++tr) {
+        const auto& abg = triplets_with_planes.at(tr).first.first;
+        const auto& delta = triplets_with_planes.at(tr).first.second;
+
+        // remove this triplet from A and B
+        A -= 2. * (abg * abg.transpose());
```
andiwand, line 538: `squaredNorm`?

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+    }
+
+    if (iter != m_cfg.maxIterations) {
+      // is not the last iteration
+      vtx_prev = vtx;
+
+      for (auto& triplet : triplets_with_planes) {
+        const auto& abg = triplet.first.first;
+        const auto& delta = triplet.first.second;
+        Acts::ActsScalar distance = fabs(abg.dot(vtx) + delta);
```
andiwand, line 519:
```suggestion
        Acts::ActsScalar distance = std::abs(abg.dot(vtx) + delta);
```

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+                  return lhs.second < rhs.second;
+                });
+
+      unsigned int threshold = (unsigned int)(triplets_with_planes.size() *
+                                              (1. - m_cfg.removeFraction));
+
+      for (unsigned int tr = threshold + 1; tr < triplets_with_planes.size();
+           ++tr) {
+        const auto& abg = triplets_with_planes.at(tr).first.first;
+        const auto& delta = triplets_with_planes.at(tr).first.second;
```
andiwand, line 535:
```suggestion
        const auto& abg = triplets_with_planes[tr].first.first;
        const auto& delta = triplets_with_planes[tr].first.second;
```

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+
+  // elements of the linear equations to solve
+  Acts::SymMatrix3 A = Acts::SymMatrix3::Identity() * 2. * triplets.size();
+  Acts::Vector3 B = Acts::Vector3::Zero();
+  for (const auto& triplet : triplets_with_rays) {
+    // use ray saved from earlier
+    const auto& start_point = triplet.first.first;
+    const auto& direction = triplet.first.second;
+
+    A -= 2. * (direction * direction.transpose());
```
andiwand, line 602: `squaredNorm`?

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+                  return lhs.second < rhs.second;
+                });
+
+      unsigned int threshold = (unsigned int)(triplets_with_rays.size() *
+                                              (1. - m_cfg.removeFraction));
+
+      for (unsigned int tr = threshold + 1; tr < triplets_with_rays.size();
+           ++tr) {
+        const auto& start_point = triplets_with_rays.at(tr).first.first;
+        const auto& direction = triplets_with_rays.at(tr).first.second;
```
andiwand, line 643:
```suggestion
        const auto& start_point = triplets_with_rays[tr].first.first;
        const auto& direction = triplets_with_rays[tr].first.second;
```

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+        triplet.second = distance;
+      }
+
+      std::sort(triplets_with_rays.begin(), triplets_with_rays.end(),
+                [](const auto& lhs, const auto& rhs) {
+                  return lhs.second < rhs.second;
+                });
+
+      unsigned int threshold = (unsigned int)(triplets_with_rays.size() *
+                                              (1. - m_cfg.removeFraction));
```
andiwand, line 638: this section looks duplicated to the at 524 above

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+
+      for (unsigned int tr = threshold + 1; tr < triplets_with_rays.size();
+           ++tr) {
+        const auto& start_point = triplets_with_rays.at(tr).first.first;
+        const auto& direction = triplets_with_rays.at(tr).first.second;
+
+        // remove this triplet from A and B
+        A += 2. * (direction * direction.transpose());
```
andiwand, line 646: `squaredNorm`?

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
```diff
@@ -0,0 +1,656 @@
...
+      std::vector<std::pair<spacepoint_t const*, Acts::ActsScalar>>>>>
+      sorted_spacepoints = sortSpacepoints(spacepoints);
+
+  // find triplets
+  std::vector<Acts::SeedVertexFinder<spacepoint_t>::Triplet> triplets =
+      findTriplets(sorted_spacepoints);
+
+  // if no valid triplets found
+  if (triplets.empty()) {
+    return {};
```
andiwand, line 117: should we rather return an optional then or `Acts::Result`?

Examples/Io/Root/include/ActsExamples/Io/Root/RootSeedVertexPerformanceWriter.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+namespace ActsExamples {
+
+/// @class RootSeedVertexPerformanceWriter
+///
+/// Writes out the number of reconstructed seed vertices along with
+/// the number of truth vertices in detector acceptance.
+/// Additionally it matches the reco vertices to their truth vertices
+/// and write out the difference in x, y, and z position.
+class RootSeedVertexPerformanceWriter final
```
andiwand, line 29: any chance to use the existing vertex performance writer?

andiwand (commented):
thank you for your contribution @pbalek !

I did not check the algorithm but left a few comments concerning the code

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
pbalek: Hi @andiwand, yes, `abg` is a vector but the desired output is a matrix. I believe `squaredNorm` would return a scalar.

Cheers, Petr

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
pbalek: Hi @andiwand, same as above, we'd like to make a matrix.

Cheers, Petr

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
pbalek: Hi @andiwand, same as above, we'd like to make a matrix.

Cheers, Petr

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
pbalek: Hi @andiwand, same as above, we'd like to make a matrix.

Cheers, Petr

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+/// seeds
+/// 4. Returns the point position as the vertex
+
+template <typename spacepoint_t>
+class SeedVertexFinder {
+ public:
+  /// Configuration struct
+  struct Config {
+    // maximum deviation in phi between the near and middle spacepoints or
+    // middle and far spacepoints
```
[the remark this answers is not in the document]
pbalek: Hi @andiwand, I've added the third slash wherever applicable.

Cheers, Petr

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+  SeedVertexFinder(const Config& cfg,
+                   std::unique_ptr<const Logger> lgr =
+                       getDefaultLogger("SeedVertexFinder", Logging::DEBUG));
+
+  /// @brief Destructor
+  ~SeedVertexFinder() = default;
+
+  /// @brief Finds the vertex based on the provided spacepoints
+  /// @param spacepoints Vector of the input spacepoints; they do not need to be sorted anyhow
+  /// @return Position of the vertex along z-axis
```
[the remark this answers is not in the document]
pbalek: That was a relict from the previous version, should be just a position of the vertex.

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+  /// @param spacepoints Vector of the input spacepoints;
+  /// @return Vector of vectors for each set of spacepoints
+  std::vector<std::vector<std::vector<
+      std::vector<std::pair<spacepoint_t const*, Acts::ActsScalar>>>>>
+  sortSpacepoints(const std::vector<spacepoint_t>& spacepoints) const;
+
+  /// @brief
+  /// @param a
+  /// @param b
+  /// @return
```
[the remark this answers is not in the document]
pbalek: Removed, it wasn't used anywhere in the end.

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+  /// @return True if the deviations and fitted ray are within the configured ranges
+  ///         If "minimalizeWRT"=="rays", then the fitted ray is also saved to
+  ///         the triplet for later
+  bool isTripletValid(Triplet& triplet);
+
+  /// @brief Calculates equation of the plane (alpha*x + beta*y + gamma*z + delta = 0), given the three points
+  /// @param triplet A single triplet (with 3 spacepoints)
+  /// @return A pair of {{alpha,beta,gamma},delta}
+  std::pair<Acts::Vector3, Acts::ActsScalar> makePlaneFromTriplet(
+      const Triplet triplet) const;
```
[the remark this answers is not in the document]
pbalek: Fixed, also for `makeRayFromTriplet`.

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+
+    // number of phi slices, at least 3 to avoid duplicities
+    // it should be less than 2*pi/maxPhideviation in order not to loop over
+    // triplets that will be rejected by maxPhideviation anyway
+    int numPhiSlices = 60;
+    // use only a fraction of available phi slices to speed up calculations;
+    float useFracPhiSlices = 0.5;
+
+    // number of z slices
+    int numZSlices = 150;
```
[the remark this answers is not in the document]
pbalek: No, we don't depend in signed `int`. In fact, unsigned integer makes more sense.

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
pbalek: Hi @andiwand, I've deleted those `std::isnan`. I copied it from `SeedFinder` but I see now that it's not widely used.
For checks of reasonable values, I would prefer to keep them. It's an easy and fast way to let a user know that something went wrong.

Cheers, Petr

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
pbalek: Hi @andiwand, I hope I changed everything to the `camelCase`.

Cheers, Petr

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
pbalek: I changed it to `else if` when possible. Also note that logic above leaves "holes", as spacepoints e.g. between `rMaxMiddle` and `rMinFar` are not assigned anywhere.

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+    // chi^2 minimalization will happen with respect to "planes" or "rays"
+    std::string minimalizeWRT = "planes";
+
+    // maximum number of iterations when discarding triplets with the largest
+    // chi^2
+    int maxIterations = 20;
+    // each iteration, discard this fraction of triplets with the largest chi^2
+    float removeFraction = 0.10;
+    // if the vertex estimation moves less than this, stop iterations
+    float minVtxShift = 0.3f * Acts::UnitConstants::mm;
```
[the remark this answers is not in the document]
asalzburger: I would be very much in favour to use `Acts::ActsScalar` throughout - explicitly using `float` should only be - in limited use cases a storage decision.

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+  ~SeedVertexFinder() = default;
+
+  /// @brief Finds the vertex based on the provided spacepoints
+  /// @param spacepoints Vector of the input spacepoints; they do not need to be sorted anyhow
+  /// @return Position of the vertex along z-axis
+  Acts::Vector3 findVertex(const std::vector<spacepoint_t>& spacepoints);
+
+ private:
+  /// @brief Struct to store spacepoint combinations from near, middle, and far parts of the detector
+  struct Triplet {
```
asalzburger, line 116: Is there any way we can harmonise this with, e.g. the Triplet definition from seeding?

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+  /// @param triplet A single triplet to be validated
+  /// @return True if the deviations and fitted ray are within the configured ranges
+  ///         If "minimalizeWRT"=="rays", then the fitted ray is also saved to
+  ///         the triplet for later
+  bool isTripletValid(Triplet& triplet);
+
+  /// @brief Calculates equation of the plane (alpha*x + beta*y + gamma*z + delta = 0), given the three points
+  /// @param triplet A single triplet (with 3 spacepoints)
+  /// @return A pair of {{alpha,beta,gamma},delta}
+  std::pair<Acts::Vector3, Acts::ActsScalar> makePlaneFromTriplet(
```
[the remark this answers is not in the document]
asalzburger: Up!

Examples/Algorithms/Vertexing/include/ActsExamples/Vertexing/SeedVertexFinderAlgorithm.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+#pragma once
+
+#include "ActsExamples/EventData/SimSpacePoint.hpp"
+#include "ActsExamples/Framework/DataHandle.hpp"
+#include "ActsExamples/Framework/IAlgorithm.hpp"
+
+#include <string>
+
+namespace ActsExamples {
```
asalzburger, line 17: Same comment: I think we should make it explicit that it does single vertex finding.

asalzburger (changes requested):
Can we also add UnitTests for this, please?

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
pbalek: Hi @andiwand, we can't normalize it because `zdist` needs its square and it relies on the fact that `norm` is a cross product of `direction` and `(0,0,1)`, while both `direction` and `(0,0,1)` are normalized.
On the other hand, it seems beneficial to save size of this vector and use it instead of `norm.dot(norm)`, so I modified it in that direction.

Cheers, Petr

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+  struct Triplet {
+    Triplet(const spacepoint_t& aa, const spacepoint_t& bb,
+            const spacepoint_t& cc)
+        : a(aa),
+          b(bb),
+          c(cc),
+          ray(Acts::Vector3::Zero(), Acts::Vector3::Zero()) {}
+    const spacepoint_t &a, &b, &c;
+
+    Acts::Ray3D ray;
```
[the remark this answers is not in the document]
pbalek: Hi @andiwand, it is used to store fit with a straight line through the triplet, so we don't need to do it for the second time. I've added a comment.

Cheers, Petr

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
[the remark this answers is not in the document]
pbalek: Hi @andiwand and @asalzburger,
what is the motivation to make it `static`? The method is used only within this class and I don't think it would be useful elsewhere. I get the motivation to make them `const` (which I overlooked for some of them) but I don't understand what we'll gain with `static` methods?
Also, what do you mean by "Up!"? I sorted the methods by the order they are used. Thus, first is sorting spacepoints, then finding triplets, then finding the vertex.

Thank you, Petr

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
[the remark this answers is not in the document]
pbalek: Hi @andiwand and @asalzburger,
I changed all `float`s to `Acts::ActsScalar`s. I originally used `float` as I had some problem with compilation but now, I don't see any problem...

Cheers, Petr

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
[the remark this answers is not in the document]
andiwand: hey @pbalek ! this method depends on the state of the algorithm i.e. the config, right? I think this is what I later discovered during review.

In general I would say that all methods that do not depend on the object state should be static to explicitly tell this fact.

I think `Up!` just meant :+1: to my suggestion (which does not seem to apply anyways) :slightly_smiling_face:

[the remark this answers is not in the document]
pbalek: Hi @andiwand,
> In general I would say that all methods that do not depend on the object state should be static to explicitly tell this fact.

Ah, this makes sense! Thank you, I'll change them where applicable.

Cheers, Petr

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
pbalek: This depends on `m_cfg.minVtxShift` and `m_cfg.maxIterations`, so it can't be static. It's `const` though.

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+  /// @brief Calculates equation of the plane (alpha*x + beta*y + gamma*z + delta = 0), given the three points
+  /// @param triplet A single triplet (with 3 spacepoints)
+  /// @return A pair of {{alpha,beta,gamma},delta}
+  std::pair<Acts::Vector3, Acts::ActsScalar> makePlaneFromTriplet(
+      const Triplet triplet) const;
+
+  /// @brief Find a point (=the vertex) that has minimum chi^2 with respect to all planes defined by the triplets
+  /// @param triplets Vector of all valid triplets
+  /// @return Position {x,y,z} of the vertex
+  Acts::Vector3 findClosestPointFromPlanes(
```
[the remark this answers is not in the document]
pbalek: This depends on `m_cfg.minVtxShift` and `m_cfg.maxIterations`, so it can't be static. It's const though.

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+      const std::vector<std::vector<std::vector<
+          std::vector<std::pair<spacepoint_t const*, Acts::ActsScalar>>>>>&
+          sorted_spacepoints);
+
+  /// @brief Validate the triplet based on "maxXYdeviation", "maxXYZdeviation", "maxZPosition", and "maxRPosition"
+  /// @param triplet A single triplet to be validated
+  /// @return True if the deviations and fitted ray are within the configured ranges
+  ///         If "minimalizeWRT"=="rays", then the fitted ray is also saved to
+  ///         the triplet for later
+  bool isTripletValid(Triplet& triplet);
```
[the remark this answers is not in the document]
pbalek: It depends on `m_cfg` so it can't be `static`. I made it `const` and rename to `tripletValidationAndUpdate`.

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+
+  /// Configuration instance
+  Config m_cfg;
+
+  /// @brief Sorts spacepoints into a separate vectors for near, middle, and far spacepoints; and for each slice of phi
+  /// @param spacepoints Vector of the input spacepoints;
+  /// @return Vector of vectors for each set of spacepoints
+  std::vector<std::vector<std::vector<
+      std::vector<std::pair<spacepoint_t const*, Acts::ActsScalar>>>>>
+  sortSpacepoints(const std::vector<spacepoint_t>& spacepoints) const;
```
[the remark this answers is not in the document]
pbalek: I made a struct for that, so it's the same type, only a little bit hidden.
Also, I changed the first `vector` to `arra< ... ,3>`.

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
pbalek: I didn't want to copy the whole spacepoints but at the same time, I needed to put them into several vectors. Pointer to the constant worked well. There's no way the spacepoints themselves can be modified (and they shouldn't).
Also, it allows to modify the vectors later. Sorting the spacepoints according their phi might speed up the creation of the triplets later. I decided not include it in the first version as it will require more testing, however we are prepared for that.

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
pbalek: I don't think we can get rid of the nesting. 
Theoretically, we could define some simple lambda functions but each lambda would be used maybe twice, so it's a very little benefit for lower readability.
Also, some pieces of code are _almost_ the same but they have e.g. `+1` here and `-1` there; have `true` here and `false` there and so on. If we were to define such function to accommodate that, it would be as complex as the actual expression which defies the purpose.
Thus I think it's better to leave it as it is now.

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
pbalek: Hi @andiwand,
to be honest, the whole function `findClosestPointFromRays` looks very similar to `findClosestPointFromPlanes` and it _almost_ looks they could be a single function. However, there are many small difference, so the single function would need to be very complex. The same goes with the iteration loops. They are very similar, but not quite.
The lines that you mentioned are indeed duplicates but I don't think it would be beneficial to define another function as it would do two different and independent things: sort a vector and then get an index based on its size and `removeFraction`.

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
```diff
@@ -0,0 +1,193 @@
...
+#include <algorithm>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace Acts {
+
+/// @class SeedVertexFinder
+///
+/// @brief Implements the vertex finder based on the track seeds
```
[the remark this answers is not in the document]
pbalek: Hi @asalzburger,
do you have some suggestion for a name? For example `SingleSeedVertexFinder` seems quite long and `SingleVertexFinder` sounds as a standard algorithm...

Thank you, Petr

andiwand:
let us know when you need a second round of review @pbalek :slightly_smiling_face:

Core/include/Acts/Vertexing/SeedVertexFinder.ipp
pbalek: Hi @andiwand, I've changed it to `Acts::Result` and then check later if it's fine.

Examples/Io/Root/include/ActsExamples/Io/Root/RootSeedVertexPerformanceWriter.hpp
pbalek: Hi @andiwand and @asalzburger, I've change it to `VertexPerformanceWriter`. I needed to modify `VertexPerformanceWriter` a little bit so it can handle vertices without any reconstructed tracks.
(Default behaviour of `VertexPerformanceWriter` is unchanged.)

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
pbalek: Hi @asalzburger, unfortunatelly neither of the triplet definitions works well here.
Moreover, we might be able to get rid of it completely but I would leave it for the next iteration.

Core/include/Acts/Vertexing/SeedVertexFinder.hpp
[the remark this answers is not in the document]
pbalek: Hi @asalzburger, I renamed it to `SingleSeedVertexFinder`.

Examples/Algorithms/Vertexing/include/ActsExamples/Vertexing/SeedVertexFinderAlgorithm.hpp
pbalek: Renamed to `SingleSeedVertexFinderAlgorithm`.

pbalek:
Hi @asalzburger,
> Can we also add UnitTests for this, please?

I've just added some unit tests.

Cheers, Petr

pbalek:
Hi @andiwand,
> let us know when you need a second round of review @pbalek 🙂

I re-requested review from you and @asalzburger.

I still have here "1 change requested" by @asalzburger, but there are no conversations left to resolve. How to address it?

Thank you,
Petr

andiwand (approved):
@asalzburger can you have a second look or approve?

asalzburger (approved):
Looks good now

pbalek:
Hi @andiwand, I think this can be merged, or is there something else I overlooked?
I can't add "automerge" label, can I?

Thank you, Petr

PR #2039, opened 2023-04-18
fix: ParticleID collision in Geant4

benjaminhuth:
This attempts to fix particle collisions in Geant4. These occur I think when the bits in the `ActsFatras::Barcode` are not enough to capture all results. This way it could happen that multiple particles where associated with a single ID:

image

I think the correct fix for this to check for collision and ignoring particles that collide. Furthermore, the barcode generation is improved, so that collisions should happen less frequently.

Since this happens quite frequently with default settings for Geant4+Pythia8(ttbar), I think all so far generated events using this configuration are suffering from this problem


andiwand:
tagging @Corentin-Allaire as this might impact the ambi performance eval?

andiwand (approved):
looks good from my side

I wonder if we can optimize the barcode or the utilization of the barcode in the G4 translation so we avoid overflows for ttbar

Corentin-Allaire:
> tagging @Corentin-Allaire as this might impact the ambi performance eval?

It definitely could affect it, if we have multiple track with the same id then one of them will be treated as a duplicate

benjaminhuth:
Actually there are quite a lot of collisions:

```
20:47:59    Geant4Simula   WARNING   2718782 particles missing in the output data because of particle ID collissions
```

Corentin-Allaire:
> Actually there are quite a lot of collisions in one event:
> 
> ```
> 20:47:59    Geant4Simula   WARNING   2718782 particles missing in the output data because of particle ID collissions
> ```

That number seems a bit too high, for how many event is that ?

andiwand:
> > Actually there are quite a lot of collisions in one event:
> > ```
> > 20:47:59    Geant4Simula   WARNING   2718782 particles missing in the output data because of particle ID collissions
> > ```
> 
> That number seems a bit too high, for how many event is that ?

I dont actually know but it might be plausible considering all the secondary particles? I once use G4 with 100 muon event and got like 10k secondaries

benjaminhuth:
> 
> That number seems a bit too high, for how many event is that ?

Its for one event, but without post selection. And probably you need to divide this number by two, because right now it counts initial and final particles.

Corentin-Allaire:
> > That number seems a bit too high, for how many event is that ?
> 
> Its for one event, but without post selection. And probably you need to divide this number by two, because right now it counts initial and final particles.

I just run the job with the default selector and I see 0 overlap so this probably doesn't change much in the end ?

asalzburger:
> looks good from my side
> 
> I wonder if we can optimize the barcode or the utilization of the barcode in the G4 translation so we avoid overflows for ttbar

Sure, if we can change the barcode behaviour to be safe against this, let's do it!

asalzburger:
> Since this happens quite frequently with default settings for Geant4+Pythia8(ttbar), I think all so far generated events using this configuration are suffering from this problem

Can you specify what this means? Which particles are disregarded?
And how does this free enough space in the barcode bits?

Examples/Algorithms/Geant4/include/ActsExamples/Geant4/EventStoreRegistry.hpp
```diff
@@ -34,24 +37,43 @@ class EventStoreRegistry {
...
+    /// Count particle ID collisions
+    std::size_t particleIdCollisionsInitial = 0;
+    std::size_t particleIdCollisionsFinal = 0;
+
+    /// Store subparticle count for {primVertex, secVertex, part, gen}
+    /// This is done using a pseudo-barcode that contains all fields but not the
+    /// subparticle counter. This can be used as key in a map to store the
+    /// subparticle information
+    using BarcodeWithoutSubparticle =
+        Acts::MultiIndex<uint64_t, 16, 16, 16, 16>;
```
andiwand, line 75: shouldn't we use the same bits as in the original barcode?

Examples/Algorithms/Geant4/src/ParticleTrackingAction.cpp
```diff
@@ -83,3 +96,34 @@ ActsExamples::SimParticle ActsExamples::ParticleTrackingAction::convert(
...
+    return std::nullopt;
+  }
+
+  auto pid = ed.trackIdMapping.at(parentId).makeDescendant();
+
+  auto key = EventStoreRegistry::State::BarcodeWithoutSubparticle::Zeros();
+  key.set(0, pid.vertexPrimary())
+      .set(1, pid.vertexSecondary())
+      .set(2, pid.particle())
+      .set(3, pid.generation());
```
andiwand, line 125: can you explain what this key is?

andiwand (approved):
thanks for turning this around so quickly :+1: 

looks good - just two questions

Examples/Algorithms/Geant4/include/ActsExamples/Geant4/EventStoreRegistry.hpp
benjaminhuth: but they must sum up to 64... and as long there is enough space it should be ok

andiwand: makes sense - my concern was that we could overflow by converting from this special bar code to the other

benjaminhuth: we do never convert them into each other, see below

Examples/Algorithms/Geant4/src/ParticleTrackingAction.cpp
benjaminhuth: The idea is that we maintain a map, that does hold the current subparticle count for a spezific combination of primary, secondary, particle and generation. This is most easily done by using another `Acts::MultiIndex` object with only four fields as key.

benjaminhuth:
> 
> Can you specify what this means? Which particles are disregarded? And how does this free enough space in the barcode bits?

Finally, I have some plots @asalzburger 

image

image

asalzburger:
@benjaminhuth  - can you fix the format, please ?

benjaminhuth:
Yeah and I need to do some clean-up still

benjaminhuth:
@asalzburger The CI *should* run through now, could you reapprove?

Corentin-Allaire (approved):

PR #2044, opened 2023-04-20
refactor: resurrect detector root volumes

andiwand:
with https://github.com/acts-project/acts/pull/2027 I removed the ability to have multiple root volumes. this functionality is brought back in this PR because it is necessary for the KDT detector builder


Core/include/Acts/Navigation/DetectorVolumeFinders.hpp
```diff
@@ -36,11 +36,13 @@ struct RootVolumeFinder : public INavigationDelegate {
           "DetectorVolumeFinders: no detector set to navigation state.");
     }
 
-    const auto& volume = nState.currentDetector->rootVolume();
-    if (volume->inside(gctx, nState.position)) {
-      nState.currentVolume = volume;
-      volume->detectorVolumeUpdator()(gctx, nState);
-      return;
+    const auto& volumes = nState.currentDetector->rootVolumes();
+    for (const auto v : volumes) {
```
asalzburger, line 40: Shouldn't that be `const auto&` (it's only a pointer in this case, but I think `clang-tidy` might be worried otherwise.

asalzburger (approved):
I like this, we have now the possibility to create one or many root volumes, bravo!

Core/include/Acts/Navigation/DetectorVolumeFinders.hpp
andiwand: I think this is okay because `auto` will resolve to a pointer and `&` would only bring another indirection

asalzburger:
You need to add the mockup detector test as well ..

andiwand:
massaging clang tidy right now otherwise this should be done I think

PR #2045, opened 2023-04-20
ci: Increase use of GitLab CI runners

paulgessinger:
WIP for testing.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -26,98 +26,98 @@ env:
 #      found a way to have Github actions not fail regularly with this job
 #      all unit tests should be reactivated.
 jobs:
-  build_debug:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


paulgessinger:
Ok, so it seems to run. I also instruct the job to install `pytest-xdist` and let it run the test on all available cores (8 on our VMs), which cuts the test execution time from 17m to 9m.

I'm inclined to let this run in addition to the current `linux_examples_test` job for the time being, to see if it behaves correctly, and then think about switching the GitHub version off.

andiwand (approved):
looks scary but makes sense to me. lets merge it and see how it behaves

benjaminhuth:
@paulgessinger This still seems to fail also if I rerun CI Bridge / linux_test_examples

paulgessinger:
Oh yeah, it's the same problem I just fixed in #2080. I'll add a commit to fix it here.

benjaminhuth (approved):

PR #2047, opened 2023-04-20
refactor: navigator volume material handling

andiwand:
split off changes from my [calorimeter demo](https://github.com/acts-project/acts/compare/main...andiwand:tmp-calorimeter-demo?expand=1). this adds a shortcut for getting the current volume material from the navigator. since the new geometry has a different volume class we cannot reach the material in a generic way out of the box


asalzburger (approved):
Happy with that.

PR #2048, opened 2023-04-20
refactor: geant4 converters

andiwand:
split off changes from my [calorimeter demo](https://github.com/acts-project/acts/compare/main...andiwand:tmp-calorimeter-demo?expand=1). this changes most of the shared pointers to unique pointers in the converter because we use unique pointers downstream sometimes. and unique pointers can be upgraded to shared pointers directly. it also includes a generalization of the material conversion and transform conversion

**update and summary**

- switched from `unique_ptr` consistently to `shared_ptr` after discussion
- added `Geant4VolumeConverter`
- added transform translation for G4 phyiscal volumes
- added generic material translation


Plugins/Geant4/include/Acts/Plugins/Geant4/Geant4Converters.hpp
```diff
@@ -51,10 +51,17 @@ struct Geant4AlgebraConverter {
 
   /// @brief  Translate a geometry transform
   ///
-  /// @param g4Trf theGeant4 transform object
+  /// @param g4Trf the Geant4 transform object
   ///
   /// @return a Acts transform
   Transform3 transform(const G4Transform3D& g4Trf);
+
+  /// @brief  Translate a geometry transform
```
asalzburger, line 59: Probably name the source (from a physical volume)

Plugins/Geant4/include/Acts/Plugins/Geant4/Geant4Converters.hpp
```diff
@@ -84,38 +91,38 @@ struct Geant4ShapeConverter {
   /// @param g4Tubs a Geant4 tube shape
   ///
   /// @return an Acts Cylinder bounds object, and thickness
-  std::tuple<std::shared_ptr<CylinderBounds>, ActsScalar> cylinderBounds(
+  std::tuple<std::unique_ptr<CylinderBounds>, ActsScalar> cylinderBounds(
```
asalzburger, line 94: My thinking here was to create what the surface constructor needs ... what's the motivation to change to `unqiue_ptr` here?

Plugins/Geant4/include/Acts/Plugins/Geant4/Geant4Converters.hpp
```diff
@@ -146,17 +153,32 @@ struct Geant4PhysicalVolumeConverter {
...
 struct Geant4MaterialConverter {
+  Material material(const G4Material& g4Material, ActsScalar compression = 1);
+
   /// @brief Convert a Geant4 material to a surface material description
   ///
   /// @param g4Material the geant4 material descrition
   /// @param original the original thickness
   /// @param compressed the compressed thickness
   ///
-  std::shared_ptr<HomogeneousSurfaceMaterial> surfaceMaterial(
```
asalzburger, line 158: same here, there ...

asalzburger (approved):
I am in principle ok with the PR, just wanted to cross-check on a few points.

Plugins/Geant4/include/Acts/Plugins/Geant4/Geant4Converters.hpp
andiwand: Maybe I went a bit overboard with this. My thinking was if we ever need a unique ptr with surface bounds it will not be possible with this interface in a straight forward way. For surface bounds it looks like we never do this anyways but for example `DetectorVolume` will take a volume bounds as a unique prt. So I think it makes sense to stay consistent

asalzburger: I see you point, the fact that `Surface` objects have shared pointers as bounds compared to the `Volume` objects that container unique bounds, is the fact that indeed when you have a.g. a straw detector with 100 thousands of straws they all could have the same bounds ...

andiwand: sure I can change that back. I think in principle factories should be using `std::unique_ptr` as a return value because you can always move them to be `std::shared_ptr` but not the other way around. if we only use `std::shared_ptr` of these it is fair to build them in that way in the first place

andiwand:
@asalzburger implemented your suggested changes

asalzburger (approved):
Good to go, questions have been addressed.

PR #2049, opened 2023-04-20
fix: enable G4 secondary hits

andiwand:
apparently G4 secondary hits were not recorded so far :(

**summary**
- enable G4 secondary hit recording by default
- improve subprocess error output in python tests
- reduce full chain pythia pileup to 50 because of OOM in CI


Corentin-Allaire (approved):
That is quite unfortunate... But great find !

benjaminhuth:
That explains also why the particleID mismatches had no effect so far on the truth hits

benjaminhuth:
@andiwand this is not configurable at the moment. should we make it configurable?

Corentin-Allaire:
If anybody is wondering what the effect of this is on the Ambiguity Solver, it is not great but it is not massive either. We lose 1% of efficiency from 98.7% to 97.6%. The fake rate is the main difference going from 0.08% (3 fake per event) to 0.41% (14 fake per event).

This doesn't include benjamin change to the ID collision yet (https://github.com/acts-project/acts/pull/2039)

benjaminhuth:
Actually I would propose to merge #2052 instead of this PR, I tried to streamline this a bit.

asalzburger:
> Actually I would propose to merge #2052 instead of this PR, I tried to streamline this a bit.

Can you guys decide/harmonize how to move forward on this, I don't know which PR is more suitable for fixing this?

andiwand:
this is what I get locally

```
14:07:13    Propagator     ERROR     Propagation reached the step count limit of 100000000 (did 100000000 steps)
14:07:13    CKF            ERROR     Propapation failed: PropagatorError:3 Propagation reached the configured maximum number of steps with the initial parameters: 
-2.74156
 33.1783
-1.69583
 2.27469
 1.96932
-116.462
14:07:13    TrackFinding   WARNING   Track finding failed for seed 23709 with errorPropagatorError:3
```

Which seems plausible that we hit this threshold with much more hits. I tried to increase the threshold but even with x1000 it will still crash which I cannot explain.

For now I just bumped the log level threshold in the G4 pythia ODD test to let the CI pass

benjaminhuth:
Hmm I agree, I think these failures are to be expected with a lot more hits and are also okay for an examples script in my opinion...
Actually out of interest, can you quantify how many more hits we get?

benjaminhuth:
The only thing is we might want to restrict the allowed errors to exactely this error... so that not silently other modules break.
Is this achievable?

andiwand:
> Hmm I agree, I think these failures are to be expected with a lot more hits and are also okay for an examples script in my opinion... Actually out of interest, can you quantify how many more hits we get?

with secondaries I see 205293 hits. without 

> The only thing is we might want to restrict the allowed errors to exactely this error... so that not silently other modules break.

that is true but I do not see how we can achieve that. maybe we can disable the secondaries somehow and keep the warning threshold there? then run it a second time with threshold and expect errors. but sounds kinda wasteful

any ideas @paulgessinger ?

Corentin-Allaire:
One thing to mention concerning this PR. I tried it loclly and I get the following warning in the CKF perf writer (a LOT of them acctually) :
`RootTrajecto   WARNING   Truth particle with barcode 124|0|7|1|8=558446353911447560`

benjaminhuth:
> that is true but I do not see how we can achieve that. maybe we can disable the secondaries somehow and keep the warning threshold there? then run it a second time with threshold and expect errors. but sounds kinda wasteful

A stupid thing to do (?) would be to kind to filter the `stdout` with a smart regex and count the errors...

andiwand:
> One thing to mention concerning this PR. I tried it locally and I get the following warning in the CKF perf writer (a LOT of them acctually) : `RootTrajecto WARNING Truth particle with barcode 124|0|7|1|8=558446353911447560`

yes that is true but this was also the case before these changes no?

> A stupid thing to do (?) would be to kind of filter the `stdout` with a smart regex and count the errors...

sure I can try that

Corentin-Allaire:
> > One thing to mention concerning this PR. I tried it locally and I get the following warning in the CKF perf writer (a LOT of them acctually) : `RootTrajecto WARNING Truth particle with barcode 124|0|7|1|8=558446353911447560`
> 
> yes that is true but this was also the case before these changes no?
> 

For me thins only occur after switching the secondary to true, without them I don't see any Warning.

andiwand:
> For me thins only occur after switching the secondary to true, without them I don't see any Warning.

Oh you are right this is the thing I fixed a while ago. Could this be due to unresolved collisions @benjaminhuth ?

andiwand:
thinking about this it might be due to some particle selection that will kick out secondaries we reconstruct with CKF?

benjaminhuth:
> thinking about this it might be due to some particle selection that will kick out secondaries we reconstruct with CKF?

That sounds reasonable.

> Oh you are right this is the thing I fixed a while ago. Could this be due to unresolved collisions @benjaminhuth ?

Hmm not sure how it could be a collision. The warning is about hits of particles that are not found in the true particles, right? So about missing particle ids, not colliding ones, right? Or do I misunderstand the warning?

andiwand:
I tried to make the error threshold more robust as you suggested @benjaminhuth 

I also added prints for failed subprocesses do we can hopefully get more output from the CI and locally without running the actual script

andiwand:
now we hit the OOM I fear :( the only solution I see is to make the event smaller. any other ideas @benjaminhuth @Corentin-Allaire

benjaminhuth:
Do we hit OOM during Geant, or afterwards in the reco?

Corentin-Allaire:
Could we split the CI in 2 ? One with small event with secondary and for all the other CI jobs with G4 run then without the secondary ?

andiwand:
> Do we hit OOM during Geant, or afterwards in the reco?

I guess during CKF because the event got much more dense

> Could we split the CI in 2 ? One with small event with secondary and for all the other CI jobs with G4 run then without the secondary ?

sounds reasonable but I fear we would have to add a bunch of cli args to the full chain to achieve this and we wanted to avoid that cc @paulgessinger

benjaminhuth:
> I guess during CKF because the event got much more dense

But then couldn't we just add a postselection after the sim? cut particles with < 1 GeV or so?

PR #2052, opened 2023-04-21
refactor: Consistent Geant4 particle selection & enable secondary hits by default

benjaminhuth:
Allow for a more consistent configuration of which particles & hits the Geant4 sim should pick up or not (options: primaries, secondaries, charged, neutral)

I think it doesn't make sense to ignore e.g. neutral hits but keep these particles, as done currently.

Also enables secondary hits & particles by default (therefore overlapping with #2049).


benjaminhuth:
@andiwand @Corentin-Allaire

Corentin-Allaire:
I am not 100% sure this is a good idea, in particular for the neutral case I an imagine some cases where you would want to write the neutral particle (because they are somehow useful for some truth study) but you don't want to have neutral hits because they do not exit.

Whether we record neutral hit depend on detector type and nothing else. So the old level of customisation made more sense (but having a consistent default is for sure better)

benjaminhuth:
Hmm okay maybe you have a good point... but for primary/secondary it would make sense, wouldn't it?

And also for neutral particles: Recording less particles in the beginning means also less chance on particle-ID collisions (and also some performance improvements, but not measured it), so if I don't want to study them, I would like to disable them maybe...

Corentin-Allaire:
For the neutral particle removing them before G4 might also cause issue because they could create charged secondary. But after G4 we should remove them for sure. 

For the primary/secondary part, I see your point if we force them to be the same we ensure that people won't do anything stupid. On the other hand, since in my opinion the tracking perf only matter for the primary, I could see a reason to only record primary particles but you would still need to primary and secondary hits for realistic event.
What make me uncomfortable is that here we are tying together config for reco and truth object and this is never a good idea.

benjaminhuth:
> For the neutral particle removing them before G4 might also cause issue because they could create charged secondary. 

So it wouldn't mean Geant4 wouldn't simulate neutral particles. But we have one problem indeed I haven't thought of: in case of a charged secondary of a neutral parent particle we need the `ActsFatras::Barcode` of the neutral parent particle to generate the new barcode...

> On the other hand, since in my opinion the tracking perf only matter for the primary, I could see a reason to only record primary particles but you would still need to primary and secondary hits for realistic event.

But currently it's the other way around: We have all particles but not their hits.

But I see you point with tying the things together. So could this be a way to go forward:
- remove the option at all to filter on charged / neutral hits & particles (we also have the particle selector algorithm)
- make primary/secondary individual configurable for hits & particles

Corentin-Allaire:
> But currently it's the other way around: We have all particles but not their hits.
I agree this is a huge issue and it should be fixed as soon as possible.

I think I am missing something (probably because it is the end of the week :) ) why do you want to change the implementation compare to how it is right now ? Obviously we need to change it to write secondary by default and have the hits writing option configurable but I don't see a need/benefit of altering the code more than this ? (it might be simpler to discuss it on zoom if you want ?)

benjaminhuth:
Closing because for now just merging #2049 after discussion with @andiwand. Decide on this later maybe.

PR #2053, opened 2023-04-21
refactor: change cut order in getCompatibleDoublets

LuisFelipeCoelho:


Core/include/Acts/Seeding/SeedFinderUtils.ipp
```diff
@@ -147,6 +147,30 @@ inline void transformCoordinates(Acts::SpacePointData& spacePointData,
   }
 }
 
+inline std::array<float, 6> transformCoordinates(
+    const std::array<float, 4>& lineCircleVariables) {
+  auto& [deltaX, deltaY, cosPhiM, sinPhiM] = lineCircleVariables;
```
CarloVarni, line 152: why passing an array if then you immediately take the difference components? Why not requiring 4 floats as inputs to the function?

LuisFelipeCoelho:
@CarloVarni I tested the modular approach and it seems to give worse results, I am afraid this is the best option

CarloVarni (approved):

PR #2056, opened 2023-04-22
feat: Allow configurable particle killing

benjaminhuth:
Allows to kill a particle at when its position exceeds a configurable r or z position.

### DIFF
--- a/Examples/Algorithms/Geant4/include/ActsExamples/Geant4/SensitiveSteppingAction.hpp
+++ b/Examples/Algorithms/Geant4/include/ActsExamples/Geant4/SensitiveSteppingAction.hpp
@@ -31,6 +31,10 @@ class SensitiveSteppingAction : public G4UserSteppingAction {
     bool neutral = false;
     bool primary = true;
     bool secondary = false;
+
+    // When to kill a track
+    double maxAbsZ = std::numeric_limits<double>::infinity();
+    double maxR = std::numeric_limits<double>::infinity();
   };
 
   /// Construct the stepping action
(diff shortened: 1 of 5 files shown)


Examples/Algorithms/Geant4/include/ActsExamples/Geant4/SensitiveSteppingAction.hpp
```diff
@@ -31,6 +31,10 @@ class SensitiveSteppingAction : public G4UserSteppingAction {
     bool neutral = false;
     bool primary = true;
     bool secondary = false;
+
+    // When to kill a track
+    double maxAbsZ = std::numeric_limits<double>::infinity();
+    double maxR = std::numeric_limits<double>::infinity();
```
asalzburger, line 37: If you have a dedicated `ParticleKillAction` - why does the stepping action need to know about this?
Isn't the idea of actions that they can be done w/o communication with another?

Examples/Algorithms/Geant4/src/SensitiveSteppingAction.cpp
```diff
@@ -38,6 +38,16 @@ void ActsExamples::SensitiveSteppingAction::UserSteppingAction(
...
+  // Decide if we want to kill the track
+  const auto pos = convertLength * track->GetPosition();
+  if (std::abs(pos.z()) > m_cfg.maxAbsZ or
+      std::hypot(pos.x(), pos.y()) > m_cfg.maxR) {
+    ACTS_DEBUG("Kill track with internal track ID "
+               << track->GetTrackID() << " at (r,z) = "
+               << std::hypot(pos.x(), pos.y()) << ", " << pos.z());
+    track->SetTrackStatus(G4TrackStatus::fStopAndKill);
+  }
+
```
asalzburger, line 50: See my comment from before - I don't really understand why the sensitive stepping action should be doing that.

asalzburger (changes requested):
Naively I would think once can write a ParticleKillAction w/o interacting with the sensitive stepping action and let the first one do the entire job?

benjaminhuth:
Will be introduced in updated #2057

PR #2057, opened 2023-04-22
feat: particle killing 4 Geant4 and refactor user actions

benjaminhuth:
Out Geant4 algorithm allows to pass `std::vector`s of user actions. However, multiple actions per type seem not to be supported by Geant4. This changes this to allow only one user action per type (run, event, particle, step).

I'm not sure if this was possible in former Geant4 versions, there is also a comment in the Geant4 source code that indicates it is possible to add multiple actions per type. However, this is not the case.


asalzburger (approved):

asalzburger:
Makes totally sense, thanks, @benjaminhuth

asalzburger:
> Out Geant4 algorithm allows to pass `std::vector`s of user actions. However, multiple actions per type seem not to be supported by Geant4. This changes this to allow only one user action per type (run, event, particle, step).
> 
> I'm not sure if this was possible in former Geant4 versions, there is also a comment in the Geant4 source code that indicates it is possible to add multiple actions per type. However, this is not the case.

As of your question, I *think* this comes from the singleton usage/desing of Geant4 in the past. But as of the chances you did, they are all in line what Geant4 expects to do - and were just missing on our side.

asalzburger:
> > Out Geant4 algorithm allows to pass `std::vector`s of user actions. However, multiple actions per type seem not to be supported by Geant4. This changes this to allow only one user action per type (run, event, particle, step).
> > I'm not sure if this was possible in former Geant4 versions, there is also a comment in the Geant4 source code that indicates it is possible to add multiple actions per type. However, this is not the case.
> 
> As of your question, I _think_ this comes from the singleton usage/desing of Geant4 in the past. But as of the chances you did, they are all in line what Geant4 expects to do - and were just missing on our side.

Oh - I misread this - I will check, I was under the impression that you can attach different UserActions, just not the the same one twice, I will check that.

asalzburger (changes requested):

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
```diff
@@ -34,20 +34,6 @@
...
-/// @tparam actions_t the actions iterable list
-template <typename manager_t, typename actions_t>
-void setUserActions(manager_t& manager, actions_t& actions) {
-  for (const auto& action : actions) {
-    if (action != nullptr) {
-      manager.SetUserAction(action);
-    }
-  }
-}
-}  // namespace
```
andiwand, line 49: never seen that before :smile:

Examples/Algorithms/Geant4/include/ActsExamples/Geant4/ActsSteppingActionList.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <optional>
+#include <string>
+
+#include <G4UserSteppingAction.hh>
+
+namespace ActsExamples {
+
+/// Geant4 only allows one user action of each type. This simple wrapper
+/// dispatches multiple actions to Geant4.
+class ActsSteppingActionList : public G4UserSteppingAction {
```
andiwand, line 25: should we template this to allow listing other actions as well?

Examples/Algorithms/Geant4/include/ActsExamples/Geant4/ActsSteppingActionList.hpp
```diff
@@ -0,0 +1,45 @@
...
+ public:
+  struct Config {
+    std::vector<G4UserSteppingAction *> actions;
+  };
+
+  ActsSteppingActionList(const Config &cfg) : m_cfg(cfg) {}
```
andiwand, line 31: maybe the config is simple enough in this case to pass the vector directly

Examples/Python/src/Geant4Component.cpp
```diff
@@ -280,34 +293,6 @@ PYBIND11_MODULE(ActsPythonBindingsGeant4, mod) {
...
-    ACTS_PYTHON_STRUCT_END();
-
-    auto cch = py::class_<ChamberConfig>(ms, "ChamberConfig").def(py::init<>());
-    ACTS_PYTHON_STRUCT_BEGIN(cch, ChamberConfig);
-    ACTS_PYTHON_MEMBER(name);
-    ACTS_PYTHON_MEMBER(SensitiveNames);
-    ACTS_PYTHON_MEMBER(PassiveNames);
-    ACTS_PYTHON_STRUCT_END();
-  }
-
```
andiwand, line 310: I guess this should stay?

Examples/Run/Geant4/Common/src/Geant4Common.cpp
```diff
@@ -89,10 +94,12 @@ void setupGeant4Simulation(
   g4Cfg.detectorConstruction = detector.release();
 
   // Set the user actions
-  g4Cfg.runActions = std::move(runActions);
-  g4Cfg.eventActions = std::move(eventActions);
-  g4Cfg.trackingActions = std::move(trackingActions);
-  g4Cfg.steppingActions = std::move(steppingActions);
+  g4Cfg.runAction = runActions.empty() ? nullptr : runActions.front();
```
andiwand, line 97: would it make sense to use a generalized the vector of actions and use it here as well?

andiwand (commented):
nice catch - well done! left a few questions but looks good in general to me

Examples/Python/src/Geant4Component.cpp
benjaminhuth: oh no idea how this came in, sure!

Examples/Run/Geant4/Common/src/Geant4Common.cpp
benjaminhuth: as these are the examples binaries I don't want to touch more code then necessary to ensure that they compile actually and somehow maintain functionality

Examples/Run/Geant4/Common/src/Geant4Common.cpp
```diff
@@ -89,10 +94,12 @@ void setupGeant4Simulation(
...
+  g4Cfg.eventAction = eventActions.empty() ? nullptr : eventActions.front();
+  g4Cfg.trackingAction =
+      trackingActions.empty() ? nullptr : trackingActions.front();
+  g4Cfg.steppingAction =
+      steppingActions.empty() ? nullptr : steppingActions.front();
```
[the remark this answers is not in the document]
benjaminhuth: Hmm I thought we might wait until it is needed...

Examples/Algorithms/Geant4/src/ParticleKillAction.cpp
```diff
@@ -0,0 +1,38 @@
+// This file is part of the Acts project.
+//
+// Copyright (C) 2021 CERN for the benefit of the Acts project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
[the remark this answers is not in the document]
benjaminhuth: I'll look into that, seems like a good idea

Examples/Run/Geant4/Common/src/Geant4Common.cpp
andiwand: oh yeah makes sense

benjaminhuth:
@asalzburger I implemented the particle killing now with `Acts::Volume`, it feels much better now actually. Regarding other actions, I would like to wait until we need a similar list implementation if its ok for you. Is it good to go now from your perspective? (will need to update hashes, so formal approval is not yet required)

I also directly included the particle killing into `full_chain_odd.py`, so I also tag @andiwand also. Is this okay or should I remove it again? I would like to execute this at least once in the CI.

andiwand:
> I also directly included the particle killing into `full_chain_odd.py`, so I also tag @andiwand as well. Is this okay or should I remove it again? I would like to execute this at least once in the CI.

I wonder if we could have a more dedicated test for that. Or will the full chain actually break if you remove it?

I only worry that we might silently remove rouge particles which should not be there in the first place and we will catch the error at a different place later in time

andiwand (approved):

asalzburger (approved):

PR #2058, opened 2023-04-25
refactor: Refactor the measurement calibrator in the examples

gagnonlg:
I'm currently working on implementing a few different measurement/cluster calibration methods for a bona fide calibration study. The easiest way to run such a study would be to have the option to swap the calibrators on the python side so one does not need to either recompile everytime or duplicate the TrackFitting example for each different methods.

Towards that end, I propose the following changes:

- Rename `MeasurementCalibrator` to `PassThroughCalibrator`;
- Create an abstract base class `MeasurementCalibrator` with a calibrate method which takes a measurement container as input;
- Created a `PairedMeasurementCalibrator` class that binds a calibrator to a `MeasurementContainer`, to pass to the actual fitter;
- Add machinery to instantiate the calibrators outside of the `execute()` loop. 

#1940 introduced a alternate path using a RefittingCalibrator instead of a MeasurementCalibrator. I did _not_ touch this code since I'm not sure if it makes sense to unify it with what I have planned -- any thoughts @benjaminhuth ? I propose to leave this as-is for the time being and re-assess once I push actual calibrators (shortly, hopefully).


gagnonlg:
I'm aware of the segfault in some of the tests, I'm investigating...

benjaminhuth (commented):
Looks good to me, just a few comments

* The code is under `Examples/Algorithms/Calibration`, but there is no `CalibrationAlgorithm`. Would it be an option to put this code under `Examples/Framework/EventData` or something like that?

* I currently find the naming of the classes not optimal, it took me quite a while to understand. Maybe `PairedMeasurementCalibrator` -> `MeasurementCallibratorInterface` to indicate that this provides the interface to the ACTS core library... 

* So far this is not generic enough to work with my refitting-calibrator, as this is not measurement based (It defines an own source-link type, which internally already has all information, so it does not need any container or something like that). But from my perspective this is not a problem.

gagnonlg:
@benjaminhuth Thanks for the feedback! I will change the location & naming of the sources as you suggest.

For the refitting calibrator: okay then it makes sense to keep this separate!

Thanks again,
L-G

gagnonlg:
@benjaminhuth I implemented your suggestion. For the renaming, it occured to me that the PairedMeasurementCalibrator is an instance of the adapter pattern, and thus I renamed it to MeasurementCalibratorAdapter.

let me know if you have more feedback :-)

benjaminhuth (approved):
I think we can get it in now!

Examples/Python/src/TrackFitting.cpp
```diff
@@ -66,6 +66,14 @@ void addTrackFitting(Context& ctx) {
         py::arg("reverseFilteringMomThreshold"),
         py::arg("freeToBoundCorrection"), py::arg("level"));
 
+    py::class_<MeasurementCalibrator, std::shared_ptr<MeasurementCalibrator>>(
+        mex, "MeasurementCalibrator");
+
+    mex.def("makePassThroughCalibrator",
```
paulgessinger, line 72: I know this is merged already, but I there a reason not to just expose the inheritance relationship and then have Python just construct the object directly instead of the factory?

gagnonlg: No real reason I guess, I just don't know what I'm doing on the python side ;-) putting this on my list

PR #2059, opened 2023-04-25
refactor: Improve measurement errors

gagnonlg:
I propose the following two changes:

- By default, if the DigitizationAlgorithm has no variance configured, it assigns `-1`. This then artificially reduces the residual error. However, at this stage in the algorithm we know the sensor widths, so we can do the "least bad" thing:  a standard deviation of the width of the cluster's central sensor divided by sqrt(12) (i.e. assume a uniform distribution). To get a "real" error, the measurement must be calibrated at some other point (or the algorithm configured differently)
- In RootTrajectoryStateWriter.cpp, the err_x,y_hit variables were filled with the error of the residual instead. Using that error only makes sense when computing the track/hit pull, so I propose to fill the err_{x,y}_hit variables with the hit errors only. This is useful to compute e.g. truth-hit pulls.


niermann999 (approved):
Looks good to me

PR #2060, opened 2023-04-25
chore: Bump clang-format to version 14

paulgessinger:
This should give us better formatting with new language constructs like concepts in C++20.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -15,7 +15,7 @@ concurrency:
 jobs:
   format:
     runs-on: ubuntu-latest
-    container: ghcr.io/acts-project/format10:v41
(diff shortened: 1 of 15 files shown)


andiwand (approved):

andiwand:
should we also update https://github.com/acts-project/acts/blob/main/CI/check_format_local ?

paulgessinger:
> should we also update https://github.com/acts-project/acts/blob/main/CI/check_format_local ?

Good point, I'm actually wondering if we should remove that one at this point?

andiwand:
I think it is actually quite practical - why would you remove it?

benjaminhuth:
Hmm I don't use it, but if it works and we have the image anyways, why not keep it?

PR #2004, opened 2023-03-30
refactor!: move SSS coordinate information to spacePointData

LuisFelipeCoelho:
This PR moves the calculation of `topStripVector` and `bottomStripVector` to `spacePointData`, so we can calculate these values only once per event, since these values are calculated for ~95% of the input space points anyway.

@CarloVarni

PR #2067, opened 2023-04-27
refactor!: optimisations to SSS measurement details

LuisFelipeCoelho:
I am splitting some of the changes from #1985 so they can go in first

Breaking change: removing `skipPreviousTopSP ` from the seeding configuration, which is now always true unless you'r using ITk SSS measurements details `useDetailedDoubleMeasurementInfo `


Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -389,17 +391,19 @@ inline void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
...
+                return state.linCircleBottom[a].cotTheta <
+                       state.linCircleBottom[b].cotTheta;
+              });
+
+    std::sort(sorted_tops.begin(), sorted_tops.end(),
+              [&state](const std::size_t& a, const std::size_t& b) -> bool {
+                return state.linCircleTop[a].cotTheta <
+                       state.linCircleTop[b].cotTheta;
+              });
+  }
```
CarloVarni, line 406: What is `skipPreviousTopSP` doing in the rest of the seeding code? Is it used to break a loop once we reach a specific condition?

Is sorting + breaking slower then not sorting at all and run on all the tops and bottoms?

CarloVarni (commented):
Changes look good from my side. I only have a comment that is only for my understanding

Core/include/Acts/Seeding/SeedFinder.ipp
LuisFelipeCoelho: `skipPreviousTopSP + sorting` is used to avoid unnecessary iterations and speeds the code for PPP. But for SSS the sorting is lost during the filter because of the strip coordinates corrections that we apply, and if we break based on the previous sorting assumption we decrease the efficiency for SSS. So for SSS we have `skipPreviousTopSP==false` and we don't need to sort them.

CarloVarni: I understand. In other words, this will benefit only SSS

CarloVarni (approved):

Core/include/Acts/Seeding/SeedFinder.ipp
asalzburger: Hi, just a procedural detail: leaving conversions open will block the `automerge` label, so if you want to get it merged, you need to close/resolve the conversations.

CarloVarni: I'm afraid I can't close the discussion despite being me the one who opened it... 

Anyway, I've given more thoughts on the `skipPreviousTopSP` and I believe that having a flag that a user can freely set  may be a huge problem. Especially if the effect is a drop of efficiency for SSS resulting from a break in the code which leads to wrong results. That smells more like a bug.

We can't rely on the user to know the seeding code that well imho. 

Maybe we should think about a solution that would make the ACTS code to figure this out by itself and properly apply the `skipPreviousTopSP` strategy or not. Something like "oh, there is a coordinate correction, so I cannot break the loop"

LuisFelipeCoelho: `skipPreviousTopSP` could be replaced by `useDetailedDoubleMeasurementInfo` since they are related and we remove `skipPreviousTopSP` from the config. That is what I was planning to do with #2004

LuisFelipeCoelho: should I do that in this PR?

Examples/Python/src/TrackFinding.cpp
```diff
@@ -98,7 +98,6 @@ void addTrackFinding(Context& ctx) {
     ACTS_PYTHON_MEMBER(impactMax);
     ACTS_PYTHON_MEMBER(deltaZMax);
     ACTS_PYTHON_MEMBER(zBinEdges);
-    ACTS_PYTHON_MEMBER(skipPreviousTopSP);
```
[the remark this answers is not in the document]
LuisFelipeCoelho: Since the sorting of SPs is done by default now, we don't need `skipPreviousTopSP` anymore, unless for SSS where we lose the sorting after some coordinates corrections that uses SSS information. From @CarloVarni comments I decided that it was better to "merge" `skipPreviousTopSP` and `useDetailedDoubleMeasurementInfo` which are related and only used by SSS seeding

LuisFelipeCoelho:
@CarloVarni the differences affect only the number of duplicates and not the efficiency. I think it comes from the fact that `skipPreviousTopSP` was set to false by default (and now it is only false for SSS). It cannot come from the other changes because they are all related to SSS only and the CI does not test SSS. If it is ok to you I can just replace the root file.

Core/include/Acts/Seeding/SeedFinderOrthogonal.ipp
```diff
@@ -405,15 +405,12 @@ void SeedFinderOrthogonal<external_spacepoint_t>::filterCandidates(
...
-          if (cotThetaB - cotThetaT < 0) {
-            break;
-          }
-          t0 = index_t + 1;
+        // break if cotTheta from bottom SP < cotTheta from top SP because
+        // the SP are sorted by cotTheta
+        if (cotThetaB - cotThetaT < 0) {
+          break;
         }
-        continue;
```
CarloVarni, line 416: where does this continue go?

Core/include/Acts/Seeding/SeedFinderOrthogonal.ipp
```diff
@@ -448,13 +445,10 @@ void SeedFinderOrthogonal<external_spacepoint_t>::filterCandidates(
...
       if (deltaCotTheta2 > (error2 + p2scatterSigma)) {
-        if (m_config.skipPreviousTopSP) {
-          if (cotThetaB - cotThetaT < 0) {
-            break;
-          }
-          t0 = index_t;
+        if (cotThetaB - cotThetaT < 0) {
+          break;
         }
-        continue;
```
CarloVarni, line 457: same here, shouldn't it be kept?

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -597,8 +603,9 @@ inline void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
         cotThetaT = zT * std::sqrt(iDeltaRT2);
 
         rMxy = std::sqrt(rMTransf[0] * rMTransf[0] + rMTransf[1] * rMTransf[1]);
-        float Ax = rMTransf[0] / rMxy;
-        float Ay = rMTransf[1] / rMxy;
+        double irMxy = 1 / rMxy;
+        float Ax = rMTransf[0] * irMxy;
+        float Ay = rMTransf[1] * irMxy;
```
CarloVarni, line 609: I'd suggest to be explicit here:
```suggestion
        float Ax = static_cast<float>(rMTransf[0] * irMxy);
        float Ay = static_cast<float>(rMTransf[1] * irMxy);
```

paulgessinger (commented):
Looks ok as far as I can tell. The references probably need to be updated because they changed in main in the meantime.

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -546,12 +550,14 @@ inline void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
         // x_0 and y_0
         float A0 = (lt.V - Vb) / dU;
 
+        float zPositionMiddle = cosTheta * std::sqrt(1 + A0 * A0);
```
andiwand, line 553:
```suggestion
        float zPositionMiddle = cosTheta * std::hypot(1, A0);
```

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -604,9 +610,10 @@ inline void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
         cotThetaB = -zB * std::sqrt(iDeltaRB2);
         cotThetaT = zT * std::sqrt(iDeltaRT2);
 
-        rMxy = std::hypot(rMTransf[0], rMTransf[1]);
-        float Ax = rMTransf[0] / rMxy;
-        float Ay = rMTransf[1] / rMxy;
+        rMxy = std::sqrt(rMTransf[0] * rMTransf[0] + rMTransf[1] * rMTransf[1]);
```
andiwand, line 613:
```suggestion
        rMxy = std::hypot(rMTransf[0], rMTransf[1]);
```

I guess `Acts::VectorHelpers::perp` might also be an option here if this is an Eigen object

Core/include/Acts/Seeding/SeedFinder.ipp
LuisFelipeCoelho: Thank you @andiwand for the suggestions. Isn't `std::hypot` a lot slower than `std::sqrt(x*x + y*y)`?

andiwand: right - I guess it depends on your accuracy vs performance targets. feel free to resolve my comments

andiwand (approved):

LuisFelipeCoelho:
For some reason I cannot resolve the conflicts with the two reference files:
CI/physmon/reference/performance_ckf_orthogonal.root
CI/physmon/reference/performance_ckf_seeded.root

PR #2068, opened 2023-04-27
refactor!: TrackStateType as bitset view instead of bitset

paulgessinger:
This decouples the storage a bit from the semantics. It allows the backends to store `uint64_t` and have the rest of the code transparently interpret this as a bitset.


andiwand (approved):

paulgessinger:
Can you have another look @andiwand?

PR #2070, opened 2023-04-27
fix: ambiguity resolution for 0 tracks

andiwand:
0 tracks as input resulted in UB. this PR is fixing that

also I bumped into the same boost bug as in https://github.com/acts-project/acts/pull/2049/ and worked around it again

fixes https://github.com/acts-project/acts/issues/2055


andiwand:
cc @Corentin-Allaire

Corentin-Allaire (approved):
Great find !

andiwand:
> This is specific the the standard solver implementation right ? The ML doesn't need change.

yeah exactly - I didn't check the ML solver but as far as I can remember you loop through the input tracks directly and 0 tracks should there be handled fine

PR #2071, opened 2023-04-27
refactor: replace warning for unmatched particles with info

andiwand:
after https://github.com/acts-project/acts/pull/2049 we see a lot of warnings like
```
RootTrajecto   WARNING   Truth particle with barcode 124|0|7|1|8=558446353911447560
```

I believe they are cause by filtered truth particles which now produce hits and get reconstructed. the root writers then find the tracks with particles which are not in the list of selected ones

after discussion with @Corentin-Allaire and @benjaminhuth it makes most sense to me to just level down the log and guarantee that those tracks are flagged as fake


Corentin-Allaire:
Shouldn't you change the level to VERBOSE or DEBUG instead of INFO ?

benjaminhuth:
`DEBUG` is good I think, because `INFO` should never fill the screen with hundreds of lines^^

PR #2072, opened 2023-04-28
feat!: add non-cost access to Surface

asalzburger:
This PR adds a non-const access to the associated `Surface` to the detector element interface and all derivates.

```c++
virtual Surface& surface() = 0;
```

Is now required to be implemented. It will allow for full const correctness of the new `Detector` geometry, having `non-const` object while detector creation (which will in turn allow for loading of material and setting of `GeometryID` and other associations, but require strict `const correctness` after geometry construction.

It removes already some now unnecessary `const_pointer_cast` or diversions via `MutableSurfacePtr` in the examples.

### DIFF
--- a/Core/include/Acts/Geometry/detail/DefaultDetectorElementBase.hpp
+++ b/Core/include/Acts/Geometry/detail/DefaultDetectorElementBase.hpp
@@ -35,8 +35,11 @@ class DetectorElementBase {
   /// @param gctx The current geometry context object, e.g. alignment
   virtual const Transform3& transform(const GeometryContext& gctx) const = 0;
 
-  /// Return surface representation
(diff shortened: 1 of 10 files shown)


AJPfleger (approved):
lgtm

paulgessinger (commented):
Ok, but we need to be very careful that we don't start keeping `std::shared_ptr<Surface>` spread across our code, because that might cause thread-safety issues down the line.

asalzburger:
> Ok, but we need to be very careful that we don't start keeping `std::shared_ptr<Surface>` spread across our code, because that might cause thread-safety issues down the line.

As said - this is only at construction time of the geometry, and only then - during event processing the access to the detector surfaces is `const Surface*`

PR #2078, opened 2023-05-02
feat: Add support for passing clusters to calibrator

gagnonlg:
Currently, the measurement calibrators in TrackFittingAlgorithm do not have access to cluster-level information. One widely-used pattern in the real world is to apply different calibrations depending on the cluster dimensions for instance, so this information is useful. This PR adds an optional input to the TrackFittingAlgorithm which is handed down through the MeasurementCalibrator interface. The reference to the cluster collection is boxed in a `std::optional<std::reference_wrapper>` to simplify the interface.

Ping @benjaminhuth for comments

### DIFF
--- a/Examples/Algorithms/TrackFitting/include/ActsExamples/TrackFitting/TrackFittingAlgorithm.hpp
+++ b/Examples/Algorithms/TrackFitting/include/ActsExamples/TrackFitting/TrackFittingAlgorithm.hpp
@@ -8,6 +8,7 @@
 
 #pragma once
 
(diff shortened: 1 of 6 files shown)


benjaminhuth (approved):
Just a comment: For me there is not real semantical advantage of an `std::optional<std::reference_wrapper<const T>>` to a `const T*`.

I think I would actually go for `const T*` in such a case, because it is less verbose. But of course this is only my opinion, feel free to ignore it :)

AJPfleger (approved):

PR #2085, opened 2023-05-04
feat: Simple scaling calibrator

gagnonlg:
This PR adds a new measurement calibrator, which offsets the measurement positions and scales the error based on a set of 2-D maps of cluster sizes (loc0 / loc1). The granularity of the maps can be different based on Volume, Layer, or Sensitive surface.

This PR does not add documentation of the method -- that will follow later once I fully develop a good setup to derive some good maps (I have a WIP implementation that I used for testing).

Lastly, note that this is not meant to be a state of the art method, it's basically to showcase the simplest non-trivial use of the measurement calibration interface. 

Ping @benjaminhuth if you can look at it, or suggest someone else!


Examples/Framework/include/ActsExamples/EventData/ScalingCalibrator.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include <TFile.h>
+#include <TH2D.h>
+
+namespace ActsExamples {
+
+class ScalingCalibrator : public MeasurementCalibrator {
+ public:
+  struct ConstantTuple {
+    double x_offset;
```
benjaminhuth, line 22: Not sure if clang-tidy will also warn you, but we should initialize these values (to 0?) to ensure we have always initialized values here.

Examples/Framework/src/EventData/ScalingCalibrator.cpp
```diff
@@ -0,0 +1,145 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#include <Acts/Definitions/TrackParametrization.hpp>
+#include <ActsExamples/EventData/ScalingCalibrator.hpp>
+
+#include <regex>
+
+#include <TKey.h>
+
+namespace internal {
```
benjaminhuth, line 16: to follow our convention, maybe rename to `namespace detail`?

Examples/Framework/include/ActsExamples/EventData/ScalingCalibrator.hpp
```diff
@@ -0,0 +1,63 @@
...
+          x_scale.GetBinContent(x_scale.FindFixBin(sizeLoc0, sizeLoc1));
+      ct.y_offset =
+          y_offset.GetBinContent(y_offset.FindFixBin(sizeLoc0, sizeLoc1));
+      ct.y_scale =
+          y_scale.GetBinContent(y_scale.FindFixBin(sizeLoc0, sizeLoc1));
+      return ct;
+    }
+  };
+
+  ScalingCalibrator(const char* path);
```
benjaminhuth, line 48: I would prefer `std::string_view` here, maybe actually `std::filesystem::path` instead of `const char *`

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -820,6 +820,8 @@ def addKalmanTracks(
     multipleScattering: bool = True,
     energyLoss: bool = True,
     clusters: str = None,
+    calibrationConfigFile: str = None,
+    clusters: str = None,
```
benjaminhuth, line 823: looks like we have two `clusters` arguments now?

benjaminhuth: A I think that has been fixed during the review :D

Examples/Framework/include/ActsExamples/EventData/ScalingCalibrator.hpp
gagnonlg: Done -- initialized the offsets to 0 and the scales to 1.

Examples/Framework/include/ActsExamples/EventData/ScalingCalibrator.hpp
gagnonlg: Done

benjaminhuth (approved):
Okay, let's get it in then!

Examples/Framework/src/EventData/ScalingCalibrator.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include <ActsExamples/EventData/ScalingCalibrator.hpp>
+
+#include <regex>
+
+#include <TKey.h>
+
+namespace detail {
+
+std::pair<Acts::GeometryIdentifier, std::string> parseMapKey(
+    const std::string& mapkey) {
```
paulgessinger, line 19: This is needed because you encode the geo id in the ROOT histogram name as a string, correct?

gagnonlg: Yes, that's the idea. Probably not the cleanest way to do it...

cmake/FindOnnxRuntime.cmake
```diff
@@ -28,6 +28,9 @@ if(NOT OnnxRuntime_INCLUDE_DIR)
 else()
   file(READ ${OnnxRuntime_INCLUDE_DIR}/onnxruntime_c_api.h ver)
   string(REGEX MATCH "ORT_API_VERSION ([0-9]*)" _ ${ver})
+  if ("${CMAKE_MATCH_1}" STREQUAL "LAST")
```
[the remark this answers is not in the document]
gagnonlg: Yeah, drive-by edit that slipped in, should I remove it?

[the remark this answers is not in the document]
paulgessinger: I'd prefer a separate PR, but up to you.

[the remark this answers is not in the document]
gagnonlg: sure -- will remove asap

paulgessinger (approved):
Seems good to go to me.

PR #2088, opened 2023-05-04
feat: add first round of DD4hep UnitTests

asalzburger:
This PR adds the first round of DD4hep unit tests:
 - this tests single shapes (Cylinder, Disc, Rectangle, Trapezoid) and their translation from DD4hep via the TGeoShape converter
 
It does not yet do any more tests (material, structure, binning), which will be added at a later stage.


asalzburger:
Hi @paulgessinger - I have ... of course .. some troubles with this new DD4hep unit tests, I think I need to source this_acts.sh in order to get the DD4hep factories found?

paulgessinger:
> Hi @paulgessinger - I have ... of course .. some troubles with this new DD4hep unit tests, I think I need to source this_acts.sh in order to get the DD4hep factories found?

At least the environment variable needs to be set, I think, yes. Maybe it's possible at the CMake level to configure the test so that this environment variable is set. Maybe @krasznaa knows if this is feasible?

asalzburger:
> > Hi @paulgessinger - I have ... of course .. some troubles with this new DD4hep unit tests, I think I need to source this_acts.sh in order to get the DD4hep factories found?
> 
> At least the environment variable needs to be set, I think, yes. Maybe it's possible at the CMake level to configure the test so that this environment variable is set. Maybe @krasznaa knows if this is feasible?

Which environment variable is that?

paulgessinger:
I believe it's looking into `LD_LIBRARY_PATH` on Linux and `DYLD_LIBRARY_PATH` on macOS.

asalzburger:
I will give this a shot probably with:

```
See also the [set()](https://cmake.org/cmake/help/v3.7/command/set.html#command:set) command to set ENV{VAR}.
```

from the Cmake documentation.

krasznaa:
> > Hi @paulgessinger - I have ... of course .. some troubles with this new DD4hep unit tests, I think I need to source this_acts.sh in order to get the DD4hep factories found?
> 
> At least the environment variable needs to be set, I think, yes. Maybe it's possible at the CMake level to configure the test so that this environment variable is set. Maybe @krasznaa knows if this is feasible?

With some considerable delay...

The [add_test(...)](https://cmake.org/cmake/help/latest/command/add_test.html) function can do a lot of things. In the very simplest case it executes a compiled binary application, which itself would've been built to run the tests. (Usually with the help of some unit testing framework/library.)

But it can pretty much execute anything. It just needs to be something that returns `0` on success, and anything else on failure. In your case, you may be able to get away with something as simple as:

```cmake
add_test( NAME dd4HepTest
   COMMAND "${CMAKE_BINARY_DIR}/bin/this_acts.sh" && testExecutable )
```

But in the offline code for instance we use a whole "script layer" between CMake and the test executables. As that script layer not only ensures that the correct environment would be set up for the application, but also can check the printed output of the test application against a reference.

The simplest component in that scripting setup is this thing:

https://gitlab.cern.ch/atlas/atlasexternals/-/blob/master/Build/AtlasCMake/modules/scripts/atlas_build_run.sh.in

Which we use all over the place. Not even just for test running, but even for linking libraries/executables correctly. (https://gitlab.cern.ch/atlas/atlasexternals/-/blob/master/Build/AtlasCMake/modules/AtlasFunctions.cmake#L108-127 It's a longer story why we need that.) Note that using an "executable script" instead of a "sourceable script" like this ensures that we would always use BASH, and would not be left up to the whims of the user.

Long story short, if some Acts test needs the runtime environment from `this_acts.sh`, then you must work that script into the `COMMAND` of the test.

paulgessinger:
I think its possible to set environment variables by wrapping the executable call in a CMake call in script mode.

krasznaa:
> I think its possible to set environment variables by wrapping the executable call in a CMake call in script mode.

Yes, that's one way of doing it. Like:

https://gitlab.cern.ch/atlas/atlasexternals/-/blob/master/External/PyAnalysis/CMakeLists.txt#L52-57

As long as you know all the environment variables that you need, this is an okay way to go about it. It's just that in the offline code we allow "packages" to set whatever custom environment variable that they want. So we have to use a single system (through generated environment setup scripts) that sets up the runtime environment of an ATLAS release.

paulgessinger:
Actually, check this out: https://cmake.org/cmake/help/latest/prop_test/ENVIRONMENT.html

krasznaa:
> Actually, check this out: https://cmake.org/cmake/help/latest/prop_test/ENVIRONMENT.html

Yepp, I sort of forgot about that. But we do use that quite a bit in the offline code as well.
  - https://gitlab.cern.ch/atlas/atlasexternals/-/blob/master/Build/AtlasCMake/modules/AtlasTestFunctions.cmake#L320-323
  - https://gitlab.cern.ch/atlas/athena/-/blob/master/Database/APR/APRTests/CMakeLists.txt#L42
  - etc.

As I wrote before, as long as you know **exactly** which environment variables you need to modify, doing this from the CMake configuration directly in some way is a good way to go about it.

asalzburger:
This will be superseded, closing this one.

PR #2089, opened 2023-05-05
refactor: add geohook for fast digitization range selection to itk example

niermann999:
Adds a geoidhook to the ```itk.py``` example that performs the matching of geometric digitization configurations to detector regions. The json config file was adjusted accordingly and will be pushed to the acts itk repository once validated again.


andiwand:
how do we proceed on this @niermann999 ?

niermann999:
> how do we proceed on this @niermann999 ?

Good question. The problem was, if I remember correctly, that I cannot use the center position of the modules to determine the digitization domain, because the center of the annulus shapes lies close to the beamline instead of the centroid. But I did not know back then how to call annulus specific code from the python bindings, since those are virtual functions

paulgessinger:
What's the plan of action on this one? Do we move ahead with this, or close it?

AJPfleger:
as discussed in the dev-meeting: let's include @benjaminhuth into the discussion and decide on how to continue with this PR :)

PR #2092, opened 2023-05-05
feat: Move MultiTrajectory tests to common suite

paulgessinger:
This allows us to use these tests more like a compliance suite, and it
can be used to test other backend implementation (like the upcoming
PODIO one)


asalzburger (approved):
Looks good.

PR #2095, opened 2023-05-05
fix: Dangling reference due to source link lifetime in fitters

### LINKED ISSUE #2000: Bug: Segmentation fault when calling GSF fitting from Athena
I'm currently doing refitting of ATLAS signle-electron tracks inside Athena (the ATLAS software environment) with the ACTS GSF, so calling the fit function of GSF.

Using a custom version of ACTS with Athena and doing several `std::cout`s to see where the segmentation fault occurs, I have located the issue to this line: https://github.com/acts-project/acts/blob/v23.5.0/Core/include/Acts/TrackFitting/GaussianSumFitter.hpp#L288

Athena is currently using ACTS v23.5.0. I have also tried it with ACTS v23.5.0 + cherry-picking the following fixes I saw in v24:

- (#1927) (7f0429f)
- (#1933) (250ea03) 
- (#1935) (0b9004f)
- (#1941) (58bb693)
- (#1963) (44eff4f)
- (#1970) (a6b3d6a)

The segmentation error still occurs, and at the same place.

[Bug]

Edit: typo

---

Created a MM channel here: https://mattermost.web.cern.ch/acts/channels/bug-gsf-segfault

andiwand:
Closes #2000

the source links in the GFS went out of scope and caused segfaults

- this PR harmonizes the source link handling for the different fitters
- fixes a potential UB caused by parameter evaluation order
- fixes a potential dangling reference in `SourceLinkAdapterIterator`


Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -289,7 +289,7 @@ class KalmanFitter {
     const Surface* targetSurface = nullptr;
 
     /// Allows retrieving measurements for a surface
-    const std::map<GeometryIdentifier, SourceLink>* inputMeasurements = nullptr;
+    std::map<GeometryIdentifier, SourceLink> inputMeasurements;
```
paulgessinger, line 292: Are you sure this doesn't cause extra copies?

paulgessinger (commented):
Looks good. 

Can you remove the changes to the citation files and version number? They are updated automatically by the release script when we make a release and I'd like to avoid merge conflicts and inconsistencies?

Core/include/Acts/TrackFitting/KalmanFitter.hpp
andiwand: I think it does but the pointer is also weird. but we can `move` if you prefer

andiwand:
> Can you remove the changes to the citation files and version number? They are updated automatically by the release script when we make a release and I'd like to avoid merge conflicts and inconsistencies?

this should be gone

paulgessinger:
Not sure what the issue is with the CI bridge here. Maybe already overloaded due to too many LCG jobs.

andiwand:
did the move now except for the GSF @benjaminhuth because there we use the map actually multiple times. I guess a reference/pointer would actually make sense here but since it will be a small map a copy should also be fine

what do you think?

paulgessinger:
Just keep in mind that copying the contained `SourceLink`s is not free for non-trivially copyable types.

paulgessinger (approved):

PR #2086, opened 2023-05-04
fix: Improve full chain pulls by enabling Fatras interactions

andiwand:
Our full chain pulls are in a bad state. Looks like the reconstruction and simulation energy loss did not match up. This PR switches the Fatras interactions on which should bring our pulls back to standard normal distribution.

Fixes
- https://github.com/acts-project/acts/issues/1643

Blocked by
- https://github.com/acts-project/acts/pull/2157
- https://github.com/acts-project/acts/pull/2239
- https://github.com/acts-project/acts/pull/2295
- https://github.com/acts-project/acts/pull/2293
- https://github.com/acts-project/acts/pull/2294

PR #2099, opened 2023-05-06
feat: push and pop for FPE monitor

andiwand:
with https://github.com/acts-project/acts/pull/2086 I am running into FPEs which seem to come from special input data configurations. one of them popped up in Eigen inverse and this is something out of our control.

my proposal is to disable the FPEs for sections that may throw one and we cannot fix them or cannot fix them right away.

I am not sure how big the performance impact of this is but I also don't really see an alternative

### DIFF
--- a/Core/include/Acts/Utilities/FpeMonitor.hpp
+++ b/Core/include/Acts/Utilities/FpeMonitor.hpp
@@ -21,8 +21,13 @@ class FpeMonitor {
   static void enable(int excepts);
   static void disable(int excepts);
 
+  static void push();
+  static void pop();
+
  private:
-  int m_excepts;
+  int m_excepts{0};
+
+  static std::vector<int>& stack();
 };
 
 }  // namespace Acts
--- a/Core/include/Acts/Utilities/FpeMonitor.hpp
+++ b/Core/include/Acts/Utilities/FpeMonitor.hpp
@@ -10,6 +10,9 @@
 
 #include "Acts/Utilities/Helpers.hpp"
 
+#include <stack>
+#include <vector>
+
 namespace Acts {
 
 class FpeMonitor {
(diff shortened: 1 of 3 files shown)


Core/include/Acts/Utilities/FpeMonitor.hpp
```diff
@@ -21,8 +21,13 @@ class FpeMonitor {
...
   static void disable(int excepts);
 
+  static void push();
+  static void pop();
+
  private:
-  int m_excepts;
+  int m_excepts{0};
+
+  static std::vector<int>& stack();
```
paulgessinger, line 30: This could literally be a `std::stack` I guess :D

paulgessinger:
How does this interact with the context manager at the python level? Is it clear what's supposed to happen when you push disable? Should we do nested context managers like 

```python
with acts.FpeMonitor() as fpe:
  with fpe.disable():
    pass
```

andiwand:
I thought about making the RAII interface using push pop but not sure what make more sense

paulgessinger:
We don't have a strong need in C++ I would say, but in python, context managers are the RAII equivalent and might be helpful. Just a thought though.

andiwand:
right now the python interface did not change. I added the push and pop functions there too but it might not be necessary

```
with acts.FpeMonitor() as fpe1:
  with acts.FpeMonitor() as fpe2:
    pass
```

will go through RAII and do the `enable` `disable` as it should in the right order

what we can do now in C++ is call `push` `pop` to get a more robust environment when needed

maybe it would good to remove the enable, disable, push, pop from python so nobody can interfere with the C++ tho

paulgessinger:
But did we want to enable FPE monitoring by default and from C++ anyway? My thinking was we only do this at the Python level when we want to catch them, but not in "production".

andiwand:
yeah you are right but I fear we will have known FPEs at some point (like nowish) and we need to mask them somehow because some of them will not be fixable I fear. (or at least not fixable right away)

andiwand:
I guess alternatively we could log an error somehow and continue execution. but you mentioned that this is also kind of tricky?

paulgessinger:
Yeah you need to hack to continue the process after the signal is raised. This is certainly possible but kind of ugly.

PR #2104, opened 2023-05-06
refactor: Deduplicate some code in SingleBoundTrackParameters

andiwand:
- deduplicate some code
- add local param accessor


Core/include/Acts/EventData/SingleBoundTrackParameters.hpp
```diff
@@ -168,6 +168,8 @@ class SingleBoundTrackParameters {
     return m_params[kIndex];
   }
 
+  /// Local spatial position two-vector.
+  Vector2 localPosition() const { return m_params.head<2>(); }
```
benjaminhuth, line 172:
```suggestion
  Vector2 localPosition() const { return m_params.segment<2>(eBoundLoc0); }
```
for consistency maybe?

benjaminhuth (approved):

Core/include/Acts/EventData/SingleBoundTrackParameters.hpp
benjaminhuth: Hmm since `eBoundLoc0` is a compile time constant, I would expect the compiler to be able to optimize this, but not sure of course.
However, isn't it kind of our style to use these `enum`-values? https://acts.readthedocs.io/en/latest/codeguide.html#a-indices-always-use-enum-values-to-access-vector-matrix-components

paulgessinger: I guess so, but you're making the assumption that `eLoc0` and `eLoc1` are consecutive in any case.

benjaminhuth: Thats true, we could also enforce this in `TrackParameterization.hpp` similar to the free coordinates I think...

PR #2105, opened 2023-05-06
fix: Add NotOnSurface error to globalToLocal in LineSurface

andiwand:
`LineSurface::globalToLocal` forces the given position on the surface without checking if it actually is

I am utilizing the `localToGlobal` function to circle back and check how far we are actually off and check that against the tolerance

### DIFF
--- a/Core/src/Surfaces/LineSurface.cpp
+++ b/Core/src/Surfaces/LineSurface.cpp
@@ -8,6 +8,7 @@
 
 #include "Acts/Surfaces/LineSurface.hpp"
 
+#include "Acts/Definitions/Algebra.hpp"
 #include "Acts/EventData/detail/TransformationBoundToFree.hpp"
 #include "Acts/Utilities/ThrowAssert.hpp"
 
@@ -63,7 +64,7 @@ Acts::Vector3 Acts::LineSurface::localToGlobal(const GeometryContext& gctx,
 
 Acts::Result<Acts::Vector2> Acts::LineSurface::globalToLocal(
     const GeometryContext& gctx, const Vector3& position,
-    const Vector3& momentum, double /*tolerance*/) const {
+    const Vector3& momentum, double tolerance) const {
   using VectorHelpers::perp;
   const auto& sTransform = transform(gctx);
   const auto& tMatrix = sTransform.matrix();
@@ -77,6 +78,12 @@ Acts::Result<Acts::Vector2> Acts::LineSurface::globalToLocal(
   // assign the right sign
   double sign = ((lineDirection.cross(momentum)).dot(decVec) < 0.) ? -1. : 1.;
   lposition[eBoundLoc0] *= sign;
+
+  if ((localToGlobal(gctx, lposition, momentum) - position).norm() >
+      tolerance) {
+    return Result<Vector2>::failure(SurfaceError::GlobalPositionNotOnSurface);
+  }
+
   return Result<Vector2>::success(lposition);
 }
 


Core/src/Surfaces/LineSurface.cpp
```diff
@@ -77,6 +78,12 @@ Acts::Result<Acts::Vector2> Acts::LineSurface::globalToLocal(
   // assign the right sign
   double sign = ((lineDirection.cross(momentum)).dot(decVec) < 0.) ? -1. : 1.;
   lposition[eBoundLoc0] *= sign;
+
+  if ((localToGlobal(gctx, lposition, momentum) - position).norm() >
```
beomki-yeo, line 82: I think  `localToGlobal` would be expensive

I suggest to use the boundaries defined `LineBounds.hpp`

`std::abs(lposition[0) < r && std::abs(lposition[1]) < halfz`

andiwand: `localToGlobal` will not only check the bounds but also how far we got projected onto the surface by the code above. the calculation above will always produce a local position even if we are not on the surface

beomki-yeo: > the calculation above will always produce a local position even if we are not on the surface

This is true but you can still use `lposition[0]` and `lposition[1]` for boundary check regardless whether it is on the surface or not.

andiwand: that is true but the `lposition` you get will not represent the global `position` if you provide input that is not in the PCA. it will project you onto the surface and I don't think it should

`globalToLocal(<random global position>)` will always give you an answer and the inverse `localToGlobal(globalToLocal(<random global position>)) != random global position` will not give you the same position back because you potentially moved it across you whole detector

beomki-yeo: oh I understand the point. In that case you can check if the track direction (`t`) is orthogonal to the radial vector (`w`)

```c++
  // direction
  vector3  t = normalize(mom);
  vector3 r = position;
 
  // line center
  c = sCenter 
 
  // line axis direction
  vector3 l = lineDirection
  
  // Projection of (track position - center) to the line 
  double A = dot(r-c,l)
  
  // Radial vector: Vector orthogonal to the line and passing the track position
  // w = r - (c + ((r - c) * l)l)
  const vector3 w = r - (c + A * l);

  if ( std::abs(dot( t* w)) > tolerance){
     // failure
  } 
```

It is also expensive but cheaper that `localToGlobal` I suppose.

andiwand: but I think `localToGlobal` does exactly that, no? at least when I look at the implementation is seems quite simple

beomki-yeo: I thought this line would be expensive.
```c++
Vector3 locZinGlobal = sTransform * Vector3(0., 0., lposition[eBoundLoc1]);
```

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
```diff
@@ -43,37 +43,9 @@ auto MultiEigenStepperLoop<E, R, A>::boundState(
...
-    std::cout << boundMeanRes.error().message() << std::endl;
-  }
-  assert(boundMeanRes.ok());
-
-  // TODO Not sure if this is entirely correct, visit again later
-  const auto cov =
+  const auto [mean, cov] =
       detail::angleDescriptionSwitch(surface, [&](const auto& desc) {
-        return detail::combineCov(states, *boundMeanRes, 1.0, Acts::Identity{}, desc);
+        return detail::combineGaussianMixture(states, Acts::Identity{}, desc);
```
paulgessinger: I see it was there before, but where does `Acts::Identity` actually come from??

Core/src/Surfaces/LineSurface.cpp
paulgessinger: Tbh, I'm also slightly worried about the performance implications.

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
andiwand: @benjaminhuth

Core/src/Surfaces/LineSurface.cpp
andiwand: I don't feel like this is a particular hot path but I am happy to improve performance on suggestions. right now I just want to fix the bug and face a ton of downstream failures

for example we have another problem with numerical instabilities. some intersections will give you global positions which are not on surface according to these changes

Core/include/Acts/Propagator/MultiEigenStepperLoop.ipp
benjaminhuth: hmm from `Utilities/Identity.hpp`. Basically from C++20 on it is 
```c++
namespace Acts {
using Identity = std::identity;
}
```
Before, it is a backport of this. I use this projection-pattern a bit in the GSF code, inspired from the ranges library.

paulgessinger: I was unaware of this.

andiwand:
@benjaminhuth will check if that works out for the GSF. otherwise I think this is ready

andiwand:
> Do the downstream issues show up as failures? How do we tackle them?

I think they are all fixed now except for GSF. I will check again tomorrow

benjaminhuth:
I'm optimistic I can fix the GSF tomorrow morning

andiwand:
@benjaminhuth fixed the GSF - looks like we can move forward with this tomorrow

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -1212,6 +1216,9 @@ class CombinatorialKalmanFilter {
       // Set accumulatd path to zero before targeting surface
       state.stepping.pathAccumulated = 0.;
 
+      navigator.targetReached(state.navigation, false);
```
benjaminhuth, line 1219: maybe a short comment why this must be done would be helpful for future developers

andiwand, line 1219:
```suggestion
      // Reset the navigation state to continue propagation towards the target surface
      navigator.targetReached(state.navigation, false);
```

Tests/UnitTests/Core/TrackFitting/GsfComponentMergingTests.cpp
```diff
@@ -173,6 +173,13 @@ BoundVector meanFromFree(std::vector<DummyComponent<eBoundSize>> cmps,
 
   mean.segment<3>(eFreeDir0).normalize();
 
+  // project the position on the surface
```
benjaminhuth, line 176:
```suggestion
  // Project the position on the surface.
  // This is mainly necessary for the perigee surface, where 
  // the mean might not fulfill the perigee condition.
```

benjaminhuth (approved):

paulgessinger:
@CarloVarni @andiwand @CarloVarni: should this be backported to ACTS `v23.5.0` so we can deploy to Athena `23.0`?

PR #2107, opened 2023-05-08
fix: SIGFPE in full_chain_odd example

### LINKED ISSUE #2046: Bug: FPE_OVERFLOW in SeedFinder.ipp
I think we found a reproducible full_chain_odd.py crash at event 322 when running Fatras with ttbar due to PE_OVERFLOW
The crash seems to be caused by  
https://github.com/acts-project/acts/blob/61c067e161839f429353b7ee79826562dbb95e2e/Core/include/Acts/Seeding/SeedFinder.ipp#L648
(@benjaminhuth nailed it down)
Can be reproduced with current acts main branch.

owtscharenko:
This PR fixes #2046, an SIGFPE in event 51 when calling the full_chain_odd example with --ttbar

The FPE occurs in line [699](https://github.com/acts-project/acts/blob/2e45ac2de349f1341f85c282a5384ae819bc4a90/Core/include/Acts/Seeding/SeedFinder.ipp#L698) in SeedFinder.ipp when calculating the pT from a helix radius as input for the scattering sigma.
The SeedFinder checks whether a maxPtScattering is set and in case pT is larger, a default p2scatterSigma is calcuated from maxPtScattering.
In case of B2==0 the calculation of p2scatterSigma now also uses maxPtScattering and skips calculation of pT to avoid 0-division.


LuisFelipeCoelho (commented):
Looks good to me. Could you just add a more informative comment explaining that if B2==0 we calculate p2scatterSigma for maxPtScattering

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -695,12 +695,21 @@ inline void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
       if (!std::isinf(m_config.maxPtScattering)) {
         // if pT > maxPtScattering, calculate allowed scattering angle using
         // maxPtScattering instead of pt.
-        float pT = options.pTPerHelixRadius * std::sqrt(S2 / B2) / 2.;
-        if (pT > m_config.maxPtScattering) {
+        // To avoid 0-divison the pT check is skipped in case of B2==0, and
+        // p2scatterSigma is calculated directly from pTscatterSigma
```
LuisFelipeCoelho, line 699:
```suggestion
        // p2scatterSigma is calculated directly from maxPtScattering
```

LuisFelipeCoelho (approved):

LuisFelipeCoelho:
Thank you @owtscharenko for fixing this

PR #2110, opened 2023-05-09
refactor: Min measurements for TrackSelector

andiwand:
currently we are passing all the tracks we find downstream without filtering them. this results in bad performance and weird plots (e.g. https://github.com/acts-project/acts/pull/2086). in this PR I attempt to combine `CKFPerformanceConfig` and `TrackSelectorRanges` into a single `TrackSelectorConfig` so we can select the tracks a single time and use them in a consistent way downstream


benjaminhuth (approved):
Looks good to me!

benjaminhuth (approved):

Examples/Scripts/Python/full_chain_itk.py
```diff
@@ -97,15 +96,13 @@
     s,
     trackingGeometry,
     field,
-    CKFPerformanceConfig(ptMin=1.0 * u.GeV if ttbar_pu200 else 0.0, nMeasurementsMin=6),
-    TrackSelectorRanges(pt=(1.0 * u.GeV, None), absEta=(None, 4.0)),
+    TrackSelectorConfig(pt=(1.0 * u.GeV, None), absEta=(None, 4.0), minMeasurements=6),
```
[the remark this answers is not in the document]
andiwand: true but that will only change the output of the CKF performance writer and I also also wanted to align the minimum pT cutoffs everywhere

CI/physmon/physmon.py
```diff
@@ -167,7 +166,7 @@ def run_ckf_tracking(truthSmearedSeeded, truthEstimatedSeeded, label):
             s,
             trackingGeometry,
             field,
-            TruthSeedRanges(pt=(500.0 * u.MeV, None), nHits=(9, None)),
+            TruthSeedRanges(pt=(400.0 * u.MeV, None), nHits=(9, None)),
```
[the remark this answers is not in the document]
andiwand: we use `400` everywhere else in the chain so I wanted to align that

Examples/Scripts/Python/full_chain_itk.py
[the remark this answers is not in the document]
paulgessinger: Ok, but it becomes completely untraceable what changes the output? We also don't automatically run the ttbar atm, so we can't tell how this change that?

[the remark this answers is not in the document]
andiwand: the downstream algorithms should be unaffected by the `CKFPerformanceConfig`. what will change is the amount of tracks the vertexing will get because of the `minMeasurements=6`.

what will also change is the IO output of the CKF because tracksummary and trackstates will also be effected by the filter. not sure how bad this is for people who use the full chain but this was the idea to make the pulls unbiased by outlier tracks

andiwand, line 99:
```suggestion
    TrackSelectorConfig(pt=(1.0 * u.GeV if ttbar_pu200 else 0.0, None), absEta=(None, 4.0), minMeasurements=6),
```

CI/physmon/physmon.py
```diff
@@ -179,7 +178,7 @@ def run_ckf_tracking(truthSmearedSeeded, truthEstimatedSeeded, label):
                 maxSeedsPerSpM=1,
                 sigmaScattering=5,
                 radLengthPerSeed=0.1,
-                minPt=500 * u.MeV,
+                minPt=400 * u.MeV,
```
andiwand, line 181:
```suggestion
                minPt=500 * u.MeV,
```

CI/physmon/physmon.py
```diff
@@ -200,10 +199,10 @@ def run_ckf_tracking(truthSmearedSeeded, truthEstimatedSeeded, label):
             s,
             trackingGeometry,
             field,
-            CKFPerformanceConfig(ptMin=400.0 * u.MeV, nMeasurementsMin=6),
-            TrackSelectorRanges(
+            TrackSelectorConfig(
                 loc0=(-4.0 * u.mm, 4.0 * u.mm),
-                pt=(500 * u.MeV, None),
+                pt=(400 * u.MeV, None),
```
andiwand, line 204:
```suggestion
                pt=(500 * u.MeV, None),
```

CI/physmon/physmon.py
```diff
@@ -298,7 +296,7 @@ def run_vertexing(fitter, mu, events):
                 maxSeedsPerSpM=1,
                 sigmaScattering=5,
                 radLengthPerSeed=0.1,
-                minPt=500 * u.MeV,
+                minPt=400 * u.MeV,
```
andiwand, line 299:
```suggestion
                minPt=500 * u.MeV,
```

CI/physmon/physmon.py
```diff
@@ -311,18 +309,17 @@ def run_vertexing(fitter, mu, events):
             s,
             trackingGeometry,
             field,
-            CKFPerformanceConfig(ptMin=400.0 * u.MeV, nMeasurementsMin=6),
-            TrackSelectorRanges(
-                pt=(500 * u.MeV, None),
+            TrackSelectorConfig(
+                pt=(400 * u.MeV, None),
```
andiwand, line 313:
```suggestion
                pt=(500 * u.MeV, None),
```

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -169,14 +168,11 @@
         s,
         trackingGeometry,
         field,
-        CKFPerformanceConfig(
-            ptMin=1.0 * u.GeV if ttbar else 0.0,
-            nMeasurementsMin=7,
-        ),
-        TrackSelectorRanges(
+        TrackSelectorConfig(
             pt=(1.0 * u.GeV, None),
```
andiwand, line 172:
```suggestion
            pt=(1.0 * u.GeV if ttbar else 0.0, None),
```

paulgessinger:
Truth estimated CKF still seems to fail: https://herald.harbinger.gessinger.dev/view/acts-project/acts/712272704/ckf_truth_estimated.html

andiwand:
yeah I think that comes from the fact that I changed the selector for this output. I will update the reference

paulgessinger (approved):
let's go

PR #2111, opened 2023-05-09
feat: NeuralCalibrator: ML-based measurement calibration

gagnonlg:
This PR implements the Mixture Density Network measurement calibration that is used in the ATLAS experiment to calibrate pixel cluster positions. For documentation at the moment, there is a comment in the .hpp with a description of the method which I reproduce here:

```
  /// Measurement position calibration based on mixture density network
  /// (MDN) model. The model takes as input:
  ///
  /// - A 7x7 charge matrix centered on the center pixel of the cluster;
  /// - The volume and layer identifiers from
  ///   the GeometryIdentifier of the containing surface;
  /// - The bound phi and theta angles of the predicted track state;
  /// - The initial estimated position
  /// - The initial estimated variance
  ///
  /// Given these inputs, a mixture density network estimates
  /// the parameters of a gaussian mixture model:
  ///
  /// P(Y|X) = \sum_i P(Prior_i) N(Y|Mean_i(X), Variance_i(X))
  ///
  /// These are translated to single position + variance estimate by
  /// taking the most probable value based on the estimated priors.
  /// The measurements are assumed to be 2-dimensional.
  ///
  /// This class implements the MeasurementCalibrator interface, and
  /// therefore internally computes the network input and runs the
  /// inference engine itself.
```
Few other things to note:

- This PR does not include a reference .onnx model; I plan to publish that for the Generic detector at least in a standalone repository along with the training script and the dataset creation code.
- At the moment, this is only implemented for the Generic detector's pixel volume. The calibrator class right now is configured with an allow-list of volume ids and a fall back PassThroughCalibrator. In the future, the calibrators should probably be setup as a GeometryHierarchyMap<MeasurementCalibrator>, but since this is a wider change (it will also make sense for the ScalingCalibrator, for instance) I propose to leave that to a subsequent PR. This will also be an opportunity to cleanup the implemementation of the calibrator configuration on the python side.
- This PR depends on #2171


Examples/Framework/src/EventData/NeuralCalibrator.cpp
```diff
@@ -0,0 +1,143 @@
...
+  m_scales = *scales;
+}
+
+void ActsExamples::NeuralCalibrator::calibrate(
+    const MeasurementContainer& measurements, const ClusterContainer* clusters,
+    const Acts::GeometryContext& /*gctx*/,
+    Acts::MultiTrajectory<Acts::VectorMultiTrajectory>::TrackStateProxy&
+        trackState) const {
+  const IndexSourceLink& sourceLink =
+      trackState.getUncalibratedSourceLink().get<IndexSourceLink>();
```
gagnonlg, line 83: Note to self: a bad_any_cast happens here

gagnonlg: Now fixed

Examples/Framework/CMakeLists.txt
```diff
@@ -34,6 +38,11 @@ target_compile_definitions(
   ActsExamplesFramework
   PRIVATE BOOST_FILESYSTEM_NO_DEPRECATED)
 
+if (ACTS_BUILD_PLUGIN_ONNX)
+  target_sources(ActsExamplesFramework PRIVATE src/EventData/NeuralCalibrator.cpp)
+  target_link_libraries(ActsExamplesFramework PRIVATE ActsPluginOnnx)
+endif()
+
```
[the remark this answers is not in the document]
Corentin-Allaire: `Framework/ML ` might be a much better idea, I agree. I don't have any strong feeling about this but I think it is something we should discuss now instead of later.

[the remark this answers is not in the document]
Corentin-Allaire: Good idea ! I will still review the rest of the PR in the mean time

Examples/Framework/src/EventData/NeuralCalibrator.cpp
```diff
@@ -0,0 +1,141 @@
...
+    acc0 += cell.bin[0] * cell.activation;
+    acc1 += cell.bin[1] * cell.activation;
+    tot += cell.activation;
+  }
+  int icntr0 = static_cast<int>(acc0 / tot);
+  int icntr1 = static_cast<int>(acc1 / tot);
+
+  // By convention, put centroid in (3, 3) cell
+  int diff0 = icntr0 - 3;
+  int diff1 = icntr1 - 3;
```
Corentin-Allaire, line 31: I am probably confused here but I though

Examples/Framework/src/EventData/NeuralCalibrator.cpp
```diff
@@ -0,0 +1,141 @@
...
+          throw std::runtime_error("Expected input size of " +
+                                   std::to_string(n_inputs) +
+                                   ", got: " + std::to_string(input.size()));
+        }
+
+        std::vector<float> output = m_model.runONNXInference(input);
+        // Assuming 2-D measurements, the expected params structure is:
+        // [      0,    nComp[ --> priors
+        // [  nComp,  3*nComp[ --> means
+        // [3*nComp,  5*nComp[ --> variances
```
Corentin-Allaire, line 113: Are the bracket backward or am I missing something ?
```suggestion
        // [      0,    nComp] --> priors
        // [  nComp,  3*nComp] --> means
        // [3*nComp,  5*nComp] --> variances
```

Corentin-Allaire (commented):
Looks good to me ! One general comment is that being more explicit with the variables name would make this easier to follow. (for example even if it is minor I would write center instead of cntr, ect). But I don't want to be too annoying so feel free to ignore this.

Examples/Framework/src/EventData/NeuralCalibrator.cpp
gagnonlg: They aren't, it's because the leftmost value is excluded. E.g. if nComp == 3, then the indices for the priors are 0, 1, 2 aka [0, 3[

gagnonlg:
> Looks good to me ! One general comment is that being more explicit with the variables name would make this easier to follow. (for example even if it is minor I would write center instead of cntr, ect). But I don't want to be too annoying so feel free to ignore this.

Yeah I can do an editing pass this afternoon. I learned programming in the C world, where spartan naming is encouraged, so, it's just a bad habit :D

Corentin-Allaire:
Thanks for the changes ! Lets discuss in which directory this should be in the next Acts meeting, then I can approve this !

Examples/Framework/CMakeLists.txt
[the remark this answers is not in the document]
gagnonlg: Was this discussed at the last dev meeting, @Corentin-Allaire @benjaminhuth ?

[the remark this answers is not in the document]
gagnonlg: Done

gagnonlg:
@Corentin-Allaire the packaging change has been done. Can you confirm this can go in?

Corentin-Allaire (approved):
Thanks for creating the new directory. And sorry for the delay...

gagnonlg:
Let's hold the merge -- There's a subtle issue with the charge matrix computation. I have a fix in place in the training code, so the computation needs to be compatible on this side. Will push the fix here shortly and we can merge.

gagnonlg:
The "fix" is in. Wasn't really a bug -- there was a bit of numerical instability when computing the offset from the cluster position on the matrix and the charge matrix. Basically to make the training easier, we want the centroid to be in the center of the charge matrix, to minimize the variance that the model has to capture. With the way it was previously done sometimes the centroid would land in the next-to-middle cell instead. I've made it more stable, which is good for the training but the definition has to be in sync with the one we perform at runtime. In the future, the code used to dump the training dataset could be checked-in the repo and we can have a single source for this computation.

TL;DR: we're good to go

Corentin-Allaire (approved):
Great to hear this is ready to go ! Please don't forget to rebase.

Corentin-Allaire (approved):

gagnonlg:
> Great to hear this is ready to go ! Please don't forget to rebase.

Thanks! I've just rebased.

PR #2113, opened 2023-05-09
ci: Add basic cmd options to physmon

benjaminhuth:
In order to debug physmon issues, I rarely need to run the whole suite, which is very time consuming. This adds a basic cmd interface to select certain tasks (not to fine-grained). It's a bit ugly, but it's bash...


paulgessinger:
Could we not split up the scripts maybe? I'll be looking at adding ttbar jobs soon anyway, where I'll probably have to do something like that anyway.

benjaminhuth:
You mean like having something like: 
```
physmon_truth_tracking.py
physmon_gsf.py
...
```

Yeah would be event better I think. I thought I go for the "least-impact" version, but a larger restructuring would be even better.
Maybe we could also get rid of the `bash` code at this point?

paulgessinger:
The `bash` script does copying and runs the plotting if needed, so I'd probably keep that in some form.

benjaminhuth:
I agree, but actually I would like to merge this in this state to have something like this in the repo at all...

andiwand (approved):
if this makes your workflow easier I think it is good enough for now

I wonder if we could merge the bash and python into a single python thing in the future

PR #2119, opened 2023-05-10
feat: Add full cov and pulls to VertexPerformanceWriter

andiwand:
since people are looking more closely into vertex performance right now this might be helpful


andiwand:
cc @vmmcairo @pbutti

paulgessinger:
This probably needs updates in the associated plotting script so it doesn't break physmon.

CI/physmon/physmon.py
```diff
@@ -344,39 +344,37 @@ def run_vertexing(fitter, mu, events):
         )
 
 
-with acts.FpeMonitor():
```
paulgessinger, line 347: Why is this removing the FPE monitoring?

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -65,154 +65,146 @@
 field = acts.ConstantBField(acts.Vector3(0.0, 0.0, 2.0 * u.T))
 rnd = acts.examples.RandomNumbers(seed=42)
 
-# TODO Geant4 currently crashes with FPE monitoring
-with acts.FpeMonitor() if not g4_simulation else contextlib.nullcontext():
```
paulgessinger, line 69: Why does this remove the FPE monitoring?

paulgessinger:
Can we keep the FPE monitoring? Also seems unrelated to what the title says this PR does?

andiwand:
> Can we keep the FPE monitoring? Also seems unrelated to what the title says this PR does?

the CI is crashing because of the FPE and I also face FPEs locally. It seems to highly depend on the rng at the moment and since we have no way of masking them and I have no time of chasing them all I dont see any other option

benjaminhuth (approved):

benjaminhuth (approved):

benjaminhuth (approved):

PR #2121, opened 2023-05-11
ci: Add AMVF to physmon

andiwand:
AMVF is not monitored in the CI at the moment. This PR adds all the checks we have for IVF also to AMVF


andiwand:
@paulgessinger the AMVF files are kinda big... maybe we should also use the hist root files for the vertexing?

CI/physmon/phys_perf_mon.sh
```diff
@@ -70,20 +70,35 @@ function full_chain() {
...
     Examples/Scripts/generic_plotter.py \
-        $outdir/performance_vertexing_${suffix}.root \
-        vertexing \
-        $outdir/performance_vertexing_${suffix}_hist.root \
+        $outdir/performance_ivf_${suffix}.root \
+        ivf \
+        $outdir/performance_ivf_${suffix}_hist.root \
         --silent \
-        --config CI/physmon/vertexing_config.yml
+        --config CI/physmon/ivf_config.yml
```
paulgessinger, line 77: Config could be common I guess?

paulgessinger:
Yes ideally we should only keep the histogram files, and I thought that's what we do already for the IVF?

CI/physmon/phys_perf_mon.sh
andiwand:
```suggestion
        --config CI/physmon/vertexing_config.yml
```

andiwand:
> Yes ideally we should only keep the histogram files, and I thought that's what we do already for the IVF?

you are right. I just messed up

benjaminhuth (approved):

PR #2122, opened 2023-05-11
fix: add protection of the matrix inverse calculations to avoid FPE

hrzhao76:


Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -703,7 +703,11 @@ inline void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
               m_config.sigmaScattering;
           p2scatterSigma = pTscatterSigma * pTscatterSigma * iSinTheta2;
         } else {
-          float pT = options.pTPerHelixRadius * std::sqrt(S2 / B2) / 2.;
+          float pT = 0.;
+          if (B2 != 0)
+            pT = options.pTPerHelixRadius * std::sqrt(S2 / B2) / 2.;
+          else
+            pT = options.pTPerHelixRadius * std::sqrt(S2 / 0.000000001) / 2.;
```
andiwand, line 710:
```suggestion
          if (B2 != 0) {
            pT = options.pTPerHelixRadius * std::sqrt(S2 / B2) / 2.;
          } else {
            pT = options.pTPerHelixRadius * std::sqrt(S2 / 0.000000001) / 2.;
          }
```

Core/include/Acts/Vertexing/AdaptiveMultiVertexFitter.ipp
```diff
@@ -321,8 +321,15 @@ bool Acts::AdaptiveMultiVertexFitter<
   for (auto vtx : state.vertexCollection) {
     Vector3 diff = state.vtxInfoMap[vtx].oldPosition.template head<3>() -
                    vtx->fullPosition().template head<3>();
-    SymMatrix3 vtxWgt =
-        (vtx->fullCovariance().template block<3, 3>(0, 0)).inverse();
+    ActsMatrix<3, 3> tmp_inverse = ActsMatrix<3, 3>::Zero();
+    bool is_invertible_vtx_cov = false;
```
andiwand, line 325:
```suggestion
    bool isInvertibleVtxCov = false;
```

Core/include/Acts/Vertexing/KalmanVertexUpdater.ipp
```diff
@@ -74,32 +74,60 @@ void Acts::KalmanVertexUpdater::updatePosition(
   const ActsSymMatrix<5> trkParamWeight =
       linTrack.weightAtPCA.block<5, 5>(0, 0);
 
+  ActsMatrix<3, 3> tmp_inverse = ActsMatrix<3, 3>::Zero();
```
andiwand, line 77:
```suggestion
  ActsMatrix<3, 3> tmpInverse = ActsMatrix<3, 3>::Zero();
```

Core/include/Acts/Vertexing/KalmanVertexUpdater.ipp
```diff
@@ -74,32 +74,60 @@ void Acts::KalmanVertexUpdater::updatePosition(
...
+  long double threshold = Eigen::NumTraits<long double>::dummy_precision();
+  // Determant threshold is Eigen::NumTraits<double>::dummy_precision() 1e-12
+  // Which is used in computeInverseWithCheck
+
+  bool is_invertible_vtx_cov = false;
+  bool is_invertible_vtx_transpose_weight_Jac = false;
+  bool is_invertible_new_vtx_cov = false;
```
andiwand, line 84:
```suggestion
  bool isInvertibleVtxCov = false;
  bool isInvertibleVtxTransposeWeightJac = false;
  bool isInvertibleNewVtxCov = false;
```

Core/include/Acts/Vertexing/AdaptiveMultiVertexFitter.ipp
```diff
@@ -321,8 +321,15 @@ bool Acts::AdaptiveMultiVertexFitter<
...
+    SymMatrix3 vtxWgt = SymMatrix3::Zero();
```
andiwand, line 326: what happens if we continue with a zero covariance?

also shouldn't we log an error here?

Core/include/Acts/Vertexing/KalmanVertexUpdater.ipp
```diff
@@ -74,32 +74,60 @@ void Acts::KalmanVertexUpdater::updatePosition(
...
+
   // Vertex to be updated
   const Vector3& oldVtxPos = vtx.position();
-  matrixCache.oldVertexWeight = (vtx.covariance()).inverse();
+  (vtx.covariance())
+      .computeInverseWithCheck(tmp_inverse, is_invertible_vtx_cov, threshold);
+  if (is_invertible_vtx_cov) {
```
andiwand, line 90: same question as above: what will happen if the inverse does not work?

Core/include/Acts/Seeding/SeedFinder.ipp
LuisFelipeCoelho, line 710: I think this was fixed already in PR #2107, right?

andiwand:
I think it should not be necessary to surround the `.inverse()` call with this sort of prevention mechanism. I could imagine that Eigen would try different inversion methods depending on the result being `nan` or not. Maybe not I don't know but I don't think we should try to handle FPEs that pop up in thirdparty code. Either we have to call a specific inversion impl that is more stable or we just error out of the calculation if we get `nan` from Eigen. But in order to do this we need some kind of FPE masking

paulgessinger:
I don't know, what if numerically the input matrix has a problem? I'm not sure if it's Eigen's job to avoid FPEs in this case.

andiwand:
The thing is that I also tried to use a different algorithm for the inversion with Eigen which then worked. The one I tried was full LU decomposition. So I wondered what `.inverse()` will do under the hood and if it might do trial and error

andiwand:
@paulgessinger do you think this is still necessary after the FPE signal improvements?

I think we could just add a check after the inverse and error out if something goes wrong. the signal will tell use where it happened but execution can continue with other vertices/events

paulgessinger:
I still believe we need to prevent the FPE from even triggering, actually.

andiwand:
I dont see the difference. either the computations fails through an FPE or more softly through Eigen. in both cases you have to handle it by at least skipping the vertex and logging an error. but handling it trough Eigen adds a lot of not so nice code IMO

paulgessinger:
FPE means Atlas fails the job during log file scanning.

andiwand:
> * If we say "let's check the **result of a computation if an FPE occurred** during the calculation and not continue then", we can't verify that we do this, because the FPE signal will have happened before that

Why can't we verify this? I see the FPE signal purely as a monitoring step. When an FPE occurs you have to log an error anyway which will fail the Atlas log scan.

But further processing can and should continue to get at least some results which tells you if that was a rare case or not.

> * If we say "let's check if the computation **will throw an FPE** before running it" we prevent the FPE from happening in the first place, which we can verify is the case

That sounds like unnecessary runtime overhead to me and will not solve any problem (as far as I can see) as you have to log an error anyways.

In principle that would mean that we surround every `inverse` operation in the codebase because it could potentially throw an FPE.

paulgessinger:
We can't verify it automatically, I mean.

The processing does continue in this way, and for example the vertexing can internally count how many failure cases like this occurred and take action based on the number.

> In principle that would mean that we surround every `inverse` operation in the codebase because it could potentially throw an FPE.

Yup. I don't see a way around it though, because FPE monitoring is happening in Athena, and if we don't do checks, we'll have to argue that it's fine if we generate 5k FPEs in run 393854 in 2022.

andiwand:
After offline discussion: I created a PR with a safe inverse helper function https://github.com/acts-project/acts/pull/2151 . After merging it would be good to Incorporate this here

Core/include/Acts/Vertexing/KalmanVertexUpdater.ipp
andiwand, line 1: here we would end up with silent FPE suppression and try to continue the math. that would be hard to debug if that results in weird numbers.

@paulgessinger I think we would need to wire a second return path here but that sounds like quite a few changes. opinion?

paulgessinger:
Closing this for now. We can reopen if this becomes relevant again.

PR #2124, opened 2023-05-11
ci: Split up physmon script into multiple ones

paulgessinger:
This PR splits up the physmon jobs into separate ones which can be run on their own.
The main `phys_perf_mon.sh` script is still used to execute them and collect the results together.

Long term we'll probably want to make physmon more extensible and modular once we potentially start adding experiment pipelines to our checks.


benjaminhuth:
So you would remove the option to selectively run pipelines + the corresponding plot generation again?

Because that is for me kind of an important feature, so if I want to know if I fixed e.g. the GSF pipeline, I don't want to run all workflows.

paulgessinger:
And just commenting out the other parts is not an option?

benjaminhuth:
Yeah me as well, and due to the python-scripts this gets easier indeed. I just wonder what speaks against a lightweight command line...

paulgessinger:
We'll want to extend this to more workflows. I'm worried about scalability and complexity in the configuration as in "what runs" that's basically it

andiwand (approved):
looks good to me

PR #2125, opened 2023-05-12
ci: Summary for physmon

andiwand:
I find myself clicking through a bunch of physmon htmls lately and thought a summary would be nice to easily spot the reference that actually changed


paulgessinger:
Where does this show up, actually? 
I've been thinking about this, but would be even nicer if it could link to the herald page. Unfortunately you can't seem to get the artifact id to build the URL, so that probably won't work.

Still nice to have though I guess.

andiwand:
my ideas was to throw this into the zip and since it has relative links it should work locally and on herald

andiwand:
it actually worked but it looks like there is a problem with the encoding https://herald.harbinger.gessinger.dev/view/acts-project/acts/693648332/summary.html

benjaminhuth (approved):

PR #2127, opened 2023-05-14
fix: Remove det(cov)<=0 check from HelicalTrackLinearizer

andiwand:
IMO this check should be removed as `det(cov)<=0` should never be true anyways and replacing it with a different cov will not improve things. I logged into this using `full_chain_odd.py` and it looks like `det(cov)<=0` will fire a bunch of times because of numerical issues and the backup cov will also have `det(cov)<=0`


pbutti (approved):
looks good to me.

pbutti (approved):

andiwand:
I had to update the reference @pbutti - could you reapprove?

benjaminhuth (approved):

PR #2129, opened 2023-05-15
fix: Protection against wrong configuration of Seed finder config

CarloVarni:
In case a misconfiguration of the `SeedFinderConfig`  happens, and the `seedFilter` is a nullptr, the code crashes.

This PR adds a protection against this eventuality. This happens in the `toInternalUnits` function, that now takes responsibility for checking the validity of its parameters.

It checks that the seedFilter is not a `nullptr` and that its config is already in internal units.

### DIFF
--- a/Core/include/Acts/Seeding/SeedFinderConfig.hpp
+++ b/Core/include/Acts/Seeding/SeedFinderConfig.hpp
@@ -181,13 +181,15 @@ struct SeedFinderConfig {
     // And make sure the seed filter config is in internal units as well
     if (not seedFilter) {
       throw std::runtime_error(
-	   "Invalid values for the seed filter inside the seed filter config: nullptr");
+          "Invalid values for the seed filter inside the seed filter config: "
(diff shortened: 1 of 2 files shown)


paulgessinger (approved):
Looks good to me. You were wondering if we ever want to actually configure the finder without the filter, I guess we don't? Thoughts @LuisFelipeCoelho ?

andiwand:
looks like our tests rely on this :thinking:

CarloVarni:
ok, then I'll get a look at the failures here and fix them. It looks like indeed we have situation where the filter is not set

CarloVarni:
I forgot in the precious commit to propagate the changes in the SeedingAlgorithm example to this MR. Now it all works

andiwand (approved):
great, thank you!

paulgessinger (approved):
This whole "internal units" mechanism is a bit of a crux that we should be looking to remove. That's independent of this change though.

PR #2132, opened 2023-05-17
fix: Correct the pTperHelixRadius calculation in SeedFinderConfig

beomki-yeo:
To synchronize with acts-project/traccc#403

The right equation is rho [mm] / pT [MeV] = 0.0033356 / B[kT] <=>  pT[MeV] / rho[mm] = 299.796 * B[kT], so the factor of 300 is wrong

Reference: https://en.wikipedia.org/wiki/Rigidity_(electromagnetism)


Core/include/Acts/Seeding/SeedFinderConfig.hpp
```diff
@@ -262,7 +262,7 @@ struct SeedFinderOptions {
     // helix radius in homogeneous magnetic field. Units are Kilotesla, MeV and
     // millimeter
     // TODO: change using ACTS units
-    options.pTPerHelixRadius = 1./0.003356 * options.bFieldInZ;
+    options.pTPerHelixRadius = 1. / 0.0033356 * options.bFieldInZ;
```
andiwand, line 265: might be worth using `Units.hpp`

stephenswat: Why would we replace a constant (which I assume is supposed to be the SOL) by one over the reciprocal of that constant? If there a good reason to do this?

andiwand: so I think with `Units.hpp` it would be
```suggestion
    options.pTPerHelixRadius = UnitConstants::T * 1e-6 * options.bFieldInZ;
```
the conversion for kT and MeV looks magical but I guess this could only be improved by changing this all across the seeding code

beomki-yeo: It is the SOL. yeah I think `UnitConstants::T * 1e6` is also OK

LuisFelipeCoelho (changes requested):
Athena uses the 300 factor, so to keep the same performance you also need to change the magnetic field in the ITk configuration. Or we can discuss changing that value in Athena as well

LuisFelipeCoelho:
FYI @noemina

LuisFelipeCoelho:
Also @beomki-yeo could you change the name of this PR to make it more informative?

beomki-yeo:
In case people want to see the physmon result

https://drive.google.com/drive/folders/1ngk3lpti-lYo1PeJFRkaGyTPJqVenLVB?usp=sharing

paulgessinger:
What's the status of this? Can we move ahead?

LuisFelipeCoelho (approved):

paulgessinger:
Not sure why the comment is missing. This is handled by a secondary github actions that's triggered separately, maybe there was a problem?

beomki-yeo:
Is it triggered only when `linux_physmon` fails? Then it makes sense

paulgessinger:
Should always be triggered in principle.

beomki-yeo:
Well I can still generate the files by myself hopefully

andiwand:
> Well I can still generate the files by myself hopefully

You can also download them from the job artifact even if the physmon post is not showing up

andiwand (approved):

PR #2133, opened 2023-05-17
feat: Option to keep only particles with hits in Geant4

benjaminhuth:
Introduces a new option (off by default) to the geant4 simulation that allows to only keep particles that also generated hits.

This is so far not solvable by particles selectors, and also most efficiently implemented directly in the simulation.

I added this mostly because of storage reasons when writing out to file, this reduces the number of particles from O(10^6) to O(10^4) for me.


Examples/Algorithms/Geant4/src/ParticleTrackingAction.cpp
```diff
@@ -59,8 +59,20 @@ void ActsExamples::ParticleTrackingAction::PostUserTrackingAction(
...
+  const auto barcode = eventData.trackIdMapping.at(aTrack->GetTrackID());
+
+  auto hasHits = eventData.particleHitCount.find(barcode) !=
+                     eventData.particleHitCount.end() and
+                 eventData.particleHitCount.at(barcode) > 0;
+
+  if (not m_cfg.keepParticlesWithoutHits and not hasHits) {
+    auto n = eventData.particlesInitial.erase(
+        ActsExamples::SimParticle{barcode, Acts::PdgParticle::eInvalid});
+    assert(n == 1);
```
andiwand, line 71: should we use an error log here?

Examples/Python/python/acts/examples/simulation.py
```diff
@@ -572,10 +572,12 @@ def addGeant4(
     preSelectParticles: Optional[ParticleSelectorConfig] = ParticleSelectorConfig(),
     postSelectParticles: Optional[ParticleSelectorConfig] = None,
     recordHitsOfSecondaries=True,
+    keepParticlesWithoutHits=True,
     outputDirCsv: Optional[Union[Path, str]] = None,
     outputDirRoot: Optional[Union[Path, str]] = None,
     logLevel: Optional[acts.logging.Level] = None,
     killVolume: Optional[acts.Volume] = None,
+    keepParticlesWithoutHits=True,
```
andiwand, line 580: duplicate

Examples/Algorithms/Geant4/src/ParticleTrackingAction.cpp
andiwand: the problem I see with this is that in release build without our assert workaround this error case will be silent while the `ACTS_ERROR` will at least show up in the log

benjaminhuth: But thats the thing with all `assert` statements...

andiwand: that is true but I mostly see this in a place where a logger is not available or might be expensive to use

but up to you

andiwand (approved):

Examples/Algorithms/Geant4/src/ParticleTrackingAction.cpp
benjaminhuth: I actually would leave it with the `assert`.

PR #2134, opened 2023-05-17
refactor: Remove some "code smells" in the clusterization code

gagnonlg:
At the ATLAS tracking workshop, it was brought up that ACTS has some "code smells" (defined as whatever is flagged by the SonarCloud tool). To get a feel for the signal-to-noise ratio of this analysis, I've had a look at the output for Clusterization.hpp/ipp and implemented some of the suggestions (not all!). There were some good ideas that did improve the readability ("using" declarations to inherit constructors, "if constexpr" instead of enable_if hacks) and some that I though led to less readable code (e.g. the infamous "1 break per loop" misra rule) that I disregarded. And a bunch of more trivial changes.

Link to the full report: https://sonarcloud.io/summary/new_code?id=Yusuf-Manjra_acts


andiwand:
could you add a link to the report for reference?

gagnonlg:
> could you add a link to the report for reference?

Sure thing! It's now in the description.

paulgessinger (commented):
I guess with the exception on whether or not to factor out code into functions, the changes are uncontroversial.

gagnonlg:
> I guess with the exception on whether or not to factor out code into functions, the changes are uncontroversial.

Yeah I don't think it should be blindly followed, but in the specific case I applied it here I think it does help

Core/include/Acts/Clusterization/Clusterization.hpp
```diff
@@ -11,8 +11,7 @@
 #include <memory>
 #include <vector>
 
-namespace Acts {
-namespace Ccl {
+namespace Acts::Ccl {
```
andiwand, line 14: not sure if this is a guideline but we only use nested namespaces in Acts as far as I could see

paulgessinger: I see them as equivalent, with a slight preference towards concatenating. I think we have this in the codebase here and there already.

PR #2135, opened 2023-05-18
refactor: move initialisation of vector in seedFilter

LuisFelipeCoelho:
This PR moves the initialisation of `compatibleSeedR` in seedFilter to reduce memory allocation

### DIFF
--- a/Core/include/Acts/Seeding/SeedFilter.ipp
+++ b/Core/include/Acts/Seeding/SeedFilter.ipp
@@ -73,13 +73,16 @@ void SeedFilter<external_spacepoint_t>::filterSeeds_2SpFixed(
               });
   }
 
+  // -> weaker requirement for a good seed
+  std::vector<float> compatibleSeedR;
+  compatibleSeedR.reserve(m_cfg.compatSeedLimit);
+
   size_t beginCompTopIndex = 0;
   // loop over top SPs and other compatible top SP candidates
   for (const std::size_t topSPIndex : topSPIndexVec) {
     // if two compatible seeds with high distance in r are found, compatible
     // seeds span 5 layers
-    // -> weaker requirement for a good seed
-    std::vector<float> compatibleSeedR;
+    compatibleSeedR.clear();
 
     float invHelixDiameter = invHelixDiameterVec[topSPIndex];
     float lowerLimitCurv = invHelixDiameter - m_cfg.deltaInvHelixDiameter;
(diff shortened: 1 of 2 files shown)


Examples/Python/python/acts/examples/itk.py
```diff
@@ -434,7 +434,7 @@ def itkSeedingAlgConfig(inputSpacePointsType: InputSpacePointsType):
         seedConfirmationFilter = True
         impactWeightFactor = 100
         compatSeedLimit = 3
-        numSeedIncrement = float("inf")
+        numSeedIncrement = 100
```
CarloVarni, line 437: is this intended? why are we changing it?

Core/include/Acts/Seeding/SeedFilter.ipp
```diff
@@ -73,14 +73,16 @@ void SeedFilter<external_spacepoint_t>::filterSeeds_2SpFixed(
               });
   }
 
+  // vector containing the radius of all compatible seeds
+  std::vector<float> compatibleSeedR;
+  compatibleSeedR.reserve(m_cfg.compatSeedLimit);
```
CarloVarni, line 78: what is the usual value for `m_cfg.compatSeedLimit`? Can we set a reasonable (i.e. 0) default value for it in order not to allocate too much memory (or worse having a `bad_alloc ` thrown at us)?

Examples/Python/python/acts/examples/itk.py
LuisFelipeCoelho: we never reach 100 compatible seeds for PPP at mu200 so I think this doesn't necessarily need to be set to infinity, but I can revert to `float("inf")` if you prefer

Core/include/Acts/Seeding/SeedFilter.ipp
LuisFelipeCoelho: the default value of `compatSeedLimit` is 2. For PPP it is 4 and for SSS it is 3

andiwand:
did you reach a conclusion on this one @LuisFelipeCoelho @CarloVarni ?

Examples/Python/python/acts/examples/itk.py
CarloVarni: better lower numbers for the memory, right?

Core/include/Acts/Seeding/SeedFilter.ipp
CarloVarni: ok with this then

CarloVarni:
@LuisFelipeCoelho @andiwand I'm ok with this PR, but I cannot resolve the discussion... looks like github is not allowing the one opening the discussion to close it ...

CarloVarni (approved):

PR #2138, opened 2023-05-19
fix: Fix physmon comment template after #2130

andiwand:
another problem I try to fix after #2130


benjaminhuth (approved):
Let's hope it works!

andiwand:
verified it locally this time

PR #2139, opened 2023-05-19
feat: Muon mockup - indexed straw surfaces

dimitra97:
This PR uses the indexed surfaces infrastructure into the mock-up builder geometry.
Some things that changed in the mock-up builder class:
-Detector Volumes are also created for the two multi-layers of each chamber, with the `MultiWireLayer` updator.

The `MultiWireLayerUpdator` created to consider the indexed surfaces of the grid in order to update the navigation state with the optimal number of surface candidates for intersection. 
The Unit test checks if we get the correct number of surface candidates in the navigation state when we move in the multi-layer chamber

There is still some work on it, since I am not sure about the way I get the bin from the position of the navigation state (e.g about the local transformation). And any other recommendations would be helpful :)


Core/include/Acts/Navigation/MultiWireLayerUpdators.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+      auto dy = binWidth[1];
+      auto phi = Acts::VectorHelpers::phi(tdirection);
+
+      BinUtility binUtility2D(nBins[0], minPos[0], maxPos[0], open, binX,
+                              transform);
+      binUtility2D +=
+          BinUtility(nBins[1], minPos[1], maxPos[1], open, binY, transform);
+
+      for (std::size_t i = 0; i < nBins[1]; i++) {
```
paulgessinger, line 146: This seems expensive to do on every update, right?

dimitra97: > This seems expensive to do on every update, right?

Oh sorry, you mean the for loop? You are right, but I just needed to go through every layer of the grid

Examples/Detectors/MuonSpectrometerMockupDetector/include/ActsExamples/MuonSpectrometerMockupDetector/MockupSectorBuilder.hpp
```diff
@@ -46,6 +54,19 @@ class MockupSectorBuilder {
...
+    // getIndexedStraws(std::vector<std::shared_ptr<Acts::Surface>>);
+
+    // Function that returns the straws of the chamber of a specific chamber
+    std::vector<std::shared_ptr<Acts::Surface>> getStraws() {
+      return ChambersStraws;
+    };
+
+   private:
+    // void setStraws
+    std::vector<std::shared_ptr<Acts::Surface>> ChambersStraws;
```
andiwand, line 69:
```suggestion
    std::vector<std::shared_ptr<Acts::Surface>> m_chambersStraws;
```

andiwand: or remove it from `private` might be even better as this is a struct

Examples/Detectors/MuonSpectrometerMockupDetector/include/ActsExamples/MuonSpectrometerMockupDetector/MockupSectorBuilder.hpp
```diff
@@ -46,6 +54,19 @@ class MockupSectorBuilder {
 
     // The names of the passive surfaces
     std::vector<std::string> PassiveNames;
```
andiwand, line 56: I overlooked these - I think they should also be starting with a lower case

Examples/Detectors/MuonSpectrometerMockupDetector/include/ActsExamples/MuonSpectrometerMockupDetector/MockupSectorBuilder.hpp
```diff
@@ -24,9 +25,13 @@ namespace ActsExamples {
 /// This is a class that builds a chamber by reading a gdml file
 
 class MockupSectorBuilder {
+  friend struct ChamberConfig;
+
  public:
-  /// Nested configuration struct
+  // Nested configuration struct
```
andiwand, line 31: `///` will instruct doxygen to pick up the comment so might be preferred

Core/include/Acts/Navigation/MultiWireLayerUpdators.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+inline static Acts::Experimental::SurfaceCandidatesUpdator MultiWireLayer() {
+  auto ap = std::make_unique<const MultiWireLayerImpl>();
+  SurfaceCandidatesUpdator nStateUpdator;
+  nStateUpdator.connect<&MultiWireLayerImpl::update>(std::move(ap));
+  return nStateUpdator;
+}
+
+}  // namespace Experimental
+}  // namespace Acts
```
andiwand, line 199:
```suggestion
}  // namespace Acts

```

Core/include/Acts/Navigation/MultiWireLayerUpdators.hpp
```diff
@@ -0,0 +1,199 @@
...
+#include "Acts/Navigation/NavigationState.hpp"
+#include "Acts/Navigation/NavigationStateFillers.hpp"
+#include "Acts/Navigation/NavigationStateUpdators.hpp"
+#include "Acts/Navigation/SurfaceCandidatesUpdators.hpp"
+#include "Acts/Surfaces/Surface.hpp"
+#include "Acts/Utilities/detail/Axis.hpp"
+#include "Acts/Utilities/detail/Grid.hpp"
+
+#include <memory>
+#include <tuple>
```
noemina, line 29:
```suggestion
#include <tuple>
#include <vector>
#include <pair>
#include <array>
```

Core/include/Acts/Navigation/MultiWireLayerUpdators.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+struct MultiWireLayerImpl : public INavigationDelegate {
+  IndexedSurfacesGridType getIndexedSurfacesFromVolume(
+      NavigationState& nState) const {
+    auto gridSurfaces = nState.currentVolume->surfaces();
+    GeometryContext gctx;
+
+    std::array<std::pair<float, float>, 3> min_max_multilayer;
+    std::fill(min_max_multilayer.begin(), min_max_multilayer.end(),
+              std::make_pair<float, float>(std::numeric_limits<float>::max(),
```
noemina, line 46: I guess `numeric_limits` needs inclusion of `algorithms`?

Core/include/Acts/Navigation/MultiWireLayerUpdators.hpp
```diff
@@ -0,0 +1,199 @@
...
+          min_max_multilayer[1].second, (float)gridSurf->center(gctx).y());
+
+      min_max_multilayer[2].first = std::min(min_max_multilayer[2].first,
+                                             (float)gridSurf->center(gctx).z());
+      min_max_multilayer[2].second = std::max(
+          min_max_multilayer[2].second, (float)gridSurf->center(gctx).z());
+    }
+    auto radius = gridSurfaces.front()->bounds().values()[0];
+
+    float binWidthX = 2 * radius, binWidthY = sqrt(3) * radius;
```
noemina, line 67: Can we make these values being configurable? I have hard time justifying `sqrt(3)` 
Maybe the default value can be the number width / number of surfaces in the two directions?

Core/include/Acts/Navigation/MultiWireLayerUpdators.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+    float rangeX1 = min_max_multilayer[2].first - radius,
+          rangeX2 = min_max_multilayer[2].second + radius;
+    float rangeY1 = min_max_multilayer[1].first - radius,
+          rangeY2 = min_max_multilayer[1].second + radius;
+
+    std::size_t nBinsX =
+                    static_cast<std::size_t>((rangeX2 - rangeX1) / binWidthX),
+                nBinsY =
+                    static_cast<std::size_t>((rangeY2 - rangeY1) / binWidthY);
```
noemina, line 77: Other possibility in that we make the number of bins customisable and evaluate the width from this.

Core/include/Acts/Navigation/MultiWireLayerUpdators.hpp
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    updateCandidates(gctx, nState);
+  }
+
+ private:
+  float tolerance = 0.5;
+};
+
+inline static Acts::Experimental::SurfaceCandidatesUpdator MultiWireLayer() {
```
noemina, line 191: I think I would put `SurfaceUpdator` or something else in the name of this function. The name, as is, is a bit misleading. Same for `MultiWireLayerImpl`

Examples/Detectors/MuonSpectrometerMockupDetector/src/MockupSectorBuilder.cpp
```diff
@@ -301,14 +356,77 @@ ActsExamples::MockupSectorBuilder::buildSector(
...
+    min_max[1].first =
+        std::min(min_max[1].first, (float)surf->center(context).y());
+    min_max[1].second =
+        std::max(min_max[1].second, (float)surf->center(context).y());
+
+    min_max[2].first =
+        std::min(min_max[2].first, (float)surf->center(context).z());
+    min_max[2].second =
+        std::max(min_max[2].second, (float)surf->center(context).z());
+  }
```
noemina, line 394: All this calculation is what is done already in the construction of the `MultiWireLayerImpl`. Is there a way to pass some of the evaluated quantities, or create here the grid that we can pass to the `NavigationDelegate`?

dimitra97: > All this calculation is what is done already in the construction of the `MultiWireLayerImpl`. Is there a way to pass some of the evaluated quantities, or create here the grid that we can pass to the `NavigationDelegate`?

Yes. Actually now it came to my mind that I don't need to do all the calculations for the grid from the surfaces. I can use the volume's bounds and position in order to define the range for the both directions for the grid. And for the number of bins I can do what you suggested, because now it is fixed to the distance of the centers of the surfaces (that's why the sqrt(3) coefficient) and the diameter. But I will try to create the grid here and pass it to the navigation delegate, as you suggest.

PR #2140, opened 2023-05-23
fix: Fix stepper charge hypothesis

andiwand:
I found at least one case where we rely on the sign of the charge hypothesis in the stepper. The charge might flip during KF but the one inside the stepper would stay the same causing inconsistencies and in my case a terrible fit.

In this PR I remove the sign of the charge hypothesis, refactored the `momentum` and `charge` accessor methods and added a `qop` accessor method that should be preferred over explicit `charge / momentum` IMO


Core/include/Acts/Propagator/EigenStepper.hpp
```diff
@@ -213,17 +213,24 @@ class EigenStepper {
...
+  ///
+  /// @param state [in] The stepping state (thread-local cache)
+  double qop(const State& state) const { return state.pars[eFreeQOverP]; }
+
   /// Absolute momentum accessor
   ///
   /// @param state [in] The stepping state (thread-local cache)
   double momentum(const State& state) const {
-    return std::abs((state.q == 0. ? 1. : state.q) / state.pars[eFreeQOverP]);
+    return std::abs(charge(state) / qop(state));
```
paulgessinger, line 225: What happens if `q==0`?

Core/include/Acts/Propagator/EigenStepper.hpp
```diff
@@ -213,17 +213,24 @@ class EigenStepper {
...
   }
 
   /// Charge access
   ///
   /// @param state [in] The stepping state (thread-local cache)
-  double charge(const State& state) const { return state.q; }
+  double charge(const State& state) const {
+    return std::copysign(state.absCharge, qop(state));
```
paulgessinger, line 232: Charge can't be zero this way can it?

paulgessinger: Ah, theres `+0` and `-0`, right.

Core/include/Acts/Propagator/EigenStepper.hpp
paulgessinger: Charge will be `+-0`, but you take `abs`, so it's `|p|`, which is all good.

andiwand: oh yeah I have to catch the q=0 case

Core/include/Acts/Propagator/EigenStepper.ipp
```diff
@@ -84,12 +84,12 @@ void Acts::EigenStepper<E, A>::update(State& state,
 
 template <typename E, typename A>
 void Acts::EigenStepper<E, A>::update(State& state, const Vector3& uposition,
-                                      const Vector3& udirection, double up,
+                                      const Vector3& udirection, double qop,
```
paulgessinger, line 87: I'm guessing you checked all uses of this function to make sure we're passing in the right value?

Core/include/Acts/Propagator/MultiEigenStepperLoop.hpp
```diff
@@ -63,23 +63,33 @@ struct WeightedComponentReducerLoop {
...
   }
 
+  template <typename stepper_state_t>
+  static ActsScalar qop(const stepper_state_t& s) {
+    return std::accumulate(
+        s.components.begin(), s.components.end(), ActsScalar{0.},
+        [](const auto& sum, const auto& cmp) -> ActsScalar {
+          return sum + cmp.weight * cmp.state.pars[eFreeQOverP];
+        });
+  }
```
paulgessinger, line 73: Should this be cached and updated when changes happen? (/cc @benjaminhuth)

Core/include/Acts/Propagator/MultiEigenStepperLoop.hpp
```diff
@@ -63,23 +63,33 @@ struct WeightedComponentReducerLoop {
     return toVector3(s.components, eFreeDir0).normalized();
   }
 
+  template <typename stepper_state_t>
```
benjaminhuth, line 66:
```suggestion
  // TODO: Maybe we can cache this value and only update it when the parameters change
  template <typename stepper_state_t>
```

Core/include/Acts/Propagator/EigenStepper.ipp
andiwand: I checked again and it seems to be only used in `PointwiseMaterialInteraction` and `SimulationActor`

paulgessinger: Don't we use it in the CKF too?

andiwand: only through `PointwiseMaterialInteraction`. the `CKF` is using the other overloads

paulgessinger: ok

asalzburger (approved):
That looks all consistent, good to go.

asalzburger (approved):
Re-approving

paulgessinger:
Just for reference, there are some discussions that this might have improved measurably the ACTS CKF performance as run in Athena (https://its.cern.ch/jira/browse/ATLIDTRKCP-472)

andiwand:
> Just for reference, there are some discussions that this might have improved measurably the ACTS CKF performance as run in Athena (https://its.cern.ch/jira/browse/ATLIDTRKCP-472)

Noice! I also saw improvements in our ODD full chain but it was mostly outliers

PR #2141, opened 2023-05-23
refactor: Testing HelicalTrackLinearizer by comparing its Jacobians to numerically computed derivatives

felix-russo:
Previously, the unit test for the HelicalTrackLinearizer only checked if its outputs were non-zero matrices. To improve the testing, a new class, NumericalTrackLinearizer, is introduced, which allows us to compute Jacobians numerically (i.e., we approximate the derivatives like f'(x) ~ (f(x + dx) - f(x)) / dx). The numerically computed Jacobians are then compared to the ones we obtain from the HelicalTrackLinearizer in the improved unit test.

Comments concerning code readability and style are highly appreciated!


Core/include/Acts/Vertexing/NumericalTrackLinearizer.ipp
```diff
@@ -0,0 +1,142 @@
...
+      Jacobian.block<eBoundSize, 3>(0, 4)};
+
+  // Constant term of Taylor expansion (Eq. 5.38 in Ref. (1))
+  BoundVector constTerm{perigeeParams - positionJacobian * pca -
+                        momentumJacobian * momentumAtPCA};
+
+  return LinearizedTrack(perigeeParams, parCovarianceAtPCA, weightAtPCA,
+                         linPoint, positionJacobian, momentumJacobian, pca,
+                         momentumAtPCA, constTerm);
+}
```
andiwand, line 142:
```suggestion
}

```

andiwand (commented):
I wonder how much we could use from `RiddersPropagator` - I guess it does basically do what you did here by hand?

beomki-yeo:
It might not have nothing to do with this PR, but there is also a way to validate `HelicalTrackLinearizer` with the analytical solution of a helix model. 

See
https://github.com/acts-project/detray/blob/51571670ea7b8e574db560c9dadf49401352583d/core/include/detray/intersection/detail/trajectories.hpp#L268

which is based on the helix model (Eq. 4.28 - 4.34) of https://doi.org/10.1007/978-3-030-65771-0

Core/include/Acts/Vertexing/NumericalTrackLinearizer.ipp
```diff
@@ -0,0 +1,143 @@
...
+    double qOvP{perigeeParams(BoundIndices::eBoundQOverP)};
+
+    curvilinearParamVec << globalCoords, globalTime, phi, theta, qOvP;
+    pca << globalCoords, globalTime;
+    momentumAtPCA << phi, theta, qOvP;
+  }
+
+  // Setting size of the perturbation delta for calculation of numerical
+  // derivatives (i.e., f'(x) ~ (f(x+delta) - f(x)) / delta)
+  double delta{1e-8};
```
beomki-yeo, line 80: I am trying to understand how the Linearizers are working so my question would be dumb.
isn't delta should be `intersection.intersection.pathLength` instead?

beomki-yeo: Or the linearizer is supposed to calculate the derivative at the fixed point rather than calculating the transport jacobian

andiwand: the delta is how much we change the initial parameters and then we propagate to the PCA again which results in different final parameters. afterwards @russofel determines the numerical derivatives by calculating `(final - initial) / delta`

using the transport jacobian should give a similar result I guess?

I think the job of the linearizer is just to provide a set of matrices that can be used in the fitter to linearly explore the parameter space in order to be able to update your prediction with the "measurement"

russofel: The only thing `intersection.intersection.pathLength` does is that it tells us in which direction to go when we propagate to a point!

Yes, we calculate the derivative at a fixed point (i.e., the PCA to a reference point in our case). 

Concerning the transport Jacobian: Bastian hints in his thesis that the transport Jacobian should correspond to our result if we let the path length between the two points go to 0 (see Fig. 5.7 in https://cds.cern.ch/record/2839800). I am not entirely sure how this works, but I will look into it.

beomki-yeo: OK. I looked into the figure and corresponding text - If it is just a jacobian transformation on a fixed point, the approach in Numerical linearizer seems correct.

Now I don't understand why HelicalTrackLinearizer is calculating two cases (b field and no b field) differently when we assume that there is no step length :p . The way I see is that the jacobian is simple a coordinate transformation from perigee to free parameter without transport jacobian

russofel: I think the Jacobian we have here is not just a coordinate transformation: It is the derivative of the Perigee parametrization with respect to the free parameters at the PCA. 

If we go to the PCA and slightly wiggle one of the track parameters there, the track will change and we will get a new PCA. The Jacobian now comprises the following information: "How do the parameters at the new PCA change compared to the old one?"

I hope this answers why we must treat the cases of no B-field and constant B-field differently: The new PCA will depend on the magnetic field - or the lack thereof!

andiwand: hm I think the jacobian should account for all of that. is the transport really just identity? maybe there are some terms that result into something else when you take the limit at the very end of the calculation? would be interesting to see

@russofel could you link some material were you calculated the derivatives? did you plug in the helix equations or how did that work?

beomki-yeo: Yeah Transport jacobian converges to the identity when delta (or step length) is closed to zero.

russofel:
> I wonder how much we could use from `RiddersPropagator` - I guess it does basically do what you did here by hand?

`RiddersPropagator` does something very similar: It aims to estimate the track parameter covariance at a target surface given the covariance at the initial surface. To this end, it wiggles each of the parameters at the initial surface and looks what happens to the parameters at the target surface. The "size" of the wiggle is determined by the covariance at the initial surface. 

Just like in the code here, it does the wiggling by hand, see: https://github.com/acts-project/acts/blob/e1a87dffb643a50b2adb99fef94741e25b3c7921/Core/include/Acts/Propagator/RiddersPropagator.ipp#L180

The difference of `RiddersPropagator` to our code is that it performs the wiggling several times and estimates the covariance via a linear fit, see https://github.com/acts-project/acts/blob/e1a87dffb643a50b2adb99fef94741e25b3c7921/Core/include/Acts/Propagator/RiddersPropagator.ipp#LL203C1-L219C1

I don't know if it makes sense to unify the code here with the one in `RiddersPropagator` - let me know what you think!

andiwand:
thanks for looking into this @russofel ! I guess if it would expose the jacobian instead of (or additional to) the covariance it would fit our use case. but right now it does not really make sense as we would have to find the jacobian from the final covariance? and I don't know if that is even possible

Core/include/Acts/Vertexing/NumericalTrackLinearizer.ipp
russofel: > It might not have nothing to do with this PR, but there is also a way to validate `HelicalTrackLinearizer` with the analytical solution of a helix model.
> 
> See https://github.com/acts-project/detray/blob/51571670ea7b8e574db560c9dadf49401352583d/core/include/detray/intersection/detail/trajectories.hpp#L268
> 
> which is based on the helix model (Eq. 4.28 - 4.34) of https://doi.org/10.1007/978-3-030-65771-0

The HelicalTrackLinearizer is purely based on analytical calculations, see Sec. 5.3.3 and 5.3.4 of https://cds.cern.ch/record/1243771?ln=en. 

I am not quite sure how these calculations relate to the resource you have linked in your comment though... I will need to look into this in detail!

Tests/UnitTests/Core/Vertexing/LinearizedTrackFactoryTests.cpp
```diff
@@ -203,41 +135,114 @@ BOOST_AUTO_TEST_CASE(linearized_track_factory_straightline_test) {
...
+      constField->makeCache(magFieldContext));
+
+  // Linearizer for 0 field and corresponding state
+  AnalyticalLinearizerStraightLine::Config straightLinConfig(
+      straightPropagator);
+  AnalyticalLinearizerStraightLine straightLinFactory(straightLinConfig);
+  AnalyticalLinearizerStraightLine::State straightLinState(
+      zeroField->makeCache(magFieldContext));
+
+  NumericalLinearizerStraightLine::Config numStraightLinConfig(
```
russofel, line 162: Are these names too long?

felix-russo:
> I wonder how much we could use from `RiddersPropagator` - I guess it does basically do what you did here by hand?

I added a comment emphasizing the connection to RiddersPropagator for future reference

Core/include/Acts/Vertexing/NumericalTrackLinearizer.ipp
```diff
@@ -0,0 +1,154 @@
...
+        intersection.intersection.pathLength);
+
+    // Unwiggle
+    paramVec(i) -= m_cfg.delta;
+
+    // Propagate to the new PCA and extract Perigee parameters
+    auto newResult = m_cfg.propagator->propagate(wiggledCurvilinearParams,
+                                                 *perigeeSurface, pOptions);
+    auto newEndParams = (*newResult->endParameters);
+    newPerigeeParams = newEndParams.parameters();
```
andiwand, line 124:
```suggestion
    newPerigeeParams = (*newResult->endParameters).parameters();
```

Core/include/Acts/Vertexing/NumericalTrackLinearizer.ipp
```diff
@@ -0,0 +1,154 @@
...
+        (newPerigeeParams.head(2) - perigeeParams.head(2)) / m_cfg.delta;
+    // We need to account for the periodicity of phi (see documentiation of
+    // difference_periodic)
+    completeJacobian(2, i) =
+        Acts::detail::difference_periodic(newPerigeeParams(2), perigeeParams(2),
+                                          2 * M_PI) /
+        m_cfg.delta;
+    // theta, q/p, and t
+    completeJacobian.block<3, 1>(3, i) =
+        (newPerigeeParams.tail(3) - perigeeParams.tail(3)) / m_cfg.delta;
```
andiwand, line 138: this could be compacter by overwriting the phi case afterwards

Core/include/Acts/Vertexing/NumericalTrackLinearizer.ipp
```diff
@@ -0,0 +1,154 @@
...
+    // Wiggle
+    paramVec(i) += m_cfg.delta;
+
+    // Create curvilinear track object from our parameters. This is needed for
+    // the propagation. Note that we work without covariance since we don't need
+    // it to compute the derivative.
+    Vector3 wiggledDir =
+        makeDirectionUnitFromPhiTheta(paramVec(4), paramVec(5));
+    CurvilinearTrackParameters wiggledCurvilinearParams(
+        paramVec.head(4), wiggledDir, paramVec(6));
```
andiwand, line 109: might be better to use enums or even define these special track params in a more central way

Core/include/Acts/Vertexing/NumericalTrackLinearizer.ipp
```diff
@@ -0,0 +1,154 @@
...
+  ActsMatrix<eBoundSize, nParams> completeJacobian;
+  completeJacobian.setZero();
+
+  // Perigee parameters wrt linPoint after wiggling
+  BoundVector newPerigeeParams;
+
+  // Check if wiggled angle theta are within definition range [0, pi]
+  assert(paramVec(5) + m_cfg.delta < M_PI &&
+         "Wiggled theta outside range, choose a smaller wiggle (i.e., delta)!"
+         "You might need to decrease targetTolerance as well.");
```
andiwand, line 95: since there is a logger available I would rather use it instead of `assert`

Core/include/Acts/Vertexing/NumericalTrackLinearizer.ipp
```diff
@@ -0,0 +1,154 @@
...
+    ActsScalar qOvP = perigeeParams(BoundIndices::eBoundQOverP);
+
+    paramVec << globalCoords, globalTime, phi, theta, qOvP;
+    pca << globalCoords, globalTime;
+    momentumAtPCA << phi, theta, qOvP;
+  }
+
+  // Complete Jacobian (consists of positionJacobian and momentumJacobian)
+  ActsMatrix<eBoundSize, nParams> completeJacobian;
+  completeJacobian.setZero();
```
andiwand, line 87:
```suggestion
  auto completeJacobian = ActsMatrix<eBoundSize, nParams>::Zero();
```

Tests/UnitTests/Core/Vertexing/LinearizedTrackFactoryTests.cpp
```diff
@@ -28,7 +29,15 @@ namespace Acts {
...
-using Linearizer = HelicalTrackLinearizer<Propagator<EigenStepper<>>>;
+// We will compare analytical and numerical computations in the case of a
+// (non-zero) constant B-field and a zero B-field.
+using AnalyticalLinearizer = HelicalTrackLinearizer<Propagator<EigenStepper<>>>;
+using AnalyticalLinearizerStraightLine =
+    HelicalTrackLinearizer<Propagator<StraightLineStepper>>;
+using NumericalLinearizer =
+    NumericalTrackLinearizer<Propagator<EigenStepper<>>>;
+using NumericalLinearizerStraightLine =
+    NumericalTrackLinearizer<Propagator<StraightLineStepper>>;
```
andiwand, line 40: might be worth typedefing the propagator

Tests/UnitTests/Core/Vertexing/LinearizedTrackFactoryTests.cpp
```diff
@@ -58,137 +67,60 @@ std::uniform_real_distribution<> resAngDist(0., 0.1);
...
 ///
 BOOST_AUTO_TEST_CASE(linearized_track_factory_test) {
-  // Number of tracks
-  unsigned int nTracks = 100;
+  // Number of tracks to linearize
+  unsigned int nTracks{100};
 
   // Set up RNG
-  int mySeed = 31415;
+  int mySeed{31415};
```
andiwand, line 81: I prefer `=` in these cases but that's a matter of taste

Tests/UnitTests/Core/Vertexing/LinearizedTrackFactoryTests.cpp
```diff
@@ -58,137 +67,60 @@ std::uniform_real_distribution<> resAngDist(0., 0.1);
...
+  Vector4 vtxPos;
+  double d0v;
+  double z0v;
+  {
+    double x = vXYDist(gen);
+    double y = vXYDist(gen);
+    double z = vZDist(gen);
+    vtxPos << x, y, z, 0;  // TODO: add time
+    d0v = std::hypot(x, y);
+    z0v = z;
```
andiwand, line 111: this was the case before but I think it is odd to use random values for a single case

andiwand:
nothing critical I think - should be ready to merge tomorrow

Tests/UnitTests/Core/Vertexing/LinearizedTrackFactoryTests.cpp
felix-russo: I agree, but I guess one could play around with the seed to see how the linearization performs for different (random) vertices.. Should I remove it and hardcode something here nonetheless?

felix-russo:
I will add the enum for the `LinearizerTrackParameters` to the `HelicalTrackLinearizer` in the PR #2179!

Tests/UnitTests/Core/Vertexing/LinearizedTrackFactoryTests.cpp
```diff
@@ -75,137 +85,60 @@ std::uniform_real_distribution<> resAngDist(0., 0.1);
...
+  auto straightPropagator =
+      std::make_shared<StraightPropagator>(straightStepper);
+
+  // Create perigee surface, initial track parameters will be relative to it
+  std::shared_ptr<PerigeeSurface> perigeeSurface{
+      Surface::makeShared<PerigeeSurface>(Vector3{0., 0., 0.})};
+
+  // Vertex position and corresponding d0 and z0
+  Vector4 vtxPos;
+  double d0v;
```
felix-russo, line 121:
```suggestion
  double d0v {};
```

Tests/UnitTests/Core/Vertexing/LinearizedTrackFactoryTests.cpp
```diff
@@ -75,137 +85,60 @@ std::uniform_real_distribution<> resAngDist(0., 0.1);
...
+  double z0v;
```
felix-russo, line 122:
```suggestion
  double z0v {};
```

Core/include/Acts/Utilities/VectorHelpers.hpp
```diff
@@ -211,5 +212,50 @@ inline auto makeVector4(const Eigen::MatrixBase<vector3_t>& vec3,
   return vec4;
 }
 
+// Computes the signed angle \in [-pi, pi] between a 2D vector vec and a 2D reference vector refVec
+// @note: the vectors should lie in the same plane
+// @note: for an explanation of why this computation works see
+//      https://stackoverflow.com/questions/5188561/
+// @note: in future Eigen releases, the nominator can be replaced by refVec.cross(vec), see
```
paulgessinger, line 219: numerator?

Core/include/Acts/Utilities/VectorHelpers.hpp
```diff
@@ -211,5 +212,50 @@ inline auto makeVector4(const Eigen::MatrixBase<vector3_t>& vec3,
...
+
+// Decomposes a 3D vector into a transverse and a longitudinal component with
+// respect to the z-direction
+template <typename Derived>
+inline Eigen::Matrix<typename Derived::Scalar, 2, 1> decomposeTransverseZ(
+    const Eigen::MatrixBase<Derived>& vec) {
+  Eigen::Matrix<typename Derived::Scalar, 2, 1> decomposedVec{perp(vec.head(2)),
+                                                              vec[2]};
+  return decomposedVec;
+}
```
paulgessinger, line 236: This is close to the threshold where I wouldn't factorize this into a separate function.

Core/include/Acts/Vertexing/NumericalTrackLinearizer.hpp
```diff
@@ -0,0 +1,164 @@
...
+#include "Acts/MagneticField/MagneticFieldProvider.hpp"
+#include "Acts/MagneticField/NullBField.hpp"
+#include "Acts/Propagator/EigenStepper.hpp"
+#include "Acts/Propagator/Propagator.hpp"
+#include "Acts/Utilities/Result.hpp"
+#include "Acts/Vertexing/LinearizedTrack.hpp"
+
+namespace Acts {
+
+/// @class NumericalTrackLinearizer
```
paulgessinger, line 24: Thumbs up for the thorough comment!

Core/include/Acts/Vertexing/NumericalTrackLinearizer.hpp
```diff
@@ -0,0 +1,164 @@
...
+    ///
+    /// @param prop Propagator
+    Config(std::shared_ptr<const Propagator_t> prop)
+        : bField{std::make_shared<NullBField>()}, propagator(std::move(prop)) {}
+
+    std::shared_ptr<const MagneticFieldProvider> bField;
+
+    std::shared_ptr<const Propagator_t> propagator;
+
+    // Tolerance determining how close we need to get to a surface to
```
paulgessinger, line 97: Could use `///` for a comment doxygen picks up.

Core/include/Acts/Vertexing/NumericalTrackLinearizer.hpp
```diff
@@ -0,0 +1,164 @@
...
+  /// @param state Linearizer state object
+  ///
+  /// @return Linearized track
+  Result<LinearizedTrack> linearizeTrack(const BoundTrackParameters& params,
+                                         const Vector4& linPoint,
+                                         const Acts::GeometryContext& gctx,
+                                         const Acts::MagneticFieldContext& mctx,
+                                         State& state) const {
+    // Using "state" to suppress unused-parameter warning
+    (void)state;
```
paulgessinger, line 133:
```suggestion
                                         State& /*state*/) const {
```
would be an alternative.

Core/include/Acts/Vertexing/NumericalTrackLinearizer.hpp
```diff
@@ -0,0 +1,164 @@
...
+  ///
+  /// @param params Parameters to linearize
+  /// @param linPoint Point which defines the Perigee.
+  /// @note Transverse plane of the Perigee corresponding to @p linPoint is
+  /// parallel to the global x-y plane
+  /// @param gctx Geometry context
+  /// @param mctx Magnetic field context
+  ///
+  /// @return Linearized track
+  Result<LinearizedTrack> linearizeTrack(
```
paulgessinger, line 149: Do you need a separate function for this? Is it ever called other than in the other method?

Core/include/Acts/Vertexing/NumericalTrackLinearizer.ipp
```diff
@@ -0,0 +1,152 @@
...
+        paramVec.head(4), wiggledDir, paramVec(eLinQOverP));
+
+    // Obtain propagation direction
+    intersection =
+        perigeeSurface->intersect(gctx, paramVec.head(3), wiggledDir, false);
+    pOptions.direction = Direction::fromScalarZeroAsPositive(
+        intersection.intersection.pathLength);
+
+    // Unwiggle
+    paramVec(i) -= m_cfg.delta;
```
paulgessinger, line 122: Applying the wiggle to a copy would also work.

Tests/UnitTests/Core/Vertexing/LinearizedTrackFactoryTests.cpp
```diff
@@ -220,41 +153,116 @@ BOOST_AUTO_TEST_CASE(linearized_track_factory_straightline_test) {
...
-    BOOST_CHECK_NE(linTrack.constantTerm, vecBoundZero);
+  // Linearizer for constant field and corresponding state
+  AnalyticalLinearizer::Config linConfig(constField, propagator);
+  AnalyticalLinearizer linFactory(linConfig);
+  AnalyticalLinearizer::State linState(constField->makeCache(magFieldContext));
+
+  NumericalLinearizer::Config numLinConfig(constField, propagator);
+  NumericalLinearizer numLinFactory(numLinConfig);
+  NumericalLinearizer::State numLinState(
+      constField->makeCache(magFieldContext));
```
paulgessinger, line 171: For constant and zero fields, the cache has infinite validity, to technically you could reuse it. It won't really make a difference here, just mentioning it.

Core/include/Acts/Vertexing/LinearizerTrackParameters.hpp
```diff
@@ -0,0 +1,47 @@
...
+/// Enum to access the components of a track parameter vector.
+///
+/// Here, we parametrize the track via a 4D point on the track, the momentum
+/// angles of the particle at that point, and q/p or 1/p.
+///
+/// @note It would make sense to rename these parameters if they are used outside of track linearization.
+/// @note This must be a regular `enum` and not a scoped `enum class` to allow
+/// implicit conversion to an integer. The enum value are thus visible directly
+/// in `namespace Acts` and are prefixed to avoid naming collisions.
+enum LinIndices : unsigned int {
```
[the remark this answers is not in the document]
andiwand: sadly the vertex fitting is relying on an alternative track parameter representation

Core/include/Acts/Vertexing/LinearizerTrackParameters.hpp
```diff
@@ -0,0 +1,47 @@
...
+  eLinQOverP = 6u,
+  // Total number of components
+  eLinSize = 7u,
+  // Number of space-time components (3+1)
+  eLinPosSize = 4u,
+  // Number of momentum components
+  eLinMomSize = 3u,
+};
+
+}  // namespace Acts
```
andiwand, line 47: newline at the end
```suggestion
}  // namespace Acts

```

Core/include/Acts/Utilities/VectorHelpers.hpp
```diff
@@ -211,5 +212,50 @@ inline auto makeVector4(const Eigen::MatrixBase<vector3_t>& vec3,
...
+    const Eigen::MatrixBase<Derived>& refVec) {
+  return (refVec[0] * vec[1] - refVec[1] * vec[0]) / refVec.dot(vec);
+}
+
+// Decomposes a 3D vector into a transverse and a longitudinal component with
+// respect to the z-direction
+template <typename Derived>
+inline Eigen::Matrix<typename Derived::Scalar, 2, 1> decomposeTransverseZ(
+    const Eigen::MatrixBase<Derived>& vec) {
+  Eigen::Matrix<typename Derived::Scalar, 2, 1> decomposedVec{perp(vec.head(2)),
```
andiwand, line 233:
```suggestion
  Eigen::Matrix<typename Derived::Scalar, 2, 1> decomposedVec{perp(vec.head<2>()),
```

Core/include/Acts/Utilities/VectorHelpers.hpp
```diff
@@ -211,5 +212,50 @@ inline auto makeVector4(const Eigen::MatrixBase<vector3_t>& vec3,
...
+}
+
+// Subtracts phi and theta
+// minPhiTheta and subPhiTheta should be 2D vectors containing phi in the 0th
+// entry and theta in the first entry
+// @note: the prefix "min" stands for minuend and the prefix "sub" for subtrahend
+// @note: Simply subtracting phis is problematic due to the periodicity of phi
+// @note: Simply subtracting thetas is problematic around the poles of the unit sphere (i.e., at theta close to 0 or pi)
+template <typename Derived>
+inline Eigen::Matrix<typename Derived::Scalar, 2, 1> subtractPhiTheta(
```
andiwand, line 245: since this breaks down at the poles I am not sure how much sense it makes to generalize it that far. I think operations near the pole should fail as the derivatives will be meaningless anyways

Tests/UnitTests/Core/Vertexing/LinearizedTrackFactoryTests.cpp
andiwand: lets leave it as is is

Core/include/Acts/Utilities/VectorHelpers.hpp
felix-russo: Ok, I will remove these functions then!

andiwand (approved):

PR #2142, opened 2023-05-23
fix: Remove unit C

andiwand:
Not sure what this is and why it was there in this place?

**Note**
This contains a very odd fix for a problem I was facing in the integration test which compares the analytical covariance with the numerical one. For some reason the numerical derivative breaks down in the case where we step towards a surface which is 1 cm away with a single step. Using 0.9 cm it works because it will do a second step which seems to provide the necessary accuracy for the differential quotient.
This is not pretty and I would like to take a second look why a single step is not sufficient in this case.


paulgessinger:
This probably needs a bit of discussion to understand it fully.

AJPfleger:
It should be [Coulomb](https://en.wikipedia.org/wiki/Coulomb). How to get to the value is explained in PRs #1179 and #1231 . But I don't know if we actually need it as a unit.

I am not sure if we should have `hbar` and `c` as units instead of keeping them in the namespace `PhysicalConstants`, because I think we don't want to set actions and velocities as multiples of `hbar` or `c`. (If we would ever use those quantities.)

andiwand:
I don't see any value in having `Coulomb` and it does not seem to be used in the codebase. It might have been confused for `c` the speed of light in the code I modified in this PR aswell.

I don't have a strong opinion towards splitting constants or not into `Units` and `PhysicalConstants`. I just thought it would be good to set them somewhere for reference. In the python bindings we do not distinguish them so also dropped it here. To me `0.5 * c` makes more sense in terms of a unit but as you said we will most likely not use that.

Tests/IntegrationTests/PropagationTests.hpp
```diff
@@ -280,7 +280,7 @@ inline std::pair<Acts::BoundTrackParameters, double> transportToSurface(
   options_t<Actions, Aborts> options(geoCtx, magCtx);
   options.direction = Acts::Direction::Forward;
   options.pathLimit = pathLimit;
-  options.maxStepSize = 1_cm;
+  options.maxStepSize = 0.9_cm;
```
Corentin-Allaire, line 283: Any reason to change this here ? Shouldn't this be part of another PR ?

Corentin-Allaire (commented):
Looks good, I think a change from another PR got in here by accident ?

Tests/IntegrationTests/PropagationTests.hpp
andiwand: Sorry I should have mentioned this somewhere. This is a very odd fix for a problem I was facing in the integration test which compares the analytical covariance with the numerical one. For some reason the numerical derivative breaks down in the case where we step towards a surface which is 1 cm away with a single step. Using 0.9 cm it works because it will do a second step which seems to provide the necessary accuracy for the differential quotient.

This is not pretty and I would like to take a second look why a single step is not sufficient in this case.

Corentin-Allaire: Do you think you could copy this information in the description of the PR ? That way in the future if we need to look at this change it will be explicit

andiwand: done :white_check_mark:

Corentin-Allaire (approved):

acts-project-service:
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/30892521) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/5862224)!

paulgessinger:
This was a false failure due to #2251

PR #2143, opened 2023-05-24
fix: Fix truth association in VertexPerformanceWriter

andiwand:
truth association is broken

I also added reco time and remove cpu time

### DIFF
--- a/CI/physmon/vertexing_config.yml
+++ b/CI/physmon/vertexing_config.yml
@@ -14,12 +14,12 @@ histograms:
     min: -6
     max: 6
 
-  "covXX|covYY|covZZ":
+  "covXX|covYY|covZZ|covTT":
     nbins: 100
     min: -0.0005
     max: 0.0005
 
-  "covXY|covXZ|covYX":
+  "covXY|covXZ|covXT|covYX|covYZ|covZT":
     nbins: 100
     min: -0.0001
     max: 0.0001
@@ -35,7 +35,7 @@ histograms:
     max: 0.06
     nbins: 100
 
-  "truthZ|recoZ":
+  "truthZ|truthT|recoZ|recoT":
     min: -200
     max: 200
     nbins: 100
@@ -56,6 +56,3 @@ extra_histograms:
     nbins: 50
     min: 0.0
     max: 1.0
-
-exclude:
-  - timeMS
--- a/Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
+++ b/Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
@@ -8,6 +8,7 @@
 
 #include "ActsExamples/Io/Performance/VertexPerformanceWriter.hpp"
 
+#include "Acts/Definitions/TrackParametrization.hpp"
 #include "Acts/EventData/MultiTrajectory.hpp"
 #include "Acts/EventData/MultiTrajectoryHelpers.hpp"
 #include "Acts/EventData/TrackParameters.hpp"
@@ -71,8 +72,6 @@ ActsExamples::VertexPerformanceWriter::VertexPerformanceWriter(
     m_inputTrajectories.initialize(m_cfg.inputTrajectories);
   }
 
-  m_inputTime.maybeInitialize(m_cfg.inputTime);
-
   // Setup ROOT I/O
   auto path = m_cfg.filePath;
   m_outputFile = TFile::Open(path.c_str(), m_cfg.fileMode.c_str());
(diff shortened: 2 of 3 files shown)


Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
```diff
@@ -85,35 +84,51 @@ ActsExamples::VertexPerformanceWriter::VertexPerformanceWriter(
...
-
-    m_outputTree->Branch("recoX", &m_recoX);
-    m_outputTree->Branch("recoY", &m_recoY);
-    m_outputTree->Branch("recoZ", &m_recoZ);
+    m_outputTree->Branch("pullT", &m_pullT);
 
     m_outputTree->Branch("truthX", &m_truthX);
     m_outputTree->Branch("truthY", &m_truthY);
     m_outputTree->Branch("truthZ", &m_truthZ);
+    m_outputTree->Branch("truthT", &m_truthT);
```
pbutti, line 110: Are these duplicated? I see them in line 87-90 as well.

Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
```diff
@@ -291,51 +306,53 @@ ActsExamples::ProcessCode ActsExamples::VertexPerformanceWriter::writeT(
         }
 
         const auto& majorityParticle = *it;
-        associatedTruthParticles.emplace_hint(associatedTruthParticles.end(),
-                                              majorityParticle);
+        trackParameters.push_back(trajectories.trackParameters(tip));
+        associatedTruthParticles.push_back(majorityParticle);
```
pbutti, line 310: I'm confused by line 309. If At line 247 you create a vector from the SimParticle Set, why are you pushing back here? Aren't you adding the particles to an already filled vector? 
We talked about this, was it checked?

Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
```diff
@@ -291,51 +306,53 @@ ActsExamples::ProcessCode ActsExamples::VertexPerformanceWriter::writeT(
...
     }
   }
 
-  ACTS_VERBOSE(
-      "Total number of reconstructed tracks : " << trackParameters.size());
-
   // Get number of track-associated true primary vertices
   m_nVtxReconstructable =
-      getNumberOfReconstructableVertices(associatedTruthParticles);
+      getNumberOfReconstructableVertices(SimParticleContainer(
```
pbutti, line 317: It would be handy to move getting the number of ReconstructableVertices in a vertexing helpers class to allow other algorithms to use it. I'm Ok to do that in a subsequent PR.

Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
```diff
@@ -291,51 +306,53 @@ ActsExamples::ProcessCode ActsExamples::VertexPerformanceWriter::writeT(
...
     for (const auto& trk : tracks) {
       Acts::BoundTrackParameters origTrack = *(trk.originalParams);
 
       // Find associated truth particle now
-      int idx = 0;
-      for (const auto& particle : associatedTruthParticles) {
-        if (origTrack.parameters() == trackParameters[idx].parameters()) {
-          particleAtVtx.insert(particleAtVtx.end(), particle);
+      for (std::size_t i = 0; i < trackParameters.size(); ++i) {
+        const auto& particle = associatedTruthParticles[i];
```
pbutti, line 342: Worth adding to the comment in line 340, that we assume that the associatedTruthParticles and the trackParameters vectors are aligned.

Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
andiwand: nice catch - thanks

Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
```diff
@@ -85,35 +84,51 @@ ActsExamples::VertexPerformanceWriter::VertexPerformanceWriter(
...
-
-    m_outputTree->Branch("recoX", &m_recoX);
-    m_outputTree->Branch("recoY", &m_recoY);
-    m_outputTree->Branch("recoZ", &m_recoZ);
+    m_outputTree->Branch("pullT", &m_pullT);
 
     m_outputTree->Branch("truthX", &m_truthX);
     m_outputTree->Branch("truthY", &m_truthY);
     m_outputTree->Branch("truthZ", &m_truthZ);
+    m_outputTree->Branch("truthT", &m_truthT);
```
andiwand, line 111:
```suggestion
```

Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
```diff
@@ -291,51 +306,53 @@ ActsExamples::ProcessCode ActsExamples::VertexPerformanceWriter::writeT(
...
+
     // Store all associated truth particles to current vtx
     SimParticleContainer particleAtVtx;
-
     std::vector<int> contributingTruthVertices;
 
     for (const auto& trk : tracks) {
       Acts::BoundTrackParameters origTrack = *(trk.originalParams);
 
       // Find associated truth particle now
```
andiwand, line 340:
```suggestion
      // Find associated truth particle now
      // We expect that the vectors `trackParameters` and `associatedTruthParticles` align
```

Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
andiwand: this is misleading in the github diff view. there is an `} else {` in-between

Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
andiwand: I agree - that should be accessible outside the writer. same is true for the truth matching itself. we can do this after this pr

pbutti (approved):

Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
```diff
@@ -291,51 +301,55 @@ ActsExamples::ProcessCode ActsExamples::VertexPerformanceWriter::writeT(
...
     }
   }
 
-  ACTS_VERBOSE(
-      "Total number of reconstructed tracks : " << trackParameters.size());
-
   // Get number of track-associated true primary vertices
   m_nVtxReconstructable =
-      getNumberOfReconstructableVertices(associatedTruthParticles);
+      getNumberOfReconstructableVertices(SimParticleContainer(
```
paulgessinger, line 312: This makes a copy and then immediately throws it away?

paulgessinger (commented):
Maybe you can elaborate how it's "broken"? From the diff I only see that you add 4D residuals and covariance terms. Also you're moving when it records the track parameters to when they've been matched against a particle and not record them until we know it matches a truth particle. 

Anything? Maybe you can add this to the PR description.

CI/physmon/vertexing_config.yml
```diff
@@ -4,41 +4,36 @@ histograms:
     min: -0.1
     max: 0.1
 
-  resZ:
```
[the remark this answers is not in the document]
andiwand: In the plots I noticed that the resolution of Z is now on the same scale as X and Y. It falls now into the same condition after removing

Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
andiwand: Yeah that is the problem with the vector vs set. We use sets for particles but when you do truth matching you want to align your particles with something which isn't straight forward. I changed the internal particle collection to a vector now but still need to convert it if we call helper functions

andiwand:
updated the description

CI/physmon/vertexing_config.yml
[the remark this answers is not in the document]
paulgessinger: But isn't that purely coincidental? Why would we not want to monitor this?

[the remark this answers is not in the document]
andiwand: we still monitor it but now it falls in the filter above `"res.*"` I thought - no?

pbutti (approved):

PR #2144, opened 2023-05-24
refactor: Align IVF and AMVF in examples

andiwand:
IVF and AMVF use different seeders and the code diverged quite a bit in the examples. In this PR I try to realign them


Core/include/Acts/Vertexing/IterativeVertexFinder.ipp
```diff
@@ -429,7 +437,7 @@ Acts::IterativeVertexFinder<vfitter_t, sfinder_t>::reassignTracksToNewVertex(
     for (auto tracksIter = tracksBegin; tracksIter != tracksEnd;) {
       // consider only tracks that are not too tightly assigned to other
       // vertex
-      if (tracksIter->trackWeight > m_cfg.cutOffTrackWeight) {
+      if (tracksIter->trackWeight > m_cfg.cutOffTrackWeightReassign) {
```
pbutti, line 440: As discussed on mattermost, we do not set the trackWeight in the FullBilloirVertexFitter. 
Unless the fitter is fixed and the track weights are updated according to https://cds.cern.ch/record/2670380/files/ATL-PHYS-PUB-2019-015.pdf page 5, this line is problematic as an user is free to set the reassign track weight cut from configuration but trackWeight == 1 always.

Core/include/Acts/Vertexing/IterativeVertexFinder.hpp
```diff
@@ -105,6 +105,7 @@ class IterativeVertexFinder {
     bool doMaxTracksCut = false;
     int maxTracks = 5000;
     double cutOffTrackWeight = 0.01;
+    double cutOffTrackWeightReassign = 1;
```
pbutti, line 108: I would suggest to add some documentation, especially with this double weight cut off, where the second is used to decide if the track will be reassigned.

pbutti (commented):
As discussed on mattermost seems like the whole performance improve comes from a different seeder.

Core/include/Acts/Vertexing/IterativeVertexFinder.ipp
andiwand: `tracksIter->trackWeight > m_cfg.cutOffTrackWeight` just does not make any sense because the tracks below that threshold will be filtered out anyways

Core/include/Acts/Vertexing/IterativeVertexFinder.hpp
andiwand:
```suggestion
    /// If `reassignTracksAfterFirstFit` is set this threshold will be used to decide if a track should be checked for reassignment to other vertices
    double cutOffTrackWeightReassign = 1;
```

Core/include/Acts/Vertexing/IterativeVertexFinder.ipp
pbutti: Right, but `tracksIter->trackWeight > m_cfg.cutOffTrackWeightReassign` doesn't make too much sense either if `tracksIter->trackWeight == 1` always. One can set `cutOffTrackWeightReassign` to whatever value in the [0,1) interval obtaining always the same result.

andiwand: but that is a different problem right? if we ignore the fact that `trackWeight` is always `0` this is an improvement IMO because you can set a threshold that will actually trigger the reassignment

pbutti: Right. I agree. Perhaps issue a warning if `cutOffTrackWeightReassign` is  != 1 in the config passed to the code unti #2149 is fixed? 

For the rest looks good. 
General comment: Is the sigma of the gaussian density seeder used as constraint in the finding step?

pbutti (approved):

pbutti (approved):

Examples/Algorithms/Vertexing/src/IterativeVertexFinderAlgorithm.cpp
```diff
@@ -90,26 +90,24 @@ ActsExamples::ProcessCode ActsExamples::IterativeVertexFinderAlgorithm::execute(
...
   Linearizer::Config linearizerCfg(m_cfg.bField, propagator);
   Linearizer linearizer(linearizerCfg, logger().cloneWithSuffix("HelLin"));
   // Setup the seed finder
-  ImpactPointEstimator::Config ipEstCfg(m_cfg.bField, propagator);
-  ImpactPointEstimator ipEst(ipEstCfg);
-  VertexSeeder::Config seederCfg(ipEst);
-  VertexSeeder seeder(seederCfg);
+  IPEstimator::Config ipEstCfg(m_cfg.bField, propagator);
+  IPEstimator ipEst(ipEstCfg);
+  Seeder seeder;
```
paulgessinger, line 102: No config anymore?

Core/include/Acts/Vertexing/IterativeVertexFinder.ipp
```diff
@@ -231,7 +239,7 @@ Acts::IterativeVertexFinder<vfitter_t, sfinder_t>::getCompatibility(
        (vertex.fullCovariance() * linTrack.positionJacobian.transpose()))
           .template block<2, 2>(0, 0);
   weightReduced += errorVertexReduced;
-  weightReduced = weightReduced.inverse();
+  weightReduced = weightReduced.inverse().eval();
```
paulgessinger, line 242: Does this do anything in this case?

paulgessinger (commented):
I know it's merged already, but anyway

Examples/Algorithms/Vertexing/src/IterativeVertexFinderAlgorithm.cpp
andiwand: I aligned this with the AMVF. Now we use the `TrackDensityVertexFinder` which seems to go fine with the default constructed `GaussianTrackDensity`

Core/include/Acts/Vertexing/IterativeVertexFinder.ipp
andiwand: For some reason I got runtime alignment errors. The error message suggested using eval

paulgessinger: Strange but ok.

PR #2146, opened 2023-05-24
chore: Wire loggers in CKF correctly

andiwand:
That way updater and smoother logs will not be automatically silenced

### DIFF
--- a/Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
+++ b/Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
@@ -954,7 +954,7 @@ class CombinatorialKalmanFilter {
         } else {
           // Kalman update
           auto updateRes = m_extensions.updater(
-              gctx, trackState, Direction::Forward, updaterLogger);
+              gctx, trackState, Direction::Forward, *updaterLogger);
           if (!updateRes.ok()) {
             ACTS_ERROR("Update step failed: " << updateRes.error());
             return updateRes.error();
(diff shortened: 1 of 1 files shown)


Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -950,7 +950,7 @@ class CombinatorialKalmanFilter {
         } else {
           // Kalman update
           auto updateRes = m_extensions.updater(
-              gctx, trackState, Direction::Forward, getDummyLogger());
+              gctx, trackState, Direction::Forward, *logger().clone("updater"));
```
paulgessinger, line 953:
```suggestion
              gctx, trackState, Direction::Forward, *logger().cloneWithSuffix("Updater"));
```

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -1129,9 +1129,9 @@ class CombinatorialKalmanFilter {
       ACTS_VERBOSE("Apply smoothing on " << nStates
                                          << " filtered track states.");
       // Smooth the track states
-      auto smoothRes =
-          m_extensions.smoother(state.geoContext, *result.fittedStates,
-                                lastMeasurementIndex, getDummyLogger());
+      auto smoothRes = m_extensions.smoother(
+          state.geoContext, *result.fittedStates, lastMeasurementIndex,
+          *logger().clone("smoother"));
```
paulgessinger, line 1134:
```suggestion
          *logger().cloneWithSuffix("Smoother"));
```

paulgessinger (commented):
How about this?

Also, the logger should be created ahead of time, otherwise we do a lot of unnecessary heap allocations. Also consider that this is only not dangling because of lifetime extension (`clone()` returns a `unique_ptr`, which you dereference), which I'd prefer to avoid.

andiwand:
good point - I adopted this now

paulgessinger:
The job queue overflow fails sentinel, I'm merging manually.

PR #2147, opened 2023-05-24
ci: Disable LCG jobs we expect to fail

paulgessinger:
To again reduce pressure


andiwand (approved):
why are those failing?

PR #2151, opened 2023-05-25
feat: Safe inverse

andiwand:
After discussion in https://github.com/acts-project/acts/pull/2122 and offline we decided to add a safe inverse that can be used in special occasions to avoid FPEs being thrown in order to handle them in a different, more controlled way.

The main motivation for this is that Athena has a very strict FPE policy which marks every occurrence as a job failure. Since we are dealing with a high number of events a certain number of inverse failures would be acceptable and should not be treated as a fatality.

### DIFF
--- a/Core/include/Acts/Utilities/Helpers.hpp
+++ b/Core/include/Acts/Utilities/Helpers.hpp
@@ -23,6 +23,7 @@
 #include <iostream>
 #include <limits>
 #include <memory>
+#include <optional>
 #include <string>
 #include <tuple>
 #include <type_traits>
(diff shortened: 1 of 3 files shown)


Core/include/Acts/Utilities/Helpers.hpp
```diff
@@ -628,4 +629,33 @@ std::tuple<typename T::value_type, ActsScalar> range_medium(const T& tseries) {
   return std::tie(range, medium);
 }
 
+/// Calculate the inverse of an Eigen matrix after checking if it can be
+/// numerically inverted. This allows to catch potential FPEs before they occur.
+///
+/// Our main motivation for this is that Athena has a very strict FPE policy
```
andiwand, line 635:
```suggestion
/// Our main motivation for this is that users might have a strict FPE policy
```

paulgessinger (approved):

PR #2159, opened 2023-05-27
ci: Physmon for single particles

andiwand:
WIPWIP


paulgessinger:
Let's see how long this run since it's a lot of combinations.

andiwand:
looks like we cannot have `addGeant4` twice - I'll have to investigate

paulgessinger:
We could add this once we're moving to pytest for physmon, I guess.

andiwand:
I will tackle this again when we have pytest for physmon and more time at hand

PR #2163, opened 2023-05-30
feat: Move greedy ambiguity solver to core

andiwand:
In this PR I move the simple `AmbiguityResolution` to the Core.

In that process I renamed this algorithm consistently to `GreedyAmbiguityResolution` and created specific directories for the ambiguity resolution in general.


Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.hpp
```diff
@@ -0,0 +1,98 @@
...
+    std::vector<int> trackTips;
+    std::vector<float> trackChi2;
+    std::vector<std::vector<std::size_t>> measurementsPerTrack;
+
+    boost::container::flat_map<std::size_t,
+                               boost::container::flat_set<std::size_t>>
+        tracksPerMeasurement;
+    std::vector<std::size_t> sharedMeasurementsPerTrack;
+
+    boost::container::flat_set<std::size_t> selectedTracks;
```
andiwand, line 63: boost 1.81 seems to provide `unordered_flat_map` which might be preferable here

Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.ipp
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Acts {
+
+template <typename track_container_t, typename traj_t,
+          template <typename> class holder_t>
+void GreedyAmbiguityResolution::computeInitialState(
+    const TrackContainer<track_container_t, traj_t, holder_t>& tracks,
+    State& state) const {
+  std::unordered_map<SourceLink, std::size_t, decltype(m_cfg.sourceLinkHash),
+                     decltype(m_cfg.sourceLinkEquality)>
```
goetzgaycken, line 23: I was wondering whether the delegates would add a noticeable overhead, and whether that could be avoided by providing the missing functionality via an adapter (extra template parameter)? But maybe my concerns are not justified.

andiwand: I am also not very happy about this delegate. IMO the source link should provide you with the functionality of checking if two are equal or not.

I guess having the hash and equality passed via a template would remove one level of indirection and allows for more compiler optimization. I can try and see if this also makes the code a bit nicer

paulgessinger: If the source link did that we'd have to use a similar type erasure mechanism as delegate itself. Delegate should have very little overhead for function calls, it's basically a function pointer call that can be inclined even. So I don't think this is something we need to be worried about.

Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.ipp
```diff
@@ -0,0 +1,74 @@
...
+#pragma once
+
+#include "Acts/AmbiguityResolution/GreedyAmbiguityResolution.hpp"
+
+#include <unordered_map>
+
+namespace Acts {
+
+template <typename track_container_t, typename traj_t,
+          template <typename> class holder_t, typename source_link_hash_t,
```
paulgessinger, line 18: Since these functions are not stored only passed to the function, I guess in this case templates do make sense.

Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.ipp
```diff
@@ -0,0 +1,74 @@
...
+    const TrackContainer<track_container_t, traj_t, holder_t>& tracks,
+    State& state, source_link_hash_t&& sourceLinkHash,
+    source_link_equality_t&& sourceLinkEquality) const {
+  auto measurementIndexMap =
+      std::unordered_map<SourceLink, std::size_t, source_link_hash_t,
+                         source_link_equality_t>(0, sourceLinkHash,
+                                                 sourceLinkEquality);
+
+  for (const auto& track : tracks) {
+    auto trajState = Acts::MultiTrajectoryHelpers::trajectoryState(
```
paulgessinger, line 30: Track should have the number of measurements already I think, so there's probably no need to recompute anymore?

Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.ipp
```diff
@@ -0,0 +1,74 @@
...
+            // yet
+            auto emplace = measurementIndexMap.try_emplace(
+                sourceLink, measurementIndexMap.size());
+            measurements.push_back(emplace.first->second);
+          }
+          return true;
+        });
+
+    state.trackTips.push_back(track.index());
+    state.trackChi2.push_back(trajState.chi2Sum / trajState.NDF);
```
paulgessinger, line 50: These should also be stored on the track already.

Examples/Algorithms/AmbiguityResolution/src/GreedyAmbiguityResolutionAlgorithm.cpp
```diff
@@ -0,0 +1,101 @@
...
+  result.nMeasurementsMin = cfg.nMeasurementsMin;
+  return result;
+}
+
+std::size_t sourceLinkHash(const Acts::SourceLink& a) {
+  return static_cast<std::size_t>(
+      a.get<ActsExamples::IndexSourceLink>().index());
+}
+
+bool sourceLinkEquality(const Acts::SourceLink& a, const Acts::SourceLink& b) {
```
paulgessinger, line 43: Just fyi, with this being in the CPP file, without LTO it's not getting inlined, which sort of negates the benefit of the template arguments.

paulgessinger (commented):
Good stuff!

Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.hpp
```diff
@@ -0,0 +1,95 @@
...
+  struct State {
+    std::size_t numberOfTracks{};
+
+    std::vector<int> trackTips;
+    std::vector<float> trackChi2;
+    std::vector<std::vector<std::size_t>> measurementsPerTrack;
+
+    boost::container::flat_map<std::size_t,
+                               boost::container::flat_set<std::size_t>>
+        tracksPerMeasurement;
```
andiwand, line 55:
```suggestion
    // TODO consider boost 1.81 unordered_flat_map
    boost::container::flat_map<std::size_t,
                               boost::container::flat_set<std::size_t>>
        tracksPerMeasurement;
```

Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.hpp
```diff
@@ -0,0 +1,95 @@
...
+    std::vector<std::size_t> sharedMeasurementsPerTrack;
+
+    boost::container::flat_set<std::size_t> selectedTracks;
```
andiwand, line 58:
```suggestion
    // TODO consider boost 1.81 unordered_flat_map
    boost::container::flat_set<std::size_t> selectedTracks;
```

Examples/Algorithms/AmbiguityResolution/src/GreedyAmbiguityResolutionAlgorithm.cpp
andiwand: I thought it can still benefit from inlining because the call which receives the template arguments is also in this CPP file

andiwand:
somehow this is changing the output... need to investigate. code wise I am already quite happy

paulgessinger:
I guess we expect exactly identical physmon results, right?

Examples/Algorithms/AmbiguityResolution/src/GreedyAmbiguityResolutionAlgorithm.cpp
paulgessinger: Yeah, maybe

Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.hpp
```diff
@@ -0,0 +1,97 @@
...
+ private:
+  Config m_cfg;
+
+  /// Logging instance
+  std::unique_ptr<const Logger> m_logger;
+
+  /// Private access to logging instance
+  const Logger& logger() const { return *m_logger; }
+
+  struct MapableSourceLink final {
```
paulgessinger, line 88:
```suggestion
  struct MappableSourceLink final {
```
?

andiwand:
yes I would think so... I think the source link order shuffled but I do not expect a different output from that when we use chi2 as weight

Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.ipp
```diff
@@ -0,0 +1,72 @@
...
+      std::unordered_map<SourceLink, std::size_t, source_link_hash_t,
+                         source_link_equality_t>(0, sourceLinkHash,
+                                                 sourceLinkEquality);
+
+  for (const auto& track : tracks) {
+    if (track.nMeasurements() < m_cfg.nMeasurementsMin) {
+      continue;
+    }
+    std::vector<std::size_t> measurements;
+    tracks.trackStateContainer().visitBackwards(
```
paulgessinger, line 34: Just fyi, I plan to add convenience to track that allows you to iterate over the measurements only without having to filter yourself.

paulgessinger: Also: you don't have to change it, but you can do

```cpp
for (auto ts : track.trackStates()) {}
```

So you don't necessarily have to use the tip index manually.

Would be nice to use so I can see if the interface works well.

paulgessinger (commented):
Sorry, couple more comments.

Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.hpp
andiwand: oh that's actually dead code - thanks :smile:

Core/include/Acts/EventData/MultiTrajectoryHelpers.hpp
```diff
@@ -70,6 +69,7 @@ TrajectoryState trajectoryState(const Acts::MultiTrajectory<traj_t>& multiTraj,
       trajState.measurementVolume.push_back(volume);
       trajState.measurementLayer.push_back(layer);
       trajState.chi2Sum += state.chi2();
+      trajState.NDF += state.calibratedSize();
```
andiwand, line 72: I did some investigation locally and found that `NDF` of the trajectory state is different from the `nDoF` in the track. With these changes they are aligned.

Not sure if this should be part of this PR or of another one. If we would merge this change first we should see that the Ambi performance aligns.

paulgessinger (approved):

paulgessinger (approved):

PR #2164, opened 2023-05-30
feat: Early version of Exa.TrkX pipeline on CPU

benjaminhuth:
This allows to compile & run the Exa.TrkX pipeline on CPU. So far, I mainly want to use this to work on the code on machines with no GPU / without the heavy GPU dependencies.

In the future, this should be improved by making this configurable and replacing the brute force NN search with an optimized one.


andiwand (approved):
looks good to me

andiwand (approved):

PR #2167, opened 2023-05-30
chore: Bump recommended boost version to 1.78

andiwand:
After running into this issue again https://github.com/boostorg/container/issues/244 I would like to bump the boost version.

In order to not break any downstream builds I added a recommended boost version but left the minimal version untouched. If the version provided from the user is below the recommended version I will show a warning. If we acquire boost ourselves via cmake I will download the recommended version.

Additionally I worked around the bug that pops up with boost <1.78 from time to time which will have some performance penalty in the examples framework but I think that is preferred over an assert that might abort execution.

I also included some fixes that were necessary for me to successfully built against our custom cmake boost.


paulgessinger:
Also needs to consider LCG and ATLAS versions of boost.

andiwand:
> Also needs to consider LCG and ATLAS versions of boost.

I ran into multiple problems so I added the concept of a "recommended" version. So I do not change the minimal version but print a warning if it is below the recommended version. If we download boost ourselves I do get the recommended version

benjaminhuth (approved):
looks good!

Examples/Algorithms/Fatras/src/FatrasSimulation.cpp
```diff
@@ -237,16 +239,36 @@ ActsExamples::ProcessCode ActsExamples::FatrasSimulation::execute(
   ACTS_DEBUG(simHitsUnordered.size() << " simulated hits");
 
   // order output containers
+#if BOOST_VERSION >= 107800
```
paulgessinger, line 242: Just for reference: I think in Core we shouldn't this but in the examples it's absolutely fine!

PR #2171, opened 2023-06-01
feat: Support multi-output models in OnnxRuntimeBase

### LINKED ISSUE #2170: Support multi-output in ONNX plugin
The current implementation of OnnxRuntimeBase assumes single-input & single-output nodes. However multi-output models are increasingly used in practice, for example when training a model with auxiliary objectives. Looking at the code, it should not be extremely hard to generalize our code to support this use-case.

tagging @benjaminhuth @Corentin-Allaire  for thoughts

gagnonlg:
As mentioned in #2170, multi-output models are increasingly used, for instance when training models with multiple auxiliary objectives. However our ONNX plugin does not support this at the moment. 

This PR adds a `runOnnxInferenceMultiOutput` that supports this usecase. The regular `runOnnxInference` methods can still be used on multi-output models -- the convention is that the output of the first node will be returned. This roughly matches the current behavior, but it was buggy in a subtle way -- the output of the first node was returned, but assuming the dimensions of the last node. This is now fixed.

The change *should* be backwards compatible, hopefully the CI agrees.

closes #2170


gagnonlg:
Failure in CI Bridge:
```
+ downloading G4EMLOW.8.2.tar.gz
...
curl: (23) Failure writing output to destination
- download of 'G4EMLOW.8.2.tar.gz' failed : CURL error 0
```
:cry:

Plugins/Onnx/src/OnnxRuntimeBase.cpp
```diff
@@ -69,28 +68,50 @@ std::vector<float> Acts::OnnxRuntimeBase::runONNXInference(
...
 
   // The first dim node should correspond to the batch size
   // If it is -1, it is dynamic and should be set to the input size
   if (inputNodeDims[0] == -1) {
     inputNodeDims[0] = batchSize;
   }
-  if (outputNodeDims[0] == -1) {
-    outputNodeDims[0] = batchSize;
+
+  bool ondMatch = true;
```
Corentin-Allaire, line 100: ond ? I think the acts coding convention is to a avoid abbreviation if possible (or make them explicit)

Plugins/Onnx/include/Acts/Plugins/Onnx/OnnxRuntimeBase.hpp
```diff
@@ -49,6 +49,12 @@ class OnnxRuntimeBase {
   std::vector<std::vector<float>> runONNXInference(
       NetworkBatchInput& inputTensorValues) const;
 
+  std::vector<std::vector<std::vector<float>>> runONNXInferenceMultiOutput(
+      NetworkBatchInput& inputTensorValues) const;
+
+  std::vector<std::vector<float>> runONNXInferenceMultiOutput(
+      std::vector<float>& inputTensorValues) const;
+
```
Corentin-Allaire, line 57: The equivalent only really exist because of a piece of legacy code somewhere (the MLTrackClassifier) that needs it (and because I didn't want to change that code). It is really inefficient to do the inference that way so I wouldn't even create this function.

Corentin-Allaire (commented):
This seems good to me. If it pass the CI then that means it doesn't change the output of the MLSolver.
It is up to you, but I think the effort to change all the use of runONNXInference to runONNXInferenceMultiOutput in ACTS is probably minor so if you don't mind doing it, only keeping the multi output one would make sense for me.

Plugins/Onnx/include/Acts/Plugins/Onnx/OnnxRuntimeBase.hpp
gagnonlg: Yeah I agree actually. Will remove.

gagnonlg:
> It is up to you, but I think the effort to change all the use of runONNXInference to runONNXInferenceMultiOutput in ACTS is probably minor so if you don't mind doing it, only keeping the multi output one would make sense for me.

I think we should still keep the "single output" runOnnxInference as a convenience, since I think the majority of usecases would still be single-output. But under the hood it is now implemented in terms of the multi-output version

Corentin-Allaire (approved):
Looks great, once we understand what is going on with the CI this can go in !

Corentin-Allaire:
Maybe rebasing will fix the Bridges issue ?

gagnonlg:
> Maybe rebasing will fix the Bridges issue ?

Let's try... :crossed_fingers:

Corentin-Allaire (approved):

paulgessinger:
Just fyi @gagnonlg (and others): if you @-mention folks in the PR description, this will propagate into the commit message, and will cause lots of notifications when the commit makes it into a release etc.

paulgessinger:
> Looks great, once we understand what is going on with the CI this can go in !

I think that was a transient failure due to the CI being unable to fetch G4 data content. This happens from time to time 🤷

PR #2174, opened 2023-06-02
refactor: Shorten long demangled names in sequencer

benjaminhuth:
This PR adds some simple name-shortening filters to the printouts of the demangled typenames of inputs/outputs in the Sequencer (remove `std::allocator` from `std::vector` if redundant, shorten `std::variant`s of measurements, strip namespaces, ...)
The aim is to not substantially reduce the information but avoid multi-line typenames, which works fairly well for me now.


andiwand (approved):
looks good to me

PR #2175, opened 2023-06-02
refactor: Export jacobian from RiddersPropagator

andiwand:
The `RiddersPropagator` calculates the jacobian numerically but did not expose this result.

### DIFF
--- a/Core/include/Acts/Propagator/RiddersPropagator.hpp
+++ b/Core/include/Acts/Propagator/RiddersPropagator.hpp
@@ -135,18 +135,18 @@ class RiddersPropagator {
   /// @param [in] derivatives Derivatives of a single parameter
   ///
   /// @return Boolean result whether a phi jump occured
-  bool inconsistentDerivativesOnDisc(
-      const std::vector<BoundVector>& derivatives) const;
+  static bool inconsistentDerivativesOnDisc(
+      const std::vector<BoundVector>& derivatives);
 
   /// @brief This function wiggles one dimension of the starting parameters,
   /// performs the propagation to a surface and collects for each change of the
   /// start parameters the slope
   ///
   /// @tparam options_t PropagatorOptions object
-  /// @tparam parameters+t Type of the parameters to start the propagation with
+  /// @tparam parameters_t Type of the parameters to start the propagation with
   ///
   /// @param [in] options Options do define how to wiggle
-  /// @param [in] startPars Start parameters that are modified
+  /// @param [in] start Start parameters which will be modified
   /// @param [in] param Index to get the parameter that will be modified
   /// @param [in] target Target surface
   /// @param [in] nominal Nominal end parameters
@@ -154,31 +154,30 @@ class RiddersPropagator {
   ///
   /// @return Vector containing each slope
   template <typename options_t, typename parameters_t>
-  std::vector<BoundVector> wiggleDimension(
-      const options_t& options, const parameters_t& startPars,
-      const unsigned int param, const Surface& target,
-      const BoundVector& nominal, const std::vector<double>& deviations) const;
+  std::vector<BoundVector> wiggleParameter(
+      const options_t& options, const parameters_t& start, unsigned int param,
+      const Surface& target, const BoundVector& nominal,
+      const std::vector<double>& deviations) const;
 
-  /// @brief This function propagates the covariance matrix
+  /// @brief This function fits the jacobian with the deviations and derivatives as input.
   ///
-  /// @param [in] derivatives Slopes of each modification of the parameters
-  /// @param [in] startCov Starting covariance
   /// @param [in] deviations Vector of deviations
+  /// @param [in] derivatives Slopes of each modification of the parameters
   ///
   /// @return Propagated covariance matrix
-  Covariance calculateCovariance(
-      const std::array<std::vector<BoundVector>, eBoundSize>& derivatives,
-      const Covariance& startCov, const std::vector<double>& deviations) const;
+  static Jacobian calculateJacobian(
+      const std::vector<double>& deviations,
+      const std::array<std::vector<BoundVector>, eBoundSize>& derivatives);
 
   /// @brief This function fits a linear function through the final state
   /// parametrisations
   ///
-  /// @param [in] values Vector containing the final state parametrisations
   /// @param [in] deviations Vector of deviations
+  /// @param [in] derivatives Vector of resulting derivatives
   ///
   /// @return Vector containing the linear fit
-  BoundVector fitLinear(const std::vector<BoundVector>& values,
-                        const std::vector<double>& deviations) const;
+  static BoundVector fitLinear(const std::vector<double>& deviations,
+                               const std::vector<BoundVector>& derivatives);
 
   /// Propagator
   propagator_t m_propagator;
--- a/Core/include/Acts/Propagator/RiddersPropagator.ipp
+++ b/Core/include/Acts/Propagator/RiddersPropagator.ipp
@@ -25,10 +25,12 @@ auto Acts::RiddersPropagator<propagator_t>::propagate(
 
   // Extract results from the nominal propagation
   auto nominalResult = std::move(nominalRet).value();
-  const BoundVector& nominalParameters =
-      nominalResult.endParameters->parameters();
+  assert(nominalResult.endParameters);
+  auto nominalFinalParameters = *nominalResult.endParameters;
+  const BoundVector& nominalFinalParameterValues =
+      nominalFinalParameters.parameters();
   // Use the curvilinear surface of the propagated parameters as target
-  const Surface& surface = nominalResult.endParameters->referenceSurface();
+  const Surface& surface = nominalFinalParameters.referenceSurface();
 
   // Steps for estimating derivatives
   std::vector<double> deviations = {-4e-4, -2e-4, 2e-4, 4e-4};
@@ -42,17 +44,21 @@ auto Acts::RiddersPropagator<propagator_t>::propagate(
 
   // Wiggle each dimension individually
   for (unsigned int i = 0; i < eBoundSize; i++) {
-    derivatives[i] =
-        wiggleDimension(opts, start, i, surface, nominalParameters, deviations);
+    derivatives[i] = wiggleParameter(opts, start, i, surface,
+                                     nominalFinalParameterValues, deviations);
   }
+
+  auto jacobian = calculateJacobian(deviations, derivatives);
+  nominalResult.transportJacobian = jacobian;
+
   if (start.covariance()) {
-    auto cov =
-        calculateCovariance(derivatives, *start.covariance(), deviations);
+    auto cov = jacobian * (*start.covariance()) * jacobian.transpose();
     // replace the covariance of the nominal result w/ the ridders covariance
-    auto& nom = *nominalResult.endParameters;
-    nom = CurvilinearTrackParameters(
-        nom.fourPosition(options.geoContext), nom.unitDirection(),
-        nom.absoluteMomentum(), nom.charge(), std::move(cov));
+    nominalResult.endParameters = CurvilinearTrackParameters(
+        nominalFinalParameters.fourPosition(options.geoContext),
+        nominalFinalParameters.unitDirection(),
+        nominalFinalParameters.absoluteMomentum(),
+        nominalFinalParameters.charge(), std::move(cov));
   }
 
   return ThisResult::success(std::move(nominalResult));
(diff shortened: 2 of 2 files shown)


andiwand:
There seems to be a workaround for `DiscSurface` I don't understand.

The code is still quite duplicated which I would like to change.

Also it would be nice if the deviations were configurable. And to check if a single deviation also works with the linearization.

cc @russofel

Core/include/Acts/Propagator/RiddersPropagator.hpp
```diff
@@ -75,10 +74,17 @@ class RiddersPropagator {
       typename result_type_helper<parameters_t, action_list_t>::type;
 
  public:
+  struct Config {
+    std::vector<double> deviations = {-4e-4, -2e-4, 2e-4, 4e-4};
+    std::vector<double> deviationsDisc = {{-3e-5, -1e-5, 1e-5, 3e-5}};
```
paulgessinger, line 79: Documentation of meaning?

Core/include/Acts/Propagator/RiddersPropagator.hpp
```diff
@@ -135,53 +157,53 @@ class RiddersPropagator {
   /// @param [in] derivatives Derivatives of a single parameter
   ///
   /// @return Boolean result whether a phi jump occured
-  bool inconsistentDerivativesOnDisc(
-      const std::vector<BoundVector>& derivatives) const;
+  static bool inconsistentDerivativesOnDisc(
```
paulgessinger, line 160: Good practice!

Core/include/Acts/Propagator/RiddersPropagator.hpp
```diff
@@ -135,53 +157,53 @@ class RiddersPropagator {
...
   ///
   /// @return Vector containing each slope
-  template <typename options_t, typename parameters_t>
-  std::vector<BoundVector> wiggleDimension(
-      const options_t& options, const parameters_t& startPars,
-      const unsigned int param, const Surface& target,
-      const BoundVector& nominal, const std::vector<double>& deviations) const;
+  template <typename propagator_options_t, typename parameters_t>
+  std::vector<BoundVector> wiggleParameter(
+      const propagator_options_t& options, const parameters_t& start,
```
paulgessinger, line 180: Any reason for the argument rename?

Core/include/Acts/Propagator/RiddersPropagator.hpp
```diff
@@ -135,53 +157,53 @@ class RiddersPropagator {
...
-  /// @param [in] derivatives Slopes of each modification of the parameters
-  /// @param [in] startCov Starting covariance
   /// @param [in] deviations Vector of deviations
+  /// @param [in] derivatives Slopes of each modification of the parameters
   ///
   /// @return Propagated covariance matrix
-  Covariance calculateCovariance(
-      const std::array<std::vector<BoundVector>, eBoundSize>& derivatives,
-      const Covariance& startCov, const std::vector<double>& deviations) const;
+  static Jacobian calculateJacobian(
```
paulgessinger, line 190: Does this actually change the meaning of the function? Where's the covariance calculated now?

Core/include/Acts/Propagator/RiddersPropagator.hpp
```diff
@@ -135,53 +157,53 @@ class RiddersPropagator {
...
+      const std::vector<double>& deviations,
+      const std::array<std::vector<BoundVector>, eBoundSize>& derivatives);
 
   /// @brief This function fits a linear function through the final state
   /// parametrisations
   ///
-  /// @param [in] values Vector containing the final state parametrisations
   /// @param [in] deviations Vector of deviations
```
paulgessinger, line 197: i.e. *wiggles*?

Core/include/Acts/Propagator/RiddersPropagator.hpp
```diff
@@ -135,53 +157,53 @@ class RiddersPropagator {
...
+  /// @param [in] derivatives Vector of resulting derivatives
   ///
   /// @return Vector containing the linear fit
-  BoundVector fitLinear(const std::vector<BoundVector>& values,
-                        const std::vector<double>& deviations) const;
+  static BoundVector fitLinear(const std::vector<double>& deviations,
+                               const std::vector<BoundVector>& derivatives);
 
-  /// Propagator
+  /// Underlying propagator
```
paulgessinger, line 204:
```suggestion
```

Core/include/Acts/Propagator/RiddersPropagator.ipp
```diff
@@ -17,42 +17,31 @@ auto Acts::RiddersPropagator<propagator_t>::propagate(
       action_list_t_result_t<CurvilinearTrackParameters,
                              typename propagator_options_t::action_list_type>>;
 
+  auto startWithoutCov = start;
```
paulgessinger, line 20: Explicit type would be preferable here I think, plus I usually try to add a comment to indicate the copy is intentional just to be safe.

Core/include/Acts/Propagator/RiddersPropagator.ipp
```diff
@@ -17,42 +17,31 @@ auto Acts::RiddersPropagator<propagator_t>::propagate(
...
-  // Steps for estimating derivatives
-  std::vector<double> deviations = {-4e-4, -2e-4, 2e-4, 4e-4};
-
-  // Allow larger distances for the oscillation
-  propagator_options_t opts = options;
-  opts.pathLimit *= 2.;
-
-  // Derivations of each parameter around the nominal parameters
-  std::array<std::vector<BoundVector>, eBoundSize> derivatives;
+  auto jacobian = wiggleAndCalculateJacobian(options, start, nominalResult);
```
paulgessinger, line 33: Generally, `auto` in combination with Eigen types has led to hard to debug bugs in the past (non evaluated expression templates etc), so ideally this would have an explicit type.

Core/include/Acts/Propagator/RiddersPropagator.ipp
```diff
@@ -17,42 +17,31 @@ auto Acts::RiddersPropagator<propagator_t>::propagate(
...
-  // Wiggle each dimension individually
-  for (unsigned int i = 0; i < eBoundSize; i++) {
-    derivatives[i] =
-        wiggleDimension(opts, start, i, surface, nominalParameters, deviations);
-  }
   if (start.covariance()) {
-    auto cov =
-        calculateCovariance(derivatives, *start.covariance(), deviations);
+    // use nominal parameters and Ridders covariance
+    auto cov = jacobian * (*start.covariance()) * jacobian.transpose();
```
paulgessinger, line 38: Ok I see, you refactored the function to only do the jacobian, and then the covariance can be calculated *inline* almost trivially. :+1:

Core/include/Acts/Propagator/RiddersPropagator.ipp
```diff
@@ -69,21 +58,50 @@ auto Acts::RiddersPropagator<propagator_t>::propagate(
...
+    const propagator_options_t& options, const parameters_t& start,
+    result_t& nominalResult) const -> Jacobian {
+  auto nominalFinalParameters = *nominalResult.endParameters;
+  // Use the curvilinear surface of the propagated parameters as target
+  const Surface& target = nominalFinalParameters.referenceSurface();
+
+  // TODO add to propagator options
   // Steps for estimating derivatives
-  std::vector<double> deviations = {-4e-4, -2e-4, 2e-4, 4e-4};
+  auto deviations = m_config.deviations;
```
paulgessinger, line 102: Copy on purpose? If it's just to allow changing between `deviations` and `deviationsDisc`, maybe a pointer is better suited?

Core/include/Acts/Propagator/RiddersPropagator.ipp
```diff
@@ -105,33 +123,23 @@ auto Acts::RiddersPropagator<propagator_t>::propagate(
...
-        if (inconsistentDerivativesOnDisc(deriv)) {
-          // Set covariance to zero and return
-          // TODO: This should be changed to indicate that something went
-          // wrong
-          return ThisResult::success(std::move(nominalResult));
-        }
-      }
+
+  // Test if target is disc - this may lead to inconsistent results
+  if (target.type() == Surface::Disc) {
```
paulgessinger, line 131: Btw, does this not also apply to cylinder surfaces?

paulgessinger (commented):
Useful change!

Core/include/Acts/Propagator/RiddersPropagator.hpp
andiwand: just out of consistency - I think we use this template arg name in the other cases

Core/include/Acts/Propagator/RiddersPropagator.hpp
andiwand: before my changes we calculated the jacobian, then the covariance and just dropped the jacobian. now I return the jacobian and do the sandwitch in the `propagate` function

Core/include/Acts/Propagator/RiddersPropagator.hpp
andiwand: yes exactly - naming is not great here

Core/include/Acts/Propagator/RiddersPropagator.ipp
```diff
@@ -197,42 +209,44 @@ Acts::RiddersPropagator<propagator_t>::wiggleDimension(
...
 
 template <typename propagator_t>
 Acts::BoundVector Acts::RiddersPropagator<propagator_t>::fitLinear(
-    const std::vector<Acts::BoundVector>& values,
-    const std::vector<double>& deviations) const {
-  BoundVector A;
-  BoundVector C;
-  A.setZero();
-  C.setZero();
+    const std::vector<double>& deviations,
```
russofel, line 233: Do you know where the formulas for this function come from?

Core/include/Acts/Propagator/RiddersPropagator.ipp
paulgessinger: Yeah you're right the dangerous area is only along the phi boundary along z, but not around the center like for a disk.

Core/include/Acts/Propagator/RiddersPropagator.ipp
```diff
@@ -105,33 +127,23 @@ auto Acts::RiddersPropagator<propagator_t>::propagate(
...
-    nom = BoundTrackParameters(nom.referenceSurface().getSharedPtr(),
-                               nom.parameters(), nom.charge(), std::move(cov));
   }
-  return ThisResult::success(std::move(nominalResult));
+
+  return calculateJacobian(*deviations, derivatives);
 }
 
 template <typename propagator_t>
 bool Acts::RiddersPropagator<propagator_t>::inconsistentDerivativesOnDisc(
```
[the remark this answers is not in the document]
andiwand: I think the problem is that the coordinates on the disc (polar coordinates) are not continuous because `2 pi = 0`

Core/include/Acts/Propagator/RiddersPropagator.ipp
andiwand: which one? the linear fit? I think this is a classical linear regression https://en.wikipedia.org/wiki/Linear_regression#Least-squares_estimation_and_related_techniques (without having it checked)

not sure how effective this is for estimating the derivative because I would not expect a slope naively ... but okay

Core/include/Acts/Propagator/RiddersPropagator.hpp
```diff
@@ -75,10 +74,19 @@ class RiddersPropagator {
       typename result_type_helper<parameters_t, action_list_t>::type;
 
  public:
+  struct Config {
+    /// Set of deltas which will be added to the nominal track parameters
+    std::vector<double> deviations = {-4e-4, -2e-4, 2e-4, 4e-4};
+    /// See `deviations` - these are applied for disc surfaces
+    std::vector<double> deviationsDisc = {{-3e-5, -1e-5, 1e-5, 3e-5}};
```
andiwand, line 81:
```suggestion
    std::vector<double> deviationsDisc = {-3e-5, -1e-5, 1e-5, 3e-5};
```

andiwand:
had a brief chat with @russofel - we think that there may be more things we could improve but that should be done in future PRs.

for reference:
- higher order derivative approximation should be done with a different method - maybe we should make it configurable?
- the fix for disc surfaces looks odd and should be revisited / refactored

from my side it is ready to merge

russofel (approved):
Seems good!

paulgessinger (approved):

PR #2177, opened 2023-06-03
fix: typo in ACTS_USE_SYSTEM_ACTSVG CMakeLists.txt

wdconinc:
Since `ACTS_USE_SYSTEM_ACTSVG` is defined as the option to "Use the ActSVG system library", the variable should be `ACTS_USE_SYSTEM_ACTSVG` and not have a trailing `N`.

(Also, removed a space in the line before for consistence formatting.)


andiwand (approved):
thank you!

PR #2180, opened 2023-06-06
feat: Multilayer structure builder

dimitra97:
This is a PR that provides the code that creates a multi wire detector volume using the surfaces from the gdml file with rhe tubes of the muon spectrometer geometry.
It uses the new interface where the` external builder` is the `volume builder` and the i`nternal builder` is the `layer structure builder`
The gdml file has been changed a little bit, in order to be able to get directly the surfaces of one multilayer of a chamber.

I think as a next step, the implementation of the navigation delegate which considers the grid with the surfaces should be implemented, as part of the internal structure builder.
Before that, I would like your comments and any advice :)


Core/include/Acts/Detector/detail/IndexedSurfacesGenerator.hpp
```diff
@@ -59,7 +59,7 @@ struct IndexedSurfacesGenerator {
   /// @param rGenerator the reference generataor
   ///
   /// @return a SurfaceCandidateUpdator delegate
-  template <typename axis_generator, typename reference_generator>
+  template <typename axis_generator, typename reference_generator, template<typename> class IndexedGridSurfacesUpdator>
   SurfaceCandidatesUpdator operator()(
```
andiwand, line 63: just for me to understand - where do you need this functionality?

dimitra97: Here, the `IndexedSurfacesImpl` is only used as navigation delegeate for indexed surfaces, as I understand. I would like to make it as we can pass other delegate types for indexed surfaces (e.g. I am working on a delegate where specific indexed surfaces will be assigned to candidates by checking with intersection and not assign all the surfaces of the volume).

PR #2182, opened 2023-06-06
fix: Smearing-digitization with Geant4

benjaminhuth:
Geant4 hits may not be exactely on the surface. This PR adjusts the tolerance for the free-to-bound conversion to address this. Also bumps the corresponding log message to a warning.

The error still occurs around 40 times for 2000 muons... but at least we are not loosing around 50% of the hits anymore...

cc @asalzburger


Fatras/include/ActsFatras/Digitization/UncorrelatedHitSmearer.hpp
```diff
@@ -62,12 +64,18 @@ struct BoundParametersSmearer {
...
+
     // construct full bound parameters. they are probably not all needed, but it
     // is easier to just create them all and then select the requested ones.
     Acts::Result<Acts::BoundVector> boundParamsRes =
-        Acts::detail::transformFreeToBoundParameters(hit.position(), hit.time(),
-                                                     hit.unitDirection(), 0,
-                                                     surface, geoCtx);
+        Acts::detail::transformFreeToBoundParameters(
+            hit.position(), hit.time(), hit.unitDirection(), 0, surface, geoCtx,
+            tolerance);
```
andiwand, line 78: does it actually make sense to use `transformFreeToBoundParameters` here? the surface method might be sufficient no?

(sorry this is of course what was there before)

Fatras/include/ActsFatras/Digitization/UncorrelatedHitSmearer.hpp
```diff
@@ -9,7 +9,9 @@
 #pragma once
 
 #include "Acts/Definitions/TrackParametrization.hpp"
+#include "Acts/Definitions/Units.hpp"
```
andiwand, line 12: I guess this can go again?
```suggestion
```

andiwand (approved):
great - thank you!

Fatras/include/ActsFatras/Digitization/UncorrelatedHitSmearer.hpp
benjaminhuth: can we leave it as it is?

andiwand (approved):

PR #2183, opened 2023-06-07
fix: include array in VectorHelpers.hpp

CarloVarni:
missing include causing compilation issues


LuisFelipeCoelho (commented):
Thank you @CarloVarni! just missing the format

beomki-yeo (approved):
LGTM as long as the format is fixed

andiwand (approved):

PR #2184, opened 2023-06-07
feat: Step merger for Geant4 and digitization energy threshold

benjaminhuth:
This PR does tow things:

* Add a step merger to the Geant4 sim. This fixes the way how we record hits: Geant4 can do multiple steps through the material, which need to be combined to one single hit. Bevor, we recorded each step as hit.

* Add a energy threshold for the hit acceptance to the digitization: This allows the digitization to reject hits with a very small energy deposit. The default value for this is `1000 * 3.65 eV` (1000 electron-hole pairs for activation).

Together, this reduces the all-over hits, that the `Geant4Simulation` produces on a ttbar event drastically. However, the number of "target hits" (hits of particles with pT > 500 MeV and nhits > 3 in this case) stays roughly the same.

image

The energy threshold seems reasonable: In a particle-gun muon-only run in the ODD we see that almost all muons pass the threshold, but a lot low energy deposit electron hits are rejected by the digitization (energy deposit is negative, since it is stored as energy loss of the particle):

image

### DIFF
--- a/Core/include/Acts/Seeding/SeedFinder.hpp
+++ b/Core/include/Acts/Seeding/SeedFinder.hpp
@@ -32,7 +32,7 @@
 
 namespace Acts {
 
-enum class SpacePointCandidateType : short { BOTTOM, TOP };
+enum class SpacePointCandidateType : short { eBottom, eTop };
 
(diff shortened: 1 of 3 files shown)


andiwand (approved):
:+1:

PR #2185, opened 2023-06-07
refactor: Rework Geant4 sim object lifetime

andiwand:
This PR is an attempt to rework our Geant4 examples implementation in order to make the lifetime of the objects more explicit, simplify the instantiation of simulation and material recording and to allow multiple Geant4 algorithms in a row.


Examples/Algorithms/Geant4/include/ActsExamples/Geant4/EventStore.hpp
```diff
@@ -0,0 +1,91 @@
...
+#include <unordered_map>
+#include <vector>
+
+#include "G4Types.hh"
+
+namespace ActsExamples {
+
+class WhiteBoard;
+
+/// Common event store for all Geant4 related sub algorithms
```
benjaminhuth, line 29: I would maybe consider only store a `shared_ptr` to this obj in the algorithm, and don't add an additional layer until it is needed.

Examples/Run/Geant4/Common/include/ActsExamples/Geant4/Geant4Common.hpp
benjaminhuth, line 1: I could also imagine dropping support for geant4 with the binaries at this point

benjaminhuth (commented):
A first round of comments, feel free to close etc.

Examples/Algorithms/Geant4/src/Geant4Manager.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  // Suppress the printing of physics information.
+#if G4VERSION_NUMBER >= 1100
+  G4HadronicParameters::Instance()->SetVerboseLevel(0);
+  G4HadronicProcessStore::Instance()->SetVerbose(0);
+  G4EmParameters::Instance()->SetIsPrintedFlag(true);
+#endif
+}
+
+std::shared_ptr<Geant4Handle> Geant4Manager::create(int logLevel,
```
benjaminhuth, line 80: maybe rename to `createHandle` to make super clear that this is not a factory of `Geant4Manager`.

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
```diff
@@ -9,151 +9,92 @@
...
-    G4FieldManager* fieldMgr = new G4FieldManager();
-    fieldMgr->SetDetectorField(m_cfg.magneticField);
-    fieldMgr->CreateChordFinder(m_cfg.magneticField);
+void ActsExamples::Geant4SimulationBase::commonInitialization() {
+  auto runManager = m_gean4Instance->runManager.get();
 
-    // Propagate down to all childrend
-    g4World->GetLogicalVolume()->SetFieldManager(fieldMgr, true);
+  // Set the detector construction
+  {
```
benjaminhuth, line 63: Extra scope necessary?

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
```diff
@@ -9,151 +9,92 @@
...
-  // Set the magnetic field
-  if (m_cfg.magneticField != nullptr) {
-    ACTS_INFO("Setting ACTS configured field to Geant4.");
-    // Get the g4World cache
-    G4VPhysicalVolume* g4World = m_cfg.detectorConstruction->Construct();
-    /// Set the field ot the G4Field manager
-    G4FieldManager* fieldMgr = new G4FieldManager();
-    fieldMgr->SetDetectorField(m_cfg.magneticField);
-    fieldMgr->CreateChordFinder(m_cfg.magneticField);
+void ActsExamples::Geant4SimulationBase::commonInitialization() {
```
benjaminhuth, line 59: Is there a  reason why this is not done in the Constructor of `Geant4SimulationBase`?

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
```diff
@@ -164,10 +105,15 @@ ActsExamples::ProcessCode ActsExamples::Geant4Simulation::execute(
 
   ACTS_DEBUG("Sending Geant RunManager the BeamOn() command.");
   // Start simulation. each track is simulated as a separate Geant4 event.
-  m_cfg.runManager->BeamOn(1);
+  runManager->BeamOn(1);
 
   // Since these are std::set, this ensures that each particle is in both sets
-  assert(eventData.particlesInitial.size() == eventData.particlesFinal.size());
+  if (eventData.particlesInitial.size() != eventData.particlesFinal.size()) {
```
benjaminhuth, line 111: Hmm why turning this into a warning? This indicates severe logical issue in the code from my perspective and should lead to immediate failure.

Examples/Algorithms/Geant4/src/ParticleTrackingAction.cpp
```diff
@@ -24,7 +24,7 @@ ActsExamples::ParticleTrackingAction::ParticleTrackingAction(
 
 void ActsExamples::ParticleTrackingAction::PreUserTrackingAction(
     const G4Track* aTrack) {
-  auto& eventData = EventStoreRegistry::eventData();
+  auto& eventData = *m_cfg.eventStore;
```
benjaminhuth, line 27: *In principle* we could add a member function returning a reference to eventData and then just add `()` to the places where this is used... this would be nice, but is of course not necessary

Of course this holds for all classes with similar interface.

Examples/Python/src/Geant4Component.cpp
```diff
@@ -51,196 +36,89 @@ using namespace pybind11::literals;
...
+    ACTS_PYTHON_MEMBER(recordHitsOfSecondaries);
+    ACTS_PYTHON_MEMBER(keepParticlesWithoutHits);
+    ACTS_PYTHON_STRUCT_END();
+  }
 
   {
-    using Detector = ActsExamples::Telescope::TelescopeDetector;
-    using DetectorConstruction =
-        ActsExamples::Telescope::TelescopeG4DetectorConstruction;
+    using Algorithm = Geant4MaterialRecording;
```
benjaminhuth, line 86: Probably this algorithm should work with the declare-algorithm-macro. But is there a `geant4Handle` accessor missing?

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
```diff
@@ -181,41 +127,213 @@ ActsExamples::ProcessCode ActsExamples::Geant4Simulation::execute(
...
+    m_fieldManager->SetDetectorField(m_magneticField.get());
+    m_fieldManager->CreateChordFinder(m_magneticField.get());
+
+    // Propagate down to all childrend
+    g4World->GetLogicalVolume()->SetFieldManager(m_fieldManager.get(), true);
+  }
+
+  // An ACTS TrackingGeometry is provided, so simulation for sensitive
+  // detectors is turned on - they need to get matched first
+  if (cfg.trackingGeometry) {
```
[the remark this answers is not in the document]
andiwand: it allows you to just simulate without mapping. not sure if that is used but I would not remove it for now

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
andiwand: the problem is that we need a run manager for this and it is coupled to the physics list now which depends on sim or material recording. I would also like to use the constructor for this but I dont think this is possible right now

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
andiwand: I will change it to `ACTS_ERROR`

benjaminhuth (approved):

PR #2189, opened 2023-06-09
fix: FPE in FATRAS hit direction

paulgessinger:
I've encountered FPEs in the calculation of the average direction of FATRAS hits.

Currently we do this:
- Divide four-momenta before and after divide by 2x the norm of the three-momentum vector
- Add both four-vectors, then take only the three-momentum out, then normalize thos

This fails if the norm of either of the three momenta is invalid or zero. I think Eigen has checks against this but we do the division manually.

I think this is equivalent to:
- Getting the normalized three-momenta
- Adding them and dividing by 2, then normalize again

Thoughts?

### DIFF
--- a/Fatras/include/ActsFatras/EventData/Hit.hpp
+++ b/Fatras/include/ActsFatras/EventData/Hit.hpp
@@ -88,9 +88,9 @@ class Hit {
   }
   /// Average normalized particle direction vector through the surface.
   Vector3 unitDirection() const {
-    auto dir0 = m_before4 / (2 * m_before4.segment<3>(Acts::eMom0).norm());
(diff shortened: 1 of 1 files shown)


andiwand (approved):

Fatras/include/ActsFatras/EventData/Hit.hpp
```diff
@@ -88,9 +88,9 @@ class Hit {
   }
   /// Average normalized particle direction vector through the surface.
   Vector3 unitDirection() const {
```
andiwand, line 90: just for reference - it might be better to not average the direction but stick to either before or after. I think for digitization we should use before?

PR #2192, opened 2023-06-09
fix: and refactor Exa.TrkX plugin

benjaminhuth:
This fixes the plugin by adding an important feature that was lost during the last refactoring. It is now possible to run run the edge classification *undirected* (feeding a bi-directional graph into the model).
Besides this, the code is refactored (logger for each module, variable names, added some warnings and debug output...)


andiwand (approved):
looks good - can you add a description to the PR?

PR #2195, opened 2023-06-12
feat: Seed selector 

CarloVarni:
Add object for selecting seeds. By default it applies no selection, but users can specify whatever criteria they desire. An example (just as a proof of concept) is provided in the SeedingAlgorithm, where we check the quality of the seeds


CarloVarni:
/cc @noemina @paulgessinger @timadye 

what is your feedback on this idea?

noemina:
I like it :)

Examples/Algorithms/TrackFinding/src/SeedingAlgorithm.cpp
```diff
@@ -1,4 +1,4 @@
-// This file is part of the Acts project.
+// This file is part of the Acts project.OA
```
timadye, line 1:
```suggestion
// This file is part of the Acts project.
```

Core/include/Acts/TrackFinding/SeedSelector.hpp
```diff
@@ -0,0 +1,66 @@
...
+  public:
+    SeedSelector(SeedSelectorOptions<external_spacepoint_t>&& options);
+    SeedSelector(const SeedSelector&) = delete;
+    SeedSelector(SeedSelector&&) = delete;
+    SeedSelector& operator=(const SeedSelector&) = delete;
+    SeedSelector& operator=(SeedSelector&&) = delete;
+    
+    ~SeedSelector() = default;
+
+    bool passesQualitySelection(const Acts::Seed<external_spacepoint_t>&) const;
```
timadye, line 49: I guess this could select on more than just the quality, eg. eta, pT etc. If so, would it be better to call this `passesQualitySelection` or even just `selected`?

timadye:
Sorry, I didn't look into the details of the rest of the seeding algorithm, so I am confused by this.

What is the purpose of the `SeedSelector` class? It seems to just pass on the test to the user-defined selector function. That would be useful if it were called inside the Core seeding code, but it is called from the Example algorithm, which doesn't need the delegate but could call `MySelector::passesQualitySelection` directly.

I guess there is more to come that would fill in the gap. Are you planning a full example (with a better name than `MySelector` 😄) with parameters (eg. quality cut value) that can be added to the full-chain? How much of that can go in the Core?

Sorry for the bother on a mere `Draft` PR.

Core/include/Acts/TrackFinding/SeedSelector.hpp
CarloVarni: good suggestion

CarloVarni:
Hi @timadye indeed you raise a good point. My problem here is how to better integrate a SeedSelection object inside the Seed Finding/Filter. For now I have extended the SeedFinder functionality adding a new method that will perform the entire seeding procedure on all the grid bin groups and then apply the selection.
This is only an addition to the already existing functionality, so it should break nothing

There is still a problem to be solved oh how to retrieve the correct index associated to a space point... that may require an internal change to the seed EDM to store a triplet of indices as well.

paulgessinger:
@timadye @CarloVarni are you coming back to this?

PR #2199, opened 2023-06-13
fix: changed keyword argument minMeasurements to nMeasurementsMin, solving…

CarloVarni:
update the full-itk example as a result of the changes in https://github.com/acts-project/acts/pull/2110 that were not propagated to this ITk script


LuisFelipeCoelho (approved):
thanks!

PR #2204, opened 2023-06-13
refactor: Allow TrackState copyFrom for different MTJs

paulgessinger:
This allows copying between MTJ backends by relaxing the template arguments.

PR #2205, opened 2023-06-13
refactor: TrackProxy::copyFrom copies track states

paulgessinger:
This uses a newly added function to reverse track states to create the track states with reverted order and then flip the order around to produce a good copy.

Blocked by:
- https://github.com/acts-project/acts/pull/2204


paulgessinger:
@goetzgaycken

goetzgaycken:
Great ! Seems to work. I suppose the  `tipIndex() = other.tipIndex();` in the version without this patch causes the invalid track states.

paulgessinger:
Ok it seems I need to debug a bit more.

paulgessinger:
Ok I think I fixed the crash, let's see if the CI agrees.

PR #2209, opened 2023-06-14
ci: Add more labels to the labeler bot

CarloVarni:
This adds new labels to the labeler plot. We introduce labels for identifying which state of the reconstruction sequence the PR is affecting, as well a label to check if a reference file has been changed (i.e. a change of performance)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -12,3 +12,21 @@
   - Plugins/**
 'Component - Documentation':
   - docs/**
+'Changes Performance':
+  - CI/physmon/reference/**
+'Event Data Model':
+  - '**/*EventData*/**'
+'Clustering':
+  - '**/*Clusterization*/**'
+'SP formation':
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -12,3 +12,21 @@
...
+'Seeding':
+  - '**/*Seeding*/**'
+'Track Finding':
+  - '**/*TrackFinding*/**'
+'Track Fitting':
+  - '**/*TrackFitting*/**'
+'Vertexing':
+  - '**/*Vertexing*/**'
+'Ambiguity Resolution':
+  - '**/*AmbiguityResolution*/**'
```
paulgessinger, line 32: Should this only apply to Core or also to the Examples?

paulgessinger: Fair enough. I don't have a strong opinion, was just thinking about it.

andiwand (approved):

PR #2211, opened 2023-06-15
fix: Add missing term in `EstimateTrackParamsFromSeed`

beomki-yeo:
I will need a crosscheck from people who are interested in this issue

`invTanTheta` is used to obtain the momentum from transverse momentum - so let me explain how we calculate the momentum in this function. The function defines a frame whose origin is at bottom spacepoint and transverse plane is defined perpendicular to the B field. Saying that $(x,y,z)$ is the position of top spacepoint in the new frame and $R$ is the radius of circle projection of helix and $\theta$ is the xy-projected angle between bottom spacepoint and top spacepoint, we have the following:

$$ \frac{q}{p} = \frac{q}{p_T}  \times \frac{1}{\sqrt{1 + (\frac{z}{R\theta})^2}} $$

$\frac{z}{R\theta}$ is equivalent to `invTanTheta` variable in the function.

If we combine with $(2R \sin{\frac{\theta}{2}} = \sqrt{x^2+y^2})$, we get the following:

$$ \frac{z}{R\theta}= \frac{z}{2R \sin^{-1}{\frac{\sqrt{x^2+y^2}}{2R}}} \approx \frac{z}{\sqrt{x^2+y^2} + \frac{(x^2+y^2)^{3/2}}{24 R^2}} = \frac{z \sqrt{\frac{1}{x^2+y^2}}}{1+\frac{x^2 + y^2}{24 R^2}} $$

In the current function, the factor of 24 seems to be missing


Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp
```diff
@@ -237,7 +237,7 @@ std::optional<BoundVector> estimateTrackParamsFromSeed(
   ActsScalar rn = local2.x() * local2.x() + local2.y() * local2.y();
   // The (1/tanTheta) of momentum in the new frame,
   ActsScalar invTanTheta =
-      local2.z() * std::sqrt(1. / rn) / (1. + rho * rho * rn);
+      local2.z() * std::sqrt(1. / rn) / (1. + rho * rho * rn / 24.f);
```
CarloVarni, line 240: Since 1/24.f is a const what about defining it as a static const value so that you avoid doing a division all the time?

beomki-yeo: Makes sense - I made 1./24 a static const double

beomki-yeo:
As a note, the reference has also the factor of 1/.24 which is approximated as 0.04

https://acode-browser.usatlas.bnl.gov/lxr/source/athena/InnerDetector/InDetRecTools/SiTrackMakerTool_xk/src/SiTrackMaker_xk.cxx
```c++
/// connecting the three points.
double A  = v2/(u2-u1)                 ;  /// Keep in mind that v1 == 0
double B  = 2.*(v2-A*u2)               ;  /// From inserting A into linear equation. Note that Igor sneaks in a factor two here
double C  = B/sqrt(1.+A*A)             ;  /// Curvature estimate. (2R)²=(1+A²)/b² => 1/2R = b/sqrt(1+A²) = B / sqrt(1+A²).
double T  = z2*sqrt(r2)/(1.+.04*C*C*rn);  /// estimate of the track dz/dr (1/tanTheta), corrected for curvature effects
```

Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp
```diff
@@ -236,8 +236,9 @@ std::optional<BoundVector> estimateTrackParamsFromSeed(
   // frame
   ActsScalar rn = local2.x() * local2.x() + local2.y() * local2.y();
   // The (1/tanTheta) of momentum in the new frame,
+  static constexpr double G = 1. / 24;
```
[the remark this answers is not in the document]
beomki-yeo: OK

CarloVarni:
Hi @beomki-yeo there are changes in the CKF and Vertex from the physmon, which I suppose are expected. You have to update the reference files and hash files as well

beomki-yeo:
I am checking out the report but [pull_eLOC1_fit](https://herald.harbinger.gessinger.dev/view/acts-project/acts/752416054/tracksummary_ckf_orthogonal.html#pull_eLOC1_fit) and [pull_eTHETA_fit](https://herald.harbinger.gessinger.dev/view/acts-project/acts/752416054/tracksummary_ckf_orthogonal.html#pull_eTHETA_fit) changes a lot for both orthogonal and regular seeding :/ It might be because that the term with 1/24. is not negligible with low pT particles.

andiwand:
is it necessary to update all reference files? usually we just overwrite the ones which fail the CI

beomki-yeo:
I don't know which one is making difference so I copied all root files from here: https://herald.harbinger.gessinger.dev/view/acts-project/acts/752630181/

andiwand:
you can check it here https://herald.harbinger.gessinger.dev/view/acts-project/acts/752630181/summary.html

so the following would be sufficient

- `CI/physmon/reference/tracksummary_ckf_orthogonal_hist.root`
- `CI/physmon/reference/tracksummary_ckf_seeded_hist.root`
- `CI/physmon/reference/tracksummary_ckf_truth_estimated_hist.root`

beomki-yeo:
Umm OK. But I think the others which have different byte sizes still should be updated 
I rollback the `performance_amvf_truth_smeared_hist.root` and `tracksummary_ckf_truth_smeared_hist.root` which have the same byte size

andiwand:
personally I would roll back all which do not fail the CI. maybe @paulgessinger can clarify

beomki-yeo:
Btw do you know when those file were changed lastly? I didnt change them when i made #2132 which would affect those files

paulgessinger:
I'd have to consult the commit log to check. I agree with @andiwand in that I would recommend we only update the files that are causing failures and keep the other ones as is.  

The physmon histogram comparison is intentionally set up to try and only fail the job if the histograms change in a meaningful way, not necessarily when they change at all, so it's expected that they don't have to be updated for every single change.

beomki-yeo:
OK

beomki-yeo:
:cry: Why does Docs CI fail

stephenswat:
Just an interesting little tidbit:

$$\frac{z}{\sqrt{x^2 + y^2} + \frac{(x^2 + y^2)^\frac{3}{2}}{24R^2}} = \frac{24R^2z}{\sqrt{x^2 + y^2}(24R^2 + x^2 + y^2)}$$

Which involves fewer divisions and allows more common subexpression elimination, so better performance (see https://quick-bench.com/q/mdh8wF983P5q9UQDyqXH8t_dZKU) and possibly better numerical stability.

andiwand:
> cry Why does Docs CI fail

random failure I restarted the workflow

CarloVarni:
> Just an interesting little tidbit:
> 
> zx2+y2+(x2+y2)3224R2=24R2zx2+y2(24R2+x2+y2)
> 
> Which involves fewer divisions and allows more common subexpression elimination, so better performance (see https://quick-bench.com/q/mdh8wF983P5q9UQDyqXH8t_dZKU) and possibly better numerical stability.

@stephenswat are you suggesting to modify this PR with this new computation? Doing it in a follow-up PR? Or something else?

stephenswat:
> @stephenswat are you suggesting to modify this PR with this new computation? Doing it in a follow-up PR? Or something else?

I'm not really suggesting anything right now, I was just messing around with some algebra. But we can look into doing this in a separate PR if it makes sense. :slightly_smiling_face:

CarloVarni (approved):

PR #2214, opened 2023-06-15
refactor: Explicit and type safe navigation delegates

andiwand:
In this PR I try to make the navigation delegates more explicit and type safe by splitting the interface of volume finders and surface candidate lookups. That way the navigation state is just a view to the delegates and cannot be modified which also lowers bug risk IMO.

- rename `DetectorVolumeUpdator` to `DetectorVolumeFinder`
- rename `SurfaceCandidatesUpdator` to `SurfaceCandidatesDelegate`
- add uniform utility functions to construct delegates
- reorganize file structure of new navigation and new geometry


andiwand:
TODO
- run unit tests
- add doc
- add unit tests

Core/include/Acts/Navigation/NavigationHelpers.hpp
```diff
@@ -0,0 +1,69 @@
...
+/// Helper method to update the candidates (portals/surfaces),
+/// this can be called for initial surface/portal estimation,
+/// but also during the navigation to update the current list
+/// of candidates.
+///
+/// @param gctx is the Geometry context of this call
+/// @param nState [in,out] is the navigation state to be updated
+///
+/// @todo for surfaces skip the non-reached ones, while keep for portals
+inline static void updateCandidates(
```
paulgessinger, line 28: Do these need to be inline?

Core/src/Detector/LayerStructureBuilder.cpp
```diff
@@ -49,17 +49,20 @@ namespace {
 ///
 /// @return a configured surface candidate updators
 template <Acts::detail::AxisBoundaryType aType>
-Acts::Experimental::SurfaceCandidatesUpdator createUpdator(
+Acts::Experimental::SurfaceCandidatesDelegate createUpdator(
     const Acts::GeometryContext& gctx,
-    const std::vector<std::shared_ptr<Acts::Surface>>& lSurfaces,
-    const std::vector<size_t>& assignToAll,
+    std::vector<std::shared_ptr<Acts::Surface>> lSurfaces,
+    std::vector<size_t> assignToAll,
```
paulgessinger, line 55: Just need to make sure we don't copy these needlessly.

Core/src/Detector/VolumeStructureBuilder.cpp
```diff
@@ -175,8 +175,9 @@ Acts::Experimental::VolumeStructureBuilder::construct(
     default:
       break;
   }
+
   // Return the transform, the volume bounds, and some default portal
   // generators
   return {Transform3(m_cfg.transform * eTransform), std::move(volumeBounds),
-          defaultPortalGenerator()};
+          makePortalGenerator<const DefaultPortalGenerator>()};
```
paulgessinger, line 182: Just force `const`?

paulgessinger (commented):
I think the naming is clearer this way and it makes sense to me to separate these two.

Core/src/Detector/VolumeStructureBuilder.cpp
andiwand: what do you mean by force? I wondered if I should add it to the helper directly. but at the same time I wondered why we require const because the method being called is const anyways :thinking:

Core/src/Detector/LayerStructureBuilder.cpp
andiwand: agreed - I just think by value gives you more flexibility when you are sinking the param anyways

Core/include/Acts/Navigation/NavigationHelpers.hpp
andiwand: I guess not - I just saw that removing that would lead to linker errors in some cases but maybe that was also in combination with templates. the question is also if these kind of methods really have to have their impl in the header

asalzburger:
@andiwand how do we go about this?

andiwand:
> @andiwand how do we go about this?

I can update this again - in principle it is ready to review just the CI is failing

There is one detail about the grids I was unsure about. I had to touch it without really knowing what is going on behind the scenes

andiwand:
this will be followed up in other, smaller PRs

PR #2217, opened 2023-06-19
fix: GCC segfault in TrackTests.cpp

paulgessinger:
We've observed the Track unit test executable to SEGFAULT during compilation. 
This seems to be a combination of different things in this compilation unit. I haven't been able to fully figure out if there's a single reason it gets triggered, but in local testing I've found that the problem goes away when I split out this test function into a separate compilation unit. I'm doing that here, and I need to make a small change to our CMake config to make that work.

### DIFF
--- a/Tests/UnitTests/CMakeLists.txt
+++ b/Tests/UnitTests/CMakeLists.txt
@@ -3,14 +3,16 @@
 # the common libraries which are linked to every unittest can be
 # extended by setting the `unittest_extra_libraries` variables before
 # calling the macro.
-macro(add_unittest _name)
+macro(add_unittest _name _source)
   # automatically prefix the target name
   set(_target "ActsUnitTest${_name}")
-  add_executable(${_target} ${ARGN})
+  add_executable(${_target} ${_source})
   # define required BOOST_TEST_... macros here to ensure consistent names
   target_compile_definitions(
     ${_target}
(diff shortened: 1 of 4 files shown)


andiwand (approved):

andiwand:
interesting interference

paulgessinger:
I think I have a fix already so let's hold this.

Tests/UnitTests/CMakeLists.txt
```diff
@@ -3,14 +3,16 @@
...
   set(_target "ActsUnitTest${_name}")
-  add_executable(${_target} ${ARGN})
+  add_executable(${_target} ${_source})
   # define required BOOST_TEST_... macros here to ensure consistent names
   target_compile_definitions(
     ${_target}
-    PRIVATE "-DBOOST_TEST_DYN_LINK" "-DBOOST_TEST_MODULE=${_target}")
+    PRIVATE "-DBOOST_TEST_DYN_LINK")
+  set_source_files_properties(${_source} PROPERTIES 
+    COMPILE_DEFINITIONS "BOOST_TEST_MODULE=${_target}")
```
andiwand, line 15: what is the change here?

andiwand (approved):
this is very odd but I guess what counts is that it works

Tests/UnitTests/CMakeLists.txt
paulgessinger: `BOOST_TEST_MODULE` is only applied to that single source file now, instead of all source files in the target.

PR #2222, opened 2023-06-20
chore: Align event generation in physics monitoring and ODD full chain

andiwand:
Pulled these changes out of https://github.com/acts-project/acts/pull/2086 in order to factorize the difference we see from the event gen and from the interactions turned on.


CI/physmon/workflows/physmon_vertexing.py
```diff
@@ -60,10 +60,13 @@ def run_vertexing(fitter, mu, events):
             s,
             MomentumConfig(1.0 * u.GeV, 10.0 * u.GeV, transverse=True),
             EtaConfig(-3.0, 3.0),
+            PhiConfig(0.0, 360.0 * u.degree),
```
paulgessinger, line 63: This is not the default?

andiwand: it is but I decided not to trust the defaults :smile:

paulgessinger:
Looks good to me. Let's update the references and then merge.

PR #2223, opened 2023-06-20
feat: Enable resolved stack traces in clang

paulgessinger:
This was complicated by having to locate the GCC backtrace header and supplying it to the build.


Plugins/FpeMonitoring/CMakeLists.txt
```diff
@@ -28,6 +30,67 @@ else()
...
+          COMPILE_DEFINITIONS -DBOOST_STACKTRACE_USE_BACKTRACE
+          OUTPUT_VARIABLE __OUTPUT)
+
+
+      if(_backtrace_default_header)
+        message(CHECK_PASS "yes")
+      else()
+        message(CHECK_FAIL "no")
+
+        file(GLOB hints "/usr/lib/gcc/*/*/include")
```
benjaminhuth, line 58: Not sure if I understand this code correctely: Do we search here for the file, and if found, we use a hardcoded path?
Why not something like `if(EXISTS "/usr/lib/gcc/x86_64-linux-gnu/11/include/backtrace.h")`?

paulgessinger: Not quite. The `find_file` will look in the standard paths. I'm using this glob to supplement the search paths in `find_file` with all folder matching `/usr/lib/gcc/*/*/include`, because that's where @andiwand and I found this header locally. 

Does that make sense?

benjaminhuth: Hmm not 100% yet, because I would excpect then
```cmake
set(backtrace_include "-DBOOST_STACKTRACE_BACKTRACE_INCLUDE_FILE=${backtrace_header})
```
and not the hardcoded path...

paulgessinger: Ah right, I forgot to change that. Changed now.

benjaminhuth (approved):

PR #2236, opened 2023-06-22
feat: Improve Acts::zip to take arbitrary number of ranges

benjaminhuth:
This allows finally something like this for an arbitray amount of ranges / containers:
```c++
for(const auto &[a, b, c] : Acts::zip(ra, rb, rc)) {
  // do something
}
```

Also adds unit test for this.


paulgessinger (approved):
:+1:

Issue #2226, opened 2023-06-21
Bug: Combinatorial Kalman filter fails during track finding


felix-russo:
When running `CI/physmon/workflows/physmon_vertexing.py`, I encountered the following error (see screenshot). I don't know how much of a problem that is, but I wanted to mention it here!
![image](https://github.com/acts-project/acts/assets/72298366/ebee4c1c-68e3-4e63-9f6d-d2fefd288a06)

paulgessinger:
It's definitely something we'll need to follow up.
Could you try to narrow down where exactly this gets triggered? I'm not sure if this could be a navigation failure, because I wouldn't expect this to cause the global position not to end up on surface. Could it be related to the global-to-local changes that @andiwand made recently?

andiwand:
I would think so yeah - it might have the same origin. So that would be another case were we do not properly propagate back to the perigee and the line surface was happy with that before my change

felix-russo:
When executing `physmon_ckf_tracking.py`, I saw the same error message.

This gets triggered in the function `globalToLocal` in `Linesurface.cpp`!

paulgessinger:
@asalzburger mentions that in case there's an RK step in-between the creation of the parameters and the global to local, a slightly changed momentum could affect if the parameters are on surface or not.

andiwand:
this is what I tried to quick fix here https://github.com/acts-project/acts/pull/2239

paulgessinger:
@felix-russo can you verify that you see this as fixed on your end?

felix-russo:
@paulgessinger all working!

PR #2239, opened 2023-06-23
fix: Quick fix `LineSurface` tolerance inconsistency in `globalToLocal`

andiwand:
This is meant as a quick fix for the problem I am experiencing in https://github.com/acts-project/acts/pull/2086. It seems like it also appeared somewhere else already and maybe in one of the recent issues.

The problem seems to be that the tolerance is not consistent between `intersection` and `globalToLocal`. Since these are used in sequence with the same tolerance `globalToLocal` will fail as it reports a higher distance than the `intersection`.

A proper fix would be to align the math of the intersection and conversion so the tolerance has the same meaning in both cases.

should also fix https://github.com/acts-project/acts/issues/2226


andiwand:
logged into this issue again and this is the output I get

```
intersect distance 0.940883 position  7.57337e-05 -6.11766e-05      44.2963 momentum -0.949711  0.197314  -0.24314
globalToLocal distance 0.000101456 position  7.57337e-05 -6.11766e-05      44.2963 momentum  0.949711 -0.197314   0.24314
```

asalzburger:
> logged into this issue again and this is the output I get
> 
> ```
> intersect distance 0.940883 position  7.57337e-05 -6.11766e-05      44.2963 momentum -0.949711  0.197314  -0.24314
> globalToLocal distance 0.000101456 position  7.57337e-05 -6.11766e-05      44.2963 momentum  0.949711 -0.197314   0.24314
> ```
> 
> you can see that both functions get the same input but conclude to a different distance.
> 
> @paulgessinger @asalzburger

There is a sign flip in the momentum, this looks like it's the cause.

andiwand:
Nope.

```
intersect distance 0.940883 position  7.57337e-05 -6.11766e-05      44.2963 momentum -0.949711  0.197314  -0.24314
distance 0.000101456 position  7.57337e-05 -6.11766e-05      44.2963 momentum -0.949711  0.197314  -0.24314
```

I don't think the direction should make a difference in this case and it seems to not do so.

If I remember correctly there are a bunch of random sign flips in the navigation which might cause this inconsistency. But I don't think this is relevant now.

@paulgessinger @asalzburger

asalzburger:
> Nope.
> 
> ```
> intersect distance 0.940883 position  7.57337e-05 -6.11766e-05      44.2963 momentum -0.949711  0.197314  -0.24314
> distance 0.000101456 position  7.57337e-05 -6.11766e-05      44.2963 momentum -0.949711  0.197314  -0.24314
> ```
> 
> I don't think the direction should make a difference in this case and it seems to not do so.
> 
> If I remember correctly there are a bunch of random sign flips in the navigation which might cause this inconsistency. But I don't think this is relevant now.
> 
> @paulgessinger @asalzburger

ok, understood - probably time to have a look at the code lines.

andiwand:
linked https://github.com/acts-project/acts/issues/2226 in the description

andiwand:
updated the hashes @paulgessinger

benjaminhuth (approved):

PR #2232, opened 2023-06-21
ci: Physmon for event simulation

andiwand:
All our physics monitors depend on the simulation which is not directly monitored at the moment. In this PR I add a first monitoring version for Fatras and Geant4.

The monitoring is not optimal yet but should be sufficient if something changes or breaks for simulation.

Possible improvements:
- record deltas for momenta / eta / phi and so on instead of initial and final

Blocked by
- https://github.com/acts-project/acts/pull/2241
- https://github.com/acts-project/acts/pull/2260
- https://github.com/acts-project/acts/pull/2262
- https://github.com/acts-project/acts/pull/2263
- https://github.com/acts-project/acts/pull/2264
- https://github.com/acts-project/acts/pull/2265
- https://github.com/acts-project/acts/pull/2266

PR #2241, opened 2023-06-23
fix: Fix FPEOVF in `RootParticleWriter`

andiwand:
I run into a few occasions where `float -> double` caused an FPEOVF. In this PR I change the output precision accordingly.

Discovered in https://github.com/acts-project/acts/pull/2232


paulgessinger (approved):
Ok. The alternative would be to wrap all of them into the `float` range. I don't think we actually have values that meaningfully exceed `float`.

andiwand:
I was not sure why this occurred. Maybe G4 produces a max double in some cases. I will do the float wrapping tho so I do not have to change the reference

andiwand:
@paulgessinger can you re-approve?

paulgessinger (approved):
🚀

PR #2242, opened 2023-06-23
ci: Add Unit Test paths to auto-labeler

paulgessinger:


.github/labeler.yml
```diff
@@ -4,12 +4,16 @@
...
+  - Tests/UnitTests/Core/**
 'Component - Examples':
   - Examples/**
+  - Tests/UnitTests/Examples/**
 'Component - Fatras':
   - Fatras/**
+  - Tests/UnitTests/Fatras/**
 'Component - Plugins':
   - Plugins/**
+  - Tests/UnitTests/Plugins/**
```
CarloVarni, line 16: what about `'**/Plugins/**'` (and similar for others)? The `'` are needed if you start with `**`

paulgessinger: I don't have a strong opinion. I'm leaning towards being more explicit.

PR #2244, opened 2023-06-26
fix: Prototracks to seeds spacepoint livetime

benjaminhuth:
Fixes a problem in the newly introduced PrototracksToSeeds algorithm regarding the lifetime of the spacepoints.

As a side note: I think the interface of `Acts::Seed` is not well designed at the moment, and also not documented. The fact that the class only stores pointers to the objects I pass in as reference, is quite surprising to me. Maybe worth discussing an interface change in the dev meeting.


Examples/Framework/include/ActsExamples/Utilities/EventDataTransforms.hpp
```diff
@@ -15,7 +15,7 @@ namespace ActsExamples {
 
 ProtoTrack seedToPrototrack(const SimSeed &seed);
 
-std::optional<SimSpacePoint> findSpacePointForIndex(
+std::optional<const SimSpacePoint *> findSpacePointForIndex(
```
andiwand, line 18: hmmm how bad would it be to just return a raw pointer and null it if not found? optional of raw pointer seems redundant in some way

andiwand (approved):
as always feel free ignoring

Examples/Framework/include/ActsExamples/Utilities/EventDataTransforms.hpp
benjaminhuth: actually you are right, I'll change it

andiwand (approved):

andiwand (approved):

PR #2250, opened 2023-06-29
refactor: Explicit direction parameter instead of momentum

andiwand:
Currently the `Surface` classes have `momentum` parameters which is misleading IMO as we do not care about the magnitude of that input. Here I try to improve things by renaming the occasions with `direction`.

### DIFF
--- a/Core/include/Acts/Surfaces/ConeSurface.hpp
+++ b/Core/include/Acts/Surfaces/ConeSurface.hpp
@@ -111,12 +111,12 @@ class ConeSurface : public Surface {
   /// @param gctx The current geometry context object, e.g. alignment
   /// @param position is the global position where the measurement frame is
   /// constructed
-  /// @param momentum is the momentum used for the measurement frame
+  /// @param direction is the momentum direction used for the measurement frame
   /// construction
   /// @return matrix that indicates the measurement frame
   RotationMatrix3 referenceFrame(const GeometryContext& gctx,
                                  const Vector3& position,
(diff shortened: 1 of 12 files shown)


paulgessinger:
@asalzburger do you remember if there was any reason the momentum is needed in some case in ATLAS which we haven't implemented in our code? If that's not the case, I'm in favor of changing this like proposed here.

paulgessinger:
This is not actually breaking, right? I thought so first, but the symbol names don't include the argument names, right?

andiwand:
> @asalzburger do you remember if there was any reason the momentum is needed in some case in ATLAS which we haven't implemented in our code? If that's not the case, I'm in favor of changing this like proposed here.

from my understanding the momentum itself would not be useful because without the field we do not know the geometry of the track

> This is not actually breaking, right? I thought so first, but the symbol names don't include the argument names, right?

yes this should be transparent in the ABI

asalzburger:
> > @asalzburger do you remember if there was any reason the momentum is needed in some case in ATLAS which we haven't implemented in our code? If that's not the case, I'm in favor of changing this like proposed here.
> 
> from my understanding the momentum itself would not be useful because without the field we do not know the geometry of the track
> 
> > This is not actually breaking, right? I thought so first, but the symbol names don't include the argument names, right?
> 
> yes this should be transparent in the ABI


I think most of the time we actually use direction, momentum is only used in the Stepping and in Fatras, so these changes make per se sense!

paulgessinger (approved):
Then let's go!

PR #2181, opened 2023-06-06
refactor!: Particle hypothesis for track parameters

andiwand:
In this PR I make the particle hypothesis part of the track parameters. This allows us to consistently use the same hypothesis across various parts of our code. Before that the particle hypothesis was a bit scattered in the propagator and stepper.

- `TrackParameters` gain `ParticleHypothesis`
- `ParticleHypothesis` is a new class which holds a charge type, the mass and the absolute PDG value
  - This interface is consistently used to get the charge or momenta from `qOverP` or vice versa
- `Single*TrackParameters` are renamed and typedefed in `TrackParameters.hpp`
  - The templated `TrackParameters` implementations are now called `Generic*TrackParameters`
- I aligned `absoluteMomentum`, `direction`, `momentum` in the steppers, truth particles and track parameters

alternative to https://github.com/acts-project/acts/pull/2178

PR #2253, opened 2023-06-30
refactor: `qOverP` and `absoluteMomentum` for Stepper interface

andiwand:
With this PR I try to align the naming in the Stepper interface to `qOverP` and `absoluteMomentum` which is also my goal for the track parameters.

I pulled these changes out of https://github.com/acts-project/acts/pull/2181.


paulgessinger (approved):
Seems fine from my side.

paulgessinger:
I guess this is also not really breaking if you don't use the internals of the stepper?

andiwand:
> I guess this is also not really breaking if you don't use the Stepper directly?

Yeah depends on what we view as a public interface. The stepper interface is breaking but it is absorbed by the Propagator for example.

PR #2255, opened 2023-07-03
chore: Improve CMake formatting

andiwand:
minor formatting change in our central `CMakeLists.txt`


paulgessinger (approved):
Sure, let's go.

PR #2256, opened 2023-07-03
docs: Ignore if `eigen.tuxfamily.org` is online

andiwand:
Ignore `eigen.tuxfamily.org` which is down atm

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -55,6 +55,7 @@ linkcheck_ignore = [
     r"https://doi.org/.*",
     r"https://cernvm.cern.ch/.*",
     r"https://tavianator.com/.*",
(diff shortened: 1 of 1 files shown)


paulgessinger (approved):

docs/conf.py
```diff
@@ -55,6 +55,7 @@
     r"https://doi.org/.*",
     r"https://cernvm.cern.ch/.*",
     r"https://tavianator.com/.*",
+    r"https://eigen.tuxfamily.org/.*",
```
andiwand, line 58:
```suggestion
    r"http://eigen.tuxfamily.org/.*",
```

paulgessinger (approved):

PR #2257, opened 2023-07-03
ci: Add Folder with hashes to the labeler paths

CarloVarni:


andiwand (approved):

.github/labeler.yml
```diff
@@ -18,6 +18,7 @@
   - docs/**
 'Changes Performance':
   - CI/physmon/reference/**
+  - Examples/Python/tests/**
```
paulgessinger, line 21: Wouldn't it be better to check if the hash text file changes specifically? A code change here does not necessarily mean any of the outputs change.

AJPfleger:
I think there might be a typo in the title: `ashes` -> `hashes`

.github/labeler.yml
CarloVarni: right. Targeting now only the `root_file_hashes.txt` file

andiwand (approved):

PR #2260, opened 2023-07-04
fix: Remove potential FPE in `RootTrajectorySummaryWriter`

andiwand:
`NAN` default values are evil


Examples/Io/Root/src/RootTrajectorySummaryWriter.cpp
```diff
@@ -345,22 +347,21 @@ ActsExamples::ProcessCode ActsExamples::RootTrajectorySummaryWriter::writeT(
...
+              param[Acts::eBoundQOverP] - t_qop,
+              param[Acts::eBoundTime] - t_time};
+
+          if (boundParam.covariance().has_value()) {
+            const auto& covariance = *boundParam.covariance();
+            for (unsigned int i = 0; i < Acts::eBoundSize; ++i) {
+              error[i] = std::sqrt(covariance(i, i));
               pull[i] = res[i] / error[i];
-            } else {
-              pull[i] = std::numeric_limits<double>::quiet_NaN();
```
paulgessinger, line 363: But then the vector sizes are not the same anymore, are they?

andiwand: I think it should be. my goal is to write `NAN` if we don't know the reconstructed params or true params but maybe I messed this up

paulgessinger: Ok i might have to reread the diff more closely.

andiwand:
I do not expect any hash changes. Lets see if this is fixed now

paulgessinger (approved):

acts-project-service:
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/30813174) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/5845949)!

PR #2266, opened 2023-07-04
refactor: Drop bool conversion in favor of `isAlive` method in `Fatras::Particle`

andiwand:
Makes the state check for the particle more explicit


paulgessinger (approved):
🚀

PR #2273, opened 2023-07-05
refactor: add time to IVF

felix-russo:
Until now only small changes, made PR to keep track of CI


felix-russo:
will work on this later

PR #2269, opened 2023-07-04
refactor: Rename `Single*TrackParameters` to `Generic*TrackParameters`

andiwand:
I am unhappy with the name `Single*TrackParameters` as it could refer to singly charged or single track. The track actually supports different kinds of charge hypothesis so I thought `Generic*TrackParameter` might fit better.

I also merged `NeutralTrackParameters` into `TrackParameters` and removed the external template instantiation as it could be a performance hit because it disables inlining.

Pulled these changes out of https://github.com/acts-project/acts/pull/2181

PR #2287, opened 2023-07-08
fix: Align `globalToLocal` and `intersection` in `LineSurface`

andiwand:
This is supposed to be a proper fix after the quick fix in https://github.com/acts-project/acts/pull/2239

In `globalToLocal` we did not respect the direction which is important for the line surface.

I tried to make the code a little more clear and move away from constructor initialization in the mathy part. I also added a unit test to check the intersection against the propagation.

I will let the CI judge how much this breaks


Core/src/Surfaces/LineSurface.cpp
```diff
@@ -80,26 +79,21 @@ Acts::Result<Acts::Vector2> Acts::LineSurface::globalToLocal(
...
-  if ((localToGlobal(gctx, lposition, direction) - position).norm() >
-      tolerance * 1.5) {
+  Vector3 lposition3 = totalTransform * position;
+
+  if (std::abs(lposition3.z()) > std::abs(tolerance)) {
     return Result<Vector2>::failure(SurfaceError::GlobalPositionNotOnSurface);
   }
 
+  // Construct local position from local x,y
+  Vector2 lposition = lposition3.head<2>();
```
felix-russo, line 95:
```suggestion
  Vector2 localXY = localPosition.head<2>();
```

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -80,26 +79,21 @@ Acts::Result<Acts::Vector2> Acts::LineSurface::globalToLocal(
     const GeometryContext& gctx, const Vector3& position,
     const Vector3& direction, double tolerance) const {
   using VectorHelpers::perp;
-  const auto& sTransform = transform(gctx);
-  const auto& tMatrix = sTransform.matrix();
-  Vector3 lineDirection(tMatrix(0, 2), tMatrix(1, 2), tMatrix(2, 2));
+
+  Transform3 totalTransform =
+      referenceFrame(gctx, position, direction).inverse() *
+      transform(gctx).inverse();
```
felix-russo, line 85: Also here I am not sure if I understand the difference between the old and the new code

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -151,45 +144,51 @@ Acts::SurfaceIntersection Acts::LineSurface::intersect(
     const GeometryContext& gctx, const Vector3& position,
     const Vector3& direction, const BoundaryCheck& bcheck,
     ActsScalar tolerance) const {
-  // following nominclature found in header file and doxygen documentation
-  // line one is the straight track
+  // The nominclature is following the header file and doxygen documentation
```
felix-russo, line 147:
```suggestion
  // The nomenclature is following the header file and doxygen documentation
```

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -110,16 +104,15 @@ std::string Acts::LineSurface::name() const {
 Acts::RotationMatrix3 Acts::LineSurface::referenceFrame(
     const GeometryContext& gctx, const Vector3& /*position*/,
     const Vector3& direction) const {
+  Vector3 measZ0 = transform(gctx).rotation() * Vector3::UnitZ();
+  Vector3 measD0 = measZ0.cross(direction).normalized();
+  Vector3 measDepth = measD0.cross(measZ0);
```
felix-russo, line 109:
```suggestion
  Vector3 measDepth = direction.normalized();
  Vector3 meadD0 = measZ0.cross(measDepth);
```

felix-russo: The variable names are not obvious for me, but this might be due to my ignorance

Tests/UnitTests/Core/Surfaces/LineSurfaceTests.cpp
```diff
@@ -104,7 +111,8 @@ BOOST_AUTO_TEST_CASE(LineSurface_allNamedMethods_test) {
   // globalToLocal()
   Vector3 gpos{0., 1., 0.};
   const Vector3 mom{20., 0., 0.};  // needs more realistic parameters
-  Vector2 localPosition = line.globalToLocal(tgContext, gpos, mom).value();
+  Vector2 localPosition =
+      line.globalToLocal(tgContext, gpos, mom.normalized()).value();
```
felix-russo, line 115: I guess the momentum normalization should happen in `globalToLocal` to avoid the use of the function with unnormalized momenta

Core/src/Surfaces/LineSurface.cpp
andiwand: I also don't like them but I had no better idea. maybe `unitD0` `unitZ0` `unitNormal` is better?

I wanted to make the same change but `direction` is actually not pointing in the same way as `measDepth` in the general case.

Tests/UnitTests/Core/Surfaces/LineSurfaceTests.cpp
andiwand: not sure. I recently changed this so the surface base class explicitly says that this is a unit direction. my intention was to make the interface clear:
- we do not depend on the magnitude
- we need to have a direction
- we will no recalculate the length of this vector

andiwand (commented):
thanks for the review @felix-russo  !

Core/src/Surfaces/LineSurface.cpp
felix-russo: Maybe something like `dirD0` etc.?

Hm this I don't quite understand... Does that mean that `direction` and `measZ0` are not orthogonal to each other in the general case?

Tests/UnitTests/Core/Surfaces/LineSurfaceTests.cpp
felix-russo: Ah, ok, now I see that the normalization does not matter. Why do you normalize in the unit test then?

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -80,27 +79,20 @@ Acts::Result<Acts::Vector2> Acts::LineSurface::globalToLocal(
...
-
-  // TODO make intersection and local<->global tolerances sound. quick fixed
-  // this with a 50% tolerance increase for now
-  if ((localToGlobal(gctx, lposition, direction) - position).norm() >
-      tolerance * 1.5) {
+
+  // Bring the global position into the local frame. First remove the
+  // translation then the rotation.
+  Vector3 localPosition = referenceFrame(gctx, position, direction).inverse() *
+                          (position - transform(gctx).translation());
```
felix-russo, line 86: Ok this is very clear now, thanks! Did you check how the output of the function changes with the new code?

Core/include/Acts/Surfaces/LineSurface.hpp
```diff
@@ -90,12 +90,12 @@ class LineSurface : public Surface {
...
-  /// Normal vector return
+  /// The normal vector is undefined if we do not know the momentum.
   ///
   /// @param gctx The current geometry context object, e.g. alignment
   /// @param lposition is the local position is ignored
   ///
-  /// @return a Vector3 by value
+  /// @return a zero vector
   Vector3 normal(const GeometryContext& gctx,
                  const Vector2& lposition) const final;
```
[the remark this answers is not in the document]
andiwand: it has to be provided because the base class requires it. but there is no surface normal if you don't have the momentum so an invalid output makes most sense to me

Core/src/Surfaces/LineSurface.cpp
andiwand: this is validated through physmon / unit tests I'd say. the point is that the output changes tho

Tests/UnitTests/Core/Surfaces/LineSurfaceTests.cpp
andiwand: sometimes it does not matter, sometimes it does. the contract of the base class method requires a normal vector so I would always provide a normal vector

asalzburger (approved):

asalzburger:
@andiwand - are both changing output, this one and the tolerance PR?

andiwand:
> @andiwand - are both changing output, this one and the tolerance PR?

this one changes the output of the AMVF slightly. I cannot really say why but the output does not look much better or worse so I just updated the reference

https://github.com/acts-project/acts/pull/2292 is changing the hashes for the GSF but the physmon is happy

paulgessinger:
This likely changes Athena output in the vertexing. Should we hold this for v28?

paulgessinger:
We're ready to go for `v28` now. Can you update the PR so we can merge it?

andiwand:
> We're ready to go for `v28` now. Can you update the PR so we can merge it?

should be ready now @paulgessinger

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -134,10 +125,9 @@ Acts::Vector3 Acts::LineSurface::binningPosition(
   return center(gctx);
 }
 
-Acts::Vector3 Acts::LineSurface::normal(const GeometryContext& gctx,
+Acts::Vector3 Acts::LineSurface::normal(const GeometryContext& /*gctx*/,
                                         const Vector2& /*lpos*/) const {
-  const auto& tMatrix = transform(gctx).matrix();
-  return Vector3(tMatrix(0, 2), tMatrix(1, 2), tMatrix(2, 2));
+  return Vector3::Zero();
 }
```
paulgessinger, line 131: Do we ever call this? If not maybe this should be an error?

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -110,16 +102,15 @@ std::string Acts::LineSurface::name() const {
 Acts::RotationMatrix3 Acts::LineSurface::referenceFrame(
```
paulgessinger, line 102: I think we need to document more thoroughly what this ends up calculating.

- y -> line direction
- x -> perp line direction and particle direction
- z -> perp x&y

- None of the axes is parallel to the particle direction in the end
- z points in the direction of the PCA (due to perpendicularities).

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -151,45 +141,51 @@ Acts::SurfaceIntersection Acts::LineSurface::intersect(
...
-  const auto& tMatrix = transform(gctx).matrix();
-  Vector3 mb = tMatrix.block<3, 1>(0, 3).transpose();
-  Vector3 eb = tMatrix.block<3, 1>(0, 2).transpose();
-  // now go ahead and solve for the closest approach
-  Vector3 mab(mb - ma);
+
+  // Origin of the line surface
+  Vector3 mb = transform(gctx).translation();
+  // Line surface axis
+  Vector3 eb = transform(gctx).rotation() * Vector3::UnitZ();
```
paulgessinger, line 152: The extraction + transpose is probably cheaper than the full matrix multiplication, right?

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -151,45 +141,51 @@ Acts::SurfaceIntersection Acts::LineSurface::intersect(
...
+  if (std::abs(denom) < std::abs(tolerance)) {
+    // return a false intersection
+    return {Intersection3D(position, std::numeric_limits<double>::max(),
+                           Intersection3D::Status::unreachable),
+            this};
+  }
+
+  double u = (mab.dot(ea) - mab.dot(eb) * eaTeb) / denom;
+  // Check if we are on the surface already
+  Intersection3D::Status status = std::abs(u) < std::abs(tolerance)
```
paulgessinger, line 170: So the meaning of `u` is the distance along the particle direction to the PCA, so if it's 0 we're already there?

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -151,45 +141,51 @@ Acts::SurfaceIntersection Acts::LineSurface::intersect(
...
+                                      ? Intersection3D::Status::onSurface
+                                      : Intersection3D::Status::reachable;
+  Vector3 result = ma + u * ea;
```
paulgessinger, line 173: This is the vector from where we are to the PCA position in the global frame?

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -198,37 +194,39 @@ Acts::BoundToFreeMatrix Acts::LineSurface::boundToFreeJacobian(
...
-  const double sin_theta = std::sin(boundParams[eBoundTheta]);
-  const double cos_phi = std::cos(boundParams[eBoundPhi]);
-  const double sin_phi = std::sin(boundParams[eBoundPhi]);
+  double cosTheta = std::cos(boundParams[eBoundTheta]);
+  double sinTheta = std::sin(boundParams[eBoundTheta]);
+  double cosPhi = std::cos(boundParams[eBoundPhi]);
+  double sinPhi = std::sin(boundParams[eBoundPhi]);
   // retrieve the reference frame
-  const auto rframe = referenceFrame(gctx, position, direction);
+  auto rframe = referenceFrame(gctx, position, direction);
```
paulgessinger, line 206: The `const`s don't hurt, do they?

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -239,81 +237,79 @@ Acts::BoundToFreeMatrix Acts::LineSurface::boundToFreeJacobian(
...
-  const auto& localZAxis = rotation.col(2);
+  Vector3 localZAxis = transform(gctx).rotation() * Vector3::UnitZ();
   // The local z coordinate
-  const auto pz = pcRowVec * localZAxis;
+  double pz = pcRowVec.dot(localZAxis);
   // Cosine of angle between momentum direction and local frame z axis
-  const auto dz = localZAxis.dot(direction);
-  const auto norm = 1 / (1 - dz * dz);
+  double dz = localZAxis.dot(direction);
+  double norm = 1 / (1 - dz * dz);
```
paulgessinger, line 258: Same here, the `const`s don't make a difference, but communicate that these values don't change anymore.

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -239,81 +237,79 @@ Acts::BoundToFreeMatrix Acts::LineSurface::boundToFreeJacobian(
...
   FreeToPathMatrix freeToPath = FreeToPathMatrix::Zero();
+
   // The derivative of path w.r.t. position
   freeToPath.segment<3>(eFreePos0) =
       norm * (dz * localZAxis.transpose() - direction.transpose());
+
   // The derivative of path w.r.t. direction
   freeToPath.segment<3>(eFreeDir0) =
-      norm * (pz * localZAxis.transpose() - pcRowVec);
+      norm * (pz * localZAxis.transpose() - pcRowVec.transpose());
```
paulgessinger, line 269: How was this 'valid' before **and** after?

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -80,27 +79,20 @@ Acts::Result<Acts::Vector2> Acts::LineSurface::globalToLocal(
...
-  // this with a 50% tolerance increase for now
-  if ((localToGlobal(gctx, lposition, direction) - position).norm() >
-      tolerance * 1.5) {
+
+  // Bring the global position into the local frame. First remove the
+  // translation then the rotation.
+  Vector3 localPosition = referenceFrame(gctx, position, direction).inverse() *
+                          (position - transform(gctx).translation());
+
+  if (std::abs(localPosition.z()) > std::abs(tolerance)) {
```
paulgessinger, line 88: This checks if the current position is on the line (and not anything about the PCA)? Is this what we want?
The reason this goes through the trouble of calculating the reference frame etc is just that we want to rotate the residual `xy` coordinates according to the particle direction right, and this automatically gives the right sign, is that correctly?

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -80,27 +79,20 @@ Acts::Result<Acts::Vector2> Acts::LineSurface::globalToLocal(
...
-  const auto& tMatrix = sTransform.matrix();
-  Vector3 lineDirection(tMatrix(0, 2), tMatrix(1, 2), tMatrix(2, 2));
-  // Bring the global position into the local frame
-  Vector3 loc3Dframe = sTransform.inverse() * position;
-  // construct localPosition with sign*perp(candidate) and z.()
-  Vector2 lposition(perp(loc3Dframe), loc3Dframe.z());
-  Vector3 sCenter(tMatrix(0, 3), tMatrix(1, 3), tMatrix(2, 3));
-  Vector3 decVec(position - sCenter);
-  // assign the right sign
-  double sign = ((lineDirection.cross(direction)).dot(decVec) < 0.) ? -1. : 1.;
```
paulgessinger, line 93: The PR description reads 

> In `globalToLocal` we did not respect the direction which is important for the line surface.

But it did take the direction into account here no? So was there a bug here and the new implementation doesn't have it?

Tests/UnitTests/Core/Surfaces/LineSurfaceTests.cpp
paulgessinger, line 1: I'd suggest not using `auto` where you type the concrete type out anyway, as it just turns into noise in that case, imo.

Tests/UnitTests/Core/Surfaces/LineSurfaceTests.cpp
```diff
@@ -266,6 +275,59 @@ BOOST_AUTO_TEST_CASE(LineSurfaceTransformRoundTripEtaStability) {
...
+  auto direction = Vector3(1, 1, 100).normalized();
+  auto boundVector =
+      (BoundVector() << 1_cm, 1_cm, VectorHelpers::phi(direction),
+       VectorHelpers::theta(direction), 1, 0)
+          .finished();
+  auto pathLimit = 1_cm;
+
+  auto surface = std::make_shared<LineSurfaceStub>(Transform3::Identity());
+
+  auto initialParams = BoundTrackParameters(surface, boundVector, 1);
```
paulgessinger, line 292: Here for example:
```suggestion
  BoundTrackParameters initialParams{surface, boundVector, 1};
```

Core/src/Surfaces/LineSurface.cpp
```diff
@@ -151,45 +141,51 @@ Acts::SurfaceIntersection Acts::LineSurface::intersect(
...
-      const Vector3 vecLocal(result - mb);
-      double cZ = vecLocal.dot(eb);
-      double hZ = m_bounds->get(LineBounds::eHalfLengthZ) + tolerance;
-      if ((std::abs(cZ) > std::abs(hZ)) or
-          ((vecLocal - cZ * eb).norm() >
-           m_bounds->get(LineBounds::eR) + tolerance)) {
-        status = Intersection3D::Status::missed;
-      }
+
+  // `tolerance` does not really have a meaning here it is just a sufficiently
```
paulgessinger, line 159: Is that true? Aren't we using `tolerance` later on to check if we're at the PCA?

paulgessinger (commented):
Sorry maybe I'm dumb but from the diff it's not at all clear to me what's actually happening here. You're changing both `intersect` and `globalToLocal`, and they're not doing the same thing afterwards (and neither did they before I don't think).

What do we want `globalToLocal` to do (after this)? Check if we're on the line surface, or check if we're on the PCA?

Tests/UnitTests/Core/Surfaces/LineSurfaceTests.cpp
```diff
@@ -266,6 +275,59 @@ BOOST_AUTO_TEST_CASE(LineSurfaceTransformRoundTripEtaStability) {
...
+BOOST_AUTO_TEST_CASE(LineSurfaceIntersection) {
+  using namespace Acts::UnitLiterals;
+
+  auto eps = 1e-10;
+
+  auto direction = Vector3(1, 1, 100).normalized();
+  auto boundVector =
+      (BoundVector() << 1_cm, 1_cm, VectorHelpers::phi(direction),
+       VectorHelpers::theta(direction), 1, 0)
+          .finished();
```
paulgessinger, line 287: 🤯
```suggestion
  BoundVector boundVector;
  boundVector << 1_cm, 1_cm, VectorHelpers::phi(direction), VectorHelpers::theta(direction), 1, 0;
```
?

Tests/UnitTests/Core/Surfaces/LineSurfaceTests.cpp
```diff
@@ -266,6 +275,59 @@ BOOST_AUTO_TEST_CASE(LineSurfaceTransformRoundTripEtaStability) {
...
+    BOOST_CHECK(result.ok());
+    BOOST_CHECK(result.value().endParameters);
+
+    displacedParameters = result.value().endParameters.value();
+  }
+
+  auto intersection =
+      surface->intersect(tgContext, displacedParameters.position(tgContext),
+                         displacedParameters.unitDirection());
+  CHECK_CLOSE_ABS(intersection.intersection.pathLength, pathLimit, eps);
```
paulgessinger, line 313: So this takes parameters at `0,0,0`, straight line steps backwards `1cm`, then checks if it's intersecting with the surface at `1cm`? If so what's the point of the propagation? Can't you just literally move the track parameters position backwards by `1cm`?

Tests/UnitTests/Core/Surfaces/LineSurfaceTests.cpp
```diff
@@ -266,6 +275,59 @@ BOOST_AUTO_TEST_CASE(LineSurfaceTransformRoundTripEtaStability) {
...
+    options.maxStepSize = 1_mm;
+
+    auto result = propagator.propagate(displacedParameters, *surface, options);
+    BOOST_CHECK(result.ok());
+    BOOST_CHECK(result.value().endParameters);
+    CHECK_CLOSE_ABS(result.value().pathLength, pathLimit, eps);
+    endParameters = result.value().endParameters.value();
+  }
+
+  CHECK_CLOSE_ABS(initialParams.parameters(), endParameters.parameters(), eps);
```
paulgessinger, line 328: This I think makes more sense to check, i.e. that the propagation correctly uses the new intersect method to approach the surface in `1mm` steps.

Core/src/Surfaces/LineSurface.cpp
andiwand: yes this is the global position of the intersection

Core/src/Surfaces/LineSurface.cpp
andiwand: does not hurt but it is inconsistent IMO. I would either use const on all temporaries or don't const any of them

Core/src/Surfaces/LineSurface.cpp
andiwand: consistency

Core/src/Surfaces/LineSurface.cpp
andiwand: I think eigen is treating vectors a little bit differently. so some operations are fine for `3x1` and `1x3` mixed

Tests/UnitTests/Core/Surfaces/LineSurfaceTests.cpp
andiwand: oh yeah I went a bit hard on auto here. will fix that

Tests/UnitTests/Core/Surfaces/LineSurfaceTests.cpp
andiwand: it takes the bound parameters we used before and steps away. we could do the math explicitly but then I have to do manual `localToGlobal` and add a displacement vector which seemed more work at the time. this way we also check the consistency of the straight line intersection and the straight line propagation

Core/src/Surfaces/LineSurface.cpp
andiwand: I think it boils down to what is `d0` and `z0` and in which order do we store them. I always have to guess but I think it is `d0,z0` which corresponds to `x,y` and then z is the "surface normal" so just the cross product of `x` and `y` in the global frame.

Tests/UnitTests/Core/Surfaces/LineSurfaceTests.cpp
paulgessinger: ok

Core/src/Surfaces/LineSurface.cpp
paulgessinger: Right but presumably `tolerance` should be small.

Core/src/Surfaces/LineSurface.cpp
paulgessinger: I don't know this, but it's possible, yes.

andiwand:
after https://github.com/acts-project/acts/pull/2321 the conflicts should disappear :crossed_fingers:

Core/src/Surfaces/LineSurface.cpp
paulgessinger: Ok I'll follow this up by checking where we actually use this overload.

paulgessinger (approved):

acts-project-service:
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/31358306) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/5964458)!

paulgessinger:
Build failure is likely due to #2269, and not this PR. I'll update this PR once it's resolved.

PR #2291, opened 2023-07-10
WIP: feat: misalignment

LaraCalic:
This is a pull request for the implementation of the misalignment of individual sensors in a correlated way.  The status of the code is the following:  2 files fail to build (one of them is GeometryHierarchyMap.hpp, and another one that can be 'accessible'  without building the previous one).


Alignment/include/ActsAlignment/Kernel/AlignmentError.hpp
```diff
@@ -51,10 +52,133 @@ extern inline const detail::AlignmentErrorCategory& AlignmentErrorCategory() {
 inline std::error_code make_error_code(ActsAlignment::AlignmentError e) {
   return {static_cast<int>(e), ActsAlignment::AlignmentErrorCategory()};
 }
-}  // namespace ActsAlignment
 
-namespace std {
-// register with STL
-template <>
-struct is_error_code_enum<ActsAlignment::AlignmentError> : std::true_type {};
-}  // namespace std
```
paulgessinger, line 60: Does this go away?

Examples/Algorithms/Alignment/include/ActsExamples/Alignment/AlignmentAlgorithm.hpp
```diff
@@ -29,87 +30,52 @@ class AlignmentAlgorithm final : public IAlgorithm {
...
-      this, "InputInitialTrackParameters"};
-  ReadDataHandle<ProtoTrackContainer> m_inputProtoTracks{this,
-                                                         "InputProtoTracks"};
-  WriteDataHandle<AlignmentParameters> m_outputAlignmentParameters{
-      this, "OutputAlignmentParameters"};
+  ReadDataHandle<MeasurementContainer> m_inputMeasurements;
+  ReadDataHandle<IndexSourceLinkContainer> m_inputSourceLinks;
+  ReadDataHandle<TrackParametersContainer> m_inputInitialTrackParameters;
+  ReadDataHandle<ProtoTrackContainer> m_inputProtoTracks;
+  WriteDataHandle<AlignmentParameters> m_outputAlignmentParameters;
```
paulgessinger, line 78: I'm not sure I understand these changes.

Core/include/Acts/Geometry/GeometryIdentifier.hpp
```diff
@@ -33,9 +32,11 @@ class GeometryIdentifier {
   using Value = uint64_t;
 
   /// Construct from an already encoded value.
-  constexpr GeometryIdentifier(Value encoded) : m_value(encoded) {}
+  constexpr GeometryIdentifier(Value encoded)
+      : m_value(encoded), m_misalignmentX(0.0), m_misalignmentY(0.0) {}
```
paulgessinger, line 36: `GeometryIdentifier` will need to contain it's internal structure as a 64bit integer, so we'll have to find another place to put this information.

paulgessinger (commented):
Thanks for the draft @LaraCalic. I have a few general comments:

- I don't think the misalignments should be stored on the `TrackingGeometry` object. The architecture we have in place for this is the `GeometryContext`, which contains (optionally experiment specific) information on the alignment. In the context of the examples framework, we have the `AlignedDetector`, which either uses the context object to look up the misalignments in an external store, or contain the misalignments themselves. This development should use that mechanism, I believe.
- Usually, the misalignment is generalized in the form of a delta transform that is to be applied on top of the default transform. That way, it can encompass both rotations as well as shifts.
- We need to find a way to decouple the structure that the misalignments are calculated from the geometry classes themselves, since the geometry itself needs to remain agnostic of misalignment.
- There seem to be some unrelated changes in the changeset.

PR #2303, opened 2023-07-13
ci: personal physmon machine

felix-russo:
This PR is only here to compute the physmon for the grid vertex seeder. **It will not be merged.**

For this purpose I will: 
-) Generate physmon data from the grid vertex seeder 
-) Use this data as new reference for the physmon
-) See how my changes to the grid seeder affect the physmon

Commit hash and what the physmon shows:
-) 6a350e8: amvf + grid seeder (monitored) vs. amvf + gaussian seeder (reference)
-) d55f33d: amvf + grid seeder with changes from #2338 (monitored)  vs. amvf + grid seeder (reference)


felix-russo:
Not needed anymore after #2355

PR #2306, opened 2023-07-17
refactor: avoid un-necessary cleanings of vectors

CarloVarni:
Move the vectors for sorting top and bottom candidates into the Seeding State. These are `vectors<std::size_t>` (corresponding to `std::iota`) that contains indexes corresponding to doublet candidates (stored in different vectors). 

Currently they are created from scratch every time the `filterCandidates` method is called (i.e. once per middle space point), but here we are doing it only once.

Also, with this PR we avoid not necessary cleanings of `std::vector<float>` objets, i.e. `state.curvatures` and `state.impactParameters`. We simply replace the previous values with correct values and keep track of the proper valid range.


CarloVarni:
@LuisFelipeCoelho can you check the effect of this? it should be quite minimal in truth ...

paulgessinger:
Is the resizing really making a difference? I think the `vector`s should internally do exactly what you're doing in custom code now.

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -436,50 +436,61 @@ inline void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
...
-  float rM = spM.radius();
-  float cosPhiM = spM.x() / rM;
-  float sinPhiM = spM.y() / rM;
-  float varianceRM = spM.varianceR();
-  float varianceZM = spM.varianceZ();
+  const float& rM = spM.radius();
+  const float cosPhiM = spM.x() / rM;
+  const float sinPhiM = spM.y() / rM;
+  const float& varianceRM = spM.varianceR();
+  const float& varianceZM = spM.varianceZ();
```
paulgessinger, line 443: The references will require more storage (64bit) than the floats (32bit), so I think the copies are actually cheaper.

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -509,10 +520,12 @@ inline void SeedFinder<external_spacepoint_t, platform_t>::filterCandidates(
     float sinTheta = 1 / std::sqrt(iSinTheta2);
     float cosTheta = cotThetaB * sinTheta;
 
-    // clear all vectors used in each inner for loop
-    state.topSpVec.clear();
```
paulgessinger, line 513: `clear` does not in fact reduce the capacity, so this ~ only resets the last-element index to 0.

Core/include/Acts/Seeding/SeedFinder.ipp
CarloVarni: I believe references and floats are both 4 bytes, no? same for unsigned long int

CarloVarni: ok... it's 4 bytes on 32-bit, 8 bytes on 64-bit

PR #2308, opened 2023-07-18
feat: cubic Detector infrastructure

asalzburger:
This PR adds the second, rather trivial but often used Detector geometry setup, the building of telescope like detector.

BoxGeometryRotated00

Detector connection is allowed in (x,y,z,) and for arbitrarily rotated volumes.

When concentrating common code between the `CylindricalDetectorHelper` and the `CubicDetectorHelper`, a dependency problem for the `PortalHelper` and its associated Json test appeared (it would not allow to to define a dummy `Test::DetectorVolume`, hence some of the deep dependencies have been simplified by move parts of the code into implementation files.


asalzburger:
Conflicts to be resolved.

paulgessinger:
What's the plan for this @asalzburger?

PR #2312, opened 2023-07-24
refactor: Remove direction from stepper

andiwand:
The propagation direction is currently stored in multiple places which makes it tedious to change it consistently during propagation which can easily result in bugs.

In this PR I propose to remove the `navDir` from the stepper interface and replace it consistently with the propagation option for the direction.

This should not effect any physics performance as it is a code refactor of how we pass down the propagation direction to the stepper.

In a follow up PR I want to do the same thing for the navigation.


benjaminhuth (approved):
Looks good from my side!

paulgessinger:
So effectively this merges the navigation direction with propagation direction?

Core/include/Acts/Propagator/detail/SteppingHelper.hpp
```diff
@@ -33,14 +33,14 @@ namespace detail {
 template <typename stepper_t>
 Acts::Intersection3D::Status updateSingleSurfaceStatus(
     const stepper_t& stepper, typename stepper_t::State& state,
-    const Surface& surface, const BoundaryCheck& bcheck, const Logger& logger,
-    ActsScalar surfaceTolerance) {
+    const Surface& surface, Direction navDir, const BoundaryCheck& bcheck,
+    ActsScalar surfaceTolerance, const Logger& logger) {
```
paulgessinger, line 37: Any reason for changing the argument order here?

andiwand:
> So effectively this merges the navigation direction with propagation direction?

yes exactly - that way there is only a single place and it cannot go out of sync

Core/include/Acts/Propagator/AtlasStepper.hpp
```diff
@@ -311,18 +305,16 @@ class AtlasStepper {
   /// @param [in] boundParams Parameters in bound parametrisation
   /// @param [in] cov Covariance matrix
   /// @param [in] surface Reset state will be on this surface
-  /// @param [in] navDir Navigation direction
   /// @param [in] stepSize Step size
   void resetState(
       State& state, const BoundVector& boundParams, const BoundSymMatrix& cov,
-      const Surface& surface, const Direction navDir = Direction::Forward,
+      const Surface& surface,
```
paulgessinger, line 311: We do have cases where the direction was changed from inside the propagation, see below.
Should this not reset the options in this case?

Core/include/Acts/Propagator/EigenStepper.hpp
```diff
@@ -248,15 +240,17 @@ class EigenStepper {
...
   /// @param [in] surfaceTolerance Surface tolerance used for intersection
+  /// @param [in] logger A @c Logger instance
   Intersection3D::Status updateSurfaceStatus(
-      State& state, const Surface& surface, const BoundaryCheck& bcheck,
-      const Logger& logger = getDummyLogger(),
-      ActsScalar surfaceTolerance = s_onSurfaceTolerance) const {
+      State& state, const Surface& surface, Direction navDir,
+      const BoundaryCheck& bcheck,
+      ActsScalar surfaceTolerance = s_onSurfaceTolerance,
+      const Logger& logger = getDummyLogger()) const {
```
paulgessinger, line 251: Ordering change is unrelated?

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -1240,7 +1240,7 @@ class CombinatorialKalmanFilter {
         ACTS_VERBOSE(
             "Reverse navigation direction after smoothing for reaching the "
             "target surface");
-        state.stepping.navDir = state.stepping.navDir.invert();
+        state.options.direction = state.options.direction.invert();
```
paulgessinger, line 1243: Oh ok, so we do reverse indeed. Seems a bit odd that the input options are not immutable. Otoh they do get copied in, so the state does de-facto own them.

paulgessinger (commented):
The argument ordering `logger` vs `surfaceTolerance` seems unrelated / arbitrary, but not critical.

paulgessinger:
Probably debatable: I guess technically this is breaking, but I think since neither `Propagator` nor `PropagatorOptions` changes at all, and the outputs are identical, I'd probably not declare this as breaking.

Core/include/Acts/Propagator/detail/SteppingHelper.hpp
andiwand: right - I thought the logger is usually the last argument and since I broke the interface I could just apply a reorder as well. but its just a slight preference. I can revert that

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
andiwand: it also bugs me a bit that it changes the "options" now but on the other hand this is how the propagation currently works. as you point out it is a copy so ownership should be fine

Core/include/Acts/Propagator/AtlasStepper.hpp
andiwand: since the stepper state property is gone now I think it is not the steppers responsibility anymore to change the direction

paulgessinger: Right, it's not at all inside the stepper state anymore. So the e.g. CKF has to do that itself now. Fair enough.

Core/include/Acts/Propagator/detail/SteppingHelper.hpp
paulgessinger: Up to you

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -583,15 +583,15 @@ class CombinatorialKalmanFilter {
       // Update the stepping state
       stepper.resetState(state.stepping, currentState.filtered(),
```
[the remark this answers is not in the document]
andiwand: I think `reset` does not flip the direction in this case but will prepare jumping from one tip to the other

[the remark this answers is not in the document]
paulgessinger: But didn't it hard reset it to `forward` before? And now it doesn't? That's what I'm wondering.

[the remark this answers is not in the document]
andiwand: before we called reset with `state.stepping.navDir` which basically means take the same direction as before. now we just don't pass that param anymore and nobody modifies the option so it should result in the same behavior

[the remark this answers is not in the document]
paulgessinger: Fair enough!

acts-project-service:
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/31541508) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6003740)!

paulgessinger:
The build did in fact work, I believe.

PR #2316, opened 2023-07-24
refactor: Explicit and type safe surface candidates delegates for new detector navigation

andiwand:
WIPWIP

Broke this changes out of https://github.com/acts-project/acts/pull/2214


andiwand:
I will pick this up when I start working on the new geometry / navigation again

PR #2319, opened 2023-07-26
feat: Target surface for filtering phase of CKF

andiwand:
For inward track finding using the `CombinatorialKalmanFilter` we need an abort condition to stop propagation at the interaction point as we would otherwise go through the whole detector on the other side.

In this PR I introduce a `filterTargetSurface` which terminates the track finding when reached. It is optional and therefore should not modify the current behavior of the CKF.

I also aligned the naming for the existing `targetSurface` to `smoothingTargetSurface` to disambiguate those two surfaces.


Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -193,8 +193,11 @@ struct CombinatorialKalmanFilterOptions {
   /// The trivial propagator options
   PropagatorPlainOptions propagatorPlainOptions;
 
-  /// The reference Surface
-  const Surface* referenceSurface = nullptr;
+  /// The filter target surface
+  const Surface* filterTargetSurface = nullptr;
+
+  /// The smoothing target surface
+  const Surface* smoothingTargetSurface = nullptr;
```
AJPfleger, line 200: Maybe we could adapt the comment to explain what the variable is used for or leave it, if we think the variable name is already descriptive enough.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -323,8 +326,11 @@ class CombinatorialKalmanFilter {
     /// Broadcast the result_type
     using result_type = CombinatorialKalmanFilterResult<traj_t>;
 
-    /// The target surface
-    const Surface* targetSurface = nullptr;
+    /// The filter target surface
+    const Surface* filterTargetSurface = nullptr;
+
+    /// The smoothing target surface
+    const Surface* smoothingTargetSurface = nullptr;
```
AJPfleger, line 333: Copy the comment from above,

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
andiwand: I think for now it is fine but I am happy about suggestions

benjaminhuth (approved):
Nice, as I understand the code, there is no change in behaviour if `filterTargetSurface` is `nullptr`. Let's get it in.

PR #2166, opened 2023-05-30
feat: Add ability to skip grid bins in SP grouping and add ITk "fast" tracking config for seeding

LuisFelipeCoelho:
This PR adds the ITk fastTracking configuration for the seeding, which includes:
- A new feature in `binnedSPGroup` that allow us to skip z bins for middle SPs in the search for neighbours SPs
- A new flag in the python ITk configuration for high occupancy seeding

PR #2325, opened 2023-07-27
feat: configurable maximum number of bins in SP grid

LuisFelipeCoelho:
Protection for large number of bins which is set by default to `std::numeric_limits<int>::max()`

This comes from PR #2166, I am splitting the changes into smaller PRs.
@CarloVarni


Core/include/Acts/Seeding/SpacePointGrid.hpp
```diff
@@ -46,6 +46,8 @@ struct SpacePointGridConfig {
   // return 1 neighbor on either side of the current phi-bin (and you want to
   // cover the full phi-range of minPT), leave this at 1.
   int phiBinDeflectionCoverage = 1;
+  // maximum number of phi bins
+  int maxPhiBins = std::numeric_limits<int>::max();
```
CarloVarni, line 50: this is a lot. Maybe we can put a more reasonable default? like 200?

LuisFelipeCoelho: I can set it to 200 but then it will become a breaking change, right?

CarloVarni: why? it's just setting a default value

LuisFelipeCoelho: Because it could cause performance changes to clients?

Core/include/Acts/Seeding/SpacePointGrid.ipp
```diff
@@ -94,6 +94,11 @@ Acts::SpacePointGridCreator::createGrid(
     // consecutive phi bins in the seed making step.
     // Each individual bin should be approximately a fraction (depending on this
     // number) of the maximum expected azimutal deflection.
+
+    // set protection for large number of bins, by default it is infinite
```
paulgessinger, line 98:
```suggestion
    // set protection for large number of bins, by default it is large
```
It's not actually infinite anymore

paulgessinger (commented):
Looks good overall, only spotted one inconsistency with the (now changed) default of 300. I don't have a strong opinion on the default. I guess infinite is a bit dangerous since you could potentially try to allocate infinite memory that way, although I suppose it's bounded by the event activity.

I'd probably pick a "large but not infinite" number, like mayb 10k or so and then mention that in the comment. 

Up to you @LuisFelipeCoelho.

Core/include/Acts/Seeding/SpacePointGrid.ipp
LuisFelipeCoelho: Thank you! changed the default to 10k

paulgessinger (approved):

paulgessinger:
Somehow the clang-tidy job seems to get stuck repeatedly, in that it's result isn't posted back here. It's only that job it seems though, which is weird.

paulgessinger:
Failure is unrelated to us.

PR #2328, opened 2023-07-27
refactor: Separate Surface base class without "obvious" normals

paulgessinger:
So far we assume the normal logic for all surfaces is the same, which is not the case for line surfaces. This PR tries to express this via the type system.
This adds a new intermediate class

```cpp
class RegularSurface : public Surface {};
```

that gets the previous `normal` methods.

The base class `Surface` now only has

```cpp
virtual Vector3 normal(const GeometryContext& gctx, const Vector3& pos,
                                      const Vector3& direction) const = 0;
```

which is valid also for `LineSurface`s.

Another option would be to additionally remove the global position arguments, since (I think) they're not actually relevant, other than for some default calculations.

### DIFF
--- a/Core/include/Acts/Surfaces/Surface.hpp
+++ b/Core/include/Acts/Surfaces/Surface.hpp
@@ -187,41 +187,8 @@ class Surface : public virtual GeometryObject,
   /// @return center position by value
   virtual Vector3 center(const GeometryContext& gctx) const;
 
-  /// Return method for the normal vector of the surface
-  /// The normal vector can only be generally defined at a given local position
-  /// It requires a local position to be given (in general)
-  ///
-  /// @param gctx The current geometry context object, e.g. alignment
-  /// @param lposition is the local position where the normal vector is
-  /// constructed
-  ///
-  /// @return normal vector by value
-  virtual Vector3 normal(const GeometryContext& gctx,
-                         const Vector2& lposition) const = 0;
-
-  /// Return method for the normal vector of the surface
-  /// The normal vector can only be generally defined at a given local position
-  /// It requires a local position to be given (in general)
-  ///
-  /// @param position is the global position where the normal vector is
-  /// constructed
-  /// @param gctx The current geometry context object, e.g. alignment
-
-  ///
-  /// @return normal vector by value
-  virtual Vector3 normal(const GeometryContext& gctx,
-                         const Vector3& position) const;
-
-  /// Return method for the normal vector of the surface
-  ///
-  /// It will return a normal vector at the center() position
-  ///
-  /// @param gctx The current geometry context object, e.g. alignment
-  //
-  /// @return normal vector by value
-  virtual Vector3 normal(const GeometryContext& gctx) const {
-    return normal(gctx, center(gctx));
-  }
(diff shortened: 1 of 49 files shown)


Core/include/Acts/Surfaces/Surface.hpp
```diff
@@ -187,41 +187,8 @@ class Surface : public virtual GeometryObject,
...
-  /// It will return a normal vector at the center() position
-  ///
-  /// @param gctx The current geometry context object, e.g. alignment
-  //
-  /// @return normal vector by value
-  virtual Vector3 normal(const GeometryContext& gctx) const {
-    return normal(gctx, center(gctx));
-  }
+  virtual Vector3 normal(const GeometryContext& gctx, const Vector3& pos,
+                         const Vector3& direction) const = 0;
```
andiwand, line 191: I wonder if it would be safer to ask for the local position as an input. I guess that would be a problem for some of the callers?

with the global position we are not necessarily on surface which can result in ambiguous situations

Fatras/include/ActsFatras/Kernel/detail/SimulationActor.hpp
```diff
@@ -167,7 +167,8 @@ struct SimulationActor {
         // again: interact only if there is valid material to interact with
         if (slab) {
           // adapt material for non-zero incidence
-          auto normal = surface.normal(state.geoContext, local);
+          auto normal = surface.normal(state.geoContext, before.position(),
+                                       before.direction());
```
andiwand, line 171: I guess a material surface should always be a regular one right? just querying with the local position should be fine then

andiwand (commented):
makes it much cleaner :+1: 

left two conceptional questions

Fatras/include/ActsFatras/Kernel/detail/SimulationActor.hpp
paulgessinger: In principle I'd agree, but then I'd have to `dynamic_cast` and check here if it's `RegularSurface`. Maybe that's a better idea?

Core/include/Acts/Surfaces/Surface.hpp
paulgessinger: I'll try today if I can change the signature accordingly.

paulgessinger:
A couple of unit tests fail locally for me, I'm not quite sure why.
I think we have a bunch of cases were we just assume you can turn any arbitrary 3D position into a normal on any surface.

paulgessinger:
Separate branch here where I don't change the `referenceFrame` and `normal` functions to accept local positions, but just check if the global positions are on surface: https://github.com/paulgessinger/acts/tree/normal-imprv-only

This should show all the places where we rely on *projecting* / *ignoring* global positions that are not on surface and returning normals / reference frames anyway.

I think the cleanest solution would be to require the positions to be on surface, and having a well-defined way to *coerce* a positon onto a surface to *force* a normal / reference frame. I haven't been able to get there yet, as it's mostly silent failures from the geometry / navigation, which a notoriously hard to track down.

paulgessinger:
I'll reopen from a new branch.

PR #2332, opened 2023-07-29
feat: Intersection navigation order

andiwand:
This provides a central implementation for intersection ordering for navigation.

This seems to fix the error I was facing in Athena with https://github.com/acts-project/acts/pull/2322

### DIFF
--- a/Core/include/Acts/Utilities/Intersection.hpp
+++ b/Core/include/Acts/Utilities/Intersection.hpp
@@ -94,6 +94,30 @@ struct Intersection {
     // The current one wins, no re-ordering
     return true;
   }
+
+  static bool navigationOrder(const Intersection<DIM>& aIntersection,
+                              const Intersection<DIM>& bIntersection) {
+    if ((aIntersection.status == Status::unreachable) &&
+        (bIntersection.status != Status::unreachable)) {
+      return false;
+    }
+    if ((aIntersection.status != Status::unreachable) &&
+        (bIntersection.status == Status::unreachable)) {
+      return true;
+    }
+    // both are reachable or onSurface now
+
+    auto a = aIntersection.pathLength;
+    auto b = bIntersection.pathLength;
+
+    if ((a >= 0 && b >= 0) || (a <= 0 && b <= 0)) {
+      return a < b;
+    }
+    if (a > 0) {  // b < 0
+      return true;
+    }
+    return false;
+  }
 };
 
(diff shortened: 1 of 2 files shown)


Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -94,6 +94,33 @@ struct Intersection {
...
+                              const Intersection<DIM>& bIntersection) {
+    if ((aIntersection.status == Status::unreachable) &&
+        (bIntersection.status != Status::unreachable)) {
+      return false;
+    }
+    if ((aIntersection.status != Status::unreachable) &&
+        (bIntersection.status == Status::unreachable)) {
+      return true;
+    }
+    // both are reachable or onSurface now
```
felix-russo, line 108: Is there not also the case when both are unreachable?

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -94,6 +94,33 @@ struct Intersection {
     // The current one wins, no re-ordering
     return true;
   }
+
+  static bool navigationOrder(const Intersection<DIM>& aIntersection,
+                              const Intersection<DIM>& bIntersection) {
```
felix-russo, line 99: Could you add a comment here explaining the function?

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -94,6 +94,33 @@ struct Intersection {
...
+    auto a = aIntersection.pathLength;
+    auto b = bIntersection.pathLength;
+
+    if (a >= 0 && b >= 0) {
+      return a < b;
+    }
+    if (a <= 0 && b <= 0) {
+      return a > b;
+    }
+    if (a > 0) {  // b < 0
```
felix-russo, line 119: Could you explain in a comment why we favor forward propagation even if the intersection path length `a` is very large?

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -94,6 +94,33 @@ struct Intersection {
...
+      return true;
+    }
+    return false;
```
felix-russo, line 122: Maybe mention in a comment that here`b >0` and `a<0`?

andiwand:
Unfortunately this does not work as expected. Will provide another solution

PR #2334, opened 2023-07-30
fix: Fix cylinder intersection order

andiwand:
In rare cases we need to propagate through a cylinder volume and only traverse the cylinder surface and not the discs. In this case we need to consider both solutions of the intersection correctly or we end up with faulty surface order or looping navigation.

I experienced this issue in the ITk geometry given a ttbar event and possibly a particle not originating from the IP but from a conversion which can result in such a special case for the navigation.

The proposed fix is to use the navigation order implemented here https://github.com/acts-project/acts/pull/2332 to order the intersections accordingly before they will be picked up by the navigation.

blocked by
- https://github.com/acts-project/acts/pull/2332

### DIFF
--- a/Core/include/Acts/Utilities/Intersection.hpp
+++ b/Core/include/Acts/Utilities/Intersection.hpp
@@ -94,6 +94,30 @@ struct Intersection {
     // The current one wins, no re-ordering
     return true;
   }
+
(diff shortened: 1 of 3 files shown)


andiwand:
Unfortunately this does not work as expected. Will provide another solution

PR #2336, opened 2023-08-01
refactor!: Disambiguate surface intersection solutions

andiwand:
Currently our intersection call can give an alternative solution but it is not clear which one is to be preferred in which situation and the results are unstable meaning that it could be swapped depending on where you are on the ray.

Here I try to make the interface cleaner and the intersection order stable.

blocked by
- https://github.com/acts-project/acts/pull/2366
- https://github.com/acts-project/acts/pull/2368


paulgessinger:
Blocked PRs are merged, marking this as ready for review.

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -39,144 +39,210 @@ inline std::ostream& operator<<(std::ostream& os, IntersectionStatus status) {
...
+    auto b = bIntersection.pathLength();
+    return a < b;
   }
 
-  /// Greater operator for sorting,
-  /// - it respects the validity of the intersection
-  /// @param si is the intersection for testing
-  bool operator>(const Intersection<DIM>& si) const {
-    if (status == Status::unreachable) {
+  static bool closestOrder(const Intersection& aIntersection,
```
paulgessinger, line 76: This method (and `forwardOrder`) could use a doc string describing the behavior. The name helps already but why not make it explicit?

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -39,144 +39,210 @@ inline std::ostream& operator<<(std::ostream& os, IntersectionStatus status) {
...
 using Intersection3D = Intersection<3>;
 
+using MultiIntersection3D = std::array<Intersection3D, 2>;
+
 /// @brief class extensions to return also the object and a representation
 template <typename object_t, typename representation_t = object_t>
 class ObjectIntersection {
  public:
+  /// Default constructor
+  ObjectIntersection() = default;
```
paulgessinger, line 111: Was this default-constructible before (I guess implicitly so)? Do we need this for vectors or is there another reason?

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -39,144 +39,210 @@ inline std::ostream& operator<<(std::ostream& os, IntersectionStatus status) {
...
-  ///
-  /// This operator will ignore the alternative, but simply
-  /// order the representing intersection
-  ///
-  /// @param oi is the source intersection for comparison
-  bool operator>(
-      const ObjectIntersection<object_t, representation_t>& oi) const {
-    return (intersection > oi.intersection);
+  ObjectIntersection<object_t, representation_t> get(std::uint8_t index) const {
+    return {m_intersections[index], m_object, m_representation, index};
```
paulgessinger, line 234: API with `get` both with compile-time and runtime index seems complicated. Do we need this?

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -39,144 +39,210 @@ inline std::ostream& operator<<(std::ostream& os, IntersectionStatus status) {
...
-  Intersection3D intersection{};
+  Intersection3D m_intersection;
   /// The object that was (tried to be) intersected
-  const object_t* object{nullptr};
+  const object_t* m_object = nullptr;
   /// The representation of this object
-  const representation_t* representation{nullptr};
+  const representation_t* m_representation = nullptr;
+  /// The intersection index
+  std::uint8_t m_index = 0;
```
paulgessinger, line 180: Should this be part of the intersection object itself?

Core/include/Acts/Surfaces/Surface.hpp
```diff
@@ -407,8 +409,8 @@ class Surface : public virtual GeometryObject,
   /// @param bcheck the Boundary Check
   /// @param tolerance the tolerance used for the intersection
   ///
-  /// @return SurfaceIntersection object (contains intersection & surface)
-  virtual SurfaceIntersection intersect(
+  /// @return @c SurfaceMultiIntersection object (contains intersection & surface)
+  virtual SurfaceMultiIntersection intersect(
```
paulgessinger, line 413: Hmm, so **all** surfaces will now always return two intersections with up to 2 intersections, but only for cylinder (and cone?) will the second one ever be non-default.
This works, but I'm wondering if there's some way to avoid always having the extra memory footprint.

Core/include/Acts/Propagator/AtlasStepper.hpp
```diff
@@ -364,7 +364,9 @@ class AtlasStepper {
   }
 
   /// Overstep limit
-  double overstepLimit(const State& /*state*/) const { return m_overstepLimit; }
+  double overstepLimit(const State& /*state*/) const {
+    return -m_overstepLimit;
```
paulgessinger, line 368: I'm not sure I understand what's happening here. Does the sign get interpreted differently now (i.e. against the direction)?

Core/include/Acts/Propagator/AtlasStepper.hpp
```diff
@@ -1416,7 +1418,7 @@ class AtlasStepper {
   std::shared_ptr<const MagneticFieldProvider> m_bField;
 
   /// Overstep limit: could/should be dynamic
-  double m_overstepLimit = -100 * UnitConstants::um;
+  double m_overstepLimit = 200 * UnitConstants::um;
```
paulgessinger, line 1421: Numerical value probably doesn't need to change here?

Core/include/Acts/Propagator/StandardAborters.hpp
```diff
@@ -161,23 +164,28 @@ struct SurfaceReached {
...
       navigator.targetReached(state.navigation, true);
-    } else {
-      // Target is not reached, update the step size
-      const double overstepLimit = stepper.overstepLimit(state.stepping);
-      // Check the alternative solution
-      if (distance < overstepLimit and sIntersection.alternative) {
-        // Update the distance to the alternative solution
-        distance = sIntersection.alternative.pathLength;
-      }
-      stepper.setStepSize(state.stepping, distance, ConstrainedStep::aborter,
```
paulgessinger, line 172: Just to confirm, we don't need to handle this case anymore, because the intersections will both be in the candidate list, right?

Core/include/Acts/Vertexing/NumericalTrackLinearizer.ipp
```diff
@@ -115,13 +117,15 @@ Acts::NumericalTrackLinearizer<propagator_t, propagator_options_t>::
                                                    paramVecCopy(eLinTheta));
     // Since we work in 4D we have eLinPosSize = 4
     CurvilinearTrackParameters wiggledCurvilinearParams(
-        paramVecCopy.head(eLinPosSize), wiggledDir, paramVecCopy(eLinQOverP));
+        paramVecCopy.template head<4>(), wiggledDir, paramVecCopy(eLinQOverP));
```
paulgessinger, line 120:
```suggestion
        paramVecCopy.template head<eLinPosSize>(), wiggledDir, paramVecCopy(eLinQOverP));
```
?

Core/src/Geometry/GenericApproachDescriptor.cpp
```diff
@@ -23,25 +23,25 @@ void Acts::GenericApproachDescriptor::registerLayer(const Layer& lay) {
...
-Acts::GenericApproachDescriptor::approachSurface(
+Acts::SurfaceIntersection Acts::GenericApproachDescriptor::approachSurface(
     const GeometryContext& gctx, const Vector3& position,
     const Vector3& direction, const BoundaryCheck& bcheck) const {
   // almost always 2
-  boost::container::small_vector<ObjectIntersection<Surface>, 2> sIntersections;
+  boost::container::small_vector<SurfaceIntersection, 2> sIntersections;
   sIntersections.reserve(m_surfaceCache.size());
-  for (auto& sf : m_surfaceCache) {
+  for (auto sf : m_surfaceCache) {
```
paulgessinger, line 32: Surface copy on purpose?
```suggestion
  for (const auto& sf : m_surfaceCache) {
```
?

paulgessinger (commented):
I like the concept. I'm wondering if we can find a better way for the API other than always returning a pair of intersections. Maybe I'm overthinking it, but it seems wasteful for the other surfaces.

Core/include/Acts/Utilities/Intersection.hpp
benjaminhuth: Agree, would probably use `operator[]` or `at()` for runtime as in stdlib.

andiwand:
> I like the concept. I'm wondering if we can find a better way for the API other than always returning a pair of intersections. Maybe I'm overthinking it, but it seems wasteful for the other surfaces.

agreed but this was already the case before with the alternative

the easiest solution (IMO) would be to enforce single intersections which would make the code downstream much easier but then we need to cut cylinders and cones into pieces

Core/include/Acts/Propagator/StraightLineStepper.hpp
```diff
@@ -222,7 +222,7 @@ class StraightLineStepper {
   /// @param state The stepping state (thread-local cache)
   double overstepLimit(const State& state) const {
     (void)state;
-    return s_onSurfaceTolerance;
+    return -m_overstepLimit;
```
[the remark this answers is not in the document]
andiwand: right but I guess `s_onSurfaceTolerance` is also not a great return value - maybe we should default `m_overstepLimit` to something smaller or to `s_onSurfaceTolerance` (using the correct sign)?

Core/include/Acts/Propagator/StandardAborters.hpp
andiwand: right this should be equivalent to the loop below

Core/include/Acts/Propagator/EigenStepper.hpp
```diff
@@ -156,7 +156,7 @@ class EigenStepper {
 
   /// Constructor requires knowledge of the detector's magnetic field
   EigenStepper(std::shared_ptr<const MagneticFieldProvider> bField,
-               double overstepLimit = 100 * UnitConstants::um);
+               double overstepLimit = 200 * UnitConstants::um);
```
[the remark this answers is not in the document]
andiwand: I was facing overstepping issues at the perigee surface and fixed them temporarily with the increased value

Core/include/Acts/Propagator/AtlasStepper.hpp
andiwand: aligned this one to the `EigenStepper` one

Core/include/Acts/Propagator/AtlasStepper.hpp
andiwand: I aligned this to the `EigenStepper` impl. we store the `m_overstepLimit` with a positive value and then flip it on return

Tests/UnitTests/Core/Geometry/GenericApproachDescriptorTests.cpp
```diff
@@ -110,15 +109,10 @@ BOOST_AUTO_TEST_CASE(GenericApproachNoOverstepping) {
...
   // No overstepping allowed, the preferred solution should be the forward one
-  CHECK_CLOSE_ABS(sfIntersection.intersection.pathLength, 10.5, s_epsilon);
-  CHECK_CLOSE_ABS(sfIntersection.intersection.position.x(), 0., s_epsilon);
-  CHECK_CLOSE_ABS(sfIntersection.intersection.position.y(), 10., s_epsilon);
-  CHECK_CLOSE_ABS(sfIntersection.intersection.position.z(), 1., s_epsilon);
-
-  CHECK_CLOSE_ABS(sfIntersection.alternative.pathLength, -9.5, s_epsilon);
-  CHECK_CLOSE_ABS(sfIntersection.alternative.position.x(), 0., s_epsilon);
-  CHECK_CLOSE_ABS(sfIntersection.alternative.position.y(), -10., s_epsilon);
-  CHECK_CLOSE_ABS(sfIntersection.alternative.position.z(), 1., s_epsilon);
```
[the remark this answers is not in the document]
andiwand: `approachSurface` is does already disambiguate the solutions - not sure if that is good after all :thinking:

Core/include/Acts/Surfaces/Surface.hpp
andiwand: I think this was also the case before with `alternative`

Core/include/Acts/Utilities/Intersection.hpp
andiwand: I think my motivation was that the context of the object will tell you why an index is necessary and the intersection can stay context-less this way

Core/include/Acts/Utilities/Intersection.hpp
andiwand: it was default constructible before. for some reason the diff moved here quite a bit. we need to make it explicit because the other constructor would shadow the default one

Core/include/Acts/Propagator/StraightLineStepper.hpp
```diff
@@ -418,6 +418,9 @@ class StraightLineStepper {
     // return h
     return h;
   }
+
+ private:
+  double m_overstepLimit = 100 * UnitConstants::um;
```
andiwand, line 423:
```suggestion
  double m_overstepLimit = s_onSurfaceTolerance;
```

Core/include/Acts/Utilities/Intersection.hpp
benjaminhuth: but would be interesting to see where a default constructed Intersection is needed, and why...

andiwand: I think currently this is the mechanism for an invalid intersection but this could also be solved with `std::optional`

benjaminhuth: Hmm if this works reliably, I would prefer a private default constructor and a factory `ObjectIntersection::Invalid()` probably...

Core/include/Acts/Utilities/Intersection.hpp
paulgessinger: ok

Tests/UnitTests/Core/Geometry/GenericApproachDescriptorTests.cpp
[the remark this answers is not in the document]
paulgessinger: Ok, so do we not have access to this second intersection anymore at all, and so we can't check it anymore in this unit test?

Core/include/Acts/Utilities/Intersection.hpp
paulgessinger: Personally, I think I'd prefer explicit `std::optional`, but I won't block another solution if you think it makes more sense.

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -39,145 +39,227 @@ inline std::ostream& operator<<(std::ostream& os, IntersectionStatus status) {
...
+
+  constexpr const representation_t* representation() const {
+    return m_representation;
   }
 
-  /// Allow swapping the intersection and the alternative
-  void swapSolutions() { std::swap(intersection, alternative); }
-};
+  constexpr std::array<ObjectIntersection<object_t, representation_t>, 2>
+  split() const {
```
paulgessinger, line 245: Doing this, for example, with a `static_vector`, could explicitly communicate if we have one or two intersections, and not require checking if the intersection object itself is valid.

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -39,145 +39,227 @@ inline std::ostream& operator<<(std::ostream& os, IntersectionStatus status) {
...
+  constexpr ObjectIntersection<object_t, representation_t> closest() const {
+    auto splitIntersections = split();
+    return *std::min_element(
+        splitIntersections.begin(), splitIntersections.end(),
+        ObjectIntersection<object_t, representation_t>::closestOrder);
   }
+
+ private:
+  /// The intersections
+  MultiIntersection3D m_intersections;
```
paulgessinger, line 258: This typedef could literally be a `static_vector<Intersection3D, 2>`, couldn't it?

Core/include/Acts/Propagator/StandardAborters.hpp
```diff
@@ -170,7 +170,8 @@ struct SurfaceReached {
 
     const double pLimit =
         state.stepping.stepSize.value(ConstrainedStep::aborter);
-    const double oLimit = stepper.overstepLimit(state.stepping);
+    // TODO
```
paulgessinger, line 173: TODO what?

Core/include/Acts/Propagator/StandardAborters.hpp
```diff
@@ -170,7 +170,8 @@ struct SurfaceReached {
...
+    const double oLimit = 10 * stepper.overstepLimit(state.stepping);
```
paulgessinger, line 174: Where does the 10 come from?

Tests/UnitTests/Core/TrackFitting/KalmanFitterTests.cpp
```diff
@@ -80,10 +80,10 @@ Acts::CurvilinearTrackParameters makeParameters() {
 const FitterTester tester;
 
 // reconstruction propagator and fitter
+const auto kfLogger = getDefaultLogger("KalmanFilter", Logging::INFO);
 const auto kfZeroPropagator =
     makeConstantFieldPropagator<ConstantFieldStepper>(tester.geometry, 0_T);
-const auto kfZero = KalmanFitter(
-    kfZeroPropagator, getDefaultLogger("KalmanFilter", Logging::INFO));
+const auto kfZero = KalmanFitter(kfZeroPropagator);
```
paulgessinger, line 86: The logger now doesn't go into the fitter anymore?

Core/include/Acts/Propagator/StandardAborters.hpp
andiwand:
```suggestion
    // TODO remove magic number
```

Tests/UnitTests/Core/TrackFitting/KalmanFitterTests.cpp
paulgessinger: Nevermind, I think I tricked myself with the diff.

paulgessinger (commented):
I'd probably take this opportunity to avoid the default-constructed-means-invalid in favor of explicit optionals, but I understand if you don't want to hold this PR for that.

andiwand:
> I'd probably take this opportunity to avoid the default-constructed-means-invalid in favor of explicit optionals, but I understand if you don't want to hold this PR for that.

I guess the only downside of an optional is extra flag bytes if the intersection is present or not but that might be negligible

Core/include/Acts/Propagator/MultiEigenStepperLoop.hpp
```diff
@@ -765,27 +767,17 @@ class MultiEigenStepperLoop
   /// @param release [in] boolean to trigger step size release
   template <typename object_intersection_t>
   void updateStepSize(State& state, const object_intersection_t& oIntersection,
-                      bool release = true) const {
-    const Surface& surface = *oIntersection.representation;
+                      Direction direction, bool release = true) const {
```
andiwand, line 770: this is rather ugly but I don't see a better solution atm

ideas @paulgessinger @benjaminhuth ?

andiwand:
physmon diff seems to be down to https://github.com/acts-project/acts/pull/2441 and CKF diff with truth smeared seeds. To me it looks like we pick up more surfaces which is coming from a diff in the navigation but it is hard to tell which one is correct. Since it is only happening for truth smeared it could point to an existing/fixed navigation issue before the first sensitive surfaces because the other seeders would start directly from that surface

Core/include/Acts/Propagator/MultiEigenStepperLoop.hpp
benjaminhuth: Hmm another possibility would be to make the direction a member of the `MultiEigenStepperLoop::State` and force its initializationi at the beginning of the propagation.
That would keep the interface cleaner...

andiwand: the problem I see with making the stepper aware of the propagation direction is that we need to keep it aligned. some code might flip it in the propagator options on the go but forget about the stepper state property

benjaminhuth: yeah makes sense. then I think this is the cleanest version.

andiwand:
ttbar and GSF changes did not fully disappear but don't look critical to me

Tests/UnitTests/Core/Geometry/GenericApproachDescriptorTests.cpp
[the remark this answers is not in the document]
andiwand: I moved the decision into the `GenericApproachDescriptor`. Before it also allowed for reordering of `solution` and `alternative` and I though the interface is nicer if we can hide that detail. The downside is that we now need to pass in the details of `pLimit, oLimit, tolerance`

Core/include/Acts/Propagator/StandardAborters.hpp
```diff
@@ -161,23 +167,31 @@ struct SurfaceReached {
...
+      return true;
     }
-    // path limit check
-    return targetReached;
+
+    const double pLimit =
+        state.stepping.stepSize.value(ConstrainedStep::aborter);
+    // not using the stepper overstep limit here because it does not always work
+    // for perigee surfaces
+    const double oLimit = overstepLimit;
```
andiwand, line 178: I needed to increase the overstep limit here because perigee surface do not play as nice as plane surfaces do

before my changes we ignored the overstepping tolerance completely but I don't think that is great either

Core/include/Acts/Geometry/ApproachDescriptor.hpp
```diff
@@ -43,11 +43,17 @@ class ApproachDescriptor {
...
   /// @return is a surface intersection
-  virtual SurfaceIntersection approachSurface(
-      const GeometryContext& gctx, const Vector3& position,
-      const Vector3& direction, const BoundaryCheck& bcheck) const = 0;
+  virtual SurfaceIntersection approachSurface(const GeometryContext& gctx,
+                                              const Vector3& position,
+                                              const Vector3& direction,
+                                              const BoundaryCheck& bcheck,
+                                              double pLimit, double oLimit,
+                                              double tolerance) const = 0;
```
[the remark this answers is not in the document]
andiwand: that would make a lot of sense but this would touch a lot of code if we want to align everything. in this case it might be for another pr

Core/include/Acts/Propagator/MultiEigenStepperLoop.hpp
```diff
@@ -772,29 +772,21 @@ class MultiEigenStepperLoop
   ///
   /// @param state [in,out] The stepping state (thread-local cache)
   /// @param oIntersection [in] The ObjectIntersection to layer, boundary, etc
+  /// @param direction [in] The propagation direction
   /// @param release [in] boolean to trigger step size release
   template <typename object_intersection_t>
   void updateStepSize(State& state, const object_intersection_t& oIntersection,
-                      bool release = true) const {
+                      Direction direction, bool release = true) const {
```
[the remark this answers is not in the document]
andiwand: right, the `MultiEigenStepperLoop` is the only one trying to intersect with the surface which requires the overall navigation direction

Core/include/Acts/Propagator/StraightLineStepper.hpp
```diff
@@ -218,10 +218,7 @@ class StraightLineStepper {
   double time(const State& state) const { return state.pars[eFreeTime]; }
 
   /// Overstep limit
-  ///
-  /// @param state The stepping state (thread-local cache)
```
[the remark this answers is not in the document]
paulgessinger: Oh I see, doxygen complains about the parameter if it's not named...

andiwand:
this is causing navigation failure in the `full_chain_odd.py` which was not the case in a previous version I believe. I am investigating

andiwand:
this seems to work now - even tho I am not super happy with the fix but I don't see an alternative

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -1268,7 +1271,7 @@ class CombinatorialKalmanFilter {
     source_link_accessor_t m_sourcelinkAccessor;
 
     /// The Surface being targeted
-    SurfaceReached targetReached;
+    SurfaceReached targetReached{std::numeric_limits<double>::lowest()};
```
andiwand, line 1274: here

Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -323,7 +324,7 @@ class KalmanFitter {
     KalmanFitterExtensions<traj_t> extensions;
 
     /// The Surface being
-    SurfaceReached targetReached;
+    SurfaceReached targetReached{std::numeric_limits<double>::lowest()};
```
andiwand, line 327: here

andiwand (commented):
these are the new changes in question

paulgessinger:
Is this why the physmon still shows differences?

andiwand:
> Is this why the physmon still shows differences?

I just didn't update the reference yet because I was unsure if this one or https://github.com/acts-project/acts/pull/2470 should go in first

paulgessinger (approved):

acts-project-service:
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/32661641) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6236237)!

paulgessinger:
This does break compilation in the material track writer in Athena.

PR #2339, opened 2023-08-02
ci: Create job summaries in python

paulgessinger:


CI/physmon/phys_perf_mon.sh
```diff
@@ -247,7 +247,10 @@ if [[ "$mode" == "all" || "$mode" == "simulation" ]]; then
     simulation geant4
 fi
 
-CI/physmon/summary.py $outdir/*.html $outdir/summary.html
+CI/physmon/summary.py $outdir/*.html summary.md
```
andiwand, line 250: maybe the python script could output both `md` and `html` just to remove a bit of logic from bash

andiwand: yeah I think we should preserve single source. so in python we could copy paste to the other output and prepend the html stuff

paulgessinger:
The texme header doesn't work as well as I thought it would, and also the links don't currently work in the job summary (I have an idea how to fix that) so I'll switch to just using multiple separate outputs from python

andiwand (dismissed):
nice!

paulgessinger:
Ok, I added some functionality to herald to redirect to an artifact from a run id while maintaining paths. Now the physmon summary markdown should contain working links.

andiwand (dismissed):

CI/physmon/summary.py
```diff
@@ -34,29 +38,51 @@
         except Exception as e:
             print(r"could not parse {h}", e)
 
-with open(args.output, mode="w", encoding="utf-8") as f:
-    f.write(
-        """<!DOCTYPE html>
+output = "# physmon summary\n"
```
andiwand, line 41:
```suggestion
```
dead variable?

CI/physmon/summary.py
```diff
@@ -34,29 +38,51 @@
...
+            """
         )
 
-    f.write(
-        """
-  </ul>
-</body>
-</html>"""
-    )
+is_ci = "GITHUB_ACTIONS" in os.environ
```
andiwand, line 73:
```suggestion
```
should we move this up?

CI/physmon/summary.py
```diff
@@ -6,18 +6,22 @@
 import functools
 import os
 
+HERALD_URL = "https://herald.dokku.paulgessinger.com/view/{repo}/runs/{run_id}/artifacts/{artifact_name}/{path}"
```
andiwand, line 9:
```suggestion
HERALD_URL = "https://herald.dokku.paulgessinger.com/view/{repo}/runs/{run_id}/artifacts/{artifact_name}/{path}"
IS_CI = "GITHUB_ACTIONS" in os.environ
```

CI/physmon/summary.py
```diff
@@ -34,29 +38,51 @@
...
-</body>
-</html>"""
-    )
+is_ci = "GITHUB_ACTIONS" in os.environ
+if args.md:
+    with open(args.md, mode="w", encoding="utf-8") as f:
+        f.write("# physmon summary\n")
+        for h, s in summary.items():
+            path = os.path.relpath(h, args.base)
+            if is_ci:
```
andiwand, line 79:
```suggestion
            if IS_CI:
```

andiwand (approved):

andiwand (approved):

PR #2340, opened 2023-08-02
refactor!: Introduce regular surface intermediate class

paulgessinger:
This adds a new class

```cpp
class RegularSurface : public Surface {};
```
that reimplements a number of functions while discarding the `direction` argument.

PR #2344, opened 2023-08-03
refactor: Require position for normal calculation to be on-surface

paulgessinger:
We currently internally project / coerce to the respective surface, but I feel like this is non obvious, and the caller has no control over this. Here, I propose to make this explicit by having another function `coerceToSurface` which projects a global position onto a surface, where the details of how that is done are depending on the surface.

Blocked by 
- #2340

### DIFF
--- a/Core/src/Surfaces/DiscSurface.cpp
+++ b/Core/src/Surfaces/DiscSurface.cpp
@@ -347,6 +347,10 @@ Acts::ActsMatrix<2, 3> Acts::DiscSurface::localCartesianToBoundLocalDerivative(
 
 Acts::Vector3 Acts::DiscSurface::normal(const GeometryContext& gctx,
                                         const Vector2& /*lposition*/) const {
+  return normal(gctx);
+}
+
+Acts::Vector3 Acts::DiscSurface::normal(const GeometryContext& gctx) const {
   // fast access via transform matrix (and not rotation())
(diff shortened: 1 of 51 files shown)


Core/src/Surfaces/PlaneSurface.cpp
```diff
@@ -173,6 +173,14 @@ Acts::Vector3 Acts::PlaneSurface::binningPosition(
   return center(gctx);
 }
 
+Acts::Vector3 Acts::PlaneSurface::coerceToSurface(
+    const GeometryContext& gctx, const Vector3& position) const {
+  // @TODO: Bypass the transform matrix multiplication
```
paulgessinger, line 178: This can be optimized I think.

Core/src/Surfaces/DiscSurface.cpp
```diff
@@ -360,6 +360,14 @@ Acts::Vector3 Acts::DiscSurface::normal(const GeometryContext& gctx) const {
   return Vector3(tMatrix(0, 2), tMatrix(1, 2), tMatrix(2, 2));
 }
 
+Acts::Vector3 Acts::DiscSurface::coerceToSurface(
+    const GeometryContext& gctx, const Vector3& position) const {
```
paulgessinger, line 364: This can be optimized I think.

paulgessinger:
We're not going ahead with this in the near future, so closing.

PR #2349, opened 2023-08-04
refactor: FPE masking updates & improvements

paulgessinger:
This PR does three things:

1. change fpe mask comment format `issue:1234` -> `#1234`
2. Line ranges now possible in FPE masking
  This works both with the explicit mask lists / yaml inputs, but also with markers like
  ```cpp
  // MARK: fpeMaskBegin(FLTINV, 1, issue:2284)
  thisMakesAnFLTINV();
  // MARK: fpeMaskEnd(FLTINV)
  ```
3. Improved FPE mask reporting in interactive running: When launching the Sequencer, the masks will now be checked if their file paths are valid, and it will show the masked pieces of code.


Examples/Python/python/acts/examples/__init__.py
```diff
@@ -441,21 +472,165 @@ def _getAutoFpeMasks(cls) -> List[FpeMask]:
...
+
+            print = rich.print
+            print(rich.rule.Rule("FPE masks"))
+
+            nonlocal error
+
+            for i, mask in enumerate(masks):
+                if i > 0:
+                    print(rich.rule.Rule())
+                #  print(f"- {mask.file}:{mask.lines[0]}: {mask.type.name}: {mask.count}")
```
niermann999, line 563: leftover?

niermann999 (approved):

Examples/Python/python/acts/examples/__init__.py
paulgessinger: Dang

paulgessinger, line 563:
```suggestion
```

CI/physmon/workflows/physmon_track_finding_ttbar.py
```diff
@@ -38,32 +38,38 @@
         Path(__file__).parent.parent / "fpe_masks.yml"
     ) + [
         acts.examples.Sequencer.FpeMask(
-            "Examples/Algorithms/Fatras/src/FatrasSimulation.cpp:172",
+            "Examples/Algorithms/Fatras/src/FatrasSimulation.cpp",
+            (172, 173),
             acts.FpeType.FLTINV,
             1,
```
andiwand, line 44: should we also require an issue number on these kind of masks?

Examples/Framework/include/ActsExamples/Framework/Sequencer.hpp
```diff
@@ -58,7 +58,8 @@ class SequenceConfigurationException : public std::runtime_error {
 class Sequencer {
  public:
   struct FpeMask {
-    std::string loc;
+    std::string file;
+    std::pair<unsigned int, unsigned int> lines;
```
andiwand, line 62:
```suggestion
    std::pair<std::size_t, std::size_t> lines;
```
nitpick: or `std::uint32_t`

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -655,9 +656,17 @@ std::pair<std::string, std::size_t> Sequencer::fpeMaskCount(
...
+    for (const auto& [file, lines, fType, count] : m_cfg.fpeMasks) {
+      const auto [start, end] = lines;
+      if (boost::algorithm::ends_with(locFile, file) &&
+          (start <= locLine && locLine < end) && fType == type) {
+        std::string ls = start + 1 == end ? std::to_string(start)
+                                          : "(" + std::to_string(start) + ", " +
+                                                std::to_string(end) + "]";
+        return {file + ":" + ls, count};
       }
     }
```
andiwand, line 671: maybe this is worth an utility parsing function

nitpick: `unsigned int`

Examples/Python/python/acts/examples/__init__.py
```diff
@@ -441,21 +472,164 @@ def _getAutoFpeMasks(cls) -> List[FpeMask]:
...
+            import rich
+
+            have_rich = True
+        except ImportError:
+            have_rich = False
+
+        error = False
+        srcdir = cls.srcdir()
+
+        def rich_summary():
```
andiwand, line 546: since this is only called once and would be intended at the same level I would maybe inline it but not sure

andiwand (approved):
left a few suggestions but already looks good as is. feel free to automerge

CI/physmon/workflows/physmon_track_finding_ttbar.py
paulgessinger: We could, but it's not trivial to ensure they're valid. For the code comments, there is a specific syntax that I can look for, whereas this here is only accessible from the job run itself. Any ideas?

Examples/Python/python/acts/examples/__init__.py
paulgessinger: Can do.

Examples/Framework/src/Framework/Sequencer.cpp
paulgessinger: I think this **is** the utility parsing function 😅

Examples/Framework/include/ActsExamples/Framework/Sequencer.hpp
andiwand: no particular - I guess mostly to ensure the size and consistency e.g. we use `std::size_t count;` below

but its only a small preference from my side. not blocking in any way

paulgessinger: I'll change it.

CI/physmon/workflows/physmon_track_finding_ttbar.py
andiwand: sure!

felix-russo (approved):

PR #2358, opened 2023-08-09
fix: Generic Detector negative strip endcap rotation

paulgessinger:
Sensors were flipped along local z, meaning they are the wrong way around.


paulgessinger:
As expected, physmon is happy but a lot of hashes change.

andiwand:
I fear we also need some updates in `test_digitization_example` - there are hardcoded volumes and layers

paulgessinger:
@andiwand That's confusing actually. Does this change the module numbers maybe?

andiwand:
If I understand this correctly the geoid should be the same no? because we only rotate/flip the modules

but we might hit them differently because of the geometry change which might cause this test to fail

paulgessinger:
Does this rotate the module about the center of the disk and not the module center? Or why would the phi values changes?

niermann999:
Nothing is rotated compared to the previous version. The translation was such that a correctly rotated modules was placed on the wrong side of the disc (mirrored at the origin compared to the positive endcap) thus effectively flipping it. If this becomes a problem with the module numbers, then we have to solve it indeed by changing the rotation instead of the translation and flip the modules instead of placing them differently, but that would make the code less intuitive, I thought

paulgessinger:
Why does this keep flip-flopping between failed hashes and conflicts in the hash file 😠

acts-project-service:
@paulgessinger
🟡 I'm going to trigger an ATLAS experiment pipeline for you:


| Variable | Value |
|------|------|
| `ACTS_GIT_REPO` | https://github.com/paulgessinger/acts.git |
| `ACTS_REF` | `unbreak-generic-strip-endcaps` |
| `SOURCE_SHA` | 4f14f26e4bcd5213941a30b9c26bbb349603be49 |
| `REVERT_SHAS` |  |

niermann999 (approved):

PR #2359, opened 2023-08-11
refactor: fittedParams -> fittedMomentum in TrackAtVertex struct

felix-russo:
The output of a vertex fitter is

- a vertex position
- the track _momenta_ at this position

Previously, the track momenta from the vertex fit were saved as `GenericTrackParameters`, where we set `d0 = z0 = 0`. For the corresponding covariance matrices, we partly used incorrect values (one cannot, in fact, compute entries like `Cov(d0, phi)` or `Cov(z0, z0)`  as `d0` and `z0` are not part of the vertex fitting model).

In this PR, `fittedParams` are replaced by `fittedMomenta` (i.e., a 3 vector + its covariance matrix). My aim is 

- to make the way the vertex fitter works more obvious and 
- to remove the incorrect covariances.


felix-russo:
Physmon failure is expected:
Before, we had a lot of tracks whose `fittedParams` were set to `originalParams` upon initialization and that were not updated during the vertex fit, see #2364. This explains why the monitored histograms have less entries.

felix-russo:
Explanation of the physmon failure:

- In the old code, `fittedParams` were set to `originalParams` upon initialization. For some tracks, the former were never updated during the AMVF vertex fit due to a bug, see https://github.com/acts-project/acts/issues/2364. Now, `fittedMomentum` is `std::optional`, and, instead of not being updated, it remains `std::nullopt`. This explains why there are less track momenta monitored than in the reference for the AMVF.
- The computation of the covariance of phi was faulty in the old code. This explains why `trk_pullPhiFitted` now has a wider distribution.

Core/include/Acts/Vertexing/TrackAtVertex.hpp
```diff
@@ -15,45 +15,67 @@
 
 namespace Acts {
 
-/// @class TrackAtVertex
+/// @struct FittedMomentum
+///
+/// @brief Vertex fitters return a vertex position and updated track momenta at said position (+corresponding covariances). The updated track momenta and their covariances are saved in the following struct.
+struct FittedMomentum {
+  FittedMomentum(Vector3 mom, ActsSymMatrix<3> cov)
```
andiwand, line 22:
```suggestion
  FittedMomentum(const Vector3& mom, const ActsSymMatrix<3>& cov)
```

we cannot `std::move` those so getting a `const &` is avoiding a second copy

Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
```diff
@@ -673,13 +673,11 @@ ActsExamples::ProcessCode ActsExamples::VertexPerformanceWriter::writeT(
               }
 
               // Save track parameters after the vertex fit
-              const auto paramsAtVtxFitted = propagateToVtx(trk.fittedParams);
-              if (paramsAtVtxFitted != std::nullopt) {
+              if (trk.fittedMomentum != std::nullopt) {
```
andiwand, line 676:
```suggestion
              if (trk.fittedMomentum) {
```

nitpick: or `has_value` is preferred IMO

Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
```diff
@@ -673,13 +673,11 @@ ActsExamples::ProcessCode ActsExamples::VertexPerformanceWriter::writeT(
...
                 Acts::ActsVector<3> recoMomFitted =
-                    paramsAtVtxFitted->parameters().segment(Acts::eBoundPhi, 3);
+                    trk.fittedMomentum.value().momentum;
```
andiwand, line 678:
```suggestion
                    trk.fittedMomentum->momentum;
```

`value()` will actually check if there is a value and throw and exception otherwise. since we know that the optional has a value we can directly access the memory and skip the condition

Examples/Io/Performance/ActsExamples/Io/Performance/VertexPerformanceWriter.cpp
```diff
@@ -673,13 +673,11 @@ ActsExamples::ProcessCode ActsExamples::VertexPerformanceWriter::writeT(
...
                 const Acts::ActsMatrix<3, 3>& momCovFitted =
-                    paramsAtVtxFitted->covariance()->block<3, 3>(
-                        Acts::eBoundPhi, Acts::eBoundPhi);
+                    trk.fittedMomentum.value().covariance;
```
andiwand, line 680:
```suggestion
                    trk.fittedMomentum->covariance;
```

Tests/UnitTests/Core/Vertexing/AdaptiveMultiVertexFitterTests.cpp
```diff
@@ -213,20 +213,18 @@ BOOST_AUTO_TEST_CASE(adaptive_multi_vertex_fitter_test) {
     int vtxIdx = (int)(iTrack / nTracksPerVtx);
     state.vtxInfoMap[&(vtxList[vtxIdx])].trackLinks.push_back(
         &(allTracks[iTrack]));
-    state.tracksAtVerticesMap.insert(
-        std::make_pair(std::make_pair(&(allTracks[iTrack]), &(vtxList[vtxIdx])),
-                       TrackAtVertex<BoundTrackParameters>(
-                           1., allTracks[iTrack], &(allTracks[iTrack]))));
+    state.tracksAtVerticesMap.insert(std::make_pair(
+        std::make_pair(&(allTracks[iTrack]), &(vtxList[vtxIdx])),
+        TrackAtVertex<BoundTrackParameters>(&(allTracks[iTrack]), 1.)));
```
andiwand, line 218:
```suggestion
        std::make_pair(&allTracks[iTrack], &vtxList[vtxIdx]),
        TrackAtVertex<BoundTrackParameters>(&allTracks[iTrack], 1.)));
```
nitpick: that `()` syntax looks a bit weird to me. same below

andiwand (commented):
one conceptional question:

I guess the fitter is only modifying the momentum but not the position of the track right? Do we constrain the position somehow, e.g. the position of the vertex?

Is it still possible to access the full refitted track parameters in some way? I don't have a use case in mind but I guess it could be handy if somebody wants to use them for something.

paulgessinger:
> I guess the fitter is only modifying the momentum but not the position of the track right? Do we constrain the position somehow, e.g. the position of the vertex?

that's the thing (if i understand correctly). previously we would use the vertex position, but that gives an inconsistent parameter set and an inconsistent covariance, which I found confusing / surprising.

felix-russo:
> I guess the fitter is only modifying the momentum but not the position of the track right? Do we constrain the position somehow, e.g. the position of the vertex?

Yes, the model assumes that position of the track is exactly at the reconstructed vertex.

> Is it still possible to access the full refitted track parameters in some way? I don't have a use case in mind but I guess it could be handy if somebody wants to use them for something.

The full refitted track parameters would be `(0, 0, phiFitted, thetaFitted, qOverPFitted)`. I will add a comment to explain this better! 

Since the vertex fitting model does not provide covariances for the impact parameters, we previously used the vertex covariance matrix to compute them. For example `Cov(d0, d0)` was a function of `Cov(xVertex, xVertex), Cov(xVertex, yVertex), etc. ` as Paul mentioned correctly.

One could argue that we could set `Cov(d0, d0) = Cov(sqrt(xVertex^2+yVertex^2), sqrt(xVertex^2+yVertex^2))`, but we agreed that this would be a bit sketchy. If they want to, users could do this computation themselves since we provide the vertex covariance.

andiwand (approved):
do we know if this interface is consumed in Athena? if not we can go forward and merge this

andiwand (approved):

acts-project-service:
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/31887589) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6079983)!

paulgessinger:
Hey @felix-russo, @andiwand. I realize now that the output of this is extensively used in Athena.
Maybe we need to discuss this in the ATLAS vertexing meeting before making this change?

felix-russo:
> Hey @felix-russo, @andiwand. I realize now that the output of this is extensively used in Athena. 

Oh really? I thought they didn't care about the refitted momenta?

> Maybe we need to discuss this in the ATLAS vertexing meeting before making this change?

Egoistically, I would like to get this in as fast as possible to continue working on the negative AMVF variances. However, I realize that it's you who will have to pick up the pieces. So, in the end, it's your call!

> In any case, this will need some amount of validation.

I am happy to provide it (please let me know what you would like to see).

PR #2365, opened 2023-08-14
ci: Fix clang tidy job

andiwand:
WIPWIP

Apparently our clang tidy is not doing its job correctly for a while. I am trying to fix this here.

This will also require code changes that were not compliant from the past

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,7 +40,7 @@ clang_tidy:
     - mkdir clang-tidy
 
     # Main clang-tidy run during cmake compilation
-    - CI/clang_tidy/run_clang_tidy.sh build | tee clang-tidy/clang-tidy.log
+    - CI/clang_tidy/run_clang_tidy.sh build -j2 | tee clang-tidy/clang-tidy.log
 
     # Install dependencies for processing scripts
     - pip install -r CI/clang_tidy/requirements.txt
--- a/CI/clang_tidy/run_clang_tidy.sh
+++ b/CI/clang_tidy/run_clang_tidy.sh
@@ -1,8 +1,12 @@
 #!/bin/bash
+
+set -e
+
 build_dir=$1
+shift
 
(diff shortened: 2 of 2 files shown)


paulgessinger:
I think it did run most of the time, and only gets killed when running on a low-ish memory machine.

.gitlab-ci.yml
```diff
@@ -40,7 +40,7 @@ clang_tidy:
     - mkdir clang-tidy
 
     # Main clang-tidy run during cmake compilation
-    - CI/clang_tidy/run_clang_tidy.sh build | tee clang-tidy/clang-tidy.log
```
paulgessinger, line 43: maybe we should do something like 

```
min(nproc, memory / 2)
```

or so?

andiwand: hmm how can this be done in this config file?

paulgessinger:
Hmm this runs for almost 2 hours now. I think we probably have to stick to full core count so it doesn't take forever. 

I've gone back to full VMs per job so hopefully the clang-tidy should be less crashy.

andiwand:
makes sense! I will revert the `-j` but otherwise I think the reporting and error handling still improved

.gitlab-ci.yml
```diff
@@ -37,10 +37,8 @@ clang_tidy:
         -DACTS_BUILD_EVERYTHING=on
         -DACTS_RUN_CLANG_TIDY=on
 
-    - mkdir clang-tidy
-
     # Main clang-tidy run during cmake compilation
-    - CI/clang_tidy/run_clang_tidy.sh build | tee clang-tidy/clang-tidy.log
+    - CI/clang_tidy/run_clang_tidy.sh clang-tidy build -j2
```
andiwand, line 41:
```suggestion
    - CI/clang_tidy/run_clang_tidy.sh clang-tidy build
```

PR #2341, opened 2023-08-03
refactor!: Particle hypothesis for track parameters

andiwand:
In this PR I make the particle hypothesis part of the track parameters. This allows us to consistently use the same hypothesis across various parts of our code. Before that the particle hypothesis was a bit scattered in the propagator and stepper.

replaces https://github.com/acts-project/acts/pull/2181 after splitting it up

blocked by
- https://github.com/acts-project/acts/pull/2366
- https://github.com/acts-project/acts/pull/2396
- https://github.com/acts-project/acts/pull/2398
- https://github.com/acts-project/acts/pull/2397

PR #2366, opened 2023-08-14
refactor: Rework interactions interface - more explicit params and types

andiwand:
Refactors the interactions interface to be minimal. That means only passing absolute charge and absolute PDG and only where necessary.

At the same time I made the types explicit. For some reason we use `float` all across the interaction code which I kept and made more explicit.

also fixes a weird FPE discovered in
- https://github.com/acts-project/acts/pull/2336

pulled these changes out of
- https://github.com/acts-project/acts/pull/2341
- https://github.com/acts-project/acts/pull/2336

blocked by
- https://github.com/acts-project/acts/pull/2375
- https://github.com/acts-project/acts/pull/2372


andiwand:
AMVF physmon is failing for unknown reasons. I experienced this already multiple times - minimal changes not directly related to the AMVF will be amplified greatly.

I will try a few things here to nail this down https://github.com/acts-project/acts/pull/2370

My guess is that either there is some weird interference because we run the IVF in the same chain or the AMVF is operating on a critical point in our setup which causes vertices to appear/disappear based on small numerical differences which then show up as bigger differences in the hist cmp

benjaminhuth (dismissed):
Looks good!

Core/include/Acts/Material/Interactions.hpp
```diff
@@ -22,29 +23,15 @@ struct RelativisticQuantities {
...
-    const auto mOverP = mass * std::abs(qOverP / q);
-    const auto pOverM = 1.0f / mOverP;
-    // beta² = p²/E² = p²/(m² + p²) = 1/(1 + (m/p)²)
-    beta2 = 1.0f / (1.0f + mOverP * mOverP);
-    // beta*gamma = (p/sqrt(m² + p²))*(sqrt(m² + p²)/m) = p/m
-    betaGamma = pOverM;
-    // gamma = sqrt(m² + p²)/m = sqrt(1 + (p/m)²)
-    gamma = std::sqrt(1.0f + pOverM * pOverM);
-  }
+  RelativisticQuantities(float mass, float qOverP, float absQ);
```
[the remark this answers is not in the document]
andiwand: there was one use in the header but I eliminated it

Core/include/Acts/Propagator/MaterialInteractor.hpp
```diff
@@ -94,8 +94,8 @@ struct MaterialInteractor {
 
       if (energyLoss) {
         using namespace UnitLiterals;
-        ACTS_VERBOSE(d.slab << " pdg=" << d.pdg << " mass=" << d.mass / 1_MeV
-                            << "MeV"
+        ACTS_VERBOSE(d.slab << " absPdg=" << d.absPdg
+                            << " mass=" << d.mass / 1_MeV << "MeV"
```
paulgessinger, line 98: Do we have the charge sign here? We could present a signed PDG just for information purposes.

Core/include/Acts/Propagator/detail/GenericDenseEnvironmentExtension.hpp
paulgessinger, line 1: The charges were defaulted to `e` before. Was this intentional maybe? Or does `absCharge` de-facto contain `e` also after this change?

Core/src/Material/Interactions.cpp
```diff
@@ -175,56 +214,51 @@ float Acts::deriveEnergyLossBetheQOverP(const MaterialSlab& slab,
...
   }
 
-  const auto I = slab.material().meanExcitationEnergy();
-  const auto Ne = slab.material().molarElectronDensity();
-  const auto thickness = slab.thickness();
-  const auto rq = Acts::RelativisticQuantities(m, qOverP, q);
-  const auto eps = computeEpsilon(Ne, thickness, rq);
-  const auto dhalf = computeDeltaHalf(I, Ne, rq);
-  const auto u = computeMassTerm(Me, rq);
+  const auto rq = RelativisticQuantities(m, qOverP, absQ);
```
paulgessinger, line 235: It's a style choice, but I personally don't like `auto` when the standard construction syntax could be used.

Arguably this could be

```cpp
const RelativisticQuantities rq{m, qOverP, absQ};
```

Also in some cases below.

Core/src/Material/Interactions.cpp
```diff
@@ -75,95 +102,107 @@ inline float logDeriveWMax(float mass, float qOverP,
...
+inline float deriveDeltaHalf(float qOverP, const RelativisticQuantities& rq) {
   // original equation is of the form
   //     log(beta*gamma) + log(eplasma/I) - 1/2
   // which the resulting derivative as
   //     d(beta*gamma)/(beta*gamma)
   return (rq.betaGamma < 10.0f) ? 0.0f : (-1.0f / qOverP);
 }
 
-}  // namespace
+inline float _computeEnergyLossLandauFwhm(const Acts::MaterialSlab& slab,
```
paulgessinger, line 140: Consider using a `detail` namespace (even if it's just in the `.cpp`) instead of prepending the `_` to disambiguate the function names.

Core/src/Material/Interactions.cpp
```diff
@@ -367,51 +374,57 @@ inline float deriveMuonDirectPairPhotoNuclearLossMeanE(double energy) {
...
+  const float momentum = absQ / qOverP;
+  const float energy = std::hypot(m, momentum);
+
+  float dEdx = computeBremsstrahlungLossMean(m, energy);
 
-  auto dEdx = computeBremsstrahlungLossMean(m, energy);
-  if (((pdg == PdgParticle::eMuon) or (pdg == PdgParticle::eAntiMuon)) and
-      (8_GeV < energy)) {
+  // muon- or muon+
+  if ((absPdg == PdgParticle::eMuon) and (8_GeV < energy)) {
```
paulgessinger, line 398: Do we actually know where the 8 GeV come from? 😄

Fatras/include/ActsFatras/EventData/Particle.hpp
```diff
@@ -152,6 +152,10 @@ class Particle {
   constexpr ProcessType process() const { return m_process; }
   /// PDG particle number that identifies the type.
   constexpr Acts::PdgParticle pdg() const { return m_pdg; }
+  /// Absolute PDG particle number.
+  constexpr Acts::PdgParticle absolutePdg() const {
```
paulgessinger, line 156: `absolutePdg` vs `absPdg`

paulgessinger (commented):
You remove some assertions. Are you removing them because we're confident they should never trigger?

Core/include/Acts/Propagator/MaterialInteractor.hpp
andiwand: we could get the charge from the stepper. would it be as easy as `absPdg * charge * -1`?

Fatras/include/ActsFatras/EventData/Particle.hpp
andiwand: below we have `absoluteCharge` that is why I was a bit conflicted with `absPdg` :smile:

Core/src/Material/Interactions.cpp
andiwand: I don't :smile:

Core/src/Material/Interactions.cpp
andiwand: right, I left this as is but I can adopt that

Core/include/Acts/Propagator/detail/GenericDenseEnvironmentExtension.hpp
andiwand: I think this was just implicitly a single charge hypothesis. At least I don't see why we would not forward the actual charge hypothesis

andiwand:
> You remove some assertions. Are you removing them because we're confident they should never trigger?

I moved the assertions into the constructor of `RelativisticQuantities` so it should result in the same behavior

Core/include/Acts/Propagator/MaterialInteractor.hpp
paulgessinger: I suppose so?

Core/src/Material/Interactions.cpp
andiwand: added a `TODO`

paulgessinger (dismissed):

benjaminhuth (approved):

acts-project-service:
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/31872486) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6076851)!

PR #2368, opened 2023-08-15
fix: Reset navigation state correctly in KF reverse pass

andiwand:
We use a different abort mechanism in the KF and do not need to tell the navigator about the target surface

Pulled this out of https://github.com/acts-project/acts/pull/2336 where it fixed a bug

### DIFF
--- a/Core/include/Acts/TrackFitting/KalmanFitter.hpp
+++ b/Core/include/Acts/TrackFitting/KalmanFitter.hpp
@@ -538,7 +538,7 @@ class KalmanFitter {
       navigator.resetState(
           state.navigation, state.geoContext, stepper.position(state.stepping),
           state.options.direction * stepper.direction(state.stepping),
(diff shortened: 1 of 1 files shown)


Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -538,7 +538,7 @@ class KalmanFitter {
       navigator.resetState(
           state.navigation, state.geoContext, stepper.position(state.stepping),
           state.options.direction * stepper.direction(state.stepping),
-          &st.referenceSurface(), targetSurface);
+          &st.referenceSurface(), nullptr);
```
benjaminhuth, line 541: Just a suggestion: Maybe add a comment, that no target must be set here because there is a custom aborter, that stops on `result.finished` or something like that.

benjaminhuth (approved):

benjaminhuth (approved):

PR #2369, opened 2023-08-15
perf: `LineSurface` back to `col(2)` for local z axis

andiwand:
Back to `col(2)` as it seems more optimal in terms of resources

PR #2382, opened 2023-08-18
perf: `LineSurface` back to `col(2)` for local z axis #2

andiwand:
second try after https://github.com/acts-project/acts/pull/2369

Back to `col(2)` as it seems more optimal in terms of resources


paulgessinger (approved):
Let's go

paulgessinger:
Build failure unrelated, but this did in fact change the Athena outputs.

PR #2387, opened 2023-08-22
refactor!: Rename ActsSymMatrix -> ActsSquareMatrix

felix-russo:
A square matrix has the same number of rows and columns.
A symmetric matrix is a square matrix with transposition symmetry, i.e., `A.transpose() == A`.

Since we don't impose transposition symmetry when initializing an `ActsSymMatrix`, I propose to rename it `ActsSquareMatrix`.


benjaminhuth:
Good thing! :+1: I thought about this as well a lot of times, but was to lazy to make a PR...

benjaminhuth (approved):
Probably this should go into v29, right?

felix-russo:
> Probably this should go into v29, right?

I would say so, yes!

paulgessinger:
I'm going to guess that this will break a LOT of downstream code...

andiwand:
if so we can try to split up the work of fixing Athena

felix-russo:
I would help too of course!

paulgessinger:
Mostly I think HUGE changes like this probably need to be announced / discussed in the meeting.

benjaminhuth:
Hmm but it is only huge in terms of linediff, but should be solvable by a single global find-replace operation... I believe most interface changes are more annoying to clients...
But sure, should be announced.

paulgessinger:
I mean "huge" as in it breaks for almost everyone who's ever used ACTS.

andiwand:
should we revert this then? better sooner than later as this will conflict a lot of branches

benjaminhuth:
So my peronal opinion is not to revert it. Such changes are never painless, and I think its a good change since it expresses clearer what the type is

PR #2392, opened 2023-08-23
feat: Exa.TrkX abstract pipeline & metric hook

benjaminhuth:
This Does the following:
* Abstracts the Exa.TrkX pipeline from the examples algorithm to the plugin. Exposes the pipeline to python, so it is possible to run and test the pipeline without the examples framework (e.g., directly with `*.pyg` or `*.csv` files):

```python
spacepoints = list(np.random.uniform((100,3)).flatten())
spacepointIDs = list(np.arange(100))

pipeline = acts.examples.Pipeline(emb, [flt,gnn], trk, acts.logging.VERBOSE)
pipeline.run(spacepoints, spacepointIDs)
```

* Adds a interface for a pipeline hook, that is invoked after each stage with `nodes` and `edges` as arguments.
* Adds a implementation of the hook that prints the edge-based metrics of the graph with respect to the truth graph
* Extends the examples algorithm so it can generate the truth graph and a metric-hook:

```
08:52:52    EdgeClassifi   VERBOSE   Memory (used / total) [in MB]: 22155.4 / 40377.2
08:52:52    MetricsHook    INFO      Metrics for total graph:
08:52:52    MetricsHook    INFO      Efficiency=0.579324, purity=0.319214
08:52:52    MetricsHook    INFO      Metrics for target graph (pT > 0.5 GeV, nHits >= 3):
08:52:52    MetricsHook    INFO      Efficiency=0.968275, purity=0.138659
08:52:52    EdgeClassifi   DEBUG     Start edge classification
```


benjaminhuth:
@xju2 This implements part of the things we discussed some time ago at CERN.

Plugins/ExaTrkX/include/Acts/Plugins/ExaTrkX/Pipeline.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <vector>
+
+#include <boost/multi_array.hpp>
+
+namespace Acts {
+
+class PipelineHook {
+ public:
+  virtual ~PipelineHook(){};
+  virtual void operator()(const std::any &, const std::any &) const {};
```
paulgessinger, line 26: I'm assuming these two arguments are always `nodes` and `edges`? Can you add thos names here?
Also this is not a pure-virtual class, is this on purpose? Otherwise consider making both of these pure-virtual.

Plugins/ExaTrkX/src/Pipeline.cpp
```diff
@@ -0,0 +1,58 @@
...
+  std::any edge_weights;
+
+  for (auto edgeClassifier : m_edgeClassifiers) {
+    auto [newNodes, newEdges, newWeights] =
+        (*edgeClassifier)(std::move(nodes), std::move(edges));
+    nodes = std::move(newNodes);
+    edges = std::move(newEdges);
+    edge_weights = std::move(newWeights);
+
+    hook(nodes, edges);
```
paulgessinger, line 51: The hook implementation needs to handle being called multiple times, once for the input graph and then for each classifier?

paulgessinger (commented):
Here's a few comments on the naming and the interface specification for these pipeline hooks.

Just to make sure the idea: the non-examples-FW python bindings are still delivered via the examples build. But the idea is that you can launch this without running the examples sequence by calling this pipeline infrastructure directly?

Plugins/ExaTrkX/include/Acts/Plugins/ExaTrkX/Pipeline.hpp
benjaminhuth: yeah, the non-purely-`virtual` method is on purpose to allow defaulting the argument in `Pipeline::run` like this: 
```c++
std::vector<std::vector<int>> run(std::vector<float> &features,
                                  std::vector<int> &spacepointIDs,
                                  const PipelineHook &hook = {}) const;
``` 
Other possibility would be to pass as pointer and check for nullptr... then it could be purely virtual...
Would you prefer such a solution?

Plugins/ExaTrkX/include/Acts/Plugins/ExaTrkX/Pipeline.hpp
```diff
@@ -0,0 +1,50 @@
...
+};
+
+class Pipeline {
```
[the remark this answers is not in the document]
benjaminhuth: Good point

Plugins/ExaTrkX/src/Pipeline.cpp
paulgessinger: No problem per so, but the assumption is that the hook doesn't (need to) differentiate whether it's called the first time or any subsequent time? I.e. there's no need for special processing for the graph-construction-only call?

PR #2399, opened 2023-08-25
feat: update ODD version

asalzburger:
This updates to the new ODD version, which will change the physics output ... we can use this PR to discuss if the changes are reasonable and eventually update the references then.


asalzburger:
Ok, differences are really marginal, the slight changes in the DD4hep description make already slightly different truth particles.

paulgessinger:
Can we hold this until Monday?

asalzburger:
I've updated the material maps file - which now should include the beam pipe

asalzburger:
Hey @paulgessinger  - the truth tracking shows a strong improvement in the `d0` pulls and also some improvement in the resolution, so - indeed - the beam pipe translation was missing and adding it brings a good next step to pulls -> 1

https://herald.harbinger.gessinger.dev/view/acts-project/acts/893582628/truth_tracking.html#pullwidth_d0_vs_eta

asalzburger:
Mmmmm, it looks like it was the omission of the time component that caused the improvement

paulgessinger:
Hm, yeah the time breaks completely. Do you understand what that's coming from?

andiwand:
I disabled time measurements on ODD master some time ago because this fixed a bunch of issues downstream. I think this can be reverted now since we fixed a bunch of related bugs

asalzburger:
> I disabled time measurements on ODD master some time ago because this fixed a bunch of issues downstream. I think this can be reverted now since we fixed a bunch of related bugs

Yes, but `acts` didn't use the head of ODD/main - I think we should still investigate a bit.

PR #2404, opened 2023-08-28
feat: allow gdml for material recording

asalzburger:
This PR adds arguments parser to the `material_recording` and allows thus to run from GDML file as well.


Corentin-Allaire (approved):
Looks good thanks for adding this.

PR #2405, opened 2023-08-28
feat: TrackSelector supports eta-dependent cuts

paulgessinger:
This changes `TrackSelector` so that eta dependent cuts can be supplied. To this end, a new nested struct `CutConfig` is introduced. The existing `Config` struct is auto-convertible from that struct, and will produce a single cut set for all eta.

You can also supply `Config` with a vector of absolute eta bin edges and correspondingly sized (`N-1`) vector of `CutConfig` structs. The selector will then look up the associated cut value.

Note: the `CutConfig` properties `etaMin`, `etaMax`, `absEtaMin` and `absEtaMax` are only supported if there is a single cut configuration with an abs eta bin from `0.0` to `+infinity`. 

The `TrackSelector` constructor checks consistency of the configuration.


asalzburger (approved):
We initially discussed using the Range1D class, but this is not necessary here it seems.

PR #2407, opened 2023-08-29
fix: Stringop overread warning in GCC12

paulgessinger:
This showed up in our GCC12 based LCG builds, and only in Release mode. I don't think there's a way for us to fix this because the warning actually comes from `std::move`'ing `boost::small_vectors`. I'm suppressing the warnings here and remove the job from being allowed to fail.

PR #2410, opened 2023-08-30
ci: Update LCG tests

paulgessinger:
Removes the forces assertions from the LCG builds, because I've observed these to fail in clang15 on LCG 103.
I don't think we necessarily need to have these turned on, and we know this is a hack anyway.

This PR then also includes builds for LCG104, and switches the LCG102 build over to LCG102b_ATLAS_23 (which is what Athena currently builds on)

Blocked by:
- #2407


paulgessinger:
Ok:
- clang12 is broken as expected
- Seems I missed a problem in gcc12/13
- gcc11 on LCG_102b_ATLAS_23 has a git clone issue of all things?
To be investigated.

Examples/Io/Root/src/RootTrajectoryStatesWriter.cpp
```diff
@@ -492,7 +492,10 @@ ActsExamples::ProcessCode ActsExamples::RootTrajectoryStatesWriter::writeT(
               auto H = state.effectiveProjector();
               auto V = state.effectiveCalibratedCovariance();
               auto resCov = V + H * covariance * H.transpose();
-              auto res = state.effectiveCalibrated() - H * parameters;
+              Acts::ActsDynamicVector res(state.calibratedSize());
+              res.setZero();
+
+              res = state.effectiveCalibrated() - H * parameters;
```
andiwand, line 498: eigen :smiling_face_with_three_hearts:

andiwand (approved):

PR #2416, opened 2023-09-04
refactor: Store phi inside space point instead of computing atan2 

CarloVarni:
In addition to what the title says I've also made few changes around. For instance avoiding passing primitive types by reference when not necessary


CarloVarni:
@paulgessinger I think this can be reviewed

paulgessinger (approved):

PR #2418, opened 2023-09-05
feat: Add (optional) forward linking of track states

paulgessinger:
A track can now initiate forward linking of its track states. The track
states have a new property, `next` that optionally makes the sequence a
doubly linked list. This is not filled by default (not even by
`appendTrackState`) because for branching cases, it's not unambiguous.

`linkForward()` on a track will initiate the forward linking and store
the innermost track state index for easy access. `reverseTrackStates`
also does this, since it's only valid for non-branching trackstate
sequences anyway.

Also adds a `innermostTrackState()` getter, which returns a track state proxy in case the track is forward linked and has a `stemIndex`

Blocked by:
- #2425 
- #2426


asalzburger:
Very nice, indeed! @EdwardMoyse you mind reviewing this one?

paulgessinger:
Rebased, conflicts resolved.

paulgessinger:
Conflicts resolved.

andiwand (approved):
lgtm :+1:

PR #2425, opened 2023-09-06
refactor: Use separate column for previous in MTJ

paulgessinger:
This was previously still using the `iprevious` member in `IndexData`, but we did have a separate `m_previous` column already that was being ignored.

Preparation for #2418


felix-russo (approved):
It is strange that the m_previous member was not used previously

paulgessinger:
I just messed that up yeah 🙄

felix-russo:
Happens to the best :)

PR #2426, opened 2023-09-06
refactor!: Rename trackStates to trackStatesReverse

paulgessinger:
This better indicates that the iterations is outside in

Preparation for #2418.


felix-russo:
So if I understand this correctly, a trajectory is a set of points in space (?), and when we call `trackStateRange` we get a set of points ranging from `iendpoint` to the starting point of the trajectory?

paulgessinger:
@felix-russo It's a sequence of track states, but in the CKF case, they can branch out, i.e. form a tree instead of a linear sequence.

When appending track states, they are linked from the end (outermost, or last) track state to the innermost one. This is the direction that the range allows you to iterate over, but from the name it was previously not clear.

asalzburger (approved):
Clean replace, which. makes the code easier to read, because it clarifies in which direction e.g. the surfaces in the refit mode are collected.

asalzburger:
@paulgessinger - this is breaking, do you want to go ahead and bring it in?

paulgessinger:
This should go into v30, so we can merge it next week.

acts-project-service:
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/32490335) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6203664)!

PR #2427, opened 2023-09-06
fix: Make eta-dependent TrackSelector backwards compatible

paulgessinger:
I noticed that the eta-dependent cuts introduced in #2405 are a not actually backwards compatible. This should hoopefully fix that.

### DIFF
--- a/Core/include/Acts/TrackFinding/TrackSelector.hpp
+++ b/Core/include/Acts/TrackFinding/TrackSelector.hpp
@@ -25,7 +25,7 @@ class TrackSelector {
  public:
   /// Configuration of a set of cuts for a single eta bin
   /// Default construction yields a set of cuts that accepts everything.
-  struct CutConfig {
+  struct Config {
     // Minimum/maximum local positions.
     double loc0Min = -inf;
     double loc0Max = inf;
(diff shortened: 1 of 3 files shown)


paulgessinger:
Ok, I confirmed locally that this compiles on Athena main without code changes (which was the goal).

Core/include/Acts/TrackFinding/TrackSelector.hpp
```diff
@@ -348,7 +353,7 @@ bool TrackSelector::isValidTrack(const track_proxy_t& track) const {
     return false;
   }
 
-  const CutConfig& cuts = m_cfg.getCuts(eta);
+  const Config& cuts = m_cfg.getCuts(eta);
```
timadye, line 356: This seems to be a good API, so for the purposes of this PR, a good thing.

Would it be worth making a shortcut for the eta-independent case, that doesn't calculate or use `eta` until it is needed. This would be especially useful if there are no cuts on `eta` either, eg. for detectors where &eta; isn't a relevant parameter.

That would probably make the logic more complicated, so fine to postpone to another PR.

timadye (commented):
Looks good!

It would be nice to have an eta-dependent example in `full_chain_itk.py`. Is that easy to add here, or do later?

paulgessinger:
@timadye I changed this a bit so eta should only get calculated when actually needed. If possible, I'd like to get this in today, so would postpone changing the ITk example.

Core/include/Acts/TrackFinding/TrackSelector.hpp
```diff
@@ -340,38 +346,58 @@ bool TrackSelector::isValidTrack(const track_proxy_t& track) const {
...
   }
 
-  const CutConfig& cuts = m_cfg.getCuts(eta);
+  const Config& cuts = *cutsPtr;
 
   return within(track.transverseMomentum(), cuts.ptMin, cuts.ptMax) and
-         within(std::abs(eta), cuts.absEtaMin, cuts.absEtaMax) and
-         within(eta, cuts.etaMin, cuts.etaMax) and
+         within(absEta(), cuts.absEtaMin, cuts.absEtaMax) and
+         within(_eta, cuts.etaMin, cuts.etaMax) and
```
timadye, line 378: Doesn't this still calculate `_eta` when there is no eta cut? How about:
```suggestion
         (noEtaCuts ||
         within(absEta(), cuts.absEtaMin, cuts.absEtaMax) and
         within(_eta, cuts.etaMin, cuts.etaMax)) and
```
where `noEtaCuts` is fairly simply (but verbosely) calculated. This could also be stored in the `Config` for speed, or combined with `m_isUnbinned` by making that an `enum`.

It will still skip `absEta()` if it the pT cut fails. I guess you could quickly gain a bit more by moving this to the last check.

paulgessinger: Okay, I pushed another version where the `TrackSelector` constructor calculates `m_noEtaCuts` that is then used in the `isValidTrack` function.

acts-project-service:
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/32288355) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6164437)!

PR #2444, opened 2023-09-13
refactor: Optional curvilinear state in Propagator

paulgessinger:
So far, without a target surface, the propagator unconditionally produces curvilinear parameters, and allocates a plane surface for this. This adds an argument, which if set to false tells the Propagator not to do that. In particular, we don't use this curvilinear state at all in the CKF and KF (I'm not sure about the GSF).


benjaminhuth:
> In particular, we don't use this curvilinear state at all in the CKF and KF (I'm not sure about the GSF).

It isn't used in the GSF either. Rather, the requirement to provide this functionality is a bit annoying (but it should be there I guess)

PR #2454, opened 2023-09-15
feat: make stepping logger volume agnostic

asalzburger:
This PR removes the explicit setting of `TrackingVolume` in favour of the `GeometryIdentifier` (which is the only thing that is accessed from the `TrackingVolume` in the `Step` struct of the `SteppingLogger` and thus makes the step logging infrastructure accessible to the new `Acts::Detector` geometry as well.


paulgessinger:
Should this change the actual output of the logger? It seems like it does, but I don't think we expect that.

asalzburger:
> Should this change the actual output of the logger? It seems like it does, but I don't think we expect that.

It eventually could - the root-outout logger has had an internal logic when to use which volumeID, which for boundary surfaces can be ambiguous. Now, this is changed to a very simple logic:

- if a surface exists (i.e. the Logger logs an intersection with a surface) the surface purely determines the geoID
- if no surface exist, the geoID is given by the volume where the logging happened

asalzburger:
@paulgessinger can I download the artifacts of the failed job, so I can inspect if that's what is actually happening?

paulgessinger:
@asalzburger I'm afraid the file is not in the artifact. But if you confirm this change is intentional, I think it's fine to merge, I just want to avoid an "accidental" change we don't intend, that's all.

Examples/Io/Root/src/RootPropagationStepsWriter.cpp
```diff
@@ -127,37 +127,21 @@ ActsExamples::ProcessCode ActsExamples::RootPropagationStepsWriter::writeT(
...
-        approachID = geoID.approach();
-        sensitiveID = geoID.sensitive();
         if (step.surface->surfaceMaterial() != nullptr) {
           material = 1;
         }
       }
-      // a current volume overwrites the surface tagged one
-      if (step.volume != nullptr) {
-        volumeID = step.volume->geometryId().volume();
-      }
```
asalzburger, line 154: Hey @paulgessinger  - this is the changed logic, the step writer actually chose to overwrite the volume identification value from the surface, which is ok as well, it just didn't puts logic into a writer ... and if another writer does this differently it will never be caught.

asalzburger:
Why @andiwand - this should be relatively quickly. The reason why the stepping test fails is explained above ... some boundary surfaces are now differently flagged.

andiwand (approved):

PR #2455, opened 2023-09-15
feat: make material interaction volume agnostic

asalzburger:
This PR makes the (recording of) material interaction work with `Acts::TrackingVolume` and `Acts::DetectorVolume` by introducing a small helper switch class between the two.

If this concept is successful, this can be used also for the `MaterialMapping` to let the two geometries coexist and co-use the same tooling chain.


asalzburger:
Hey @Corentin-Allaire - as discussed, this is the first PR for the interaction recording, probably we can use the same strategy for the mapping. Would you agree to review this?

asalzburger:
Some problem with rebasing, closing and updating .. it

PR #2459, opened 2023-09-18
fix: Remove reserve from compatibleSurfaces

paulgessinger:
This would previously always heap allocate, since our small vector has local capacity 10 and we unconditionally ask for 20 entries. In ODD almost all cases are below 10, so I'm just removing the `reserve()` call.


paulgessinger:
Pre-fix: (heap allocated):
image


With fix (no heap allocation):
image

andiwand (approved):

PR #2463, opened 2023-09-19
fix: Navigator correction

dimitra97:
This PR just inserts a line` fillNavigationState` in the `prestep` Detector Navigator method
@andiwand


asalzburger:
Correct! We have seen that this is missing. Thanks - I will review and let you know.

asalzburger (approved):
Makes totally sense.

PR #2465, opened 2023-09-19
feat: volume agnostic material interaction

asalzburger:
This PR introduces an `InteractionVolum` helper class that allows to switch between `TrackingVolume` and `DetectorVolume` and through that makes the material interaction recording struct quasi volume agnostic.


Core/include/Acts/Material/MaterialInteraction.hpp
```diff
@@ -9,13 +9,52 @@
...
+  /// @param tv The tracking volume
+  InteractionVolume(const TrackingVolume* tv) : trackingVolume(tv) {}
+
+  /// Constructor from detector volume
+  /// @param dv The detector volume
+  InteractionVolume(const Experimental::DetectorVolume* dv)
+      : detectorVolume(dv) {}
+
+  /// Forward the geometry identifier
+  GeometryIdentifier geometryId() const {
```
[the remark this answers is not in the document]
asalzburger: No, that will never happen - you can not run in parallel geometries, as the Navigator only knows one.

Corentin-Allaire (approved):

acts-project-service:
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/32602853) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6225344)!

Core/include/Acts/Material/MaterialInteraction.hpp
```diff
@@ -9,13 +9,52 @@
...
-class TrackingVolume;
+
+/// @brief The Material interaction volume struct
+/// It acts as a switch between detctor and tracking volume
+/// as long as those co-exist alongside
+struct InteractionVolume {
+  /// The tracking volume
+  const TrackingVolume* trackingVolume = nullptr;
+  /// The detector volume
+  const Experimental::DetectorVolume* detectorVolume = nullptr;
```
paulgessinger, line 28: Could this have been an `std::variant`?

asalzburger: Yes and no (I tried this first), indeed we could have a variant here, but I also need the option that both are `nullptr`, so either it would be a variant with a `std::monostate` to catch the latter case, or a `std::optional<std::variant<...,...>>` which I found overly complicated.

paulgessinger: Ok. Just seemed like a very custom tagged union to me, but I won't insist on changing anything.

PR #2474, opened 2023-09-22
ci: Bump build procs in GitLab CI

paulgessinger:
We currently limit to `-j3`, presumably because we were running into memory consumption issues.

I'm playing with the sizing of the GitLab CI build VMs at the moment, and I'd like to see if we can go to the full core count at this point.


andiwand (approved):

paulgessinger:
Milestone bot seems to have issues (github token rate limit?)

PR #2476, opened 2023-09-22
feat: Write measurement indices in `RootTrajectorySummaryWriter`

andiwand:
This allows accessing the measurement indices of tracks inside a `tracksummary.root`. That information can then be used to aggregate efficiencies offline and outside of ACTS.

### DIFF
--- a/Examples/Io/Root/include/ActsExamples/Io/Root/RootTrajectorySummaryWriter.hpp
+++ b/Examples/Io/Root/include/ActsExamples/Io/Root/RootTrajectorySummaryWriter.hpp
@@ -116,6 +116,8 @@ class RootTrajectorySummaryWriter final
   std::vector<unsigned int>
       m_NDF;  ///< The number of ndf of the measurements+outliers
   std::vector<std::vector<double>>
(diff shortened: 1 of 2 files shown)


Examples/Io/Root/include/ActsExamples/Io/Root/RootTrajectorySummaryWriter.hpp
```diff
@@ -115,6 +115,8 @@ class RootTrajectorySummaryWriter final
   std::vector<float> m_chi2Sum;               ///< The total chi2
   std::vector<unsigned int>
       m_NDF;  ///< The number of ndf of the measurements+outliers
+  std::vector<std::vector<double>>
```
paulgessinger, line 118: double?

andiwand:
what I want to do is actually much easier with `nMajorityHits`

beomki-yeo:
I know the PR is closed but a question: the measurement index here is equivalent to what is in `event0000000XX-measurement-simhit-map.csv` or is it totally different thing?

PR #2478, opened 2023-09-22
fix: Fix `RootSimHitReader` for empty events

andiwand:
Looking at the writer this can happen for non recorded hits so the reader should also support that.

I discovered this while simulating single particles with Geant4 in our Examples framework


benjaminhuth:
Hmm I'm not totally convinced on this approach yet...
a) We cannot respect such empty events as part of the determination of the event range, so if the empty event is at the borders of the event range, it would be ignored as well
b) In general this would consider every not found event id as empty event... not sure if this is what we want...

IMO we should switch the root simhit IO to one tree entry per event. Then we could also remove this map stuff...

benjaminhuth:
Yeah, file IO breaking changes are not nice... Should we let the writer warn if the first or the last event is empty?

benjaminhuth (approved):

paulgessinger:
There seems to be one physmon failure. Oddly it only fails in the GitHub actions job, not the one on GitLab...

andiwand:
hmm that has to be a random error because I don't think we use the reader in the physmon chains

I will poke the CI and see if it helps

acts-project-service:
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/32723565) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6249965)!

PR #2480, opened 2023-09-25
docs: introducing acts/notes

felix-russo:
Proposal to introduce a folder, where one can upload:

- analytical calculations
- plots with explanations
- basically everything that helps understanding acts code

The idea is to link the files with the code via comments. An example is provided in this PR.


paulgessinger:
Hmm, can't we do this in Markdown, and host it on the documentation? That way it's searchable and we can link to it from other parts of the documentation.

benjaminhuth:
Good idea indeed!

My first thought would be that maybe `docs/notes` would be more appropriate.

We could also add a top level site to read-the-docs named "Additional material" or so and link the files there for easy access

felix-russo:
We will find a different approach that does not involve uploading pdfs to the acts repo!

PR #2482, opened 2023-09-26
feat: dd4hep and odd detector

asalzburger:
This PR adds the necessary new code to build a first version of the ODD detector in the new detector structure.

It also adds UnitTests for DD4hep.


asalzburger:
I do not understand the difference in performance though - I should not have touched any used functionality, will split the PR up and do step by step to catch this.

asalzburger:
Close, Superseded.

PR #2483, opened 2023-09-26
refactor: Updators fix

dimitra97:
This PR introduces some changes in the surface candidates updators. 
I moved some functions from the `SurfaceCandidatesUpdator.hpp` to the `NavigationStateUpdators.hpp` so they can be visible to the updators not defined in the `SurfaceCandidatesUpdators.hpp`
The flag `sort` indicates whether sorting is needed or not. A case where the individual updators should not do sorting is when they are passed as updators to the `Chained Updator`, where the sorting needs to be done in the end in the `Chained Updator`. 
I think doing twice the sorting when using the` Chained Updator` takes a lot of time.


andiwand:
With https://github.com/acts-project/acts/pull/2316 I would move the sorting to the navigator so the job of the delegate is just to provide the candidates. But I haven't followed up on this effort in a while so happy to merge your fix in first.

dimitra97:
> With #2316 I would move the sorting to the navigator so the job of the delegate is just to provide the candidates. But I haven't followed up on this effort in a while so happy to merge your fix in first.

I think then the sort flag is not needed?

andiwand:
yes exactly I think this would simplify things

andiwand (dismissed):
nicely done!

I left one suggestion

but happy to merge as is

andiwand (approved):

asalzburger (approved):
Great!

asalzburger:
doc build timed out again, but this I will overrule if that's the only check that fails.

PR #2485, opened 2023-09-27
refactor: rename local variables and improve comments for IVF

felix-russo:
- Rename some variables 
- Improve some comments
- One little change: Previously `hypotVariance` was set to 1 if it was 0, which seems like a random quick fix. I think it is more correct to skip the track in this case because the expression in line 408 would diverge and the if-condition should evaluate to `false`.
image


paulgessinger (approved):
Looks good to me. We should keep an eye out to see if this changes Athena outputs.

acts-project-service:
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/32816335) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6267660)!

PR #2486, opened 2023-09-27
feat: Make CKF branching configurable from python

andiwand:
Currently we hardcode the branching config in python which makes it hard to find the default and unchangeable for the user. With this PR I add a struct `TrackFindingConfig` which defines the branching params and lower the defaults in hope to spare some resources with ttbar events.


Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -119,6 +119,12 @@
     defaults=[(None, None)] * 7 + [None],
 )
 
+TrackFindingConfig = namedtuple(
```
benjaminhuth, line 122: Maybe we can find a name that is a bit less generic, like `CkfConfig` or so?

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -119,6 +119,12 @@
...
+    "TrackFindingConfig",
+    ["chi2CutOff", "numMeasurementsCutOff"],
+    defaults=[50.0, 10],
```
andiwand, line 125:
```suggestion
    defaults=[15.0, 10],
```

benjaminhuth (approved):
Looks good!

PR #2489, opened 2023-09-28
perf: do not sort track state candidates with chi2 above cut-off.

goetzgaycken:
When selecting measurements for the next ckf step, only sort
the candidate track states with chi2 below the cut-off by chi2.

This is a minor improvement which reduces the track finding time
of the ckf in ttbar PU200 with the ITK detector by 1-2%. It does not
have any impact on the tracks found by the ckf.

### DIFF
--- a/Core/include/Acts/TrackFinding/MeasurementSelector.hpp
+++ b/Core/include/Acts/TrackFinding/MeasurementSelector.hpp
@@ -24,6 +24,7 @@
 #include <limits>
 #include <utility>
 #include <vector>
+#include <cassert>
 
 namespace Acts {
 
@@ -106,60 +107,81 @@ class MeasurementSelector {
       return CombinatorialKalmanFilterError::MeasurementSelectionFailed;
     }
 
+    assert( !cuts->chi2CutOff.empty() );
+    const auto& chi2CutOff = cuts->chi2CutOff;
+    auto maxChi2Cut  = *std::max_element(chi2CutOff.begin(), chi2CutOff.end());
     double minChi2 = std::numeric_limits<double>::max();
     size_t minIndex = 0;
-    size_t index = 0;
-    // Loop over all measurements to select the compatible measurements
-    for (auto& trackState : candidates) {
-      // Take the parameter covariance
-      // const auto predicted = tackState.predicted();
-      // const auto predictedCovariance = trackState.predictedCovariance();
-
-      double chi2 = calculateChi2(
-          // This abuses an incorrectly sized vector / matrix to access the
-          // data pointer! This works (don't use the matrix as is!), but be
-          // careful!
-          trackState
-              .template calibrated<MultiTrajectoryTraits::MeasurementSizeMax>()
-              .data(),
-          trackState
-              .template calibratedCovariance<
-                  MultiTrajectoryTraits::MeasurementSizeMax>()
-              .data(),
-          trackState.predicted(), trackState.predictedCovariance(),
-          trackState.projector(), trackState.calibratedSize());
-
-      trackState.chi2() = chi2;
-
-      // Search for the measurement with the min chi2
-      if (chi2 < minChi2) {
-        minChi2 = chi2;
-        minIndex = index;
-      }
-
-      index++;
-    }
-
-    const auto& chi2CutOff = cuts->chi2CutOff;
-
+    auto trackStateIterEnd = candidates.end();
     {
-      // If there is no selected measurement, return the measurement with the
-      // best chi2 and tag it as an outlier
-      const auto bestIt = std::next(candidates.begin(), minIndex);
-      const auto chi2 = bestIt->chi2();
-      const auto chi2Cut =
-          VariableCut<traj_t>(*bestIt, cuts, chi2CutOff, logger);
-      ACTS_VERBOSE("Chi2: " << chi2 << ", max: " << chi2Cut);
-      if (chi2 >= chi2Cut) {
-        ACTS_VERBOSE(
-            "No measurement candidate. Return an outlier measurement.");
-        isOutlier = true;
-        // return single item range, no sorting necessary
-        return Result::success(std::pair{bestIt, std::next(bestIt, 1)});
-      }
+       auto trackStateIter = candidates.begin();
+       // Loop over all measurements to select the compatible measurements
+       // Sort track states which do not satisfy the chi2 cut to the end.
+       // When done trackStateIterEnd will point to the first element that
+       // does not satisfy the chi2 cut.
+       assert(trackStateIter != trackStateIterEnd);
+       for (;;) {
+          // Take the parameter covariance
+          // const auto predicted = tackState.predicted();
+          // const auto predictedCovariance = trackState.predictedCovariance();
+
+          double chi2 = calculateChi2(
+            // This abuses an incorrectly sized vector / matrix to access the
+            // data pointer! This works (don't use the matrix as is!), but be
+            // careful!
+            trackStateIter
+                ->template calibrated<MultiTrajectoryTraits::MeasurementSizeMax>()
+                .data(),
+            trackStateIter
+                ->template calibratedCovariance<
+                    MultiTrajectoryTraits::MeasurementSizeMax>()
+                .data(),
+            trackStateIter->predicted(), trackStateIter->predictedCovariance(),
+            trackStateIter->projector(), trackStateIter->calibratedSize());
+
+          trackStateIter->chi2() = chi2;
+
+          // only consider track states which pass the chi2 cut
+          if (chi2 >= maxChi2Cut || chi2 >= VariableCut<traj_t>(*trackStateIter, cuts, chi2CutOff, logger) ) {
+             --trackStateIterEnd;
+             // still check whether the element has the smallest chi2 in case an outlier is returned.
+             if (chi2<minChi2) {
+                minChi2 = chi2;
+                // the current element will be swapped with the last unchecked element if they are
+                // different
+                minIndex = std::distance(candidates.begin(),trackStateIterEnd);
+             }
+
+             if (trackStateIter == trackStateIterEnd) {
+                break;
+             }
+             else {
+                // swap rejected element with last element in list
+                std::swap(*trackStateIter,*trackStateIterEnd);
+             }
+          }
+          else {
+             // Search for the measurement with the min chi2
+             // @if there is a track state which passes the cut-off there is
+             // no need to remember the track state with the smallest chi2.
+             ++trackStateIter;
+             if (trackStateIter == trackStateIterEnd) break;
+          }
+       }
     }
 
-    std::sort(candidates.begin(), candidates.end(),
+
+    // If there are no measurements below the chi2 cut off, return the measurement with the
+    // best chi2 and tag it as an outlier
+    if (candidates.begin() == trackStateIterEnd) {
+      const auto bestIt = std::next(candidates.begin(), minIndex);
+      ACTS_VERBOSE("No measurement candidate. Return an outlier measurement chi2=" << bestIt->chi2());
+      isOutlier = true;
+      // return single item range, no sorting necessary
+      return Result::success(std::pair{bestIt, std::next(bestIt, 1)});
+    }
+
+    std::sort(candidates.begin(), trackStateIterEnd,
               [](const auto& tsa, const auto& tsb) {
                 return tsa.chi2() < tsb.chi2();
               });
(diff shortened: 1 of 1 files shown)


Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -106,60 +107,81 @@ class MeasurementSelector {
...
+                ->template calibratedCovariance<
+                    MultiTrajectoryTraits::MeasurementSizeMax>()
+                .data(),
+            trackStateIter->predicted(), trackStateIter->predictedCovariance(),
+            trackStateIter->projector(), trackStateIter->calibratedSize());
+
+          trackStateIter->chi2() = chi2;
+
+          // only consider track states which pass the chi2 cut
+          if (chi2 >= maxChi2Cut || chi2 >= VariableCut<traj_t>(*trackStateIter, cuts, chi2CutOff, logger) ) {
```
paulgessinger, line 145: `VariableCut<traj_t>(*trackStateIter, cuts, chi2CutOff, logger)` could probably be constructed outside of the loop.

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -106,60 +107,81 @@ class MeasurementSelector {
...
+             --trackStateIterEnd;
+             // still check whether the element has the smallest chi2 in case an outlier is returned.
+             if (chi2<minChi2) {
+                minChi2 = chi2;
+                // the current element will be swapped with the last unchecked element if they are
+                // different
+                minIndex = trackStateIterEnd-candidates.begin();
```
paulgessinger, line 152: I didn't know that `vector` iterators support this. Maybe still better to use `std::distance`?

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -106,60 +107,81 @@ class MeasurementSelector {
...
-      index++;
+    auto trackStateIterEnd = candidates.end();
+    {
+       auto trackStateIter = candidates.begin();
+       // Loop over all measurements to select the compatible measurements
+       // Sort track states which do not satisfy the chi2 cut to the end.
+       // When done trackStateIterEnd will point to the first element that
+       // does not satisfy the chi2 cut.
+       assert(trackStateIter != trackStateIterEnd);
+       for (;;) {
```
paulgessinger, line 123: I suppose this could still be
```suggestion
       for (trackStateIter = candidates.begin();trackStateIter != trackStateIterEnd;) {
```
? It's still potentially an infinite loop (if the loop body is bugged) but a bit more explicit up top?

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -106,60 +107,81 @@ class MeasurementSelector {
...
+                // different
+                minIndex = trackStateIterEnd-candidates.begin();
+             }
+
+             if (trackStateIter == trackStateIterEnd) {
+                break;
+             }
+             else {
+                // swap rejected element with last element in list
+                std::swap(*trackStateIter,*trackStateIterEnd);
```
paulgessinger, line 160: Ok and just for my understanding, you pull in the `End` here above, so that's how you accumulate rejected elements at the end, right?

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -168,30 +190,16 @@ class MeasurementSelector {
     const auto numMeasurementsCut = VariableCut<traj_t>(
         *candidates.begin(), cuts, cuts->numMeasurementsCutOff, logger);
 
-    auto endIterator = candidates.begin();
-    auto maxIterator = candidates.end();
-    if (candidates.size() > numMeasurementsCut && numMeasurementsCut > 0) {
-      maxIterator = std::next(candidates.begin(), numMeasurementsCut);
+    if ( static_cast<std::size_t>(std::distance(candidates.begin(),trackStateIterEnd)) > numMeasurementsCut && numMeasurementsCut > 0) {
+       trackStateIterEnd = std::next(candidates.begin(), numMeasurementsCut);
```
paulgessinger, line 194: Minor point and irrelevant for `std::vector`: if `numMeasurementsCut > 0` comes first, the call to `std::distance` can be avoided if that's not the case.

Actually: if `numMeasurementsCut` is 0 we can probably early return from the function anyway?

paulgessinger (commented):
Very nice!
Can you revert the changes to `.zenodo.json`, `CITATION.cff` and `version_number`? Those are only modified in the releases branch.

andiwand (commented):
the commit history looks a bit busy. may be worth to rebase the changes to the main branch and force push

paulgessinger:
@andiwand It's just based off of the tag. If the three files are reverted and we squash-and-merge, the other commits should go away anyway.

goetzgaycken:
> Very nice! Can you revert the changes to `.zenodo.json`, `CITATION.cff` and `version_number`? Those are only modified in the releases branch.

Strange the PR should only modify  Core/include/Acts/TrackFinding/MeasurementSelector.hpp. Maybe the problem is that it is based on v29.2. Hopefully a rebase will fix that

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
goetzgaycken: With future changes the loop could be turned into an infinite loop, In the current state either the front iterator or the end iterator are advanced and in each loop iteration the distance shrinks. both branches will break the loop if the distance is zero. Before the first iteration the loop cannot be empty because the entry requires that candiates is not empty. Not that it matters much wrt. the other computations but it would still add and extra comparison which was already done.
I could change it to:
```
for(  auto trackStateIter = candidates.begin(); ;) {
     assert(trackStateIter != trackStateIterEnd);
}
```
to make the intention clear.

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
goetzgaycken: Yes, all elements with too large chi2 will be moved to the end. This partitions the vector into a part with elements that pass the chi2 cut and the upper part with elements which do not pass the chi2 cut. In principle one could, I guess, just ignore the elements which do not pass the chi2 cut with the exception of the element which has the minimum chi2. So a simple one way copy would be good enough. But I suppose that would make it a bit more complicated.

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
goetzgaycken: If distance was an expensive operation this would be worthwhile, but likely distance is cheaper than the comparison. 
This is mustly just:
`return std::next(candidates.begin(), min( numMeasurementsCut,s_c<>( std::distance(candidates.begin(),trackStateIterEnd))) );`
one cannot return much earlier.

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
goetzgaycken: This depends on *trackStateIter which chages with each iteration.

goetzgaycken:
update from e672b18cc83c655dea48ee30a0bf57b110883f8c to e5d3d602f6a09c19c91b08e96c3f1d083e7a85f3
- reapplied changes to main
- use distance instead of iterator arithmetic for minIndex

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
paulgessinger: `std::distance` is O(1) for `vector`, but not for other containers. But yeah, in this case it does not matter, really.

paulgessinger:
`clang-format` still complains about the `MeasurementSelector.hpp` formatting. Can you fix that?

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -106,60 +107,85 @@ class MeasurementSelector {
       return CombinatorialKalmanFilterError::MeasurementSelectionFailed;
     }
 
+    assert(!cuts->chi2CutOff.empty());
+    const auto& chi2CutOff = cuts->chi2CutOff;
+    auto maxChi2Cut = *std::max_element(chi2CutOff.begin(), chi2CutOff.end());
```
andiwand, line 112: I wonder if we should add a global maximum to the config. that way we would not need to look for the maximum all the time

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -106,60 +107,85 @@ class MeasurementSelector {
...
+      // Loop over all measurements to select the compatible measurements
+      // Sort track states which do not satisfy the chi2 cut to the end.
+      // When done trackStateIterEnd will point to the first element that
+      // does not satisfy the chi2 cut.
+      assert(trackStateIter != trackStateIterEnd);
+      for (;;) {
+        // Take the parameter covariance
+        // const auto predicted = tackState.predicted();
+        // const auto predictedCovariance = trackState.predictedCovariance();
+
```
andiwand, line 127: dead code?
```suggestion
```

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -106,60 +107,85 @@ class MeasurementSelector {
...
-        minChi2 = chi2;
-        minIndex = index;
+    auto trackStateIterEnd = candidates.end();
+    {
+      auto trackStateIter = candidates.begin();
+      // Loop over all measurements to select the compatible measurements
+      // Sort track states which do not satisfy the chi2 cut to the end.
+      // When done trackStateIterEnd will point to the first element that
+      // does not satisfy the chi2 cut.
+      assert(trackStateIter != trackStateIterEnd);
```
andiwand, line 122: this will always be `false` no? since we check `candidates.empty()` above
```suggestion
```

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -106,60 +107,85 @@ class MeasurementSelector {
...
+            // swap rejected element with last element in list
+            std::swap(*trackStateIter, *trackStateIterEnd);
+          }
+        } else {
+          // Search for the measurement with the min chi2
+          // @if there is a track state which passes the cut-off there is
+          // no need to remember the track state with the smallest chi2.
+          ++trackStateIter;
+          if (trackStateIter == trackStateIterEnd)
+            break;
```
andiwand, line 171:
```suggestion
          if (trackStateIter == trackStateIterEnd) {
            break;
          }
```

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -106,60 +107,85 @@ class MeasurementSelector {
...
-          trackState.predicted(), trackState.predictedCovariance(),
-          trackState.projector(), trackState.calibratedSize());
-
-      trackState.chi2() = chi2;
-
-      // Search for the measurement with the min chi2
-      if (chi2 < minChi2) {
-        minChi2 = chi2;
-        minIndex = index;
+    auto trackStateIterEnd = candidates.end();
```
andiwand, line 115: @paulgessinger do the track states share the predicted parameters and covariances?

I would have expected that we only get one prediction and all the measurements and then chose on the measurements

paulgessinger: I believe the track states share the predicted parameters and covariances, yes.

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
goetzgaycken: would make sense to me.

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
goetzgaycken: But old out-commented code for @paulgessinger  to decide.

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
goetzgaycken: - Yes unless the code is altered. It is just there to state the intention.

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
paulgessinger: I think this is a leftover from when I factorized out the chi2 calculation. I think this can go.

goetzgaycken:
> `clang-format` still complains about the `MeasurementSelector.hpp` formatting. Can you fix that?

ok

paulgessinger:
@goetzgaycken I pushed a commit with fixed formatting earlier, sorry.

andiwand (approved):

PR #2500, opened 2023-10-03
feat: Python bindings for Geant4 manager and handle

andiwand:
This enables us to inspect the current handle of Geant4 and therefore adjust the log level from python which is useful while debugging.


Examples/Python/src/Geant4Component.cpp
```diff
@@ -69,7 +69,14 @@ PYBIND11_MODULE(ActsPythonBindingsGeant4, mod) {
...
 
-  py::class_<Geant4Handle, std::shared_ptr<Geant4Handle>>(mod, "Geant4Handle");
+  py::class_<Geant4Manager, std::unique_ptr<Geant4Manager, py::nodelete>>(
+      mod, "Geant4Manager")
+      .def_static("instance", &Geant4Manager::instance,
+                  py::return_value_policy::reference)
+      .def("currentHandle", &Geant4Manager::currentHandle);
+
+  py::class_<Geant4Handle, std::shared_ptr<Geant4Handle>>(mod, "Geant4Handle")
+      .def("tweekLogging", &Geant4Handle::tweekLogging);
```
felix-russo, line 79: Why was this line not needed before?

felix-russo (approved):
Looks good, just one question for my personal understanding

Examples/Python/src/Geant4Component.cpp
andiwand: It was not accessed through Python yet which I did in another repository

Examples/Python/src/Geant4Component.cpp
```diff
@@ -69,7 +69,14 @@ PYBIND11_MODULE(ActsPythonBindingsGeant4, mod) {
              std::shared_ptr<DetectorConstructionFactory>>(
       mod, "DetectorConstructionFactory");
 
-  py::class_<Geant4Handle, std::shared_ptr<Geant4Handle>>(mod, "Geant4Handle");
+  py::class_<Geant4Manager, std::unique_ptr<Geant4Manager, py::nodelete>>(
```
paulgessinger, line 72: Unique nodelete means that Python will not own this object. Does G4 own it?

Examples/Python/src/Geant4Component.cpp
paulgessinger, line 79: Tweak?

Examples/Python/src/Geant4Component.cpp
andiwand: `Geant4Manager` is a singleton and has a private destructor so python should really not own it. without this it also complains about not being able to correctly deconstruct it

paulgessinger: Huh, I think I didn't appreciate that this was an option. I worked around this using factory functions in some cases, but this might be a bit nicer.

paulgessinger (approved):

PR #2492, opened 2023-09-29
fix: Fix Geant4 initial particle translation

andiwand:
Observed problem: Particles put into G4 will not have the same properties after `BeanOn` is called. We relied on this in order to translate them back into Acts.

The following image shows how pt=1GeV looked like in `particles_initial` after a G4 forth and back.

image

Solution: Put particles into the initial collection on translation and adopt generic translation code so it will accept already translated particles.

PR #2502, opened 2023-10-03
fix: root hashes after tgeo thickness gap bug

asalzburger:
This updates the root hashes after the TGeo bugfix.

The fix in thickness calculation changes slightly the layer building.


andiwand (approved):

andiwand:
this explains the failure in https://github.com/acts-project/acts/pull/2492

PR #2509, opened 2023-10-04
refactor!: move AMVF and IVF config explicitly

felix-russo:
Title says it all!

Might be breaking as now `config` must be moved in the parent scope.

It might be better to not move the `config` and pass it as a const reference, not sure though.


paulgessinger:
I don't think it has performance implications. The move most likely is going to result in a regular copy anyway, but the compiler might end up optimizing this out the same way it does for the const ref. We also don't (or shouldn't) instantiate this very often (I think we do at this point, but only for convenience reasons) so I don't think it's a critical optimization.

felix-russo:
Should I just use a const ref then?

acts-project-service:
## :red_circle: Athena integration test results [be529c8cb083bf33ac822c787501b10965639bd4]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/33710174) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6440201)!

paulgessinger (approved):
Anyway, implicitly moving mutable references is not a good idea anyway, so let's do this.

paulgessinger:
This required minor compile fixes in Athena.

felix-russo:
I realized that we pass `config` structs almost always as `const ref`.

Should we do this here too? Or should we also use `std::move` for the other `configs`?

paulgessinger:
@felix-russo It doesn't really matter. If you pass a const ref and copy into the member in the constructor or you pass a copy and move it into the container. The const ref might be slightly more efficient, because the move for a POD is effectively a copy. 

But for config structs this should not be a concern.

BUT: we should never pass a mutable ref and move that, because it moves from the source value that the caller holds, which is super unexpected.

PR #2514, opened 2023-10-05
fix: (Layer-based) Geometry construction handles layerless volumes

paulgessinger:
In calorimeter developments with ODD we noticed that if DD4hep produces a volume without layers, which currently is the case with the calorimeter implementation that we have there, the geometry construction will fail to produce a geometry. This changes the layer-based geometry construction to ignore volumes without layers for now.


andiwand (approved):
lgtm

PR #2519, opened 2023-10-08
feat: Random transformations of aligned sensors

LaraCalic:


LaraCalic:
In this PR, I am implementing the following changes:
- taking the 'aligned' decorator and adding random alignment for each sensor, and integrating instances of that specific struct
  - decorator uses random translations and rotations (these changes enable the decorator to introduce misalignments in structures and refresh its internal mapping of transformation differences)

Examples/Detectors/ContextualDetector/include/ActsExamples/ContextualDetector/AlignmentDecorator.hpp
```diff
@@ -50,42 +55,86 @@ class AlignmentDecorator : public IContextDecorator {
...
-      }
-      if (cfg.aSigmaZ != 0.) {
-        trf *=
-            Acts::AngleAxis3(cfg.aSigmaZ * gauss(rng), Acts::Vector3::UnitZ());
-      }
+    if (iov < cfg.misalignmentInstructions.size()) {
+      const MisalignmentStruct& misalignment = cfg.misalignmentInstructions[iov];
+
+      // place where we take care of transformations before misalignment (for each event)
+      cfg.preMisalignmentTransforms[iov] = trf;
```
paulgessinger, line 113: Is this only for groups then? For the sensors we can get this directly from the surfaces?

Examples/Detectors/ContextualDetector/include/ActsExamples/ContextualDetector/ExternalAlignmentDecorator.hpp
```diff
@@ -65,10 +61,10 @@ class ExternalAlignmentDecorator : public AlignmentDecorator {
   /// added in order.
   ///
   /// @param context the bare (or at least non-const) Event context
-  ProcessCode decorate(AlgorithmContext& context) override;
+  ProcessCode decorate(AlgorithmContext& context) final override;
```
paulgessinger, line 64: Generally we only do one out of `virtual`, `override` and `final`, as `final` implies `override`.

Examples/Detectors/ContextualDetector/include/ActsExamples/ContextualDetector/ExternallyAlignedDetectorElement.hpp
```diff
@@ -38,33 +39,57 @@ namespace Contextual {
...
+    Acts::Translation3 translation; 
+    Acts::Rotation3 rotation;        
+};
 class ExternallyAlignedDetectorElement
     : public Generic::GenericDetectorElement {
  public:
   struct AlignmentStore {
     // GenericDetector identifiers are sequential
     std::vector<Acts::Transform3> transforms;
-    size_t lastAccessed = 0;
```
paulgessinger, line 47: I thought I remember that this control IOVs, does it not?

Examples/Detectors/ContextualDetector/include/ActsExamples/ContextualDetector/AlignmentDecorator.hpp
paulgessinger, line 1: I don't fully understand this change. This already previously implemented a mechanism to calculate random rotation and translation shifts, but by configuring the individual sigmas.

PR #2535, opened 2023-10-11
refactor: Refactor and fix loop protection

andiwand:
The loop protection should not be direction dependent anymore. The limit should be set even if it is higher.

### DIFF
--- a/Core/include/Acts/Propagator/detail/LoopProtection.hpp
+++ b/Core/include/Acts/Propagator/detail/LoopProtection.hpp
@@ -41,16 +41,16 @@ void setupLoopProtection(propagator_state_t& state, const stepper_t& stepper,
   // Transverse component at start is taken for the loop protection
   const double p = stepper.absoluteMomentum(state.stepping);
   // Calculate the full helix path
-  const double helixPath = state.options.direction * 2 * M_PI * p / B;
-  // And set it as the loop limit if it overwrites the internal limit
-  double loopLimit = state.options.loopFraction * helixPath;
-  double pathLimit = pathAborter.internalLimit;
-  if (std::abs(loopLimit) < std::abs(pathLimit)) {
-    pathAborter.internalLimit = loopLimit;
+  const double helixPath = 2 * M_PI * p / B;
+  // Apply the loop fraction
+  const double loopLimit = state.options.loopFraction * helixPath;
 
-    ACTS_VERBOSE("Path aborter limit set to "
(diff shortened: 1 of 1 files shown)


paulgessinger (commented):
Ok, we lose the direction sign, but that's intended because we don't want to apply this when stepping back (e.g. overstepping), and when we're actually going backward, the path length is also positive?

Core/include/Acts/Propagator/detail/LoopProtection.hpp
```diff
@@ -18,11 +18,17 @@ namespace detail {
...
-                         path_aborter_t& pathAborter, const Logger& logger) {
+                         path_aborter_t& pathAborter, bool releaseLimit,
+                         const Logger& logger) {
   if (!state.options.loopProtection) {
     return;
   }
 
+  if (releaseLimit) {
+    pathAborter.internalLimit =
+        state.options.direction * std::numeric_limits<double>::max();
```
paulgessinger, line 29: You recover the sign here for the internal limit in this case?

andiwand: yeah not happy about the sign of this property but currently we use a sign for the accumulated path length and this property is compared to it and it seems natural to keep the sign on this one too

Core/include/Acts/Propagator/Propagator.hpp
```diff
@@ -128,7 +128,7 @@ struct PropagatorOptions : public PropagatorPlainOptions {
     eoptions.maxRungeKuttaStepTrials = maxRungeKuttaStepTrials;
     eoptions.maxStepSize = maxStepSize;
     eoptions.targetTolerance = targetTolerance;
-    eoptions.pathLimit = direction * std::abs(pathLimit);
+    eoptions.pathLimit = pathLimit;
```
paulgessinger, line 131: It's still not obvious for me that this doesn't break for backwards propagation. The treatment of how the path aborter applies this is unchanged. Is the pathlength accumulation in the aborter signed or unsigned?

andiwand: right this should go again. I attempted to remove the sign of the path length until I realized that I kept it intentionally before

I think this still works because the path limit aborter is very sign agnostic. we apply `abs` multiple times which seems a bit of a too wide contract to me but I would keep it for now

paulgessinger (dismissed):

paulgessinger (approved):

PR #2537, opened 2023-10-11
feat: component builder proxies 

asalzburger:
This PR introduces the proxy system for building the new Detector model.

When parsing the Geometry modeller (DD4Hep, GeoModel, etc.), this can populate proxies, i.e. a description how the builder should be set up. 
In a second step, a helper module will fill the remaining space with empty volumes - and complete the proxy tree.

Finally the top level builder can be constructed from a fully deployed proxy tree.


Core/include/Acts/Detector/ComponentBuilderProxy.hpp
```diff
@@ -0,0 +1,168 @@
...
+
+  /// The transform
+  Transform3 transform = Transform3::Identity();
+
+  /// Request the builder of this object
+  std::shared_ptr<const IDetectorComponentBuilder> builder() const {
+    if (std::holds_alternative<VolumeProxy>(this->holder)) {
+      return std::get<VolumeProxy>(this->holder).builder;
+    } else {
+      std::runtime_error(
```
paulgessinger, line 75:
```suggestion
      throw std::runtime_error(
```

Core/include/Acts/Detector/ComponentBuilderProxy.hpp
```diff
@@ -0,0 +1,168 @@
...
+  };
+
+  /// The proxy name - this needs to be uniuque
+  std::string name = "";
+
+  /// The Parent proxy, optionally set
+  std::shared_ptr<ComponentBuilderProxy> parent = nullptr;
+
+  /// The volume proxy, optionally set
+  std::variant<VolumeProxy, ContainerProxy> holder = VolumeProxy{};
```
paulgessinger, line 59: If I'm reading correctly what you're doing with this, wouldn't this be much easier done using another interface and inheriting from it? Then you can get rid of the manual checking of alternatives, and just use virtual function calls, maybe.

Core/include/Acts/Detector/ComponentBuilderProxy.hpp
```diff
@@ -0,0 +1,168 @@
...
+  /// @brief  Definition of a container
+  struct ContainerProxy {
+    /// The binning instruction
+    std::vector<BinningValue> binning = {};
+    /// The contained component builders, in case of a container builder
+    std::vector<std::shared_ptr<ComponentBuilderProxy>> children = {};
+  };
+
+  /// The proxy name - this needs to be uniuque
+  std::string name = "";
```
paulgessinger, line 53: Do these have to be settable from the outside? Otherwise I'd suggest following the usual naming of `m_name` etc, and then also renaming the method arguments in this class to not be single letters, where possible.

Core/include/Acts/Detector/ComponentBuilderProxy.hpp
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /// @brief Create a child volume proxy
+  ///
+  /// @param n The name of the child volume proxy
+  /// @param t The transform of the child volume proxy
+  /// @param bt The boundary type of the child volume proxy
+  /// @param bv The boundary values of the child volume proxy
+  /// @param b The builder of the child volume proxy
+  std::shared_ptr<ComponentBuilderProxy> addChildVolumeProxy(
```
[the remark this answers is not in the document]
asalzburger: Fair enough, this can be changed.

Core/include/Acts/Detector/ComponentBuilderProxy.hpp
```diff
@@ -0,0 +1,168 @@
...
+    std::vector<BinningValue> binning = {};
+    /// The contained component builders, in case of a container builder
+    std::vector<std::shared_ptr<ComponentBuilderProxy>> children = {};
+  };
+
+  /// The proxy name - this needs to be uniuque
+  std::string name = "";
+
+  /// The Parent proxy, optionally set
+  std::shared_ptr<ComponentBuilderProxy> parent = nullptr;
```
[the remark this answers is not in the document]
asalzburger: I think the link to the parent is not needed, actually, the building call will be a direct tree call:
-> starting from the top builder downwards.

I will remove the parent then.

Core/include/Acts/Detector/ComponentBuilderProxy.hpp
asalzburger: I would have to do dynamic_cast then, I don't mind - either way works.

Core/include/Acts/Detector/ComponentBuilderProxy.hpp
asalzburger: In general, if the name of the proxies are used to then generate the builder tools (that's their only reason to be there), then it could be useful to harmonise namings when the full tree is constructed, but again, that's not really necessary.

Core/include/Acts/Detector/ComponentBuilderProxy.hpp
paulgessinger: But can't you have a method which just returns what you create in the `if`/`else` now?

Core/include/Acts/Detector/ComponentBuilderProxy.hpp
```diff
@@ -0,0 +1,168 @@
...
+    proxy->boundsType = bt;
+    proxy->boundaryValues = bv;
+    proxy->holder = VolumeProxy{b};
+    if (std::holds_alternative<ContainerProxy>(this->holder)) {
+      std::get<ContainerProxy>(this->holder).children.push_back(proxy);
+    } else {
+      throw std::runtime_error(
+          "ComponentBuilderProxy::addChildVolumeProxy() - "
+          "Cannot add a child volume proxy to a non-container proxy");
+    }
```
paulgessinger, line 134: And here you could do `addChild` which throws for the derived class which cannot have child volume proxies.

Core/include/Acts/Detector/ComponentBuilderProxy.hpp
paulgessinger: I see your point. In the rest of the code base we really only expose members if we expect them to be set from other classes etc, and we prefix internal members with `m_`, among other things so we have fewer naming issues, and that it's clear that it's a member variable.

Core/include/Acts/Detector/ComponentBuilderProxy.hpp
```diff
@@ -0,0 +1,168 @@
...
+  /// The boundary values - to be interpreted together with boundary type
+  std::vector<ActsScalar> boundaryValues = {};
+
+  /// The transform
+  Transform3 transform = Transform3::Identity();
+
+  /// Request the builder of this object
+  std::shared_ptr<const IDetectorComponentBuilder> builder() const {
+    if (std::holds_alternative<VolumeProxy>(this->holder)) {
+      return std::get<VolumeProxy>(this->holder).builder;
```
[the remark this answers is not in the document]
asalzburger: True, of course.

Core/include/Acts/Detector/ComponentBuilderProxy.hpp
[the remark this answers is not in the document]
asalzburger: Changed to a bare pointer.

Core/include/Acts/Detector/ComponentBuilderProxy.hpp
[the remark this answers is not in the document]
asalzburger: Actually, there return of the object is not necessary - so, first I will change this into void, then we can see - probably internally I will still need something copyable.

asalzburger:
Closing this one, the development is superseded by #2550

PR #2538, opened 2023-10-11
fix: Correct the Sign of Drift Distance

beomki-yeo:
The PR makes the drift distance of line measurement to be postivie always.


Tests/UnitTests/Core/TrackFitting/GainMatrixUpdaterTests.cpp
```diff
@@ -73,10 +74,13 @@ BOOST_AUTO_TEST_CASE(Update) {
...
   // Gain matrix update and filtered state
-  BOOST_CHECK(GainMatrixUpdater()
-                  .
-                  operator()<VectorMultiTrajectory>(tgContext, ts)
-                  .ok());
+  auto surface =
+      Surface::makeShared<Acts::DiscSurface>(Acts::Transform3::Identity());
+  BOOST_CHECK(
+      GainMatrixUpdater()
+          .
```
andiwand, line 81: is this what clang format does? :scream:

Core/include/Acts/Surfaces/Surface.hpp
```diff
@@ -389,6 +389,8 @@ class Surface : public virtual GeometryObject,
   virtual FreeToPathMatrix freeToPathDerivative(
       const GeometryContext& gctx, const FreeVector& parameters) const;
 
+  virtual BoundVector correctSign(const BoundVector& boundParams) const;
```
andiwand, line 392: I am not too happy with this name. Maybe
```suggestion
  virtual BoundVector boundToMeasurement(const BoundVector& boundParams) const;
```
?

paulgessinger: On thing we need to be careful here is that in principle we do not want to have separate coordinate systems between surface and measurement. As we discussed, the sign in the line surface is an edge case. 

I think we should really avoid introducing the concept of a more or less generic coordinate transform between surface frame and measurement frame.

Core/include/Acts/TrackFitting/GlobalChiSquareFitter.hpp
```diff
@@ -58,7 +58,7 @@ struct Gx2FitterExtensions {
       Delegate<void(const GeometryContext&, const CalibrationContext&,
                     const SourceLink&, TrackStateProxy)>;
 
-  using Updater = Delegate<Result<void>(const GeometryContext&, TrackStateProxy,
+  using Updater = Delegate<Result<void>(const GeometryContext&, const Surface*, TrackStateProxy,
```
paulgessinger, line 61: Can you not get the surface from the track state?

Core/src/TrackFitting/GainMatrixUpdater.cpp
```diff
@@ -64,8 +64,9 @@ std::tuple<double, std::error_code> GainMatrixUpdater::visitMeasurement(
       return false;                                           // abort execution
     }
 
+    auto signCorrected = surface->correctSign(trackState.predicted);
```
paulgessinger, line 67: I think you should be able to do 
```cpp
trackState.referenceSurface()
```
to get the surface of the track state. This way, you don't need to pipe through the surface explicitly.

paulgessinger (commented):
I'm wondering if, in order to not introduce the "measurement frame" as a first-class concept, this should be done inside the the KF updater and use a dynamic cast to figure out if it's looking at a `LineSurface`.

Core/include/Acts/Surfaces/Surface.hpp
andiwand: hm makes sense but in principle this is what we do here right?

something else: do we have to propagate that `abs` into the jacobian?

paulgessinger: > hm makes sense but in principle this is what we do here right?

Basically, I think that having this be an interface method promotes this as a concept.

> something else: do we have to propagate that `abs` into the jacobian?
Now that's a good question. It's not obvious to me if the covariance also contains information on the radius sign.

XiaocongAi:
[like]  Xiaocong Ai reacted to your message:
________________________________

XiaocongAi:
> I'm wondering if, in order to not introduce the "measurement frame" as a first-class concept, this should be done inside the the KF updater and use a dynamic cast to figure out if it's looking at a `LineSurface`.

I totally agree. This is what we are exactly doing for a drift chamber tracking locally.

Core/include/Acts/Surfaces/Surface.hpp
beomki-yeo: > It's not obvious to me if the covariance also contains information on the radius sign.

I don't think we need to do this for any jacobians during propagation
But I guess the off diagonal elements of <2x2> measurement in KF update should be flipped with drift distance (L)
```
L -> -L
cov <L ,z> -> - cov<L,z>
```

Core/src/TrackFitting/GainMatrixUpdater.cpp
beomki-yeo: I cannot find any surface method in `InternalTrackState` :thinking: 
```
/home/beomki/projects/acts/acts/Core/src/TrackFitting/GainMatrixUpdater.cpp:67:37: error: ‘struct Acts::GainMatrixUpdater::InternalTrackState’ has no member named ‘referenceSurface’
   67 |     auto signCorrected = trackState.referenceSurface()->correctSign(trackState.predicted);
```

Core/src/TrackFitting/GainMatrixUpdater.cpp
```diff
@@ -43,9 +44,16 @@ std::tuple<double, std::error_code> GainMatrixUpdater::visitMeasurement(
...
 
-    const auto H = trackState.projector
-                       .template topLeftCorner<kMeasurementSize, eBoundSize>()
-                       .eval();
+    auto H = trackState.projector
+                 .template topLeftCorner<kMeasurementSize, eBoundSize>()
+                 .eval();
+
+    const LineSurface* line = dynamic_cast<const LineSurface*>(surface);
+    if (line) {
```
andiwand, line 52:
```suggestion
    if (const LineSurface* line = dynamic_cast<const LineSurface*>(surface); line) {
```

Examples/Io/Csv/src/CsvMeasurementWriter.cpp
```diff
@@ -109,6 +113,17 @@ ActsExamples::ProcessCode ActsExamples::CsvMeasurementWriter::writeT(
           meas.local_key = 0;
           // Create a full set of parameters
           auto parameters = (m.expander() * m.parameters()).eval();
+
+          // Take a non-negative drfit distance measurement for Line Surface
+          const Acts::Surface* surface =
+              m_cfg.trackingGeometry->findSurface(geoId);
+          const Acts::LineSurface* line = dynamic_cast<const Acts::LineSurface*>(surface);
+          if (line) {
```
andiwand, line 121:
```suggestion
          if (dynamic_cast<const Acts::LineSurface*>(surface) != nullptr) {
```

Core/src/TrackFitting/GainMatrixUpdater.cpp
beomki-yeo: Fixed

Examples/Io/Csv/src/CsvMeasurementWriter.cpp
beomki-yeo: Fixed

Core/include/Acts/TrackFitting/GlobalChiSquareFitter.hpp
paulgessinger: What do you think about this one @beomki-yeo? I think it's preferable to not pass in the surface explicitly, if you can get it from the track state.

Core/src/TrackFitting/GainMatrixUpdater.cpp
paulgessinger: Yeah, that's an implementation detail of the gain matrix updater. Can you just add the surface pointer here, and fill it from the gain matrix updater template function with the pointer you get from the track state proxy?

acts-project-service:
## :red_circle: Athena integration test results [eb66084e5ffdde8bb5d68fa5a71216a5765430d7]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/33473926) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6397248)!

andiwand:
How should we proceed on this @beomki-yeo @paulgessinger ?

beomki-yeo:
I have to fix the PR as suggested by Paul. I have been just procrastinating. Do you need this PR merged soon?

andiwand:
> I have to fix the PR as suggested by Paul. I have been just procrastinating. Do you need this PR merged soon?

No hurry just wanted to see what the status is since this PR is getting quite outdated

beomki-yeo:
I will open it later when I can work on it.

PR #2559, opened 2023-10-19
feat: Proposal for required config parameters in Examples

andiwand:
A common pitfall with our Examples is that all the config parameters have and must have a default value. Many of these parameters are experiment dependent and users end up using parameters they didn't know about in the first place.

I propose to improve this situation by boxing these parameters into a `Required` type which acts like a `std::optional`. The config object is still default constructible and the boxing type is wired to Python.

My current implementation is asserting on copies of uninitialized `Required` objects. In Python this will throw an error directly at the algorithm construction because we copy the config into the algorithm which will fail if an uninitialized value was encountered. 

That interface might be unintuitive. Another idea was to validate the config in the algorithms constructor. This requires more code but provides opportunity for better error messages.

I exercised this only on the `ParametricParticleGenerator` for now.

### DIFF
--- a/Examples/Algorithms/Generators/ActsExamples/Generators/ParametricParticleGenerator.cpp
+++ b/Examples/Algorithms/Generators/ActsExamples/Generators/ParametricParticleGenerator.cpp
@@ -1,6 +1,6 @@
 // This file is part of the Acts project.
 //
-// Copyright (C) 2017-2020 CERN for the benefit of the Acts project
+// Copyright (C) 2017-2023 CERN for the benefit of the Acts project
 //
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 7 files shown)


paulgessinger:
For my understanding: this is supposed to allow us from having non-default config parameters that the user needs to explicitly set? We can still have default config parameters where this is not necessary.

andiwand:
> For my understanding: this is supposed to allow us from having non-default config parameters that the user needs to explicitly set? We can still have default config parameters where this is not necessary.

yes exactly. we have a bunch of fields where we except user input but default it anyways because we want to have a default constructible config

benjaminhuth:
So in principle I like the idea to wrap the requirement of explicitly set parameters into a concrete type.

The only thing which feels a bit weird, that the config `struct` can neither be copied nor moved when not having set value to the required parameters... 

But I see that copying is the only way to hook into the algorithm construction here generically without weird introspection macros, but it still feels weird...

andiwand:
yeah I agree. as you said this was the only way to hook into the config passing to the algorithm but it might not be worth the convenience.

the other option I see is to validate the config in the algorithm constructor or to add a validation function to the config struct. this would also have a more explicit way of dealing with errors.

let me try the second approach

andiwand:
closing as a more general solution would be preferable

PR #2532, opened 2023-10-11
refactor: Refactor and fix `accuracy` property in `ConstrainedStep`

andiwand:
I think the accuracy property of the `ConstrainedStep` should always be positive as we want to use it consistently in back stepping.

I encountered a situation where the overstepping was not accurate enough (the overstepping step exceeded the eigen stepper error tolerance), then the step length reduction mechanism fails and ends in an endless loop.

Pulled this out of https://github.com/acts-project/acts/pull/2518

PR #2561, opened 2023-10-20
perf: Improve `ConstrainedStep` footprint after #2532

andiwand:
Fixes an oversight in https://github.com/acts-project/acts/pull/2532


paulgessinger (approved):
Thanks!

paulgessinger:
Athena failure is unrelated.

PR #2564, opened 2023-10-21
feat!: GSF mixture mode and refactor

benjaminhuth:
Adds a mode reduction method, and also changes the `enum` name for the option from `FinalReductionMethod` to `ComponentMergeMethod` (thus breaking!)

Also fixes a bug in the covariance computation which now results in (slightly) better pulls


Core/include/Acts/TrackFitting/detail/MergeGaussianMixture.hpp
```diff
@@ -195,20 +196,150 @@ auto gaussianMixtureMeanCov(const components_t components,
...
+
+  // We only store the highest mode
+  // TODO Replace with Acts::Result
+  std::optional<ActsVector<D>> mode;
+  double mode_val = 0;
+
+  // Algorithm parameters
+  constexpr double tol = 1.e-8;
+  constexpr double eigv_max = 0.01;
+  constexpr int iter_max = 50;
```
andiwand, line 299: are these params stable/optimal enough to be used in all scenarios?

andiwand (commented):
looks good from my side. just one comment

Core/include/Acts/TrackFitting/detail/MergeGaussianMixture.hpp
benjaminhuth: Hmm so it was quite a while ago that I implemented this, and I'm actually not totally sure. But since I test several mixture configurations in the unit test, it can't be too bad... But it could be certainly a performance optimization to tune these parameters...

Another point: Currently the interface (an `enum`) wouldn't allow for configuration anyways, so this would have to be changed to a functor-based interface (which would be good...). However, I definitely would to it in a separate PR.

andiwand (dismissed):

paulgessinger:
Doc job complains about a missing enum target:

```
/home/runner/work/acts/acts/docs/core/reconstruction/track_fitting.md:70: WARNING: cpp:enum reference target not found: Acts::MixtureReductionMethod
```

PR #2576, opened 2023-10-24
feat: ML seed selection

Corentin-Allaire:
This pull request add a ML based seed selection that can be used to reduce the number of seed after the seeding step.

This also add : 

- Performances plots based on the distance to the closest track 
- A CSV writer for seed (used in the training of the NN)
- the option to run on GPU for the ML training script


Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -43,12 +45,18 @@
     help="Use the Ml Ambiguity Solver instead of the classical one",
     action="store_true",
 )
+parser.add_argument(
+    "--MLSeedFilter",
+    help="Use the Ml seed filter to select seed after the seeding step",
+    action="store_true",
+)
```
paulgessinger, line 52: Cue my obligatory comment that I'm not a fan of just adding extra features to this script steered by command line options.

Corentin-Allaire: I don't disagree and I am 100% ready to implement this any other way (but I am not sure which one)

PR #2588, opened 2023-10-27
try: CKF with fuzzy navigation

andiwand:
just hacked this together to see some physmon results


andiwand:
closing after demonstration is done

PR #2595, opened 2023-10-29
refactor: Remove path limit from navigation

andiwand:
I don't think this is generally necessary for navigation. The path limit is handled by an aborter which can be orthogonal with the navigation. In principle it could be a loss of optimization but usually our geometries are segmented well enough that this will not remove a meaningful amount of surfaces IMO

blocked by
- https://github.com/acts-project/acts/pull/2599


andiwand:
that navigation failure in the test geometry is very concerning

andiwand:
this is causing navigation failures which I did not expect and don't think this should happen.

when we intersect cylinders, the path limit discards one of the solutions which seems to keep the navigation stable in back propagation. removing the limit causes the navigation to fail. https://github.com/acts-project/acts/pull/2599 improves the situation but does not fix it completely. I found at least one case where we resolve a layer that was not resolved before

having a single surface stream in the navigator would fix this I think but this would be a bigger change

andiwand:

PR #2598, opened 2023-10-30
refactor: Remove target estimation from navigator

andiwand:
This is very unintuitive and might result in unwanted behavior


andiwand:
> And the idea is to have this functionality fully contained in the surface aborter?

the code I remove here is trying to estimate the target layer and volume based on an intersection with the target surface. this is done several times during navigation since this estimate could change.

I do not think that is it necessary to estimate those and if the user wants to decrease navigation candidates the target layer or volume should be provided directly

andiwand:

PR #2601, opened 2023-10-31
feat: Add high-stats fuzzy random navigation test

paulgessinger:


andiwand:
what is the benefit of having this partially in Python instead of having a pure C++ unit / integration test?

btw I did something similar here https://github.com/acts-project/acts/pull/2625/files#diff-424460bede45a6537feb4431a533a08105032049974660609be2ff03d950047b

paulgessinger:
@andiwand Firstly, I was just playing around with this for now. Secondly, does your test run arbitrary start and target points (that's what I was trying to go for here). Thirdly, I wanted to use ODD so we actually can theoretically hit all the edge cases, so that's why it's split in C++ and Python.

andiwand:
right I didn't think of ODD yet. but I guess this could still be a test inside the C++ examples right?

but having it within the sequencer might be interesting if we could couple it to the event generation somehow, run through a generic propagation algorithm and then compare the surfaces in python

andiwand:
oh I should also mention that I wrote 2 try all navigator in https://github.com/acts-project/acts/pull/2625 which we can use to validate other

paulgessinger:
@andiwand Honestly, I was just thinking that we have one canonical way to setup ODD now, and I thought we could just use it for this.

PR #2544, opened 2023-10-14
refactor: use time info in Kalman vertex fit

felix-russo:
Since we introduced the time coordinate in tracking, I propose to the same in the Kalman vertex fitter now that the Jacobians of `HelicalTrackLinearizer` are filled entirely.

Starting to factor this out, blocked by:

- #2707
- #2708

PR #2604, opened 2023-10-31
refactor: Find physmon crash

felix-russo:
Trying to find where physmon crashed in #2544


paulgessinger:
So this is a subset of the other PR?

andiwand:
does it not crash locally?

felix-russo:
Sorry, I should have tested locally (which is what I continued to do and forgot about this pr)

PR #2608, opened 2023-11-01
feat: Update Json converters and enhance Detray version writing

asalzburger:
This PR updates the Json writers for the new Detector geometry, and particular for the detray export.

The final output needs still a bit fine-tuning and the material files are still empty.

It adds also support for the `IndexedRootVolumes`, a grid to find the initial volume to start from, it uses the same infrastructure as the `IndexedSurfacesUpdators` hence this code was centrally put into a helper function.


Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp
```diff
@@ -26,6 +26,13 @@ namespace AxisJsonConverter {
 /// @return a json object to represent the axis
 nlohmann::json toJson(const IAxis& ia);
 
+/// Convert an axis to json - detay style
```
paulgessinger, line 29:
```suggestion
/// Convert an axis to json - detray style
```

Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp
```diff
@@ -83,30 +144,29 @@ nlohmann::json toJson(const grid_type& grid) {
 /// @note the axis generator also defines the grid dimension
 ///
 /// @return a grid object
-template <typename axis_generator_type>
+template <typename axis_generator_type,
+          typename value_type = std::vector<std::size_t>>
 auto fromJson(const nlohmann::json& jGrid,
               const axis_generator_type& aGenerator) {
```
paulgessinger, line 150: Why not just pass the grid rather than the generator?

Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp
```diff
@@ -0,0 +1,253 @@
...
+#include <tuple>
+
+namespace Acts {
+using namespace Experimental::detail::GridAxisGenerators;
+
+namespace IndexedGridJsonHelper {
+
+// Generate the possible axes in this case
+static auto s_possibleAxes =
+    std::tuple<EqBound, EqOpen, EqClosed,
```
paulgessinger, line 26: Just to mention that `tuple` is not very efficient with many types, so if at all possible, avoid using a giant tuple.

paulgessinger:
Could you attach a sample json file to the PR? Just wondering what this ends up looking like.
If I have some time, I might look into writing a schema to validate the format.

Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp
asalzburger: This is for JsonIO, so I think we might be anyway limited by the IO and this is not performance critical.

Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp
asalzburger: We use everywhere the following pattern:
```Object fromJson(const nlohmann::json& jInput, const Options& a)```

I can keep this pattern, if the Options is the way how to generate the grid, otherwise I would have to change it to

```void fromJson(const nlohmann::json& jinput, grid_type& grid)```

Where the grid is filled in the `fromJson` method - which I don't like.

**Comment by asalzburger** [comment]
> Could you attach a sample json file to the PR? Just wondering what this ends up looking like. If I have some time, I might look into writing a schema to validate the format.

Yes, I will, there's still some refinement to be done in the interplay with detray, so we will do this once the chain is closed.

File: Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp (hunk 2 above)
Comment (asalzburger) [reply Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp:150]: @paulgessinger does this satisfy you as an answer?

**Comment by niermann999** [comment]
> Could you attach a sample json file to the PR? Just wondering what this ends up looking like. If I have some time, I might look into writing a schema to validate the format.

Some Json schemas for the detray files exist in https://github.com/acts-project/detray/tree/main/tests/validation/python/validation/json_schema

File: Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp (hunk 3 above)
Comment (paulgessinger) [reply Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp:26]: It's inefficient at compile time, since it recursively instantiates structs to generate the type.

**Comment by paulgessinger** [comment]
@niermann999 now that is excellent! Is this the standard json schema format? It might make sense to put them into separate JSON files so they are usable with standard command line tools.

File: Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp (hunk 3 above)
Comment (asalzburger) [reply Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp:26]: Oh, I see - so you have another suggestion how to implement this then?

File: Plugins/Json/include/Acts/Plugins/Json/DetrayJsonHelper.hpp [hunk 4]
```diff
@@ -90,5 +93,42 @@ inline static void addVolumeLink(nlohmann::json& jSurface, int vLink) {
...
+  // Default is `brute_force`
+  size_t accLink = 0u;
+  if (casts.size() == 2u) {
+    if (casts[0u] == binX && casts[1u] == binY) {
+      accLink = 1u;
+    } else if (casts[0u] == binR && casts[1u] == binPhi) {
+      accLink = 3u;
+    } else if (casts[0u] == binZ && casts[1u] == binPhi) {
+      accLink = 4u;
+    } else if (casts[0u] == binZ && casts[1u] == binR) {
```
niermann999, line 120: Is this possible in ACTS, to have 3D grid with only two axes?

Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp
```diff
@@ -66,12 +75,64 @@ nlohmann::json toJson(const grid_type& grid) {
...
+        typename grid_type::index_t lbin;
+        lbin[0u] = ib0;
+        lbin[1u] = ib1;
+        nlohmann::json jBin;
+        jBin["loc_index"] = lbin;
+        jBin["content"] = grid.atLocalBins(lbin);
+        jData.push_back(jBin);
+      }
+    }
+  }
```
niermann999, line 130:
```suggestion
typename grid_type::index_t lbin;
for (unsigned int ib0 = 1u; ib0 <= axes[0u]->getNBins(); ++ib0) {
      lbin[0u] = ib0;
      // 2D connections
      if constexpr (grid_type::DIM == 2u) {
          for (unsigned int ib1 = 1u; ib1 <= axes[1u]->getNBins(); ++ib1) {
              lbin[1u] = ib1;
          }
       }
   }
  nlohmann::json jBin;
  jBin["loc_index"] = lbin;
  jBin["content"] = grid.atLocalBins(lbin);
  jData.push_back(jBin);
```
Might be easier to extend to 3D grids at some point

Plugins/Json/include/Acts/Plugins/Json/IndexedSurfacesJsonConverter.hpp
```diff
@@ -99,42 +56,56 @@ void convert(nlohmann::json& jIndexedSurfaces,
...
     // Get the surface updator
     auto indexedSurfaces = std::get<SubDelegateType>(castedDelegate->updators);
-    jIndexedSurfaces = convertImpl<SubDelegateType>(indexedSurfaces);
+    jIndexedSurfaces = IndexedGridJsonHelper::convertImpl<SubDelegateType>(
+        indexedSurfaces, detray);
+    if (detray) {
+      nlohmann::json jAccLink;
+      jAccLink["type"] =
+          DetrayJsonHelper::accelerationLink(indexedSurfaces.casts);
+      jAccLink["index"] = 0;
```
niermann999, line 65: In detray this enumerates the grids of a given type. Otherwise, it is better to set it to `std::size_t` max, so it becomes clear that the index of the grid is unknown

Plugins/Json/src/DetectorJsonConverter.cpp
```diff
@@ -108,32 +122,55 @@ nlohmann::json Acts::DetectorJsonConverter::toJsonDetray(
...
+      detector.detectorVolumeFinder(), true);
 
+  // (1) Geometry
   // For detray, the number of surfaces and portals are collected
   // from the translated volumes
-  jHeader["type"] = "detray";
-  jHeader["surface_count"] = nSurfaces;
-  jDetector["header"] = jHeader;
-  jDetector["data"] = jData;
+  jGeometryHeader["type"] = "detray";
```
niermann999, line 132: Should be clear from the `version` field by now

Plugins/Json/src/GridJsonConverter.cpp
```diff
@@ -24,3 +25,13 @@ nlohmann::json Acts::AxisJsonConverter::toJson(const IAxis& ia) {
...
   return jAxis;
 }
+
+nlohmann::json Acts::AxisJsonConverter::toJsonDetray(const IAxis& ia) {
+  nlohmann::json jAxis;
+  jAxis["bounds"] =
+      ia.getBoundaryType() == Acts::detail::AxisBoundaryType::Bound ? 1 : 2;
+  jAxis["binning"] = ia.isEquidistant() ? 0 : 1;
+  jAxis["bins"] = ia.getNBins();
+  jAxis["edges"] = ia.getBinEdges();
```
niermann999, line 35: I am assuming these are {min, max} for regular binning?

niermann999:
> @niermann999 now that is excellent! Is this the standard json schema format? It might make sense to put them into separate JSON files so they are usable with standard command line tools.

As far as I know it is (I used https://json-schema.org to write the thing). Up to this point there is a small python executable that  runs the validation on a given input json file using the `jsonschema` module

Examples/Python/src/Json.cpp
```diff
@@ -195,9 +197,23 @@ void addJson(Context& ctx) {
...
-              out << jDetector.dump(4);
+              out.open(name + "_geometry_detray.json");
+              out << jGeometry.dump(4);
+              out.close();
+
+              out.open(name + "_surface_grids_detray.json");
+              out << jSurfaceGrids.dump(4);
+              out.close();
+
+              out.open(name + "_material_detray.json");
```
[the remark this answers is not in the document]
asalzburger: But I guess it shouldn't be homogenous only.

Plugins/Json/include/Acts/Plugins/Json/DetrayJsonHelper.hpp
asalzburger: No, I tried to adapt to the detray grids, You do expect a 3D one, right?

Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp
asalzburger: That suggestion doesn't work, the filling of the content has to be once within the tighter, once without the outer loop.  I could do a fake inner loop with dim=1 and then with an if not fill `bin[1u]` but that doesn't make it clearer.

Plugins/Json/src/DetectorJsonConverter.cpp
asalzburger: yes, this is more for the ACTS side.

Plugins/Json/include/Acts/Plugins/Json/DetrayJsonHelper.hpp
niermann999: For the volume grid at some point, yes. The surface and material grids are all 2D (so far)

Plugins/Json/src/GridJsonConverter.cpp
asalzburger: No they are all the edits, do you expect min/max ? in this case I will catch it here.

niermann999: For regular binning it is just min/max. For irregular binning it expects the lower bin edges (this will be the case e.g. for the volume grid)

Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp
paulgessinger: Not sure I get it. You're not passing an options struct here, but the generator object. What's the difference between passing the generator object that you call to make the grid, compare to just passing the grid?

Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp
```diff
@@ -83,30 +144,29 @@ nlohmann::json toJson(const grid_type& grid) {
...
-template <typename axis_generator_type>
+template <typename axis_generator_type,
+          typename value_type = std::vector<std::size_t>>
 auto fromJson(const nlohmann::json& jGrid,
               const axis_generator_type& aGenerator) {
   // Generate the grid
-  using GridType =
-      typename axis_generator_type::template grid_type<std::vector<size_t>>;
+  using GridType = typename axis_generator_type::template grid_type<value_type>;
   GridType grid(aGenerator());
```
paulgessinger, line 153: This is the only time you use the `aGenerator`, so why not just template on and accept `GridType grid`?

Plugins/Json/include/Acts/Plugins/Json/DetectorVolumeFinderJsonConverter.hpp
```diff
@@ -0,0 +1,109 @@
...
+      GridType, Acts::Experimental::IndexedDetectorVolumeExtractor,
+      Acts::Experimental::DetectorVolumeFiller>;
+  // Get the instance
+  const auto* instance = delegate.instance();
+  auto castedDelegate = dynamic_cast<const DelegateType*>(instance);
+  if (castedDelegate != nullptr) {
+    jIndexedVolumes = IndexedGridJsonHelper::convertImpl<DelegateType>(
+        *castedDelegate, detray);
+    if (detray) {
+      jIndexedVolumes["volume_link"] = 1;
```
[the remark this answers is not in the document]
asalzburger: Ok, we can leave it then for the moment.

Plugins/Json/include/Acts/Plugins/Json/DetrayJsonHelper.hpp
asalzburger: Ok, I will just take a 2D grid and make a 3D grid out of it.

Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp
paulgessinger: So if I read the code right, you're using the tuple only to communicate the types, and not actual values (correct me if I'm wrong). So effectively you're using it to wrap a parameter pack, rather than just using a parameter pack? I don't think I see anywhere the tuple values are actually unpacked. 

And if that's anywhere, what does that even mean: here's an object of ALL axis types?

So I think I'd suggest using the parameter pack directly. If you don't want to list the type more than once (which I don't think is the case because you only call `unrollConvert` in one spot, use a struct/meta-function that encodes only the types.

Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp
asalzburger: The grid is the object that is created by the `fromJson` call. Everywhere in the code we have the same signature and code usage:

```Object fromJson(const nlohmann::json& jInput, const Options& a)```

Now, in case of the Grid, the return object can have different type, so I need some sort of template mechanism. I can not do 

```template <typename GridType> GridType fromJson(const nlohmann::json& jInput)``` because there's neither default nor standard constructors for Grid objects.

Hence, the only other way to do it is:

```template <typename GridType> void fillFromJson(const nlohmann::json& jInput, GridType& grid)```

I, personally, like the generator more, because this code already exists, and I can reuse it - and, as I said before - it keeps the same/similar message signature as in most of the other json code.

File: Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp (hunk 3 above)
Comment (paulgessinger) [reply Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp:26]: Reading the code again, yeah since you never actually put values in that tuple, you really don't need the tuple at all, I don't think.

File: Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp (hunk 10 above)
Comment (asalzburger) [reply Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp:153]: Can you make a concrete suggestion how you want that method to look like?

Do you suggest to have:
```template <typename GridType> void fillFromJson(const nlohmann::json& jInput, GridType& grid)```
?

I can change it to this, but I don't fully see the benefit.  The AxisGenerator will then be used in the parent class that contains the grid, so I am not sure what's to be gained here.

File: Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp (hunk 3 above)
Comment (paulgessinger) [reply Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp:26]: I think this is normally called a "type list", and you can use it to re-extract the parameter pack wherever you need.

https://www.codingwiththomas.com/blog/getting-started-with-typelists

Comment (asalzburger) [reply Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp:26]: Ok, thanks, I will have a look at that.

File: Plugins/Json/src/GridJsonConverter.cpp (hunk 8 above)
Comment (asalzburger) [reply Plugins/Json/src/GridJsonConverter.cpp:35]: Ok, I will then catch the type and and write min/max in this case.

File: Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp (hunk 3 above)
Comment (niermann999) [reply Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp:26]: That's exactly what I used for the detray grid IO :)
https://github.com/acts-project/detray/blob/main/core/include/detray/utils/type_list.hpp

Comment (niermann999) [reply Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp:26]: With mini - unittest:
https://github.com/acts-project/detray/blob/main/tests/unit_tests/core/utils_type_list.cpp

Comment (asalzburger) [reply Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp:26]: Is this a Utility we could use in Acts/Core in general?

File: Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp (hunk 10 above)
Comment (asalzburger) [reply Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp:153]: I don't think you can reuse the grid in this case, the method is called `fromJson` hence you would like to create a grid from json.

Second, that's not about serialising here, on the contrary, you want to create an object from a serialised state, unfortunately - as the object to be created is a templated type - you need to tell the `fromJson` method what actual type you want to have returned.

That's what the generator is for: it knows the type and can help you to create it. Hence in my pseudo-code above I called it "Options", it's not really an option, it is more a directive how to create the return object.

File: Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp (hunk 3 above)
Comment (asalzburger) [reply Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp:26]: I suppose, in the detray implementation would also create this tuple of types:

```c++
template <int N, typename T, typename... Ts>
struct get_at<N, type_list<T, Ts...>> {
    using type = detail::remove_cvref_t<decltype(
        detray::get<N>(detray::tuple<T, Ts...>{}))>;
};
```

This looks to me that the tuple is created at this stage, I will try to see whether I can iterate through without instantiating that tuple.

asalzburger: I think I can just do simply with a `<Head, Tail...>` pattern.

niermann999: In the end, I don't believe I actually used that method, but yes, I guess it can be done without a tuple. In detray, the axes are built by deducing the pack in the building function:
```

    /// @brief End of recursion: build the grid from the @param grid_data
    template <typename... bounds_ts,
              typename... binning_ts>
    static void deserialize(types::type_list<bounds_ts...>,
                            types::type_list<binning_ts...>) {
 and then doing:
```
n_axis::single_axis<bounds_ts, binning_ts>...
```

Comment (paulgessinger) [reply Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp:26]: TIL you can have multiple parameter packs.

Comment (paulgessinger) [reply Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp:26]: If you end up keeping `get_at` with a roundtrip to tuple, you still avoid having tuple on the other code paths.

File: Plugins/Json/include/Acts/Plugins/Json/IndexedSurfacesJsonConverter.hpp (hunk 6 above)
Comment (asalzburger) [reply Plugins/Json/include/Acts/Plugins/Json/IndexedSurfacesJsonConverter.hpp:66]: Done.

File: Plugins/Json/src/GridJsonConverter.cpp (hunk 8 above)
Comment (asalzburger) [reply Plugins/Json/src/GridJsonConverter.cpp:35]: Done.

File: Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp (hunk 3 above)
Comment (asalzburger) [reply Plugins/Json/include/Acts/Plugins/Json/IndexedGridJsonHelper.hpp]: All done.

File: Plugins/Json/include/Acts/Plugins/Json/DetectorVolumeFinderJsonConverter.hpp [hunk 12]
```diff
@@ -0,0 +1,103 @@
...
+/// @brief Unrolling function for catching the right instance
+///
+/// @param jIndexedVolumes the json object to be filled
+/// @param delegate the delegate to be translated
+/// @param detray indicate if this is a detray json to be written out
+/// @param tList the Type list carrying all the possible types
+template <typename... Args>
+void unrollConvert(nlohmann::json& jIndexedVolumes,
+                   const Experimental::DetectorVolumeUpdator& delegate,
+                   bool detray, [[maybe_unused]] TypeList<Args...> tList) {
```
niermann999, line 69: This is always unused, right? That would make it easier to read, I think
```suggestion
                   bool detray, TypeList<Args...>) {
```

Plugins/ActSVG/include/Acts/Plugins/ActSVG/IndexedSurfacesSvgConverter.hpp
```diff
@@ -216,14 +189,17 @@ void convert(const GeometryContext& gctx, const surface_container& surfaces,
...
-template <typename surface_container, typename tuple_type, size_t... I>
+template <typename surface_container, typename Head, typename... Tail>
 void unrollConvert(const GeometryContext& gctx,
                    const surface_container& surfaces, const Options& cOptions,
                    ProtoIndexedSurfaceGrid& sgi,
                    const Experimental::SurfaceCandidatesUpdator& delegate,
-                   const tuple_type& axesTuple, std::index_sequence<I...>) {
-  (convert(gctx, surfaces, cOptions, sgi, delegate, std::get<I>(axesTuple)),
-   ...);
+                   [[maybe_unused]] TypeList<Head, Tail...> tList) {
```
niermann999, line 197:
```suggestion
                   TypeList<Head, Tail...>) {
```

Plugins/ActSVG/include/Acts/Plugins/ActSVG/IndexedSurfacesSvgConverter.hpp
```diff
@@ -216,14 +189,17 @@ void convert(const GeometryContext& gctx, const surface_container& surfaces,
...
 ///
 /// @note parameters are as of the `convertImpl` method
-template <typename surface_container, typename tuple_type, size_t... I>
+template <typename surface_container, typename Head, typename... Tail>
 void unrollConvert(const GeometryContext& gctx,
                    const surface_container& surfaces, const Options& cOptions,
                    ProtoIndexedSurfaceGrid& sgi,
                    const Experimental::SurfaceCandidatesUpdator& delegate,
-                   const tuple_type& axesTuple, std::index_sequence<I...>) {
-  (convert(gctx, surfaces, cOptions, sgi, delegate, std::get<I>(axesTuple)),
```
niermann999, line 225: Why does the call from the DetectorVolumeFinderJsonConverter
```
(convert(jIndexedVolumes, delegate, detray, Args{}), ...);
```
not work here?

Plugins/Json/include/Acts/Plugins/Json/IndexedSurfacesJsonConverter.hpp
```diff
@@ -99,42 +58,51 @@ void convert(nlohmann::json& jIndexedSurfaces,
...
-template <typename tuple_type, size_t... I>
+/// @param detray if the detray json format is written
+/// @param tList the type list of possible axes
+template <typename... Args>
 void unrollConvert(nlohmann::json& jIndexedSurfaces,
                    const Experimental::SurfaceCandidatesUpdator& delegate,
-                   const tuple_type& axesTuple,
-                   std::index_sequence<I...> /*unused*/) {
-  (convert(jIndexedSurfaces, delegate, std::get<I>(axesTuple)), ...);
+                   bool detray, [[maybe_unused]] TypeList<Args...> tList) {
```
niermann999, line 82:
```suggestion
                   bool detray, TypeList<Args...>) {
```

Plugins/ActSVG/include/Acts/Plugins/ActSVG/IndexedSurfacesSvgConverter.hpp
asalzburger: Indeed, I could - well caught!

niermann999 (dismissed):

Core/include/Acts/Utilities/TypeList.hpp
```diff
@@ -10,15 +10,17 @@
 
 #include <type_traits>
 
-namespace Acts {
+#include <stdlib.h>
```
paulgessinger, line 13: Is this for `size_t`? I think it would be a better idea to use `std::size_t` and include e.g. `cstddef`.

asalzburger: Ah, that was too late yesterday I had #2624 differently in mind.

asalzburger: fixed.

asalzburger:
Sorry @niermann999  - needs re-approving, I fixed clang-tidy and doc build.

niermann999 (dismissed):

asalzburger:
Once more, please.

niermann999 (approved):

niermann999 (approved):

asalzburger:
This ran into `CI/bridge` hard limit again, it seems, the bridge jobs didn't even start.

asalzburger:
Since all other tests ran through, and it has been two days waiting for merging - I do this manually now.

acts-project-service:
## :red_circle: Athena integration test results [1c81427c73376a48608daac71373b5d07cb18f50]
Please investigate the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6493341)!

|status|job|
|------|---|
|:red_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029781) |
|:red_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsBenchmarkWithSpot.sh 8 100](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029620) |
|:red_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsWorkflow.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029619) |
|:red_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateResolvedTracks.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029618) |
|:red_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateTracks.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029617) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateActsCoreSpacePoints.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029616) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateActsSpacePoints.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029615) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateSeeds.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029614) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateOrthogonalSeeds.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029613) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateClusters.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029612) |
|:red_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsPersistifyEDM.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029611) |
|:red_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsKfRefitting.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029610) |
|:green_circle: | [run_ci_tests: python3 ../athena/Tracking/Acts/ActsGeometry/test/ActsExtrapolationAlgTest.py](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029609) |
|:green_circle: | [run_ci_tests: python3 ../athena/Tracking/Acts/ActsGeometry/test/ActsITkTest.py](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029608) |
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029607) |
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029606) |
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029605) |
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34029604) |

PR #2611, opened 2023-11-01
refactor: Remove navigator target intersection

andiwand:
This should only be handled by the `SurfaceReached` aborter


andiwand:

PR #2621, opened 2023-11-02
feat: Introduce propagator stage

andiwand:
Currently it is not possible to discover in which stage of propagation we are in the actors and aborters. By introducing a propagator stage enum, adding it to the state and setting it in the proagator loop this can be used in the actors and aborters.

Pulled out of https://github.com/acts-project/acts/pull/2603

PR #2603, opened 2023-10-31
refactor: Refactor `SurfaceReached` aborter

andiwand:
With these changes I try to decouple the navigator and surface reached aborter further. I don't believe that the navigator has to care about the target surface other than optimization reasons while the surface reached aborter should only care about the surface it is supposed to care about without looking at or touching the navigation state.

blocked by
- https://github.com/acts-project/acts/pull/2621

PR #2625, opened 2023-11-03
refactor!: Rewrite navigation

andiwand:
this is experimental

I discovered too many navigation issues lately and decided to rewrite it in a way that is (hopefully) more robust and which fails fast. In principle I just switched the existing navigation to a single intersection stream rather than having the boundary, layer, surface hierarchy.

incorporates:
- https://github.com/acts-project/acts/pull/2598
- https://github.com/acts-project/acts/pull/2619
- https://github.com/acts-project/acts/pull/2611
- https://github.com/acts-project/acts/pull/2595

blocked by
- https://github.com/acts-project/acts/pull/2603
- https://github.com/acts-project/acts/pull/2621
- https://github.com/acts-project/acts/pull/2681
- https://github.com/acts-project/acts/pull/2691
- https://github.com/acts-project/acts/pull/2694


Tests/UnitTests/Core/Propagator/NavigatorTests.cpp
```diff
@@ -705,46 +677,6 @@ BOOST_AUTO_TEST_CASE(Navigator_target_methods) {
...
-  // Check that the currentSurface is reset to:
-  BOOST_CHECK_EQUAL(BVHState.navigation.currentSurface, nullptr);
-  // No layer has been found
-  BOOST_CHECK_EQUAL(BVHState.navigation.navLayers.size(), 0u);
-  // ACTORS-ABORTERS-TARGET
-  navigator.preStep(BVHState, BVHStepper);
-  // Still no layer
-  BOOST_CHECK_EQUAL(BVHState.navigation.navLayers.size(), 0u);
-  // Surfaces have been found
-  BOOST_CHECK_EQUAL(BVHState.navigation.navSurfaces.size(), 42u);
```
andiwand, line 747: I removed the BVH navigation for now

acts-project-service:
## :red_circle: Athena integration test results [fffac7fd4fee8feec40602ab2d414b508c69039b]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/33686333) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6434882)!

Tests/UnitTests/Core/Geometry/BVHDataTestCase.hpp
andiwand, line 1: I removed the BVH navigation for now

andiwand, line 1: I removed the BVH navigation for now

andiwand:
I looked at the 10 first events in the GSF and 4 of them have some diff
- 3 of them have some very small FP diff which could be explained by stepping difference
- 1 of them finds an additional measurement during filtering which is not found during smoothing

It seems to me that physmon is faster with these changes compared to main which would be really cool if true

open questions from my side:
- how should we proceed with the GSF?
- how do we handle the unit tests I had to adopt?

ping @benjaminhuth @paulgessinger @asalzburger

Tests/UnitTests/Core/Propagator/MultiStepperTests.cpp
andiwand, line 1: this unit test lost a bit its purpose since I removed the overstepping and everything is reachable now

either we remove the test or check step size for +/-

.github/workflows/builds.yml
```diff
@@ -83,11 +83,11 @@ jobs:
           path: ${{ github.workspace }}/ccache
           key: ${{ steps.ccache-restore.outputs.cache-primary-key }}
 
-      - name: Unit tests
-        run: ctest --test-dir build -j$(nproc)
+      #- name: Unit tests
+      #  run: ctest --test-dir build -j$(nproc)
```
andiwand, line 87:
```suggestion
      - name: Unit tests
        run: ctest --test-dir build -j$(nproc)
```

.github/workflows/builds.yml
```diff
@@ -83,11 +83,11 @@ jobs:
...
 
-      - name: Integration tests
-        run: cmake --build build --target integrationtests
+      #- name: Integration tests
+      #  run: cmake --build build --target integrationtests
```
andiwand, line 90:
```suggestion
      - name: Integration tests
        run: cmake --build build --target integrationtests
```

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -393,37 +336,92 @@ class Navigator {
...
+
+      // Estimate the surface status
+      auto surfaceStatus = stepper.updateSurfaceStatus(
+          state.stepping, surface, intersection.index(),
+          state.options.direction, boundaryCheck,
+          state.options.surfaceTolerance, logger());
+
+      if (surfaceStatus == IntersectionStatus::onSurface) {
+        ACTS_ERROR(volInfo(state) << "We are on surface before trying to reach "
+                                     "it. This should not happen. Good luck.");
```
andiwand, line 374: this is one of the errors I log right now @paulgessinger. I guess we have to either make sure it never fires or to reduce the level?

andiwand:
facing a magic bug on `linux_ubuntu_extra (ubuntu2204_cpp20, 20)` which does not want to be discovered. print statements make it disappear and I cannot reproduce them locally, not even with the docker image

Tests/IntegrationTests/PropagationBentTracks.cpp
```diff
@@ -89,50 +89,13 @@ BOOST_AUTO_TEST_CASE(with_boundary_check_no_bfield) {
                     0);
 }
 
-BOOST_AUTO_TEST_CASE(without_boundary_check_no_bfield) {
-  auto navCfg = Acts::Navigator::Config{};
-  navCfg.boundaryCheckLayerResolving = Acts::BoundaryCheck(false);
```
andiwand, line 94: works now without extended bounds

Tests/IntegrationTests/PropagationBentTracks.cpp
```diff
@@ -89,50 +89,13 @@ BOOST_AUTO_TEST_CASE(with_boundary_check_no_bfield) {
...
-                    0);
-  BOOST_CHECK_EQUAL(std::count(xPositions.begin(), xPositions.end(), 2001.0),
-                    0);
-  BOOST_CHECK_EQUAL(std::count(xPositions.begin(), xPositions.end(), 3000.0),
-                    1);
-}
-
-BOOST_AUTO_TEST_CASE(no_boundary_check_with_bfield) {
-  auto navCfg = Acts::Navigator::Config{};
-  navCfg.boundaryCheckLayerResolving = Acts::BoundaryCheck(false);
```
andiwand, line 130: works now without extended bounds

Core/include/Acts/Propagator/detail/AnyIntersection.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+#include "Acts/Geometry/TrackingVolume.hpp"
+#include "Acts/Surfaces/Surface.hpp"
+
+#include <variant>
+
+namespace Acts {
+namespace detail {
+
+class AnyIntersection {
```
andiwand, line 19: why variant instead of virtual inheritance

Core/include/Acts/Propagator/detail/AnyIntersection.hpp
andiwand, line 1: intersection could hold std::any instead of templating

Core/include/Acts/Propagator/detail/NavigationCandidate.hpp
```diff
@@ -0,0 +1,136 @@
...
+    if (std::holds_alternative<LayerObject>(object)) {
+      const auto& layer = std::get<LayerObject>(object);
+      auto intersection = representation->intersect(gctx, position, direction,
+                                                    boundaryCheck, tolerance);
+      return AnyMultiIntersection(LayerMultiIntersection(
+          intersection.intersections(), layer, representation));
+    }
+    if (std::holds_alternative<BoundaryObject>(object)) {
+      const auto& boundary = std::get<BoundaryObject>(object);
+      auto intersection = boundary->surfaceRepresentation().intersect(
```
andiwand, line 55:
```suggestion
      auto intersection = representation.intersect(
```

Core/src/Geometry/Layer.cpp
```diff
@@ -239,22 +210,18 @@ Acts::Layer::compatibleSurfaces(
...
+        return a.object() == b.object() && a.index() == b.index();
       });
 
   // resize to remove all items that are past the unique range
   sIntersections.resize(std::distance(sIntersections.begin(), it),
                         SurfaceIntersection::invalid());
 
-  // sort according to the path length
-  std::sort(sIntersections.begin(), sIntersections.end(),
-            SurfaceIntersection::forwardOrder);
```
andiwand, line 256: we could keep this

Core/src/Geometry/TrackingVolume.cpp
```diff
@@ -532,65 +501,42 @@ Acts::TrackingVolume::compatibleBoundaries(
...
-    for (auto& bsIter : bSurfaces) {
+    for (auto& boundary : boundaries) {
       // Get the boundary surface pointer
-      const auto& bSurfaceRep = bsIter->surfaceRepresentation();
-      ACTS_VERBOSE("Consider boundary surface " << bSurfaceRep.geometryId()
-                                                << " :\n"
-                                                << std::tie(bSurfaceRep, gctx));
-
-      // Exclude the boundary where you are on
-      if (excludeObject != &bSurfaceRep) {
```
andiwand, line 546: we should keep this consistently also for layer resolution

Examples/Python/src/Propagation.cpp
```diff
@@ -127,18 +127,9 @@ void addPropagation(Context& ctx) {
...
-              if (overstepLimit) {
-                return {std::move(bField), overstepLimit.value()};
-              } else {
-                return {std::move(bField)};
-              }
-            }),
-        py::arg("bField"), py::arg("overstepLimit") = std::nullopt);
+        py::init([](std::shared_ptr<const Acts::MagneticFieldProvider> bField)
+                     -> Acts::EigenStepper<> { return {std::move(bField)}; }),
+        py::arg("bField"));
```
andiwand, line 132:
```suggestion
        py::init()
```
?

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -51,147 +55,116 @@ struct NavigationOptions {
...
+    static bool forwardOrder(const NavigationCandidate& aCandidate,
+                             const NavigationCandidate& bCandidate) {
+      return Intersection3D::forwardOrder(
+          aCandidate.intersection.intersection(),
+          bCandidate.intersection.intersection());
+    }
+  };
+
+  using NavigationCandidates =
+      boost::container::small_vector<NavigationCandidate, 24>;
```
andiwand, line 124: `std::vector`

compatible surfaces/layers may wanna push directly

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -393,904 +336,382 @@ class Navigator {
...
-                         << "Navigation stepSize updated to "
-                         << stepper.outputStepSize(state.stepping));
-            return true;
-          }
-        }
-      }
+      state.navigation.currentLayer =
+          intersection.template object<LayerIntersection>();
+
+      reinitializeCandidates(state, stepper);
```
andiwand, line 516: we could only resolve the next 1-2 layers

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -393,904 +336,382 @@ class Navigator {
...
+    const auto& intersection = candidate.intersection;
+    const Surface& surface = *intersection.representation();
 
-    // Navigator post step always starts without current surface
-    state.navigation.currentSurface = nullptr;
+    // Determine the surface status without boundary check.
+    // TODO This will also update the step length which is unnecessary
+    Intersection3D::Status surfaceStatus = stepper.updateSurfaceStatus(
+        state.stepping, surface, intersection.index(), state.options.direction,
+        BoundaryCheck(false), state.options.surfaceTolerance, logger());
```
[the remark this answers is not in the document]
andiwand: ah I remember now @paulgessinger - in principle we can be out of bounds but on surface after the step while we thought we would be in bounds before the step

if this is a sensitive it does not really matter and we will skip it in the next `preStep` call. for layers and boundaries this would cause a renavigation and we might be out of the current volume in an unfortunate edge case. in such a case it would be nice to see a log that we actually hit the surface but out of bounds

I guess this is more of a debugging case then but if the performance hit is not significant I would like to keep it

andiwand:
For some reason this has quite some performance implications. Quickly check against main with ttbar PU 10 in the ODD with Fatras simulation +20% CPU time. I didn't profile this thoroughly but it looks like the impact comes from additional cylinder intersections which is very unexpected to me.

Unfortunately I don't see much optimization opportunities. I tried a few things like reducing layer and boundary lookups which improved performance but at the cost of navigation failures.

Don't really know where to go with this now. Either we buy the performance decrease or we have to close this and wait for the new geometry.

paulgessinger:
This is difficult. I see benefit improving the correctness of the navigation over the performance, but 20% is **significant**.

andiwand:
I broke this one up into a few follow up PRs
- https://github.com/acts-project/acts/pull/2846
- https://github.com/acts-project/acts/pull/2849
- https://github.com/acts-project/acts/pull/2850
- https://github.com/acts-project/acts/pull/2851

I fear I cannot make this fast enough to compete with the existing navigation. I will rather shift my focus on the new geometry and its navigation next year

closing

PR #2643, opened 2023-11-09
docs: Add white paper on the correction term in transport jacobian

beomki-yeo:
@felix-russo @paulgessinger Could you guys transfer the ownership?

Abstract:
```
In the track extension model of ACTS, the covariance of a track is updated
for every surface intersection by applying a jacobian matrix, the composition of
the coordinate transformation and free space transport. In the derivation of the
transport jacobian, it is necessary to consider the path difference coming from the
track displacement, which appears as a direction variation term with respect to the
path length.
The motivation of this white paper is that the author failed in finding any refer-
ence explaining this clearly. (Please let the author know if anyone finds it) There
are bunch of references which calculate this term with an ideal helix model, but
it is not very useful with the runge-kutta based track extension model. There-
fore, in this note, we are deriving the term thoroughly and demonstrate how it is
implemented in the source code.
```

PR #2657, opened 2023-11-13
docs: Add a white paper for line surface jacobian

beomki-yeo:
This PR adds a white paper for line surface jacobian

Abstract:
```
Line surface used for wire measurements or perigee surface is an unique type of surface where the intersections of a track is not defined on a physical plane. 
Due to this trait, the underlying mathematics gets quite complicated especially with coordinate transform jacobian. 
In this white paper, we attempt to derive the jacobians for both of local-to-global and global-to-local transforms.
```

~~The whitepaper contains the calculation of local-to-global transformation jacobian but it doesn't include global-to-local one, which I left for future authors (I didn't manage to solve it)~~

Update: global to local jacobian calculation is updated.


docs/white_papers/figures/correction-for-transport-jacobian.png
andiwand, line 1: somehow some of the pngs changed

beomki-yeo: I guess they are regenerated somehow when I run `docs/white_papers.py pull --github-token ${GITHUB_TOKEN}`. 
That also happened in previous PR #2643

beomki-yeo: I manually reverted the changes in pngs but I think the auto-regeneration of other pngs should be fixed (@paulgessinger  @felix-russo) Not sure if this only happens to me

felix-russo: It happens also to be! I guess that's because we get a new screenshot from each whitepaper when we execute `white_papers.py` -  but I am not sure why this screenshot differs from the old ones...

beomki-yeo: idk maybe metadata?

paulgessinger: Sorry, I missed this. Yes, I think the PNGs are recreated and are not bit-identical, hence new git commits. Could be avoided by comparing the source commit hash, but this is not currently implemented.

paulgessinger:
Sorry, this got held up unfortunately. @beomki-yeo, could you fix the conflict?

beomki-yeo:
I guess it is good to go now

beomki-yeo:
@andiwand @paulgessinger  Any review on this?

andiwand (dismissed):
haven't read the paper yet but I am happy to put it in as is

andiwand:
sorry I just took a quick look and I have a few comments but no need to implement this right away
- the local parameter is hard to read for me as is almost looks like a `1`
- I thought that it might be good make $\phi$ distinguishable in the two representations but I guess it is fine as it is exactly the same thing
- maybe a full vector representation of the global and local params would help
- personally I would make the subscript for the jacobian more explicit like `L->G` and `G->L`
- a full matrix representation of the J would be nice

beomki-yeo:
Thanks I will modify the manuscript except this:

> a full matrix representation of the J would be nice

It's super painful to write a full matrix representation, and it is not very necessary I suppose

beomki-yeo:
@paulgessinger @andiwand can you merge this?

andiwand (dismissed):

paulgessinger (approved):

beomki-yeo:
This got hung up again :/ Shouldn't we have to reduce the threads in gitlab CI as we did for traccc?

paulgessinger:
The job that fails is the most important one, and it's because of OOM. 

I think we increased our build memory consumption again in the recent weeks.

acts-project-service:
## :red_circle: Athena integration test results [bbbbf6b078215c6caf906d14f70bae86e51f8a32]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34718860) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6615938)!

PR #2660, opened 2023-11-13
fix: GSF early stopping now should really stop

benjaminhuth:
Now really stop early (before this did not stop propagation). This now adds an extra aborter. Would be good to refactor the propagation at some point to allow easy stopping in actors...


Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -317,7 +317,8 @@ struct GsfActor {
     // Break the navigation if we found all measurements
     if (m_cfg.numberMeasurements &&
         result.measurementStates == m_cfg.numberMeasurements) {
-      navigator.targetReached(state.navigation, true);
+      ACTS_VERBOSE("Stop navigation because all measurements are found");
+      navigator.navigationBreak(state.navigation, true);
```
andiwand, line 321: I think it would be even cleaner to not go through the navigator state. one option would be to use the shared state mechanism for aborters

andiwand (approved):

asalzburger:
I saw not that the Athena tests fail, but I suppose that's expected?

PR #2658, opened 2023-11-13
refactor!: Consistent naming of GSF mixture reduction and component merging

benjaminhuth:
This PR applies consistent naming for two similar operations on Gaussian mixtures:
* *component merging* is the procedure of merge Gaussian components to a single set of parameters+covariance
* *mixture reduction* is reducing the number of components in a mixture

PR #2661, opened 2023-11-13
feat: Add GSF component momentum cut to stabilize fit

benjaminhuth:
Adds an optional momentum cut, which drops low weight component. This can stabilize the fit and the navigation.

Should go in after #2658

### DIFF
--- a/Core/include/Acts/TrackFitting/detail/GsfActor.hpp
+++ b/Core/include/Acts/TrackFitting/detail/GsfActor.hpp
@@ -20,6 +20,7 @@
 #include "Acts/TrackFitting/GsfError.hpp"
 #include "Acts/TrackFitting/GsfOptions.hpp"
 #include "Acts/TrackFitting/KalmanFitter.hpp"
+#include "Acts/TrackFitting/detail/GsfComponentMerging.hpp"
 #include "Acts/TrackFitting/detail/GsfUtils.hpp"
 #include "Acts/TrackFitting/detail/KalmanUpdateHelpers.hpp"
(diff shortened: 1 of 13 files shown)


Examples/Algorithms/TrackFitting/src/GsfFitterFunction.cpp
```diff
@@ -101,6 +102,7 @@ struct GsfFitterFunctionImpl final : public ActsExamples::TrackFitterFunction {
         &Acts::GainMatrixUpdater::operator()<Acts::VectorMultiTrajectory>>(
         &updater);
 
+    const double momentumCutoff = 500_MeV;
```
andiwand, line 105: can we move that to the struct?

andiwand (approved):

AJPfleger (changes requested):

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -523,11 +508,14 @@ struct GsfActor {
                             const SourceLink& source_link) const {
     const auto& surface = *navigator.currentSurface(state.navigation);
 
+    // This allows to easily project the to weight, filtered pars, filtered cov
```
benjaminhuth, line 511:
```suggestion
    // This allows to easily project the state to weight, filtered pars, filtered cov
```

asalzburger:
We picked a lot of conflicts on that, what's the plan here?

benjaminhuth:
After the physmon with Geant4 is merged, I check again if this PR makes sense or not, otherwise I close it.

PR #2684, opened 2023-11-16
feat: Structured logging

paulgessinger:
This is a proposal to add structured logging to the logging infrastructure.
In the current draft it works like this:

You call a special overload of the `Logger::log` method with a level, message and an arbitrary number of key-value pairs. The key-value pairs are supplied using a special helper type `structured_log_key_v` which encodes the value as an rvalue. I added an operator overload syntax and a literal for this, so you can write

```cpp
logger().log(Logging::INFO, "Message", Logging::slog("key") = "value");
// or
logger().log(Logging::INFO, "Message", "key"_slog = "value");
```

Internally, the key value pairs are added to a json dictionary, but only if the log level is met. This means the conversion doesn't run if the log does not happen, and the arguments should not even be copied at all since they're fully passed through as rvalues.

The output then looks like this:

```
MyLogger INFO STRUCT: {"message":"Message","key":"value"}
```

---
Note: this now relies on having `nlohmann::json` available as a public dependency for the core lib. I think it might be time to consider adding this, since it could enable this kind of stuff, but also potentially simplify our JSON conversion helpers, since we'd be able to just add 

```cpp
friend void to_json(nlohmann::json& j, my_type&)
```

to struct definitions etc.


acts-project-service:
@paulgessinger
🔴 I'm sorry, I couldn't run your command because of an error:
```
No such command 'cc'.
```

paulgessinger:
@krasznaa

paulgessinger:
I'll revisit this some other time.

PR #2685, opened 2023-11-16
refactor: Delete the examples binaries

paulgessinger:
This PR deletes the Examples binaries


paulgessinger:
I'm ok to delete them for v33, yes.

paulgessinger:
There's still the decision remaining what to do with the "hello world" example that one or our unit tests runs (@gagnonlg), and @dimitra97 (I think) added a binary that runs the muon mockup geometry.

How do we best resolve these?

gagnonlg:
You are referring to the no-deps build that runs HelloWorld to check that boost got linked properly, right? IIRC we build the boost filesystem, program_options, and unit_test libraries (the rest is header-only). I guess if we don't do binaries anymore we can drop the program_options dependency? And do we even use boost::Filesystem anymore? If we don't, we can simply replace the HelloWorld check by an arbitrary unit test, perhaps.

AJPfleger (approved):

PR #2701, opened 2023-11-20
feat: adding new IndexedSurfaceMaterial

asalzburger:
This PR adds a new derivate for BinnedMaterial, the `IndexedSurfaceMaterial`. This will allow to run material mapping and write it into a `detray` ready format such that it can be used for the Acts -> detray conversion.

It also adds a complete set of unit testing for this class.

While writing, I realised I was implementing the third time an access method to the grid from 
`Vector2` and `Vector3` to `grid::point_t`, so I centralised this code.

Also, I promoted the GridAxisGenerator to Utilities.
And I added more UnitTests.


niermann999:
Hm, detray actually holds the material directly in the grids, since I thought that material maps cannot be reused between surfaces?
https://github.com/acts-project/detray/blob/fcb1c85ba8bdb8799b2f42a4bba37b1ef35858dd/core/include/detray/materials/material_map.hpp#L18-L22

asalzburger:
> Hm, detray actually holds the material directly in the grids, since I thought that material maps cannot be reused between surfaces? https://github.com/acts-project/detray/blob/fcb1c85ba8bdb8799b2f42a4bba37b1ef35858dd/core/include/detray/materials/material_map.hpp#L18-L22

Oh, I see, I still would like to keep the indexed one, because it will give us the chance to re-use the same material in several bins, for symmetric detectors, this could be a real saver.

asalzburger:
Doc build should be fixed with #2702

Core/include/Acts/Material/IndexedSurfaceMaterial.hpp
```diff
@@ -0,0 +1,136 @@
...
+    // Access via bin0 and bin1
+    Acts::Vector2 lposition =
+        GridAccessHelpers::toLocal(m_grid, bin0, bin1, m_localAccessors);
+    return materialSlab(lposition);
+  }
+
+  /// Scale operator
+  ///
+  /// @param scale is the scale factor applied
+  ISurfaceMaterial& operator*=(ActsScalar scale) final {
```
niermann999, line 106: Just for my benefit: in which cases do we need global scaling for the material?

Core/include/Acts/Material/IndexedSurfaceMaterial.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+  /// @copydoc ISurfaceMaterial::materialSlab(std::size_t bin0, std::size_t bin1) const
+  ///
+  /// The ISurface material class expects bins in [ 0 - n ] x [ 0, m ], we will
+  /// convert into this format, but it is gonna be slow
+  const MaterialSlab& materialSlab(std::size_t bin0,
+                                   std::size_t bin1) const final {
+    // Access via bin0 and bin1
+    Acts::Vector2 lposition =
+        GridAccessHelpers::toLocal(m_grid, bin0, bin1, m_localAccessors);
```
niermann999, line 99: I thought ```toLocal``` should not be used in performance relevant code?

Core/include/Acts/Material/IndexedSurfaceMaterial.hpp
asalzburger: Oh, we just did a little study for the Z mass analysis @noemina  will tell you more about it if you want.

noemina: Yes, this feature revealed to be useful if you want to test by how much tracking performance change when you have different material assumptions in the simulated detector. We can have a chat on this if you are interested :)

Core/include/Acts/Material/IndexedSurfaceMaterial.hpp
asalzburger: Guess we need to profile this. At the end, this will happen ~10-20 times per track, so it should be ok.

Tests/UnitTests/Core/Navigation/NavigationStateUpdatorsTests.cpp
```diff
@@ -128,7 +128,9 @@ class TestAxis : public IAxis {
 
 class MultiGrid1D {
  public:
-  static constexpr std::size_t DIM = 1u;
+  using point_t = std::array<ActsScalar, 1u>;
+
+  static constexpr std::size_t DIM = std::tuple_size_v<point_t>;
```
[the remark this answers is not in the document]
asalzburger: fixed.

asalzburger:
Superseded.

PR #2705, opened 2023-11-21
refactor: Rewrite KF with less code recicle

CarloVarni:
Isolate common pieces of code for the `fit` methods of the KF. The difference between the two `fit` methods is simply the list of surface sequence, and that ends in the KalmanFitterOptions. The rest of the code is the same


asalzburger (approved):
From my side looks ok, let's see what the tests say.

PR #2708, opened 2023-11-21
refactor: KalmanVertexUpdater

felix-russo:
Refactors `KalmanVertexUpdater` and `KalmanVertexTrackUpdater`
- `KalmanVertexTrackUpdater` is moved to `KalmanVertexUpdater`. There was duplicate code and I don't think it makes sense to separate the two.
- Variable names are updated
- Comments are added

Should not change physmon.

blocked by #2713


felix-russo:
Replace by #2713

PR #2715, opened 2023-11-22
refactor!: remove variable useVertexCovForIPEstimation from AMVFinder config

felix-russo:
BREAKING CHANGE: Removes member variable `useVertexCovForIPEstimation` of AMVFinder config


Core/include/Acts/Vertexing/AdaptiveMultiVertexFinder.hpp
```diff
@@ -150,13 +150,6 @@ class AdaptiveMultiVertexFinder {
     // Default fitQuality for constraint vertex in case no beamspot
     // constraint is provided
     std::pair<double, double> defaultConstrFitQuality{0., -3.};
-
-    // Use the full available vertex covariance information after
-    // seeding for the IP estimation. In original implementation
-    // this is not (!) done, however, this is probably not correct.
-    // So definitely consider setting this to true.
-    bool useVertexCovForIPEstimation = false;
```
paulgessinger, line 158: So this was previously already the default, meaning that with the change, if you were using the default implementation, there should be no changes. As far as I can tell, the default in Athena is also `false` and is not configured to `true` anywhere, so I think there should be no output changes in Athena from this.

Tests/UnitTests/Core/Vertexing/AdaptiveMultiVertexFinderTests.cpp
```diff
@@ -197,7 +197,7 @@ BOOST_AUTO_TEST_CASE(adaptive_multi_vertex_finder_test) {
     CHECK_CLOSE_OR_SMALL(recoVtx.position(), expVtx.position, relTol, small);
     CHECK_CLOSE_OR_SMALL(recoVtx.covariance(), expVtx.covariance, relTol,
                          small);
-    BOOST_CHECK_EQUAL(recoVtx.tracks().size(), expVtx.nTracks);
+    // BOOST_CHECK_EQUAL(recoVtx.tracks().size(), expVtx.nTracks);
```
paulgessinger, line 200: Is there a reason to keep these commented out? Otherwise I'd just remove them.

It's not entirely clear to me how these checks relate to the overwritten covariance, could you elaborate?

paulgessinger (commented):
Seems logical, I had one clarifying question on the tests.

Core/include/Acts/Vertexing/AdaptiveMultiVertexFinder.hpp
felix-russo: It seems to have an effect on Athena though since it does not compile, right?

Tests/UnitTests/Core/Vertexing/AdaptiveMultiVertexFinderTests.cpp
felix-russo: I think that just everything changes since I removed the default behavior

felix-russo: I just commented it out to see the impact on athena and on our CI

felix-russo:
Breaks Athena because I removed the member variable
image

What I also really don't understand: Our unit test fails at several places. How come our physmon does not detect anything?

PR #2717, opened 2023-11-22
test: Hacky two way CKF

andiwand:
Not meant for merging but just wanted to put it out so we can easily display and discuss the code and results.

This is a very early attempt to use our existing CKF in both directions.

First we run the CKF with forward propagation with smoothing and then use the smoothed params to run an additional backward propagation for each forward track. Afterwards we can reverse the backward track states and modify the parent index that it becomes a single, longer track.

For now I drafted this in a new algorithm but I think we can extend the existing track finding algorithm to optionally do that and add a config with which direction to start.

Discovered problems
- CKF does not produce tracks without smoothing turned on

Discovered difficulties
- accessing the nth track state is a bit cumbersome

cc @noemina @paulgessinger


andiwand:
closing after demonstration is finished

PR #2539, opened 2023-10-12
refactor: Refactor target direction logic in KF/CKF

andiwand:
Not sure what the use case for this was but to me it seems like the target propagation logic is flawed in KF/CKF. If we propagate to target from the last state the smoothing seems unnecessary. Also since the target surface is a perigee surface for us there are edge cases where the last state has a smaller "distance" to the perigee than the first one.

In general I'd say the user should take such a decision but this is too big of a change at the moment.

discovered with https://github.com/acts-project/acts/pull/2518

PR #2722, opened 2023-11-23
refactor!: Remove smoothing and extrapolation from core CKF

andiwand:
After https://github.com/acts-project/acts/pull/2539 I want to finally refactor the CKF and remove the smoothing and the reference surface targeting from the core algorithm in order to make it more composable and less complicated.

Before the CKF was basically only usable with smoothing turned on otherwise we would not receive any tracks. Now it always runs without smoothing and the user can decide what and how to smooth. Afterwards the user can decide which track state they want to extrapolate to a reference surface if necessary.

IMO this makes the algorithm flow much easier to comprehend and gives the user more flexibility. This also makes an implementation of a two way finding easier as discovered in https://github.com/acts-project/acts/pull/2717

blocked by
- https://github.com/acts-project/acts/pull/2723

---

Users can smooth and extrapolate tracks after the CKF returns them. Extrapolation can be achieved using the `Propagator` while smoothing can be done by the `GainMatrixSmoother`.

For convenience this PR also includes helper functions to smooth and extrapolate single tracks or a whole track container.

Example of smoothing and extrapolating tracks after the CKF

https://github.com/acts-project/acts/blob/98186891d4bdc5b211f6af3e96ca59ace7c315cc/Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp#L172-L189


Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -160,31 +156,10 @@ struct CombinatorialKalmanFilterOptions {
...
-      : geoContext(gctx),
-        magFieldContext(mctx),
-        calibrationContext(cctx),
-        sourcelinkAccessor(std::move(accessor_)),
-        extensions(extensions_),
-        propagatorPlainOptions(pOptions),
-        smoothingTargetSurface(rSurface),
-        multipleScattering(mScattering),
-        energyLoss(eLoss),
-        smoothing(rSmoothing) {}
```
paulgessinger, line 187: There do all these go?

andiwand: to me the constructor seemed overloaded so I reduced it to a minimum as setting params after the constructor looks more explicit and readable to me. but that is only my preference

I guess extensions and source link accessor are mandatory so maybe they should remain. but the interactions flags and target surface and propagation options seem way to specialized to me

AJPfleger:
Please add diff to description as discussed

andiwand:
@osbornjd @plariono do you have feedback on this one?

I tried now to add helper functions for extrapolation and smoothing. This is exercised in the examples framework and I linked a code section to the PR description.

plariono:
Hi @andiwand, the concept I like a lot, the details still have to understand

osbornjd (commented):
How does this affect the KF? Does this mean that the CKF and KF have different behavior in terms of smoothing + extrapolation vs. not? My other question is does the CKF now return the track state by default at the innermost or outermost layer? My understanding is that the `TrackExtrapolationStrategy` allows you to extrapolate to the target (perigee) surface from either the first or last layer, but what state is actually returned from the CKF now?

andiwand:
For the KF I would argue that the smoothing is intended as it should be called with a complete track already. The extrapolation in the KF core I would remove.

The tracks which come out of the finders and fitters there will be not reference surface set. The tracks are only a collections of track states.

The extrapolator can then be configured to start form the inside/outside or automatically deduced by distance

andiwand:
Did not alter the fitters for now. I think this can be done in a second step. The code is ready to be reviewed. I will try to get some performance eval done for this before we get this in

andiwand:
I wondered if you have more feedback or if you are fine with the changes @osbornjd 

Also tagging @pbutti, @timadye and @XiaocongAi as you also use the CKF

osbornjd:
I have no additional feedback, this seems to be well documented in the Examples framework and in this PR. It will obviously be flagged in the release notes as a breaking change so I think that is presumably enough to give everyone a heads up about the change in functionality.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
timadye: I wondered about this when I saw the `extensions` and `propagatorPlainOptions` set in the `Examples` `TrackFindingAlgorithm`. There is the small complication/overhead of initialising those in the constructor and then overwriting them later.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -158,9 +169,27 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
     for (auto& track : tracksForSeed) {
-      // Set the seed number, this number decrease by 1 since the seed number
-      // has already been updated
+      auto smoothingResult =
+          Acts::smoothTrack(ctx.geoContext, tracksTemp, track, logger());
+      if (!smoothingResult.ok()) {
+        ACTS_ERROR("Smoothing for seed "
+                   << iseed << " and track " << track.index()
+                   << " failed with error " << smoothingResult.error());
+        continue;
```
timadye, line 178: Should this count towards `m_nFailedSeeds`? Similarly for `extrapolationResult` below?

Maybe only if all `tracksForSeed` fail, ie. none of them reach the track selector.

Previously, did it count as an error if smoothing or extrapolation failed for **any** track? That wasn't ideal, so progress!

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -77,42 +67,47 @@ struct CombinatorialKalmanFilterTipState {
...
           candidate_container_t& trackStates, bool&, const Logger&)>;
   using BranchStopper = Delegate<bool(const CombinatorialKalmanFilterTipState&,
                                       typename traj_t::TrackStateProxy&)>;
 
-  /// The Calibrator is a dedicated calibration algorithm that allows
-  /// to calibrate measurements using track information, this could be
-  /// e.g. sagging for wires, module deformations, etc.
-  typename KalmanFitterExtensions<traj_t>::Calibrator calibrator;
+  /// The source link accessor
+  SourceLinkAccessor sourcelinkAccessor;
```
timadye, line 94: Sorry, for my understanding, why was the `sourcelinkAccessor` moved out of the main `CombinatorialKalmanFilterOptions` into the `CombinatorialKalmanFilterExtensions`?

I suppose it could live in either, but this seems more core-CKF than an extension.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -403,7 +351,21 @@ class CombinatorialKalmanFilter {
       }
 
       if (!result.filtered &&
-          filterTargetReached(state, stepper, navigator, logger())) {
+          targetReached(state, stepper, navigator, logger())) {
```
paulgessinger, line 354: What does this target reached look for now? Filter target I suppose?

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -209,25 +191,14 @@ struct CombinatorialKalmanFilterOptions {
...
-
-  /// The smoothing target surface
-  const Surface* smoothingTargetSurface = nullptr;
-
-  /// Strategy to propagate to reference surface
-  CombinatorialKalmanFilterTargetSurfaceStrategy
-      smoothingTargetSurfaceStrategy =
-          CombinatorialKalmanFilterTargetSurfaceStrategy::firstOrLast;
+  /// The target surface
+  const Surface* targetSurface = nullptr;
```
paulgessinger, line 195: Maybe clarify in the comment that this is the filter target surface now?

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -1508,18 +1207,16 @@ class CombinatorialKalmanFilter {
...
-      }
-
       auto track = trackContainer.getTrack(trackContainer.addTrack());
       track.tipIndex() = tip;
 
-      const BoundTrackParameters& parameters = it->second;
-      track.parameters() = parameters.parameters();
-      track.covariance() = *parameters.covariance();
-      track.setReferenceSurface(parameters.referenceSurface().getSharedPtr());
+      if (auto it = combKalmanResult.fittedParameters.find(tip);
```
paulgessinger, line 1213: 1. So with this change we create tracks even if we don't have fitted parameters for them? 
2. Is this needed so we can do the smoothing? 
3. The fitted parameters are still at a reference surface, i.e. the perigee? We propagate there from the final surface, right?

I would imagine we still don't want to return tracks at all where the "fitted parameters" are empty, doesn't this mean that the target propagation failed?

Core/include/Acts/Utilities/PropagatorHelpers.hpp
```diff
@@ -0,0 +1,133 @@
...
+StandardStepper buildStandardStepper(
+    std::shared_ptr<const MagneticFieldProvider> bField) {
+  auto stepper = StandardStepper(std::move(bField));
+  return stepper;
+}
+
+/// @brief Build a blind navigator
+///
+/// @return the blind navigator
+BlindNavigator buildBlindNavigator() {
```
paulgessinger, line 52: Blind vs void? Why not just use void?

Core/include/Acts/Utilities/PropagatorHelpers.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include "Acts/Propagator/VoidNavigator.hpp"
+
+#include <memory>
+
+namespace Acts {
+
+using StandardStepper = EigenStepper<>;
+
+using StandardNavigator = Navigator;
+using BlindNavigator = VoidNavigator;
```
paulgessinger, line 27: Should we really alias this?

Core/include/Acts/Utilities/PropagatorHelpers.hpp
paulgessinger, line 1: This seems like fairly trivial aliasing to me. The tool constructors are not meaningfully more complex than the added functions. 
What's the benefit of adding these?

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -0,0 +1,339 @@
...
+/// @param geoContext The geometry context
+/// @param trackContainer The track container
+/// @param track The track to smooth
+/// @param logger The logger
+///
+/// @return The result of the smoothing
+template <typename track_container_t>
+Result<void> smoothTrack(
+    const GeometryContext &geoContext, track_container_t &trackContainer,
+    typename track_container_t::TrackProxy &track,
```
paulgessinger, line 55: You can get the track container from the track proxy with `.container()`, so this function could accept the track proxy by itself. I'm not sure the template deduction would work identically though so you might have to template on `track_proxy_t`.

Aside: I should probably add concepts for track container and track proxies at this point.

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -0,0 +1,339 @@
...
+    const GeometryContext &geoContext, const track_container_t &trackContainer,
+    const Logger &logger = *getDefaultLogger("TrackSmoother", Logging::INFO)) {
+  Result<void> result = Result<void>::success();
+
+  for (const auto &track : trackContainer) {
+    auto smoothingResult =
+        smoothTrack(geoContext, trackContainer, track, logger);
+
+    if (!smoothingResult.ok()) {
+      result = smoothingResult.error();
```
paulgessinger, line 92: Theoretically, there could be multiple errors I guess, do we think this could be a problem due to only the last error being visible?

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -0,0 +1,339 @@
...
+///
+/// @tparam track_proxy_t The track proxy type
+///
+/// @param geoContext The geometry context
+/// @param track The track
+/// @param referenceSurface The reference surface
+/// @param strategy The extrapolation strategy
+/// @param logger The logger
+///
+/// @return The result of the search
```
paulgessinger, line 110: Maybe specify what the result means, because it's an unnamed pair.

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -0,0 +1,339 @@
...
+                   BoundaryCheck(true), s_onSurfaceTolerance)
+        .closest();
+  };
+
+  if (strategy == TrackExtrapolationStrategy::first) {
+    ACTS_VERBOSE("looking for first track state");
+
+    auto first = findFirst();
+    if (!first.ok()) {
+      ACTS_ERROR("no first track state found");
```
paulgessinger, line 169: We'll have to see if these ERRORs cause problems downstream, might have to make the logging behavior configurable again.

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -0,0 +1,339 @@
...
+                   << intersectionLast.pathLength());
+      return std::make_pair(*last, intersectionLast.pathLength());
+    }
+
+    ACTS_ERROR("no intersection found");
+    return Result<std::pair<TrackStateProxy, double>>::failure(
+        TrackExtrapolationError::ReferenceSurfaceUnreachable);
+  }
+
+  throw std::runtime_error("invalid extrapolation strategy");
```
paulgessinger, line 242: Philosophical question if this should also use the return value.

Also (NIT): consider making this a switch over the enum instead of an if-chain.

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -0,0 +1,339 @@
...
+}
+
+/// @brief Extrapolate a track to a reference surface
+///
+/// @tparam track_proxy_t The track proxy type
+/// @tparam propagator_t The propagator type
+/// @tparam propagator_options_t The propagator options type
+///
+/// @param track The track
```
paulgessinger, line 251: Maybe specify that the track is updated in-place.

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -0,0 +1,339 @@
...
+  return Result<void>::success();
+}
+
+/// @brief Extrapolate tracks to a reference surface
+///
+/// @tparam track_container_t The track container type
+/// @tparam propagator_t The propagator type
+/// @tparam propagator_options_t The propagator options type
+///
+/// @param trackContainer The track container
```
paulgessinger, line 303: Again, maybe specify that the tracks are changed in-place.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -159,9 +168,29 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
 
     auto& tracksForSeed = result.value();
     for (auto& track : tracksForSeed) {
-      // Set the seed number, this number decrease by 1 since the seed number
```
paulgessinger, line 162: Consider keeping the comment

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -159,9 +168,29 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
+
+      auto extrapolationResult = Acts::extrapolateTrackToReferenceSurface(
+          track, *pSurface, extrapolator, extrapolationOptions,
+          m_cfg.extrapolationStrategy, logger());
+      if (!extrapolationResult.ok()) {
+        m_nFailedExtrapolation++;
+        ACTS_ERROR("Extrapolation for seed "
+                   << iseed << " and track " << track.index()
+                   << " failed with error " << extrapolationResult.error());
+        continue;
```
paulgessinger, line 189: If this happens, we don't copy to the output track container. Is this intended?

paulgessinger (commented):
❤️ love it

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
andiwand: right now I remember - I though of the case where we run the CKF inwards which needs to terminate at the perigee in case of a cylindrical detector

Core/include/Acts/Utilities/PropagatorHelpers.hpp
andiwand: I was also unsure about this. maybe this can be separated from this PR to have a separate discussion

I feel. like setting up the propagator got quite involved and some factory / helpers would be nice but not sure if this is overgeneralizing stuff

Core/include/Acts/Utilities/TrackHelpers.hpp
andiwand: Yes I guess this is the tradeoff for that interface. I think it is "okay" as we still propagate at least one error and the user will decide what to do with it.

Alternatively we could use error counters or a vector of error results?

Core/include/Acts/Utilities/TrackHelpers.hpp
andiwand: I think right now this can only happen if we do not have any measurements states at all. In this case we should really not try to extrapolate the track. One way to avoid this would be to use a track selector before extrapolation which could also save some CPU time

Core/include/Acts/Utilities/TrackHelpers.hpp
andiwand: My favorite solution would be a compile time check if all cases are handled, but I think this is not possible in C++?

I kind of avoid switch statements for such big blocks of code but this is only a personal preference. Happy to change this if you think this aligns more with the rest of our repo

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
andiwand: My thought here was that all output tracks should be expressed at the reference surface. If a track cannot be expressed at the provided reference it will not be of any use to the user.

Alternatively we could flag the track as failed?

andiwand:
The diff in the track states is coming from the fact that we didn't write them correctly out of the CKF. While branching the smoothed track states will overwrite each other leading to the same states. This is fixed with this PR.

Now it is down to the physmon diff of the track summary. I believe this is due to extrapolation and material effects. Not sure yet how to prove this and to check which one is more correct.

Core/include/Acts/Utilities/TrackHelpers.hpp
timadye: Maybe this is what you meant, and I'm telling you what you well know, but in case it's useful: if you use a `switch` with no `default` and compile with `-Wall` (we do, don't we?), the compiler should tell you if you missed a `case`.

Core/include/Acts/Utilities/TrackHelpers.hpp
timadye: Would it be more useful to keep only the *first* error?

Core/include/Acts/Utilities/TrackHelpers.hpp
andiwand: but this is rather a compiler feature than a language feature right? not sure if we usually rely on this in acts

Core/include/Acts/Utilities/TrackHelpers.hpp
andiwand: good question - do you see a benefit in using the first instead of the last?

andiwand:
the difference in pulls is coming from the fact that the `MaterialInteractor` is doing a bit more than the CKF while we extrapolate to the reference surface. This is the section of the code which makes the difference

https://github.com/acts-project/acts/blob/9eac779e07d8a4a34e94eb4afdb48b3c78a7393f/Core/include/Acts/Propagator/MaterialInteractor.hpp#L100-L104

I think this is done to accumulate position uncertainty due to multiple scattering. The additional covariance on the angles will have an effect on the position covariance but only if it is propagated correctly.

So to me it looks like as this should be more accurate but will also increase CPU a bit.

I briefly looked at the pulls and interestingly they get a bit wider and are moving a bit away from the ones on main. But I am not sure about the significance of this result as I only looked at fast simulation.

andiwand:
Performance seems to be mostly identical. I used this script for measuring https://github.com/andiwand/cern-scripts/blob/f4fb4867912a315ab2d080ba3404427b26b9d8a7/tmp/full_chain_perf.py and a release build of ACTS main vs this branch.

image

slow down factor 1.005

Core/include/Acts/Utilities/TrackHelpers.hpp
timadye: GCC, Clang, and MSVC all have this warning. I think that is a fairly reliable compile-time check, even if it isn't (I believe) formally against the standard.

Core/include/Acts/Utilities/TrackHelpers.hpp
timadye: if it were complaining about the same track, then the first error (eg. filter failure) might lead to the second (eg. smoothing failure). If it is separate seeds/tracks, then I think it is all the same.

andiwand:
I checked the `CKF` code against the `MaterialInteractor` to see how compatible they are.

In the `CKF` we have a helper function to do the material update https://github.com/acts-project/acts/blob/0c753b10f586cf8c17d2d00bb9b33b0a04f5edd3/Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp#L1139

It does basically the same as what is done in the `MaterialInteractor` https://github.com/acts-project/acts/blob/0c753b10f586cf8c17d2d00bb9b33b0a04f5edd3/Core/include/Acts/Propagator/MaterialInteractor.hpp#L58

The essential difference is that the `MaterialInteractor` also deals with volume material (it looks like it is only recording it 🤔  but not relevant here). Apart from that the `MaterialInteractor` will call `stepper.transportCovarianceToCurvilinear(state.stepping);` before the material update if we doe covariance transport, which is the case in the CKF as well. This part is missing in the `CKF` helper function, but I think the reason is that we expect that happening before calling the function. Looking at the code this seems to be the case.

@paulgessinger @noemina this is what I found out reading the code. To me it looks like the same interactions are happening but I didn't prove it by running the code. I think this could be done by using the CKF as an extrapolator but requires more work. What do you think - is this satisfying for now or should we go deeper?

paulgessinger:
@andiwand Do we understand why it does the conversion to curvilinear? Other than that, we can go ahead with this in my opinion.

andiwand:
> @andiwand Do we understand why it does the conversion to curvilinear? Other than that, we can go ahead with this in my opinion.

This is part of our interaction model / how we propagate the MSC. The MSC interaction will add some angle uncertainty to the bound covariance but we should add it to a covariance that is bound close to the material. Otherwise we apply the angle uncertainty to the position uncertainty for the whole propagation path and not just the part after the material.

The same is done during KF/CKF filtering but is basically missing for the extrapolation.

acts-project-service:
## :white_check_mark: Athena integration test results [c1e658a110e854acf6598b06e10aba2aec760ace]

|status|job|report|
|------|---|------|
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517248)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517247)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517246)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517245)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517244)||
|:green_circle: | [test_ActsConversionWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517243)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517242)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517241)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517240)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517239)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517238)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517237)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517236)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517235)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517234)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517233)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517232)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517231)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517230)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517229)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517228)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517227)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517226)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517225)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517224)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517223)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517222) | [:paperclip:](https://acts-athena-outputs.web.cern.ch/test_data18_13TeV_1000evt/pulls/2722/)|
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37517221) | [:paperclip:](https://acts-athena-outputs.web.cern.ch/test_ttbarPU40_reco/pulls/2722/)|

paulgessinger (approved):
Let's go!

PR #2744, opened 2023-11-29
fix: Fix CKF pathlimit abort

andiwand:
The CKF has special path limit treatment during propagation which is overruled by propagation internal path limit aborter in some cases. Here I try to improve this by putting a stub into the propagation


asalzburger (approved):
Very clear - makes sense.

PR #2746, opened 2023-11-29
refactor: Portal fusing and remove shared-from-this

paulgessinger:
The shared-from-this functionality in `Portal` is only needed because `Portal::fuse` uses it to reset the other fuse partner to itself. This PR changes `Portal::fuse` into a static symmetric function which returns a new merged portal.

The input portals are invalidated.

It then removes the shared-from-this functionality from portal, as it is no longer needed.

PR #2753, opened 2023-11-30
fix: set fused portals correctly to former keep volume

asalzburger:
This PR sets the fused portals correctly back to the former keep volumes that relied on keeping the same shared pointer.

This is an update after #2746 which omitted that.

### DIFF
--- a/Core/src/Detector/detail/CylindricalDetectorHelper.cpp
+++ b/Core/src/Detector/detail/CylindricalDetectorHelper.cpp
@@ -959,7 +959,7 @@ Acts::Experimental::detail::CylindricalDetectorHelper::connectInR(
 
     // Fuse containers, and update the attached volumes
     std::shared_ptr<Portal> innerCylinder = containers[ic - 1].find(2u)->second;
-    // Direction is explicitely up
+    // Direction is explicitly addressed with a direction index
     auto innerAttachedVolumes =
         innerCylinder
             ->attachedDetectorVolumes()[Direction(Direction::Backward).index()];


Core/src/Detector/detail/CylindricalDetectorHelper.cpp
```diff
@@ -423,15 +423,12 @@ Acts::Experimental::detail::CylindricalDetectorHelper::connectInR(
...
-      //
-      // In this case the outer cylinder portal of the inner volume is kept,
-      // the inner cylinder of the outer portal goes to waste
-      auto& keepCylinder = volumes[iv - 1]->portalPtrs()[2u];
-      auto& wasteCylinder = volumes[iv]->portalPtrs()[3u];
-      keepCylinder = Portal::fuse(keepCylinder, wasteCylinder);
-      volumes[iv]->updatePortal(keepCylinder, 3u);
+      // Fusing cylinders from inner and outer volume
+      auto innerCylinder = volumes[iv - 1]->portalPtrs()[2u];
+      auto outerCylinder = volumes[iv]->portalPtrs()[3u];
```
paulgessinger, line 428: We used to have named accessor enums for this, didn't we?

paulgessinger (approved):
Thanks!

Core/src/Detector/detail/CylindricalDetectorHelper.cpp
asalzburger: We did, do you suggest we use them?

asalzburger:
Hey @paulgessinger  - I had a spelling error, can you re-approve, please?

paulgessinger (approved):

PR #2755, opened 2023-11-30
chore: Introduce macOS arm64 CI

paulgessinger:


asalzburger:
How many threads did you use for compiling?

paulgessinger:
I'll revisit this some other time.

PR #2760, opened 2023-12-01
refactor: Remove `representation` from `ObjectIntersection`

andiwand:
As the templating of the representing object was rather a burden in https://github.com/acts-project/acts/pull/2625 I propose to remove it all together and compose the intersection if necessary. Here I used a `std::pair` in the `Navigator` for composition for now which may be not optimal but it works as it should for now


asalzburger (approved):
Fine with me!

paulgessinger:
Seems some webhooks were missed. Merging this manually!

PR #2766, opened 2023-12-05
feat: build open data detector from dd4hep in experimental detector schema

asalzburger:
This PR includes everything to build the first version of the open data detector in the `Experimental::Detector` schema, purely by using variant parameters.

For doing this, I needed to introduce:
- place holder proto binning (the range will be set after surface parsing)
- a new method in the `ActsExamples::DD4hepDetector` 
- some more python bindings

Currently, the `thirdparty/OpenDataDetector` is not yet updated, this will be done in a follow-up PR.


dimitra97:
> Hey @dimitra97 - this one is ready for reviewing.

I am starting looking into it!

Core/src/Detector/LayerStructureBuilder.cpp
```diff
@@ -38,6 +39,47 @@ class DetectorVolume;
...
+      const auto& range = extent.range(pb.binValue);
+      // Patch the edges values from the range
+      vmin = range.min();
+      vmax = range.max();
+    } else if (pb.binValue == Acts::binPhi && fullPhiBinning) {
+      vmin = -M_PI;
+      vmax = M_PI;
+      pb.boundaryType = Acts::detail::AxisBoundaryType::Closed;
+    }
+    // Eventually update the edges
```
paulgessinger, line 67:
```suggestion
    // Possibly update the edges
```

Examples/Detectors/DD4hepDetector/src/DD4hepDetector.cpp
```diff
@@ -51,6 +50,32 @@ auto DD4hepDetector::finalize(
...
+        "No DD4hep geometry service configured, can not build "
+        "TrackingGeometry."};
+  }
+
+  auto world = geometryService->geometry();
+  // Build the detector structure
+  Acts::Experimental::DD4hepDetectorStructure dd4hepStructure(
+      Acts::getDefaultLogger("DD4hepDetectorStructure", options.logLevel));
+
+  /// @return a detector, a blue node and a store
```
paulgessinger, line 69: Blue node?

Plugins/DD4hep/include/Acts/Plugins/DD4hep/DD4hepDetectorStructure.hpp
```diff
@@ -39,8 +39,9 @@ class DD4hepDetectorStructure {
   struct Options {
     /// The log level of the tools
     Logging::Level logLevel = Logging::INFO;
-    /// Boolean to flag if blueprint should be written out
-    bool blueprintDot = false;
+    /// Chose to emulate the building and plot a graph instead with the given
+    /// name
+    std::string emulateToGraph = "";
```
paulgessinger, line 44:
```suggestion
    /// Choose to emulate the building and plot a graph instead with the given
    /// name
    std::string emulateToGraph = "";
```

Can you elaborate what "emulate" means here? Do you mean that it will **only** build the graph and not run the construction itself?

The graph writing works in the same blueprint tree that is used for construction, right?

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
```diff
@@ -0,0 +1,131 @@
...
+    int nExpansion =
+        Acts::getAttrValueOr<int>(xmlBinning, std::string(bv + "expansion"), 0);
+    // Auto-range detection
+    bool autoRange = Acts::getAttrValueOr<bool>(
+        xmlBinning, std::string(bv + "autorange"), false);
+    variantParams.set<bool>(bname + "_" + bv + "_autorange", autoRange);
+    variantParams.set<int>(bname + "_" + bv + "_exp", nExpansion);
+    // Equidistant binning detected
+    if (nBins > 0) {
+      // Set the type identificatio
```
paulgessinger, line 29:
```suggestion
      // Set the type identification
```

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
```diff
@@ -0,0 +1,131 @@
...
+      variantParams.set<std::string>(bname + "_" + bv + "_type", "equidistant");
+      // Set the number of bins
+      variantParams.set<int>(bname + "_" + bv + "_n", nBins);
+      // Set min/max paraeter
```
paulgessinger, line 33:
```suggestion
      // Set min/max parameter
```

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
```diff
@@ -0,0 +1,131 @@
...
+    // Gather the bin expansion parameter, expansion of 0 is default
+    int nExpansion =
+        Acts::getAttrValueOr<int>(xmlBinning, std::string(bv + "expansion"), 0);
+    // Auto-range detection
+    bool autoRange = Acts::getAttrValueOr<bool>(
+        xmlBinning, std::string(bv + "autorange"), false);
+    variantParams.set<bool>(bname + "_" + bv + "_autorange", autoRange);
+    variantParams.set<int>(bname + "_" + bv + "_exp", nExpansion);
+    // Equidistant binning detected
+    if (nBins > 0) {
```
paulgessinger, line 28: `nBins == 0` means variable binning? How are the number of variable bins defined? Explicitly from the boundaries it seems?

It might be a bit confusing that `nBins>0` and boundaries are mutually exclusive.

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
```diff
@@ -0,0 +1,131 @@
...
+      std::string del = ",";
+      int end = boundaries.find(del);
+      int ib = 0;
+      // Unit conversion
+      double unitScalar = 1.;
+      if (bv != "phi") {
+        unitScalar = Acts::UnitConstants::mm / dd4hep::millimeter;
+      }
+      // Split and convert
+      while (end != -1) {
```
paulgessinger, line 57:
```suggestion
      while (end != boundaries.end()) {
```

?

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
```diff
@@ -0,0 +1,131 @@
...
+      int nExpansion =
+          getParamOr<int>(bname + "_" + ab + "_exp", dd4hepElement, 0);
+      // Indicate auto-range checking
+      bool autoRange = getParamOr<bool>(bname + "_" + ab + "_autorange",
+                                        dd4hepElement, false);
+      // Equidistant binning
+      if (aType == detail::AxisType::Equidistant) {
+        if (autoRange) {
+          protoBinnings.push_back(
+              Experimental::ProtoBinning(bVal, bType, nBins, nExpansion));
```
paulgessinger, line 99: Ok so for my understanding, proto binning indicates auto binning based on discovered surfaces?

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
```diff
@@ -0,0 +1,131 @@
...
+          // Check for closed phi binning
+          if (bVal == binPhi && (max - min) > 1.9 * M_PI) {
+            bType = Acts::detail::AxisBoundaryType::Closed;
+          }
+          protoBinnings.push_back(Experimental::ProtoBinning(
+              bVal, bType, min, max, nBins, nExpansion));
+        }
+      } else {
+        // Variable binning
+        std::vector<ActsScalar> edges;
```
paulgessinger, line 116: There's no auto binning for variable bin edges?

Plugins/DD4hep/src/DD4hepBlueprint.cpp
paulgessinger, line 1: Sorry, I'm slow again: why is there a specific "DD4hepBlueprint" needed? Is this encapsulated by the structure builders that are connected to the blueprint nodes?

Plugins/DD4hep/src/DD4hepDetectorSurfaceFactory.cpp
```diff
@@ -43,6 +43,11 @@ void Acts::DD4hepDetectorSurfaceFactory::construct(
   ACTS_DEBUG("Recursive search did yield: "
              << cache.sensitiveSurfaces.size() << " sensitive surface(s), "
              << cache.passiveSurfaces.size() << " passive surface(s)");
+
+  // Check for binning patching
+  if (!cache.binnings.empty() && cache.sExtent.has_value()) {
+    ACTS_DEBUG("Checking for patching the binnig structure ... ");
```
paulgessinger, line 49:
```suggestion
    ACTS_DEBUG("Checking for patching the binning structure ... ");
```

Also: what does "patching" mean here?

Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp
```diff
@@ -93,11 +93,15 @@ nlohmann::json toJson(const grid_type& grid) {
 template <typename grid_type>
 nlohmann::json toJsonDetray(const grid_type& grid, bool swapAxis = false) {
   nlohmann::json jGrid;
+  // Get the grid axes & potentiall swap them
```
paulgessinger, line 96:
```suggestion
  // Get the grid axes & potentially swap them
```

Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp
```diff
@@ -109,27 +113,32 @@ nlohmann::json toJsonDetray(const grid_type& grid, bool swapAxis = false) {
...
   // 2D connections
   if constexpr (grid_type::DIM == 2u) {
-    unsigned int iaxis0 = swapAxis ? 1u : 0u;
-    unsigned int iaxis1 = swapAxis ? 0u : 1u;
-    for (unsigned int ib0 = 1u; ib0 <= axes[iaxis0]->getNBins(); ++ib0) {
-      for (unsigned int ib1 = 1u; ib1 <= axes[iaxis1]->getNBins(); ++ib1) {
+    for (unsigned int ib0 = 1u; ib0 <= axes[0u]->getNBins(); ++ib0) {
+      for (unsigned int ib1 = 1u; ib1 <= axes[1u]->getNBins(); ++ib1) {
         typename grid_type::index_t lbin;
+        // Lookup bin - respect swap if it happen for the lookup
```
paulgessinger, line 133:
```suggestion
        // Lookup bin - respect swap if it happened for the lookup
```

Plugins/DD4hep/include/Acts/Plugins/DD4hep/DD4hepDetectorStructure.hpp
asalzburger: Exactly, it will not do the building, because,  e.g. if the building throws and exception, this will show you how/where.

asalzburger: I will clarify in the comments!

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
```diff
@@ -0,0 +1,131 @@
...
+//
+// Copyright (C) 2023 CERN for the benefit of the Acts project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#include "Acts/Plugins/DD4hep/DD4hepBinningHelpers.hpp"
+
+void Acts::DD4hepBinningHelpers::decodeBinning(
```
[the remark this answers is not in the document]
asalzburger: You are absolutely right, I use that only in the `UnitTest` - I will shift the code then there. 
For the ODD, this is not used at all, but variant parameters are used.

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
asalzburger: The `ProtoBinning` class is basically the stripped out (without actual bin calculation) binning description, which we used to do with the `BinUtility`. The latter is not used in the new `Detector` schema, as the actual bin lookup and storage is all done via grids.

One option of the `ProtoBinning` is to define the binning without knowing the edges yet (this practically only makes sense for equidistant), e.g. you want to say: I want 100 material bins on this portal surface in `z`, when you do not know the actual necessary z boundary yet.

Plugins/DD4hep/include/Acts/Plugins/DD4hep/DD4hepDetectorStructure.hpp
asalzburger: I clarified with:

```
    /// If this string is not empty, the detector is not built, but only
    /// a building graph is generated as `.dot` file with the given name
    std::string emulateToGraph = "";
```

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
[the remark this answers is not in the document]
asalzburger: It's moved to the unit tests now.

Examples/Detectors/DD4hepDetector/src/DD4hepDetector.cpp
asalzburger: Left over, removed.

Plugins/DD4hep/src/DD4hepDetectorSurfaceFactory.cpp
asalzburger: I changed all of this to 'automatic range adaption' because that's what it does.

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
asalzburger: Updated.

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
asalzburger: This code is only used in the UnitTests now, it only serves the local testing, so I think I will leave it as such.

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
asalzburger: Updated.

Plugins/Json/include/Acts/Plugins/Json/GridJsonConverter.hpp
asalzburger: Fixed.

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
asalzburger: Updated.

asalzburger:
> > Hey @dimitra97 - this one is ready for reviewing.
> 
> I am starting looking into it!

Ok, thanks. I have integrated Paul's changes & suggestions, so some of the work is already done.

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
paulgessinger: You're right, this is an `std::string`. However, can you use `std::string::npos` anyway? This is in fact `-1`, but it's usually used as the end sentinel value for strings.

Plugins/DD4hep/src/DD4hepDetectorSurfaceFactory.cpp
paulgessinger: That's a great improvement! Thank you!

paulgessinger (commented):
Seems in good shape, thanks for the replies. One more comment on usage of literal `-1` as a sentinel. Should be an easy fix.

Plugins/DD4hep/src/DD4hepBinningHelpers.cpp
asalzburger: Ah, thanks!

asalzburger:
It's updated now.

asalzburger:
> Seems in good shape, thanks for the replies. One more comment on usage of literal `-1` as a sentinel. Should be an easy fix.

Fixed now.

asalzburger:
Hi @paulgessinger - can you have a look please if this is ok for you merge?

paulgessinger (approved):
Let's go!

PR #2769, opened 2023-12-05
refactor: correct phi and theta for periodicity changes in KalmanVertexUpdater

felix-russo:
will break athena


paulgessinger (dismissed):
So we can end up with phi-theta values that are outside $[0,\pi)$ and $[-\pi,\pi]$ and this *just* wraps it back to the valid ranges?

Seems like a no-brainer to me.

I triggered the Athena jobs, let's see what this looks like now.

acts-project-service:
## :red_circle: Athena integration test results [0366adc930ae12f066d8acfeeda2a15fb26a00cb]
Please investigate the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6583617)!

|status|job|
|------|---|
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535653) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsBenchmarkWithSpot.sh 8 100](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535652) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsWorkflow.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535651) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateAmbiguityResolution.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535650) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateResolvedTracks.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535649) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateTracks.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535648) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateActsCoreSpacePoints.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535647) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateActsSpacePoints.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535646) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateSeeds.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535645) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateOrthogonalSeeds.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535644) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateClusters.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535643) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsPersistifyEDM.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535642) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsGSFRefitting.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535641) |
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsKfRefitting.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535640) |
|:green_circle: | [run_ci_tests: python3 ../athena/Tracking/Acts/ActsGeometry/test/ActsExtrapolationAlgTest.py](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535639) |
|:green_circle: | [run_ci_tests: python3 ../athena/Tracking/Acts/ActsGeometry/test/ActsITkTest.py](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535638) |
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535637) |
|:red_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535636) |
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535635) |
|:red_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535634) |
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535633) |
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34535632) |

felix-russo:
Yes that's why we kept this change for the next major version!

As usual, I don't know if this makes athena "worse" or "better". But I think we have enough arguments to get this in

paulgessinger:
@felix-russo I guess this we'll have to argue.

asalzburger:
Hi, I think - as the phi wrapping into range makes totally sense, we should argue for this change.

andiwand:
@paulgessinger and I had a discussion about this last week and we think if we wrap the angle we would also have to modify the chi2 code

@paulgessinger did you check the results after wrapping the difference again?

felix-russo:
FYI: this is also done at a different place of the Kalman Updater; I think it was just forgotten here:
https://github.com/acts-project/acts/blob/2dbf7682aa18e6f3972d4ecfbfbfe27f823490a8/Core/src/Vertexing/KalmanVertexTrackUpdater.cpp#L148-L153

andiwand:
the problem I see is that we later subtract other track params which are not pi/2pi wrapped

https://github.com/acts-project/acts/blob/e283c4645d031769cbf1552edf74811722a2779b/Core/src/Vertexing/KalmanVertexUpdater.cpp#L121-L135

so not correcting them in the first place will result in small differences while correcting the angle needs another difference wrapping otherwise we can end up with big angle differences. I think this is the reason why we see higher chi2 in the ref/obs plot

felix-russo:
Which params are not pi/2pi wrapped? 

We should also wrap `linearizedTrkParams`, but `trkParams` comes from the propagation to the PCA - so there the angles should be in the correct ranges

andiwand:
`linearizedTrackParameters` is not wrapped and `paramDiff = trkParams - linearizedTrackParameters` can also be out of range

this could be the reason why the chi2 gets bigger in the monitoring

not wrapping the angle here might be fine after all because we only check the difference at the end which is hopefully small after the update anyways

felix-russo:
Sorry, but I don’t follow… 
Before:
- newTrkMom was not wrapped
- linearizedTrkParams was not wrapped
- trkParams was wrapped

With the proposed change:
- newTrkMom is wrapped
- linearizedTrkParams is not wrapped
- trkParams is wrapped

Don’t we subtract unwrapped from wrapped in both cases?

Wouldn’t it be best to wrap all three quantities?

felix-russo:
I am not sure if we can ignore the wrapping: 

If `trkParams` (which is wrapped) has an azimuthal angle of `epsilon` and `linearizedTrkParams` (which is not wrapped) has an azimuthal angle of `2pi + epsilon`, they should exactly align.

If we then subtract, however, we get a contribution `~2pi/var(phi)` to the `chi2`. Or am I missing something?

andiwand:
> If we then subtract, however, we get a contribution `~2pi/var(phi)` to the `chi2`. Or am I missing something?

that would be the case after this change but the chi2 contribution should be 0 no? because they actually represent the same direction

felix-russo:
I think the undesired contribution to the chi2 would arise both with and without this change! Or am I missing shtg?

The code is not ready to go in, I am just trying to get on the same page before I update it

andiwand:
in principle I guess so yes but at the same time we see more chi2 outliers in the comparison plot https://github.com/acts-project/acts/pull/2769#issuecomment-1844776152 . the values missing from the peek end up overflowing on the right

felix-russo:
Ok I guess since we don't understand this but the previous version of the code just works better, I suggest to just leave everything as is.

The PR just removes the TODO now, should we get this in? @andiwand @paulgessinger

andiwand (approved):

PR #2782, opened 2023-12-08
refactor: Use common direction transform Jacobian

andiwand:
I found the same code in a few places and there seems to be a bit of simplification opportunity in the math

PR #2781, opened 2023-12-08
refactor!: Remove `CovarianceTransport`

andiwand:
This seems not to be used anywhere other than benchmarks

Breaking as this is part of the public interface

PR #2780, opened 2023-12-08
refactor: Remove `JacobianEngine`

andiwand:
This seems not to be used anywhere and since it is in `detail::` I thought it is safe to be removed

blocked by
- https://github.com/acts-project/acts/pull/2781

PR #2789, opened 2023-12-11
refactor: Reuse `JacobianEngine`

andiwand:
In #2780 I realized that we do not call any of the functions in `JacobianEngine` on the current main branch and most of them are duplicated in `CovarianceEngine`. Here I try to unify the duplicated code again

blocked by


Core/include/Acts/Propagator/detail/JacobianEngine.hpp
```diff
@@ -66,22 +49,23 @@ ActsMatrix<8, 7> anglesToDirectionJacobian(const Vector3& direction);
...
 /// @param [in] geoContext The geometry Context
+/// @param [in] surface Target surface
 /// @param [in] freeParameters Free, nominal parametrisation
 /// @param [in] boundToFreeJacobian Jacobian from bound to free at start
 /// @param [in] freeTransportJacobian Transport jacobian free to free
 /// @param [in] freeToPathDerivatives Path length derivatives for free
-/// parameters
-/// @param [in] surface Target surface
+///        parameters
+/// @param [out] fullTransportJacobian A 6x6 transport jacobian from bound to bound
```
paulgessinger, line 58: Do you expect this be a performance benefit, instead of returning this matrix?

Core/include/Acts/Propagator/detail/JacobianEngine.hpp
```diff
@@ -122,8 +105,7 @@ BoundToFreeMatrix boundToFreeTransportJacobian(
     const FreeMatrix& freeTransportJacobian);
 
 /// @brief This function calculates the full jacobian from a given
-/// free parameterisation to a new curvilinear bound
-/// parameterisation.
+/// free parameterisation to a new curvilinear bound parameterisation.
```
paulgessinger, line 108:
```suggestion
/// free parametrisation to a new curvilinear bound parametrisation.
```
?

Core/include/Acts/Propagator/detail/JacobianEngine.hpp
```diff
@@ -132,23 +114,18 @@ BoundToFreeMatrix boundToFreeTransportJacobian(
...
-FreeToBoundMatrix freeToBoundTransportJacobian(
-    const GeometryContext& geoContext, const FreeVector& freeParameters,
-    const ActsMatrix<7, 8>& directionToAnglesJacobian,
-    const ActsMatrix<8, 7>& anglesToDirectionJacobian,
-    const FreeMatrix& freeTransportJacobian,
-    const FreeVector& freeToPathDerivatives, const Surface& surface);
+void freeToBoundTransportJacobian(const GeometryContext& geoContext,
+                                  const Surface& surface,
+                                  const FreeVector& freeParameters,
+                                  const FreeMatrix& freeTransportJacobian,
```
paulgessinger, line 126: `freeTransportJacobian` is the contraction of the direction to angles and angles to direction jacobians?

Core/include/Acts/Propagator/detail/JacobianEngine.hpp
```diff
@@ -157,34 +134,47 @@ FreeToBoundMatrix freeToBoundTransportJacobian(
...
+///        covariance transport
+///
+/// @param [in, out] freeTransportJacobian The transport jacobian from start
+///        free to final free parameters
+/// @param [in, out] derivatives Path length derivatives of the free, nominal
+///        parameters
+/// @param [in, out] boundToFreeJacobian Projection jacobian of the last bound
+///        parametrisation to free parameters
+/// @param [in] direction Normalised direction vector
+void reinitializeJacobians(FreeMatrix& freeTransportJacobian,
```
paulgessinger, line 174: Can you clarify in the comment what the difference between the two functions here is?

Core/src/Propagator/detail/CovarianceEngine.cpp
```diff
@@ -21,324 +21,105 @@
...
-  boundToFreeJacobian(eFreePos1, eBoundLoc1) = -sinPhi * cosTheta;
-  boundToFreeJacobian(eFreePos2, eBoundLoc1) = sinTheta;
-  boundToFreeJacobian(eFreeTime, eBoundTime) = 1;
-  boundToFreeJacobian.block<3, 2>(eFreeDir0, eBoundPhi) =
-      sphericalToFreeDirectionJacobian(direction);
-  boundToFreeJacobian(eFreeQOverP, eBoundQOverP) = 1;
-}
-}  // namespace
-
-namespace detail {
```
paulgessinger, line 253: Could keep this and remove `detail::` prefixes from the method definitions.

Core/src/Propagator/detail/JacobianEngine.cpp
```diff
@@ -6,22 +6,23 @@
...
+#include "Acts/Utilities/JacobianHelpers.hpp"
 #include "Acts/Utilities/VectorHelpers.hpp"
 
-#include <algorithm>
 #include <cmath>
-#include <utility>
 
 namespace Acts {
 
-namespace detail {
```
paulgessinger, line 22: Could keep this and remove `detail::` prefixes from the method definitions.

paulgessinger:
This breaks Athena compilation. I guess technically, since this is in detail, we would not call it a breaking change, but de-facto it seems to be used publicly. Maybe this indicates we need to move this OUT of detail, and hence mark this breaking?

Core/include/Acts/Propagator/detail/JacobianEngine.hpp
andiwand: I don't know if this would have any performance implications but I copied this over from the `CovarianceEngine` where the code was duplicated and there we used a reference for returning.

Core/include/Acts/Propagator/detail/JacobianEngine.hpp
andiwand: I think this is basically what we get from the stepper. the full jacobian from point A to point B in free coordinates

andiwand:
> This breaks Athena compilation. I guess technically, since this is in detail, we would not call it a breaking change, but de-facto it seems to be used publicly. Maybe this indicates we need to move this OUT of detail, and hence mark this breaking?

puh that is unexpected. I was hoping that this is not used outside. do you know where it is used and how extensively?

Core/include/Acts/Propagator/detail/JacobianEngine.hpp
andiwand: oh I see. I think these direction to angles and vice versa are done only by the global to local now. looks like there was a different free representation once with angles instead of directions

paulgessinger (approved):

PR #2806, opened 2023-12-11
fix: increase initial vertex variance

felix-russo:


andiwand:
thanks for trying!

this seems quite odd to me. I though this caused numerical issues in the past but now I wonder if it might be messing with the number of tracks we try to fit to a single vertex and then it blows up as all tracks might be compatible with the initial vertex now

ultimately this might be the reason why AMVF without beamspot constrained it broken in Athena @paulgessinger as we use 1e8 there too I think

cc @felix-russo

Issue #2810, opened 2023-12-11
`Surface::boundToFreeJacobian` could also use `FreeVector` parameters


andiwand:
It seems inconsistent to use `BoundVector` for `Surface::boundToFreeJacobian` and `FreeVector` for `Surface::freeToBoundJacobian`. One is basically the inverse of the other and could use the same parameters.

https://github.com/acts-project/acts/blob/f7106bc1adc803700eef3690f8dc43046c73bb61/Core/include/Acts/Surfaces/Surface.hpp#L315-L333

https://github.com/acts-project/acts/blob/f7106bc1adc803700eef3690f8dc43046c73bb61/Core/include/Acts/Surfaces/Surface.hpp#L335-L350

Internally we always calculate the free parameters and externally we seem to always have the free parameters at hand.

There is also an extreme example where we do the free to bound twice

https://github.com/acts-project/acts/blob/f7106bc1adc803700eef3690f8dc43046c73bb61/Core/src/Propagator/detail/CovarianceEngine.cpp#L201-L218

PR #2811, opened 2023-12-11
refactor!: Refactor `Surface::boundToFreeJacobian` to use free instead of bound vector

andiwand:
This avoid free->bound->free roundtrips and aligns the interface with `freeToBoundJacobian`

related issues
- https://github.com/acts-project/acts/issues/2810

blocked by
- https://github.com/acts-project/acts/pull/2789


acts-project-service:
## :red_circle: Athena integration test results [dd4e386ef22fe6ca4d43342fa05d56bb110b7115]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34810750) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6633795)!

andiwand:
I hope this does not break Athena anymore

Core/src/Surfaces/Surface.cpp
```diff
@@ -254,14 +254,10 @@ Acts::RotationMatrix3 Acts::Surface::referenceFrame(
 }
 
 Acts::BoundToFreeMatrix Acts::Surface::boundToFreeJacobian(
-    const GeometryContext& gctx, const BoundVector& boundParams) const {
-  // Transform from bound to free parameters
-  FreeVector freeParams =
```
paulgessinger, line 259: Oh ok, so we would convert free to bound, only to then convert back to free, right?

Core/src/Surfaces/Surface.cpp
```diff
@@ -254,14 +254,10 @@ Acts::RotationMatrix3 Acts::Surface::referenceFrame(
...
-      detail::transformBoundToFreeParameters(*this, gctx, boundParams);
-  // The global position
-  const Vector3 position = freeParams.segment<3>(eFreePos0);
+    const GeometryContext& gctx, const FreeVector& parameters) const {
```
paulgessinger, line 257: We could consider naming this `freeParams`
```suggestion
    const GeometryContext& gctx, const FreeVector& freeParams) const {
```

but maybe not needed.

Core/src/Surfaces/DiscSurface.cpp
```diff
@@ -211,29 +209,29 @@ Acts::Vector2 Acts::DiscSurface::localCartesianToPolar(
...
-  // retrieve the reference frame
-  const auto rframe = referenceFrame(gctx, position, direction);
+  const Vector3 direction = parameters.segment<3>(eFreeDir0);
+  // The measurement frame of the surface
+  RotationMatrix3 rframeT =
+      referenceFrame(gctx, position, direction).transpose();
+  // calculate the transformation to local coordinates
+  const Vector3 pos_loc = transform(gctx).inverse() * position;
+  const double lr = perp(pos_loc);
+  const double lphi = phi(pos_loc);
```
paulgessinger, line 223: So this is a trade-off of having to do the trig here, or doing the trig when converting to free. I guess this is not *more* calculation, since we were indeed doing the trig when converting back to free.

Core/src/Surfaces/Surface.cpp
paulgessinger: I suppose one additional point is that the free-to-bound and bound-to-free roundtrip implicitly checks if we're on-surface. Should that still happen, or do we not care?

paulgessinger (commented):
Fine with this overall. Do we expect output changes from this?

Core/src/Surfaces/Surface.cpp
andiwand: I guess this is a question of the contract. I would think the free params are obtained by some bound params at some point so they should be bound. To me it sounds fair to keep this on the callers responsibility that the input is well defined because this is potentially hot code and we might not want to pay the price

Core/src/Surfaces/DiscSurface.cpp
andiwand: right and the disc is the only case where we need the local params. this is the same in `freeToBoundJacobian`

Core/src/Surfaces/Surface.cpp
paulgessinger: We could add an `assert` that only checks in debug?

Core/src/Surfaces/Surface.cpp
paulgessinger: Right, I'm just saying that prior to this PR, we would most often have free params, convert them to bound, call this function, and then convert back to free. With this PR we don't anymore.

Core/src/Surfaces/Surface.cpp
andiwand: checked the repo and `const FreeVector& parameters` seems to be consistently used in the surface code right now

andiwand: I implemented this now - one potential problem with this is that the user might use a different tolerance and then we compare against our globally set tolerance. but maybe our tolerance is the upper bound for such things?

andiwand:
> Fine with this overall. Do we expect output changes from this?

I do not expected physics relevant output changes but due to floating point magic we have seen before it can very well be that we are not bit equal

andiwand:
moved the assert stuff into https://github.com/acts-project/acts/pull/2932 as it required a bit more debugging

Tests/UnitTests/Core/Propagator/BoundToCurvilinearConversionTests.cpp
```diff
@@ -197,6 +197,9 @@ void test_bound_to_curvilinear(const std::vector<TestData> &test_data_list,
       }
     }
 
+    FreeVector free_param_vec = Acts::detail::transformBoundToFreeParameters(
+        *surface, geoCtx, param_vec);
```
AJPfleger, line 201:
```suggestion
    FreeVector freeParamVec = Acts::detail::transformBoundToFreeParameters(
        *surface, geoCtx, paramVec);
```

This file has odd variable naming. Do you know why?
Not sure, if we should keep the file consistent for now and change more of them in a later go?

andiwand: I think the code is just outdated

AJPfleger (approved):

Issue #2812, opened 2023-12-11
Redundancy of bound and curvilinear state


andiwand:
While refactoring the [`CovarianceEngine`](https://github.com/acts-project/acts/blob/main/Core/include/Acts/Propagator/detail/CovarianceEngine.hpp) and the [`JacobianEngine`](https://github.com/acts-project/acts/blob/main/Core/include/Acts/Propagator/detail/JacobianEngine.hpp) I noticed that we duplicate quite some code for handling bound and curvilinear states as they are very similar.

I am not sure if that is based on performance reasons or if this is just grown historically.

In any case I think we could remove the duplication by merging the concept or we could split the curvilinear parameters from the bound parameters which would remove the dependency on a shared pointer surface under the hood.

Ultimately this depends on the performance implications of merging them I guess since this would be the most general and easiest solution IMO.

beomki-yeo:
Curvilinear state is just a special case of bound state where the surface normal vector is direction. 

I suppose we can make our interface smarter with surface normal vector argument. (ex) put a surface normal vector for bound state and put a track direction for curvilinear state)

paulgessinger:
Just want to add that in curvilinear the plane rotation around the normal is also fixed.

andiwand:
After having a deeper look at this it seems to me that this is an optimization step. We are basically doing the same math but with less expressions in case of the curvilinear surface.

PR #2814, opened 2023-12-12
refactor: Introduce `CurvilinearSurface`

andiwand:
I noticed that the concept of a curvilinear surface is a quite spread over the codebase and I wanted to improve this. Here I introduce a surface like class which does not actually inherit from `Acts::Surface` but shares some of the functionality. This way the jacobians are put in a single place and can be used from somewhere else in an expressive fashion.

Afterwards it might make sense to let the `CurvilinearTrackParameters` depend on this special surface instead of a `PlaneSurface` or to merge them with `BoundTrackParameters` completely.

related issues
- #2812 

blocked by
- https://github.com/acts-project/acts/pull/2789
- https://github.com/acts-project/acts/pull/2811


Core/include/Acts/Surfaces/CurvilinearSurface.hpp
```diff
@@ -0,0 +1,185 @@
...
+#include <iomanip>
+#include <ios>
+#include <ostream>
+#include <string>
+
+namespace Acts {
+
+/// @brief Utility class for curvilinear surfaces
+///
+class CurvilinearSurface {
```
paulgessinger, line 26: This could potentially use the surface concept that I added a while back

andiwand: nice! I took a look and I fear it requires more functions than necessary here

acts-project-service:
## :red_circle: Athena integration test results [fd7db45e28897357497e5b6d81bd16943d801ea7]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37715848) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7168288)!

Core/src/Surfaces/CurvilinearSurface.cpp
```diff
@@ -0,0 +1,132 @@
...
+  return Surface::makeShared<PlaneSurface>(transform());
+}
+
+std::shared_ptr<Surface> CurvilinearSurface::surface() const {
+  return planeSurface();
+}
+
+std::ostream& operator<<(std::ostream& os, const CurvilinearSurface& surface) {
+  surface.toStream(os);
+  return os;
```
krasznaa, line 129:
```suggestion
  return surface.toStream(os);
```

Core/src/Surfaces/CurvilinearSurface.cpp
```diff
@@ -0,0 +1,132 @@
...
+  sl << std::setiosflags(std::ios::fixed);
+  sl << std::setprecision(4);
+  sl << std::boolalpha;
+  sl << "Curvilinear Surface" << std::endl;
+  sl << "     Center position  (x, y, z) = (" << m_position.x() << ", "
+     << m_position.y() << ", " << m_position.z() << ")" << std::endl;
+  sl << "     Direction  (x, y, z) = (" << m_direction.x() << ", "
+     << m_direction.y() << ", " << m_direction.z() << ")" << std::endl;
+  sl << std::setprecision(-1);
+  sl << std::noboolalpha;
```
krasznaa, line 109: I'm going to be **very** pedantic here... :thinking: This is the type of code that makes/made stream based printouts a pain in C\+\+. :frowning: That one has to be very careful not to change the global state of the stream passed to a function like this one.

In this case the received stream can very much walk away from this function in a "different state" than it arrived with. :frowning: The most pragmatic solution to which is to:
  - Construct the formatted output using an `std::ostringstream`;
  - Put the contents of that stream into the stream passed to this function.

I.e. you might as well remove this function, and just provide the `toString()` function. (Implemented on its own this time around.) Then you don't need to worry about leaking stream modifiers. :thinking:

If you do that, of course the "global" output operator would become:

```c++
std::ostream& operator<<(std::ostream& os, const CurvilinearSurface& surface) {
  out << surface.toSting();
  return out;
}
```

krasznaa (commented):
Since your exchange on Mattermost raised my interest, now deal with my technical (and in the grand scheme of things irrelevant) comments. :stuck_out_tongue:

Core/src/Surfaces/CurvilinearSurface.cpp
andiwand: Right, I was also worried about changing the state of the streamer but went blind since I just copied the code from https://github.com/acts-project/acts/blob/b0c64a07b4603bad9e9df94bf0b369bbbf795ae8/Core/src/Surfaces/Surface.cpp#L198-L221 (looks not exactly like this so might actually have been a different source) and wanted to be done with it 😄 

Using a string would be one option. Restoring the stream to the original state another I guess? Do you know if there are convenience functions for that or do we need to remember the properties we change before setting them?

I will open an issue about this.

andiwand: https://github.com/acts-project/acts/issues/3079

krasznaa: It wouldn't be impossible to use a combination of:
  - https://en.cppreference.com/w/cpp/io/ios_base/flags
  - https://en.cppreference.com/w/cpp/io/ios_base/setf
  - https://en.cppreference.com/w/cpp/io/ios_base/unsetf

But the last time I tried this, I gave up in the end. As the code became just too finicky. It was just easier to create a temporary stream whose flags I could modify however I wanted to. :thinking:

Core/include/Acts/Propagator/detail/JacobianEngine.hpp
AJPfleger, line 1: Do you know why this needed to be removed and why we don't want to remove it anymore?

Tests/UnitTests/Core/Navigation/DetectorNavigatorTests.cpp
AJPfleger, line 1: Do you understand what caused the additional includes of `Acts/Surfaces/PlaneSurface.hpp` in this and other tests?

Core/include/Acts/Propagator/detail/JacobianEngine.hpp
andiwand: I am not sure why I put this comment anymore but I do think we want to keep this function by now

Tests/UnitTests/Core/Navigation/DetectorNavigatorTests.cpp
andiwand: We use `PlaneSurface` in the test but it is not included. I switched to a forward declare in the curvilinear params which made the downstream include chains fail

Tests/UnitTests/Core/Surfaces/CurvilinearSurfaceTests.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "Acts/Definitions/Algebra.hpp"
+#include "Acts/Definitions/TrackParametrization.hpp"
+#include "Acts/Surfaces/CurvilinearSurface.hpp"
+#include "Acts/Tests/CommonHelpers/FloatComparisons.hpp"
+#include "Acts/Utilities/VectorHelpers.hpp"
+
+#include <cmath>
+
+namespace Acts {
+namespace Test {
```
AJPfleger, line 20:
```suggestion
namespace Acts::Test {
```

Tests/UnitTests/Core/Surfaces/CurvilinearSurfaceTests.cpp
```diff
@@ -0,0 +1,77 @@
...
+  CHECK_CLOSE_REL(c2fJacobian(eFreeDir0, eBoundTheta), cosTheta * cosPhi, 1e-5);
+  CHECK_CLOSE_REL(c2fJacobian(eFreeDir1, eBoundPhi), sinTheta * cosPhi, 1e-5);
+  CHECK_CLOSE_REL(c2fJacobian(eFreeDir1, eBoundTheta), cosTheta * sinPhi, 1e-5);
+  CHECK_CLOSE_REL(c2fJacobian(eFreeDir2, eBoundTheta), -sinTheta, 1e-5);
+  // Q/P parameter: stays as is
+  CHECK_CLOSE_REL(c2fJacobian(eFreeQOverP, eBoundQOverP), 1, 1e-5);
+}
+
+}  // namespace Test
+}  // namespace Acts
```
AJPfleger, line 77:
```suggestion
}  // namespace Acts::Test
```

AJPfleger (approved):

andiwand:
I will hold this back until I can test this in Athena

PR #2818, opened 2023-12-12
fix: Fix Geant4 `SensitiveSurfaceMapper`

andiwand:
It seems that the logic of the sensitive surface mapper was flawed.

I fixed that and tried to make the expression more explicit


andiwand:
cc @janheinri

Examples/Algorithms/Geant4/src/SensitiveSurfaceMapper.cpp
```diff
@@ -71,11 +71,16 @@ void ActsExamples::SensitiveSurfaceMapper::remapSensitiveNames(
...
+  bool isMappedMaterial =
       std::find(m_cfg.materialMappings.begin(), m_cfg.materialMappings.end(),
-                volumeMaterialName) != m_cfg.materialMappings.end() ||
+                volumeMaterialName) != m_cfg.materialMappings.end();
+  bool isMappedVolume =
       std::find(m_cfg.volumeMappings.begin(), m_cfg.volumeMappings.end(),
-                volumeName) != m_cfg.volumeMappings.end()) {
+                volumeName) != m_cfg.volumeMappings.end();
+
+  if (isSensitive && (isMappedMaterial || isMappedVolume)) {
```
janheinri, line 83: I might be completely wrong here, but wouldn't this mean that surfaces are never mapped if g4SensitiveDetector == nullptr?
Wouldn't you want this to run in this case?
Considering the message at the end maybe it has to be (isSensitive || isMappedMaterial || isMappedVolume) after all.
I think I also got confused along the way

andiwand, line 83:
```suggestion
  if (isSensitive || isMappedMaterial || isMappedVolume) {
```

andiwand: right :+1:

paulgessinger (approved):

PR #2821, opened 2023-12-12
chore: IWYU fixes

andiwand:
Another round of IWYU fixes

Again I had to do quite some manual work after IWYU


andiwand:
Sure I think this one will conflict a lot in the meantime but at the same time it is not really relevant. I just realized that it still requires a lot of manual tweaking which makes this tool questionable in the state it is right now

PR #2825, opened 2023-12-13
test: Add bound to bound roundtrip covariance tests

paulgessinger:
Currently this tests plane to plane and plane to perigee roundtrip tests.


paulgessinger:
@andiwand @beomki-yeo

Tests/UnitTests/Core/Propagator/CovarianceEngineTests.cpp
```diff
@@ -164,5 +171,186 @@ BOOST_AUTO_TEST_CASE(covariance_engine_test) {
...
+      Acts::BoundVector parOut =
+          Acts::detail::transformFreeToBoundParameters(freePars, srfB, gctx)
+              .value();
+
+      Acts::BoundToFreeMatrix boundToFreeJacobian =
+          srfA.boundToFreeJacobian(gctx, parIn);
+      Acts::FreeMatrix freeTransportJacobian = FreeMatrix::Identity();
+
+      FreeVector freeToPathDerivatives = FreeVector::Zero();
+      freeToPathDerivatives.head<3>() = freePars.segment<3>(eFreeDir0);
```
beomki-yeo, line 211: Can you put Cross(dir, any_vector) for 4th - 6th elements and see the test is OK?

If I am correct, the test should pass.

Tests/UnitTests/Core/Propagator/CovarianceEngineTests.cpp
```diff
@@ -164,5 +171,186 @@ BOOST_AUTO_TEST_CASE(covariance_engine_test) {
...
+  GeometryContext gctx;
+
+  std::size_t n = 100;
+
+  std::mt19937 rng{42};
+
+  std::uniform_real_distribution<double> locDist(-5., 5.);
+  std::uniform_real_distribution<double> globDist(-50., 50.);
+  std::uniform_real_distribution<double> angleDist(-2 * M_PI, 2 * M_PI);
+  std::uniform_real_distribution<double> unitDist(0, 1);
```
andiwand, line 185: `BOOST_DATA_TEST_CASE` can be very convenient for these things

andiwand (commented):
I like it :+1:

Tests/UnitTests/Core/Propagator/CovarianceEngineTests.cpp
paulgessinger: Right, but this here runs a number of tests for the different scenarios, so I'd have to add a **ton** of random numbers. This way I only need to define a handful of distributions.

Tests/UnitTests/Core/Propagator/CovarianceEngineTests.cpp
paulgessinger: What's `any_vector` here?

beomki-yeo: anything that is neither zero nor parallel to the direction, so that Cross(dir, anyvector) is not zero vector

beomki-yeo: We can start with ones with orthogonal to the directino though

paulgessinger: If I add

```cpp
freeToPathDerivatives.segment<3>(4) = freePars.segment<3>(eFreeDir0).cross(Vector3::Random());
```

I get failures.

beomki-yeo: For which cases and for which element? How big is the difference?

beomki-yeo: You might need to put the same vector for both directions

beomki-yeo: It is important because the `freeToPathDerivative` is (t_x,t_y,_tz,0, (txb)_x, (txb)_y, (txb)_z, 0) for both perigee and bound surface in ACTS implementation

paulgessinger: So that would mean the roundtrip works because we assume B is zero, but it's not clear if the actual covariance in the target plane surface or perigee is correct. 

Is there a good way to assert the target correctness? Other than comparing with an analytically calculated covariance?

beomki-yeo: The way to assert the target correctness is comparing with numerical differentiation.

We can numerically obtain "J_{G->L} * Correction * J_T * J_{L->G}" which is the full jacobian from initial surface to target surface. (In your setup J_T is identity matrix.)

It is simply $$\frac{\partial x^{final}_j}{\partial x^{initial}_i} = \frac{x^{final}_j(x^{initial}_i +h) - x^{final}_j(x^{initial}_i)}{h}$$

Setting h~0.0001 works OK in my experience

I have been working on this in detray library - you can check here:
https://github.com/acts-project/detray/pull/610/files
Note that it is for the target correctness with two displaced surfaces.

---------------------------------------------------------------------------


Back to the main discussion,


> So that would mean the roundtrip works because we assume B is zero, 

This is not what I meant. 
we must get the consistent result even with non-zero bfield components, which are going to be cancelled anyway.
If you are not getting the consistent result with `cross(direction, bfield)` term, there could be a bug in the test.

The test of this PR has been OK without bfield because the free space transport Jacobian is just a Identity matrix. 
But if you are trying to do the same round-trip test with two "displaced" surfaces, not intersecting, it's not going to work anymore.

I can prove why ACTS implementation must be able to give you the same bound covariance after round trip for two "intersecting" surface, with `cross(direction,bfield)` term:
https://github.com/beomki-yeo/WhyRoundTripJacobianIsIdentityMatrix/releases/download/v1.0/main.pdf

paulgessinger: @beomki-yeo Ok, I added a fixed but random B field vector, and indeed that seems to succeed in all cases.

Tests/UnitTests/Core/Propagator/CovarianceEngineTests.cpp
```diff
@@ -164,5 +171,210 @@ BOOST_AUTO_TEST_CASE(covariance_engine_test) {
...
+
+        Acts::BoundToFreeMatrix boundToFreeJacobian =
+            srfA.boundToFreeJacobian(gctx, parIn);
+        Acts::FreeMatrix freeTransportJacobian = FreeMatrix::Identity();
+
+        FreeVector freeToPathDerivatives = FreeVector::Zero();
+        freeToPathDerivatives.head<3>() = freePars.segment<3>(eFreeDir0);
+
+        freeToPathDerivatives.segment<3>(eFreeDir0) =
+            bField.cross(freePars.segment<3>(eFreeDir0));
```
paulgessinger, line 221: @beomki-yeo I'm setting the direction derivative to $B\times T$ here.

beomki-yeo: Should be OK for this test

Tests/UnitTests/Core/Propagator/CovarianceEngineTests.cpp
beomki-yeo: OK just make sure that you don't hit the divergence conditions with random variables 😃

beomki-yeo: Hmm maybe my derivations are wrong where I assumed J_{L -> G} J_{G -> L} is identity matrix at the final surface, which is not true. There must be more magic in the calculations :p ...

paulgessinger: Ok @beomki-yeo, I've done this now, but here it gets weird:

- With the direct jacobian from A->B I get the same covariance as when going trough J_{L->G} J_{G->L} **without the free-to-path derivatives!**
- With the direct jacobian, I **cannot roundtrip** A->B->A, it just gives different covariances back at A.

What's going on here?

beomki-yeo: It's very interesting. I just have a guess here. 

It might be easy to explain with an image:



In the first case, I think you are calculating direct jacobians by projecting the displaced parameters onto the final surface vertically (A point in the image). This won't need a `free-to-path-derivative` 

If you want to include the `free-to-path-derivative` in jacobian calculation, you might need to project the parameter with the track direction ($\mathbf{t}$) (B point in the image). 

The second case can be explained in the same way. I believe you are getting A -> A' point during the direct jacobian calculation. This won't give you the identity matrix:
$$\frac{u(u_0+h) - u(u_0)}{h} \neq 1 $$






In summary, I believe the `localTolocal` in direction jacobian function is not being used correctly :thinking:

paulgessinger: Ok, so with the projection I get somewhat matching covariances. 

I think I'm observing some numerical instability: the top left 2x2 corners match at 10^-5, but the rest of the matrix, including the rest of the diagonal are only up to within 0.1 of the nominal calculation.

I think for the purpose of this test this is good enough. The main goal is to verify we can roundtrip covariances, with the direct jacobian being more of a sanity check.

@beomki-yeo @andiwand

beomki-yeo: Ok for the last point, i suggest setting the cross(t×b) terms to zero because we are only projecting with t. (No propation under b field)

beomki-yeo: Everything else looks good to me

Core/include/Acts/Propagator/detail/CovarianceEngine.hpp
```diff
@@ -138,5 +138,18 @@ void transportCovarianceToCurvilinear(BoundSquareMatrix& boundCovariance,
                                       BoundToFreeMatrix& boundToFreeJacobian,
                                       const Vector3& direction);
 
+/// Convert bound track parameters to another bound surface.
+/// @pre The @p targetSurface must intersect with the surface attached to
+///      @p boundParameters, and the paremeters must be on-surface on the
```
paulgessinger, line 143:
```suggestion
///      @p boundParameters, and the parameters must be on-surface on the
```

Tests/UnitTests/Core/Propagator/CovarianceEngineTests.cpp
paulgessinger: > Ok for the last point, i suggest setting the cross(t×b) terms to zero because we are only projecting with t. (No propation under b field)

I mean technically I could also set up an call a propagator, so that I get the correct BField projection from the propagation.

paulgessinger: Ok so with the numerical propagation, the resulting covariances are completely different w.r.t. to the nominal calculation.

Tests/UnitTests/Core/Propagator/CovarianceEngineTests.cpp
```diff
@@ -164,5 +183,324 @@ BOOST_AUTO_TEST_CASE(covariance_engine_test) {
...
+  CHECK_CLOSE_ABS(parA, parA2, 1e-9);
+  CHECK_CLOSE_COVARIANCE(covA, covA2, 1e-9);
+
+  auto prop = makePropagator(bField);
+  BoundMatrix J = directJacobian(prop, parA, *planeSurfaceA, *perigee);
+  BoundMatrix covC = J * covA * J.transpose();
+  CHECK_CLOSE_ABS((covB.template topLeftCorner<2, 2>()),
+                  (covC.template topLeftCorner<2, 2>()), 1e-7);
+  CHECK_CLOSE_ABS(covB.diagonal(), covC.diagonal(), 1e-7);
+  CHECK_CLOSE_ABS(covB, covC, 1e-7);
```
beomki-yeo, line 502: Is this line comparing all elements of the covariance?

paulgessinger: Yes.

Tests/UnitTests/Core/Propagator/CovarianceEngineTests.cpp
```diff
@@ -164,5 +183,324 @@ BOOST_AUTO_TEST_CASE(covariance_engine_test) {
...
+      boundToBound(parB, covB, *perigee, *planeSurfaceA, bField);
+  CHECK_CLOSE_ABS(parA, parA2, 1e-9);
+  CHECK_CLOSE_COVARIANCE(covA, covA2, 1e-9);
+
+  auto prop = makePropagator(bField);
+  BoundMatrix J = directJacobian(prop, parA, *planeSurfaceA, *perigee);
+  BoundMatrix covC = J * covA * J.transpose();
+  CHECK_CLOSE_ABS((covB.template topLeftCorner<2, 2>()),
+                  (covC.template topLeftCorner<2, 2>()), 1e-7);
+  CHECK_CLOSE_ABS(covB.diagonal(), covC.diagonal(), 1e-7);
```
paulgessinger, line 501:
```suggestion
```

beomki-yeo (dismissed):
This is a nice test and we should get this in.

Tests/UnitTests/Core/Propagator/CovarianceEngineTests.cpp
beomki-yeo: > Yes. I added the other lines so I could apply different tolerances but now it might not be needed anymore.

Aha ok. Then it is OK to have them if you like

Tests/UnitTests/Core/Propagator/CovarianceEngineTests.cpp
```diff
@@ -164,5 +183,321 @@ BOOST_AUTO_TEST_CASE(covariance_engine_test) {
...
+  out[eBoundTime] = local[eBoundTime];
+  return out;
+}
+
+propagator_t makePropagator(const Vector3& bField) {
+  return propagator_t{EigenStepper<>{std::make_shared<ConstantBField>(bField)},
+                      detail::VoidNavigator{}};
+}
+
+BoundMatrix directJacobian(const propagator_t& prop, const BoundVector& parA,
```
andiwand, line 227:
```suggestion
BoundMatrix numericalBoundToBoundJacobian(const propagator_t& prop, const BoundVector& parA,
```

paulgessinger: Ok. Can you reapprove @andiwand?

andiwand (approved):
awesome!

acts-project-service:
## :white_check_mark: Athena integration test results [e867b1f2504bfd6b8dcda14a9da4b675ea1f2194]

|status|job|report|
|------|---|------|
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824070)||
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsBenchmarkWithSpot.sh 8 100](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824069)||
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsWorkflow.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824068)||
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateAmbiguityResolution.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824067)||
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateResolvedTracks.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824066)||
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateTracks.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824065)||
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateActsCoreSpacePoints.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824064)||
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateActsSpacePoints.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824063)||
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateSeeds.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824062)||
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateOrthogonalSeeds.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824061)||
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsValidateClusters.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824060)||
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsPersistifyEDM.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824059)||
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsGSFRefitting.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824058)||
|:green_circle: | [run_ci_tests: ../athena/AtlasTest/CITest/test/ActsKfRefitting.sh](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824057)||
|:green_circle: | [run_ci_tests: python3 ../athena/Tracking/Acts/ActsGeometry/test/ActsExtrapolationAlgTest.py](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824056)||
|:green_circle: | [run_ci_tests: python3 ../athena/Tracking/Acts/ActsGeometry/test/ActsITkTest.py](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824055)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824054)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824053)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824052)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824051)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824050)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824049)||
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34824048)||

PR #2829, opened 2023-12-14
feat!: Wire time to spacepoints and seeds

andiwand:
Currently our spacepoints and seeds do not carry any time information from a potential measurement to the initial track parameters.

In this PR I try to wire optional time information originating from the measurement, carried through spacepoints into seeds and finally into initial track params.

One caveat is that we use bound parameters in between which cannot tell if time is actually measured or not. My intermediate solution was to rely on the projected covariance entry to be 0 in this case.


Core/include/Acts/Seeding/InternalSpacePoint.hpp
```diff
@@ -27,7 +29,9 @@ class InternalSpacePoint {
   InternalSpacePoint(std::size_t index, const SpacePoint& sp,
                      const Acts::Vector3& globalPos,
                      const Acts::Vector2& offsetXY,
-                     const Acts::Vector2& variance);
+                     const Acts::Vector2& variance,
+                     const std::optional<float>& globalTime,
```
paulgessinger, line 33: Not sure the optional cover const ref is with it over just passing the optimal by value here.

paulgessinger (dismissed):
Cool. Is the statement in the description still the case? I are in principle the river is passed through as optional and the parameter estimation uses this information directly.

paulgessinger:
Physmon looks reasonable I think. Some other tests are still failing, otherwise I'd say this is good to go.

paulgessinger (approved):

PR #2833, opened 2023-12-15
feat: Update ODD to v3.0.3 and small ODD fixes

asalzburger:
This updates the ODD to v3.0.3, which brings in a fix:

- the long strip sensors were different in front and back side
- the geometric config file works (was not working)

It fixes the problem of: 
- `Examples/Scripts/Python/geoemtry.py` crashing on default because of missing output folder

And enables the building of:
- `Examples/Scripts/Python/detector_creation.py`

*Note: small performance chances are explected*


asalzburger:
As expected, this introduces fluctuations in the result due to the slight change of LongStrip description (fix that the sensors are identical).

andiwand:
looked at the physmon output - looks uncontroversial to me

asalzburger:
> looked at the physmon output - looks uncontroversial to me

Thanks, I will update the references then.

andiwand:
superseded by https://github.com/acts-project/acts/pull/2992 @asalzburger ?

PR #2835, opened 2023-12-15
refactor!: Remove now-obsolete property `skipZMiddleBinSearch`

CarloVarni:
This needs https://github.com/acts-project/acts/pull/2718 to be in.

As quoted there: The grid iterator will be able to iterate on whatever sequence of bins you like.

In other words... for a usual seed finding process we have:

`zBinsCustomLooping = [1, 2, 3, 4, 11, 10, 9, 8, 6, 5, 7]`
but for fast tracking you can simply provide a shorter vector. For instance:

`zBinsCustomLooping = [11, 10, 9, 8, 6, 5, 7]`
much simpler imho.

PR #2838, opened 2023-12-16
feat!: Generic grid bin finder

CarloVarni:
This will complete the adaptation of the grid to an N-dimentional case. The user is now able to iterate on any axis as they like without the need to rotate their geometry. 

Next step is to see how the seed finding routine can be adapted to be more general

This PR is also a bug-fix since in the past we had:
```cpp
boost::container::small_vector<std::size_t, 9>
Acts::BinFinder<external_spacepoint_t>::findBins(
    std::size_t phiBin, std::size_t zBin,
    const Acts::SpacePointGrid<external_spacepoint_t>* binnedSP) const {
  // if zBinNeighbors is not defined, get the indices using
  // neighborHoodIndices
  if (m_zBinNeighbors->empty()) {
    return binnedSP->neighborHoodIndices({phiBin, zBin}).collect();
  }
  // if the zBinNeighbors is defined, get the indices from there
  std::array<std::pair<int, int>, 2> sizePerAxis;
  sizePerAxis.at(0) = std::make_pair(-m_numPhiNeighbors, m_numPhiNeighbors);
  sizePerAxis.at(1) = (*m_zBinNeighbors)[zBin - 1];
  return binnedSP->neighborHoodIndices({phiBin, zBin}, sizePerAxis).collect();
}
```

In the case `m_zBinNeighbors` is empty, the code is instructed to search neighbours in the range `{-1, 1}` FOR EVERY AXIS (this is what `binnedSP->neighborHoodIndices({phiBin, zBin}).collect();` does), thus overwriting the value set by the user for `m_numPhiNeighbors`. 

Now, the axes are treated independently

This requires: https://github.com/acts-project/acts/pull/2835


Core/include/Acts/Utilities/GridBinFinder.hpp
```diff
@@ -22,30 +24,33 @@ namespace Acts {
...
-      m_zBinNeighbors;
-  int m_numPhiNeighbors = 1;
+  template <typename first_value_t, typename... vals>
+  void storeValue(first_value_t&& fv, vals&&... others);
+
+  std::array<std::pair<int, int>, DIM> getSizePerAxis(
+      const std::array<std::size_t, DIM>& locPosition) const;
+
+ private:
+  using stored_values_t = std::variant<int, std::vector<std::pair<int, int>>>;
```
paulgessinger, line 51: Keep in mind that the variant has the memory footprint of the vector even in case an integer is stored.

CarloVarni: good point

Core/include/Acts/Utilities/GridBinFinder.hpp
```diff
@@ -0,0 +1,114 @@
...
+  ///
+  /// @param [in] grid The Grid
+  /// @return If the GridBinFinder is compatible with the grid
+  template <typename stored_t, class... Axes>
+  bool isGridCompatible(const Acts::Grid<stored_t, Axes...>& grid) const;
+
+ private:
+  using stored_values_t = std::variant<int, std::vector<std::pair<int, int>>>;
+  /// @brief the instructions for retrieving the nieghbouring bins for each given axis in the grid
+  /// These values are provided by the user and an be either ints or a vector of
```
niermann999, line 101:
```suggestion
  /// These values are provided by the user and can be either ints or a vector of
```

Core/include/Acts/Utilities/GridBinFinder.hpp
```diff
@@ -0,0 +1,114 @@
+// This file is part of the Acts project.
+//
+// Copyright (C) 2023 CERN for the benefit of the Acts project
```
niermann999, line 3: Let's also update the copyright everywhere
```suggestion
// Copyright (C) 2024 CERN for the benefit of the Acts project
```

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/SeedingAlgorithm.hpp
```diff
@@ -63,7 +63,7 @@ class SeedingAlgorithm final : public IAlgorithm {
     std::vector<std::pair<int, int>> zBinNeighborsBottom;
     // number of phiBin neighbors at each side of the current bin that will be
     // used to search for SPs
-    int numPhiNeighbors = 0;
+    int numPhiNeighbors = 1;
```
niermann999, line 66: What happens in the grid iteration if the local position happens to be at the edge of the phi axis? Does the iteration wrap around at -pi <-> pi or does this never occur?

niermann999 (commented):
Btw, have you considered using ```std::ranges``` for this? E.g. writing a local bin indexer should be as easy as:
```
std::views::cartesian_product{
        std::views::iota{std::get<I>(search_window)}...};
```

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/SeedingAlgorithm.hpp
CarloVarni: that should be handled automatically, since the phi axes is a `detail::AxisBoundaryType::Closed`. That means (according to documentation in https://github.com/acts-project/acts/blob/main/docs/core/misc/grid_axis.md?plain=1#L32) that "out-of-bounds lookups wrap-around to the other side of the axis"

CarloVarni: But indeed we can add some unit tests

CarloVarni:
> Btw, have you considered using `std::ranges` for this? E.g. writing a local bin indexer should be as easy as:
> 
> ```
> std::views::cartesian_product{
>         std::views::iota{std::get<I>(search_window)}...};
> ```

`std::ranges` is C++20, which as of now is not yet ok with Athena. Thus I'm limiting myself to use C++17

Core/include/Acts/Utilities/GridBinFinder.hpp
```diff
@@ -0,0 +1,114 @@
...
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+#include "Acts/Seeding/SpacePointGrid.hpp"
```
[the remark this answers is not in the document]
CarloVarni: well, this is a ~long story. The point here is that a `Grid` is quite a generic object that potentially can be used anywhere and not only for the seeding. So I believe it makes more sense to have that generic object into the `Utility` project.

When it comes to `SpacePointGrid`, then we have a particular specialization of a Grid. In this case for instance the stored elements are `std::vector<std::unique_ptr<Acts::InternalSpacePoint<external_spacepoint_t>>>` and the axis are phi and z.

This is very specific to the seeding, and thus it should remain in the `Seeding` project. In another PR of mine I'm changing this further in view of adding support for a telescope-geometry grid. 

But this is out of scope w.r.t. this PR

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/SeedingAlgorithm.hpp
CarloVarni: Added a test dedicated to that

niermann999: Nice, thanks! I was just worried that it might not do that when you look up bins by any kind of bin indices instead of from a position, but fair enough

Core/include/Acts/Utilities/GridBinFinder.hpp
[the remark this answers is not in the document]
niermann999: Ah, Ok, I was just curious why it would belong here, thanks

niermann999 (approved):

niermann999 (approved):

andiwand (approved):

acts-project-service:
## :red_circle: Athena integration test results [8bb1baef5550f6e2dbadb4acf426c6239a32150b]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/35299545) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6740231)!

PR #2843, opened 2023-12-18
 feat: Misalignment shift and test check that only selected modules are shifted

LaraCalic:
`edit:` remove reminder text from this description to optimise search


acts
paulgessinger, line 1: Is this a submodule?

Examples/Algorithms/Alignment/include/ActsExamples/Alignment/AlignmentAlgorithm.hpp
```diff
@@ -52,8 +75,44 @@ class AlignmentGroup {
...
+            
+            Eigen::Affine3d transform = Eigen::Translation3d(center - centerOfMass);
+            surfaceTransforms[surface] = transform;
+        }
+    }  
+
+      // return value of this function is a unit vector that points in x - direction (x direction is a choice)
+    Eigen::vector3 ReferenceDirection() const {
+        return Eigen::vector3::UnitX();
+    }
```
paulgessinger, line 112: Should this be a seperate function? Also, does this compile?

paulgessinger (commented):
- `DetectorAlignment.{cpp,hpp}` have been deleted in #2685
- Calculation of relative transform only does translation, I think the rotation is missing
- Initialization flow looks ok in general
- I don't think I fully understand how this test is now supposed to work, there's still an input file and command line arguments?

andiwand:
archiving this due to inactivity for now

PR #2768, opened 2023-12-05
refactor: Refactor navigation

andiwand:
After #2625 turned out to be harder than initially thought I tried to factor out some changes to get them into main already.

Summary of the changes I made
- `pLimit` = path limit -> far limit = `farLimit`
- `oLimit` = overstep limit -> near limit = `nearLimit`
- replaced `DetectorNavigator` iterators with indices

PR #2846, opened 2023-12-20
refactor: Backport navigation rewrite changes

andiwand:
Backporting more changes from https://github.com/acts-project/acts/pull/2625 to main. This is a followup to #2768 

Summary
- remove overstepping from steppers
- use surface tolerance for geometry lookups
- add self consistency navigation test


Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -46,10 +39,10 @@ struct NavigationOptions {
...
   bool resolvePassive = false;
 
-  /// object to check against: at start
-  const object_t* startObject = nullptr;
-  /// object to check against: at end
-  const object_t* endObject = nullptr;
+  /// Hint for start object
+  const void* startObject = nullptr;
+  /// Hint for end object
+  const void* endObject = nullptr;
```
paulgessinger, line 45: Generally, we've tried to avoid bare void pointers in the code as much as possible. That doesn't mean to not ever use them, but to apply caution.

In the recent vertexing rewrite, I used a struct around a void pointer. The struct can be made to allow comparisons with arbitrary pointers, but the void pointer stays private. One can then also add a type index to ensure type safety, in case the pointer ever needs to be cast back to a concrete type. 

Would that be an option here?

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -65,10 +65,13 @@ class Intersection {
   }
 
   constexpr const Position& position() const { return m_position; }
+  constexpr Position& position() { return m_position; }
```
paulgessinger, line 68: Do we need this for intersection patching, or what's the reason?

Core/src/Geometry/Layer.cpp
```diff
@@ -119,16 +120,18 @@ Acts::Layer::compatibleSurfaces(
...
   // @todo: - we might be able to skip this by use of options.pathLimit
   // check if you have to stop at the endSurface
-  double nearLimit = options.nearLimit;
-  double farLimit = options.farLimit;
   if (options.endObject != nullptr) {
     // intersect the end surface
     // - it is the final one don't use the boundary check at all
     SurfaceIntersection endInter =
-        options.endObject
+        static_cast<const Surface*>(options.endObject)
```
paulgessinger, line 134: Here we do in fact cast back the void pointer. Can we add a struct with a type index and a cast function?

Core/src/Geometry/TrackingVolume.cpp
```diff
@@ -472,164 +468,158 @@ void Acts::TrackingVolume::closeGeometry(
...
 
   // Helper function to test intersection
   auto checkIntersection =
-      [&](SurfaceMultiIntersection& smIntersection,
-          const BoundarySurface* bSurface) -> BoundaryIntersection {
-    for (const auto& sIntersection : smIntersection.split()) {
-      if (!sIntersection) {
+      [&](SurfaceMultiIntersection& candidates,
+          const BoundarySurface* boundary) -> BoundaryIntersection {
+    for (const auto& intersection : candidates.split()) {
```
paulgessinger, line 488: Funky indentation?

Core/src/Geometry/TrackingVolume.cpp
```diff
@@ -472,164 +468,158 @@ void Acts::TrackingVolume::closeGeometry(
...
-          // create a layer intersection
-          lIntersections.push_back(LayerIntersection(atIntersection, tLayer));
-        }
+  if (m_confinedLayers == nullptr) {
+    return {};
+  }
+
+  // start layer given or not - test layer
+  const Layer* tLayer = options.startObject != nullptr
+                            ? static_cast<const Layer*>(options.startObject)
```
paulgessinger, line 586: Void pointer static cast here as well.

Core/src/Geometry/TrackingVolume.cpp
```diff
@@ -472,164 +468,158 @@ void Acts::TrackingVolume::closeGeometry(
...
+  // which are not valid for navigation. These are discarded here by checking
+  // against the minimum path length.
+  auto min = std::min_element(
+      lIntersections.begin(), lIntersections.end(),
+      [](const LayerIntersection& a, const LayerIntersection& b) {
+        return a.first.pathLength() < b.first.pathLength();
+      });
+  std::rotate(lIntersections.begin(), min, lIntersections.end());
+  lIntersections.resize(std::distance(min, lIntersections.end()),
+                        {SurfaceIntersection::invalid(), nullptr});
```
paulgessinger, line 621: This rotates the min element to the start and then resizes from min element to the end. But is the min iterative even valid after the rotation? 

I guess this works because it's only ever 1 or 2 elements, right?

Tests/UnitTests/Core/Propagator/NavigatorTests.cpp
paulgessinger, line 1: I like this test!

paulgessinger (commented):
Very nice!

Core/include/Acts/Propagator/Navigator.hpp
andiwand: That would be an option or I could revive the template param. I guess it does not hurt that much

Core/src/Geometry/TrackingVolume.cpp
andiwand: yeah that looks funny but seems like it is what clang-format wants

Core/src/Geometry/TrackingVolume.cpp
andiwand: in some rare cases we resolve a layer on the other side of the cylinder which then messes with the navigation. here I check if the order makes sense and remove the ones which are out of order at the beginning of the vector

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -929,7 +926,13 @@ class Navigator {
               state.geoContext, stepper.position(state.stepping),
               state.options.direction * stepper.direction(state.stepping),
               navOpts, logger());
-      // The number of boundary candidates
+      std::sort(state.navigation.navBoundaries.begin(),
+                state.navigation.navBoundaries.end(),
+                [](const auto& a, const auto& b) {
+                  return SurfaceIntersection::forwardOrder(a.first, b.first);
```
andiwand, line 932:
```suggestion
                  return SurfaceIntersection::pathLengthOrder(a.first, b.first);
```

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -1118,30 +1122,31 @@ class Navigator {
...
-        std::ostringstream os;
-        os << state.navigation.navSurfaces.size();
-        os << " surface candidates found at path(s): ";
-        for (auto& sfc : state.navigation.navSurfaces) {
-          os << sfc.pathLength() << "  ";
-        }
-        logger().log(Logging::VERBOSE, os.str());
+    std::sort(state.navigation.navSurfaces.begin(),
+              state.navigation.navSurfaces.end(),
+              SurfaceIntersection::forwardOrder);
```
andiwand, line 1135:
```suggestion
              SurfaceIntersection::pathLengthOrder);
```

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -1187,28 +1193,35 @@ class Navigator {
...
     // Request the compatible layers
     state.navigation.navLayers =
         state.navigation.currentVolume->compatibleLayers(
             state.geoContext, stepper.position(state.stepping),
             state.options.direction * stepper.direction(state.stepping),
             navOpts);
+    std::sort(state.navigation.navLayers.begin(),
+              state.navigation.navLayers.end(),
+              [](const auto& a, const auto& b) {
+                return SurfaceIntersection::forwardOrder(a.first, b.first);
```
andiwand, line 1209:
```suggestion
                return SurfaceIntersection::pathLengthOrder(a.first, b.first);
```

andiwand:
measured performance - no significant/measurable change

Tests/UnitTests/Core/Propagator/NavigatorTests.cpp
```diff
@@ -693,5 +700,362 @@ BOOST_AUTO_TEST_CASE(Navigator_target_methods) {
...
+                         bwdStepSurfacesTmp.end());
+
+  // TODO backward-backward step compatibility test
+
+  std::reverse(bwdStepSurfaces.begin(), bwdStepSurfaces.end());
+  BOOST_CHECK_EQUAL_COLLECTIONS(bwdStepSurfaces.begin(), bwdStepSurfaces.end(),
+                                fwdStepSurfaces.begin(), fwdStepSurfaces.end());
+}
+
+/// the actual test nethod that runs the test can be used with several
```
niermann999, line 925:
```suggestion
/// the actual test method that runs the test can be used with several
```

Tests/UnitTests/Core/Propagator/NavigatorTests.cpp
```diff
@@ -693,5 +700,362 @@ BOOST_AUTO_TEST_CASE(Navigator_target_methods) {
...
+        bdata::random(
+            (bdata::engine = std::mt19937(), bdata::seed = 23,
+             bdata::distribution = std::uniform_int_distribution<int>(0, 1))) ^
+        bdata::xrange(ntests),
+    pT, phi, theta, charge, index) {
+  if (index < skip) {
+    return;
+  }
+
+  double p = pT / sin(theta);
```
niermann999, line 1038:
```suggestion
  double p = pT / std::sin(theta);
```

niermann999 (commented):
Looks good to me, although I am not sure I can judge the implications of this change

Core/src/Geometry/TrackingVolume.cpp
andiwand: at the moment they have to be thrown out. otherwise we end up stepping backward since I killed to overstepping limit completely.

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -836,7 +832,7 @@ class Navigator {
       NavigationOptions<Surface> navOpts;
       // Exclude the current surface in case it's a boundary
       navOpts.startObject = state.navigation.currentSurface;
-      navOpts.nearLimit = stepper.overstepLimit(state.stepping);
+      navOpts.nearLimit = state.options.surfaceTolerance;
```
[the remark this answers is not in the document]
andiwand: Overstepping is now always resolved by the stepper by trying to reach the surface no matter how far we are overstepping. The idea of using the `surfaceTolerance` here was to only look for new intersections ahead of us since anything in range of `surfaceTolerance` would be considered to be "on surface".

Core/include/Acts/Propagator/detail/SteppingHelper.hpp
```diff
@@ -51,9 +53,8 @@ Acts::Intersection3D::Status updateSingleSurfaceStatus(
     return Intersection3D::Status::onSurface;
   }
 
-  // Path and overstep limit checking
-  const double nearLimit = stepper.overstepLimit(state);
-  const double farLimit = state.stepSize.value(ConstrainedStep::aborter);
+  const double nearLimit = std::numeric_limits<double>::lowest();
+  const double farLimit = std::numeric_limits<double>::max();
```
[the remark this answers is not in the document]
andiwand: right since I removed the overstepping tolerance and I don't think using the aborter limit makes sense we can just use the intersection without any constraints

Tests/UnitTests/Core/Propagator/MaterialCollectionTests.cpp
```diff
@@ -57,71 +57,44 @@ MagneticFieldContext mfContext = MagneticFieldContext();
...
-Navigator navigatorES({tGeometry});
-Navigator navigatorSL({tGeometry});
-
 using BField = ConstantBField;
 using EigenStepper = Acts::EigenStepper<>;
 using EigenPropagator = Propagator<EigenStepper, Navigator>;
 using StraightLinePropagator = Propagator<StraightLineStepper, Navigator>;
 
-const double Bz = 2_T;
+double Bz = 2_T;
```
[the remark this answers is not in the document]
andiwand:
```suggestion
const double Bz = 2_T;
```

andiwand:
error logs need to be addressed

andiwand:
Take a quick look into a failing propagation and it looks like these are tracks which start to loop in the solenoid. Before my change these were filtered based on the far limit.

Using the far limit to discard intersections seems more like a hack to me. A proper solution would be to stop the propagation after the long strips IMO. But I am not sure how we can achieve this easily.

Alternatively I can also backport the far limit which should give us the current results.

andiwand:
After reintroducing the `farLimit` based on the aborter step size the physmon diff is calmed down and it looks like there are some minor efficiency improvements hinting that this fixed the navigation a bit.

What is left to do for me is to check if this has any performance implications.

Core/include/Acts/Propagator/StandardAborters.hpp
```diff
@@ -101,14 +101,11 @@ struct SurfaceReached {
...
-    // always work for perigee surfaces.
+    // not using the stepper overstep limit here because it does not always work
+    // for perigee surfaces
     // note: the near limit is necessary for surfaces with more than one
-    // intersections in order to discard the ones which are behind us.
-    const double nearLimit =
-        overrideNearLimit.value_or(stepper.overstepLimit(state.stepping));
-    const double farLimit =
-        state.stepping.stepSize.value(ConstrainedStep::aborter);
+    // intersection in order to discard the ones which are behind us
```
niermann999, line 107: But the near limit is removed, is the comment still valid?

niermann999 (commented):
Just a small comment

Core/include/Acts/Propagator/StandardAborters.hpp
andiwand: For the `SurfaceReached` we still use the `nearLimit`. I wanted to drop it but when we target a cylinder with the `SurfaceReached` aborter it would break due to the ambiguity of the intersections.

acts-project-service:
## :red_circle: Athena integration test results [57fd232791810cbb1bb5a58f57c73e73c7a4a4e3]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39288909) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7443859)!

PR #2847, opened 2023-12-20
refactor!: Use Acts::Any for magnetic field cache

paulgessinger:
Uses the somewhat recently added `Acts::Any` type for the magnetic field cache storage, instead of the previously used `SmallObjectCache` object. `Acts::Any` should be more flexible, potentially faster and less specific to the magnetic field implementation.

I did not observe changes to the CPU performance with this change.


Core/include/Acts/MagneticField/MagneticFieldProvider.hpp
```diff
@@ -24,7 +24,7 @@ class MagneticFieldProvider {
   /// Opaque cache type that can store arbitrary implementation specific cache
   /// data. Examples are an interpolation cell, or an experiment specific
   /// conditions data handle.
-  using Cache = detail::SmallObjectCache;
+  using Cache = Acts::AnyBase<sizeof(char) * 512>;
```
AJPfleger, line 27: You probably just ported from `SmallObjectedCache.hpp`, but do we know why we have chosen a size of `512`?

paulgessinger: I wanted it to be large enough so that our interpolated field map cell fits inside and stays on the stack. The previous implementation did not overflow onto the heap, this one can. I don't think I measured the performance impact of other sizes.

acts-project-service:
## :red_circle: Athena integration test results [12321d5cb9bb4097ae662c29952a1afe709bcb3f]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/34890034) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6648250)!

andiwand (approved):

paulgessinger:
I'll fix the Athena failures tomorrow.

PR #2839, opened 2023-12-18
docs: Add comments grid iterator

CarloVarni:
This adds some comments to the code to describe what the grid iterator classes do. It also adds two methods for retrieving the current local and global positions from the iterators, as well as adding some unit tests to check these newly-added methods.

PR #2854, opened 2023-12-22
refactor!: General binned group for seeding

CarloVarni:
A general class for defining a `BinnedGroup`. This class stores a `grid` (owns it), two `GridBinFinders` and a navigation pattern, optionally defined by the user. The PR also defines its iterator.

This will replace the current `BinnedSPGroup` implementation, which is very specific.

Also, creating and filling a grid cannot be generalized imho: variable definitions, selection cuts, number of axes and their ordering. As such, the creation and the filling of the grid are now handled externally, and not inside the `BinnedGroup` class, and it's the user's duty to make sure the filling happens.

In my mind, all of this should be handled by the user for any custom way of doing it. One thing we can provide (as we currently do but I'm afraid is more a "nice consequence" then a "design") some pre-defined grids for specific geometries. 
For instance we can define a `CylindricalSpacePointGrid<external_spacepoint_t>` (as well as `CylindricalBinnedGroup<external_spacepoint_t>`) or a `TelescopeSpacePointGrid<external_spacepoint_t>` (and its `TelescopeBinnedGroup<external_spacepoint_t>`). 

These last points are not covered here, will be addressed in another PR once this one is in. My idea is to define these "specialization" as simple `typedef`s of the above generic classes, as well as defining functions for their creation and filling. This can be located inside  a new `Acts/Seeding/detail` folder, so we can add as many specializations as possible.

Requires:
- https://github.com/acts-project/acts/pull/2838
- https://github.com/acts-project/acts/pull/2835
- https://github.com/acts-project/acts/pull/2839 (maybe optional?)


Core/include/Acts/Seeding/SpacePointGrid.ipp
```diff
@@ -126,13 +122,120 @@ Acts::SpacePointGridCreator::createGrid(
...
   detail::Axis<detail::AxisType::Variable, detail::AxisBoundaryType::Bound>
       zAxis(std::move(zValues));
-  return std::make_unique<Acts::SpacePointGrid<SpacePoint>>(
+  return Acts::SpacePointGrid<SpacePoint>(
       std::make_tuple(std::move(phiAxis), std::move(zAxis)));
 }
+
+template <typename external_spacepoint_t,
+          typename external_spacepoint_iterator_t, typename callable_t>
+void Acts::SpacePointGridCreator::fillGrid(
```
paulgessinger, line 139: Is this not effectively a method on `grid`? Or is there a reason this is better implemented as an external static method?

Tests/UnitTests/Core/Seeding/BinnedGroupTest.cpp
```diff
@@ -0,0 +1,256 @@
...
+  // We do not provide navigation, the constructor will define it
+  Acts::BinnedGroup<grid_3d_t> group_3d(std::move(grid_3d), *binFinder_3d.get(),
+                                        *binFinder_3d.get());
+
+  // Move Constructor/Assignment
+  const Acts::BinnedGroup<grid_1d_t> group_1d_moved(std::move(group_1d));
+  const Acts::BinnedGroup<grid_2d_t> group_2d_moved(std::move(group_2d));
+  const Acts::BinnedGroup<grid_3d_t> group_3d_moved = std::move(group_3d);
+
+  [[maybe_unused]] const grid_1d_t& retrievedGrid = group_1d.grid();
```
paulgessinger, line 70: Do you need the `maybe_unused` here?

Core/include/Acts/Seeding/BinnedGroupIterator.hpp
paulgessinger, line 1: How do these relate to the bin iterators added in previous PRs?

Tests/UnitTests/Core/Seeding/BinnedGroupTest.cpp
```diff
@@ -0,0 +1,256 @@
...
+#include "Acts/Utilities/GridBinFinder.hpp"
+#include "Acts/Utilities/GridIterator.hpp"
+
+#include <array>
+#include <memory>
+#include <vector>
+
+namespace Acts::Test {
+
+BOOST_AUTO_TEST_CASE(binned_group_constructor) {
```
paulgessinger, line 22: This test is not actually testing anything, is it?

Core/include/Acts/Seeding/BinnedGroup.hpp
```diff
@@ -0,0 +1,104 @@
...
+///
+/// The assumption is that the grid has ownership of the space points and
+/// that the grid value_type (i.e. T in Grid<T, Axes ...>) is an iterable
+/// object of space points, such as a vector< ... >
+template <typename grid_t>
+class BinnedGroup {
+ public:
+#ifndef DOXYGEN
+  friend BinnedGroupIterator<grid_t>;
+#endif
```
paulgessinger, line 33: I think I figured out what the problem with this was before. I believe you can just do

```suggestion
  friend class BinnedGroupIterator<grid_t>;
```

Tests/UnitTests/Core/Seeding/BinnedGroupTest.cpp
```diff
@@ -0,0 +1,256 @@
...
+  Acts::detail::EquidistantAxis xAxis(0, 100, 10);
+  grid_t grid(std::make_tuple(xAxis));
+  /// Add some entries to the grid
+  grid.at(1ul).push_back(4ul);
+  grid.at(1ul).push_back(1ul);
+  grid.at(8ul).push_back(7ul);
+  grid.at(9ul).push_back(2ul);
+
+  std::unique_ptr<binfinder_t> botBinfinder = std::make_unique<binfinder_t>(0);
+  std::unique_ptr<binfinder_t> topBinfinder = std::make_unique<binfinder_t>(1);
```
paulgessinger, line 133: Doesn't really matter for the test, but why `unique_ptr`?

Tests/UnitTests/Core/Seeding/BinnedGroupTest.cpp
```diff
@@ -0,0 +1,256 @@
...
+  Acts::BinnedGroup<grid_t> group(std::move(grid), *botBinfinder.get(),
+                                  *topBinfinder.get());
+
+  std::size_t nIterations = 0ul;
+  for (const auto [bottom, middle, top] : group) {
+    ++nIterations;
+    BOOST_CHECK_EQUAL(bottom.size(), 1ul);
+    BOOST_CHECK_EQUAL(top.size(), 3ul);
```
paulgessinger, line 141: Sorry, I'm slow: how do these numbers come to be? Why is it the same for all iterations?
And where does the `nIterations == 3` come from?

Tests/UnitTests/Plugins/Cuda/Seeding/SeedFinderCudaTest.cpp
```diff
@@ -245,11 +245,13 @@ int main(int argc, char** argv) {
...
-  std::unique_ptr<Acts::SpacePointGrid<SpacePoint>> grid =
+  Acts::SpacePointGrid<SpacePoint> grid =
       Acts::SpacePointGridCreator::createGrid<SpacePoint>(gridConf, gridOpts);
+  Acts::SpacePointGridCreator::fillGrid(config, options, grid, spVec.begin(),
+                                        spVec.end(), ct, rRangeSPExtent);
+
   auto spGroup = Acts::BinnedSPGroup<SpacePoint>(
-      spVec.begin(), spVec.end(), ct, bottomBinFinder, topBinFinder,
-      std::move(grid), rRangeSPExtent, config, options);
+      std::move(grid), *bottomBinFinder.get(), *topBinFinder.get());
```
paulgessinger, line 254:
```suggestion
      std::move(grid), *bottomBinFinder, *topBinFinder);
```

paulgessinger (commented):
So: a `BinnedGroup` now is a grid + an iteration sequence. Could you maybe elaborate, maybe in the tests next to the assertions, what the iteration order is in each case? 

Then I have the general point that you're using `unique_ptr` in ways where I don't think you actually need to. Can you double check this? Are you trying to optimize the number of copies this way, or is there another reason that I'm missing.

Core/include/Acts/Seeding/SpacePointGrid.ipp
CarloVarni: In my opinion `fillGrid` should not be a method of the grid. The grid itself should know nothing about that, since that is an object that defines some axes.

The `fillGrid` methods requires the knowledge of what axis corresponds to what variable and also how to retrieve/compute those variables given (e.g.) a space points. I don't think the grid should care about this. It is the duty of the costumer to define a function for filling their own grid.

So in my mind one will have to create thei own CustomSpacePointGridCreator class and define there the way their grid is defined and the instructions on how to fill it.

Tests/UnitTests/Core/Seeding/BinnedGroupTest.cpp
CarloVarni: not sure. Wouldn't you have compilation warnings if we don't use it?

Tests/UnitTests/Core/Seeding/BinnedGroupTest.cpp
CarloVarni: true ... a remnant of previous versions of the contructors

Tests/UnitTests/Core/Seeding/BinnedGroupTest.cpp
CarloVarni: no idea, i forgot ... I'll remove the unique ptrs

Tests/UnitTests/Core/Seeding/BinnedGroupTest.cpp
CarloVarni: well... not really. This is checking all the constructors are not having problems (including copy and handling the navigation)

Tests/UnitTests/Core/Seeding/BinnedGroupTest.cpp
CarloVarni: The binned group iterator will only iterate on the bins that are not empty. This is to avoid calling the `operator*` which is expensive (it handles the identification of neighbouring bins).

In this test, I'm only filling 3 bins with some values. Thus `nIterations == 3`.

Then, the bottom and top sizes are given by the `GridBinFinder`s we defined above. 

Imagine you are in the bin `(N)`, the top bin finder is requiring to find the neighbours within 1 grid bins. Thus it will return `N-1`, `N` and `N+1`. `Thus top.size() == 3`.

The bottom bin finder is requiring to find the neighbours within 0 grid bins. So it will only return `N`. As such `bottom.size() == 3`

CarloVarni: you can also do more complicated bin searches by providing pairs of ints btw

For instance, if you provide `{k, l}` to the bin finders, then you are instructing the code to fine the neighbouring bins  in the range `N-k, ... N .... N+l`

And you can do it for every axis independently. As well as providing different `k` and `l` for each bin in the same axis. For instance a set of pairs like this will give you different bottom and top sizes depending of which `N` bin you are: `{0, 1}, {1, 1}, {1, 0}, {2, 5}`

Core/include/Acts/Seeding/SpacePointGrid.ipp
paulgessinger: Meaning `SpacePointGridCreator` is supposed be just one way to use this grid type?

Tests/UnitTests/Core/Seeding/BinnedGroupTest.cpp
paulgessinger: I guess it's a compile test.

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [cbd38cd023e71a189b9845c4932c104f12cb7b28]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/35344578) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6748292)!

PR #2887, opened 2024-01-22
feat: Cuboidal detector builder from blueprint

ssdetlab:
Adding the Unit Test for the Cuboidal Container Builder. The Unit Test creates a simple Telescope-style geometry with two arms and four layers from a Blueprint. The Builder chain is tested to reproduce the detector with expected parameters. 

Changes in BlueprintHelper are made to accommodate the Cuboidal geometry. The old fillGaps method is separated into fillGapsCylindrical, fillGapsCuboidal and the interface method that calls the specific implementations. Modifications are made in the endPoint finding function to accommodate not only Z, but also X and Y directions. It should be noted that without adjustToParent flag being set to true, gaps are not guaranteed to be filled properly.  

A small fix related to the selection of the Portal volume attachment direction is added to the CuboidalDetectorHelper. Now it is possible to connect containers and create >1D connected structures. It should be noted that the portal merging is not yet consistent between the volume and container connections.


Core/src/Detector/detail/CuboidalDetectorHelper.cpp
```diff
@@ -202,10 +202,19 @@ Acts::Experimental::detail::CuboidalDetectorHelper::connect(
       auto portalSurface =
           Surface::makeShared<PlaneSurface>(portalTransform, portalBounds);
       auto portal = std::make_shared<Portal>(portalSurface);
+
+      // Assign the portal direction
+      // in a consistent way
```
asalzburger, line 207: You can soften that by 

```
Acts::Direction dir = ( index % 2 == 0 ) ? Direction::Forward : Direction::Backward;
```

asalzburger (changes requested):
Minor suggestion.

Core/src/Detector/detail/CuboidalDetectorHelper.cpp
ssdetlab: Done

asalzburger (approved):

asalzburger:
Download error, I re-started the relevant CI jobs.

acts-project-service:
## :red_circle: Athena integration test results [ccd2d3e4390c28b50938976f85b24e2acfca1894]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/35439454) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6768075)!

PR #2896, opened 2024-01-25
ci: Disable nodeps job because th RPM pull keeps failing

paulgessinger:
Also removes some trailing whitespace.


asalzburger (approved):

asalzburger:
> I think it is back up?

Super unstable - following PR #2897 has already a failure again.

PR #2907, opened 2024-01-29
revert: Use common direction transform Jacobian (#2782)

andiwand:
Output changes are not tolerated by Athena and will be patched to v32.


andiwand:
This should not target main but a release branch which does not exist yet.

Waiting for the CI to complete to update the references

paulgessinger:
Hashes still need to be updated, I believe.

AJPfleger (approved):

PR #2908, opened 2024-01-29
chore: Update ParticleDataTable.hpp

paulgessinger:
This updates the script to generate our particle data table from scikithep, and updates the table.


AJPfleger:
Does `clang-format` do anything after generating the data base?
Maybe we should remove it? Since we don't expect any changes, an ill setup system could even mal-format the file.

paulgessinger:
This is not intended to be run very often, so I would say it's not super critical.

paulgessinger:
Memory consumption of compiling this header file goes from 60M to 77M, which I think we can tolerate.

andiwand (approved):

andiwand:
Are the output changes coming from changes in particle properties like mass @paulgessinger ? Let's look at the new plots and check how big the variation is

paulgessinger:
I haven't looked into this further. I'd wouldn't have expected masses to change actually but who knows.

andiwand:
physmon looks like outliers moved a bit - unproblematic from my side

AJPfleger:
@andiwand suggested in the dev-meeting:
- physmon is fine again, maybe just clash with different PR
- 1 hash changed, maybe because one of the masses changed. We should go with the updated mass.

Next step: @andiwand will update the hash

AJPfleger (approved):

AJPfleger (dismissed):

AJPfleger (dismissed):

AJPfleger (approved):

PR #2909, opened 2024-01-30
chore: material mapper uses visitor

asalzburger:
This PR changes to use the `visitSurfaces(...)` method and introduces a virtual interface `IMaterialMapper`, in order to eventually change the MaterialMapping example to a sequenced mapping.

1) first run surface mapping where you want
2) then run volume mapping 

It also removes 5 times the definition of `DetectorMaterialMaps` in favour of a central file.


asalzburger:
Hi @Corentin-Allaire - this is some long, standing thing I wanted to clean-up, the first PR is neutral, basically I remove the same amount of lines which I add :-)

In a second step, I will change a the way the `MaterialMapping` talks to the new interface now.

Core/src/Material/SurfaceMaterialMapper.cpp
```diff
@@ -49,73 +50,43 @@ Acts::SurfaceMaterialMapper::SurfaceMaterialMapper(
...
+  for (auto& smg : mState->accumulatedMaterial) {
     ACTS_VERBOSE(" -> Surface in with id " << smg.first);
   }
   return mState;
 }
 
 void Acts::SurfaceMaterialMapper::resolveMaterialSurfaces(
     State& mState, const TrackingVolume& tVolume) const {
-  ACTS_VERBOSE("Checking volume '" << tVolume.volumeName()
-                                   << "' for material surfaces.")
```
Corentin-Allaire, line 76: I think this line of info might still be useful ?

Corentin-Allaire (commented):
So the changes looks good but we are losing a one feature compared to the previous implementation (we can discuss it directly if it is an issue or not). Basically with the previous implementation the reason both surface and volume mapper carried both type of material in their state was to allow you to start mid mapping process. Basically you could launch one job doing the volume mapping, get the output file (a material map with volume material but only proto material for the surface). And then in a later job restart from that file to do the surface mapping (this let you do both separately). I don't think you can do that anymore with our implementation.

asalzburger:
Will be superseded by the MM1-MM4 PRs.

PR #2913, opened 2024-01-30
fix: DirectNavigator causes incorrect propagation finalization

paulgessinger:
This is observed to break the KalmanFitter if run with the direct navigator, at least in some cases. To be followed up with a more robust solution.


andiwand (approved):
great thank you!

PR #2916, opened 2024-01-31
refactor: Consistently use `mm` for time in examples IO

andiwand:
We are mixing time units in examples currently. I propose to solve this by consistently using `mm` which is the native internal core unit


paulgessinger:
Hm, shouldn't the `ns` literal then give us the conversion factor into `mm`, so that you can write `ns` and it just does the right thing?

andiwand:
> Hm, shouldn't the `ns` literal then give us the conversion factor into `mm`, so that you can write `ns` and it just does the right thing?

the thing is that we are mixing `ns` and `mm` in the output files. we should either consistently use `ns` or `mm`. I opted for `mm` for now as it is also the native internal unit for time

paulgessinger:
Right and the way in principle that the unit system is designed is that it converts to our internal units. So in my opinion, it the internal unit is `mm`, when you write `25_ns` it should convert this to the equivalent length in `mm` and of you want `ns` back, you divide a time by the constant.

andiwand:
> Right and the way in principle that the unit system is designed is that it converts to our internal units. So in my opinion, if the internal unit is `mm`, when you write `25_ns` it should convert this to the equivalent length in `mm` and if you want `ns` back, you divide a time by the constant.

yes this is what it does. I am not touching this. I just noticed that we use different units for time when we write it to file. sometimes we do no conversion which results in `mm` and sometimes we convert to `ns`. I would argue that this should be consistent to reduce confusion. and I would argue that `mm` would be easier as we would not have to convert everything including covariance matrices

beomki-yeo:
Sorry to jump in but I strongly suggest using one of time units (`ps` `ns` `us` `ms` `s`) for time components

beomki-yeo:
Just to demonstrate how it can go wrong in deriving $c=1$  in the natural unit 

$$c  = \frac{299792458 * unit::m}{1 * unit::s} = \frac{299792458000}{299792458000} $$

$$c  = \frac{299792458 * unit::m}{1 * unit::mm} = ? $$

andiwand:
not sure if I can follow. I don't suggest to change our unit system.

our units are defined as
```
constexpr double mm = 1.0;
constexpr double s = 299792458000.0;
constexpr double c = 1.0;
```

but in examples we are mixing time I/O with `mm` and `ns`. I suggest to use one of both and personally I would use `mm` as it matches up with out internal representation and it might be less surprising during debugging.

but I am happy to use `ns` if you see a benefit.

note that our digitization config also uses `mm` for time

beomki-yeo:
Of course we should not mix ns and mm for time -- I am not raising a question on this. 

What I suggest is not using length unit for time component which will confuse things a lot as I demonstrated above

paulgessinger:
Should we discuss this in the meeting on Tuesday?

andiwand:
`c = 1` is a definition which ties or time unit to the space unit. not sure why you want to derive that

sure lets discuss on tuesday

beomki-yeo:
Discussing why we should use _time_ unit for _time_ quantity - seriously?

beomki-yeo (approved):
As concluded in the meeting today, let's go with the unified representation for length and time

beomki-yeo (approved):

PR #2917, opened 2024-01-31
fix!: Fix vertex finding for seeds with `z=0`

andiwand:
While working on the 4D vertexing performance I realized that our vertex finders will abort if a seed with `z=0` is found. This can happen using a grid based seeder as it can round to exactly 0.

As far as I could see the vertex constraint was only used to signal the finish line for the vertex finding. I exchanged that with an optional and also removed the vector which had no use as far as I could tell


pbutti (commented):
Looks reasonable to me

pbutti (dismissed):

Core/include/Acts/Vertexing/AdaptiveMultiVertexFinder.ipp
```diff
@@ -33,44 +33,44 @@ auto Acts::AdaptiveMultiVertexFinder<vfitter_t, sfinder_t>::find(
...
-         iteration < m_cfg.maxIterations) {
-    // Tracks that are used for searching compatible tracks
-    // near a vertex candidate
-    std::vector<const InputTrack_t*> searchTracks;
-    if (m_cfg.doRealMultiVertex) {
-      searchTracks = origTracks;
-    } else {
-      searchTracks = seedTracks;
-    }
+  while (!seedTracks.empty() && iteration < m_cfg.maxIterations) {
```
felix-russo, line 36: So the previous code was obviously nonsensical, but maybe it would be better to replace it with something like
```suggestion
while (((m_cfg.addSingleTrackVertices && !seedTracks.empty()) ||
          (seedTracks.size()>1)) &&
          iteration < m_cfg.maxIterations) {
```

felix-russo: Otherwise we should remove the member variable `addSingleTrackVertices`, but I don't know if anyone ever sets this to true

Tests/UnitTests/Core/Vertexing/IterativeVertexFinderTests.cpp
```diff
@@ -577,10 +574,7 @@ BOOST_AUTO_TEST_CASE(iterative_finder_test_athena_reference) {
   IPEstimator::Config ipEstimatorCfg(bField, propagator);
   IPEstimator ipEstimator(ipEstimatorCfg);
 
-  using ZScanSeedFinder = ZScanVertexFinder<BilloirFitter>;
-
-  static_assert(VertexFinderConcept<ZScanSeedFinder>,
```
felix-russo, line 582: Why are we not asserting this anymore?

felix-russo (commented):
Very good idea; I don't know why we did this but I am guessing that it is a relict from athena.

- Do you know why the physmon changes for the IVF? I don't see why it should
- Is this breaking because of the (very sensible) renaming?

Tests/UnitTests/Core/Vertexing/IterativeVertexFinderTests.cpp
andiwand: `ZScanVertexFinder` is actually a seed finder and does not comply with the vertex finder interface

Core/include/Acts/Vertexing/AdaptiveMultiVertexFinder.ipp
andiwand: right I also thought about fixing this but was a bit scared about performance changes. we can try

andiwand:
> Very good idea; I don't know why we did this but I am guessing that it is a relict from athena.
> 
> * Do you know why the physmon changes for the IVF? I don't see why it should
> * Is this breaking because of the (very sensible) renaming?

I am also a bit puzzled about the performance changes in the IVF. I can try to track it down

Yeah I think flagging as breaking makes a lot of sense here

Tests/UnitTests/Core/Vertexing/IterativeVertexFinderTests.cpp
felix-russo: lol yes that's true

Core/include/Acts/Vertexing/AdaptiveMultiVertexFinder.ipp
felix-russo: Alternatively, we could leave this for a separate PR

andiwand: pulled this into https://github.com/acts-project/acts/pull/2931

andiwand:
blocked by the vertexing refactor atm

andiwand:
I decoupled this one now from https://github.com/acts-project/acts/pull/2930 and pulled the vertex refactor changes in. Ready to go from my side

Core/include/Acts/Vertexing/ZScanVertexFinder.ipp
```diff
@@ -95,11 +95,5 @@ inline auto Acts::ZScanVertexFinder::find(
...
   Vertex vtxResult = Vertex(output);
 
-  // Vector to be filled with one single vertex
-  std::vector<Vertex> vertexCollection;
-
-  // Add vertex to vertexCollection
-  vertexCollection.push_back(vtxResult);
-
-  return vertexCollection;
+  return std::vector<Vertex>{vtxResult};
```
paulgessinger, line 98:
```suggestion
  return {vtxResult};
```

paulgessinger (approved):
Seems reasonable, I only have nitpicks.

Output change is likely due to extra vertices being found?

Core/include/Acts/Vertexing/AdaptiveMultiVertexFinder.ipp
```diff
@@ -158,8 +161,15 @@ auto Acts::AdaptiveMultiVertexFinder<vfitter_t>::doSeeding(
   if (!seedResult.ok()) {
     return seedResult.error();
   }
+  const auto& seedVector = *seedResult;
+
+  ACTS_DEBUG("Found " << seedVector.size() << " seeds");
+
+  if (seedVector.empty()) {
+    return std::optional<Vertex>();
```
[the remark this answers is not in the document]
andiwand: I think I encountered problems with the template auto deduction but maybe this is gone now with untemplating stuff

andiwand:
the problem with the construction comes from `Result` so I fear we have to keep the calls to the `std::vector` constructor

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [123746ce319d7782288af1b67db5f7aa18937669]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36332448) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6934160)!

paulgessinger:
This PR changes Athena outputs

Core/include/Acts/Vertexing/AdaptiveMultiVertexFinder.ipp
andiwand, line 1: The code shuffle here might have unexpected effects. I will revert and see if that is it.

Otherwise I can only suspect that the number of vertices is changing. If so I would think it is a bug fix but we have to check the plots.

andiwand:
https://github.com/acts-project/acts/pull/3002 will hopefully resolve this

PR #2922, opened 2024-02-01
test: Test KF direct navigation

andiwand:
Since this broke lately without noticing in ACTS but later on in ATHENA we would like to add this to the Python level tests


andiwand:
looks like this is not producing any errors @paulgessinger 

I guess we have to inspect the track summary to check if there are entries?

andiwand:
@paulgessinger I fear my setup is not reproducing the bug

asalzburger (approved):
Thanks, it's good to have this monitored.

PR #2927, opened 2024-02-05
refactor!: Change CMake defaults to be able to run the full chain

andiwand:
I think the CMake defaults should be tuned in a way that our standard users can use ACTS as is without tweaking options. To me it seems like the default workflow a new user would like to run is the ODD full chain. Therefore I enabled DD4hep, Fatras and the python bindings to `ON` by default.


paulgessinger:
I think out of the box, without changing the options, the software should build with only eigen and boost.

paulgessinger:
Could we support this using CMake presets, like what @krasznaa set up for the GPU projects?

andiwand:
Right maybe this was a bit premature. I just think that this is one of the barriers one encounters when using ACTS the first time and I wondered if we can resolve this by just toggling the defaults a bit

paulgessinger:
I think presets would be perfect for this. I'm also thinking this could simplify the various build configurations that we run in the CI.

PR #2932, opened 2024-02-06
refactor: Assert to be `on surface` for surface functions with free param input

andiwand:
This popped up in https://github.com/acts-project/acts/pull/2811 - as this can be a pitfall we are asserting if we are really on surface. This is done via an `assert` which is not evaluated in standard release build

This also fixes a bug in the covariance engine. We used to first do the cov transport and then check if we are actually on surface. I turned these checks around.


Tests/UnitTests/Core/Propagator/EigenStepperTests.cpp
```diff
@@ -505,17 +505,14 @@ BOOST_AUTO_TEST_CASE(eigen_stepper_test) {
...
   freeParams = detail::transformBoundToFreeParameters(
       bp.referenceSurface(), tgContext, bp.parameters());
-  freeParams.segment<3>(eFreePos0) *= 2;
-  freeParams[eFreeTime] *= 2;
-  freeParams[eFreeQOverP] *= -0.5;
 
   es.update(esState, freeParams, bp.parameters(), 2 * (*bp.covariance()),
             *plane);
-  CHECK_CLOSE_OR_SMALL(es.position(esState), 2. * pos, eps, eps);
+  CHECK_CLOSE_OR_SMALL(es.position(esState), pos, eps, eps);
```
paulgessinger, line 511: Why are these different?

andiwand: I think what the test was doing before was scaling up the global position while leaving the local one unchanged. this fails now because we check if we are actually on surface. I changed the test now to set the original position which corresponds to the bound position

paulgessinger:
So the assert triggers very frequently. I think I faintly remember observing this in past when we force free parameters to be on surface during navigation, effectively doing layer projection. 

I tried working around this in the past by having a method where you coerce free parameters to a surface regardless of compatibility but I think we abandoned this. 

Not sure if it's worth pursuing this in this PR and for this use case but we probably should keep this in mind.

acts-project-service:
## :red_circle: Athena integration test results [3fa1300db51178db48f288982bfe7f45e0b77fcf]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/35938577) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6861978)!

PR #2473, opened 2023-09-21
feat!: volume agnostic material mapping

asalzburger:
This PR changes the material mapper logic and cleans up the code quite a bit.

While the algorithmic code rests identical (output unchanged), the `SurfaceMaterialMapper` and `VolumeMaterialMapper` have now entirely separate functionality:

  * the surface mapper does only surface mapping and already implements `TrackingGeometry` and `Detector` schema
  * the volume mapper does only volume mapping (currently not yet fully supporting the `Detector` schema)

It does a small renaming:
  * `MappingType` -> `MaterialMappingType`
In order to harmonise with:
  * `MaterialMappingState` the state base class
  * `MaterialMappingResult` the surface/volume maps produced

The Mappers are now designed to run in a chained setup, i.e. one mapper runs and returns a pair of `mapped` and `unmapped` tracks, the `unmapped` part is then processed by the subsequent mapper.

By doing so, the Mappers became `const correct` (except for the finalizeMaps) call, as modification of the input collection is not allowed anymore. 

Execution time / json map of vanilla ODD *before* change:
```
09:48:55    Sequencer      INFO      Processed 1000 events in 8.657600 s (wall clock)
09:48:55    Sequencer      INFO      Average time per event: 8.594614 ms/event
```

```
-rw-r--r--  1 salzburg  staff  238417544 Sep 22 09:48 material-map.json
```

Execution time / json map of vanilla ODD *after* change:
```
09:49:33    Sequencer      INFO      Processed 1000 events in 8.856570 s (wall clock)
09:49:33    Sequencer      INFO      Average time per event: 8.160427 ms/event
```

```
-rw-r--r--  1 salzburg  staff  238417544 Sep 22 09:49 material-map.json
```

It also includes a comprehensive set of new UnitTests for the `SurfacesMaterialMapper` which reflects in a sizeable increase of test coverage.

The parameter optimisation and the full validation of the `VolumeMaterialMapper` as well as the follow-up for the `Detector` model will come in separate PRs.

PR #2935, opened 2024-02-07
feat: Geometry-agnostic material mapping for both surfaces and volumes 

ssdetlab:
This is an attempt to finish the work started by the #2473

In addition to the changes described in the original PR: 

1) Support of both the `TrackingGeometry` `Navigator` and the `DetectorNavigator` in both `SurfaceMaterialMapper` and `VolumeMaterialMapper` is implemented. 

2)  Volume material mapping is finished with conversion of the `VolumeHit` struct into something that can support both the geometries for the time being. Now it is inheriting from the `InteractionVolume` which already has some infrastructure for the `Detector`/`TrackingGeometry` switching. Slight renaming is introduced.

3) Minor fixes in Python bindings are made. The segfault problem of the original PR is resolved. Root hashes were modified due to the renaming of the output files. 

4) Modifications are made to the `DetectorNavigator` to resolve the problem of stepping loop abortion before the first step is even made, and the problem of exiting the stepping loop due to the `VolumeBounds::inside()` returning true when the stepper just exited the `Portal` and is on the verge of the volume. The latter prevented End Of World check from executing correctly. These are, most likely, temporary solutions, as it feels that something smarter can be done there. 

5) Unit Tests are expanded for both Volume and Surface material mapping.


ssdetlab:
Looks like segfaults in the python tests were caused by the discrepancy in the root hash names. What is interesting, locally it didn't lead to segfaults. The tests were just failing. Have to do some cleanup now

ssdetlab:
Ok, seems that the problem wasn't only in the hash checks. Exposing the bindings code to the containers of `IMaterialMapper` brings the segfaults back.

ssdetlab:
Separated the Navigator part into the #2988

Closing for now, will come back later.

PR #2936, opened 2024-02-07
fix: Propagate initial vertex time variance for AMVF w/o time

andiwand:
Logs like
```
21:16:17    VertexPerfor   WARNING   Nonpositive variance after vertex fit: Var(T) = 0 <= 0.
```
are gone after this change.

To me it seems more sensical to propagate the initial time variance instead of hard setting it to `0` which causes the cov matrix to be invalid.


andiwand:
@felix-russo do you have an opinion on this?

pbutti:
@andiwand Alternatively one could use a very large positive value to the initial for sigma_tt and set the other terms in the last row/column of the covariance matrix to 0 and just call the KF update on it. 
Removing the lines where the time row/column are set to 0 makes sense to me.
In principle in the case of no time information, the last row/column should not provide any weight.

andiwand:
> @andiwand Alternatively one could make sure that a very large positive value to the initial for sigma_tt is present in the initial cov matrix (or set it to a very large value) and check/set that the other terms in the last row/column of the covariance matrix to 0 and just call the KF update on it. Removing the lines where the time row/column are set to 0 makes sense to me. In the case of no time information, the last row/column should not provide any weight and setting it as above should achieve that.

@pbutti We are actually forwarding the initial covariance into the vertex fit which, like you described, has a big diagonal covariance. To me it seems good enough to use the initial cov. What do you think?

felix-russo:
> @felix-russo do you have an opinion on this?

Looks very good to me!

andiwand:
I will go ahead and update the reference here

pbutti (approved):

acts-project-service:
## :red_circle: Athena integration test results [b50dd086d76b0a858952563cd713510ad55ae73b]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36099300) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6891174)!

paulgessinger:
Athena failures unrelated.

PR #2945, opened 2024-02-14
feat: Propagator Helpers

andiwand:
Convenience helpers to construct `Propagator` and `PropagatorOption` objects.

Setting up a propagator is quite involved as you need to know what stepper and what navigator to use. Usually there is a default setup for this which expressed here as `buildStandardPropagator`.

Disclaimer: It feels a bit like an anti-pattern to me but I see people struggle with setting up a propagator frequently. This procedure is also abstracted a bunch of times in our unit tests which could be unified by such a factory / helper.

pulled out of https://github.com/acts-project/acts/pull/2722


paulgessinger:
This seems out of date. Will you get back to this @andiwand or can I close it?

andiwand:
can be closed for now

PR #2946, opened 2024-02-14
refactor!: Use `Delegate` for track linearizers

paulgessinger:
This allows removing the linearizer template from: 
- Billoir fitter
- ZScan finder
- IVF
- AMVFitter

Part of:
- #2842 

Blocked by:
- #2886

PR #2842, opened 2023-12-18
refactor: Vertex implementation

paulgessinger:
Big refactor of the vertexing code to rely much much less on templating.

Blocked by:
- https://github.com/acts-project/acts/pull/2874
- https://github.com/acts-project/acts/pull/2876
- #2877
- #2878 
- #2880
- #2881
- #2886 
- #2946 
- #2948
- #2951
- #2955
- #2952 
- #2953
- #2971 
- #2973 

TODO:
- [ ] Move `Vertex.ipp` to `Vertex.cpp`

PR #2948, opened 2024-02-14
feat!: Add `IVertexFinder` interface, use in vertexing

paulgessinger:
This PR introduces a new `IVertexFinder` with an interface that can support the existing vertex finder implementations, with the exception of the seed vertex finder (which depends on specific spacepoint triplets).

The interface introduces a type-erased state holder that the client of the finder initializes via the `makeState` interface function, and which the finder implementation downcasts to manipulate the state. This enables the concrete interface without templating.

Since some vertex finders use *other* vertex finders as seeders, this also allows to have the seed finder property use this `IVertexFinder` interface.

I did not observe CPU performance degradation as documented in #2842.

Part of:
- #2842 

Blocked by:
- #2946


Core/include/Acts/Vertexing/IVertexFinder.hpp
paulgessinger, line 1: This design I think I really need feedback on!

paulgessinger:
@andiwand this is ready now.

Core/include/Acts/Vertexing/AdaptiveGridDensityVertexFinder.ipp
```diff
@@ -6,24 +6,24 @@
...
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-template <typename vfitter_t>
-auto Acts::AdaptiveGridDensityVertexFinder<vfitter_t>::find(
+inline auto Acts::AdaptiveGridDensityVertexFinder::find(
     const std::vector<InputTrack>& trackVector,
-    const VertexingOptions& vertexingOptions, State& state) const
+    const VertexingOptions& vertexingOptions, IVertexFinder::State& state) const
     -> Result<std::vector<Vertex>> {
+  auto& thisState = state.as<State>();
```
andiwand, line 13: alternatively you could rename the param and leave the concrete state with the same name. that might be diff friendlier

Core/include/Acts/Vertexing/IVertexFinder.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include <vector>
+
+namespace Acts {
+
+class Vertex;
+struct InputTrack;
+struct VertexingOptions;
+
+class IVertexFinder {
```
andiwand, line 23: I think this could use some doc

Core/include/Acts/Vertexing/IVertexFinder.hpp
```diff
@@ -0,0 +1,40 @@
...
+ public:
+  using State = Acts::AnyBase<128>;
+
+  virtual Result<std::vector<Vertex>> find(
+      const std::vector<InputTrack>& trackVector,
+      const VertexingOptions& vertexingOptions, State& state) const = 0;
+
+  virtual State makeState() const = 0;
+
+  virtual bool hasTrivialState() const = 0;
```
andiwand, line 33: what does trivial mean in this context?

Core/include/Acts/Vertexing/IVertexFinder.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+class IVertexFinder {
+ public:
+  using State = Acts::AnyBase<128>;
+
+  virtual Result<std::vector<Vertex>> find(
+      const std::vector<InputTrack>& trackVector,
+      const VertexingOptions& vertexingOptions, State& state) const = 0;
+
+  virtual State makeState() const = 0;
```
andiwand, line 31: I guess this is like a default construction right? So any initialization depending on inputs has to be done afterwards?

Core/include/Acts/Vertexing/IVertexFinder.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+  virtual bool hasTrivialState() const = 0;
+
+  virtual void setTracksToRemove(
+      State& state, const std::vector<InputTrack>& removedTracks) const = 0;
```
andiwand, line 36: why "set" and not `removeTracks`?

Core/src/Vertexing/KalmanVertexUpdater.cpp
andiwand, line 1: to me it looks like these changes are unrelated. I have no problem in keeping them but maybe we should mention that in the PR

andiwand (commented):
Left a few general comments. I am definitely in favor of removing such templated interfaces as I do not think we gain any runtime performance from it. On the other hand we lose time during compilation and non-templated code is easier to reason about IMO.

I wonder how far we are from actually using virtual inheritance and keeping the state inside the algorithm. We could even use a templated frontend to keep the ACTS state / algorithm separation intact.

I stumbled across this vertex seed finder and vertex finder interface a few weeks ago and it did not really make sense to me. I think these should be two different interfaces but maybe I am missing something.

Core/include/Acts/Vertexing/IVertexFinder.hpp
paulgessinger: Basically, this is there because the IVF has a *non-trivial* state because it needs to tell the full billoir fitter about the field cache. I'm wondering if it might be better to have the billoirfitter call take the field cache as an explicit argument, and getting rid of this extra step here.

What do you think?

Core/include/Acts/Vertexing/IVertexFinder.hpp
paulgessinger: Yes and no. Construction can depend on finder config arguments that are stored in the finder instance.

That's sort of what `hasTrivialState` is supposed to communicate, but maybe too complex. 

The basic idea is that it doesn't have to have any external input, and the state can be created by the finder implementation.

Core/include/Acts/Vertexing/IVertexFinder.hpp
paulgessinger: Because the way the removal works currently is that the e.g. `AdaptiveGridDensityVertexFinder` removes these tracks in the next find call. We could think about collapsing this to doing the track removal when the function is called.

Core/src/Vertexing/KalmanVertexUpdater.cpp
paulgessinger: I'll try to get rid of this.

Core/include/Acts/Vertexing/IVertexFinder.hpp
paulgessinger: So, looking at this more, it turns out that the `GridDensityVertexFinder` uses the result of the track removal (i.e. "no more tracks were removed") to signal a break to the parent finder. `AdaptiveGridDensityVertexFinder` handles this differently, though. I need to think more about this.

Core/include/Acts/Vertexing/AdaptiveGridDensityVertexFinder.ipp
paulgessinger: Changed.

Core/include/Acts/Vertexing/IVertexFinder.hpp
paulgessinger: Added.

Core/include/Acts/Vertexing/IVertexFinder.hpp
paulgessinger: Removed in favor of adding the magnetic field context to the `makeState` function.

Core/src/Vertexing/KalmanVertexUpdater.cpp
paulgessinger: Moved out of this PR.

Core/include/Acts/Vertexing/IVertexFinder.hpp
paulgessinger: I'd prefer to leave this for this PR, and then rework how the grid density finders handle track removal and termination.

Core/include/Acts/Vertexing/GridDensityVertexFinder.hpp
```diff
@@ -10,16 +10,22 @@
...
 #include "Acts/Vertexing/VertexingOptions.hpp"
 
 #include <map>
 
 namespace Acts {
 
+// namespace detail {
+// template <typename derived_t>
+// }
+
```
andiwand, line 28: dead code?

andiwand (approved):

Core/include/Acts/Vertexing/IVertexFinder.hpp
```diff
@@ -0,0 +1,55 @@
...
+  /// @param state The state object (needs to be created via @c makeState)
+  /// @return The found vertex candidates
+  virtual Result<std::vector<Vertex>> find(
+      const std::vector<InputTrack>& trackVector,
+      const VertexingOptions& vertexingOptions, State& state) const = 0;
+
+  /// Function to create a state object for this concrete vertex finder
+  /// @param mctx The magnetic field context
+  /// @return The state object
+  virtual State makeState(const MagneticFieldContext& mctx) const = 0;
```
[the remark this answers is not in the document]
paulgessinger: Could be, but we already have the state mechanism, and some of the state contains other stuff, so I was thinking it makes sense to colocate the field cache there as well (as before)

Core/include/Acts/Vertexing/GridDensityVertexFinder.hpp
paulgessinger: Will remove

paulgessinger:
I seem to have broken physmon with the updates I added.

andiwand (dismissed):

andiwand (approved):

Core/include/Acts/Vertexing/IterativeVertexFinder.hpp
```diff
@@ -114,21 +114,25 @@ class IterativeVertexFinder {
     /// decide if a track should be checked for reassignment to other vertices
     double cutOffTrackWeightReassign = 1;
 
-    // Function to extract parameters from InputTrack
+    /// Function to extract parameters from InputTrack
     InputTrack::Extractor extractParameters;
+
+    /// Magnetic field provider
+    std::shared_ptr<MagneticFieldProvider> field;
```
wdconinc, line 121: Belated comment, but can this not be
```suggestion
    std::shared_ptr<const MagneticFieldProvider> field;
```
(asking since I'm having to const_cast in EIC code...)

paulgessinger: I believe that's an oversight! Do you want to open a PR? We can also apply a fix to whatever version is convenient for you and make a patch release.

wdconinc: Thanks a lot!

Issue #2905, opened 2024-01-26
bad hole count passed to CKF branch stopper


timadye:
Using the CKF branch stopper from Athena cuts many more tracks than expected in the ITk for |&eta;|>2.5.

Athena's branch stopper is passed an [`Acts::CombinatorialKalmanFilterTipState`](https://github.com/acts-project/acts/blob/main/Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp#L67) object `tipState`, and returns `true` for those with `tipState.nHoles > 2`. This should have the same effect as removing those tracks from the output collection after the CKF is run but, as [these slides](https://github.com/acts-project/acts/files/14067621/ckf_5.pdf) show, there is a large reduction in efficiency.

This issue is observed with Athena `main,r2024-01-18`, which uses ACTS v31.2.0. It does not include the change to the `BranchStopper` signature introduced in #2757, which comes with v32.0.0.

This is probably a bug in ACTS Core, but not critical.

timadye:
Unless someone has a better idea, I will try to reproduce this issue in ActsExamples.

paulgessinger:
It might be a good idea to use the track state that is now passed to the branch stopper to count the holes the track has so far and just compare that with what the tip state claims.

timadye:
I checked whether I could reproduce the issue with standalone ACTS, but it seems to work fine - at least with the latest `main`. I tried introducing the same selections that Athena uses (#2934), but still no problem.
The problem is still there with Athena + ACTS v32.0.2, so unless it was fixed since v32.0.2, it is probably something in how the CKF is called from Athena.

timadye:
Now I am looking at the track state passed from Athena. I guess I can use `TrackStateProxy::typeFlags()` to see the holes. 
Once I found the `Acts::ConstTrackStateType` printout label listed the flags the wrong way round (I'll submit a PR fix), I can see the holes that are shown in the tip state counts.
I continue to investigate...

timadye:
OK, I think I see what's going wrong.

In Athena (at least), the `BranchStopper` is called several times after the last measurement. The additional track states are on the same branch (increasing `previous`, `pathLength`, and radius). They are flagged as either material or holes. Without the branch stopper, the returned trajectory ends with the last measurement, so the extra holes are not counted when checked outside the CKF.

[...]

PR #2950, opened 2024-02-14
fix: TrackStateType printout in wrong order

timadye:
`operator<<(std::ostream& os, ConstTrackStateType t)` prints out the flag bits as a binary number, with bit 0 (`MeasurementFlag`) at the end. This is belied by the label `MPOHMS`, which should be `SMHOPM`.

This PR fixes the label. Another option would be to reverse the order of bit printout. Please comment if you think that would be better.


timadye:
This was found while investigating #2905

andiwand (approved):
lgtm

Core/include/Acts/EventData/TrackStateType.hpp
```diff
@@ -115,8 +115,8 @@ class ConstTrackStateType {
     for (std::size_t i = 0; i < TrackStateFlag::NumTrackStateFlags; i++) {
       trunc[i] = bs[i];
     }
-    // MeasurementParameterOutlierHoleMaterialSharedhit
-    os << "MPOHMS=" << trunc;
+    // SharedhitMaterialHoleOutlierParameterMeasurement
```
paulgessinger, line 118: Ok I see.
Should this print the literal byte content, or should it print the ordering in the enum though?

timadye: I'd be happy to do it the other way. This way seemed the simplest unless someone had an opinion one way or the other.

andiwand: I have no opinion on the order as long as it is consistent 😄 anyways if somebody sees a problem we can always change it

paulgessinger: Let's merge it as is.

PR #2951, opened 2024-02-15
refactor: Remove `VertexFitterConcept`

paulgessinger:
This is not currently useful, as we don't actually have a generic vertex fitter interface to begin with.

Part of:
- #2842

Blocked by:
- #2948


paulgessinger:
Rolled into #2948 after refactoring.

PR #2904, opened 2024-01-26
refactor: Central truth matching for tracks in Examples

andiwand:
An attempt to move to a central truth matching in the Examples framework. Currently all our writers have a separate truth matching implementation with some amount of common code. I propose to move the truth matching to its own algorithm and then use the output as an input for the writers.

This also allows for simple truth matching in case of truth tracking to be specialized and for a common mechanism for how we infer the truth information. E.g. some algorithms right now expect particles and tracks in a 1:1 correspondence in the input containers which is not achievable with CKF tracking.

blocked by
- https://github.com/acts-project/acts/pull/2964

PR #2963, opened 2024-02-19
refactor: Remove `deltaR` from performance writers in Examples

andiwand:
While working on a centralized truth matching (https://github.com/acts-project/acts/pull/2904) I stumbled over `deltaR` in the performance writers which seems to be a measure of how close truth and reco are and based on that we match them.

To me that measure seems very arbitrary as we use an unweighted euclidean distance of phi and eta to calculate it. In https://github.com/acts-project/acts/pull/2904 I do not use this measure for truth tracking but only the association between measurements and truth particles.

In preparation of getting https://github.com/acts-project/acts/pull/2904 merged soon I split this change out so we can have a separate discussion and keep the physmon diff to a minimum.

### DIFF
--- a/Examples/Framework/include/ActsExamples/Validation/EffPlotTool.hpp
+++ b/Examples/Framework/include/ActsExamples/Validation/EffPlotTool.hpp
@@ -33,8 +33,7 @@ class EffPlotTool {
     std::map<std::string, PlotHelpers::Binning> varBinning = {
         {"Eta", PlotHelpers::Binning("#eta", 40, -4, 4)},
         {"Phi", PlotHelpers::Binning("#phi", 100, -3.15, 3.15)},
-        {"Pt", PlotHelpers::Binning("pT [GeV/c]", 40, 0, 100)},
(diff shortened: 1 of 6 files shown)


paulgessinger:
This is a very common measure actually, so I'd probably argue we can keep it.

andiwand:
> This is a very common measure actually, so I'd probably argue we can keep it.

Interesting I didn't come across this yet. Do you know in what context this is used?

What does `deltaR` stand for? First I confused it with radius but it is actually just angles in different units.

Since this is strictly using truth information it should actually not effect https://github.com/acts-project/acts/pull/2904

paulgessinger:
It's very frequently used as sort of angular distance, it's kind of like cone radius between pointing objects. 

One super common context is tracks in jets, where you measure efficiency or resolution of tracks vs $\Delta R$ between the track and the jet axis. 

It's also used (in ATLAS) for overlap removal, to avoid double counting signals in the detector (e.g. an electron AND a calorimeter signal)

PR #2965, opened 2024-02-19
refactor: Change the masking of the Geometry Identifiers

dimitra97:
With this PR I make an attempt to change the masking of the geometry Identifiers. This comes from the realization that for the muon spectrometer the number of volumes is bigger than the one allowed in the identifiers. A suggestion would be to move 4 bits from the `kApproachMask` to the `kVolumeMask.`


paulgessinger:
Fyi @andiwand @asalzburger @noemina

paulgessinger:
Strange, I can't reproduce the failure locally 🤔

dimitra97:
> Strange, I can't reproduce the failure locally 🤔

Ok that's strange. Also,  I do not know if this is the case but I see in the` odd-material-mapping-config.json` that e.g the `'RadialBounds' ` has a `geoId =  72339069014638592` (line 20). By converting it to binary this is `0000000100000001000000000000000000000000000000000000000000000000` . I guess this will use the `kBoundaryMask=0x000ff00000000000` (not sure if it works like that) so the value of geoId would need more 5 zero bits shifted?

PR #2969, opened 2024-02-21
feat: Write truth matching in `VertexPerformanceWriter`

andiwand:
Writes primary and secondary vertex number after matching to truth.

I also reorganized the branches and the order of writing them to follow reco -> truth.

blocked by
- https://github.com/acts-project/acts/pull/2970

PR #2976, opened 2024-02-21
fix: Match particles without cutting on track weight in `VertexPerformanceWriter`

andiwand:
Removed `trk.trackWeight > m_cfg.minTrkWeight` from the matching code as this is checked later anyways and the log afterwards is fired in cases where we can actually match the track.


andiwand:
the cut is actually relevant but should not mess with the truth matching. I integrated this into https://github.com/acts-project/acts/pull/2969 now

PR #2980, opened 2024-02-21
refactor: Safety check for spherical coordinates poles 

ssdetlab:
Adding a small check for zero values of the `SinTheta` inside the `evaluateTrigonomics` of the `VectorHelpers`. It was causing the NaNs in the update Jacobians inside the KF when the geometry is oriented along z-axis and the initial momentum estimate is directed as (0,0,1).


andiwand:
I fear that won't solve the problem with `(0,0,1)` as we assume valid `(phi, theta)` angles across the whole code base. Spherical coordinates will always have problems with the poles if derivatives matter.

Usually this is solved by rotating the tracking geometry so `(0,0,1)` can be avoided.

ssdetlab:
> I fear that won't solve the problem with `(0,0,1)` as we assume valid `(phi, theta)` angles across the whole code base.

Seemed to work for our case. The starting vector was (0,0,1) and the initial step was performed without any issues. After that the magnetic field/multiple scattering kicked in and the pole was no longer a problem.

andiwand:
I agree, the whole situation with `(0,0,1)` far from optimal. I just don't want to put in partial solutions to this where it then works for a specific case (like exactly `(0,0,1)`) but not any other case in the vicinity (e.g. `(~0,~0,~1)`).

This problem does not exist with free parameters since we use a direction vector instead of spherical coordinates. But any kind of conversion global->local or local->global will have this problem because the bound representation is using spherical coordinates.

But that is just my opinion. Maybe @paulgessinger or @asalzburger think differently about this.

beomki-yeo:
I want to know how the KF with (0,0,1) direction is causing problem. :thinking: 
If I am not wrong, we need the inverse of sin theta only for global -> bound transform.

Thing is that we are also supposed to call global -> bound transform when the track hits the material, where the direction is already scattered from (0,0,1). So technically the pole should not cause the problem unless the scattering angle is zero

andiwand:
Right scattering only happens during simulation so it will not effect the KF by changing direction but only by increasing the covariance.

`(phi, theta)` are inherently unstable at the vicinity of `(0,0,1)` since a small change in direction can make a big change in angles. I think transforming covariances in this region will be unstable in both directions, local to global and global to local.

ssdetlab:
> I think transforming covariances in this region will be unstable in both directions, local to global and global to local.

Yes, should be the case. The way the trigonomics are calculated right now, both `cosPhi` and `sinPhi` are dependent on the `sinTheta` inversion. This will propagate to both `freeToSphericalDirectionJacobian` and `sphericalToFreeDirectionJacobian`.

The numerical stability is indeed something to test. In my understanding, the problem is going to be hard to get rid of without introducing some additional algorithm (which can be expensive to propagation), because of the way the track is parametrized. Please, correct me if I'm wrong.

I propose to settle down for an explicit warning that tells you about the proximity to `(0,0,1)` for now. Like I mentioned above, the NaNs sent me on the investigation across ~10 files, so I'd want to prevent the scenarios like that.

andiwand:
can you update the pr title @ssdetlab ?

Tests/UnitTests/Core/Surfaces/PlaneSurfaceTests.cpp
```diff
@@ -271,22 +271,32 @@ BOOST_AUTO_TEST_CASE(PlaneSurfaceAlignment) {
...
   Translation3 translation{0., 1., 2.};
-  auto pTransform = Transform3(translation);
+  RotationMatrix3 rotation = RotationMatrix3::Identity();
+  double rotationAngle = M_PI_2;
+  Vector3 xPos(cos(rotationAngle), 0., -sin(rotationAngle));
+  Vector3 yPos(0., 1., 0.);
+  Vector3 zPos(sin(rotationAngle), 0., cos(rotationAngle));
+  rotation.col(0) = xPos;
+  rotation.col(1) = yPos;
+  rotation.col(2) = zPos;
```
andiwand, line 281: this can also be done with `AngleAxis` which might be more expressive

Tests/UnitTests/Core/Surfaces/PlaneSurfaceTests.cpp
andiwand, line 1: I guess the tests you modify here have some axis alignment which is problematic?

ssdetlab: Yes, the changed tests assumed z-directed momenta. The changes boil down to excluding spherical poles from the tests (as it was already done for the Cylindrical geometries, btw), and rotating the geometry in the  `PlaneSurface` tests

Tests/UnitTests/Core/Surfaces/PlaneSurfaceTests.cpp
ssdetlab: Checked `AngleAxis`. Indeed, looks much more intuitive. Commited changes

acts-project-service:
## :white_check_mark: Athena integration test results [a1b40bc16ba0bed854c5158f0cdac110f74ffa05]

|status|job|report|
|------|---|------|
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684334)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684333)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684332)||
|:green_circle: | [test_ActsConversionWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684331)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684330)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684329)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684328)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684327)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684326)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684325)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684324)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684323)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684322)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684321)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684320)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684319)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684318)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684317)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684316)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684315)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684314)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684313)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684311)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684310)||
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36684308)||

PR #2983, opened 2024-02-22
test: Test field invariance of direction uncertainty transport

andiwand:
This is testing a suspicion I had about direction uncertainty transported with and without magnetic field. The covariance of the direction does not care if it is transported in a straight line or along a curved track. This allows us to apply MSC effects after traversing the material by only remembering at which distances they occurred. Afterwards the covariance can be treated as an additional term to the final propagation covariance.


andiwand:
@beomki-yeo @paulgessinger fyi

andiwand:
Hm the SL propagator is just undoing it's own work right now so this does not prove anything yet. I think I have to introduce an additional direction uncertainty for both cases before propagating.

beomki-yeo:
I am confused by the title. Doesn't direction uncertainty increase with the presence of the magnetic field?

paulgessinger:
This has picked up conflicts. What's the plan for this @andiwand?

andiwand:
I will try to follow up with this in the next few weeks

PR #2985, opened 2024-02-23
fix: Tweak AMVF config with time in Examples

andiwand:
I observed higher vertex efficiency with time setting `maxMergeVertexSignificance` to default. `tracksMaxSignificance` and `doFullSplitting` were also reset in order to use the same config with and without time.

PR #2986, opened 2024-02-23
fix: Tweak AMVF config `tracksMaxSignificance` with time in Examples

andiwand:
Similar to https://github.com/acts-project/acts/pull/2985 we would like to see how this drives the performance.


andiwand:
integrated changed into https://github.com/acts-project/acts/pull/2985

PR #2991, opened 2024-02-27
fix: Tweak AMVF config `do3dSplitting` with time in Examples

andiwand:
Similar to https://github.com/acts-project/acts/pull/2985 we would like to see how this drives the performance.

blocked by
- https://github.com/acts-project/acts/pull/2828


andiwand:
integrated changes into https://github.com/acts-project/acts/pull/2985

PR #3001, opened 2024-02-29
refactor: Refactor wiring for number of hits in simulation

andiwand:
Currently we recalculate the number of hits per particles in different places. Here I try to unify this by outputting the number of hits from the simulation and attaching it to the final particle of the simulation.


Examples/Io/Root/src/RootParticleWriter.cpp
```diff
@@ -105,18 +102,16 @@ ActsExamples::ProcessCode ActsExamples::RootParticleWriter::finalize() {
 
 ActsExamples::ProcessCode ActsExamples::RootParticleWriter::writeT(
     const AlgorithmContext& ctx, const SimParticleContainer& particles) {
+  const SimParticleContainer* finalParticles = nullptr;
+  if (m_inputFinalParticles.isInitialized()) {
+    finalParticles = &m_inputFinalParticles(ctx);
+  }
+
```
AJPfleger, line 109: Could we move this block down, next to
```
    bool wroteFinalParticle = false;
    if (finalParticles != nullptr) {
```
?

andiwand: the other section is in a loop and we usually query all the whiteboard entries at the beginning of the algorithm

AJPfleger (approved):

PR #3002, opened 2024-02-29
fix: Wire optional seed result in `TrackDensityVertexFinder` after #2917

andiwand:
In https://github.com/acts-project/acts/pull/2917 I changed the way how "no more seeds" is communicated in our vertexing code but forgot to wire this for the `TrackDensityVertexFinder` correctly.


acts-project-service:
## :red_circle: Athena integration test results [f64f72672685dae9f1d78629e3141e9511d1f8dc]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/36594961) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/6980452)!

paulgessinger:
I think the output failures indicate that we're going back to the original outputs. I'd be inclined to either merge this and then manually remove the reference overrides (thus going back to the vanilla references active in Athena main) or doing that before merging and triggering here again. In any case, it's a bit of a manually timed operation.

Core/src/Vertexing/GaussianTrackDensity.cpp
```diff
@@ -58,15 +59,25 @@ std::pair<double, double> Acts::GaussianTrackDensity::globalMaximumWithWidth(
...
-  return (maxSecondDerivative == 0.)
-             ? std::make_pair(0., 0.)
-             : std::make_pair(maxPosition,
-                              std::sqrt(-(maxDensity / maxSecondDerivative)));
+  if (maxSecondDerivative == 0.) {
+    return std::nullopt;
+  }
+
+  return std::make_pair(maxPosition,
+                        std::sqrt(-(maxDensity / maxSecondDerivative)));
```
paulgessinger, line 67: NIT, this with deduction guides, I think this can be 
```suggestion
  return std::pair{maxPosition,
                        std::sqrt(-(maxDensity / maxSecondDerivative))};
```

Core/src/Vertexing/GaussianTrackDensity.cpp
```diff
@@ -14,11 +14,12 @@
 
 namespace Acts {
 
-std::pair<double, double> Acts::GaussianTrackDensity::globalMaximumWithWidth(
+Result<std::optional<std::pair<double, double>>>
```
paulgessinger, line 17: This is a bit of an awkward type, but I don't see a better alternative either.

paulgessinger (dismissed):

Core/src/Vertexing/GaussianTrackDensity.cpp
andiwand: agreed - we have this a couple of times now in the vertexing but I also don't see a way around

andiwand:
> I think the output failures indicate that we're going back to the original outputs. I'd be inclined to either merge this and then manually remove the reference overrides (thus going back to the vanilla references active in Athena main) or doing that before merging and triggering here again. In any case, it's a bit of a manually timed operation.

What suits better for you? From my side both is fine. If you revert the output changes we can just make this PR the next to go in

paulgessinger (approved):
Let's go!

PR #3005, opened 2024-03-04
feat: adding relative transformation + calculation of com function

LaraCalic:
`edit:` remove reminder text from this description to optimise search

### DIFF
--- a/Tests/UnitTests/Examples/Algorithms/Alignment/AlignmentGroupingTests.cpp
+++ b/Tests/UnitTests/Examples/Algorithms/Alignment/AlignmentGroupingTests.cpp
@@ -8,7 +8,6 @@
 
 #include <boost/test/data/test_case.hpp>
 #include <boost/test/unit_test.hpp>
-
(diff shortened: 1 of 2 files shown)


Tests/UnitTests/Examples/Algorithms/Alignment/AlignmentGroupingTests.cpp
```diff
@@ -92,3 +91,26 @@ BOOST_AUTO_TEST_CASE(Multiple_test) {
...
+
+  // Calculate center of mass
+  Eigen::Vector3d calculatedCOM = ActsExamples::calculateCenterOfMass(surfaces, gctx);
+
+  BOOST_CHECK(calculatedCom.isApprox(expectedCom, 1e-3)); 
+
+  
+  for (auto* surface : surfaces) {
+    delete surface;
+  }
```
AJPfleger, line 115:
```suggestion
```
Is there a strong reason, why you do a manual clean-up? At the end of the scope the surfaces will be removed anyway.

PR #3009, opened 2024-03-05
refactor: Unconditionally use blocked mult in EigenStepper

paulgessinger:
Just to try if this also returns clean outputs, I'm downgrading the `if` added in #1181 to an `assert`. Let's see what the CI says here.


stephenswat:
Nice!

@andiwand do you have any performance numbers for this?

paulgessinger:
@stephenswat I can run a quick test.

paulgessinger:
This is from 10 runs:

image

Core/include/Acts/Propagator/EigenStepper.ipp
```diff
@@ -250,24 +250,20 @@ Acts::Result<double> Acts::EigenStepper<E, A>::step(
...
-      // For safety purposes, we provide a full matrix multiplication as a
-      // backup strategy.
-      state.stepping.jacTransport = D * state.stepping.jacTransport;
-    }
+    assert((D.topLeftCorner<4, 4>().isIdentity()) &&
+           (D.bottomLeftCorner<4, 4>().isZero()) &&
+           (state.stepping.jacTransport.template topLeftCorner<4, 4>()
+                .isIdentity()) &&
+           (state.stepping.jacTransport.template bottomLeftCorner<4, 4>()
+                .isZero()));
```
andiwand, line 258: nitpick: should we make multiple asserts for this?

andiwand (approved):
very interesting! I thought the assert would fire

Core/include/Acts/Propagator/EigenStepper.ipp
paulgessinger: Can do

paulgessinger:
image

More stats confirm the measurement from before, I think we're looking at about a 5% speedup.

paulgessinger:
Athena is green now, should we get this in? @andiwand

andiwand (approved):
lets go

PR #818, opened 2021-05-27
chore: Update release script

paulgessinger:
This is the release script I've been using for about half a year. This doesn't run automatically at all, and right now I don't think it's very important as I can just run it manually. This PR is to add this version to the main branch, so I have to do less switching around when making a release.

PR #3011, opened 2024-03-06
chore: Avoid double listing breaking changes

paulgessinger:


CI/release.py
```diff
@@ -131,14 +131,6 @@ def generate_changelog(commits, commit_parser=_default_parser) -> dict:
         )
         changes[message.type].append((commit.sha, capital_message, commit.author))
 
-        if message.breaking_descriptions:
-            for paragraph in message.breaking_descriptions:
-                changes["breaking"].append((commit.sha, paragraph, commit.author))
-        elif message.bump == 3:
```
AJPfleger, line 137: What was the idea behind this check? (added in #818)

paulgessinger: I think the logic that identified which changes were breaking was different.

AJPfleger (approved):

PR #3012, opened 2024-03-06
feat: harmonize local and global grid access

asalzburger:
This PR harmonises the way the grid entries are accessed using local and global positions.

Several classes such as `GridSurfaceMaterial` and to some limit also the `IndexedSurfaces` may require lookup in an attached grid from a *global* and *local* frame position.

To translate the global lookup and local lookup into the grid frame (which could be less dimensional, e.g. 1-DIm on a 2-DIM surface), or re-ordered (flipped axes, e.g.), or even in a different definition (a phi-z grid on a Cylinder with local coordinates `[r*Phi, z]` those classes have a global and local casting mechanism which utilises the `GridAccessHelper`.

This PR mirrors the global casting to also be working the local casting, in order to e.g. `ISurfaceMaterial` local lookup interface.

Adding @dimitra97  - as I had to slightly update code from the MultiWire builders


dimitra97:
Looks interesting. So in the `MultiWire Builders` and the `Multi Layer Surfaces Updater `the grid frame is assumed to be the same with the local/volume's frame. And in the `Updater` I do the cast using directly the local position, according to a PR I opened yesterday: https://github.com/acts-project/acts/pull/3010
I guess in this case is the same if I leave it as it is, right? However, I can try adapt the usage of the functionality you introduce here :)

asalzburger:
> Looks interesting. So in the `MultiWire Builders` and the `Multi Layer Surfaces Updater `the grid frame is assumed to be the same with the local/volume's frame. And in the `Updater` I do the cast using directly the local position, according to a PR I opened yesterday: #3010 I guess in this case is the same if I leave it as it is, right? However, I can try adapt the usage of the functionality you introduce here :)

Yes, you can leave everything as is. This was just for your information.

asalzburger:
Hey @dimitra97  - since you looked over it already, do you mind doing a review, please ?

asalzburger:
> I'm wondering:
> 
> we've have the exact same problem (I think) for the B field lookup. We used to have a fairly complex type-erased explicit field mapper structure, that I ended up removing in favor of a function to convert the 3-dim query position to an N-dim lookup position in the internal coordinate frame of the interpolation domain (rz, xyz, for example).
> 
> Can't we do a very similar thing here? This whole "cast" mechanism spreading more or less everywhere seems fairly heavy for what we're trying to accomplish.
> 
> If I understand what's happening here, would an alternative be to have `ProtoBinning` and `BinningDescription` (which is a vector of `ProtoBinning`, I'm assuming 1 per axis) be the dynamic structure thats built at runtime, and then provide a function that turns this in to a compile-time grid with all ifs/elses in a single place?
> 
> If that function erases the concrete grid type with a lookup conversion function that always takes 2D inputs (like is needed for material lookup), that could become a single return type, similar to `InterpolatedMagneticField` (which like mentioned before, I think, does the same thing)

Hi @paulgessinger - the intention here is:

If you are on the optimal frame (i.e. local frame is grid frame), then this becomes purely an access method.

I.e. a LocalPosition gets translated to a grid_point type (and depending if you have a 1-D measurement you still can address which dimension is actually binned). In that case there's minimal overhead, it will create the correct dimensioned point and fill in the grid point (even from global it will be just an access).

It would still allow you to put a 1D grid on a plane, binned in Y though, without any additional operation rather than creation and filling, the unrolling of the filling loop should be done at compile time, so we do not pay any function call. 

My main argument agains if/else cascades is that both the dimension of the grid_point and the source access (is it the first local parameter, the second one, both) can vary & some may or may not have manipulations - so it quickly becomes a very long cascade.

paulgessinger:
About the if/else cascades: we have them anyway, but they're currently spread widely across the code. 

My argument about "direct access" and unrolling of dimension lookups here is that it's vastly more complex, and in the case of the B field it had a negligible performance benefit. You remember I was able to drastically simplify the B field code when I removed. It does exactly the same thing: go from global 3D to an N dimensional lookup domain. 

One of two function calls, which can potentially even be inlined, can abstract almost all of this, while (I think) reducing complexity and number of needed structs, typedefs and functions considerably.

asalzburger:
After discussions, this will be superseded.

PR #3037, opened 2024-03-13
refactor: TrackingVolume and VolumeBounds mutability

paulgessinger:


Core/include/Acts/Geometry/Volume.hpp
```diff
@@ -67,20 +63,22 @@ class Volume : public virtual GeometryObject {
...
   /// @param envelope Optional envelope to add / subtract from min/max
   /// @return Constructed bounding box pointing to this volume
   BoundingBox boundingBox(const Vector3& envelope = {0, 0, 0}) const;
 
   /// Construct oriented bounding box for this shape
+  /// @note This will build an oriented bounding box with an
+  ///       envelope value of (0.05, 0.05, 0.05)mm
   /// @return Constructed oriented bounding box pointing to this volume
-  const BoundingBox& orientedBoundingBox() const;
+  BoundingBox orientedBoundingBox() const;
```
asalzburger, line 81: Is the bounding box now created on the fly and not stored?

Core/include/Acts/Geometry/Volume.hpp
```diff
@@ -105,7 +103,6 @@ class Volume : public virtual GeometryObject {
   Transform3 m_itransform;
   Vector3 m_center;
   std::shared_ptr<const VolumeBounds> m_volumeBounds;
-  BoundingBox m_orientedBoundingBox;
```
asalzburger, line 108: I see.

Core/src/Geometry/CuboidVolumeBuilder.cpp
```diff
@@ -186,14 +186,14 @@ std::shared_ptr<Acts::TrackingVolume> Acts::CuboidVolumeBuilder::buildVolume(
...
-                               nullptr, cfg.trackingVolumes, cfg.name);
+    trackVolume = std::make_shared<TrackingVolume>(
+        trafo, bounds, cfg.volumeMaterial, nullptr, nullptr,
+        cfg.trackingVolumes, cfg.name);
   } else {
     // Build TrackingVolume
-    trackVolume = TrackingVolume::create(trafo, bounds, cfg.volumeMaterial,
-                                         std::move(layArr), nullptr,
-                                         cfg.trackingVolumes, cfg.name);
+    trackVolume = std::make_shared<TrackingVolume>(
```
asalzburger, line 194: Ok,  makes sense, we just use the `std::make_shared<...>`

Core/src/Geometry/CylinderVolumeBounds.cpp
```diff
@@ -20,13 +21,36 @@
 #include "Acts/Utilities/BoundingBox.hpp"
 
 #include <cmath>
-#include <type_traits>
```
asalzburger, line 23: I guess this was just a leftover?

Tests/UnitTests/Core/Geometry/CuboidVolumeBoundsTests.cpp
```diff
@@ -176,6 +176,32 @@ BOOST_AUTO_TEST_CASE(CuboidVolumeBoundarySurfaces) {
   BOOST_CHECK(pFaceZX.col(2).isApprox(yaxis));
 }
 
+BOOST_AUTO_TEST_CASE(CuboidVolumeBoundsSetValues) {
```
asalzburger, line 179: This is the first set of the new style tests, I suppose.

asalzburger (approved):
Looks good.

Core/include/Acts/Geometry/Volume.hpp
paulgessinger: It is, because I wanted to avoid having to also refresh it when volume bounds are reset. It's also a non-zero footprint, that we maybe don't want to store for every volume, but rather in the acceleration structure using it. I would imagine we don't call this function very often.

Core/src/Geometry/CuboidVolumeBuilder.cpp
paulgessinger: Because there's no need to guarantee shared_ptr ownership for TrackingVolume like we do for Surface.

Core/src/Geometry/CylinderVolumeBounds.cpp
paulgessinger: It's unused in any case, yeah.

Tests/UnitTests/Core/Geometry/CuboidVolumeBoundsTests.cpp
paulgessinger: This just tests the setting of values. I'm planning to add construction tests when the functionality is actually added, like for the cylinder volume stack.

acts-project-service:
## :red_circle: Athena integration test results [d411085e682e11667b443a925bd3d7c448f152dc]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37452105) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7119331)!

PR #3042, opened 2024-03-19
fix: change the order in geant4 converters transform

dimitra97:
I think in the `Geant4Converters` when we apply the transform, the rotation should be first in the code and then the translation. Also, I noticed that the opposite breaks the transform of the straw surfaces of the muon mockup detector.
@asalzburger @ssdetlab

### DIFF
--- a/Plugins/Geant4/src/Geant4Converters.cpp
+++ b/Plugins/Geant4/src/Geant4Converters.cpp
@@ -59,8 +59,8 @@ Acts::Transform3 Acts::Geant4AlgebraConverter::transform(
   rotation << g4Rot.xx(), g4Rot.yx(), g4Rot.zx(), g4Rot.xy(), g4Rot.yy(),
       g4Rot.zy(), g4Rot.xz(), g4Rot.yz(), g4Rot.zz();
   Transform3 transform = Transform3::Identity();
-  transform.pretranslate(translation);
   transform.prerotate(rotation);
+  transform.pretranslate(translation);  
   return transform;
 }
 
--- a/Plugins/Geant4/src/Geant4Converters.cpp
+++ b/Plugins/Geant4/src/Geant4Converters.cpp
@@ -60,7 +60,7 @@ Acts::Transform3 Acts::Geant4AlgebraConverter::transform(
       g4Rot.zy(), g4Rot.xz(), g4Rot.yz(), g4Rot.zz();
   Transform3 transform = Transform3::Identity();
   transform.prerotate(rotation);
-  transform.pretranslate(translation);  
+  transform.pretranslate(translation);
   return transform;
 }
 


andiwand:
This breaks a unit test - I cannot judge what is more correct but maybe the unit test reveals something

dimitra97:
> This breaks a unit test - I cannot judge what is more correct but maybe the unit test reveals something

Hm yes I saw. Maybe for the telescope detector we need the opposite order (but I cannot tell why). But for the straws, they have rotation 90deg on y-axis and if I first translate and then rotate they will fall in the same z position. 
Also I noticed that if the sign in x rotation in this unittest is omitted, then it is ok.


```
"x=\"-90\" y=\"0\" z=\"0\"/>"
        << '\n';
  bgdml << "              </physvol>" << '\n';
  bgdml << "    </volume>" << '\n';
  bgdml << "    <volume name=\"wl\">" << '\n';
  bgdml << "         <materialref ref=\"G4_Galactic\"/>" << '\n';
  bgdml << "         <solidref ref=\"wb\"/>" << '\n';
  bgdml << "             <physvol name=\"cl_pv\">" << '\n';
  bgdml << "                    <volumeref ref=\"cl\"/>" << '\n';
  bgdml << "                    <rotation name=\"cl_pv_rot\" unit=\"deg\" "
           "x=\"-90\" y=\"0\" z=\"0\"/>"
        << '\n';
```

`

 I do not know if we really want -90 deg rotation along x though.

andiwand:
there should definitely be a way that works for all cases. maybe @asalzburger @noemina can help. otherwise I can have a look next week

noemina:
I think rotating and then translating makes more sense.

asalzburger:
> > This breaks a unit test - I cannot judge what is more correct but maybe the unit test reveals something
> 
> Hm yes I saw. Maybe for the telescope detector we need the opposite order (but I cannot tell why). But for the straws, they have rotation 90deg on y-axis and if I first translate and then rotate they will fall in the same z position. Also I noticed that if the sign in x rotation in this unittest is omitted, then it is ok.
> 
> ```
> "x=\"-90\" y=\"0\" z=\"0\"/>"
>         << '\n';
>   bgdml << "              </physvol>" << '\n';
>   bgdml << "    </volume>" << '\n';
>   bgdml << "    <volume name=\"wl\">" << '\n';
>   bgdml << "         <materialref ref=\"G4_Galactic\"/>" << '\n';
>   bgdml << "         <solidref ref=\"wb\"/>" << '\n';
>   bgdml << "             <physvol name=\"cl_pv\">" << '\n';
>   bgdml << "                    <volumeref ref=\"cl\"/>" << '\n';
>   bgdml << "                    <rotation name=\"cl_pv_rot\" unit=\"deg\" "
>            "x=\"-90\" y=\"0\" z=\"0\"/>"
>         << '\n';
> ```
> 
> `
> 
> I do not know if we really want -90 deg rotation along x though. But in general, my concern is that we wouldn't want a mechanism that would work for all the cases?

Yes, it has to work in this case as well ... this was an actual case in NA60+

ssdetlab:
We raised this question with Andreas. I think there has to be a general agreement on the order inside the code base, because right now it does not look consistent. 

I believe rotating and then translating follows the Eigen convention, so maybe we should stick to that. But then some refactoring has to be done. From what I saw, a lot of code decouples translation and rotation as separate calls where order matters. 

Should we somehow enforce the usage of the constructor only for the transforms?

andiwand:
What I am confused about is that this is G4 conversion code no? And G4 surly has a defined transformation order.

If so we just need to match that in our code and we should be fine no?

ssdetlab:
The looked into the problem a bit deeper. The issue seems stem from the way the `Geant4SurfaceFactory` handles the relative transformation propagation to the global frame. 

Right now the mother transformation is multiplied with the daughter one, but it looks like the it's not the correct way of shifting the daughter into the global frame. From a simple test with a mockup Geant4 geometry, it looks like you have to manually handle the translations rotation and keep track of the inversion. 

Ican run the tests locally and prepare the PR.

PR #3060, opened 2024-03-26
feat: Add TransformRange helper

paulgessinger:
This implements a transforming range adapter, similar to what will be available with the C++20+ ranges library, but much less powerful.

I'm planning to use this to gate mutable access to inner structures like child volumes in `TrackingVolume`, by using the `ConstDereference` policy for this range helper, which will always dereference and convert to const.


Core/include/Acts/Utilities/TransformRange.hpp
```diff
@@ -0,0 +1,133 @@
...
+          bool force_const>
+struct TransformRangeIterator {
+ private:
+  using internal_value_type = typename iterator_t::value_type;
+
+ public:
+  using value_type = std::conditional_t<
+      force_const,
+      decltype(std::as_const(
+          std::declval<typename Callable<internal_value_type>::value_type>())),
```
niermann999, line 43: I would tend to use ```declval<typename Callable<internal_value_type>::value_type &>``` (this is at least what I saw used most and I think its otherwise an r-value or smth? Don't know how that works with ```as_const```)

Core/include/Acts/Utilities/TransformRange.hpp
```diff
@@ -0,0 +1,133 @@
...
+      typename Callable<internal_value_type>::value_type>;
+
+  using iterator = TransformRangeIterator<
+      Callable,
+      std::conditional_t<std::is_const_v<container_t>,
+                         typename container_t::const_iterator,
+                         typename container_t::iterator>,
+      std::is_const_v<container_t>>;
+  using const_iterator =
+      TransformRangeIterator<Callable, typename container_t::const_iterator,
```
niermann999, line 88: Does this not already make everything const just because ```const_iterator``` dereferences to const?

Core/include/Acts/Utilities/TransformRange.hpp
```diff
@@ -0,0 +1,133 @@
...
+                             true>;
+
+  explicit TransformRange(container_t& container) : m_container(&container) {}
+
+  value_type operator[](std::size_t i) {
```
niermann999, line 93: Would also just put ```decltype(auto)``` here, but not sure

Core/include/Acts/Utilities/TransformRange.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+  iterator end() { return iterator{m_container->end()}; }
+
+  const_iterator begin() const { return const_iterator{m_container->begin()}; }
+
+  const_iterator end() const { return const_iterator{m_container->end()}; }
+
+  std::size_t size() const { return m_container->size(); }
+  bool empty() const { return m_container->empty(); }
+
```
niermann999, line 121: Do you want to add ```cbegin()```  for this class? The other stuff  ```back(), front()``` etc is probably not so important

Core/include/Acts/Utilities/TransformRange.hpp
```diff
@@ -0,0 +1,133 @@
...
+  bool operator==(const TransformRangeIterator& other) const {
+    return m_iterator == other.m_iterator;
+  }
+
+ private:
+  iterator_t m_iterator;
+};
+
+template <template <typename> typename Callable, typename container_t>
```
niermann999, line 69: The idea is to use this only on containers not more generally on other ranges?

Core/include/Acts/Utilities/TransformRange.hpp
paulgessinger: I'm forcing this to always be a reference in that case right? Is it a requirement on iterators that they return references when being dereferenced?

Core/include/Acts/Utilities/TransformRange.hpp
paulgessinger: I mean I guess having it be more generic would be nice. I want to expose `size`, `empty` and indexed access. I could implement this on top of an iterator pair and rely on operations on them, but wasn't sure if that's worth it here.

Core/include/Acts/Utilities/TransformRange.hpp
paulgessinger: I don't think so, since the iterator itself can point at a pointer, which in turn might be mutable. 

In that case I still want to always dereference the target pointer to const. 

But I also want this to work with outer const references as well, like `const std::vector<const int*>>`, which is why I think I need to list the `const_reference` variants as well.

Core/include/Acts/Utilities/TransformRange.hpp
```diff
@@ -0,0 +1,133 @@
...
+template <template <typename> typename Callable, typename iterator_t,
+          bool force_const>
+struct TransformRangeIterator {
+ private:
+  using internal_value_type = typename iterator_t::value_type;
+
+ public:
+  using value_type = std::conditional_t<
+      force_const,
+      decltype(std::as_const(
```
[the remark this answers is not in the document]
paulgessinger: I think 
```cpp
const typename Callable<internal_value_type>::value_type
```
does not do anything, i.e. if 
```cpp
typename Callable<internal_value_type>::value_type // -> int&
```
then
```cpp
const typename Callable<internal_value_type>::value_type // -> int&
```

Core/include/Acts/Utilities/TransformRange.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+namespace Acts {
+
+namespace detail {
+
+template <typename input_t>
+struct Dereference {
+  using value_type = decltype(*std::declval<input_t>());
+  constexpr static value_type apply(input_t& value) { return *value; }
+  constexpr static value_type apply(const input_t& value) { return *value; }
```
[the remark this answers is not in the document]
paulgessinger: Ok, I think if I rewrite this to use a method template with deduced parameter and universal-reference, I don't need separate overloads.

Core/include/Acts/Utilities/TransformRange.hpp
```diff
@@ -0,0 +1,133 @@
...
+};
+
+template <template <typename> typename Callable, typename iterator_t,
+          bool force_const>
+struct TransformRangeIterator {
+ private:
+  using internal_value_type = typename iterator_t::value_type;
+
+ public:
+  using value_type = std::conditional_t<
```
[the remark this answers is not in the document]
paulgessinger: I guess the iterator category is whatever the underlying iterator is, right?

Core/include/Acts/Utilities/TransformRange.hpp
niermann999: I was just wondering if the ```force_const``` was needed for ```const_iterator```, but you are right, I don't think this propagates constness such that we get pointers to const

Core/include/Acts/Utilities/TransformRange.hpp
niermann999: Does that mean, the responsibility for guaranteeing constness lies partly with the Range and partly with the ```Dereference``` struct?

paulgessinger:
@niermann999 I've simplified (I think) the typedefs, and straightened out (hopefully) the type deductions. Could you have another look?

Core/include/Acts/Utilities/TransformRange.hpp
niermann999: I don't know of any such requirement. The enumerate view, I think, returns an r-value. Does ```as_const``` not also make it a reference (I am not sure what it does precisely)?

Core/include/Acts/Utilities/TransformRange.hpp
```diff
@@ -0,0 +1,133 @@
...
+#pragma once
+
+#include <type_traits>
+#include <utility>
+
+namespace Acts {
+
+namespace detail {
+
+template <typename input_t>
```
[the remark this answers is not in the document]
niermann999: OK, I see

Core/include/Acts/Utilities/TransformRange.hpp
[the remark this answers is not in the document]
niermann999: No, I think for that you would need to make sure that ```TransformRangeIterator``` iterator can expose all functionality of ```iterator_t```. So, I think for that you would need to conditionally define more methods like ```operator+, operator-, operator--``` etc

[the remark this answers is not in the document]
paulgessinger: Fair enough.

Core/include/Acts/Utilities/TransformRange.hpp
niermann999: Ah, wait, no ```Derefernce``` is used *only* to abstract if the ```iterator_t``` value type is a pointer, and the range has to guarantee "deep" constness for pointers?

Core/include/Acts/Utilities/TransformRange.hpp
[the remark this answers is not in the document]
niermann999: I don't think it is necessary either. Just add the forward iterator category tag and all client code *should* know not to try to use any of these operators

Core/include/Acts/Utilities/TransformRange.hpp
paulgessinger: I think `as_const` does you're right. I've gotten rid of that in this spot now.

Core/include/Acts/Utilities/TransformRange.hpp
paulgessinger: That's the idea at least. With the callables now using universal references, I think they shouldn't change the constness of the inputs at all anymore.

Core/include/Acts/Utilities/TransformRange.hpp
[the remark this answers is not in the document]
paulgessinger: I've added a `DotGet` struct now that calls `.get()` and added a test for `std::reference_wrapper`. This works just as well, so as long as the relationship between the input type and the output type is *dereference-like* this seems to be working fine.

I don't think we need this to do more.

Core/include/Acts/Utilities/TransformRange.hpp
niermann999: Ok, that makes sense. Will you add some comments in the final version?

Core/include/Acts/Utilities/TransformRange.hpp
paulgessinger: I guess I can just implement `cbegin` / `cend` like

```cpp
const_iterator cbegin() const { return begin(); }
```

Core/include/Acts/Utilities/TransformRange.hpp
paulgessinger: will do.

Core/include/Acts/Utilities/TransformRange.hpp
paulgessinger: Also, now that I'm thinking about it, it is in fact a requirements, returning rvalues will result in dangling references now. 

I don't think I need to generalize this a this point to also support rvalues at this time, however.

paulgessinger:
Added doc comments now.

Core/include/Acts/Utilities/TransformRange.hpp
```diff
@@ -0,0 +1,228 @@
...
+  constexpr static decltype(auto) apply(input_t&& value) {
+    return std::as_const(*value);
+  }
+};
+
+/// Callable that calls the @c get method of a value
+struct DotGet {
+  template <typename input_t>
+  constexpr static decltype(auto) apply(input_t&& value) {
+    return value.get();
```
niermann999, line 223: I don't know if it makes a difference, but maybe use ```std::forward```?

Core/include/Acts/Utilities/TransformRange.hpp
```diff
@@ -0,0 +1,228 @@
...
+
+#include <iterator>
+#include <type_traits>
+#include <utility>
+
+namespace Acts {
+
+namespace detail {
+
+template <typename Callable, typename iterator_t, bool force_const>
```
niermann999, line 19: One last thing that came to mind: Since the ```TransformRange``` name seems to suggest it is quite generic, why not make ```Callable``` a functor, instead of ducktyping the ```apply``` method and ```value_type```? You could then do:
```
std::invoke_result_t<Callable, internal_value_type>
```
or
```
std::invoke_result_t<Callable, decltype(std::declval<internal_value_type>())>
```

niermann999 (approved):
Looks good. Just minor suggestions

Core/include/Acts/Utilities/TransformRange.hpp
niermann999: I was a bit confused, because I had expected ```TransformRange``` to be generic and ```Dereference``` to bring the specific behaviour

Core/include/Acts/Utilities/TransformRange.hpp
paulgessinger: And then create `Callable{}` on the fly when I need to invoke it?

niermann999: Or hold it as a member and use ```std::invoke```? Idk, just a suggestion

paulgessinger: I think I'll keep `apply` for right now.

Core/include/Acts/Utilities/TransformRange.hpp
paulgessinger: Conclusion: we're not sure if that would do anything here.

Core/include/Acts/Utilities/TransformRange.hpp
paulgessinger: I didn't see a way to mediate this between the range, the iterator, and the callable. Maybe I'll come back to this in the future.

acts-project-service:
## :white_check_mark: Athena integration test results [2b51bcb1de16eaad87eb7e52c717df2915708e20]

|status|job|report|
|------|---|------|
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526418)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526417)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526416)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526415)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526414)||
|:green_circle: | [test_ActsConversionWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526413)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526412)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526411)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526410)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526409)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526408)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526407)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526406)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526405)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526404)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526403)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526402)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526401)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526400)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526399)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526398)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526397)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526396)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526395)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526394)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526393)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526392)||
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37526391)||

PR #3066, opened 2024-03-27
feat: Two-way CKF example

timadye:
Resurrect #2717 from @andiwand, updating to the latest `main`, and removing earlier test (`MyTrackFindingAlgorithm`).

blocked by
- https://github.com/acts-project/acts/pull/3074
- https://github.com/acts-project/acts/pull/3075
- https://github.com/acts-project/acts/pull/3076
- https://github.com/acts-project/acts/pull/3077
- https://github.com/acts-project/acts/pull/3078
- https://github.com/acts-project/acts/pull/3086


timadye:
Currently this has the default `twoWay = true` (in `TrackFindingAlgorithm.hpp` and `reconstruction.py`). Should we make it `false` to keep the old behaviour, or make this the new default?

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -145,27 +157,87 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
+          // We are excluding non measurement states and outlier here. Those can
+          // decrease resolution because only the smoothing corrected the very
+          // first prediction as filtering is not possible.
+          if (isMeasurement && !isOutlier) {
+            firstState = st;
+          }
+        }
+
+        Acts::BoundTrackParameters secondInitialParameters(
+            firstState->referenceSurface().getSharedPtr(),
```
andiwand, line 196: if a `firstState` is not found it will crash here no?

timadye: good point. I fixed it, pulled, fixed again, pulled, and then find you fixed it :)

andiwand: oh sorry 😅

timadye: NP 😎

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -159,14 +160,21 @@ Result<void> smoothTracks(
 template <typename track_proxy_t>
 Result<std::pair<typename track_proxy_t::ConstTrackStateProxy, double>>
 findTrackStateForExtrapolation(
-    const GeometryContext &geoContext, track_proxy_t &track,
+    const GeometryContext &geoContext, const track_proxy_t &track,
```
timadye, line 163: another API change that might count as a breaking change, but again probably won't actually break any real code.

Core/src/TrackFitting/GainMatrixSmoother.cpp
```diff
@@ -35,8 +35,16 @@ Result<void> GainMatrixSmoother::calculate(
...
+    // ACTS_ERROR("Gain smoothing matrix G has NaNs");
+
+    // ACTS_ERROR("Filtered covariance:\n" << filteredCovariance(ts));
+    // ACTS_ERROR("Jacobian:\n" << jacobian(prev_ts));
+    // ACTS_ERROR("Predicted covariance:\n" << predictedCovariance(prev_ts));
+    // ACTS_ERROR("Inverse of predicted covariance:\n"
+    //            << predictedCovariance(prev_ts).inverse());
+
+    // ACTS_ERROR("Gain smoothing matrix G:\n" << G);
+
```
timadye, line 47: were these commented changes intended?

Core/src/Utilities/TrackHelpers.cpp
```diff
@@ -35,5 +35,6 @@ class TrackExtrapolationErrorCategory : public std::error_category {
 }  // namespace
 
 std::error_code Acts::make_error_code(Acts::TrackExtrapolationError e) {
-  return {static_cast<int>(e), TrackExtrapolationErrorCategory()};
+  static TrackExtrapolationErrorCategory c;
```
timadye, line 38:
```suggestion
  static const TrackExtrapolationErrorCategory c;
```

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -152,51 +161,141 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
       continue;
     }
 
-    auto& tracksForSeed = result.value();
-    for (auto& track : tracksForSeed) {
-      auto smoothingResult = Acts::smoothTrack(ctx.geoContext, track, logger());
-      if (!smoothingResult.ok()) {
+    auto& firstTracksForSeed = firstResult.value();
+    for (auto& firstTrack : firstTracksForSeed) {
+      // TODO select track
```
timadye, line 185: why might we want to select the track at this stage?

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -152,51 +161,141 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
+            ACTS_WARNING("Second track finding failed for seed "
+                         << iSeed << " with error" << secondResult.error());
+          } else {
+            auto firstFirstState =
+                std::next(firstTrack.trackStatesReversed().begin(),
+                          firstTrack.nTrackStates() - 1);
+
+            auto& secondTracksForSeed = secondResult.value();
+            for (auto& secondTrack : secondTracksForSeed) {
+              // TODO select track
```
timadye, line 237: again, why might we want to select the track at this stage?

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -152,51 +161,141 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
+
+              (*firstFirstState).previous() =
+                  (*std::next(secondTrack.trackStatesReversed().begin()))
+                      .index();
+              secondTrack.tipIndex() = firstTrack.tipIndex();
+
+              Acts::calculateTrackQuantities(secondTrack);
+
+              // TODO this extrapolation should not be necessary as the CKF will
+              // target this surface anyway and could communicate the parameters
```
timadye, line 253: Why can't we remove this extrapolation now? Maybe explain in the `TODO` comment.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -935,20 +927,15 @@ class CombinatorialKalmanFilter {
     /// @param prevTip The index of the previous state
     ///
     /// @return The tip of added state
-    std::size_t addNonSourcelinkState(const TrackStatePropMask& stateMask,
+    std::size_t addNonSourcelinkState(TrackStatePropMask stateMask,
```
timadye, line 930: another API change that might count as a breaking change, but again probably won't actually break any real code.

Core/include/Acts/EventData/MultiTrajectory.hpp
```diff
@@ -240,9 +240,8 @@ class MultiTrajectory {
   /// @param iprevious index of the previous state, kInvalid if first
   /// @return Index of the newly added track state
   template <bool RO = ReadOnly, typename = std::enable_if_t<!RO>>
-  constexpr IndexType addTrackState(
-      TrackStatePropMask mask = TrackStatePropMask::All,
-      IndexType iprevious = kInvalid) {
+  IndexType addTrackState(TrackStatePropMask mask = TrackStatePropMask::All,
+                          IndexType iprevious = kInvalid) {
```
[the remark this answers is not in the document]
andiwand: I wonder if this ever really was constexpr because the underlying vector impl will not be constexpr for example?

This is one of the changes I would factor out

Core/include/Acts/Utilities/TrackHelpers.hpp
andiwand: I think this one should be fine as something non const will still pass and something const too

Core/src/TrackFitting/GainMatrixSmoother.cpp
andiwand: Yes I think this could be useful for debugging in the future

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
andiwand: This could be an opportunity to throw tracks out already. Either before smoothing or after or both. Cuts on the impact parameters are not possible but we could cut on momentum at innermost hit or number of measurements or holes.

But this makes only sense if we can cut on more or "better" information compared to the branch aborter

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
andiwand: discussion above

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
andiwand: That looked like a limitation of the CKF. Maybe this is gone now - I can try again. I think the problem was / is that an already complete track will not generate any new states in the second search and will not have the extrapolated parameters attached to a second track.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
andiwand: I think this should be unproblematic. The mask is actually an 8 bit value so it does not make sense to reference to it.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -152,51 +161,141 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
+              if (secondTrack.nTrackStates() < 2) {
+                continue;
+              }
+
+              secondTrack.reverseTrackStates(true);
+              seedNumber(secondTrack) = nSeed - 1;
+
+              (*firstFirstState).previous() =
+                  (*std::next(secondTrack.trackStatesReversed().begin()))
+                      .index();
```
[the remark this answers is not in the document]
andiwand: I think this is fine as the first state of the first track will either be the same and we just overwrite `previous` all the time or it will be a different state and we just set a new `previous`.

The second tracks does not change so it should be fine.

andiwand:
Started breaking things off
- https://github.com/acts-project/acts/pull/3073
- https://github.com/acts-project/acts/pull/3074
- https://github.com/acts-project/acts/pull/3075

Core/include/Acts/Utilities/TrackHelpers.hpp
timadye: yeah, I think it only matters if the exact signature needs to match. I assume not here.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
timadye: yeah. I think this is best handled in the branch aborter which can stop once it has enough information to make the cut.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
timadye: it might be best to have a bit more information in the comments - or just remove them. But it's fine for me if you prefer to leave it.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
timadye: either way, can you update or remove the comment?

Core/include/Acts/EventData/MultiTrajectory.hpp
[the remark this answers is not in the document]
andiwand: I moved this to https://github.com/acts-project/acts/pull/3073

We can clean this PR out as soon as the others are merged. Or you can do it already if you like!

andiwand:
I think this now works as it should. The only thing which is a bit annoying is the second extrapolation which would not be necessary if the CKF would spit out parameters at the target surface.

But it might be good enough for now and we improve the Core CKF in the next iteration.

What do you think @timadye ?

timadye:
The changes look OK to me.

Unfortunately the results aren't so good running on ITk with ttbar PU200.
1. It crashes after 81 events with error:
```
/home/ppd/adye/acts/build/src/Core/include/Acts/Utilities/VectorHelpers.hpp:142: std::array<double, 4> Acts::VectorHelpers::evaluateTrigonomics(const Acts::Vector3&): Assertion `sinTheta != 0 && "VectorHelpers: Vector is parallel to the z-axis " "which leads to division by zero"' failed.
```
2. It has a **lot** more `ERROR`s. Here's a count of the most common:
```
 284888 TrackFinding   ERROR     Smoothing track 1 failed with error KalmanFitterError:3
 284888 TrackFinding   ERROR     First smoothing for seed 22 and track 1 failed with error KalmanFitterError:3
 140630 TrackFinding   ERROR     no intersection found
 140630 TrackFinding   ERROR     failed to find track state for extrapolation
  74918 TrackFinding   ERROR     Extrapolation for seed 14 and track 0 failed with error TrackExtrapolationError:2
  65712 TrackFinding   ERROR     Second extrapolation for seed 14 and track 6 failed with error TrackExtrapolationError:2
   8238 TrackFinding   ERROR     Propagation reached the step count limit of 1000 (did 1000 steps)
```
3. If I set `twoWay=false`, it seems to hang on event 97. All other events finished, but an hour later one thread is still running with 100% CPU, and all the other threads are idle. The 8-thread process is using 64.6GB RAM.
4. With `twoWay=false`, I still get quite a lot of `ERROR`s, but this is only about twice as many as we had before #2722, albeit different messages:
```
  94977 TrackFinding   ERROR     no intersection found
  94977 TrackFinding   ERROR     failed to find track state for extrapolation
  94977 TrackFinding   ERROR     Extrapolation for seed 2 and track 0 failed with error TrackExtrapolationError:2
```

Since it misbehaves even with `twoWay=false`, I think there is still a problem with other core changes, notably #2722. So maybe not specific to this PR, but needs fixing before we can see the performance implications, at least for the ITk.

andiwand:
Thank you for the detailed investigation @timadye !

I will look into this next week.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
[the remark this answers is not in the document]
andiwand: You were right here @timadye this messed with `firstTrack` which caused errors down the line. I patched this now by restoring the `previous` index.

timadye:
Just checked the latest version with the ITk, including `restore first track` patch. The number of `ERROR` messages are now similar between runs with `twoWay=true` and `false`. I still get a crash and hang as reported above.

I think that means we still have a problem running with the ITk, but it doesn't seem to be specifically associated with this PR.

AJPfleger (dismissed):

andiwand:
This seems to produce new FPEs - I will take a look

andiwand:
FPE should be hopefully fixed now

timadye (commented):
Sorry, these comments may just be paranoia on my part, but I just wanted to check...

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -152,51 +170,149 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
+                         << iSeed << " with error" << secondResult.error());
+          } else {
+            auto firstFirstState =
+                std::next(firstTrack.trackStatesReversed().begin(),
+                          firstTrack.nTrackStates() - 1);
+
+            auto& secondTracksForSeed = secondResult.value();
+            for (auto& secondTrack : secondTracksForSeed) {
+              if (secondTrack.nTrackStates() < 2) {
+                continue;
```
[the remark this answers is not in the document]
andiwand: My thinking was that this will restore to the behavior without two way

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -152,51 +170,149 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
+              }
+
+              if (!m_trackSelector.has_value() ||
+                  m_trackSelector->isValidTrack(secondTrack)) {
+                auto destProxy = tracks.makeTrack();
+                destProxy.copyFrom(secondTrack, true);
+              }
+
+              // restore first track
+              (*firstFirstState).previous() = Acts::kTrackIndexInvalid;
```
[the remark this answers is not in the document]
andiwand: I think this is protecting us from the forward branching which is still happening

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -152,51 +170,149 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
+              firstState->referenceSurface().getSharedPtr(),
+              firstState->parameters(), firstState->covariance(),
+              firstInitialParameters.particleHypothesis());
+
+          auto secondResult = (*m_cfg.findTracks)(secondInitialParameters,
+                                                  secondOptions, tracksTemp);
+
+          if (!secondResult.ok()) {
+            ACTS_WARNING("Second track finding failed for seed "
+                         << iSeed << " with error" << secondResult.error());
```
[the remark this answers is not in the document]
andiwand: see discussion https://github.com/acts-project/acts/pull/3066#discussion_r1558112525

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -152,51 +170,149 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
+              if (!secondExtrapolationResult.ok()) {
+                // restore first track
+                (*firstFirstState).previous() = Acts::kTrackIndexInvalid;
+
+                m_nFailedExtrapolation++;
+                ACTS_ERROR("Second extrapolation for seed "
+                           << iSeed << " and track " << secondTrack.index()
+                           << " failed with error "
+                           << secondExtrapolationResult.error());
+                continue;
```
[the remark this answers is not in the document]
andiwand: see discussion https://github.com/acts-project/acts/pull/3066#discussion_r1558112525

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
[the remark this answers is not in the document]
timadye: yes, that was my understanding. I was worried that `findTracks(secondInitialParameters,...)` could change the first track's parameters/covariance. That would be bad if it turns out we don't use the second track.

I guess you are saying that it isn't `findTracks` that can affect the whole track, but only the subsequent smoothing which is done once we are committed to the second track. Is that right?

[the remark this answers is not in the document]
andiwand: The second pass should not affect the `firstTrack`. So the track states should really stay the same. The final params are then determined using the extrapolation which will use one of the unchanged track states. So yes, the `findTracks` will not affect the whole track.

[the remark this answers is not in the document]
timadye: Thanks for confirming. I guess this could still be an issue with the original all-in-one CKF (prior to #2722) that we still have to use in Athena. We don't have to worry about multiple branches (because we only use one branch at the moment), but maybe do for cases where the second track fails and we revert to the first - most common where we have < 2 track states. Is it worth trying to mitigate this? I thought to at least add the restoration of the original initial index as you do in this PR.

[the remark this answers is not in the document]
andiwand: We can take care of that when we bump the Acts version. I plan to release tomorrow and open a MR in Athena with the bumped version. Then we can check the two way implementation in Athena

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -113,14 +114,32 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
+  ActsExamples::TrackFindingAlgorithm::TrackFinderOptions firstOptions(
       ctx.geoContext, ctx.magFieldContext, ctx.calibContext, slAccessorDelegate,
-      extensions, pOptions);
+      extensions, firstPropOptions);
+
+  // For the second pass we do not want to branch which is the default behavior
+  // of the MeasurementSelector
+  Acts::MeasurementSelector secondMeasSel{Acts::MeasurementSelector::Config{
+      {Acts::GeometryIdentifier(),
+       Acts::MeasurementSelectorCuts{{}, {5}, {1}}}}};
```
timadye:
```suggestion
  Acts::MeasurementSelector::Config secondMeasSelCfg{m_cfg.measurementSelectorCfg};
  for (auto cuts : m_cfg.measurementSelectorCfg) {
    cuts.numMeasurementsCutOff = {1};
  }
  Acts::MeasurementSelector secondMeasSel{std::move(secondMeasSelCfg)};
```
or
```suggestion
  std::vector<Acts::MeasurementSelector::Config::InputElement> secondMeasSelCfg;
  secondMeasSelCfg.resize(m_cfg.measurementSelectorCfg.size());
  for (std::size_t index = 0; index < m_cfg.measurementSelectorCfg.size(); ++index) {
    secondMeasSelCfg.emplace_back(
        m_cfg.measurementSelectorCfg.idAt(index),
        Acts::MeasurementSelectorCuts{
            m_cfg.measurementSelectorCfg.valueAt(index).etaBins,
            m_cfg.measurementSelectorCfg.valueAt(index).chi2CutOff,
            {1}});
  }
  Acts::MeasurementSelector secondMeasSel{std::move(secondMeasSelCfg)};
```
to save a `vector` allocation by copying each member. (Note, it is OK to specify a single value, even if we have eta bins.)

Or did you mean to have a different `chi2CutOff` as well? The default is `15` (in `reconstruction.py`), but the user could change it.

timadye: Wow! It looks like you removed the need for this suggestion at just the same time I posted it 😁

andiwand: Oh yeah sorry ... I just realized that this is still in there. That was a workaround to only process a single branch in the second pass. But with the recent changes it should be possible again to branch there too.

andiwand:
Made a quick performance comparison between https://github.com/acts-project/acts/pull/3066/commits/de9a0c915b3a7e994ac1d948593af05b23324e89 and current

image

Looks like the additional track copies cost about 3%

@timadye

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -152,51 +162,162 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
+            Acts::calculateTrackQuantities(trackCandidate);
+          }
+        }
+      }
 
-      if (!m_trackSelector.has_value() ||
-          m_trackSelector->isValidTrack(track)) {
-        auto destProxy = tracks.makeTrack();
-        destProxy.copyFrom(track, true);  // make sure we copy track states!
+      if (nSecond == 0) {
```
AJPfleger, line 302: Could you add a comment, when we will do the one-way-extrapolation? eg reasons why two-way might fail

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -152,51 +162,162 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
+                continue;
+              }
+
+              if (!m_trackSelector.has_value() ||
+                  m_trackSelector->isValidTrack(trackCandidate)) {
+                auto destProxy = tracks.makeTrack();
+                destProxy.copyFrom(trackCandidate, true);
+              }
+
+              ++nSecond;
```
AJPfleger, line 291: Do we want to count them even, if the candidate is not a valid track?

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -152,51 +162,162 @@ ActsExamples::ProcessCode ActsExamples::TrackFindingAlgorithm::execute(
...
-          track, *pSurface, extrapolator, extrapolationOptions,
-          m_cfg.extrapolationStrategy, logger());
-      if (!extrapolationResult.ok()) {
-        m_nFailedExtrapolation++;
-        ACTS_ERROR("Extrapolation for seed "
-                   << iseed << " and track " << track.index()
-                   << " failed with error " << extrapolationResult.error());
-        continue;
-      }
+      std::size_t nSecond = 0;
```
AJPfleger, line 203: Could you add a comment, what this variable is for?

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
andiwand: nope this just indicates how many second tracks we actually found

AJPfleger (approved):

timadye:
I ran the ITk ttbar+PU200 tests again (this is standalone ACTS). This still crashed (`twoWay=true`) or went into a loop (`twoWay=false`) within 100 events. As suggested before, this is probably a result of #2722 and only gives different symptoms with this PR.

andiwand:
I will take a look @timadye

PR #3069, opened 2024-04-01
refactor: Do not allocate memory if the surface is connected with a detector el…

junggjo9:
If the surface has an associated detector element its internal transform is never accessed consuming memory for no-reason. If the transform is a unique_ptr which is released in such cases, the footprint should be reduced


paulgessinger:
This is what we used to have before. I think @asalzburger at some point measured that having the transform locally improved performance. But I forget the details.

junggjo9:
Hmm, the transform is never used if the detector element is present.... Confused, I am

paulgessinger:
The transform is only used for surfaces which don't have a detector element. Otherwise the transform is retrieved from the detector element, and that's where for example the alignment is handled.

junggjo9:
Yep, that what I'm saying too ;)

paulgessinger:
So for all other surfaces, boundaries etc, the transform is accessed locally.

asalzburger:
> This is what we used to have before. I think @asalzburger at some point measured that having the transform locally improved performance. But I forget the details.

Hi, yes I remember that there was a visible (but small performance change), however, we can just see with this PR if this is still a measurable effect?

junggjo9:
Results from the propagation test on 3 threads

patched:
```
10:03:28    Sequencer      INFO      Processed 100000 events in 1223.381458 s (wall clock)
10:03:28    Sequencer      INFO      Average time per event: 48.693846 ms/event
```
main:

```
22:24:31    Sequencer      INFO      Processed 100000 events in 1237.661429 s (wall clock)
22:24:31    Sequencer      INFO      Average time per event: 49.130005 ms/event
```

Core/src/Surfaces/PlaneSurface.cpp
```diff
@@ -38,7 +38,24 @@ Acts::PlaneSurface::PlaneSurface(const GeometryContext& gctx,
...
+  Vector3 V = T.cross(U);
+  RotationMatrix3 curvilinearRotation;
+  curvilinearRotation.col(0) = U;
+  curvilinearRotation.col(1) = V;
+  curvilinearRotation.col(2) = T;
+
+  // curvilinear surfaces are boundless
+  Transform3 trf{curvilinearRotation};
+  trf.pretranslate(center);
+  m_transform = std::make_unique<Transform3>(std::move(trf));
```
andiwand, line 58: is that change intentional?

junggjo9: Nope. That was by mistake. Will fix it

AJPfleger: Hey, what is the status on this? :)

AJPfleger:
@junggjo9 @paulgessinger @asalzburger @andiwand
In the tests, it makes a difference of around 1% (not sure if this is significant). How should we proceed with this?

Core/src/Surfaces/Surface.cpp
```diff
@@ -22,21 +22,26 @@ std::array<std::string, Acts::Surface::SurfaceType::Other>
...
 
 Acts::Surface::Surface(const DetectorElementBase& detelement)
     : GeometryObject(), m_associatedDetElement(&detelement) {}
 
 Acts::Surface::Surface(const Surface& other)
     : GeometryObject(other),
       std::enable_shared_from_this<Surface>(),
-      m_transform(other.m_transform),
-      m_surfaceMaterial(other.m_surfaceMaterial) {}
+
```
andiwand, line 33:
```suggestion
```

Core/src/Geometry/CylinderLayer.cpp
```diff
@@ -37,7 +37,8 @@ Acts::CylinderLayer::CylinderLayer(
   auto cVolumeBounds = std::make_shared<CylinderVolumeBounds>(
       *CylinderSurface::m_bounds, thickness);
   // @todo rotate around x for the avePhi if you have a sector
-  m_representingVolume = std::make_unique<Volume>(m_transform, cVolumeBounds);
+  Transform3 volTrf = m_transform ? *m_transform : Transform3::Identity();
+  m_representingVolume = std::make_unique<Volume>(volTrf, cVolumeBounds);
```
andiwand, line 41: I think it is safe to assume that `m_transform` will be set
```suggestion
  m_representingVolume = std::make_unique<Volume>(*m_transform, cVolumeBounds);
```

Core/src/Geometry/DiscLayer.cpp
```diff
@@ -39,7 +39,9 @@ Acts::DiscLayer::DiscLayer(const Transform3& transform,
     auto rVolumeBounds =
         std::make_shared<CylinderVolumeBounds>(*rBounds, thickness);
     // @todo rotate around x for the avePhi if you have a sector
-    m_representingVolume = std::make_unique<Volume>(m_transform, rVolumeBounds);
+    Transform3 volTrf = m_transform ? *m_transform : Transform3::Identity();
+    m_representingVolume =
+        std::make_unique<Volume>(std::move(volTrf), rVolumeBounds);
```
andiwand, line 44: same here

```suggestion
    m_representingVolume = std::make_unique<Volume>(*m_transform, rVolumeBounds);
```

Core/src/Surfaces/Surface.cpp
```diff
@@ -22,21 +22,26 @@ std::array<std::string, Acts::Surface::SurfaceType::Other>
...
+      m_associatedDetElement(other.m_associatedDetElement),
+      m_surfaceMaterial(other.m_surfaceMaterial) {
+  if (other.m_transform) {
+    m_transform = std::make_unique<Transform3>(*other.m_transform);
+  }
```
andiwand, line 38: this sort of handling makes me wonder if a `std::shared_ptr` would be the better option

Core/src/Surfaces/Surface.cpp
```diff
@@ -240,7 +250,7 @@ const Acts::Transform3& Acts::Surface::transform(
   if (m_associatedDetElement != nullptr) {
     return m_associatedDetElement->transform(gctx);
   }
-  return m_transform;
+  return (*m_transform);
```
andiwand, line 253:
```suggestion
  return *m_transform;
```

andiwand:
no significant change in performance - good from my side

image

image

Core/src/Surfaces/Surface.cpp
paulgessinger: I don't think we copy surfaces very often.

andiwand: Right I also don't think it would result in any significant memory reduction but it might make the code a bit easier to read

andiwand:
There are a couple of CI failure
- Can you check the formatting @junggjo9 ?
- I guess the `CI Bridge` ones are false positives
- codecov complains - I guess because there are a couple more branches which are not directly tested right now? Should we add some unit tests to capture the new behavior?

Core/src/Surfaces/Surface.cpp
```diff
@@ -156,7 +161,11 @@ Acts::Surface& Acts::Surface::operator=(const Surface& other) {
   if (&other != this) {
     GeometryObject::operator=(other);
     // detector element, identifier & layer association are unique
-    m_transform = other.m_transform;
+    if (other.m_transform) {
+      m_transform = std::make_unique<Transform3>(*other.m_transform);
+    } else {
+      m_transform.reset();
```
AJPfleger, line 167: Coverage is now resolved. Just this single untested branch is added. Maybe you can think of a test for it. Otherwise only the formatting fails.

Core/src/Surfaces/Surface.cpp
junggjo9: so shared or unique? That's the question ;)

Core/src/Surfaces/Surface.cpp
junggjo9: If the transform is not there then it's connected with the detector element. Can't think about a test right now which would be sensible

Core/src/Surfaces/Surface.cpp
andiwand: Let's leave it as it is. I don't think this is a critical thing and it is just a personal preference on my side

andiwand (approved):
lgtm 👍

PR #3073, opened 2024-04-04
refactor: Remove `constexpr` from `MultiTrajectory` functions

andiwand:
I believe the `constexpr` cannot be guaranteed with C++17 for the vector backend and even with C++20 it would limit other backends to be `constexpr`.

From https://en.cppreference.com/w/cpp/language/constexpr
> The constexpr specifier declares that it is possible to evaluate the value of the function or variable at compile time. Such variables and functions can then be used where only compile time [constant expressions](https://en.cppreference.com/w/cpp/language/constant_expression) are allowed (provided that appropriate function arguments are given).

If not all backends can be guaranteed to be `constexpr`, the frontend cannot be `constexpr`.


andiwand:
What do you think about this one @paulgessinger ?

paulgessinger:
I was originally thinking of the backend declares its methods as constexpr this might allow exposing this. But the actual operations further in are not const evaluated I don't think. 

Not sure, we can drop them here, but they also don't break anything.

andiwand:
But declaring it `constexpr` in the frontend would force the backend also being `constexpr` no? Otherwise the `constexpr` on the frontend is invalid I thought.

paulgessinger:
Since constexpr in functions is mostly a compiler hint, it doesn't make a difference. The compiler would only complain for a constexpr variable where the function might not in fact be const executable. 

In any case, I don't think stripping these qualifiers hurts, so let's do it.

AJPfleger (approved):

acts-project-service:
## :red_circle: Athena integration test results [1374baab7433e367bf41a9f2c69af6624f8a2d19]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/37771196) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7178219)!

PR #3099, opened 2024-04-12
build: setup covfie, detray, traccc, and shared libraries

fredevb:
Updated cmake files to fetch covfie, detray, traccc, thrust, algebra-plugins and set the cmake options in the projects to use shared libraries instead of setting them up for themselves.
Update dfelibs version to v20211029 and vecmem version to v1.4.0.
Added new cmake options:
- added ACTS_SETUP_<name> and ACTS_USE_SYSTEM_<name> for covfie, detray, tracc, thrust, algebra-plugins, and dfelibs.
- added ACTS_BUILD_PLUGIN_COVFIE and ACTS_BUILD_PLUGIN_TRACCC for building plugins that would allow acts to easily make use of covfie and traccc (future work)
Added tests for covfie and traccc to check if the libraries and plugins are available.


asalzburger:
Hi @krasznaa - can we kindly ask you to review this one?

Plugins/Covfie/include/Acts/Plugins/Covfie/CovfieConverter.hpp
```diff
@@ -0,0 +1,14 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+namespace acts::covfie_conversion{
+
+    bool test();
+}
```
asalzburger, line 14: You will get notified about End of line / newline - so you can start fixing the files below,

CMakeLists.txt
```diff
@@ -197,6 +212,12 @@ set(_acts_nlohmanjson_version 3.2.0)
 set(_acts_onnxruntime_version 1.12.0)
 set(_acts_root_version 6.20)
 set(_acts_tbb_version 2020.1)
+#set(_acts_vecmem_version 1.4.0)
+#set(_acts_algebraplugins_version 0.22.0)
+#set(_acts_covfie_version 0.9.0)
+#set(_acts_detray_version 0.55.0)
+#set(_acts_traccc_version 0.8.0)
+#set(_acts_thrust_version 0.8.0)
```
asalzburger, line 220: Did you hard-code these in the `<package>/CMakeFiles.txt`? I like that they are centrally set.

krasznaa: I know I'm stirring a deeper pot right here, but I'm not the biggest fan of this setup. :frowning: I understand the appeal of seeing all version numbers side-by-side, but:
  - These "hidden variables" are passed around, and turned into actual cache variables in some really awkward ways at the moment. See for instance:

https://github.com/acts-project/acts/blob/main/thirdparty/actsvg/CMakeLists.txt

(That thing turns `_acts_actsvg_version` into `ACTS_VERSION`, and then the actual cache variable, `ACTS_ACTSVG_GIT_TAG`, which gets used by CMake.)

  - I switched both Athena and all the R&D projects to always download TGZ files instead of doing Git clones. With the ability for the user to switch to git clones if they absolutely want to. This is done by using a single `<FOO>_SOURCE` cache variable, like in:
    * https://github.com/acts-project/traccc/blob/main/extern/vecmem/CMakeLists.txt#L20-L22
    * https://gitlab.cern.ch/atlas/atlasexternals/-/blob/main/External/VecMem/CMakeLists.txt?ref_type=heads#L15-17
    * etc.

The latter, picking up TGZ files, is really hard to sync with specifying version number in a single place. :thinking: You'd have to specify the whole "source strings" in a single place (including the MD5 hashes when needed), or the whole thing just doesn't work.

So all in all, not the biggest fan of the current Acts setup. Reminds me way too much of what [lcgcmake](https://gitlab.cern.ch/sft/lcgcmake) does. (Which is not a good thing. :frowning:)

CMakeLists.txt
```diff
@@ -205,6 +226,7 @@ set(_acts_boost_recommended_version 1.78.0)
 # required packages
 if (ACTS_SETUP_BOOST)
   if (ACTS_USE_SYSTEM_BOOST)
+    message(STATUS "Using system installation of Boost")
```
krasznaa, line 229: This is just temporary, right?

CMake out of the box can indeed be short on info on what it is looking for, but there are better ways of printing more info in case the `find_package(Boost)` call doesn't provide enough of it. :thinking: 

(The same comment goes for all similar printouts...)

Plugins/Traccc/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+add_library(
+  ActsPluginTraccc SHARED
```
krasznaa, line 2: In the end this PR should not contain any of these plugins. :thinking:

But if, for the moment, you want to test something meaningful, at least include some very basic Covfie / traccc calls in these dummy functions. To test that the compilation / linking works on at least the most basic level.

thirdparty/thrust/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
+# This file is part of the Acts project.
```
krasznaa, line 1: Do we expect that we'll want to use Thrust directly in Acts? :thinking: If not, we could leave it up to traccc to build whichever version it wants. (Still telling Detray that it wouldn't build it, so that only traccc does.)

thirdparty/vecmem/CMakeLists.txt
```diff
@@ -16,11 +16,15 @@ message( STATUS "Building VecMem as part of the Acts project" )
...
-set( ACTS_VECMEM_GIT_TAG "v0.4.0" CACHE STRING "Version of VecMem to build" )
+set( ACTS_VECMEM_GIT_TAG "v1.4.0" CACHE STRING "Version of VecMem to build" )
 mark_as_advanced( ACTS_VECMEM_GIT_REPOSITORY ACTS_VECMEM_GIT_TAG )
 FetchContent_Declare( VecMem
    GIT_REPOSITORY "${ACTS_VECMEM_GIT_REPOSITORY}"
    GIT_TAG "${ACTS_VECMEM_GIT_TAG}" )
 
+# Options used in the build of VecMem.
+set( VECMEM_BUILD_TESTING FALSE CACHE BOOL
+   "Turn off the build of the VecMem unit tests" )
```
krasznaa, line 27: It is the default (currently), but I wonder if it may make sense to set `VECMEM_BUILD_BENCHMARKING` to `FALSE` explicitly here as well. :thinking:

krasznaa (commented):
As I wrote, I would very much prefer to control the "source" of these externals through single `<FOO>_SOURCE` cache variables. As that has served us a little better in other places.

But in general, I think this looks good. After some cleanup, I think this will be the way to go with it. :wink:

thirdparty/OpenDataDetector
andiwand, line 1: I guess this one is unintentional?

This is likely the origin of the physmon regression

AJPfleger:
Discussion from today's weekly core-meeting (Revising a few points already mentioned):
@krasznaa:
Current cmake has many options for getting external configurations.
In many projects we choose one single way to download things.
In offline code: download from git with tags is not the best choice, because cmake can get confused.
Some time ago, we switched to tar-files, because you can also add hashes.
git would be not so great for extra hashes.

Suggested solution: make long string, that consists types and values like "download from this git with that specific tag"
We already have for ATLAS offline build:
- download special ACTS git tag
- download ACTS nightly
All R&D projects are already set up this way, and we could also keep all the strings from above in a single place.

@fredevb @asalzburger agree on this solution and plan to go in that direction.

paulgessinger:
I'm wondering: would we want all of these libs to live in separate plugins, or might it be worth considering bundling them together? I can imagine we'll have a lot of code that either uses none of the libs, or all of them. In this case, we can possibly avoid a lot of conditional checks.

krasznaa:
:thinking: Generally this may be a good idea. Since especially with the current code, we have to make sure that the build is configured consistently between detray and traccc. (For the floating point precision decision for instance.)

Still, since in the long run traccc will definitely stop being an external (even if vecmem and detray for instance remain as such even in the longer term), it may be better to just handle everything individually from the start. :thinking:

paulgessinger:
@krasznaa but will there be a use case where we want to build, say the detray plugin but not the traccc plugin?

Even if traccc as a separate library goes away, it would be simpler imo to have a single traccc plugin that builds everything, and eventually this plugin switch just configures compilation of traccc sources instead of building the external. 

I'm just a bit worried we'll end up having a lot of conditional logic that we don't really need.

krasznaa:
First off: I agree, we must put as little logic into the code as possible. I'm definitely not looking for that, as such things are always super fragile and hard to maintain.

But to ask your first question: We absolutely consider options where we'd want to use detray without traccc. :thinking: We want to do that relatively soon, to produce simulated files for our traccc developments. (Which will still be happening outside of Acts at that point.)

My preference here is to:
  - Allow a great flexibility with what would be built by Acts, and how;
  - Provide a meaningful default for what would be built by Acts.

And that's it. Nothing more. If we want to come up with many different "working modes" with Acts, we should very much consider using [CMake presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) for that purpose. As we already do in the R&D projects. :thinking: (While still avoiding any complicated inter-dependent CMake flags.)

paulgessinger:
Sounds good. Anyway I'm just raising the point here, not arguing we shouldn't go ahead with this or anything. 

> But to answer your first question: We absolutely consider options where we'd want to use detray without traccc.

Sure, but I'm thinking more about the plugin build, not necessarily workflow. Let's say there's a sim job that uses the numerical integration from detray:

Is it necessary to be able to configure the build with only detray enabled, in a dedicated detray plugin where that sim job lives? 

Or would it be flexible enough if there's a single plugin which also builds traccc, but that workflow uses just detray?

cmake/ActsExternSources.cmake
```diff
@@ -0,0 +1,55 @@
...
+set( ACTS_VECMEM_SOURCE
+   "URL;https://github.com/acts-project/vecmem/archive/refs/tags/v1.4.0.tar.gz;URL_MD5;af5434e34ca9c084678c2c043441f174" CACHE STRING "Source to take VECMEM from")
+mark_as_advanced( ACTS_VECMEM_SOURCE )
+
+set( ACTS_ALGEBRAPLUGINS_SOURCE
+   "URL;https://github.com/acts-project/algebra-plugins/archive/refs/tags/v0.22.0.tar.gz;URL_MD5;42bcaad8d19a2c773993a974816dfdf5" CACHE STRING "Source to take ALGEBRAPLUGINS from")
+mark_as_advanced( ACTS_ALGEBRAPLUGINS_SOURCE )
+
+set( ACTS_COVFIE_SOURCE
+   "URL;https://github.com/acts-project/covfie/archive/refs/tags/v0.9.0.tar.gz;URL_MD5;b310712c6dd1acc8104c734086f40fc0" CACHE STRING "Source to take COVFIE from")
```
stephenswat, line 14:
```suggestion
   "URL;https://github.com/acts-project/covfie/archive/refs/tags/v0.10.0.tar.gz;URL_MD5;af59c6e2a1eebfa765b29f0af9fc70f7" CACHE STRING "Source to take COVFIE from")
```

thirdparty/dfelibs/README.MD
AJPfleger, line 1: I think the file ending shouldn't be in caps.

thirdparty/thrust/CMakeLists.txt
fredevb: I don't think we'll use Thrust directly in Acts. Should I change it so that traccc sets it up for itself and detray?

asalzburger:
Hi @fredevb - I rebased this after #3186 was merged and resolved the conflicts.

CMakeLists.txt
asalzburger: Ok, let's move this to another PR as another cleanup.

Plugins/Traccc/CMakeLists.txt
asalzburger: That's the idea here.

thirdparty/thrust/CMakeLists.txt
asalzburger: I think thrust will currently always be coupled to the traccc + detray, let's leave it with them for the moment.

CMakeLists.txt
```diff
@@ -178,17 +194,13 @@ endif()
...
-set(_acts_boost_version 1.71.0)
 set(_acts_dd4hep_version 1.21)
 set(_acts_edm4hep_version 0.7)
-set(_acts_geomodel_version 4.6.0)
+set(_acts_eigen3_version 3.3.7)
 set(_acts_podio_version 0.16)
 set(_acts_doxygen_version 1.9.4)
-set(_acts_eigen3_version 3.3.7)
 set(_acts_hepmc3_version 3.2.1)
-set(_acts_nlohmanjson_version 3.2.0)
```
andiwand, line 191: isn't this still in use?

same for the others above

fredevb: I don't think its used anymore since it's not set in the external sources cmake file:

set( ACTS_NLOHMANNJSON_SOURCE
   "URL;https://github.com/nlohmann/json/archive/refs/tags/v3.10.5.tar.gz;URL_HASH;SHA1=8969f5ad1a422e01f040ff48dcae9c0e6ad0811d" CACHE STRING "Source to take nlohmann_json from")

CMakeLists.txt
```diff
@@ -287,15 +349,13 @@ if(ACTS_BUILD_PLUGIN_DD4HEP)
 endif()
 if(ACTS_BUILD_PLUGIN_JSON)
   if(ACTS_USE_SYSTEM_NLOHMANN_JSON)
-    message(STATUS "Using system installation of nlohmann::json")
     find_package(nlohmann_json ${_acts_nlohmanjson_version} REQUIRED CONFIG)
```
andiwand, line 352: referring to my comment here https://github.com/acts-project/acts/pull/3099#discussion_r1601010108

CMakeLists.txt
fredevb: I've reverted the changes

CMakeLists.txt
```diff
@@ -184,6 +200,7 @@ set(_acts_boost_version 1.71.0)
 set(_acts_dd4hep_version 1.21)
 set(_acts_edm4hep_version 0.7)
 set(_acts_geomodel_version 4.6.0)
+set(_acts_eigen3_version 3.3.7)
```
andiwand, line 203: this is duplicated on line 206

asalzburger (approved):

asalzburger:
Let's bring this in - so we can also get Eleni's work forward.

thirdparty/boost/CMakeLists.txt
```diff
@@ -16,13 +15,7 @@ set(Boost_B2_FLAGS variant=release threading=multi runtime-link=shared link=shar
...
 # download the boost release
 
-ExternalProject_Add(
-  boost
-  URL ${Boost_URL}
-  CONFIGURE_COMMAND ""
-  BUILD_COMMAND ""
-  INSTALL_COMMAND ""
-)
+ExternalProject_Add( boost "${ACTS_BOOST_SOURCE};CONFIGURE_COMMAND;;BUILD_COMMAND;;INSTALL_COMMAND;" )
```
andiwand, line 18: that command looks a bit obscure - is this a functional change?

asalzburger: Oh, I missed that - @fred can you comment?

fredevb: I changed it so that the source is list (which can contain URL, MD5, etc) and also to make it consistent with how the rest of the external projects are fetched.

krasznaa: But don't be worried to mix the two formalisms. The following should work fine:

```suggestion
ExternalProject_Add(
   boost  
   ${ACTS_BOOST_SOURCE}
   CONFIGURE_COMMAND "" 
   BUILD_COMMAND ""
   INSTALL_COMMAND "" )
```

CMake will understand what you meant by that. This is how I do things in the offline code as well.
  - https://gitlab.cern.ch/atlas/atlasexternals/-/blob/main/External/CLHEP/CMakeLists.txt?ref_type=heads#L61-71
  - https://gitlab.cern.ch/atlas/atlasexternals/-/blob/main/External/Acts/CMakeLists.txt?ref_type=heads#L58-74
  - etc.

andiwand: To me that version is more comprehensible. Maybe something we can follow up on

acts-project-service:
## :red_circle: Athena integration test results [210b8979078704af25154b594071826f3dbb2a59]
Please investigate the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7426052)!

|status|job|report|
|------|---|------|
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168436)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168435)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168434)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168433)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168432)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168431)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168430)||
|:green_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168429)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168428)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168427)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168426)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168425)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168424)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168423)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168422)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168421)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168420)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168419)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168418)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168417)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168416)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168415)||
|:red_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168414)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168413)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168412)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168411)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168410)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168409)||
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39168408)||

cmake/ActsExternSources.cmake
```diff
@@ -0,0 +1,52 @@
...
+   "URL;https://github.com/acts-project/dfelibs/archive/refs/tags/v20211029.tar.gz;URL_MD5;87fb09c5a11b98250f5e266e9cd501ea" CACHE STRING "Source to take dfelibs from")
+mark_as_advanced( ACTS_DFELIBS_SOURCE )
+
+set( ACTS_FRNN_SOURCE
+   "GIT_REPOSITORY;https://github.com/lxxue/FRNN;GIT_TAG;3e370d8d9073d4e130363faf87d2370598b5fbf2" CACHE STRING "Source to take FRNN from")
+mark_as_advanced( ACTS_FRNN_SOURCE )
+
+set( ACTS_GEOMODEL_SOURCE
+   "GIT_REPOSITORY;https://gitlab.cern.ch/GeoModelDev/GeoModel;GIT_TAG;4.6.0;PATCH_COMMAND;git am ${CMAKE_CURRENT_SOURCE_DIR}/0001-Add-option-to-skip-setting-up-json-completely.patch" CACHE STRING "Source to take GeoModel from")
+mark_as_advanced( ACTS_GEOMODEL_SOURCE )
```
andiwand, line 35: I fear this broke the patch apply from @paulgessinger 

cc @fredevb @asalzburger @krasznaa

PR #3101, opened 2024-04-13
perf: Remove covariance from material states and holes in CKF

andiwand:
I briefly looked at some flamegraphs of the CKF and realized that `boundState` is one of the major contributors to the CPU time. Most of the `boundState` calculation is done for covariance transport. The transport is essential for the KF formalism and correctly treating measurements but I don't think it is necessary to transport the covariance to material states.

As a consequence I also remove the jacobian from the material states which will be compensated by having the accumulated transport jacobian on the next measurement state.

I imagine that this can make a big difference especially at the end of the detector where all the sensitives are processed already.

### DIFF
--- a/Core/include/Acts/EventData/TrackProxy.hpp
+++ b/Core/include/Acts/EventData/TrackProxy.hpp
@@ -730,7 +730,7 @@ class TrackProxy {
       prev = ts.previous();
       ts.template component<IndexType>(hashString("next")) = prev;
       ts.previous() = next;
-      if (invertJacobians) {
+      if (invertJacobians && ts.hasJacobian()) {
         if (next != kInvalid) {
           BoundMatrix curJacobian = ts.jacobian();
           ts.jacobian() = nextJacobian.inverse();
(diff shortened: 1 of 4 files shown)


andiwand:
4mu PU 200
default seeding
fatras

image

andiwand:
ttbar PU 200
default seeding
fatras

image

paulgessinger:
We need to be careful not to target the material surfaces on the way back in this case. If we do this anyway, we also might want to not even create track states for the material surfaces to begin with, because we won't need them anymore.

andiwand:
> We need to be careful not to target the material surfaces on the way back in this case. If we do this anyway, we also might want to not even create track states for the material surfaces to begin with, because we won't need them anymore.

I though the reason to have material states on the track was that we can use a direct refitter afterwards. So no need for full navigation to pick up material surfaces again.

But I don't really see a reason to store the full covariance or the jacobian for these states as they are unobservable.

andiwand:
@asalzburger pointed out that these `boundState` calls on material surfaces are actually crucial to correctly account for the MSC and Eloss covariance. I will check if there is something else we can do to reduce the CPU time of the material states or if this is a lost cause.

paulgessinger:
Ah, I was thinking of just not storing the covariance, but the numerical value of the local covariance is needed for the calculation, indeed.

andiwand:
I fear at the moment this is unavoidable during filtering and for the smoothing we need to incorporate the MSC and Eloss uncertainties for correctness too which forces us to store the jacobian and the covariance

Issue #2967, opened 2024-02-19
Improvement: allow CKF to keep stopped branch


timadye:
Following on from #2905, @andiwand suggested:

> Maybe the branch stopper can also return if the track should be kept or not? I could imagine that we could use both scenarios: early removal of uninteresting tracks and stopping the propagation to save CPU and keeping the tracks.

Maybe instead of the CKF's `BranchStopper` returning a `bool`, it could return an:

```
enum BranchStopperResult {
  ContinueBranch = 0,
  DiscardBranch = 1,
  StopAndKeepBranch = 2
};
```

Although the `enum` and `bool` values are upwardly compatible, the type is obviously different, so I guess this would require a breaking change.

If we are already changing the API, it might be useful to simplify the [call signature](https://github.com/acts-project/acts/blob/main/Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp#L89) to pass the track state by `const &`:
```
  using BranchStopper = Delegate<TrackStateFlag(const CombinatorialKalmanFilterTipState&,
                                      const typename traj_t::TrackStateProxy&)>;
```

This proposal is not urgent, so it could be included with the next breaking release.

AJPfleger:
@andiwand @timadye what is the status on this?

andiwand:
this is handled with https://github.com/acts-project/acts/pull/3102

PR #3102, opened 2024-04-15
refactor!: Refactor CKF branch stopper to allow stop and keep tracks

andiwand:
Changes the CKF branch stopper to allow keeping the tracks after the branch is stopped. This is useful if a track already collected enough measurements but then starts to accumulate holes.

This change should allow us to stop branches more aggressively because we don't have to be worried anymore to throw out valid track candidates.

closes
- https://github.com/acts-project/acts/issues/2967

blocked by
- https://github.com/acts-project/acts/pull/3164
- https://github.com/acts-project/acts/pull/3163


AJPfleger (approved):
lgtm, as we already discussed this morning.

I don't see any problems to keep the track, if it has a lot of "holes" in the end. In most (all?) places we consider holes only as such, when they appear etween the first and last measurements.

timadye (dismissed):
Looks good.

It's a pity this change can't be made to be compatible, since an explicit cast from `bool` to `BranchStopperResult` would return the right result. But I can't see any way to make that work, especially for a `Delegate` call.

andiwand:
I am measuring a speedup of 3x which makes me wonder if that can be real 🤔 

image

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -904,13 +921,10 @@ class CombinatorialKalmanFilter {
           tipState.nMeasurements++;
         }
 
-        // Check if need to stop this branch
-        if (!m_extensions.branchStopper(tipState, trackState)) {
```
timadye, line 908: I don't understand why the branch stopper was previously called again here, so I guess this is an improvement. Can you just confirm that the change was intended?

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -73,14 +73,21 @@ struct CombinatorialKalmanFilterExtensions {
   using candidate_container_t =
       typename std::vector<typename traj_t::TrackStateProxy>;
 
+  enum class BranchStopperResult {
```
timadye, line 76: It's a bit awkward that this `enum` that is needed by user code is inside a templated class (`Acts::CombinatorialKalmanFilterExtensions<traj_t>::BranchStopperResult`), even though the template parameter isn't relevant for the definition.

Would it make sense to move it out as `Acts::CombinatorialKalmanFilterBranchStopperResult`?

timadye (commented):
Unfortunately, when I run `full_chain_itk.py`, I get a crash:
```
Core/include/Acts/Utilities/VectorHelpers.hpp:141: std::array<double, 4> Acts::VectorHelpers::evaluateTrigonomics(const Acts::Vector3&): Assertion `sinTheta != 0 && "VectorHelpers: Vector is parallel to the z-axis " "which leads to division by zero"' failed.
```
after ~47 events. With `main HEAD`, it runs OK, at least to >90 events. Of course that difference could just be bad luck.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
andiwand: That is intentional yes.  I think this was used to potentially abort after a measurement was selected but I don't think this should be done. We should only call the branch aborter once after filtering. The measurement selector can already cut with all the information at hand.

andiwand:
> Unfortunately, when I run `full_chain_itk.py`, I get a crash:
> 
> ```
> Core/include/Acts/Utilities/VectorHelpers.hpp:141: std::array<double, 4> Acts::VectorHelpers::evaluateTrigonomics(const Acts::Vector3&): Assertion `sinTheta != 0 && "VectorHelpers: Vector is parallel to the z-axis " "which leads to division by zero"' failed.
> ```
> 
> after ~47 events. With `main HEAD`, it runs OK, at least to >90 events. Of course that difference could just be bad luck.

Hm that looks unrelated. I can try to see if this is easily fixable but this should go into another PR in any case.

andiwand:
@timadye after https://github.com/acts-project/acts/pull/3164 and https://github.com/acts-project/acts/pull/3163 the problem was gone for me

timadye:
> @timadye after #3164 and #3163 the problem was gone for me

Thanks! That should make testing the performance of this PR much easier. I will do that today.

timadye:
> I merged them in now @timadye . Can you try again?

Running... it's already got further than my last attempt 😄

timadye:
It seems to run fine on 100 ttbar+PU200 events, single threaded. I get a modest speedup 7.7&rarr;7.1 s/event (x1.08) for the whole job, but that improvement is all in the `TrackFindingAlgorithm`, 2.5&rarr;1.8 s/event (x1.39). So, nothing like the x3 speedup you saw, but still nice.

Perhaps more interestingly, I also see the hint of an improvement in the efficiency in the transition region:
trackeff_vs_eta

I'm running a larger job with 1000 events and 8 threads to see if these changes hold up there.

timadye:
For 1000 ttbar+PU200 events, with 8 threads, I get an overall speedup 9.1→8.2 s/event (x1.10) for the whole job, and in the `TrackFindingAlgorithm`, 2.7→1.9 s/event (x1.43). The efficiency is a bit weird, but I suppose mostly better than `main`:
trackeff_vs_eta

timadye:
Some more plots with interesting features:
trackeff_vs_pT
trackeff_vs_phi
nOutliers_vs_eta

andiwand:
Just to be sure: these plots come from Athena reconstruction? And all of them are ttbar PU 200?

- I am a little bit surprised that the efficiency is sometimes lower. While in phi it looks like it is almost consistently larger?
- For efficiency over pT I guess our signal is concentrated on the left where we do see some improvements.
- Over eta it looks more symmetrical which is nice.

I believe the increase in efficiency comes from https://github.com/acts-project/acts/pull/3164. I found a spot where we stop the branch but never stored the track.

Speedup 1.4 is great! I am surprised that it is so different between ODD Acts and ITk Athena-Acts. Maybe the previous branch aborter was already more effective than the ODD Acts one because of the detector geometry.

I guess you are using the same holes/outlier cuts as before on Athena main?

timadye:
> Just to be sure: these plots come from Athena reconstruction? And all of them are ttbar PU 200?

No, they are ACTS stand-alone with the ITk. All with ttbar PU 200.

I can't easily test it in Athena until the PRs are merged - and even then, the `main--ACTS` build will fail due to the breaking change - at least to start with.

> * I am a little bit surprised that the efficiency is sometimes lower. While in phi it looks like it is almost consistently larger?

That is &eta; dependent. Overall, the efficiency is larger in more &eta; regions than regions where it is smaller.

> * For efficiency over pT I guess our signal is concentrated on the left where we do see some improvements.

right. Most ttbar PU 200 tracks are low p<sub>T</sub>.

> * Over eta it looks more symmetrical which is nice.

👍 

> I believe the increase in efficiency comes from #3164. I found a spot where we stop the branch but never stored the track.

Cool! I'll test this with #3164 on its own to separate the effect of the two PRs.

> Speedup 1.4 is great! I am surprised that it is so different between ODD Acts and ITk Athena-Acts. Maybe the previous branch aborter was already more effective than the ODD Acts one because of the detector geometry.

Did you check this is still the case? You expressed some doubt about the x3 number before, and the lack of spread in times for different events is a little suspicious.

I assume your plot before was for the `TrackFindingAlgorithm` alone, so comparable with my x1.4 number. Mine was based on `timing.tsv` which only gives an average, not a histogram. How do yet get the timing for each event?

> I guess you are using the same holes/outlier cuts as before on Athena main?

This is acts #3102 vs acts `main`.  I am using the same holes/outlier cuts for both, ie. both with #3163 merged in.

andiwand:
> Did you check this is still the case? You expressed some doubt about the x3 number before, and the lack of spread in times for different events is a little suspicious.

I will run it again. The output seems to have changed significantly with https://github.com/acts-project/acts/pull/3102/commits/2b068eabb436842d9aa364bce6e8ee8155cd156d which is unexpected but the change looks more correct than what was there before.

One suspicion I have is that we end up in the Solenoid quite often in the ODD and just waste a lot of time propagating there. This is prohibited in a lot of cases not by stopping the finding earlier.

timadye:
#3164 didn't show any changes, so the efficiency changes must be down to this PR (#3102). I will also run this one again, following your latest updates.

andiwand:
image

Speedup is still significant - seems to be real.

Here are some stats from the track finding which gives a little bit more insight.

main

```
19:53:34    TrackFinding   INFO      TrackFindingAlgorithm statistics:
19:53:34    TrackFinding   INFO      - total seeds: 1533560
19:53:34    TrackFinding   INFO      - deduplicated seeds: 363980
19:53:34    TrackFinding   INFO      - failed seeds: 0
19:53:34    TrackFinding   INFO      - failed smoothing: 0
19:53:34    TrackFinding   INFO      - failed extrapolation: 0
19:53:34    TrackFinding   INFO      - failure ratio seeds: 0
19:53:34    TrackFinding   INFO      - found tracks: 1028500
19:53:34    TrackFinding   INFO      - selected tracks: 33380
19:53:34    TrackFinding   INFO      - stopped branches: 141080
```

changed

```
19:50:37    TrackFinding   INFO      TrackFindingAlgorithm statistics:
19:50:37    TrackFinding   INFO      - total seeds: 1533560
19:50:37    TrackFinding   INFO      - deduplicated seeds: 288710
19:50:37    TrackFinding   INFO      - failed seeds: 0
19:50:37    TrackFinding   INFO      - failed smoothing: 0
19:50:37    TrackFinding   INFO      - failed extrapolation: 0
19:50:37    TrackFinding   INFO      - failure ratio seeds: 0
19:50:37    TrackFinding   INFO      - found tracks: 50090
19:50:37    TrackFinding   INFO      - selected tracks: 33950
19:50:37    TrackFinding   INFO      - stopped branches: 1196310
```

stopped branches is ~10x

timadye:
Same test as [before](https://github.com/acts-project/acts/pull/3102#issuecomment-2089288414), but comparing #3102 after 2b068ea was added against the old `main`. Similar speedup as before: `TrackFindingAlgorithm` 2.7&rarr;1.9 s/event (x1.42), but even better efficiency improvement:
trackeff_vs_eta
trackeff_vs_phi
nOutliers_vs_eta

timadye:
[again](https://github.com/acts-project/acts/pull/3102#issuecomment-2089288414) for the ITk:
#### `main`:
```
00:14:50    TrackFinding   INFO      TrackFindingAlgorithm statistics:
00:14:50    TrackFinding   INFO      - total seeds: 20906970
00:14:50    TrackFinding   INFO      - deduplicated seeds: 18113959
00:14:50    TrackFinding   INFO      - failed seeds: 0
00:14:50    TrackFinding   INFO      - failed smoothing: 0
00:14:50    TrackFinding   INFO      - failed extrapolation: 0
00:14:50    TrackFinding   INFO      - failure ratio seeds: 0
00:14:50    TrackFinding   INFO      - found tracks: 4771749
00:14:50    TrackFinding   INFO      - selected tracks: 3037718
00:14:50    TrackFinding   INFO      - stopped branches: 636123
```
#### #3102 (2b068ea):
```
18:58:50    TrackFinding   INFO      TrackFindingAlgorithm statistics:
18:58:50    TrackFinding   INFO      - total seeds: 20894532
18:58:50    TrackFinding   INFO      - deduplicated seeds: 17905047
18:58:50    TrackFinding   INFO      - failed seeds: 0
18:58:50    TrackFinding   INFO      - failed smoothing: 0
18:58:50    TrackFinding   INFO      - failed extrapolation: 0
18:58:50    TrackFinding   INFO      - failure ratio seeds: 0
18:58:50    TrackFinding   INFO      - found tracks: 3527696
18:58:50    TrackFinding   INFO      - selected tracks: 3193132
18:58:50    TrackFinding   INFO      - stopped branches: 1946303
```
so x3.1 stopped branches. Comparing with ~x10 for the ODD @andiwand found, is in line with the smaller speedup.

timadye (approved):
This is clearly an improvement. Unless you are planning other breaking changes to be included at the same time, is it ready to go? 🚀

acts-project-service:
## :red_circle: Athena integration test results [bed53b2fd3ddffc36cc7c8176ea4bbb447e6ec2e]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38590937) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7327077)!

PR #3115, opened 2024-04-18
refactor: Make pile-up configurable for `full_chain_odd.py` and change default to 200

andiwand:
Makes the chain more configurable and changes the default pile-up for ttbar to 200.

### DIFF
--- a/Examples/Python/tests/test_examples.py
+++ b/Examples/Python/tests/test_examples.py
@@ -1173,7 +1173,15 @@ def test_full_chain_odd_example_pythia_geant4(tmp_path):
     env["ACTS_LOG_FAILURE_THRESHOLD"] = "ERROR"
     try:
         stdout = subprocess.check_output(
-            [sys.executable, str(script), "-n1", "--geant4", "--ttbar"],
+            [
+                sys.executable,
+                str(script),
+                "-n1",
+                "--geant4",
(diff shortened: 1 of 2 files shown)


Corentin-Allaire (dismissed):
I think the reason we kept it at 50 was that the CI had issue with PU=200. But if it now work that would be great !
I guess you will need to change all the ref later so ping me so that I approve afterward

Examples/Python/tests/test_examples.py
```diff
@@ -1173,7 +1173,15 @@ def test_full_chain_odd_example_pythia_geant4(tmp_path):
...
         stdout = subprocess.check_output(
-            [sys.executable, str(script), "-n1", "--geant4", "--ttbar"],
+            [
+                sys.executable,
+                str(script),
+                "-n1",
+                "--geant4",
+                "--ttbar",
+                "--ttbar-pu",
+                "50",
```
andiwand, line 1183: I made this explicit there @Corentin-Allaire

Corentin-Allaire: Ah yes I missed that ! Then this is perfect !

Corentin-Allaire (approved):

Corentin-Allaire (approved):

PR #3116, opened 2024-04-18
feat: Stop branches based on number of outliers in Examples track finding

andiwand:
The current implementation of the measurement selector will flag any measurement at least as an outlier if there are measurements on the surface. Holes are only produced in case there are no measurements on the surface. This was overlooked while introducing a branch aborter that cuts on the number of holes.

Here I add an additional cut on the number of outliers in the branch aborter to effectively remove fake tracks already during the search.

I wonder if our current implementation of the measurement selector makes sense. Maybe we should not flag outliers during the search and handle them rather as holes. Alternatively one can introduce another chi2 cut to allow cutting on outliers too.


AJPfleger (approved):

andiwand:
this gives some serious speedups

ttbar PU 200
default seeding
fatras

image

timadye:
given this observation, would it make more sense to count outliers and holes an cut on `nHoles+nOutliers <= maxHoles` ?

AJPfleger (approved):

asalzburger:
> this gives some serious speedups
> 
> ttbar PU 200 default seeding fatras
> 
> image

Excellent job, Andreas.

asalzburger:
> given this observation, would it make more sense to count outliers and holes an cut on `nHoles+nOutliers <= maxHoles` ?

Hi @timadye - I think we need to distinguish between holes and outliers in the track fit and in the track finding. In the current implementation, as soon as there is even one measurement on the module (which usually in ttbar 200 has a probability of 1), this will be flagged as an outlier and not a hole .. so holes in track finding do not make too much sense. In track fitting, however, when there is a predefined set of hits, it does!

acts-project-service:
## :white_check_mark: Athena integration test results [edd5661cdb0fb0f7e76848d196c626bf07c356e3]

|status|job|report|
|------|---|------|
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183657)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183656)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183655)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183654)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183653)||
|:green_circle: | [test_ActsConversionWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183652)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183651)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183650)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183649)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183648)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183647)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183646)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183645)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183644)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183643)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183642)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183641)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183640)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183639)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183638)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183637)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183636)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183635)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183634)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183633)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183632)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183631)||
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38183630)||

timadye:
Right. I think we need two `chi2` cuts, one for outliers and one for holes.

I've tried to compare what Athena (`SiCombinatorialTrackFinder_xk`) and Acts (`MeasurementSelector`) do for the ITk.

#### Acts
* `chi2CutOff` = 15
* `maxHoles` = 2

#### Athena
* `Xi2max` = 9 (max chi2 for updators)
* `Xi2maxNoAdd` = 25 (max chi2 for outliers)
* `Xi2maxlink` = 200 (max chi2 for clusters)

the `Xi2`* properties are inconsistently labelled in different places in Athena, so I may have gotten them confused.

* `maxHoles` = 2
* `maxPixelHoles` = 2
* `maxSctHoles` = 2  (&rarr; strip holes)
* `maxDoubleHoles` = 1 (I think this is only used by `InDetNNScoringTool` for SCT, so may not affect the track finding)

The Acts standalone `chi2CutOff` setting was copied over to Athena/Acts, while most other properties are set to their Athena equivalents.

To follow the Athena settings for the ITk, we could add a `chi2CutOffOutlier` and set the two cuts to 9 and 25. Can we also implement something equivalent to  `maxPixelHoles` and `maxSctHoles`? To do that in the branch stopper we would have to somehow keep our own counts, since the `tipState` doesn't keep separate hole counts.

andiwand:
About the chi2 cuts: I think this should be purely user defined. It might be more convenient to implement a measurement selector in Athena. Possibly using inheritance or CRTP.

I also wonder if a chi2 cut is actually what we want or rather a p-value since we are dealing with different DoF.

About the different cuts for pixel and strips: I think this should also be purely experiment driven. One solution could be to also carry a track trough the CKF and not just track states which already has the ability of attaching arbitrary data. I think we should avoid putting state into the branch stopper. If not possible I would rather use inheritance and build an interface for this interaction.

The current 15 chi2 cut might also be a bit too large. It would be interesting to see how much this drives the performance by lowering it to 9 for example.

timadye:
Hi @andiwand,

> About the chi2 cuts: I think this should be purely user defined. It might be more convenient to implement a measurement selector in Athena. Possibly using inheritance or CRTP.

This might be where we have to go eventually, but we still have to solve the problem of the outliers. Adding a second chi2 cut seems like the natural way to solve that, regardless of the detector.

> I also wonder if a chi2 cut is actually what we want or rather a p-value since we are dealing with different DoF.

Perhaps, but maybe a simpler solution would be to introduce different chi2 cuts for pixels and strips. If we wanted to keep it generic, we could name them as chi2 cuts for 1D and 2D.

> About the different cuts for pixel and strips: I think this should also be purely experiment driven. One solution could be to also carry a track trough the CKF and not just track states which already has the ability of attaching arbitrary data. I think we should avoid putting state into the branch stopper. If not possible I would rather use inheritance and build an interface for this interaction.

Yeah, we shouldn't have state in the ActsExamples branch stopper. I would only consider that as stop-gap measure for Athena, until we can support it in core Acts.

Another idea that would work but that some might consider too ugly: could we count measurements, holes, etc. separately for 1D and 2D hits? That would at least be generic, even if not always used.

Another option would be to define an experiment-defined index type to classify surfaces. For ATLAS, that might be pixel, strip, HGTD, etc. This could be used to index a well-defined set of counts, stats, cuts etc in the Core Acts algorithms.

> The current 15 chi2 cut might also be a bit too large. It would be interesting to see how much this drives the performance by lowering it to 9 for example.

Yes, I can test this in Athena. I wouldn't hold out too much hope, because we don't yet have a real calibrator.

Thanks,
Tim.

PR #3124, opened 2024-04-19
feat: Add per event timing

### LINKED ISSUE #1700: Detailed timing per event
Hello,
would it be possible to get the detailed timing per event for each algorithm instead of the total?
@stephenswat @LuisFelipeCoelho

CouthuresJeremy:
Add a parameter to the sequencer to save the timing per event in addition to the default averaged timing.
Closes #1700.
More details in the issue.

### DIFF
--- a/Examples/Framework/src/Framework/Sequencer.cpp
+++ b/Examples/Framework/src/Framework/Sequencer.cpp
@@ -637,7 +637,7 @@ int Sequencer::run() {
         joinPaths(m_cfg.outputDir, m_cfg.outputEventTimingFile);
     dfe::NamedTupleTsvWriter<EventTimingInfo> eventTimingWriter(eventTimingPath,
                                                                 4);
-    for (EventTimingInfo info : eventTimingData) {
+    for (const EventTimingInfo& info : eventTimingData) {
       eventTimingWriter.append(info);
     }
   }


Examples/Framework/include/ActsExamples/Framework/Sequencer.hpp
```diff
@@ -78,6 +78,8 @@ class Sequencer {
     std::string outputDir;
     /// output name of the timing file
     std::string outputTimingFile = "timing.tsv";
+    // Enable or disable detailed per event timing writing
+    bool enableEventTiming = false;
```
andiwand, line 82: I think this should be steered with an output file path which can be present or not. Similar to `outputTimingFile`

CouthuresJeremy: I updated it, is it what you had in mind?

Examples/Framework/include/ActsExamples/Framework/Sequencer.hpp
```diff
@@ -78,6 +78,9 @@ class Sequencer {
     std::string outputDir;
     /// output name of the timing file
     std::string outputTimingFile = "timing.tsv";
+    /// Output name of the per event timing file
+    /// Enable detailed per event timing writing
+    std::optional<std::string> outputEventTimingFile = std::nullopt;
```
andiwand, line 83:
```suggestion
    std::string outputEventTimingFile = "timing-events.tsv;
```

I think a string is sufficient. We usually check if it is empty to see if the output is wanted or not.

I would also argue that the new output can be useful by default and does not harm anyone.

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -394,6 +394,14 @@ struct TimingInfo {
   DFE_NAMEDTUPLE(TimingInfo, identifier, time_total_s, time_perevent_s);
 };
 
+// struct to store timing data
+struct EventTimingInfo {
+  std::string identifier;
+  double time_s = 0;
+  int eventId;
```
andiwand, line 401:
```suggestion
  double time_s = 0;
  int eventId = 0;
```

I would either initialize both or none of both

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -542,6 +552,19 @@ int Sequencer::run() {
               context.fpeMonitor = nullptr;
             }
 
+            if (enableEventTiming) {
```
andiwand, line 555:
```suggestion
            if (!m_cfg.outputEventTimingFile.empty()) {
```

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -598,6 +621,17 @@ int Sequencer::run() {
     return EXIT_FAILURE;
   }
 
+  if (enableEventTiming) {
```
andiwand, line 624:
```suggestion
  if (!m_cfg.outputEventTimingFile.empty()) {
```

Examples/Framework/include/ActsExamples/Framework/Sequencer.hpp
CouthuresJeremy: In the issue, @paulgessinger thought that it would be better off by default as it might produce very large files (on row for each algorithm for each event).
Ref: [issue  #1700](https://github.com/acts-project/acts/issues/1700#issuecomment-1330332667)

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -542,6 +552,19 @@ int Sequencer::run() {
...
+            if (enableEventTiming) {
+              // Collect the event timing data
+              for (std::size_t i = 0; i < names.size(); i++) {
+                EventTimingInfo eventInfo;
+                eventInfo.identifier = names[i];
+                eventInfo.time_s = std::chrono::duration_cast<Seconds>(
+                                       localClocksAlgorithms[i])
+                                       .count();
+                eventInfo.eventId = event;
+                eventTimingData.push_back(eventInfo);
```
paulgessinger, line 564: I don't think this is thread-safe.

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -416,6 +424,7 @@ int Sequencer::run() {
   std::vector<std::string> names = listAlgorithmNames();
   std::vector<Duration> clocksAlgorithms(names.size(), Duration::zero());
   tbbWrap::queuing_mutex clocksAlgorithmsMutex;
+  std::vector<EventTimingInfo> eventTimingData;
```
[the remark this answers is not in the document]
andiwand: Writing to disk after each event sounds easiest to me. We just need to synchronize on the output file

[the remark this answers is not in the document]
CouthuresJeremy: I agree, mimicking the way the writer algorithms are handled would probably make more sense but I do not get how it is done.

Examples/Framework/src/Framework/Sequencer.cpp
CouthuresJeremy: From [stackoverflow](https://stackoverflow.com/a/9042750), indeed, it is not.
I am not familiar enough with threading to spot this, thanks!

paulgessinger: You'll probably need to use a `mutex` to synchronize the vector accesses, but that also might impact the actual performance that you're trying to measure.

CouthuresJeremy: For now, I made the update be done using unique indices, which, from what I understood, is thread safe.

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -460,9 +468,19 @@ int Sequencer::run() {
...
   std::atomic<std::size_t> nProcessedEvents = 0;
   std::size_t nTotalEvents = eventsRange.second - eventsRange.first;
+
+  // prepare event timing data
+  bool enableEventTiming = m_cfg.outputEventTimingFile.has_value();
+  std::size_t nEventTimingData = 0;
+  if (enableEventTiming) {
+    nEventTimingData = nTotalEvents * names.size();
+  }
+  std::vector<EventTimingInfo> eventTimingData(nEventTimingData);
```
andiwand, line 481:
```suggestion
  std::vector<EventTimingInfo> eventTimingData;
  if (enableEventTiming) {
    eventTimingData.resize(nTotalEvents * names.size());
  }
```

matter of taste but to me it looks a bit simpler

Examples/Framework/src/Framework/Sequencer.cpp
```diff
@@ -460,9 +468,19 @@ int Sequencer::run() {
     }
   }
 
-  // execute the parallel event loop
+  // loop preparation
   std::atomic<std::size_t> nProcessedEvents = 0;
   std::size_t nTotalEvents = eventsRange.second - eventsRange.first;
+
+  // prepare event timing data
+  bool enableEventTiming = m_cfg.outputEventTimingFile.has_value();
```
andiwand, line 476: we usually use `outputEventTimingFile.empty()` as the switch

Examples/Framework/include/ActsExamples/Framework/Sequencer.hpp
andiwand: I would still remove the optional

CouthuresJeremy: Done

Examples/Framework/src/Framework/Sequencer.cpp
CouthuresJeremy: Is it better to use `.empty()` everywhere rather than using a boolean? 
I would naïvely expect that using `.empty()` would have more negative impact on the performance than using a boolean.

andiwand: I think `empty` would just check against the size which results in an integer zero check so there should not be any overhead. Also, I believe the path is not that hot to leverage these kind of optimization.

CouthuresJeremy: I personally prefer setting a `boolean` and using it everywhere as I am less likely to write a wrong condition by mistake by using it. But that might not be the best way to do it.
Should I change it to `.empty()` instead or leave it like that?

andiwand: No need to change it - also just a personal preference on my side

AJPfleger:
@CouthuresJeremy Hi what's the status on this? :)

CouthuresJeremy:
> @CouthuresJeremy Hi what's the status on this? :)

@AJPfleger 

After a discussion with @asalzburger, it would be better to have the timing at the end of the `Execute` of the algorithms in the chain. I am not an expert of the sequencer and I will not have time to work on it in the following weeks.

I marked this pull request as draft then.

andiwand:
closing due to inactivity

PR #3125, opened 2024-04-19
perf: Use curvilinear for material states in CKF

andiwand:
Instead of transporting the covariance to the surface we can also use a curvilinear surface which is computationally a little bit more optimal. This should not have any physics impacts.


andiwand:
This seems to achieve another 3% speedup

ttbar PU 200
default seeding
fatras

image

andiwand:
> This should do the same thing, but why are the hashes changing? :/

This is expected since the math changes a little we get slightly different results. In physmon we see that the results are still compatible with the reference which is the important thing here.

AJPfleger (dismissed):

andiwand:
This required an update of the Athena CKF reference objects counts. That is a bit surprising but the difference does not look terrible. /cc @timadye

timadye:
> This required an update of the Athena CKF reference objects counts. That is a bit surprising but the difference does not look terrible. /cc @timadye

Sounds great!

In case you know... I wanted to look at the integration test results, but don't see what it is comparing to say they are successful. Can you point me to an example? (even if we ignore the `run_unit_tests` failure)

andiwand:
As far as I know this test here https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38326028 compares the output to https://gitlab.cern.ch/atlas/athena/-/blob/main/Tracking/Acts/ActsConfig/share/ActsCheckObjectCounts.ref?ref_type=heads

Is that what you meant?

timadye:
Ah OK, so `run_unit_tests` fails because it includes the object counts check. The others pass because they just need the job to run without errors.

Thanks, that's what I wanted.
Tim.

timadye:
... and the change in the number of CKF output tracks is small.

PR #3150, opened 2024-04-27
feat: `SympyStepper`

andiwand:
An attempt to use generated code instead of Eigen for stepping, jacobians and covariance transport.


andiwand:
image

andiwand:
current performance measurements

I think after the branch stopper improved we are not extrapolating too far outside of the ID anymore which reduced the impact on the track finding

image

image

Core/src/Propagator/codegen/sympy_stepper_math.hpp
```diff
@@ -0,0 +1,349 @@
...
+  new_J[57] = J[33] * dGdTL[9] + J[41] * dGdTL[10] + J[49] * dGdTL[11] + J[57] +
+              dFdTL[10];
+  new_J[58] = J[34] * dGdTL[9] + J[42] * dGdTL[10] + J[50] * dGdTL[11] + J[58] +
+              dFdTL[11];
+  new_J[59] = J[59] + h * lambda * x35 / dtds;
+  new_J[60] = J[36] * dGdTL[9] + J[44] * dGdTL[10] + J[52] * dGdTL[11] + J[60];
+  new_J[61] = J[37] * dGdTL[9] + J[45] * dGdTL[10] + J[53] * dGdTL[11] + J[61];
+  new_J[62] = J[38] * dGdTL[9] + J[46] * dGdTL[10] + J[54] * dGdTL[11] + J[62];
+  new_J[63] = 1;
+  J[0] = new_J[0];
```
asalzburger, line 284: Are they continuous in memory, ie. could this one memory copy of a block ?

andiwand: I am also not happy with this assignment. I played a bit with it to see if it has any performance implications and it seems like it doesn't. At the moment this is code that is directly generated through sympy but we could overload the behavior with `std::copy` in the future to make it more concise

asalzburger: Ok, sounds good.

andiwand:
I reverted to using the `EigenStepper` in the Examples after testing the `SympyStepper` in the CI. I think for now it makes sense to consistently rely on the `EigenStepper`.

At some point we should invest some time to properly monitor the different steppers CPU and physics wise. That way we can guarantee that they all work as they should.

The user should be able to transparently swap the `EigenStepper` out and use the `SympyStepper` after this PR which should come with some CPU time improvements while physics should stay (mostly up to FP magic) the same.

asalzburger (approved):

asalzburger:
Let's go for it.

acts-project-service:
## :red_circle: Athena integration test results [d7aa30390ea945bf4663eba32e1e100b4ca35c1d]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39290878) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7444274)!

paulgessinger:
As discussed, if we want to rely on this more going forward, we will want to improve the documentation of the generating code. 

I would also suggest we try to avoid committing the generated code to the repository.

andiwand:
> As discussed, if we want to rely on this more going forward, we will want to improve the documentation of the generating code.
> 
> I would also suggest we try to avoid committing the generated code to the repository.

Agreed - Generating the code in a CMake step or pulling it in if this is not possible would be much better. Since it is not a huge amount of code we decided to put it in temporarily. I added a small CI check to see if the generated code is what we expect to avoid this from getting out of sync.

I guess we have a similar problem here https://github.com/acts-project/acts/blob/a48fa8ac4d74f6647c64bb97e14d6d21276636f1/Core/src/Definitions/ParticleDataTable.hpp. Maybe we can find a common solution for this.

Documentations is lacking far behind on this I agree. I will try to put this in as soon as possible

PR #3161, opened 2024-04-30
refactor!: Use track container in Core `CKF` and pass track proxies to delegates

andiwand:
Wires tracks through Core CKF to track branches and passes those tracks into the branch stopper delegate. This allows the user to count experiment specific track quantities on those branches and then cut on those (e.g. pixel holes).

blocked by
- https://github.com/acts-project/acts/pull/3187


andiwand:
@timadye this should allow us to count experiment specific stopping criteria inside the branch stopper without adding state to it but rather store this information on the branches

andiwand:
no significant performance changes measured

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -1175,9 +1136,26 @@ class CombinatorialKalmanFilter {
...
         m_propagator.template makeState(initialParameters, propOptions);
 
-    auto& r = propState.template get<CombinatorialKalmanFilterResult<traj_t>>();
-    r.fittedStates = &trackContainer.trackStateContainer();
-    r.stateBuffer = std::make_shared<traj_t>();
+    auto& r =
+        propState
+            .template get<CombinatorialKalmanFilterResult<track_container_t>>();
+    r.tracks = &trackContainer;
+    r.trackStates = &trackContainer.trackStateContainer();
```
timadye, line 1143: This are a bit of a code smell, keeping pointers to a `findTracks` argument in the result. It was there before this PR, but now we have an extra pointer.
Would it be possible to remove this and pass `trackContainer` through to where it is needed? I guess not, but just thought I'd check. Alternatively, would we lose much by removing the `trackStates` pointer?

timadye (commented):
A couple of comments, while I'm still mid-review.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
andiwand: Yes this is not great. The lifetime should be ensured since these pointers are only used while we are inside the `findTracks` function.

timadye (commented):
As far as I can understand it, this PR looks good. I have a few naive questions:
1. This seems a big change adding the track container just to be able to access the track proxy from the branch stopper. Is this the only benefit, or is it useful in the other delegates or for another reason?
2. I had previously hoped that we could use the track state to count experiment specific track quantities as they were visited. Is this not possible?
3. Before introducing such a change, it would be good to check that it actually can do what it claims to. Would it be possible to have an example in the ActsExamples branch stopper? An actually functional example might be to cut on the number of 1D or 2D measurements.

andiwand:
> 1. This seems a big change adding the track container just to be able to access the track proxy from the branch stopper. Is this the only benefit, or is it useful in the other delegates or for another reason?

Right I also think this is a big change but there are other benefits from that:
- The branches in the CKF are now modeled as `tracks` in the track container. Instead of having `activeTips` which was only valid after we created a track state we can now talk about `braches` which do not necessarily have to have track states.
- Because of that we can also attach the extrapolated track parameters at the target surface to the track within the CKF which was previously not possible. I exercised that in https://github.com/acts-project/acts/pull/3195.

> 2. I had previously hoped that we could use the track state to count experiment specific track quantities as they were visited. Is this not possible?

This is possible by adding custom columns to the track container and counting the stats within the branch stopper now. Although I didn't exercise that yet.

> 3. Before introducing such a change, it would be good to check that it actually can do what it claims to. Would it be possible to have an example in the ActsExamples branch stopper? An actually functional example might be to cut on the number of 1D or 2D measurements.

I agree, I will try to make some reasonable showcase in the Examples. I am not sure if we will be able to merge that but I guess it would be already valuable for just pointing at some code.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -365,7 +364,9 @@ ProcessCode TrackFindingAlgorithm::execute(const AlgorithmContext& ctx) const {
   TrackContainer tracksTemp(trackContainerTemp, trackStateContainerTemp);
 
   tracks.addColumn<unsigned int>("trackGroup");
+  tracks.addColumn<Acts::CombinatorialKalmanFilterTipState>("CkfTipState");
```
timadye, line 367: I don't understand why this has to be done in `Examples`, rather than in Core. There is even a check at L1302 that this is defined, so why not define it by default, or in a Core factory?

If we have to define it here, shouldn't we use `Acts::s_combinatorialKalmanFilterTipStateColumn`?

timadye (commented):
apart from the one comment, this looks good to me.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
andiwand: Maybe this should be put into a `CKF` function - not sure. The main reason I made it explicit on the caller side is that they have to care about this column later. If you copy containers you have to ensure the columns exist before copy is called. We also discussed this briefly @paulgessinger - any opinion?

andiwand: @timadye and I agreed on creating a factory method in the CKF which sets the track container up

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -864,15 +872,14 @@ class CombinatorialKalmanFilter {
           // Add a hole or material track state to the multitrajectory
           std::size_t currentTip = addNonSourcelinkState(
               stateMask, boundState, result, isSensitive, prevTip);
-          result.activeTips.emplace_back(currentTip, tipState);
-
           auto nonSourcelinkState =
-              result.fittedStates->getTrackState(currentTip);
+              result.trackStates->getTrackState(currentTip);
+          currentBranch.tipIndex() = currentTip;
```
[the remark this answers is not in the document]
andiwand: I tried to harmonize these types. Lets see if that solves the problem

paulgessinger:
@andiwand I was just playing around with the setting for now. No need to overcomplicate our process right now. I updated the setting now (to 25%) and it should be updated when the CI runs the next time.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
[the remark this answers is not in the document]
paulgessinger: Good idea, looks easy enough!

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
andiwand: I implemented this now but did not utilize it in the Examples because we abstract the track finding there further

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -1287,9 +1305,25 @@ class CombinatorialKalmanFilter {
...
+        propState
+            .template get<CombinatorialKalmanFilterResult<track_container_t>>();
+    r.tracks = &trackContainer;
+    r.trackStates = &trackContainer.trackStateContainer();
+    r.stateBuffer = std::make_shared<
+        typename track_container_t::TrackStateBackendContainer>();
+
+    if (!trackContainer.hasColumn(tipStateAccessor.key)) {
+      ACTS_ERROR(
+          "Track container does not have the tip state column \"CkfTipState\"");
```
timadye, line 1318: Just checking... I guess this is an `ERROR` because it probably should have been done earlier.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -368,6 +367,13 @@ ProcessCode TrackFindingAlgorithm::execute(const AlgorithmContext& ctx) const {
   tracksTemp.addColumn<unsigned int>("trackGroup");
   Acts::ProxyAccessor<unsigned int> seedNumber("trackGroup");
 
+  // alternatively you can use `CKF::prepareContainer` but that is not available
+  // here
+  tracks.addColumn<Acts::CombinatorialKalmanFilterTipState>(
+      Acts::s_combinatorialKalmanFilterTipStateColumn);
+  tracksTemp.addColumn<Acts::CombinatorialKalmanFilterTipState>(
+      Acts::s_combinatorialKalmanFilterTipStateColumn);
```
timadye, line 375: Wouldn't this work?

```suggestion
    Acts::CombinatorialKalmanFilter::prepareContainer(tracks);
    Acts::CombinatorialKalmanFilter::prepareContainer(tracksTemp);
```
Maybe it would be clearer to put it before `addColumn<unsigned int>("trackGroup")`. I think it's OK to have separate examples' `addColumn`.

timadye (commented):
With 190e9a5, ttbar+PU200 ITk no longer crashes! 😎 
I'm now testing with the later commits and 1000 events.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
timadye: Looks good! I commented on the Examples in-place, so I think we can close this thread.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
andiwand: Right I think we should communicate this somehow to the user otherwise the component access will throw an exception / signal

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
andiwand: I fear we would need the template params of the `Acts::CombinatorialKalmanFilter`. Alternatively we could make the utility a free function I guess?

timadye: ah yes, I see the problem. That will be a problem for all the users of this function, so it's good to fix it now :) . Can we move it to a free function, eg. `Acts::CombinatorialKalmanFilterPrepareContainer()`.

This is a similar issue with had with the `BranchStopperResult`. Maybe we should have made a CKF namespace, but it's probably too late now.

andiwand: added a fee function as a helper

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
paulgessinger: I guess you can get the track state container from the track container? Do we need the extra pointer?

paulgessinger (approved):

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -84,9 +91,10 @@ struct CombinatorialKalmanFilterExtensions {
       Delegate<Result<std::pair<typename candidate_container_t::iterator,
                                 typename candidate_container_t::iterator>>(
           candidate_container_t& trackStates, bool&, const Logger&)>;
-  using BranchStopper =
-      Delegate<BranchStopperResult(const CombinatorialKalmanFilterTipState&,
-                                   typename traj_t::TrackStateProxy&)>;
+  using BranchStopper = Delegate<BranchStopperResult(
+      const CombinatorialKalmanFilterTipState&,
+      const typename track_container_t::TrackProxy&,
+      const typename track_container_t::TrackStateProxy&)>;
```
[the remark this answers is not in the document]
andiwand: It should be - I think it gets only called in case of a new track state which is appended to the end of the track. Either a new measurement or a new hole / material state.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -328,10 +332,15 @@ class CombinatorialKalmanFilter {
 
   const Logger& logger() const { return *m_logger; }
 
+  static constexpr ProxyAccessor<CombinatorialKalmanFilterTipState>
+      tipStateAccessor = ProxyAccessor<CombinatorialKalmanFilterTipState>(
+          hashString(s_combinatorialKalmanFilterTipStateColumn));
```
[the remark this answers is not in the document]
andiwand: That does not compile for some reason. It does not recognize `hashString` as a function

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -628,33 +647,35 @@ class CombinatorialKalmanFilter {
             ACTS_ERROR("Error while acquiring bound state for target surface: "
                        << res.error() << " " << res.error().message());
             result.lastError = res.error();
-          } else if (!result.activeTips.empty()) {
-            const auto& fittedState = *res;
-            std::size_t currentTip = result.activeTips.back().first;
+          } else {
+            auto& [boundParams, jacobian, pathLength] = *res;
+            auto& currentBranch = result.activeBranches.back();
```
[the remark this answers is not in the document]
andiwand: I can turn this into a value since the proxy is not heavy to copy anyways

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -674,8 +695,8 @@ class CombinatorialKalmanFilter {
               typename navigator_t>
     void reset(propagator_state_t& state, const stepper_t& stepper,
                const navigator_t& navigator, result_type& result) const {
-      auto currentState =
-          result.fittedStates->getTrackState(result.activeTips.back().first);
+      auto currentState = result.trackStates->getTrackState(
+          result.activeBranches.back().tipIndex());
```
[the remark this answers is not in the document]
andiwand: Makes sense 👍 I can do that

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/TrackFindingAlgorithm.hpp
```diff
@@ -122,6 +123,14 @@ class TrackFindingAlgorithm final : public IAlgorithm {
     bool stayOnSeed = false;
     /// Compute shared hit information
     bool computeSharedHits = false;
+
+    // Pixel and strip volume ids to be used for maxPixel/StripHoles cuts
+    std::set<Acts::GeometryIdentifier::Value> pixelVolumes;
+    std::set<Acts::GeometryIdentifier::Value> stripVolumes;
```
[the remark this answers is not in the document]
andiwand: Ah sorry this is not used for the whole ID but just for the volume part. Maybe we can use `std::uint32_t` or `std::size_t`? To make this a bit more explicit

andiwand (commented):
Thanks @timadye - I left a few comments

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/TrackFindingAlgorithm.hpp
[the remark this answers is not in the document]
timadye: isn't is better to have the explicit type `Acts::GeometryIdentifier::Value` the same as the lookup type? If that changes size, it should match, I think.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -227,7 +257,7 @@ class BranchStopper {
   }
 
  private:
-  Config m_config;
+  const TrackFindingAlgorithm::Config& m_cfg;
```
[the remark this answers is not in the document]
timadye: yes. and yes. The original copied the cuts config from the parent, but we probably don't want to copy the whole `Config` object.

Examples/Scripts/Python/full_chain_itk.py
```diff
@@ -124,6 +124,11 @@
     ckfConfig=CkfConfig(
         seedDeduplication=True,
         stayOnSeed=True,
+        # ITk volumes from Noemi's plot
+        pixelVolumes={8, 9, 10, 13, 14, 15, 16, 18, 19, 20},
+        stripVolumes={22, 23, 24},
+        maxPixelHoles=1,
+        maxStripHoles=2,
```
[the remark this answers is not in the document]
timadye: I don't think it is needed, just to have an example. But should work for the ODD. You need to find out the volume ids, though.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -205,16 +213,38 @@ class BranchStopper {
...
 
     if (singleConfig == nullptr) {
       ++m_nStoppedBranches;
       return BranchStopperResult::StopAndDrop;
     }
 
+    bool tooManyHolesPS = false;
+    if (!(m_cfg.pixelVolumes.empty() && m_cfg.stripVolumes.empty()) &&
+        !(m_cfg.maxPixelHoles == std::numeric_limits<std::size_t>::max() &&
+          m_cfg.maxStripHoles == std::numeric_limits<std::size_t>::max())) {
```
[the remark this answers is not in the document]
timadye: Right. I wanted to avoid any time taken if the new feature is not used, but we could just check the two sets and not check the max holes. Is that a good compromise?

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/TrackFindingAlgorithm.hpp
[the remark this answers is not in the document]
andiwand: In principle yes. For the volume ID it does not matter much I suppose because they are fairly small numbers. It is just a personal preference on the interface. No need to change it

Examples/Scripts/Python/full_chain_itk.py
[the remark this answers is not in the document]
andiwand: I will add it

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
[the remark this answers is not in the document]
andiwand: Sounds good 👍

andiwand:
@timadye one thing I noticed is that we have a hole cut now in the track selector and in the `CkfConfig`. I guess it is okay for the Examples for now because aligning this looks complicated

timadye:
yeah, and the pix/str hole cuts aren't eta-dependent. Unfortunately the `Acts::TrackSelector` is a core class, so we can't add detector-dependent selections to it. I think that could be improved, but that is a whole other thing.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -280,6 +285,16 @@ struct CombinatorialKalmanFilterResult {
...
 };
 
+/// Prepare the track container for the CombinatorialKalmanFilter
+///
+/// @param trackContainer The track container to be prepared
+template <typename track_container_t>
+static void prepareTrackContainerForCombinatorialKalmanFilter(
+    track_container_t& trackContainer) {
+  trackContainer.template addColumn<CombinatorialKalmanFilterTipState>(
+      s_combinatorialKalmanFilterTipStateColumn);
```
andiwand, line 295: As @paulgessinger noticed: This does not work with all backend because some only support a subset of primitive data types but definitely not custom PODs. We have two options here: Either fan out the properties or go back to store these info in a vector

paulgessinger:
This has some output changes, and the ExaTrkX job seems to fail to build. Will you look into this @andiwand?

andiwand:
I will have a look. I think the output changes appeared with https://github.com/acts-project/acts/pull/3161/commits/255906dbd1fc3ed5f8af2d0b374eb1370bc300ac and exatrkx failed for a while but I didn't realize it. I guess I just have to absorb the braking change there

paulgessinger:
@andiwand I'm just a bit confused why that would be. Does it make sense to you?

andiwand:
The output changed again after I refactored the tip state counting to use the track interface. Looks already way healthier than before. I don't know exactly what causes the difference now but my hunch is that the branch stopper behaves a little bit differently than before.

Looking at the track classification
image
Which shows a reduction in duplicate tracks which also affects all the other plots to change.

paulgessinger:
@andiwand physmon changes largely don't seem to be very large.

andiwand:
I will update the reference

acts-project-service:
## :red_circle: Athena integration test results [ffdf2cc9986903f95edab13614777d0e7e50153e]
Please investigate the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7683431)!

|status|job|report|
|------|---|------|
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770307)||
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770306)||
|:red_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770305)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770304)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770303)||
|:red_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770302)||
|:red_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770301)||
|:red_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770300)||
|:red_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770299)||
|:red_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770298)||
|:red_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770297)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770296)||
|:red_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770295)||
|:red_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770294)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770293)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770292)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770291)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770290)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770289)||
|:red_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770288)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770287)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770286)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770285)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770284)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770283)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770282)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770281)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770280)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770279)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770278) | [:paperclip:](https://acts-athena-outputs.web.cern.ch/test_data18_13TeV_1000evt/pulls/3161/)|
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40770277) | [:paperclip:](https://acts-athena-outputs.web.cern.ch/test_ttbarPU40_reco/pulls/3161/)|

andiwand:
I executed `test_ActsWorkflow` manually with `main--ACTS` and it seems to work

paulgessinger:
So we think it's just the inconsistency?

andiwand:
I think so - I triggered another run with the modified CI setup. Lets see if that goes through. Otherwise I would suggest to merge these PRs at the end of the day to get them into the nightly and then fix potential problems afterwards.

andiwand:
Athena is happy now

paulgessinger (approved):

PR #3166, opened 2024-05-02
fix: Fix Fatras Highland MSC

andiwand:
I believe we should not use the `sqrt(2)` factor here since we are actually working in the plane and not in 3D later.


andiwand:
closing after https://github.com/acts-project/acts/pull/3168

PR #3167, opened 2024-05-02
refactor: Rework Core `CKF` material interactions a bit

andiwand:
Even tho we usually do not put material on the sensitives and don't start branches from material states this seems a bit more correct to me.


andiwand:
> Why the changes from `FullUpdate` to `PostUpdate`? And why the `PreUpdate`?

`PreUpdate` will account for material before we arrive on the surface and `PostUpdate` for the material after the surface.  This is usually not used but should still be handled carefully since we support this option.

When we reset we start from a surface so the `PreUpdate` already happened. `FullUpdate` would double count the `pre` material.

Since we do not distinguish between measurement states and material states in the reset we should also split the material for material surfaces.

AJPfleger (approved):

acts-project-service:
## :red_circle: Athena integration test results [813057475c8c10f3f382dcb816f212c9fde8af18]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/38583365) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7325480)!

PR #3170, opened 2024-05-02
refactor!: Rewrite `BoundaryCheck`; Rename to `BoundaryTolerance` and split out inside checks

andiwand:
- Switch to a variant backend which holds different boundary tolerances
- Break parametrization and box/polygon check apart `BoundaryTolerance` vs `insideAlignedBox` vs `insidePolygon`
- Make tolerances explicitly cartesian or bound
- Implement absolute cartesian tolerance which allows us to extend surfaces by 1mm in the local axes for example
- Implement absolute euclidean tolerance which allows us to extend surfaces by 1mm in any direction for example
- Throw exceptions if requested boundary tolerance is not implemented

blocked by
- https://github.com/acts-project/acts/pull/3204
- https://github.com/acts-project/acts/pull/3205
- https://github.com/acts-project/machines/pull/109


andiwand:
This turned out as a chunky change.

All tests seem to pass now. I still have to check if the jacobian math I changed is correct.

Would any/all of you be up for a (partial) review @asalzburger @AJPfleger @niermann999 ?

benjaminhuth (commented):
A few comments about documentation and throw, let's do the discussed isolation and I will look again.

paulgessinger:
Before reviewing this one, might it be worth it splitting these changes up? It's a pretty big changeset.

andiwand:
> Before reviewing this one, might it be worth it splitting these changes up? It's a pretty big changeset.

I also discussed this with @benjaminhuth but we came to the conclusion that any kind of splitting would break the build.

I could try to reorganize the commits so we could see the refactor separated from the downstream changes.

What do you think @paulgessinger ?

benjaminhuth:
Regarding the separate commits: In the end, for the review it would make no difference, since the whole code must be reviewed at once.

I think we cannot reduce it further, since the majority of the changes are simply updating the interface and cannot be isolated.

paulgessinger:
@andiwand maybe you can prefix the commit messages with numbers, so it's clear which groups of commits go together?

paulgessinger:
Are we removing Mahalanobis distance in favor of Chi2 with this?

andiwand:
I think it is somehow possible to only review commit by commit, right @paulgessinger?

> @andiwand maybe you can prefix the commit messages with numbers, so it's clear which groups of commits go together?

Right - I think I will have to rewrite the commit history anyways. I will try to split this up

> Are we removing Mahalanobis distance in favor of Chi2 with this?

Not intentionally - I thought the chi2 is using the Mahalanobis distance implicitly by transforming with the Jacobian.

paulgessinger:
@andiwand its definitely possible, but you still have to decide which commits to look at together. 

I think mahalanobis distance is separate from the chi2, it's the generalization of the "sigma distance". Maybe it comes out the same thing.

andiwand:
I reorganized the commits and rebased on main

andiwand:
Blocked by a compiler bug which should disappear with the next machines update https://github.com/acts-project/machines/pull/109

Core/include/Acts/Surfaces/SurfaceConcept.hpp
```diff
@@ -51,10 +51,10 @@ concept SurfaceConcept = requires(S s, const S cs, S s2, const S cs2,
     } -> std::same_as<void>;
   {
     cs.isOnSurface(gctx, Vector3{}, Vector3{},
-                   std::declval<const BoundaryCheck&>())
+                   std::declval<const BoundaryTolerance&>())
```
paulgessinger, line 54: Consider using the `requires` clause in L28/29 to create a value of type `const BoundaryTolerance&` instead of using declval.

Core/src/Surfaces/AnnulusBounds.cpp
```diff
@@ -172,19 +174,35 @@ bool Acts::AnnulusBounds::inside(const Vector2& lposition, double tolR,
...
+    const BoundaryTolerance& boundaryTolerance) const {
+  if (boundaryTolerance.isInfinite()) {
+    return true;
+  }
+
+  if (boundaryTolerance.isNone()) {
+    return inside(lposition, 0., 0.);
+  }
+
+  if (auto absoluteBound = boundaryTolerance.asAbsoluteBoundOpt();
```
paulgessinger, line 188: So this `asAbsoluteBoundOpt` is sort of the alternative to returning the variant and visiting it?

Core/src/Surfaces/AnnulusBounds.cpp
```diff
@@ -172,19 +174,35 @@ bool Acts::AnnulusBounds::inside(const Vector2& lposition, double tolR,
...
+                  absoluteBound->tolerance1);
   }
 
   // first check if inside. We don't need to look into the covariance if inside
   if (inside(lposition, 0., 0.)) {
     return true;
   }
 
+  if (!boundaryTolerance.hasChi2Bound()) {
+    throw std::logic_error("not implemented");
```
paulgessinger, line 200: It's cool that we have a good way to decide and communicate this now.

Core/include/Acts/Surfaces/BoundaryTolerance.hpp
```diff
@@ -0,0 +1,154 @@
...
+namespace Acts {
+
+/// Captures different types of boundary tolerances
+/// - Infinite: infinite tolerance i.e. no boundary check
+/// - None: no tolerance i.e. exact boundary check
+/// - AbsoluteBound: absolute tolerance in bound coordinates
+/// - AbsoluteCartesian: absolute tolerance in Cartesian coordinates
+/// - AbsoluteEuclidean: absolute tolerance in Euclidean distance
+/// - Chi2Bound: chi2 tolerance in bound coordinates
+class BoundaryTolerance {
```
paulgessinger, line 28: I think we might need some more documentation about what these cases are, possibly with pictures so make it clear. At least the euclidean distance should be described in a bit more detail here.

Core/src/Surfaces/BoundaryTolerance.cpp
```diff
@@ -0,0 +1,198 @@
...
+    const std::optional<SquareMatrix2>& jacobianOpt) const {
+  if (isInfinite()) {
+    return true;
+  }
+
+  if (isNone()) {
+    return distance == Vector2::Zero();
+  }
+
+  if (auto absoluteBound = getVariantPtr<AbsoluteBound>();
```
paulgessinger, line 143: Nitpick:

```suggestion
  if (const auto* absoluteBound = getVariantPtr<AbsoluteBound>();
```

also applies below.

Core/include/Acts/Surfaces/BoundaryTolerance.hpp
```diff
@@ -0,0 +1,154 @@
...
+  AbsoluteBound asAbsoluteBound(bool isCartesian = false) const;
+  /// Get the tolerance as absolute Cartesian.
+  const AbsoluteCartesian& asAbsoluteCartesian() const;
+  /// Get the tolerance as absolute Euclidean.
+  const AbsoluteEuclidean& asAbsoluteEuclidean() const;
+  /// Get the tolerance as chi2 bound.
+  const Chi2Bound& asChi2Bound() const;
+
+  /// Get the tolerance as absolute bound if possible.
+  std::optional<AbsoluteBound> asAbsoluteBoundOpt(
```
paulgessinger, line 120: This makes a copy right? If this is a performance concern (which I'm thinking it might be), we could instead return a pointer (instead of a reference) and use that to communicate the optional aspect. That's less safe though, of course.

paulgessinger (approved):
Looks good to me, no major feedback.

Core/src/Surfaces/AnnulusBounds.cpp
andiwand: Right this does some logic inside which handles the cases of `None`, `AbsoluteBound`, `AbsoluteCartesian` in case we are cartesian anyways

Core/include/Acts/Surfaces/BoundaryTolerance.hpp
andiwand: True that will be used in a critical path and it results in a copy. Since `AbsoluteBound` is quite small I didn't expect any big overheads from that but this is something we can try to measure.

Core/include/Acts/Surfaces/BoundaryTolerance.hpp
andiwand: True documentation is lacking quite a big even before. Do you think this should be done directly in this PR or it could potentially wait for a follow-up?

andiwand: I tried to improve the doc

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [6010b1f2c22e278c8dc0724c87d3b5129ea305b5]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40717246) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7675624)!

andiwand:
@paulgessinger with https://github.com/acts-project/acts/pull/3170/commits/f0bf983d76d7d0c62f74e10ec87f8210c664f97c I test the infinite bounds case. I tried using the `FpeMonitor` but since it is part of a plugin I wasn't sure if I should wire it in. I also checked the code and only spotted cases where we do `<= tolerance` so in principle we should be safe doing that

paulgessinger:
@andiwand ah I forgot that I ended up moving this into a plugin. 

I think boost test has some mechanism for this, which might be worth a shot: https://www.boost.org/doc/libs/1_85_0/boost/test/execution_monitor.hpp

paulgessinger:
Also, consider having a look at the sonarqube issues, most of them seem fairly easy fixes.

andiwand:
This is currently blocked because the Athena integration breaks. Currently the `BoundaryCheck` symbol leaks into a lot of objects there which would require almost a full rebuild of Muon software to avoid runtime linker errors.

We have two options here
1. Try to adapt the muon headers / libraries so ACTS does not leak through so much code (currently under investigation)
2. Merge this change as the last one in the chain for v36 and buy into integration failures until we get a new nightly that was built against the new symbols

andiwand:
This is not green in Athena - thanks for the intervention @junggjo9 @paulgessinger

.github/workflows/analysis.yml
```diff
@@ -29,7 +29,7 @@ env:
 jobs:
   build_debug:
     runs-on: ubuntu-latest
-    container: ghcr.io/acts-project/ubuntu2404:51
+    container: ghcr.io/acts-project/ubuntu2204:51
```
paulgessinger, line 32: Is this intentional?

paulgessinger (approved):

.github/workflows/analysis.yml
andiwand: Merge failure 🤦  I will revert this

paulgessinger (approved):

PR #3190, opened 2024-05-15
refactor: Simplify layer handling in `Navigator`

andiwand:
The `startSurface` should never be null which is checked and protected against now


andiwand:
I don't observe any CPU performance changes but it looks like the efficiency slightly increases with this change

asalzburger (approved):

PR #3192, opened 2024-05-16
refactor!: Track EDM brush over

andiwand:
- use `calibrated` instead of `measurement`
- expose typedefs
- correct index params


Core/include/Acts/EventData/MultiTrajectory.hpp
```diff
@@ -432,92 +432,94 @@ class MultiTrajectory {
...
 
   /// Retrieve a parameter proxy instance for parameters at a given index
-  /// @param parIdx Index into the parameter column
+  /// @param istate The track state
   /// @return Mutable proxy
   template <bool RO = ReadOnly, typename = std::enable_if_t<!RO>>
-  typename TrackStateProxy::Parameters parameters(IndexType parIdx) {
-    return self().parameters_impl(parIdx);
+  typename TrackStateProxy::Parameters parameters(IndexType istate) {
+    return self().parameters_impl(istate);
```
paulgessinger, line 439: I don't think this is the track state index, or at least it didn't use to be. This is a separate index to allow sharing of parameters + covariances.

This applies to the other parameter and covariance related functions as well.

andiwand: right I will revert these cases

acts-project-service:
## :red_circle: Athena integration test results [0fd5753529ad6816e153c25b2ce9bdb95f7e1d77]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40428391) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7627774)!

paulgessinger:
I guess it does in fact break Athena compilation, doesn't it?

andiwand:
> I guess it does in fact break Athena compilation, doesn't it?

Yes I think at least the `measurement` -> `calibrated` method name change will break Athena. I can patch this tomorrow before we eventually merge this

andiwand:
This seems to break with `rootcling` in the Athena build e.g. https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40986739

Most likely because root is picking the wrong ACTS headers which are not compatible with the changes here.

Maybe we just merge this PR as the last one for v36 or directly before a new nightly is triggered

cc @paulgessinger

Core/include/Acts/TrackFinding/MeasurementSelector.ipp
```diff
@@ -62,19 +62,11 @@ MeasurementSelector::select(
...
-            .template calibratedCovariance<
-                MultiTrajectoryTraits::MeasurementSizeMax>()
-            .data(),
-        trackState.predicted(), trackState.predictedCovariance(),
-        trackState.projector(), trackState.calibratedSize());
+    double chi2 =
+        calculateChi2(trackState.effectiveCalibrated().data(),
+                      trackState.effectiveCalibratedCovariance().data(),
+                      trackState.predicted(), trackState.predictedCovariance(),
+                      trackState.projector(), trackState.calibratedSize());
```
paulgessinger, line 69: Can you leave in a comment clarifying that this does not use the dynamic objects, but only the relevant data pointers?

paulgessinger (commented):
Just one more minor thing.

paulgessinger (approved):

PR #3233, opened 2024-05-29
test: Add option to only build matching unittests

benjaminhuth:
As we have a lot of unit tests, I find it convenient to only enable some of them during developement. This can be done like this with this PR:

```cmake
cmake .. -D ACTS_BUILD_ONLY_MATCHING_UNITTESTS="ActsUnitTestGeo.*"
```

This will only build unit tests that match the pattern (regex).


Tests/UnitTests/Core/EventData/CMakeLists.txt
```diff
@@ -9,7 +9,9 @@ add_unittest(MultiTrajectory MultiTrajectoryTests.cpp)
 add_unittest(TransformHelpers TransformHelpersTests.cpp)
 add_unittest(CorrectedTransformFreeToBound CorrectedTransformFreeToBoundTests.cpp)
 add_unittest(Track TrackTests.cpp)
-target_sources(ActsUnitTestTrack PUBLIC TrackTestsExtra.cpp)
+if(("${ACTS_BUILD_ONLY_MATCHING_UNITTESTS}" STREQUAL "") OR ("ActsUnitTestTrack" MATCHES "${ACTS_BUILD_ONLY_MATCHING_UNITTESTS}"))
+  target_sources(ActsUnitTestTrack PUBLIC TrackTestsExtra.cpp)
+endif()
```
andiwand, line 14: can we rather check if the target exists?

andiwand (commented):
I usually disable them if compile time matters and then enable all again. How much different makes it for you?

AJPfleger:
Could you provide some more details on a use-case?
During my development, only 1 unit test changes and needs to be recompiled. Then I only execute that one.

AJPfleger:
If we converge to use the regex targets, I would like to have a line in the documentation mentioning it.

andiwand:
I guess one advantage of using CMake for this is that you do not have to explicitly build tests with `ninja` and the `all` target will just by default only build the tests which match. It reminds me a bit of the Athena mechanism to select packages to build. Maybe that would also be an option so we can have `+` and `-` listings. But that might be over-engineering

Tests/UnitTests/Core/EventData/CMakeLists.txt
```diff
@@ -9,7 +9,9 @@ add_unittest(MultiTrajectory MultiTrajectoryTests.cpp)
 add_unittest(TransformHelpers TransformHelpersTests.cpp)
 add_unittest(CorrectedTransformFreeToBound CorrectedTransformFreeToBoundTests.cpp)
 add_unittest(Track TrackTests.cpp)
-target_sources(ActsUnitTestTrack PUBLIC TrackTestsExtra.cpp)
+if(("${ACTS_BUILD_ONLY_MATCHING_UNITTESTS}" STREQUAL "") OR ("ActsUnitTestTrack" MATCHES "${ACTS_BUILD_ONLY_MATCHING_UNITTESTS}"))
+  target_sources(ActsUnitTestTrack PUBLIC TrackTestsExtra.cpp)
```
paulgessinger, line 13: Maybe we can just extend the macro to accept multiple source files, and a list of targets. The we can roll these extra calls into the main macro call, and it falls within the if on its own. 

What do you think @benjaminhuth?

benjaminhuth: Yeah, that sounds good.
But before I get my hands on it, we should decide if we want to go forward in this direction at all

benjaminhuth:
I would be fine closing this, since most people seem fine with using ninja + regex, and this should also work fine for me. In the spirit "the less build system code the better".
Adding a mechanism to do module selection with `+` / `-` listings sounds way to overengineered to me.

paulgessinger:
If we contain it to the unit test macro + top level CMake option (so avoid having the extra ifs in the config) I'm on board with this. I also would stick to a single regex you supply with a single option.

benjaminhuth:
Let's close it for now, I don't have the motivation currently to make this clean. If I ever feel like needing this in the future, I can still reopen it.

PR #3241, opened 2024-05-31
revert: Remove target volume estimation from `Navigator` #3217

andiwand:
Reverting #3217 as this has bigger implications in Athena than expected. The point where this fails at the moment is in a special extrapolation where the whole tracking geometry is "cut" by a cylinder after the first pixel barrel layer. Somehow the previous version of the code was able to handle this case, I am not sure how, since we are also violating our navigation concepts with overlaps there.

I would like to follow up on this on at a later stage where I will try to make the change sound with Athena before we bring it in. One idea is to make the `SurfaceReached` aborter more robust by giving it a state (if this is possible) about the intersections we are targeting. This should make it possible to remove the overstepping condition and to safely target cylinders with that aborter even if there are overlaps in the geometry.

PR #3217, opened 2024-05-24
refactor: Remove target volume estimation from `Navigator`

andiwand:
This is an attempt to remove (IMO) unnecessary complexity from the `Navigator`. We have a target volume estimation which runs a couple of times to guess the target volume based on an intersection with the target surface. IMO if we know the target surface we should also know the target volume and the user is responsible for passing this information.

PR #3242, opened 2024-06-01
refactor: Remove target volume estimation from `Navigator`

andiwand:
This is https://github.com/acts-project/acts/pull/3217 after it was reverted with https://github.com/acts-project/acts/pull/3241. There were a couple of failures monitored in Athena which could not be fixed right away.

This is an attempt to remove (IMO) unnecessary complexity from the Navigator. We have a target volume estimation which runs a couple of times to guess the target volume based on an intersection with the target surface. IMO if we know the target surface we should also know the target volume and the user is responsible for passing this information.


acts-project-service:
## :red_circle: Athena integration test results [184a688dc18ff5b9fc1fe95b52f6017baa3764ba]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42089756) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7892920)!

AJPfleger (approved):

PR #3257, opened 2024-06-06
refactor!: Axis improvements: move from detail + CTAD

paulgessinger:
This PR moves the `Axis` types from the `detail` namespace to the top level `Acts` namespace, following what was done for `Grid`.

In addition, I'm adding deduction guides so that you can more conveniently create axes and grids like:
 
```cpp
Axis eqOpen{0.0, 10., 10}; // Open is default
Axis eqBound{AxisBound, 0.0, 10., 10};
Axis eqClosed{AxisClosed, 0.0, 10., 10};
Axis varOpen{{0, 1, 2., 3, 4}}; // Open is default
Axis varBound{AxisBound, {0, 1, 2., 3, 4}};
Axis varClosed{AxisClosed, {0, 1, 2., 3, 4}};

Grid g{Type<double>, eqOpen, varClosed};
```

**Note**: The old way still works, I just think we can use this to make our code a lot less verbose when working with axes and grids.

Finally, I'm adding an `IGrid` base class to all grid types, what allow generically getting the axes as `IAxis`, while also making that method in `Axis` have non-fixed return type size. This is not supposed to be used on any hot path anyway, and is used for grid inspection.


Core/include/Acts/Geometry/SurfaceArrayCreator.hpp
```diff
@@ -374,35 +374,35 @@ class SurfaceArrayCreator {
...
-      detail::Axis<detail::AxisType::Variable, bdtA> axisA(pAxisA.binEdges);
-      detail::Axis<detail::AxisType::Equidistant, bdtB> axisB(pAxisB.min, pAxisB.max, pAxisB.nBins);
+      Axis<AxisType::Variable, bdtA> axisA(pAxisA.binEdges);
+      Axis<AxisType::Equidistant, bdtB> axisB(pAxisB.min, pAxisB.max, pAxisB.nBins);
 
       using SGL = SurfaceArray::SurfaceGridLookup<decltype(axisA), decltype(axisB)>;
       ptr = std::unique_ptr<ISGL>(static_cast<ISGL*>(
             new SGL(globalToLocal, localToGlobal, std::make_tuple(axisA, axisB), {pAxisA.bValue, pAxisB.bValue})));
 
     } else /*if (pAxisA.bType == arbitrary && pAxisB.bType == arbitrary)*/ {
```
andiwand, line 402: outside this PR's changes, but that confused me a bit

andiwand (approved):
LGTM 👍 

What is the breaking aspect of this PR?

paulgessinger:
@andiwand the grid interface type changes because the axis types move namespaces. Maybe we don't consider the Grid constructors to be public API but I wasn't sure. 

We could also typedef the `Axis` types back into the `detail` and deprecate those I suppose.

andiwand:
You mean the `GridAxisGenerators`? If we coupled against a `detail` I would also argue that this could be seen as a fix by moving out of `detail` - up to you!

paulgessinger:
I was just talking about regular old `Grid`. That used to be in `detail` too, but I moved that out a while ago.

Core/include/Acts/Geometry/SurfaceArrayCreator.hpp
paulgessinger: You mean the extra comment?

paulgessinger:
It seems clang cannot reliably deprecate `using` declarations.

Core/include/Acts/Geometry/SurfaceArrayCreator.hpp
andiwand: yes exactly `} else /*if (pAxisA.bType == arbitrary && pAxisB.bType == arbitrary)*/ {`

not a big deal it just popped into my eye. at first it looked like valid code then I saw the comment and then I wondered if this is now an inverted condition or the same condition of the upper `if`.

paulgessinger: I like to do this sometimes to spell out what the condition is that the `else` is catching. 🤷

acts-project-service:
## :red_circle: Athena integration test results [e88b8cce99df1caa7f1df47a678b7e32de95c0fb]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40290952) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7604557)!

paulgessinger:
Patch for Athena is live on the canary.

PR #3193, opened 2024-05-16
refactor!: Template algorithms on track container frontend `TrackContainer`

andiwand:
Instead of pulling all the track container template parameters into the algorithms we can template on the `TrackContainer` and get the details from there.

blocked by
- https://github.com/acts-project/acts/pull/3192

PR #3272, opened 2024-06-11
refactor: Move MLAmbiguitySolver to Core

Corentin-Allaire:
This PR moves the MLAmbiguitySolver to Core, this will allow us to test it more easily with ATLAS in the future. It also removes the DBScan version of this algorithm as it was shown to be way less effective.

## Summary by CodeRabbit

- **New Features**
	- Enhanced performance monitoring capabilities with the addition of machine learning-based metrics.
	- Integrated a new machine learning function for ambiguity resolution in track finding workflows.
	- Introduced a new `TrackTruthMatcher` algorithm to improve track validation.

- **Bug Fixes**
	- Corrected naming conventions for the `CsvSpacePointWriter` class and its methods.

- **Chores**
	- Removed deprecated configurations and algorithms to streamline the codebase.
	- Updated header inclusions to reflect new naming conventions and class structures.


Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
benjaminhuth, line 1: I'm not an expert of Ambiguity solution, but looking at this it raised the questions if it would be possible to unify the interface with the `Acts::GreedyAmbiguitySolution`, of course not in terms of configuration, but in terms of how to call it. The ways to call these to implementations seem very different to me.

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
```diff
@@ -0,0 +1,126 @@
...
+#include <memory>
+#include <string>
+#include <tuple>
+#include <vector>
+
+namespace Acts {
+
+/// Generic implementation of the machine learning ambiguity resolution
+/// Contains method for data preparations
+template <typename AmbiguityNetwork>
```
benjaminhuth, line 27: This seems very generic to me. It would be great to either have a virtual interface or a C++20 concept that further constrains the required type, also maybe some more documentation.
As a third alternative, also 2 `Acts::Delegate` could be a good interface, assuming the algorithm uses two independent neural networks.

CarloVarni, line 27: maybe we can add some concept for  `AmbiguityNetwork` ?

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
```diff
@@ -0,0 +1,126 @@
...
+class AmbiguityResolutionML {
+ public:
+  struct Config {
+    /// Path to the ONNX model for the duplicate neural network
+    std::string inputDuplicateNN;
```
CarloVarni, line 32: let's be explicit
```suggestion
    std::string inputDuplicateNN = "";
```

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
```diff
@@ -0,0 +1,126 @@
...
+        m_logger{std::move(logger)} {}
+
+  /// Associated measurements ID to Tracks ID
+  ///
+  /// @param tracks is the input track container
+  /// @param sourceLinkHash is the hash function for the source link, will be used to associate to tracks
+  /// @param sourceLinkEquality is the equality function for the source link used used to associated hits to tracks
+  /// @return an ordered list containing pairs of track ID and associated measurement ID
+  template <typename track_container_t, typename traj_t,
+            template <typename> class holder_t, typename source_link_hash_t,
```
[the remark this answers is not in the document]
andiwand: https://github.com/acts-project/acts/pull/3193

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
Corentin-Allaire: Good idea, I haven't had the time to fully study the c++20 concept but it is probably a great opportunity !

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+  // The neural network for duplicate classification, the network implementation
+  // is chosen with the AmbiguityNetwork template parameter
+  AmbiguityNetwork m_duplicateClassifier;
+
+  /// Logging instance
+  std::unique_ptr<const Logger> m_logger = nullptr;
+
+  /// Private access to logging instance
+  const Logger& logger() const;
```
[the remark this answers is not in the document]
Corentin-Allaire: I forgot to implement it. Will do

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
Corentin-Allaire: I am not fully sure how we could unify those. The interface is pretty much the same as the one used by the score base ambiguity solving as they both work similarly : score all track then remove the lowest score. While the greedy one use an internal state as it operate recursively on the tracks. 

Maybe we could try to change the score based and the ML solver to a more "state based" system. But since this would mean modify multiple solver maybe that would be better suited for a separate PR ?

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
Corentin-Allaire: This has been added in the form of the AmbiguityNetworkConcept

Corentin-Allaire: This now uses a AmbiguityNetworkConcept

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
```diff
@@ -0,0 +1,135 @@
...
+  };
+  /// Construct the ambiguity resolution algorithm.
+  ///
+  /// @param cfg is the algorithm configuration
+  /// @param logger is the logging instance
+  AmbiguityResolutionML(const Config& cfg,
+                        std::unique_ptr<const Logger> logger = getDefaultLogger(
+                            "AmbiguityResolutionML", Logging::INFO))
+      : m_cfg{cfg},
+        m_duplicateClassifier(m_cfg.inputDuplicateNN.c_str()),
```
benjaminhuth, line 45: Hmm at this point we assume that the `AmbiguityNetwork` is constructible with a `const char *`, which is not required by the concept... 
I think we should either enforce this in the concept or directly pass the `AmbiguityNetwork` in via the `config` and leave the construction to the caller...

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
```diff
@@ -0,0 +1,135 @@
...
+
+namespace Acts {
+
+/// Generic implementation of the machine learning ambiguity resolution
+/// Contains method for data preparations
+template <AmbiguityNetworkConcept AmbiguityNetwork>
+class AmbiguityResolutionML {
+ public:
+  struct Config {
+    /// Path to the ONNX model for the duplicate neural network
```
benjaminhuth, line 32: Nitpick: maybe not write ONNX if this is now a generic concept

Corentin-Allaire: Good point !

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
Corentin-Allaire: You are right, I added a check in the concept to enforce the existence of a constructor with a string

Core/include/Acts/AmbiguityResolution/AmbiguityNetworkConcept.hpp
```diff
@@ -0,0 +1,56 @@
...
+///
+/// @tparam N the type of the network
+template <typename network_t>
+concept AmbiguityNetworkConcept = requires(
+    DummyTrackContainer &tracks,
+    std::unordered_map<std::size_t, std::vector<std::size_t>> &clusters,
+    std::vector<std::vector<float>> &outputTensor, const char *modelPath,
+    network_t &n) {
+  requires TrackContainerFrontend<DummyTrackContainer>;
+
```
CarloVarni, line 45: I would suggest
```suggestion
concept AmbiguityNetworkConcept = TrackContainerFrontend<DummyTrackContainer> && requires(
    DummyTrackContainer &tracks,
    std::unordered_map<std::size_t, std::vector<std::size_t>> &clusters,
    std::vector<std::vector<float>> &outputTensor, const char *modelPath,
    network_t &n) {
```
btw, I don't think the & are necessary here.

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
```diff
@@ -0,0 +1,135 @@
...
+  /// @param tracks is the input track container
+  /// @param sourceLinkHash is the hash function for the source link, will be used to associate to tracks
+  /// @param sourceLinkEquality is the equality function for the source link used used to associated hits to tracks
+  /// @return an ordered list containing pairs of track ID and associated measurement ID
+  template <TrackContainerFrontend track_container_t,
+            typename source_link_hash_t, typename source_link_equality_t>
+  std::multimap<int, std::pair<std::size_t, std::vector<std::size_t>>>
+  mapTrackHits(const track_container_t& tracks,
+               source_link_hash_t&& sourceLinkHash,
+               source_link_equality_t&& sourceLinkEquality) const {
```
CarloVarni, line 64: are these global references necessary? from what I see from this function these may be `const source_link_hash_t&` and `const source_link_equality_t&`

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
```diff
@@ -0,0 +1,135 @@
...
+    std::multimap<int, std::pair<std::size_t, std::vector<std::size_t>>>
+        trackMap;
+    std::size_t trackIndex = 0;
+    // Loop over all the trajectories in the events
+    for (const auto& track : tracks) {
+      // Kick out tracks that do not fulfill our initial requirements
+      if (track.nMeasurements() < m_cfg.nMeasurementsMin) {
+        continue;
+      }
+      std::vector<std::size_t> measurements;
```
CarloVarni, line 83: can you call a `reserve`?

Examples/Algorithms/TrackFindingML/src/AmbiguityResolutionMLAlgorithm.cpp
```diff
@@ -8,18 +8,29 @@
...
 #include "ActsExamples/TrackFindingML/AmbiguityResolutionMLAlgorithm.hpp"
 
+#include "ActsExamples/EventData/IndexSourceLink.hpp"
+#include "ActsExamples/EventData/Measurement.hpp"
 #include "ActsExamples/Framework/ProcessCode.hpp"
 
 #include <iterator>
 #include <map>
 
+std::size_t sourceLinkHash(const Acts::SourceLink& a) {
```
CarloVarni, line 18: If this is used only here, then let's go for internal linkage
```suggestion
static std::size_t sourceLinkHash(const Acts::SourceLink& a) {
```

Examples/Algorithms/TrackFindingML/src/AmbiguityResolutionMLAlgorithm.cpp
```diff
@@ -8,18 +8,29 @@
...
+  return static_cast<std::size_t>(
+      a.get<ActsExamples::IndexSourceLink>().index());
+}
+
+bool sourceLinkEquality(const Acts::SourceLink& a, const Acts::SourceLink& b) {
```
CarloVarni, line 23: same:
```suggestion
static bool sourceLinkEquality(const Acts::SourceLink& a, const Acts::SourceLink& b) {
```

CarloVarni (changes requested):

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
Corentin-Allaire: But in any case it should be of the order of ~12 elements so I am not sure this will make a big difference (the greedy solver uses the same approach I think)

Examples/Algorithms/TrackFindingML/src/AmbiguityResolutionMLAlgorithm.cpp
paulgessinger: I usually do this with an anonymous namespace.

Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
Corentin-Allaire: @CarloVarni This is the only remaining point, I am not sure it make a real difference but what do you think ?

CarloVarni: for 12 elements, that would be something like 5 allocations. And that is for every single track. 
However, since computing the proper size every single time would be prohibitive, I'd suggest to have something like:
```cpp
std::vector<std::size_t> measurements;
for (const auto& track : tracks) {
    measurements.clear();
    for (auto ts : track.trackStatesReversed()) { 
        ...
    }
}
```
i.e. moving the vector outside the track loop and clear it every time. This way you reduce the allocations.

Corentin-Allaire: Yes this seem much better ! I don't know how I didn't think of that earlier. Thanks

CarloVarni (dismissed):

CarloVarni:
@paulgessinger any idea why Bridge CI / linux_physmon is failing?

paulgessinger:
@CarloVarni Seems like all of the ttbar outputs change over on GitLab. Not quite sure why it would not affect the GitHub run. Maybe we have another non-reproducibility issue? (@andiwand)

I don't see anything in this PR that would explain this.

Examples/Io/Csv/include/ActsExamples/Io/Csv/CsvSpacePointWriter.hpp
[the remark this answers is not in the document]
Corentin-Allaire: Your right I found that the naming there was inconsistent and probably corrected it without thinking. It feel like a waste to open a PR to just fix an incorrect name (but I can if you prefer)

CI/physmon/workflows/physmon_trackfinding_ttbar_pu200.py
```diff
@@ -134,13 +136,25 @@
...
+    )
+
     addAmbiguityResolution(
         s,
         AmbiguityResolutionConfig(
             maximumSharedHits=3,
             maximumIterations=100000,
             nMeasurementsMin=6,
         ),
+        tracks="ckf_tracks",
```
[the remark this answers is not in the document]
Corentin-Allaire: tracks is the name of the input track not the output. So nothing gets overwritten.

Corentin-Allaire:
@paulgessinger @CarloVarni It seem the issue with the CI bridge is gone ? Could someone reapprove ?

AJPfleger (approved):
re-approving. the additional changes seem fine :)

PR #3280, opened 2024-06-12
fix: Switch to layer targeting in any case, remove condition to skip to boundaries

benjaminhuth:
This is a fix for an issue encountered by @janheinri 

The issue was that if `startSurface->associatedLayer() == nullptr`, then the navigation would never resolve the layers in the current volume, because it switches to boundary resolution immediately.


benjaminhuth:
tagging @andiwand

andiwand:
Lets see if this 💣 💥 the CI - if this goes green we can trigger the Athena CI and we should also check against ITk standalone

acts-project-service:
## :red_circle: Athena integration test results [3b63722144080201b952325408d223f8f2441b6a]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/39985933) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7555870)!

janheinri:
So in case any feedback might be helpful: This seems to have fixed my issue

andiwand:
This is a good example of how obscure our gen1 navigation code is 🤢

paulgessinger (approved):

PR #3282, opened 2024-06-13
refactor!: Remove `Measurement` from Core

andiwand:
This was relatively long on our wishlist and we hope to be able to make it for v36. The idea here is that `Acts` does not really need a measurement EDM since we abstract this away with `SourceLink` for uncalibrated measurements and `TrackState` for calibrated measurements. A calibrated measurement is transformed using the uncalibrated measurement and the track parameter at a given surface therefore it cannot be reused for other tracks.


andiwand:
cc @paulgessinger

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
```diff
@@ -1,21 +1,254 @@
...
+
+  std::ostream& operator<<(std::ostream& os) const {
+    Acts::detail::printMeasurement(os, static_cast<indices_t>(kSize),
+                                   m_subspace.indices().data(), m_params.data(),
+                                   m_cov.data());
+    return os;
+  }
+
+ private:
+  Acts::SourceLink m_source;
```
paulgessinger, line 163: Should this even have a source link itself? Are we accessing this at all?

paulgessinger (approved):
Looks good to me,

paulgessinger (dismissed):
/cc @fredevb re: removal of `Acts::Measurement` from Core.

acts-project-service:
## :red_circle: Athena integration test results [749489f2d806427a376fa163e3f434f262791c2a]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40180710) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7588258)!

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
andiwand: I think we can remove this

andiwand: This seems to be used in some writer. Maybe we can push this to a second iteration if necessary?

andiwand:
Prepared a patch for this on the canary - should be ready now

andiwand:
This seems to work with Athena now @paulgessinger. `test_ActsWorkflow` seems to be a random failure

paulgessinger (approved):
LGTM

PR #3318, opened 2024-06-20
build: Change the way coverage is enabled

paulgessinger:
Also introduces a number of presets and adds them to our CI configuration instead of inline cmake options.

See also: https://its.cern.ch/jira/browse/ATLINFR-5390 (ATLAS internal)


cmake/ActsCompilerOptions.cmake
```diff
@@ -2,10 +2,23 @@
...
   message(STATUS "Setting default build type: ${CMAKE_BUILD_TYPE}")
-endif() 
+endif()
+
+set(extra_flags "-Wall;-Wextra;-Wpedantic;-Wshadow;-Wno-unused-local-typedefs")
+
+# This adds some useful conversion checks like float-to-bool, float-to-int, etc.
+# However, at the moment this is only added to clang builds, since GCC's -Wfloat-conversion
+# is much more aggressive and also triggers on e.g., double-to-float
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|AppleClang")
```
krasznaa, line 12: I know this is how it was before as well, but why not:

```suggestion
if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```

? If "AppleClang" matches, "Clang" will too.

cmake/ActsCompilerOptions.cmake
```diff
@@ -2,10 +2,23 @@
 if (NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Build type configuration" FORCE)
   message(STATUS "Setting default build type: ${CMAKE_BUILD_TYPE}")
-endif() 
+endif()
+
+set(extra_flags "-Wall;-Wextra;-Wpedantic;-Wshadow;-Wno-unused-local-typedefs")
```
krasznaa, line 7: You're doing quite some back-and-forth conversion here. :confused: Why not do the whole thing with "string operations"?

I.e.

```cmake
set(extra_flags "-Wall -Wextra -Wpedantic -Wshadow -Wno-unused-local-typedefs")
if(...)
   set(extra_flags "${extra_flags} -Wfloat-conversion")
endif()
```

Then you can add the flags coming "from the outside" with the same formalism as well.

cmake/ActsCompilerOptions.cmake
```diff
@@ -2,10 +2,23 @@
...
+  list(APPEND extra_flags "-Wfloat-conversion")
+endif()
 
-set(ACTS_CXX_FLAGS "-Wall -Wextra -Wpedantic -Wshadow -Wno-unused-local-typedefs")
-set(ACTS_CXX_FLAGS_DEBUG "--coverage")
+string(REPLACE " " ";" user_flags "${ACTS_CXX_FLAGS}")
+
+list(APPEND extra_flags ${user_flags})
+
+string(REPLACE ";" " " ACTS_CXX_FLAGS "${extra_flags}")
```
krasznaa, line 20: This code, I believe, will keep the `ACTS_CXX_FLAGS` variable growing at each re-configuration. :frowning: Remember, since it's coming from the command line, `ACTS_CXX_FLAGS` is a cache variable. It's value is remembered. So when you do an incremental build that requires a re-configuration, all the string manipulation will happen once again on this variable.

Unfortunately there's no perfect solution to this that I would know of. :frowning: I can only suggest to declare the `ACTS_CXX_FLAGS_<FOO>` variables properly as cache variable (otherwise they are really of not much use), and then `FORCE` set `CMAKE_CXX_FLAGS_<FOO>` based on them on every re-configuration with some logic.

Though that is also not super nice when trying to use Acts as a sub-project of something else. :thinking:

krasznaa (commented):
Unfortunately there is no clear best practice here that I would know of. :frowning: Let me think a little more of what may be best here...

cmake/ActsCompilerOptions.cmake
paulgessinger: I guess I was trying to avoid the extra space, but might not be important.

cmake/ActsCompilerOptions.cmake
paulgessinger: I guess it depends on what we're actually trying to accomplish. I'm not HUGE fan of the separate `ACTS_CXX` variables to begin with. 

Going by the principle of least surprise, I'd expect to be able to set the C++ flags by setting `CMAKE_CXX_FLAGS` from the outside.

How about checking if our forced `extra_flags` are already present in `CMAKE_CXX_FLAGS`, and if not appending them? This would avoid the cache cycle problem, and allow the flags to be set at the command line or in presets.

cmake/ActsCompilerOptions.cmake
paulgessinger: Probably. The matching was there previously, but I can change it anyway.

cmake/ActsCompilerOptions.cmake
krasznaa: Like we do in every R&D project (and in atlasexternals)?
  - https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-functions.cmake#L140-L160
  - https://github.com/acts-project/detray/blob/main/cmake/detray-functions.cmake#L194-L214
  - ...
  - https://gitlab.cern.ch/atlas/atlasexternals/-/blob/main/Build/AtlasCMake/modules/AtlasCompilerSettings.cmake?ref_type=heads#L83-101

Yeah, I think that's a fine approach. :laughing:

paulgessinger: Hm, yeah but that comes at the cost of using a bunch of user functions, which to me is an antipattern.
I'll stick to the list mechanism and removal of duplicates.

paulgessinger: I've tested this locally, and there's no cache retention problem (`CMAKE_CXX_FLAGS`) is not a cache variable. You can also set it just fine from an enclosing project, so I think it checks all of our boxes.

krasznaa: To be more correct, `CMAKE_CXX_FLAGS` **is** a cache variable "in general", but since in this directory you set it as a regular variable, for Acts it becomes a regular variable. :thinking:

So okay, that should generally be fine then. :thinking: 

But the "bunch of user functions"... I think it's much worse of a pattern that Acts doesn't provide any helper functions for building its libraries/executables. This means that any systematic change has to be done in N places. Making it very easy to arrive at a setup where some subdirectories do thing in one way, while others in some different one. :frowning: I myself am a **big** fan of the project-specific helper functions. (Athena would've drowned already without the `atlas_<foo>(...)` functions.)

paulgessinger: I think we'll have to agree to disagree on this.

The custom functions are a big contributing factor to most people not really knowing what's going on in the CMake configuration, since everything is hidden behind magic functions. For example, we can probably count the number of people who understand how packages **actually** work on one hand. 

I do concede, however, that the ATLAS build is a different beast. 

For ACTS (and imo also for the R&D projects) I think we're much better off without various opaque layers of CMake scripting.

krasznaa (changes requested):
Some further thoughts...

cmake/ActsCompilerOptions.cmake
krasznaa: You know that I personally like CMake to a reasonable amount. But even I don't think that everybody would need to know as much as I do, just to develop some C\+\+ code for Acts. :thinking: Not everybody has to know everything. That's the whole point of using "functions" in the first place, in any language.

So yeah, I'm on a very different opinion.

cmake/ActsCompilerOptions.cmake
```diff
@@ -21,32 +26,19 @@ if(DEFINED CMAKE_CXX_STANDARD)
...
-  set(ACTS_CXX_FLAGS "${ACTS_CXX_FLAGS} -g")
+  list(APPEND cxx_flags "-g")
 endif()
 
-# Acts linker flags
-set(ACTS_EXE_LINKER_FLAGS_DEBUG "--coverage")
-set(ACTS_SHARED_LINKER_FLAGS_DEBUG "--coverage ")
+string(REPLACE " " ";" explicit_flags "${CMAKE_CXX_FLAGS}")
+list(APPEND cxx_flags "${explicit_flags}")
+list(REMOVE_DUPLICATES cxx_flags)
```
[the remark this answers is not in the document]
paulgessinger: And then we ignore duplicates? Probably doesn't break anything, so might be ok.

CMakePresets.json
```diff
@@ -0,0 +1,65 @@
...
+      "displayName": "Build Performance",
+      "inherits": "ci",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "ACTS_BUILD_EVERYTHING": true,
+        "CMAKE_EXPORT_COMPILE_COMMANDS": true
+      }
+    },
+    {
+      "name": "linux_ubuntu",
```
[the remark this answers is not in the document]
paulgessinger: Yeah I wasn't sure about the name anyway. Right now, these are mainly meant for the CI jobs, so I named them after the respective CI jobs. But if we want to use presets more generally, more general names might be desirable.

cmake/ActsCompilerOptions.cmake
```diff
@@ -21,32 +26,19 @@ if(DEFINED CMAKE_CXX_STANDARD)
...
-# However, at the moment this is only added to clang builds, since GCC's -Wfloat-conversion 
-# is much more aggressive and also triggers on e.g., double-to-float
-if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|AppleClang") 
- set(ACTS_CXX_FLAGS "${ACTS_CXX_FLAGS} -Wfloat-conversion")
-endif()
 
 if(ACTS_ENABLE_CPU_PROFILING OR ACTS_ENABLE_MEMORY_PROFILING)
   message(STATUS "Added -g compile flag")
-  set(ACTS_CXX_FLAGS "${ACTS_CXX_FLAGS} -g")
+  list(APPEND cxx_flags "-g")
```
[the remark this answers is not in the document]
krasznaa: Just make sure that a 3.11 CMake minimum version is enforced. Not sure what the current minimum for Acts is... :thinking:

cmake/ActsCompilerOptions.cmake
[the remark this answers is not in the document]
krasznaa: :thinking: On first order, yes... I think it will anyway be very rare that this would come up. But even when it does, the compiler frontends don't tend to mind duplicate flags all too much.

At the same time, a really messed up set of flags like `"-std=c++20 -std=c++17 -std=c++20"` would result in a different outcome with the duplicates removed. :thinking: I think it's because of all of these possibilities that the CMake core code doesn't try to be too smart with this.

paulgessinger:
The failures are a bit confusing. Something's messing up the python binding installation. I'll separate the presets from the change to the coverage flags.

cmake/ActsCompilerOptions.cmake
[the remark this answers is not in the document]
paulgessinger: Ok, switched back to just string composition now.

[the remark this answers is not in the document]
paulgessinger: Can you check again, and hit approve if you don't have further suggestions?

cmake/ActsCompilerOptions.cmake
```diff
@@ -2,13 +2,18 @@
...
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  set(cxx_flags "${cxx_flags} -Wfloat-conversion")
+endif()
 
-set(ACTS_CXX_FLAGS "-Wall -Wextra -Wpedantic -Wshadow -Wno-unused-local-typedefs")
-set(ACTS_CXX_FLAGS_DEBUG "--coverage")
-set(ACTS_CXX_FLAGS_MINSIZEREL "")
-set(ACTS_CXX_FLAGS_RELEASE "")
-set(ACTS_CXX_FLAGS_RELWITHDEBINFO "")
+# set(ACTS_CXX_FLAGS "${extra_flags}" CACHE STRING "Extra C++ compiler flags")
```
andiwand, line 16: dead code?

krasznaa (dismissed):
Apart from Andi's dead code comment, the rest seems okay like this. :thinking: At least as okay as I can think of. Since this type of code is just inherently always a little fragile. :frowning:

cmake/ActsCompilerOptions.cmake
paulgessinger, line 16:
```suggestion
```

andiwand (approved):

acts-project-service:
## :red_circle: Athena integration test results [a75cf66ed0414e384781a8424f969bb5e108ec95]
Please investigate the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7653799)!

|status|job|report|
|------|---|------|
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588483)||
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588482)||
|:green_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588481)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588480)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588479)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588478)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588477)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588476)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588475)||
|:green_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588474)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588473)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588472)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588471)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588470)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588469)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588468)||
|:red_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588467)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588466)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588465)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588464)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588463)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588462)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588461)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588460)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588459)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588458)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588457)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588456)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588455)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588454)||
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40588453)||

PR #3320, opened 2024-06-20
feat: traccc plugin (common functionality and CPU chain)

fredevb:
Added traccc to examples (Acts/Examples/Traccc), making it possible to run the traccc chain on CPU.

### In Acts/Examples/Traccc/Common:

- Added functions to connect the EDM in examples to the traccc edm conversion functions provided by the traccc plugin, enabling conversion of the example EDM to the traccc EDM.
- Introduced features for debugging and ensuring that measurements created by ACTS and traccc are consistent. Also, added code to associate traccc measurements with their corresponding ACTS measurements (as the ordering in the measurement containers may be different). This is required for truth matching since we need to know which hits caused which measurements.
- Developed a base traccc chain algorithm class, containing code that is independent of whether the chain runs on CPU or GPU.

### In Acts/Examples/Traccc/Host:

- Implemented the traccc chain algorithm to run on CPU.

Additionally, added Python bindings for the traccc chain and its configuration.


andiwand:
closing due to massive conflicts and inactivity

PR #3328, opened 2024-06-24
chore: Fix a number of SonarQube reliability issues

paulgessinger:
This PR 

- Modifies the logging macros so that they require a `;` now
- Modifies the DD4hep conversion to not leak memory, but instead attach the ACTS detector element lifetime to the DD4hep detector element
- Adds an explicit cast from `uint64_t` to `uint32_t`
- Makes a number of constructors `explicit` and/or `noexcept`
- Avoids some problematic integer manipulations
- Avoids throwing `std::string` as an exception
- \+ a few smaller fixes


benjaminhuth (approved):
Looks good!

PR #3335, opened 2024-06-28
feat: `Surface::toStream` overload without stream argument

paulgessinger:
I'm adding a way that allows you to avoid having to write

```cpp
surface->toStream(gctx, std::cout);
std::cout << std::endl;
```

and instead write

```cpp
std::cout << surface->toStream(gctx) << std::endl;
```

This is done by returning a temporary helper struct that implements `operator<<`.

### DIFF
--- a/Core/include/Acts/Geometry/GeometryContext.hpp
+++ b/Core/include/Acts/Geometry/GeometryContext.hpp
@@ -25,6 +25,35 @@ namespace Acts {
 
 using GeometryContext = ContextType;
 
+/// Helper struct that stores an object and a context, and will print it to
+/// an outstream.
+/// This allows you to write
+/// ```cpp
+/// std::cout << surface->toStream(geoContext) << std::endl;
+/// ```
+template <typename T>
+struct GeometryContextOstreamWrapper {
+  GeometryContextOstreamWrapper(const T& object, const GeometryContext& gctx)
+      : m_object(object), m_gctx(gctx) {}
+
+  GeometryContextOstreamWrapper(const GeometryContextOstreamWrapper&) = delete;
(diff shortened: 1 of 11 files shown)


andiwand:
I think @benjaminhuth did something similar here https://github.com/acts-project/acts/pull/1327 which requires `std::tie`ing the surface and context

benjaminhuth:
yeah, maybe this is more intuitive than `std::cout << std::tie(surface, ctx) << std::endl`. It will definitively give better compiler errors :D

andiwand:
Should we replace it then and converge on one version?

benjaminhuth:
Would be fine for me.

benjaminhuth:
Could you add this also to the doxygen description of `Acts::Surface`, so that people easily find this method @paulgessinger ?

paulgessinger:
Ha, I had forgotten all about #1327, I was playing around with a `tuple` argument, but I didn't actually know it would work with `tie`.

Up to you, I can also close this in favor of just sticking to #1327.

paulgessinger:
@andiwand That's a breaking change then?

benjaminhuth:
@paulgessinger Really up to you... If your unawareness of `std::tie` is a hint that this is to non-obvious, we can use this. Otherwise we just have one PR less, which is also not too bad^^

benjaminhuth (approved):
Let's get it in!

acts-project-service:
## :red_circle: Athena integration test results [7d825f614faa5688281fb75d3d49ff33725b2875]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/40682654) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7670183)!

PR #2830, opened 2023-12-14
refactor: Refactor `AdaptiveGridTrackDensity`

andiwand:
Sorry this got more radical then I anticipated

Summary of the changes:
- dynamic `spatialTrkGridSize` and `temporalTrkGridSize`
  - replaced it with a sigma scale and maximum window for space and time
- refactored the optional time handling
- truncated pair construction
- tried to simplify some of the density map handling e.g. default zero, no need to count

PR #2745, opened 2023-11-29
refactor!: make trkGridSize non-template parameter in AdaptiveGridTrackDensity

felix-russo:
BREAKING CHANGE: `spatialTrkGridSize` and `temporalTrkGridSize` need to be set in `Config` instead of being passed as template parameters. Concerns `AdaptiveGridTrackDensity` and `AdaptiveGridDensityVertexFinder`.

PR #2460, opened 2023-09-18
feat: time vertex seeding

felix-russo:
WIP

PR #3340, opened 2024-07-01
feat!: Resurrect original `AdaptiveGridTrackDensity`; Move new impl to `SparseGridTrackDensity`

andiwand:
This is an attempt to resurrect the original `AdaptiveGridTrackDensity` to compare it's performance against the new `SparseGridTrackDensity`.

The main reason for this is that the algorithms diverged significantly and it would be good to capture this by different names. This also allows for CPU and physics performance comparisons which is currently looked at.

blocked by
- https://github.com/acts-project/acts/pull/3352


andiwand:
This is a bit more complicated than I anticipated. There have been a serious of changes to the `AdaptiveGridTrackDensity`, some of them we want to keep and some of them we want to revert.
- revert https://github.com/acts-project/acts/pull/2338
- keep https://github.com/acts-project/acts/pull/2387
- keep https://github.com/acts-project/acts/pull/2376
- revert https://github.com/acts-project/acts/pull/2460
- revert https://github.com/acts-project/acts/pull/2745
- revert https://github.com/acts-project/acts/pull/2830

I think this has to be done manually since this also interfered with large scale vertexing refactor. Overall we should keep the improved comments and interfaces while the algorithm should stay as it was with its data structures.

paulgessinger:
You don't think we can keep #2745 @andiwand?

andiwand:
> You don't think we can keep #2745 @andiwand?

I would definitely keep this in `SparseGridTrackDensity` which is the current implementation of `AdaptiveGridTrackDensity`. Looking at the old implementation of `AdaptiveGridTrackDensity` it seems to be that the constant size has performance benefits. If there is an alternative I would also like to drop the template param.

paulgessinger:
Did we not measure this back then? I'd be extremely surprised if that's what makes a difference in performance here.

andiwand:
I don't think we did a full comparison of the `std::unordered_map` vs the `std::vector` implementation. Part of this is that track density objects can live on the stack with known size.

andiwand:
This is the performance I measured for CPU and physics. Looks surprisingly comparable to me.

image

I am not sure if we want to proceed with porting the original `AdaptiveGridTrackDensity` now. One reason to do so is that the algorithm quite diverged and it might be confusing to keep the name for the newer version. Also the newer one was extended to time while the old one was not adoptable in this direction. On the other hand the principle is quite the same while we use a different data structure as a backend.

andiwand:
dropping this for now as there are no apparent gains in merging this

PR #3348, opened 2024-07-03
feat: Introduce the usage of the geomodel converters for the Muon System

dimitra97:
This PR introduces the usage of the Geomodel Converters for the conversion of the tubes of the MDTs chambers as a starting point. I guess there are some things for discussion.
I want to clarify some things though:
1. The sqlite database that I use is:  ATLAS-R3-MUONTEST_v3.db 
 1.a I have checked it with the Inner Detector and the `geomodel.py` script b`y python3 geomodel.py -i ATLAS-R3-MUONTEST_v2.db -q PIXEL SCT  --output-obj -o id -ml std::Silicon`, without using the blueprint for now just to get the sensitive surfaces with Pixel and SCT and it works, this is what I got: 
image

 2. I needed to change some things in the `GeomodelDetectorSurfaceFactory` and the python script `geomodel.py` that we use to generate the converted surfaces and dump them in the obj file:
  2.a The Muon System's database apparently has a different hierarchy that the one of the ID. The top-level full physical volume is actually the MDT Multilayer, the children of which are the tubelayers and the tubefoam. I am interested in the children of the tubelayers which are the sensitive tubes I want to convert. For this reason:
- The` convert-subvolumes` argument was introduced, that allows for subvolumes extraction from the hierarchy.
- These are not full phys vol, but `PVConstLink`  from Geomodel. So I also needed to change what the converters accept as arguments (Now they accept the PVConstLink volume and the transform that would be handled before passing to the converter). 
-The DetectorElement does not hold any more the full physical volume though and I do not know if this will be a problem (I do not know what is happening with the Alignment for example etc, if it will be needed there).

 2.b The Muon System is huge, so if I build everything (all the tubes from all the chambers) the file will be huge. For visualization     reasons I changed the `match function`.
 - I can define the name of the chamber that I want to build (e.g ` BIL3_MDT02_-1_4_2 `and the name comes from the databse). In case no name and no material provided, it goes and builds everything. 
 - In general it checks both material and name (because for the sensitive I want to build Tubes that have the material `std::Aluminium` ) and if I only say the name `Tube` it will also build the layers and the foams that carry a different material. If the material list is empty or we have a full phys vol ---> it checks just the name list.
 If the name list is empty ---> it checks just the material list

3.  I changed how the `converters` are picked up and turned it to a function with a map, in order not to loop every time to all the converters passed in the configuration.

For the Muon System, if I run the python script and get the obj with the surfaces by: 
`python3 geomodel.py -i ATLAS-R3-MUONTEST_v2.db -q Muon --convert-subvols --output-obj -o muons -ml Aluminium -n Tube BIL3_MDT02_-1_4_2`
I get this:

image
for now just floating tubes, but we will continue with the other technologies with the summer student.

4. Also, one last thing is that in order to use some functions that allow to get the subvolumes from the full physical volume, I neede to change the tag of Geomode. But this can also be discussed. 

5. Also by ` python3 geomodel.py -i ATLAS-R3-MUONTEST_v3.db -q Muon --output-obj -o muons  -n MDT -ml Air` (this picks the names and the materials of the top-level physical volumes which are the bounding boxes of the tubes --> multilayers). For the moment it is only converted as plane surfaces. **Next step with the summer student:**  Get the converters that convert them to detector volumes (most probably take advantage of what @Matthewharri has implemented!! )
image

Sorry for the big explanation.

I am pinging some people that may be interested and I have touched their code: @asalzburger @benjaminhuth @noemina @junggjo9 @paulgessinger 
 

## Summary by CodeRabbit

- **New Features**
	- Enhanced command-line interface for geometry modeling with new options for input handling (`--name-list`, `--material-list`, `--convert-subvols`, and `--enable-blueprint`).
	- Improved functionality in `GeoModelDetectorSurfaceFactory` for handling subvolumes and sensitive surfaces.
	- Added new function for mapping geometric shape IDs to converter instances.

- **Bug Fixes**
	- Updated internal references to ensure compatibility with new types in various conversion methods.

- **Documentation**
	- Updated CMake configuration to include new dependency checks for `GeoModelCore` and `GeoModelIO`.

- **Tests**
	- Refactored tests to utilize smart pointers for improved memory management and updated parameter types for method calls.


Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelConverters.hpp
```diff
@@ -68,4 +69,24 @@ using GeoUnionDoubleTrdConverter =
     detail::GenericGeoShapeConverter<GeoShapeUnion,
                                      detail::GeoUnionDoubleTrdConverter>;
 
+/// @brief The map that maps the converters with the shapes
+
+std::shared_ptr<const IGeoShapeConverter> GeoShapesConverters(int geoShapeId) {
+  using ConverterLookUpMap =
+      std::map<int, std::shared_ptr<const IGeoShapeConverter>>;
```
benjaminhuth, line 76: `std::unordered_map` or `boost::flat_map` would be better suited the use case here I think.

In the future, I think there might be the possibilities that you have multiple matching converters per ShapeType (e.g., multiple union converter), so than we would need a `multi_map` and return a `vector`... but I think we can worry about this in the future when it happens.

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
```diff
@@ -81,62 +97,100 @@ void Acts::GeoModelDetectorSurfaceFactory::construct(
...
+  const std::vector<GeoChildNodeWithTrf> children =
+      getChildrenWithRef(geoPV, false);
+  std::vector<GeoChildNodeWithTrf> foundVols{};
+
+  for (const auto &child : children) {
+    if (matchFunc(child.nodeName, child.volume)) {
+      ACTS_VERBOSE("The subvol" << child.nodeName << "mathes the queries");
+      foundVols.push_back(child);
+    }
+    if (!child.volume->getNChildVols()) {
```
benjaminhuth, line 181: Probably also clang-tidy will complain about this, but we should not implicit convert int to bool like this.
```suggestion
    if (child.volume->getNChildVols() != 0) {
```

benjaminhuth (commented):
I have a first round of comments, nothing very critical here.

I like the changes, especially it seems more logically to get the converters from the shape type and not pass them in from python.
In the future we might want to generalize the subvolume search a bit (the `else` statement which currently allows either to convert the top level volume or the subvolumes).

dimitra97:
> I have a first round of comments, nothing very critical here.
> 
> I like the changes, especially it seems more logically to get the converters from the shape type and not pass them in from python. In the future we might want to generalize the subvolume search a bit (the `else` statement which currently allows either to convert the top level volume or the subvolumes).

Thanks Benjamin for the comments! Yes I agree with you with the else statement I am also not happy. As we discussed also offline, for the moment I had it like that because for the Muons the top-level physical volume is actually a bounding box, which is the multilayer and has the shape of std::Trd. So from the converters it will just pick the trapezoid converter and it will be turned into a plane surface which does not make sense. I am thinking about also converting the bounding box to a volume and then the else statement could be gone actually.

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelConverters.hpp
dimitra97: changed to unorder_map!

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
```diff
@@ -81,62 +97,100 @@ void Acts::GeoModelDetectorSurfaceFactory::construct(
...
+void Acts::GeoModelDetectorSurfaceFactory::convertSensitive(
+    PVConstLink geoPV, const Acts::Transform3 &transform,
+    std::vector<Acts::GeoModelSensitiveSurface> &sensitives) {
+  const GeoLogVol *logVol = geoPV->getLogVol();
+  const GeoShape *shape = logVol->getShape();
+  int shapeId = shape->typeID();
+  std::string name = logVol->getName();
+  std::shared_ptr<const Acts::IGeoShapeConverter> converter =
+      Acts::GeoShapesConverters(shapeId);
+  if (converter != nullptr) {
```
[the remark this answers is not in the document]
dimitra97: changed!

Examples/Scripts/Python/ATLAS-R3-MUONTEST_v2.db
andiwand, line 1: is this file supposed to be committed here?

Examples/Scripts/Python/geomodel.py
```diff
@@ -98,11 +130,14 @@ def main():
...
-    gmFactoryConfig.shapeConverters = [
-        gm.GeoBoxConverter(),
-        gm.GeoTrdConverter(),
-        gm.GeoIntersectionAnnulusConverter(),
-    ]
+    # gmFactoryConfig.shapeConverters = [
+    #     gm.GeoBoxConverter(),
+    #     gm.GeoTrdConverter(),
+    #     gm.GeoIntersectionAnnulusConverter(),
+    # ]
```
andiwand, line 137: temporary change?

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelConverters.hpp
```diff
@@ -21,6 +21,7 @@
 #include "Acts/Plugins/GeoModel/detail/GeoUnionDoubleTrdConverter.hpp"
 #include "Acts/Utilities/Result.hpp"
 
+#include <map>
```
andiwand, line 24:
```suggestion
#include <unordered_map>
```

will require reformatting

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorElement.hpp
```diff
@@ -15,7 +15,10 @@
 
 #include <memory>
 
+#include <GeoModelKernel/GeoFullPhysVol.h>
+
 class GeoFullPhysVol;
```
andiwand, line 20: I guess this forward declare can be dropped after the include?

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorElement.hpp
```diff
@@ -48,9 +51,8 @@ class GeoModelDetectorElement : public DetectorElementBase {
   /// @return a shared pointer to an instance of the detector element
   template <typename SurfaceType, typename BoundsType>
   static std::shared_ptr<GeoModelDetectorElement> createDetectorElement(
-      const GeoFullPhysVol& geoPhysVol,
-      const std::shared_ptr<BoundsType> bounds, const Transform3& sfTransform,
-      ActsScalar thickness) {
+      PVConstLink geoPhysVol, const std::shared_ptr<BoundsType> bounds,
```
andiwand, line 54: Is `PVConstLink` a typedef from GeoModel? I wonder if we should rather unroll that in ACTS since we usually want pointer types to be explicit.

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorSurfaceFactory.hpp
```diff
@@ -38,12 +40,16 @@ class GeoModelDetectorSurfaceFactory {
...
-    /// The shape converters to be use
-    std::vector<std::shared_ptr<const IGeoShapeConverter>> shapeConverters = {};
-    /// List for names to match
+    // /// The shape converters to be use
+    // std::vector<std::shared_ptr<const IGeoShapeConverter>> shapeConverters =
+    // {};
+    // /// List for names to match
     std::vector<std::string> nameList;
-    /// List for materials to match
+    // /// List for materials to match
```
andiwand, line 48:
```suggestion
    /// List for materials to match
```

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorSurfaceFactory.hpp
```diff
@@ -38,12 +40,16 @@ class GeoModelDetectorSurfaceFactory {
 
   // Configuration struct for the Detector surface factory
   struct Config {
-    /// The shape converters to be use
-    std::vector<std::shared_ptr<const IGeoShapeConverter>> shapeConverters = {};
-    /// List for names to match
+    // /// The shape converters to be use
+    // std::vector<std::shared_ptr<const IGeoShapeConverter>> shapeConverters =
+    // {};
+    // /// List for names to match
```
andiwand, line 46: dead code?

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
```diff
@@ -6,16 +6,19 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-#include "Acts/Plugins/GeoModel/GeoModelDetectorSurfaceFactory.hpp"
+// Needs to be included first to avoid some conflict between forward declare and
+// typedef from Geomodel side
+//  clang-format off
```
andiwand, line 11:
```suggestion
// clang-format off
```
nitpick

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
```diff
@@ -55,24 +59,37 @@ void Acts::GeoModelDetectorSurfaceFactory::construct(
...
-    auto matches = [&](const std::string &name, const GeoVFullPhysVol &fpv) {
-      bool match = m_cfg.nameList.empty() && m_cfg.materialList.empty();
-      if (match) {
+    auto matches = [&](const std::string &name, PVConstLink physvol) {
+      if (m_cfg.nameList.empty() && m_cfg.materialList.empty()) {
         return true;
       }
 
-      match |= std::any_of(
+      bool matchName{false}, matchMaterial{false};
```
andiwand, line 67:
```suggestion
      bool matchName{false};
      bool matchMaterial{false};
```

nitpick; clang-tidy might also pick on that

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
```diff
@@ -55,24 +58,37 @@ void Acts::GeoModelDetectorSurfaceFactory::construct(
...
+
+      matchName |= std::any_of(
```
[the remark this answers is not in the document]
dimitra97: it changed!

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
```diff
@@ -6,16 +6,18 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+// clang-format off
```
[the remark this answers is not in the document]
dimitra97: I added the comment

Examples/Scripts/Python/geomodel.py
dimitra97: ah no, sorry forgot to remove!!

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorElement.hpp
dimitra97: Yes it is from Geomodel.

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
dimitra97: actually I removed those:)

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorElement.hpp
junggjo9: You could also just rely on the GeoVFullPhysVol

junggjo9:
I would not put the ATLAS database file into the repository.

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorElement.hpp
dimitra97: You are right!

Plugins/GeoModel/include/Acts/Plugins/GeoModel/detail/GeoTubeConverter.hpp
```diff
@@ -35,7 +35,7 @@ struct GeoTubeConverter {
   /// @param bool sensitive
   ///
   /// @return The detector element and surface
-  Result<GeoModelSensitiveSurface> operator()(const GeoFullPhysVol& geoFPV,
+  Result<GeoModelSensitiveSurface> operator()(PVConstLink geoPV,
```
junggjo9, line 38: Do you want to parse them by reference?

Plugins/GeoModel/include/Acts/Plugins/GeoModel/detail/GeoUnionDoubleTrdConverter.hpp
```diff
@@ -16,6 +16,7 @@
 #include <memory>
 #include <tuple>
 
+#include <GeoModelKernel/GeoFullPhysVol.h>
```
junggjo9, line 19: That's not really needed, I guess

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
```diff
@@ -81,62 +96,103 @@ void Acts::GeoModelDetectorSurfaceFactory::construct(
     std::set<std::string> materials;
 
     for (auto &[name, fpv] : qFPV) {
-      if (fpv == nullptr) {
-        ACTS_WARNING("Pointer to volume '" << name << "' is null");
+      PVConstLink physVol{fpv};
+      // GeoIntrusivePtr<const GeoVPhysVol> physvol{fpv};
```
junggjo9, line 101: Commented line.

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
```diff
@@ -81,62 +96,103 @@ void Acts::GeoModelDetectorSurfaceFactory::construct(
...
+    PVConstLink geoPV, const Acts::Transform3 &transform,
+    std::vector<Acts::GeoModelSensitiveSurface> &sensitives) {
+  const GeoLogVol *logVol = geoPV->getLogVol();
+  const GeoShape *shape = logVol->getShape();
+  int shapeId = shape->typeID();
+  std::string name = logVol->getName();
+  std::shared_ptr<const Acts::IGeoShapeConverter> converter =
+      Acts::GeoShapesConverters(shapeId);
+  if (converter == nullptr) {
+    ACTS_DEBUG("converter is a nullptr");
```
junggjo9, line 149: No puff? If a converter is not available, I'd expect more brimbamborium.

dimitra97: fiiine

dimitra97: actually it can be like that I guess, in case the shape is not found, not need to puff? just move on

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorElement.hpp
andiwand: okay 👍 I guess there is not much we can do if GeoModel wants it this way

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelConverters.hpp
```diff
@@ -68,4 +69,24 @@ using GeoUnionDoubleTrdConverter =
     detail::GenericGeoShapeConverter<GeoShapeUnion,
                                      detail::GeoUnionDoubleTrdConverter>;
 
+/// @brief The map that maps the converters with the shapes
+
+std::shared_ptr<const IGeoShapeConverter> GeoShapesConverters(int geoShapeId) {
+  using ConverterLookUpMap =
+      std::unordered_map<int, std::shared_ptr<const IGeoShapeConverter>>;
```
paulgessinger, line 76: I guess the local typedef is not really needed since you only use it in this location

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
```diff
@@ -6,16 +6,20 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+// Needs to be included first to avoid some conflict between forward declare and
+// typedef from Geomodel side
+// clang-format off
+#include "Acts/Plugins/GeoModel/GeoModelTree.hpp"
```
paulgessinger, line 12: This seems funky. Can you elaborate what this is for?

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
paulgessinger: Is this something you expect to happen and move on, or does this usually indicate some underlying issue that should be fixed?

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
```diff
@@ -81,62 +97,102 @@ void Acts::GeoModelDetectorSurfaceFactory::construct(
...
+  auto converted = converter->toSensitiveSurface(geoPV, transform);
+  if (converted.ok()) {
+    sensitives.push_back(converted.value());
+    const auto &[el, sf] = converted.value();
+
+    ACTS_VERBOSE("(successfully converted: "
+                 << name << " / " << recType(*shape) << " / "
+                 << logVol->getMaterial()->getName() << ")");
+
+    if (!(el && sf)) {
```
paulgessinger, line 161:
```suggestion
    if (!el || !sf)) {
```
?

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
```diff
@@ -81,62 +97,102 @@ void Acts::GeoModelDetectorSurfaceFactory::construct(
...
+      throw std::runtime_error("The Detector Element or the Surface is nllptr");
```
paulgessinger, line 162:
```suggestion
      throw std::runtime_error("The Detector Element or the Surface is nullptr");
```

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
```diff
@@ -81,62 +97,102 @@ void Acts::GeoModelDetectorSurfaceFactory::construct(
...
+    }
+    return;
+  }
+  ACTS_DEBUG(name << " / " << recType(*shape)
```
paulgessinger, line 166: Is this an error?

cmake/ActsExternSources.cmake
```diff
@@ -31,7 +31,7 @@ set( ACTS_FRNN_SOURCE
 mark_as_advanced( ACTS_FRNN_SOURCE )
 
 set( ACTS_GEOMODEL_SOURCE
-   "GIT_REPOSITORY;https://gitlab.cern.ch/GeoModelDev/GeoModel;GIT_TAG;4.6.0;" CACHE STRING "Source to take GeoModel from")
+   "GIT_REPOSITORY;https://gitlab.cern.ch/GeoModelDev/GeoModel;GIT_TAG;5.3.0;" CACHE STRING "Source to take GeoModel from")
```
paulgessinger, line 34: Will 6.2.0 require additional changes?

paulgessinger (commented):
What's the purpose of `PVConstLink`? Does it have reference semantics?

cmake/ActsExternSources.cmake
dimitra97: This I will need to check. I think I will make this PR draft until I see what is happening with 6.2.0

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
dimitra97: I do not think that it indicates an underlying issue, because if we get back a nullptr it would just mean that the converter for this shape is not in the `unordered_map` (for the moment that would mean that the converter is not implemented). 
I do not know if we really want to throw exception here.
I guess it depends if we want to convert everything or also skip some shapes and continue with others that can be converted?

Plugins/GeoModel/src/GeoModelDetectorSurfaceFactory.cpp
dimitra97: So actually, we need to include on the top the `ReadGeomode.hpp `(which is included in the GeomodelTree.hpp) because it has the `#define _GeoShapePersistification_On_` that if it is not defined before including the GeoKernel files then it goes and does the forward declararion of the `class Persistifer ` and creates the conflict declaration

dimitra97:
Ok so for the 6.2 Geomodel, nothing really special needed to change and also the database is now: ATLAS-R3-MUONTEST_v3.db

dimitra97:
> What's the purpose of `PVConstLink`? Does it have reference semantics?

it is a smart pointer defined in geomodel

dimitra97:
This was done for the summer student's project in order to use it and develop on top of it. So it was done, I am closing it.

PR #2925, opened 2024-02-05
refactor: unify the GPU device selection ExaTrkX; local variables declaration in FRNN lib 

hrzhao76:
This PR moves `deviceHint` in previous implementations to the Config constructor, and create a `torch::Device` type with some protections to ensure both model and input tensors are loaded to a specific GPU. The base of FRNN repo is changed to avoid declaration of global variables in CUDA codes which causes segmentation fault in run time when running with Triton Inference Server. 

Tagging ExaTrkX aaS people here
@xju2 @ytchoutw @asnaylor @yongbinfeng @y19y19

PR #3353, opened 2024-07-08
fix: Re-enable Exa.TrkX build without CUDA, make configuration explicit

benjaminhuth:
#2925 broke build for the Exa.TrkX plugin CPU only, this PR fixes this. It changes this from an implicit choice (depending on wether or not we find cuda) to an explicit choice (cmake flag).
Also adds CI job to build CPU only.


.gitlab-ci.yml
```diff
@@ -54,6 +54,46 @@ clang_tidy:
...
     - codereport clang-tidy/clang-tidy.json clang-tidy/html
 
+build_exatrkx_cpu:
+  stage: build
+  image: ghcr.io/acts-project/ubuntu2004_exatrkx:v43
+  tags:
+    - large
+
+  cache:
+    key: ccache-exatrkx-$CI_COMMIT_REF_SLUG
```
paulgessinger, line 64: I gutss this should be separate from the cache key of the other job.

paulgessinger (approved):

paulgessinger (approved):

PR #3357, opened 2024-07-09
feat: Prototracks-to-Tracks converter & refactor

benjaminhuth:
This algorithm is helpful for using CKF performance writer with pure TrackFinding output. Some additional refactoring in the python bindings is included.


andiwand (approved):
LGTM 👍

andiwand:
> This algorithm is helpful for using CKF performance writer with pure TrackFinding output. Some additional refactoring in the python bindings is included.

One thing that just came to my mind - Is this for pure track finding performance measurements? I wonder if this should be rather done with the proto tracks instead of tracks. There is also `TrackFindingPerformanceWriter` which is not really used atm but I wanted to merge it with the `CKFPerformanceWriter` a while ago because I feel like they have the same purpose.

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
```diff
@@ -0,0 +1,93 @@
...
+          mtj->makeTrackState(Acts::TrackStatePropMask::Calibrated, tip);
+      tip = trackStateProxy.index();
+
+      IndexSourceLink sl(Acts::GeometryIdentifier{}, idx);
+
+      calibrator.calibrate({}, {}, Acts::SourceLink{sl}, trackStateProxy);
+      trackStateProxy.typeFlags().set(Acts::TrackStateFlag::MeasurementFlag);
+    }
+
+    auto track = tracks.makeTrack();
```
paulgessinger, line 69: Consider using the `appendTrackState` to avoid having to manually manage the tip index here. Ultimately it does the same thing, but the API is a bit nicer I think.

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    std::size_t tip = Acts::kTrackIndexInvalid;
+    for (auto idx : protoTrack) {
+      auto trackStateProxy =
+          mtj->makeTrackState(Acts::TrackStatePropMask::Calibrated, tip);
+      tip = trackStateProxy.index();
+
+      IndexSourceLink sl(Acts::GeometryIdentifier{}, idx);
+
+      calibrator.calibrate({}, {}, Acts::SourceLink{sl}, trackStateProxy);
```
paulgessinger, line 65: Does this do anything? IIRC there's no predicted state to speak of, right? So the calibration will effectively be identity only. 

OTOH, we should avoid using default constructed contexts, so you might want to at least use the contexts from the event context.

Examples/Python/src/Utilities.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <pybind11/stl.h>
+
+namespace py = pybind11;
+
+using namespace ActsExamples;
+using namespace Acts;
+
+namespace Acts::Python {
+
+void addUtilities(Context& ctx) {
```
paulgessinger, line 29: Maybe something like track conversion utilities to make it a bit clearer what the contents are?

benjaminhuth: It is named after the folder where the algorithms reside in (`Examples/Algorithms/Utilities`), so I would leave it like that...

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
benjaminhuth: Good point with the contexts... Yeah, but without the calibration the matching to particles won't work later right?

andiwand (dismissed):

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
paulgessinger: Not sure I follow how the matching and the calibration are connected.

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
```diff
@@ -0,0 +1,91 @@
...
+    avgSize += protoTrack.size();
+    minSize = std::min(minSize, protoTrack.size());
+    maxSize = std::max(maxSize, protoTrack.size());
+
+    auto track = tracks.makeTrack();
+    for (auto idx : protoTrack) {
+      auto trackStateProxy =
+          track.appendTrackState(Acts::TrackStatePropMask::Calibrated);
+
+      IndexSourceLink sl(Acts::GeometryIdentifier{}, idx);
```
paulgessinger, line 62: Sorry I didn't see this before. Shouldn't the source links already exist? The geometry identifier is also incorrect (0) but should probably be the actual id.

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+      calibrator.calibrate(ctx.geoContext, ctx.calibContext,
+                           Acts::SourceLink{sl}, trackStateProxy);
+      trackStateProxy.typeFlags().set(Acts::TrackStateFlag::MeasurementFlag);
+    }
+
+    track.nMeasurements() = protoTrack.size();
+    track.nHoles() = 0;
+    track.nOutliers() = 0;
+    track.setReferenceSurface(refSurface->getSharedPtr());
```
andiwand, line 72: I think the ref surface should be omitted

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
```diff
@@ -0,0 +1,91 @@
...
+      IndexSourceLink sl(Acts::GeometryIdentifier{}, idx);
+
+      calibrator.calibrate(ctx.geoContext, ctx.calibContext,
+                           Acts::SourceLink{sl}, trackStateProxy);
+      trackStateProxy.typeFlags().set(Acts::TrackStateFlag::MeasurementFlag);
+    }
+
+    track.nMeasurements() = protoTrack.size();
+    track.nHoles() = 0;
+    track.nOutliers() = 0;
```
andiwand, line 71: These could be filled by `calculateTrackQuantities`

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
```diff
@@ -0,0 +1,91 @@
...
+    track.setReferenceSurface(refSurface->getSharedPtr());
+    track.parameters() = Acts::BoundVector::Ones();
+    track.covariance() = Acts::BoundSquareMatrix::Identity();
```
andiwand, line 74: We could also omit these if the ref surface is not set

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
benjaminhuth: A reference surface is required by the `CKFPerformanceWriter`

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
benjaminhuth: see above, is required

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
benjaminhuth: I mean they can, but this is a dummy track, so they can also just be set like this I think....

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
benjaminhuth: Yeah so this was supposed to be the minimal implementation that make the CKF Performance Writer work.
Currently, no downstream process of interest requires the geometry identifier.

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
andiwand: But in that case we should change the `CKFPerformanceWriter` IMO. Setting information we don't have can be a footgun

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
benjaminhuth: This line which is executed during calibration:
`trackState.setUncalibratedSourceLink(sourceLink);`

Of course we could skip calibration and just execute this line...

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
paulgessinger: I think I agree with @andiwand that this is inconsistent though. They look like regular tracks this way but are not. 

The track is equipped to not have a reference surface, and the writer should be able to handle these.

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
paulgessinger: Sure, but if we start writing silently corrupt information, we'll eventually have to figure this out again through debugging when things break. So I think we should not introduce this information in the first place.

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
```diff
@@ -0,0 +1,91 @@
...
+          track.appendTrackState(Acts::TrackStatePropMask::Calibrated);
+
+      IndexSourceLink sl(Acts::GeometryIdentifier{}, idx);
+
+      calibrator.calibrate(ctx.geoContext, ctx.calibContext,
+                           Acts::SourceLink{sl}, trackStateProxy);
+      trackStateProxy.typeFlags().set(Acts::TrackStateFlag::MeasurementFlag);
+    }
+
+    track.nMeasurements() = protoTrack.size();
```
paulgessinger, line 69: This is an implicit conversion from `size_t` to `uint32_t`.

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
```diff
@@ -0,0 +1,91 @@
...
+  float avgSize = 0;
+  std::size_t minSize = std::numeric_limits<std::size_t>::max();
+  std::size_t maxSize = 0;
+
+  for (const auto& protoTrack : prototracks) {
+    if (protoTrack.empty()) {
+      continue;
+    }
+
+    avgSize += protoTrack.size();
```
paulgessinger, line 53: This is an implicit conversion from `size_t` to `float`.

benjaminhuth: Can chagne, but if we care, we should make this a warning / clang tidy thing somewhere.

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
benjaminhuth: yeah i can change but uint64_t to uint32_t but is not really dangerous in this case. If we care, we should make this a warning / clang tidy thing somewhere.

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
paulgessinger: SonarQube reports on this now (see report)

Examples/Algorithms/Utilities/src/PrototracksToTracks.cpp
paulgessinger: SonarQube does produce a warning for this, actually.

benjaminhuth:
@andiwand @paulgessinger I think I have addressed everything, could you have a look again if CI goes through?

andiwand (approved):

andiwand (approved):

PR #3361, opened 2024-07-10
refactor: Model prototracks as tracks in Examples

andiwand:
Remodel prototracks as tracks with our existing track EDM with the benefit of streamlining common I/O and monitoring without conversion.


andiwand:
After discussion with @benjaminhuth we think it makes most sense to go with https://github.com/acts-project/acts/pull/3357 and at the same time remove/minimize the input variation in our algorithms to either tracks, or if there is no use case for tracks, just prototracks but not both. Closing this because it is therefore obsolete

PR #3345, opened 2024-07-03
build: Bump minimum required C++ standard to 20

paulgessinger:
Blocked by:

- #3227
- https://github.com/acts-project/machines/pull/113

PR #3367, opened 2024-07-11
refactor!: use c++20 concept to define insert function for different types of seed collections

CarloVarni:
Taking from @CouthuresJeremy 's PR I'm here proposing a possible alternative that would not even be a breaking change (unless I'm missing something). This will use c++20 concepts to define different versions of an `insert` function that will add a seed to the output collection.

This will work with 'back_insert_operator' (as we currently do) as well as collection that support `push_back` and/or `insert`

blocked by

- https://github.com/acts-project/acts/pull/3345
- https://github.com/acts-project/acts/pull/3377

edit: Now this is breaking since we do not support back_insert_operator option anymore


CarloVarni:
@paulgessinger @andiwand @CouthuresJeremy I have provided support also for the non C++20 code without changes to the API, so this may be already ok to go. 

The seed finder and filter now accept a global reference to the seed container (for the current code that is a `std:: back_insert_iterator<std::vector<Seed>>`) but you can also pass directly `std::vector` or `std::set` or anything else. **Provided there is an adequate `insert` function in `Acts/Utilities/Functions.hpp` that should work.**

Also, the signatures of the `insert` functions should automatically prohibit rvalues for collections such as `std::vector` or `std::set` for the C++20 part.

I'll add some tests to check and verify this. But this is good to be reviewed already, I think.

/cc @pbutti who may be interested in seeding developments

Core/include/Acts/Utilities/Functions.hpp
paulgessinger, line 1: I think these should be located with the seed filter and in a detail namespace.

Core/include/Acts/Utilities/Concepts.hpp
paulgessinger, line 1: This file will go away when we move to C++20 entirely.

CarloVarni: can't we repurpose it to contain all generic concepts for ACTS?

Core/include/Acts/Utilities/Functions.hpp
CarloVarni: My idea was to make them general enough so that they could be used in the rest of the ACTS code

Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -201,7 +199,7 @@ void SeedFinder<external_spacepoint_t, grid_t, platform_t>::createSeedsForGroup(
 
     m_config.seedFilter->filterSeeds_1SpFixed(
         state.spacePointData, state.candidates_collector,
-        seedFilterState.numQualitySeeds, outIt);
+        seedFilterState.numQualitySeeds, std::forward<container_t>(outIt));
```
CarloVarni, line 202: ops ... just realized this is may be a `std::move` performed multiple times on the same object

Core/include/Acts/Utilities/Functions.hpp
paulgessinger: This seems like a niche-need though, so I don't think we should populate the general namespace with these very generic functions.

Core/include/Acts/Utilities/Concepts.hpp
paulgessinger, line 1: In any case, this would have to move anyway. 
OTOH, C++20-by-default is right around the corner, so maybe we just wait until this switching becomes redundant?

Core/include/Acts/Seeding/detail/UtilityFunctions.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+template <typename external_t, typename... args_t>
+concept isCollectionThatSupportsEmplace = std::ranges::range<external_t> &&
+    requires(external_t coll, args_t... vals) {
+  coll.emplace_back(vals...);
+};
+
+// Define some functions
+template <typename value_t>
+void insert(std::output_iterator<value_t> auto& storage, value_t&& value) {
```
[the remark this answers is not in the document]
CarloVarni: done

Core/include/Acts/Seeding/SeedFinderOrthogonal.ipp
```diff
@@ -676,9 +676,10 @@ void SeedFinderOrthogonal<external_spacepoint_t>::processFromMiddleSP(
    */
   if ((!bottom_lh_v.empty() && !top_lh_v.empty()) ||
       (!bottom_hl_v.empty() && !top_hl_v.empty())) {
+    auto backIns = std::back_inserter(out_cont);
     m_config.seedFilter->filterSeeds_1SpFixed(
         spacePointData, candidates_collector, seedFilterState.numQualitySeeds,
-        std::back_inserter(out_cont));
+        backIns);
```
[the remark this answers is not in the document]
CarloVarni: fixed

Core/include/Acts/Seeding/detail/UtilityFunctions.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+template <std::ranges::range storage_t, typename value_t>
+requires(!Acts::detail::isCollectionThatSupportsPushBack<storage_t> &&
+         Acts::detail::isCollectionThatSupportsInsert<
+             storage_t>) void insert(storage_t& storage, value_t&& value) {
+  storage.insert(std::ranges::end(storage), std::forward<value_t>(value));
+}
+
+template <typename... value_t>
+void emplace(
```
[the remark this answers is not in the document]
CarloVarni: removed

paulgessinger (approved):
Thanks! Let's get this in.

acts-project-service:
## :red_circle: Athena integration test results [88c12234fd4faf5cd5d9d8309deda91b493f30e0]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/41264716) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7761936)!

PR #3372, opened 2024-07-12
feat: traccc cuda example

fredevb:
Added an example to run the full traccc chain on the GPU.
Added new cmake option ACTS_BUILD_CUDA which controls whether the CUDA chain example is built.
Updated the python binding for the traccc chain example to include cuda chains (the platform option when adding the algorithm can be used to pick between "host" and "cuda").

Currently not tested as I do not have a GPU with CUDA support.


andiwand:
closing due to massive merge conflicts and inactivity

PR #3387, opened 2024-07-15
refactor!: Clean up `GainMatrixUpdater` a bit

andiwand:
Shuffle / remove comments in `GainMatrixUpdater`.


Core/src/TrackFitting/GainMatrixUpdater.cpp
```diff
@@ -58,10 +58,11 @@ std::tuple<double, std::error_code> GainMatrixUpdater::visitMeasurement(
     ACTS_VERBOSE("Gain Matrix K:\n" << K);
 
     if (K.hasNaN()) {
+      // set to error abort execution
       error = (direction == Direction::Forward)
```
andiwand, line 62: this is actually the only reason we pull in the direction which seems a bit odd to me. maybe we should drop that and generalize the error

paulgessinger (dismissed):

paulgessinger:
I guess this is breaking now @andiwand?

andiwand:
I guess so - I will flag it

Core/src/TrackFitting/GainMatrixUpdater.cpp
```diff
@@ -58,10 +57,9 @@ std::tuple<double, std::error_code> GainMatrixUpdater::visitMeasurement(
     ACTS_VERBOSE("Gain Matrix K:\n" << K);
 
     if (K.hasNaN()) {
-      error = (direction == Direction::Forward)
-                  ? KalmanFitterError::ForwardUpdateFailed
-                  : KalmanFitterError::BackwardUpdateFailed;  // set to error
-      return false;                                           // abort execution
+      // set to error abort execution
+      error = KalmanFitterError::UpdateFailed;
```
paulgessinger, line 61: I guess this was primarily there for diagnostics?

paulgessinger (approved):

Core/src/TrackFitting/GainMatrixUpdater.cpp
paulgessinger: Also (not critical): can we catch this case in the unit test? Sonar reports this as uncovered.

andiwand: Potentially one could count the error cases but that can still be done by the KF itself. If the update is symmetrical I think it is beneficial to reflect that also in the interface.

I added a unit test to check the error case.

Tests/UnitTests/Core/TrackFitting/GainMatrixUpdaterTests.cpp
```diff
@@ -94,4 +95,41 @@ BOOST_AUTO_TEST_CASE(Update) {
...
+  // Check that the state has storage available
+  BOOST_CHECK(ts.hasPredicted());
+  BOOST_CHECK(ts.hasFiltered());
+  BOOST_CHECK(ts.hasCalibrated());
+
+  // Gain matrix update and filtered state
+  BOOST_CHECK(GainMatrixUpdater()
+                  .
+                  operator()<VectorMultiTrajectory>(tgContext, ts)
+                  .error() == KalmanFitterError::UpdateFailed);
```
paulgessinger, line 132: Go home clang-format, you're drunk 🙄

paulgessinger (approved):

Tests/UnitTests/Core/TrackFitting/GainMatrixUpdaterTests.cpp
andiwand: Yeah I couldn't believe it but it really wants it that way

acts-project-service:
## :red_circle: Athena integration test results [218d8020a313335ecb625613c90fb5a70ecc3e7d]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/41276118) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7763533)!

PR #3391, opened 2024-07-16
refactor!: Return all track states from Core CKF

andiwand:
Since the Core CKF is growing into its own components which is then absorbed by a track finding algorithm it also allows us to extend tracks into different regions of the detector. While extending tracks we map want to keep eventual holes/material/outliers states at the end or beginning of the track. For this reason the CKF is modified to return all the states it found.

For convenience of the user I added helper functions which allow to trim tracks. This is then exercised in the Examples track finding.

blocked by
- https://github.com/acts-project/acts/pull/3597
- https://github.com/acts-project/acts/pull/3644
- https://github.com/acts-project/acts/pull/3648


andiwand:
Discussed this briefly with @noemina - Potentially this should rather be a flag and this flag also toggles if we store anything before the first measurement. In this case we can also store holes at the end of the track.

cc @paulgessinger

andiwand:
@timadye it would be great to have your opinion on this one

timadye:
I guess this is fine, but I'm not yet sure I understand when it would be useful. Is this for when you want to run another extension pass starting where you left off (or before where you started)?
Wouldn't it be simpler (for the user) and (potentially a lot) faster to use a flag in the Core CKF to disable the dropping of the holes etc at the ends. Then you wouldn't normally need to trim them later, though having the trim function could still be useful, eg. if an extension failed.

andiwand:
My main motivation is making the Core CKF simpler and dropping some additional logic which should potentially be switchable giving the user control over that. If there would be a performance penalty I would agree with you and make this switchable in the Core CKF but since we collect the states anyways and then trim them in the CKF finalization I would rather push this out since the user may want to have the additional states and if not can explicitly filter them to their need. Potentially people want to only trim holes or holes and outliers and making this configurable is a mess for this component IMO.

timadye:
I guess we should check the performance - or have you already done that? If not, when would be a good time to try?
The reason I thought it might be slower is that holes at the start don't have to be added at all. I'm less sure about the end, but perhaps there is more to copy before it is trimmed.

andiwand:
image

timadye:
With 100 events ttbar_pu200, `full_chain_itk.py`, #3391 TrackFindingAlgorithm is 0.6% *faster* than it's base main (26b059f3a). The efficiency is roughly the same, but there are 14% more fakes. There are also fewer holes and outliers.

Is this expected? I would have expected identical results. Maybe this PR includes the changes from #3597 (which do change things, but I thought not so much), but the `main` reference does not (#3597 was merged after this PR was branched).

nHoles_vs_eta
nOutliers_vs_eta

andiwand:
I am not sure where the additional fakes are coming from. Are these there after the ambiguity resolution step?

It is also quite surprising to me that this change has such effects. Either it is fixing some undiscovered bug from before or introducing a new one. I think we have to pick out some tracks from before and after and look at the hit content to better understand this.

timadye:
Let me rerun the test now that `main` (and #3597) have been merged in. That should separate out the two changes.

timadye:
With 100 events ttbar_pu200, `full_chain_itk.py`, #3391 TrackFindingAlgorithm is now the same speed as it's base `main` (3fb892e77). The efficiency is roughly the same, but there still are 14% more fakes. There are also fewer holes and outliers.

fakerate_vs_eta
nHoles_vs_eta
nOutliers_vs_eta

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -401,6 +402,97 @@ void calculateTrackQuantities(track_proxy_t track) {
...
+
+  for (TrackStateProxy trackState : track.trackStatesReversed()) {
+    TrackStateType typeFlags = trackState.typeFlags();
+    if (trimHoles && typeFlags.test(TrackStateFlag::HoleFlag)) {
+      continue;
+    }
+    if (trimOutliers && typeFlags.test(TrackStateFlag::OutlierFlag)) {
+      continue;
+    }
+    if (trimMaterial && typeFlags.test(TrackStateFlag::MaterialFlag) &&
```
timadye, line 430: This will trim holes, outliers, and pure material. The [old behaviour](https://github.com/acts-project/acts/pull/3391/files#diff-55dd901875f5a87b6a9fee495de245296efbfd45aa9f9483036bcd299a277e09L855) skipped everything except *material* (not measurements, as I would have expected). This is clearly different. I think the new behaviour makes more sense, but this might explain the differences we see.

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -401,6 +402,97 @@ void calculateTrackQuantities(track_proxy_t track) {
...
+  std::optional<TrackStateProxy> back;
+
+  for (TrackStateProxy trackState : track.trackStatesReversed()) {
+    back = trackState;
+
+    TrackStateType typeFlags = trackState.typeFlags();
+    if (trimHoles && typeFlags.test(TrackStateFlag::HoleFlag)) {
+      continue;
+    }
+    if (trimOutliers && typeFlags.test(TrackStateFlag::OutlierFlag)) {
```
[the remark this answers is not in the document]
andiwand: I think outliers will have `MeasurementFlag` set yes

Core/include/Acts/Utilities/TrackHelpers.hpp
andiwand: Good point!

timadye:
would it be useful to check this change in behaviour with a test reproducing the old logic (material or measurement) and see if we still get a change?

Core/include/Acts/Utilities/TrackHelpers.hpp
andiwand: - https://github.com/acts-project/acts/pull/3648

timadye:
I have checked #3644, #3648, and #3391 (this PR) and compared against `main` (65aff940a). I rebased #3644 and merged #3391 with `main` to allow a fair comparison. With 100 ttbar_pu200 events, `full_chain_itk.py` gave the following results:
|  | CPU \[s/ev\] | speedup | efficiency | fake rate |
| --- | --- | --- | --- | --- |
| [`main`](https://github.com/acts-project/acts/commit/65aff940aaf32ee54a954e795b9ce8bf559afa35) | 1.72 | &ndash; | 99.19% | 3.03% |
| #3648 | 1.69 | 1.4% | 99.19% | 3.02% |
| #3644 | 1.72 | \-0.4% | 99.27% | 3.43% |
| #3391 | 1.70 | 0.8% | 99.27% | 3.42% |

where the CPU time is just for `TrackFindingAlgorithm` and is with 8 threads on the same machine. The speedup in each case is relative to `main`.

trackeff_vs_eta
fakerate_vs_eta
nHoles_vs_eta
nOutliers_vs_eta

I think this shows that the tracking performance changes come from #3644, while #3648 improves the speed slightly. #3391 includes the changes from the other two, but doesn't have any additional effect on CPU or performance.

andiwand:
Should we go ahead with #3391 and dump #3644 and #3648 @timadye ? Or do you prefer to merge them sequentially #3644 #3648 #3391?

timadye:
I think it is cleaner to merge #3644 and #3648 (can be done in parallel), then #3391.

andiwand:
@timadye the physmon diff disappeared now

timadye:
With 100 ttbar_pu200 events running `full_chain_itk.py`, #3470 `TrackFindingAlgorithm` is 1% faster than its base `main` (495c60df4). I don't see any appreciable change in the efficiency or fakes.

andiwand:
> With 100 ttbar_pu200 events running `full_chain_itk.py`, #3470 `TrackFindingAlgorithm` is 1% faster than its base `main` ([495c60d](https://github.com/acts-project/acts/commit/495c60df45baf716bbdca583e14d586a43657dc5)). I don't see any appreciable change in the efficiency or fakes.

@timadye is this meant for #3470 or this PR ( #3391 ) ?

timadye:
> @timadye is this meant for #3470 or this PR ( #3391 ) ?

sorry for the typo. I have fixed the original comment. It was #3391.

timadye (approved):

acts-project-service:
## :red_circle: Athena integration test results [ca7021e8be480e42e0764f13df88f022408c53d1]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/44059168) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8218297)!

PR #3403, opened 2024-07-17
build: physmon for traccc chains (host and cuda)

fredevb:
Added physmon for a traccc chain example. This includes both traccc on the host and device (cuda). These can be run in the phymon script using mode traccc_host and traccc_cuda.


andiwand:
closing due to massive merge conflicts and inactivity

PR #3425, opened 2024-07-22
refactor: Lift GenericDetector to a Plugin

benjaminhuth:
The `GenericDetector` has no real dependencies on the Examples types, and thus can be easily compiled without those. By lifting it to a plugin, it could be used in UnitTests for navigation testing e.g.


paulgessinger:
This has picked up a lot of conflicts, unfortunately. How do you want to proceed @benjaminhuth?

benjaminhuth:
Let's close this for now due to low priority.

PR #3426, opened 2024-07-23
chore: Do not copy parameters without ref Surface in TrackProxy

andiwand:
Minor tweak of the `TrackProxy::copyFrom`. `parameters` and `covariance` are only meaningful if a reference surface is given.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,10 +222,17 @@ set(_acts_boost_recommended_version 1.78.0)
 # Include the sources for the external dependencies.
 include(ActsExternSources)
 
+# Controls behavior of DOWNLOAD_EXTRACT_TIMESTAMP
+if(POLICY CMP0135)
+  cmake_policy(SET CMP0135 NEW)
+endif()
+
 # required packages
 if (ACTS_SETUP_BOOST)
   if (ACTS_USE_SYSTEM_BOOST)
-    # NOTE FindBoost.cmake looks for BoostConfig.cmake first, before running it's own logic.
+    if(POLICY CMP0167)
+      cmake_policy(SET CMP0167 NEW)
+    endif()
 
     # Enable both program_options and unit_test_framework to reduce complexity
     # Also Cuda tests seem to use program_options
--- a/cmake/FindFilesystem.cmake
+++ b/cmake/FindFilesystem.cmake
@@ -104,7 +104,6 @@ if(TARGET std::filesystem)
     return()
 endif()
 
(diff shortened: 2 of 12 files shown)


Core/include/Acts/EventData/TrackProxy.hpp
```diff
@@ -687,14 +687,20 @@ class TrackProxy {
 
       // reverse using standard linked list reversal algorithm
       reverseTrackStates();
+    } else {
+      // just copy the track state indices
+      tipIndex() = other.tipIndex();
+      stemIndex() = other.stemIndex();
```
andiwand, line 693: @paulgessinger I put this in here just to have a brief discussion. To me this seemed like the expected behavior if somebody copies a track and just modifies the properties but not the states that also the tip would point to the same as the previous track. What do you think?

Just a slight preference from my side happy to revert

paulgessinger, line 693: These are only valid if the track is copied within the same container though, right? If you do this generically, you might get corrupted tracks.

I see how this can be useful as a "shallow copy", however.

paulgessinger: One problem is that these are only valid if the track is copied within the same container though, right? If you do this generically, you might get corrupted tracks.

I see how this can be useful as a "shallow copy", however.

andiwand: Right I did not think of that case... that "shallow copy" would look like this `trackProxy.copy(withStates=false)`? So we can guarantee it lands in the same container?

That might be useful. For this PR I will just drop the change

paulgessinger: > Right I did not think of that case... that "shallow copy" would look like this `trackProxy.copy(withStates=false)`? So we can guarantee it lands in the same container?

I'd probably literally call it `trackProxy.copyShallow` and the only thing it does is copy without deep copying the track states. If you want a deep copy, you can still call the other function.

> That might be useful. For this PR I will just drop the change

Sounds good!

paulgessinger (approved):

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [87022bd2670deba2a269b19d23e54dd431f511f0]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/41546526) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7806929)!

PR #3429, opened 2024-07-23
feat: detray plugin surface grids

EleniXoch:
Addition of surface grids to existing detray plugin geometry implementation. 

This is still a work in progress :)

### DIFF
--- a/Plugins/Detray/include/Acts/Plugins/Detray/DetrayConversionHelper.hpp
+++ b/Plugins/Detray/include/Acts/Plugins/Detray/DetrayConversionHelper.hpp
@@ -91,9 +91,9 @@ inline static std::tuple<unsigned int, std::vector<ActsScalar>> maskFromBounds(
 ///
 /// @return the acceleration link idnetifier
 template <typename binning_values_t>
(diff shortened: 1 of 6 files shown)


Plugins/Detray/include/Acts/Plugins/Detray/DetrayConversionHelper.hpp
paulgessinger, line 1: I had removed this file in #3509 because it's essentially duplicated in the Json plugin: https://github.com/acts-project/acts/blob/main/Plugins/Json/include/Acts/Plugins/Json/DetrayJsonHelper.hpp

It should not be added without deduplicating this code. I think the cleaned up version that's now in `DetrayJsonHelper.hpp` should still have all the functionality.

EleniXoch: Oh yes thank you

asalzburger:
I guess this can be closed.

asalzburger:
This is superseded, we have introduced the converter of Eleni already.

PR #3435, opened 2024-07-24
feat: speed up detector navigator

asalzburger:
With this PR, the Straight line propagation through the OpenDataDetector:
- remains identical regarding the reached sensitive elements
- gets very close in execution speed 

The latter improvement is on the `DetectorNavigator` to change the `surfaceCandidates` of the state in situ and break after the first successfully hit candidate.

The plot shows a comparison of the distribution of number of sensitives hit and positions between Gen1 and Gen2 geometry, showing that this is now identical:

SteppingComparison

The bottom row shows the non-sensitive steps taken, which are obviously different due to the different geometry portals.

### DIFF
--- a/Core/include/Acts/Navigation/NavigationStateUpdaters.hpp
+++ b/Core/include/Acts/Navigation/NavigationStateUpdaters.hpp
@@ -39,31 +39,25 @@ inline void updateCandidates(const GeometryContext& gctx,
   const auto& position = nState.position;
   const auto& direction = nState.direction;
 
-  NavigationState::SurfaceCandidates nextSurfaceCandidates;
-
-  for (NavigationState::SurfaceCandidate c : nState.surfaceCandidates) {
+  for (NavigationState::SurfaceCandidate& c : nState.surfaceCandidates) {
     // Get the surface representation: either native surface of portal
-    const Surface& sRep =
+    const Surface& surface =
         c.surface != nullptr ? *c.surface : c.portal->surface();
 
     // Only allow overstepping if it's not a portal
     ActsScalar overstepTolerance =
         c.portal != nullptr ? s_onSurfaceTolerance : nState.overstepTolerance;
 
-    // Get the intersection @todo make a templated intersector
-    // TODO surface tolerance
-    auto sIntersection = sRep.intersect(
+    // Check the surface intersection
+    auto sIntersection = surface.intersect(
         gctx, position, direction, c.boundaryTolerance, s_onSurfaceTolerance);
     for (auto& si : sIntersection.split()) {
-      c.objectIntersection = si;
-      if (c.objectIntersection.isValid() &&
-          c.objectIntersection.pathLength() > overstepTolerance) {
-        nextSurfaceCandidates.emplace_back(c);
+      if (si.isValid() && si.pathLength() > overstepTolerance) {
+        c.objectIntersection = si;
+        break;
       }
     }
   }
-
-  nState.surfaceCandidates = std::move(nextSurfaceCandidates);
 }
 
 /// @brief This sets a single object, e.g. single surface or single volume


andiwand (approved):

asalzburger:
Unit tests are failing, I didn't check ... will look at them. But also, I will pull in @dimitra97 for cross-checking as this might also affect the multi-wire navigation.

Core/include/Acts/Navigation/NavigationStateUpdaters.hpp
```diff
@@ -39,31 +39,25 @@ inline void updateCandidates(const GeometryContext& gctx,
   const auto& position = nState.position;
   const auto& direction = nState.direction;
 
-  NavigationState::SurfaceCandidates nextSurfaceCandidates;
-
-  for (NavigationState::SurfaceCandidate c : nState.surfaceCandidates) {
+  for (NavigationState::SurfaceCandidate& c : nState.surfaceCandidates) {
```
andiwand, line 42: I think we need some kind of `intersection candidate` model to correctly handle cylinders

Core/include/Acts/Navigation/NavigationStateUpdaters.hpp
```diff
@@ -39,31 +39,25 @@ inline void updateCandidates(const GeometryContext& gctx,
...
     ActsScalar overstepTolerance =
         c.portal != nullptr ? s_onSurfaceTolerance : nState.overstepTolerance;
 
-    // Get the intersection @todo make a templated intersector
-    // TODO surface tolerance
-    auto sIntersection = sRep.intersect(
+    // Check the surface intersection
+    auto sIntersection = surface.intersect(
         gctx, position, direction, c.boundaryTolerance, s_onSurfaceTolerance);
     for (auto& si : sIntersection.split()) {
```
andiwand, line 54: the splitting should not be necessary if we knew already with intersection index we aim for in case of an `intersection candidate`

Core/include/Acts/Navigation/NavigationStateUpdaters.hpp
```diff
@@ -39,31 +39,25 @@ inline void updateCandidates(const GeometryContext& gctx,
...
-      c.objectIntersection = si;
-      if (c.objectIntersection.isValid() &&
-          c.objectIntersection.pathLength() > overstepTolerance) {
-        nextSurfaceCandidates.emplace_back(c);
+      if (si.isValid() && si.pathLength() > overstepTolerance) {
+        c.objectIntersection = si;
+        break;
```
andiwand, line 57: in principle this does not update all candidates anymore right as the method name suggests?

also do we sort them afterwards again or does that only happen once? because the intersection distances will not be in line anymore

asalzburger: So, I don't think we need the break, actually ... I will do some systematic checks tomorrow.

Core/include/Acts/Navigation/NavigationStateUpdaters.hpp
asalzburger: correct, the only issue would be if a formally non-reachable split intersection becomes reachable right?

I thought about it a lot (too much) yesterday, but probably the solution is the following:

- get the list of surfaces (and keep them separate from the intersection)
- have a buffer of intersection solutions on the candidates and remember the actual buffer size (which can change with re-navigation)

andiwand: Yes I think this is similar to what I do in the try all navigators. I don't really see a way around this having surfaces which can be intersected twice.

andiwand: In the Gen1 navigator I disambiguated that by having a vector of `SurfaceIntersections` as candidates. These can also be used to reintersect.

dimitra97:
> Unit tests are failing, I didn't check ... will look at them. But also, I will pull in @dimitra97 for cross-checking as this might also affect the multi-wire navigation.

i am having a look!!

asalzburger:
This will be superseded.

PR #3438, opened 2024-07-24
fix: Correct handling of outliers and holes for track statistics

andiwand:
I discovered in https://github.com/acts-project/acts/pull/3413 that we do not accumulate outliers and holes correctly in the different accumulators we in multiple places.

Also makes me wonder if an enum would be better in handling these cases compared to an unconstrained bit set.


timadye:
looks good to me.

paulgessinger (approved):

andiwand:
> ## 🔴 Athena integration test results [[ed53797](https://github.com/acts-project/acts/commit/ed5379705803243732383599d66c1f0ef3366711)]
> ### Build job with this PR failed!
> Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/41625319) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7818976)!

re-running here https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/41627033

paulgessinger (approved):

Core/include/Acts/EventData/MultiTrajectoryHelpers.hpp
```diff
@@ -44,21 +45,27 @@ using VolumeTrajectoryStateContainer =
...
   TrajectoryState trajState;
-  multiTraj.visitBackwards(entryIndex, [&](const auto& state) {
+  multiTraj.visitBackwards(tipIndex, [&](const auto& state) {
     // Get the volume Id of this surface
     const auto& geoID = state.referenceSurface().geometryId();
     const auto& volume = geoID.volume();
     const auto& layer = geoID.layer();
     trajState.nStates++;
     auto typeFlags = state.typeFlags();
-    if (typeFlags.test(Acts::TrackStateFlag::MeasurementFlag)) {
```
timadye, line 61: is this reordering of the `if` clauses (here and elsewhere) because we can have a measurement and an outlier (ie. both bits set) for the same track state. Is that expected? I don't think I saw that before, when I printed out the flags.

andiwand: I think that was the intention of having flags there yes. But that freedom can also be a problem as we discover here in the code and from just thinking differently about it.

Generally the rotation of the `if` should provide a more stable solution

timadye (approved):

timadye:
hang on a moment: I am getting a python crash in `full_chain_itk.py` setup for `ActsExamples::AdaptiveMultiVertexFinderAlgorithm::Config`. Let me check that isn't me doing something stupid.

timadye:
> hang on a moment: I am getting a python crash in `full_chain_itk.py` setup for `ActsExamples::AdaptiveMultiVertexFinderAlgorithm::Config`. Let me check that isn't me doing something stupid.

We realised this is from another MR. No need to delay this one.

timadye:
I checked ITk ttbar_pu200 100 events, 8 threads. Compared to `main` this PR has almost identical CPU time, efficiency, and hit/hole/outlier etc counts.

andiwand:
MacOS build is broken. I am trying to fix it

PR #3434, opened 2024-07-24
refactor: Generalize propagation validation in Examples

andiwand:
Some preparation work to monitor the propagation including stepping and navigation tightly with CI

PR #3448, opened 2024-07-26
feat: Bind some Gen1 geometry building functionality to python 

benjaminhuth:
Gen1 Geometry building ist better than its reputation, if you bind it to python. Was used to build ITk from GeoModel.

image

Necessary changes to the GeoModel Plugin will come in a subsequent PR, the script that actually builds the ITk won't live in the ACTS repository.


benjaminhuth:
cc @asalzburger @noemina

Examples/Scripts/Python/propagation.py
andiwand, line 1: should this go in?

Examples/Io/Root/include/ActsExamples/Io/Root/RootPropagationStepsWriter.hpp
andiwand, line 1: note that this will change the hashes

Examples/Scripts/Python/propagation.py
benjaminhuth: In principle yes I would have said, because it makes the thing working... 
I would leave the objstepwriter commented out until it is un-broken
The sterile option could be made configurable.

What is your take on this?

Examples/Io/Root/include/ActsExamples/Io/Root/RootPropagationStepsWriter.hpp
benjaminhuth: Yeah but its convenient to have r in the set, because people who don't know any root can plot stuff then with the clicky-colored-ui then.

paulgessinger: We went back and forth about this and eventually removed it. 

Also, should be a separate PR.

andiwand: we can absorb this here if we think this makes sense https://github.com/acts-project/acts/pull/3434

Examples/Scripts/Python/itk_from_geomodel_gen1.py
```diff
@@ -0,0 +1,442 @@
...
+            for key in self.keys():
+                l += self[key].flatten()
+        else:
+            for key in self.keys():
+                l += self[key]
+
+        return l
+
+
+class ItkBuilder:
```
paulgessinger, line 103: Can't most (all?) of this be done in C++? I'm not sure it's necessarily a good idea to put so much logic and one-off infrastructure into python like this. If you write this in C++, you can just bind the outer interface to python and achieve the same thing with fewer downsides.

Examples/Python/src/GeometryBuildingGen1.cpp
paulgessinger, line 1: I'm kind of hesitant to expose so much of the Core API to python to be honest. The idea was to keep this as restricted as possible. In this case, can't this be done in C++?

Examples/Scripts/Python/propagation.py
paulgessinger: Not sure I follow. What's broken about the obj writer? Changes seem a bit arbitrary to me.

Examples/Python/src/Geometry.cpp
```diff
@@ -146,6 +146,13 @@ void addGeometry(Context& ctx) {
   {
     py::class_<Acts::TrackingGeometry, std::shared_ptr<Acts::TrackingGeometry>>(
         m, "TrackingGeometry")
+        .def(py::init([](const MutableTrackingVolumePtr& v,
+                         // std::shared_ptr<const IMaterialDecorator> m,
+                         const GeometryIdentifierHook& h,
+                         const PythonLogger& logger) {
+          return std::make_shared<Acts::TrackingGeometry>(v, nullptr, h,
+                                                          logger.logger());
+        }))
```
paulgessinger, line 155: Don't use `PythonLogger` as argument here. Also, can't you just bind the constructor without a separate functions? Also also, I don't think we should be creating these objects from python.

Examples/Python/src/Detector.cpp
paulgessinger, line 1: Random change?

Examples/Python/src/Base.cpp
paulgessinger, line 1: Random change?

paulgessinger (commented):
Sorry to push back on this a little bit, but I don't think we should substantially increase the amount of Core API surface that we expose to python, and then implement logic that can be implemented in C++ in python instead. 

The python bindings are supposed to steer the examples, and help with configuration, not contain sophisticated logic and algorithms.

paulgessinger:
Can we hold this until Thursday so we can discuss how to proceed with this?

Examples/Python/src/Geometry.cpp
benjaminhuth: The constructor takes a reference to a logger. That means, the logger must live on the call-site. This is most naturally achieved by our logger that lives in python.
Option B would be to let the callee take a `std::unique_ptr<Logger>`, but that would require technically an API break I think...

Examples/Python/src/Base.cpp
benjaminhuth: Not random, moved to header. See discussion on `PyLogger` usage.

benjaminhuth:
Unfortunately I won't be there on thursday, and also not in the first week of August. But maybe we can resolve this online...

Regarding the question if it at all should be in python and cannot be in C++: I think that adding bindings to a few building structures does not substentially increase the API surface, and is well justified in that case: I don't think its good to have details like regex parsing of ITK database strings, number of layers, radii of volumes etc. in C++ rather then in a python script. I think the chosen approach is actually quite in the spirit of having configuration logic in python, and generic functionality in C++.  
Last but not least I'm not super eager to rewrite the code in C++ now where it is already done in python^^

In todays standup meeting we also clarified, that the build script itself will not become part of the ACTS repository, but rather the ITk repository where also the database is located. Also from this perspective I think a python script is favourable over a C++-compiled version of the geometry building.

benjaminhuth:
For the record: build script has been removed, unrelated/propagation related changes have been reverted

Examples/Python/src/GeometryBuildingGen1.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+  using SurfacePtrVector = std::vector<std::shared_ptr<const Surface>>;
+
+  py::class_<Acts::Layer, std::shared_ptr<Acts::Layer>>(m, "Layer");
+
+  {
+    auto creator =
+        py::class_<Acts::LayerCreator>(m, "LayerCreator")
+            .def(py::init<Acts::LayerCreator::Config>())
+            .def("cylinderLayer",
```
paulgessinger, line 43: Can this be bound directly without the lambda?

Examples/Python/src/GeometryBuildingGen1.cpp
```diff
@@ -0,0 +1,155 @@
...
+                 [](const Acts::LayerCreator &self, const GeometryContext &gctx,
+                    SurfacePtrVector surfaces, std::size_t binsPhi,
+                    std::size_t binsZ) {
+                   return self.cylinderLayer(gctx, surfaces, binsPhi, binsZ);
+                 })
+            .def("discLayer",
```
paulgessinger, line 49: Same here

Examples/Python/src/GeometryBuildingGen1.cpp
```diff
@@ -0,0 +1,155 @@
...
+                       m, "TrackingVolumeArrayCreator")
+                       .def(py::init<Config>());
+
+    py::class_<Config>(creator, "Config").def(py::init<>());
+  }
+
+  py::class_<Acts::VolumeBounds, std::shared_ptr<Acts::VolumeBounds>>(
+      m, "VolumeBounds");
+
+  py::class_<Acts::CylinderVolumeBounds,
```
paulgessinger, line 109: **IF** we're deciding that volume bounds etc will be exposed in Python, would it be a good idea to add all of the other types as well? Or can we leave this on a case by case basis?

Examples/Python/src/GeometryBuildingGen1.cpp
```diff
@@ -0,0 +1,155 @@
...
+             std::shared_ptr<Acts::CylinderVolumeBounds>, Acts::VolumeBounds>(
+      m, "CylinderVolumeBounds")
+      .def(py::init<ActsScalar, ActsScalar, ActsScalar, ActsScalar, ActsScalar,
+                    ActsScalar, ActsScalar>(),
+           "rmin"_a, "rmax"_a, "halfz"_a, "halfphi"_a = M_PI, "avgphi"_a = 0.,
+           "bevelMinZ"_a = 0., "bevelMaxZ"_a = 0.);
+
+  {
+    // NOTE Here we break the interface down from a transform to a zshift, since
+    // this is the only thing used
```
paulgessinger, line 119: Is that necessary? Generally, I find it to be the least surprising if the Python version does not introduce extra logic over C++, whenever that's possible (which it isn't always of course)

Examples/Python/src/GeometryBuildingGen1.cpp
```diff
@@ -0,0 +1,155 @@
...
+                             Acts::Logging::Level level) {
+              return Acts::CylinderVolumeHelper(
+                  cfg, Acts::getDefaultLogger("CylinderVolumeHelper", level));
+            }))
+            .def("createTrackingVolume",
+                 [](const Acts::CylinderVolumeHelper &self,
+                    GeometryContext gctx, const LayerVector &layers,
+                    std::shared_ptr<VolumeBounds> volumeBounds,
+                    ActsScalar zShift, const std::string &name) {
+                   auto trafo = Transform3::Identity();
```
paulgessinger, line 132: If there's not good way to create the transform in Python without adding a bunch of API from Eigen, maybe this could be a separate Python helper, and this function still accepts the transform and doesn't convert internally?

paulgessinger (commented):
@benjaminhuth Sure. 

Mainly I'm concerned with creating volume bounds and tracking volumes directly in Python, and doing any of the sizing calculations in Python. 

Especially the latter clearly deviates from configuration over calculation. If that's outside the repository and hence specific to the ITk test, I'm ok with that. 

If you ask me, I would suggest to find a way to put the sizing logic into C++ with some form of configuration at the boundary and bind that to Python.

benjaminhuth:
I think the main difference here in our approach is that I only aime for "making it work until NextGen Geometry is ready", so the approach was never to port the full flexibility of Gen1 geometry building to python. I don't think we open the box of pandora with that.

I think it would be far from straight forward to do a generic C++ sizing logic for Gen1 geometry (I think that is done in the `GenericDetector` to some degree, but turned out to not be portable enough for this use case). Trying this would end up in an almost as complex system as the blueprint method or so... And I would not invest time in such a thing for an already deprecated geometry model.

In my opinion it is be okay to have incomplete and inconsistent bindings for this for the time being, just make things work in this geometry transistion phase that will be anyways a bit unpleasent...

Examples/Python/src/GeometryBuildingGen1.cpp
benjaminhuth: If I rember correctly I did that because full binding would require the binding of the types of some default arguments, that are not necessary currently...

Examples/Python/src/GeometryBuildingGen1.cpp
benjaminhuth: Same as above, I think case-by-case is enough for the purpose of these bindings...

Examples/Python/src/GeometryBuildingGen1.cpp
benjaminhuth: as above, I would stick to the "make ITk work" type of bindings...

paulgessinger: Sure, but this is just confusing. I don't think we can and should just throw all other considerations over board for "making it work".

paulgessinger:
I see your point. But to some degree the thing is that yes, the sizing logic it possibly non-trivial, I wrote the code to do that in Athena for the Gen1 ACTS geometry. 

> I think it would be far from straight forward to do a generic C++ sizing logic for Gen1 geometry

Correct, the interfaces were meant to be used in experiment specific code to write geometry construction. The dd4hep and TGeo plugins achieve this through configuration, which is necessary because we ship them as plugins. 

GeoModel is a bit different because it's only used by ATLAS, but I don't think we should ship a plugin that's "GeoModel, but only for ITk kind of".

Now, with the Python building code gone from the repository, I don't have fundamental objections against the rest of this PR. 

I would still encourage evaluating the code that we do put in, so that it makes sense, and is not just whatever needed to be added to get something done.

Examples/Scripts/Python/propagation.py
andiwand: I think the problem @benjaminhuth is experiencing is that types in the whiteboard do not match up. I guess it is the typeinfo mechanism we use which is not stable across compilation units?

Examples/Python/src/Geometry.cpp
andiwand: I think what we usually do is pass in a log level and construct the logger in the binding. The logger will not be hold on in C++ so that should be safe

Examples/Scripts/Python/propagation.py
paulgessinger: I think it should be possible to make it stable enough if it isn't. We didn't run into any issues with this anywhere else so far, did we?

benjaminhuth: Hmm in the meanwhile I encountered the issue a second time totally unrelated to that algorithm. So I suspect that it has to do with my specific setup... could be that I did a (minor) compiler version upgrade and through ccache I got a messed up setup there...
But I wouldn't take any action until reproducible see that on my system...

Examples/Python/src/GeometryBuildingGen1.cpp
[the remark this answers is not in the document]
benjaminhuth: If I understood correctly this should also be used to build a geometry in Gen1 and NextGen from single source in order to help validating navigation and such things.

benjaminhuth:
@paulgessinger @andiwand I changed the PR as follows:
- Changes to the GeoModel Plugin are postponed to a subsequent PR
- The `PythonLogger` is not used anymore

Is this good to go in from your perspective?

Examples/Python/src/Geometry.cpp
```diff
@@ -146,6 +146,15 @@ void addGeometry(Context& ctx) {
   {
     py::class_<Acts::TrackingGeometry, std::shared_ptr<Acts::TrackingGeometry>>(
         m, "TrackingGeometry")
+        .def(py::init([](const MutableTrackingVolumePtr& v,
+                         std::shared_ptr<const IMaterialDecorator> m,
```
paulgessinger, line 150: Can you change these to nontrivial variable names?

Examples/Python/src/Geometry.cpp
```diff
@@ -146,6 +146,15 @@ void addGeometry(Context& ctx) {
...
+                         const GeometryIdentifierHook& h,
+                         Acts::Logging::Level level) {
+          auto logger = Acts::getDefaultLogger("TrackingGeometry", level);
+          auto trkGeo = std::make_shared<Acts::TrackingGeometry>(
+              v, m ? &*m : nullptr, h, *logger);
```
paulgessinger, line 155:
```suggestion
              v, m.get(), h, *logger);
```

Examples/Python/src/GeometryBuildingGen1.cpp
paulgessinger: Can you unify this with the `makeCylinderVolume` function that is in `Geometry.cpp`. I guess we don't need both?

benjaminhuth: `makeCylinderVolume` was actually unused, so I removed it. Moved Binding for `CylinderVolumeBounds` to `Geometry.cpp` anyways because I think its better there.

Examples/Python/src/GeometryBuildingGen1.cpp
benjaminhuth: I changed it to take a `Transform3`.

benjaminhuth:
I think I addressed your comments, can you have a look again @paulgessinger ?

Examples/Python/src/Geometry.cpp
```diff
@@ -146,13 +146,13 @@ void addGeometry(Context& ctx) {
...
-                         std::shared_ptr<const IMaterialDecorator> m,
-                         const GeometryIdentifierHook& h,
+        .def(py::init([](const MutableTrackingVolumePtr& volPtr,
+                         std::shared_ptr<const IMaterialDecorator> matDec,
+                         const GeometryIdentifierHook& hook,
                          Acts::Logging::Level level) {
           auto logger = Acts::getDefaultLogger("TrackingGeometry", level);
           auto trkGeo = std::make_shared<Acts::TrackingGeometry>(
-              v, m ? &*m : nullptr, h, *logger);
+              volPtr, matDec ? matDec.get() : nullptr, hook, *logger);
```
paulgessinger, line 155:
```suggestion
              volPtr, matDec.get(), hook, *logger);
```

`shared_ptr<T>::get()` returns `nullptr` if the pointer is empty, no need to check.

paulgessinger (approved):

paulgessinger:
Merging this manually.

acts-project-service:
## :white_check_mark: Athena integration test results [ecad6fd51783e7de306910e2a95eae98e4e0f944]

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633169)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633168)||
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633167)||
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633166)||
|:green_circle: | [test_ActsCheckObjectCountsCachedWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633165)||
|:green_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633164)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633163)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633162)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633161)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633160)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633159)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633158)||
|:green_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633157)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633156)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633155)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633154)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633153)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633152)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633151)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633150)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633149)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633148)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633147)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633146)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633145)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633144)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633143)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633142)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633141)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633140)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633139)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633138)||
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42633137)||

PR #3449, opened 2024-07-27
refactor!: Decouple navigation and stepping

andiwand:
This is a proposal to decouple navigation from stepping by
- changing the navigator interface to only consume **its** state and not the propagation state
- providing position and direction as an input to the navigation
- putting the `Propagator` in charge of communicating the target and estimated distance to the stepper

Originally I wanted the navigator to return the next step size but that clashes with the multi stepper. The current solution allows the stepper to judge on the step size given a surface. This allows the stepper to tweak the step size or to just take what the navigator proposed. In case of the multi stepper we will still just run intersections with the individual components.

The navigator is reduced to providing surface candidates and the surface status will be purely determined by the stepper. The propagator passes the information so the navigator knows when to switch surface / layer / volume.

Ultimately this removes all the templating from the navigator and the interface captures well what the navigator is supposed to do by having dedicated and visible I/O paths.

In case this is where we want to go I would like to do something similar to the steppers.

blocked by
- https://github.com/acts-project/acts/pull/3984
- https://github.com/acts-project/acts/pull/3985
- https://github.com/acts-project/acts/pull/3986

## Release Notes

- **Navigation Improvements**
	- Introduced a new `NavigationTarget` structure to enhance navigation precision.
	- Streamlined navigator initialization and target resolution processes.
	- Added more robust surface and target validation mechanisms.
	- Removed unnecessary member variables from navigators to simplify state management.

- **Propagator Enhancements**
	- Updated propagation logic to improve step size management.
	- Introduced configurable maximum target skipping limit.
	- Enhanced error handling for navigation scenarios.
	- Updated error enumerations to reflect new conditions.

- **Stepper Refinements**
	- Simplified stepper state initialization.
	- Improved step size update and release mechanisms.
	- Added more flexible context and options handling.
	- Removed redundant tolerance parameters from stepper states.

- **Testing and Validation**
	- Updated test cases to reflect new navigation and propagation interfaces.
	- Enhanced logging and error reporting.
	- Improved test coverage for various navigation scenarios.
	- Streamlined test setup and initialization processes.

- **Performance Optimizations**
	- Removed redundant state management variables.
	- Simplified navigation and stepping logic.
	- Improved memory management in propagation states.


andiwand:
The current change avoids double intersection while approaching the target which result in CPU performance improvements. Interestingly this is only pronounced for Fatras and quite suppressed for the CKF

image

image

paulgessinger:
Probably makes my life easier implementing a Gen3 navigator

Core/include/Acts/Propagator/EigenStepper.hpp
```diff
@@ -83,7 +84,7 @@ class EigenStepper {
     explicit State(const GeometryContext& gctx,
                    MagneticFieldProvider::Cache fieldCacheIn,
                    const BoundTrackParameters& par,
-                   double ssize = std::numeric_limits<double>::max())
+                   double ssize = 10 * Acts::UnitConstants::m)
```
paulgessinger, line 87: Could make this a central static default at the `EigenStepper` class level.

Is / should this be synced to the `StepperOptions` max step size?

Core/include/Acts/Propagator/Propagator.ipp
```diff
@@ -27,80 +30,154 @@ template <typename S, typename N>
...
+
+    state.stage = PropagatorStage::postPropagation;
+
+    state.options.actorList.act(state, m_stepper, m_navigator, logger());
+
+    return Result<void>::success();
+  }
+
+  auto getNextTarget = [&]() -> Result<NavigationTarget> {
+    for (unsigned int i = 0; i < state.options.maxTargetSkipping; ++i) {
```
paulgessinger, line 52: I guess this is essentially a safety mechanism?

Core/include/Acts/Propagator/NavigationTarget.hpp
```diff
@@ -0,0 +1,39 @@
...
+namespace Acts {
+
+class Surface;
+
+struct NavigationTarget {
+  const Surface* surface = nullptr;
+  std::uint8_t surfaceIntersectionIndex = 0;
+  BoundaryTolerance boundaryTolerance = BoundaryTolerance::None();
+
+  static NavigationTarget invalid() { return NavigationTarget(); }
```
paulgessinger, line 24:
```suggestion
  static constexpr NavigationTarget invalid() { return NavigationTarget(); }
```
?

Core/include/Acts/Propagator/Propagator.ipp
```diff
@@ -27,80 +30,154 @@ template <typename S, typename N>
...
-      if (state.options.actorList.checkAbort(state, m_stepper, m_navigator,
-                                             logger())) {
-        terminatedNormally = true;
-        break;
+      IntersectionStatus preStepSurfaceStatus = m_stepper.updateSurfaceStatus(
+          state.stepping, *nextTarget.surface,
+          nextTarget.surfaceIntersectionIndex, state.options.direction,
+          nextTarget.boundaryTolerance, state.options.surfaceTolerance,
+          logger());
+      if (preStepSurfaceStatus >= Acts::IntersectionStatus::reachable) {
```
paulgessinger, line 63: The `>=` operator is a bit funky. I think it's sort of clear what the comparison encodes, but can't this be achieved with a `!=` as well?

Core/include/Acts/Propagator/Propagator.ipp
```diff
@@ -27,80 +30,154 @@ template <typename S, typename N>
...
+        return nextTarget;
       }
     }
-  } else {
-    ACTS_VERBOSE("Propagation terminated without going into stepping loop.");
+
+    ACTS_ERROR("getNextTarget failed to find a valid target surface.");
```
paulgessinger, line 68:
```suggestion
    ACTS_ERROR("getNextTarget failed to find a valid target surface after " 
               << state.options.maxTargetSkipping << " attempts.");
```

Core/include/Acts/Propagator/Propagator.ipp
```diff
@@ -27,80 +30,154 @@ template <typename S, typename N>
...
+      // navigator step constraint is not valid anymore
+      m_stepper.releaseStepSize(state.stepping, ConstrainedStep::navigator);
+
+      nextTargetResult = getNextTarget();
+      if (!nextTargetResult.ok()) {
+        return nextTargetResult.error();
+      }
+      nextTarget = *nextTargetResult;
+    }
+  }
```
paulgessinger, line 164: We don't conventionally do this, but this could benefit from a 


```suggestion
  } // end stepping loop
```

Core/include/Acts/Propagator/Propagator.ipp
```diff
@@ -27,80 +30,154 @@ template <typename S, typename N>
...
+
+    if (!nextTarget.isValid()) {
+      // navigator step constraint is not valid anymore
+      m_stepper.releaseStepSize(state.stepping, ConstrainedStep::navigator);
+
+      nextTargetResult = getNextTarget();
+      if (!nextTargetResult.ok()) {
+        return nextTargetResult.error();
+      }
+      nextTarget = *nextTargetResult;
```
paulgessinger, line 162: If this is invalid, what's the path we take? I assume this eventually leads to the normal termination of the propagation / navigation but it's not clear to me how that happens.

Core/include/Acts/Propagator/PropagatorOptions.hpp
```diff
@@ -30,6 +30,9 @@ struct PurePropagatorPlainOptions {
   /// Maximum number of steps for one propagate call
   unsigned int maxSteps = 1000;
 
+  /// Maximum number of next target calls for one step
```
paulgessinger, line 33: Maybe we could elaborate more what this protects against, and why this might happen.

Core/include/Acts/Propagator/StraightLineStepper.hpp
```diff
@@ -274,7 +274,7 @@ class StraightLineStepper {
   /// @param stype [in] The step size type to be set
   /// @param release [in] Do we release the step size?
   void updateStepSize(State& state, double stepSize,
-                      ConstrainedStep::Type stype = ConstrainedStep::actor,
+                      ConstrainedStep::Type stype = ConstrainedStep::navigator,
```
paulgessinger, line 277: Does this even need a default?

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -682,15 +682,21 @@ class CombinatorialKalmanFilter {
...
       // to consider the material effects here
       materialInteractor(navigator.currentSurface(state.navigation), state,
                          stepper, navigator, MaterialUpdateStage::PostUpdate);
 
+      // Set path limit based on loop protection
       detail::setupLoopProtection(state, stepper, result.pathLimitReached, true,
                                   logger());
+
+      // Set path limit based on target surface
+      targetReached.checkAbort(state, stepper, navigator, logger());
```
paulgessinger, line 699: Is this new behavior or was this done some other way before?

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -317,13 +317,20 @@ struct GsfActor {
       applyMultipleScattering(state, stepper, navigator,
                               MaterialUpdateStage::PostUpdate);
     }
+  }
 
-    // Break the navigation if we found all measurements
+  template <typename propagator_state_t, typename stepper_t,
+            typename navigator_t>
+  bool checkAbort(propagator_state_t& /*state*/, const stepper_t& /*stepper*/,
```
paulgessinger, line 324: This contracts the main actor with the separate aborter, which is possible after you merged the two lists, right?

Tests/IntegrationTests/NavigatorConsistency.cpp
```diff
@@ -300,45 +300,46 @@ void runConsistencyTest(const propagator_probe_t& propProbe,
                                 refSurfaces.begin(), refSurfaces.end());
 }
 
+Acts::Logging::Level logLevel = Acts::Logging::INFO;
```
paulgessinger, line 303: This is not used everywhere. I'd probably go with a global `Logger` instance and clone it as needed.

Tests/UnitTests/Core/Propagator/NavigatorTests.cpp
```diff
@@ -58,212 +40,17 @@ namespace Acts::Test {
...
-void step(stepper_state_t& sstate) {
-  // update the cache position
-  sstate.pos4[Acts::ePos0] += sstate.stepSize.value() * sstate.dir[Acts::eMom0];
-  sstate.pos4[Acts::ePos1] += sstate.stepSize.value() * sstate.dir[Acts::eMom1];
-  sstate.pos4[Acts::ePos2] += sstate.stepSize.value() * sstate.dir[Acts::eMom2];
-  // create navigation parameters
-  return;
+void step(Vector3& pos, const Vector3& dir, double stepSize) {
+  pos += stepSize * dir;
+}
```
paulgessinger, line 45: 🤣

paulgessinger (commented):
- The interaction between this change and the constrained step changes.

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -248,937 +259,483 @@ class Navigator {
...
-  void initialize(propagator_state_t& state, const stepper_t& stepper) const {
-    // Call the navigation helper prior to actual navigation
+  /// @param state The navigation state
+  /// @param position The start position
+  /// @param direction The start direction
+  /// @param propagationDirection The propagation direction
+  void initialize(State& state, const Vector3& position,
+                  const Vector3& direction,
+                  Direction propagationDirection) const {
+    (void)propagationDirection;
```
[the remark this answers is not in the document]
andiwand: I need this parameter in the direct navigator to know in which direction the surface sequence has to be processed. In any other navigator this is not used. Not great but I didn't see any alternative

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -248,937 +259,483 @@ class Navigator {
...
+
+      // current volume and layer search through global search
+      state.startVolume = m_cfg.trackingGeometry->lowestTrackingVolume(
+          state.options.geoContext, position);
+
+      if (state.startVolume != nullptr) {
+        state.startLayer = state.startVolume->associatedLayer(
+            state.options.geoContext, position);
       } else {
         ACTS_VERBOSE(volInfo(state)
```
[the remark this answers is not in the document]
andiwand: Good point. I guess this only happens if the tracking geometry is broken or somebody starts the propagation outside. The numerical integration would still be valid as long as there is a magnetic field but we will never be able to enter the tracking geometry. So I guess it is an error? Otherwise we silently fall back to void navigation.

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -93,12 +91,13 @@ class Navigator {
 
   /// The navigation stage
   enum struct Stage : int {
-    undefined = 0,
     surfaceTarget = 1,
     layerTarget = 2,
-    boundaryTarget = 3
+    boundaryTarget = 3,
+    noTarget = 4,
```
andiwand, line 97: `initial`

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -248,937 +259,483 @@ class Navigator {
...
-  void postStep(propagator_state_t& state, const stepper_t& stepper) const {
-    // Check if the navigator is inactive
-    if (inactive(state, stepper)) {
-      return;
-    }
-
-    // Set the navigation stage
-    state.navigation.navigationStage = Stage::undefined;
+  /// @return True if the target is valid
+  bool checkTargetValid(const State& state, const Vector3& position,
```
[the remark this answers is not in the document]
andiwand: naming: `haveValidTarget`?

Core/include/Acts/Propagator/NavigationTarget.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+  NavigationTarget(const Surface& surface_,
+                   std::uint8_t surfaceIntersectionIndex_,
+                   BoundaryTolerance boundaryTolerance_)
+      : surface(&surface_),
+        surfaceIntersectionIndex(surfaceIntersectionIndex_),
+        boundaryTolerance(std::move(boundaryTolerance_)) {}
+
+  bool isValid() const { return surface != nullptr; }
```
andiwand, line 33:
```suggestion
  bool isNone() const { return surface != nullptr; }
```

Core/include/Acts/Propagator/NavigationTarget.hpp
andiwand: call it `none`

Core/include/Acts/Propagator/NavigationTarget.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include "Acts/Surfaces/BoundaryTolerance.hpp"
+
+#include <utility>
+
+namespace Acts {
+
+class Surface;
+
+struct NavigationTarget {
```
andiwand, line 19: doc + mention this is basically also a `std::optional`

paulgessinger:
After discussion with @andiwand:
- Change the `-1` sentinel `int` to an `std::optional<std::size_t>`
- `Stage::noTarget` -> `Stage::initial`
- Disambiguate NavigationTarget::invalid and checkTargetValid: valid does not mean the same thing currently.

Tests/IntegrationTests/NavigatorConsistency.cpp
andiwand: It is used a bit further down
```c++
  ACTS_LOCAL_LOGGER(Acts::getDefaultLogger("NavigatorTest", logLevel))
```

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
andiwand: Yes exactly there is not reason to separate them anymore. I did this on the side after realizing the abort condition was communicated through the navigator state which is really not ideal. This could be factored out of this PR in principle

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
andiwand: This was missing before and I put it in because I thought it became a problem. Turned out it was just the invalidation of the current target which was the problem. I would still like to get this is but we could factor this out

Core/include/Acts/Propagator/StraightLineStepper.hpp
andiwand: Good point I think we should just use the param without default. I will also translate this to the other overload

Core/include/Acts/Propagator/Propagator.ipp
andiwand: I agree I am also not a big fan of this. It relies on the fact that we order `missed < reachable < onSurface`. Here it is catching `reachable` and `onSurface`. I will replace it by two `==`

Core/include/Acts/Propagator/Propagator.ipp
andiwand: Yes this is very similar to the maximum number of steps. A misconfigured or bugged navigator might never provide a reachable target and we would be stuck in a loop forever.

Core/include/Acts/Propagator/EigenStepper.hpp
andiwand: I think you are right, this can be replaced by the options. Let me try that

Core/include/Acts/Propagator/Navigator.hpp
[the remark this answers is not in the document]
andiwand: This is a callback to the navigator to check if the current target is still valid if it was not reached by the first step. For potential future implementations this can be used invalidate the state and to cause a reinitialization.

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -248,937 +259,483 @@ class Navigator {
...
+        ++state.navSurfaceIndex;
+        if (!state.endOfSurfaces()) {
+          ACTS_VERBOSE(volInfo(state) << "Target set to next surface.");
+          return NavigationTarget(*state.navSurface().object(),
+                                  state.navSurface().index(),
+                                  BoundaryTolerance::None());
+        } else {
+          // This was the last surface, switch to layers
+          ACTS_VERBOSE(volInfo(state) << "Target layers.");
+          state.navigationStage = Stage::layerTarget;
```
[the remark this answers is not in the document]
andiwand: Yes I adopted all the navigators to really only resolve targets when it is necessary. So `nextTarget` will have the actual performance hit and the other functions should be very lightweight.

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -130,71 +142,73 @@ class Navigator {
...
+      return navBoundaries.at(navBoundaryIndex);
+    }
 
-    /// Navigation state: the world volume
-    const TrackingVolume* worldVolume = nullptr;
+    bool endOfSurfaces() const {
+      return navSurfaceIndex >= static_cast<int>(navSurfaces.size());
+    }
+    bool endOfLayers() const {
+      return navLayerIndex >= static_cast<int>(navLayers.size());
```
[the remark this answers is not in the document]
andiwand: Yes exactly

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -248,937 +259,483 @@ class Navigator {
...
 
-    auto tryTargetNextSurface = [&]() {
+    auto tryEstimateNextTarget = [&]() -> NavigationTarget {
       // Try targeting the surfaces - then layers - then boundaries
 
-      if (state.navigation.navigationStage <= Stage::surfaceTarget &&
-          targetSurfaces(state, stepper)) {
-        ACTS_VERBOSE(volInfo(state) << "Target set to next surface.");
-        return true;
+      if (state.navigationStage == Stage::noTarget) {
```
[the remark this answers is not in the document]
andiwand: renamed to `initial` to make more sense

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -248,937 +259,483 @@ class Navigator {
...
   ///
-  /// @param [in,out] state is the mutable propagator state object
-  /// @param [in] stepper Stepper in use
-  template <typename propagator_state_t, typename stepper_t>
-  void preStep(propagator_state_t& state, const stepper_t& stepper) const {
-    // Check if the navigator is inactive
-    if (inactive(state, stepper)) {
-      return;
+  /// @return The next target surface
+  NavigationTarget estimateNextTarget(State& state, const Vector3& position,
```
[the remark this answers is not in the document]
andiwand: I believe my thought here was that we actually do not know the real next target so we "estimate" it. But maybe this is something that is clear with the knowledge of a navigator does anyways. I will rename it to `nextTarget`

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
paulgessinger: Your call.

Core/include/Acts/Propagator/Propagator.ipp
andiwand: Discussed offline: we tried to improve the flow a bit by disambiguating `invalid` and `none` target

andiwand:
The stepper options changes escalated a bit and I failed to put this in a different commit. The changes could be factorized into the following
- GSF actor / aborter
- Stepper update signature
- Stepper option handling
- CKF / KF path length aborter
- Constrained step
- Actual navigator refactor

andiwand:
Add navigator concept

paulgessinger (dismissed):
:+1: 

Output changes are probablly expected, but should be sanity checkeed.

EDIT: I don't see anything that's not consistent with statistics.

andiwand:
With https://github.com/acts-project/acts/pull/3449/commits/cf2b28aa8f0cec124833d3b2184cca6fe367fd22 I am trying to work around an FPE that popped up in the GSF. What do you think about that handling @benjaminhuth @paulgessinger ?

paulgessinger:
Let's go 🚀

paulgessinger (approved):

PR #3450, opened 2024-07-28
feat: Add statistics to propagation

andiwand:
This is an attempt to centralize statistics about the propagation which includes stepping and navigation. The intention is to monitor these quantities in the CI and to use them as comparison for different stepper / navigator implementations.


paulgessinger (commented):
Can't the stepping actor do this?

andiwand:
> Can't the stepping actor do this?

Doing this with an actor would be nice because we could also easily switch it on and off then. But I think we generally depend on internal information for the stepping / navigation which would require to wire specific flags and properties out. So accumulating the stats directly in the stepper / navigator seems easier even though it is not switchable anymore.

asalzburger:
> > Can't the stepping actor do this?
> 
> Doing this with an actor would be nice because we could also easily switch it on and off then. But I think we generally depend on internal information for the stepping / navigation which would require to wire specific flags and properties out. So accumulating the stats directly in the stepper / navigator seems easier even though it is not switchable anymore.

Exactly, I had a test implementation of this, where I have an Actor that does it - but this again bundles stepper + navigator ... I Haden't found a better solution so far.

andiwand:
@asalzburger @paulgessinger finally made some progress on this one

Core/include/Acts/Propagator/EigenStepper.ipp
```diff
@@ -286,6 +291,18 @@ Acts::Result<double> Acts::EigenStepper<E>::step(
     }
   }
 
+  state.stepping.pathAccumulated += h;
+  ++state.stepping.nSteps;
+  state.stepping.nStepTrials += nStepTrials;
+
+  ++state.stepping.statistics.nSuccessfulSteps;
+  if (state.options.direction !=
+      Direction::fromScalarZeroAsPositive(initialH)) {
```
paulgessinger, line 300: This is checking if we're going opposite of the original stepping direction? Is this the most robust way to determine this? Maybe it doesn't matter for the statistic even if it isn't.

Core/include/Acts/Propagator/NavigatorStatistics.hpp
```diff
@@ -0,0 +1,21 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+#include <cstddef>
+
+namespace Acts {
+
+struct NavigatorStatistics {
+  std::size_t nRenavigations = 0;
```
paulgessinger, line 16: Should we explain what this counts in a comment? If yes, this applies also to the other statistics structs.

Core/include/Acts/Propagator/EigenStepper.ipp
```diff
@@ -286,6 +291,18 @@ Acts::Result<double> Acts::EigenStepper<E>::step(
     }
   }
 
+  state.stepping.pathAccumulated += h;
```
paulgessinger, line 294: This confused me a bit. This is not for statistics strictly speaking. Why do we do this before the Cov transport now? In general with the statistics, do we want to include the numbers even if the cov transport branch can terminate? Just wondering what the most unsurprising location for the counting is.

Core/include/Acts/Propagator/SympyStepper.hpp
```diff
@@ -132,15 +133,8 @@ class SympyStepper {
...
 
-    /// @brief Storage of magnetic field and the sub steps during a RKN4 step
-    struct {
-      /// Magnetic field evaulations
-      Vector3 B_first, B_middle, B_last;
-      /// k_i of the RKN4 algorithm
-      Vector3 k1, k2, k3, k4;
-      /// k_i elements of the momenta
-      std::array<double, 4> kQoP{};
-    } stepData;
```
paulgessinger, line 143: Was this unused?

Core/src/Propagator/SympyStepper.cpp
```diff
@@ -164,6 +165,8 @@ Result<double> SympyStepper::stepImpl(
       break;
     }
 
+    ++state.statistics.nFailedSteps;
```
paulgessinger, line 168: Is this necessarily an error here? 
Ah, failed probably refers to the number of trials until the accurracy target was reached. 

Maybe the name could be more explicit about that?

Core/include/Acts/Propagator/SympyStepper.hpp
andiwand: Yes, I just stumbled across this while adding the statistics. The sympy stepper does not rely on these state variables.

Core/include/Acts/Propagator/EigenStepper.ipp
andiwand: I believe this is the only robust way to do this. Do you have an alternative way in mind?

`state.options.direction` represents the direction the propagator was configured to use. The sign of `initialH` will represent the actual direction we will go.

Separation wise this is not great. If we want to remove the propagator state from the stepper we need another mechanism but I would push this to the actual refactor.

Core/include/Acts/Propagator/EigenStepper.ipp
andiwand: True I can move that back

Core/src/Propagator/SympyStepper.cpp
andiwand: maybe `rejected`?

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [b30f8dfe86e2ed940fac02bbe54efdfe2cddaac0]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/46406915) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8575496)!

PR #3451, opened 2024-07-29
refactor: Disable copy&move for `TrackStateType`

andiwand:
This makes the behavior explicit for copy&move for `TrackStateType`.

I pulled this out from https://github.com/acts-project/acts/pull/3413 as I saw sonarcloud complaining about this.


andiwand:
cc @timadye

Core/include/Acts/EventData/TrackStateType.hpp
```diff
@@ -41,7 +41,20 @@ class TrackStateType {
       std::numeric_limits<std::make_unsigned<raw_type>::type>::digits;
   /// Constructor from a reference to the underlying value container
   /// @param raw the value container
-  TrackStateType(raw_type& raw) : m_raw{&raw} {}
+  TrackStateType(raw_type& raw) : m_raw{&raw} { assert(m_raw != nullptr); }
```
paulgessinger, line 44: Isn't this redundant? The reference can't be to nullptr.

Core/include/Acts/EventData/TrackStateType.hpp
```diff
@@ -41,7 +41,20 @@ class TrackStateType {
...
+  TrackStateType(const TrackStateType& other) : m_raw{other.m_raw} {
+    assert(m_raw != nullptr);
+  }
+
+  /// Move constructor
+  /// @param other the other object to move from
+  TrackStateType(TrackStateType&& other) : m_raw{other.m_raw} {
+    assert(m_raw != nullptr);
+    other.m_raw = nullptr;
+  }
```
paulgessinger, line 57: What's the motivation for this? Is this not a bit confusing?

paulgessinger (commented):
I'm not sure I understand the motivation for the change. Does this prevent undesirable behavior that I'm missing? I would argue it's safer to keep disallowing copying and try to avoid having these types as lvalues at all.

andiwand:
> I'm not sure I understand the motivation for the change. Does this prevent undesirable behavior that I'm missing? I would argue it's safer to keep disallowing copying and try to avoid having these types as lvalues at all.

I think right now we end up getting a default copy constructor. This popped out in https://github.com/acts-project/acts/pull/3413 because I did `ACTS_LOG(trackStateFlags)` which seems to do a copy somehow.

I think we should either disallow copying by `= delete` but in that case we should also disallow conversion from non-const to const because this is a copy. Otherwise I would allow copy and add all the copy constructors.

@paulgessinger

Core/include/Acts/EventData/TrackStateType.hpp
andiwand: Hm can't you just do `TrackStateType(*nullptr)` and end up with `m_raw = nullptr`? Because in principle the `nullptr` will not be dereferenced but just captured by reference.

timadye: or even use the default constructor

andiwand: I think the default constructor is implicitly deleted because there is one with an argument but one could also explicitly delete it which is best practice in C++ I believe https://en.cppreference.com/w/cpp/language/rule_of_three

paulgessinger:
I see. I'd probably be in favor of deleting the copy.

Core/include/Acts/EventData/TrackStateType.hpp
```diff
@@ -39,27 +39,31 @@ class TrackStateType {
...
   static constexpr std::size_t kRawBits =
       std::numeric_limits<std::make_unsigned<raw_type>::type>::digits;
+
+  // Delete default constructor
+  TrackStateType() = delete;
+
   /// Constructor from a reference to the underlying value container
   /// @param raw the value container
-  TrackStateType(raw_type& raw) : m_raw{&raw} {}
+  TrackStateType(raw_type& raw) : m_raw{&raw} { assert(m_raw != nullptr); }
```
paulgessinger, line 48: I still think this is unconventional, since using a reference puts this in the caller explicitly, but it doesn't hurt either.

paulgessinger (approved):

Core/include/Acts/EventData/TrackStateType.hpp
andiwand: I saw that we put asserts on this pointer in all the member functions and thought I could just lift this to the constructor so it is guaranteed once

paulgessinger: Again, no strong opinion, but they're not quite the same imo. 

The assert in the constructor is shielding against an invalid argument, which in principle is avoided by the reference. The assert in the methods is shielding against the member pointer becoming null for some reason. 

It's a bit academic here, so anything you choose is fine.

timadye (approved):

acts-project-service:
## :red_circle: Athena integration test results [1886a94cebc1a6b9d66eccf51e52918ecfd6fde4]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/41803389) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7848107)!

PR #3453, opened 2024-07-29
refactor: Rework projector

andiwand:
Reworks the projector storage in the track EDM and the operations on vectors matrices. The bitset in the track EDM is replaced by a simple sequence of `std::uint8_t` which provide the mapping `measurement index -> bound index`. A helper is provided, `SubspaceHelper`, which collects convenience functions that convert to the legacy format and to matrix form and also do operations on vectors and matrices e.g. `projectVector` and `projectMatrix`.

The benefit of this change is that the mapping and storage of the mapping becomes easier to grasp and also computationally less expensive. Instead of relying on eigen matrix operations we map directly to the desired object.

A similar change in Athena improved the overall track finding CPU time by 30%.

In favor of `SubspaceHelper` I removed `FixedSizeSubspace` and `VariableSizeSubspace` and made the necessary changes in the Examples.

In principle these changes should be non-breaking. The switch in API should be scheduled for v37.


andiwand:
cc @paulgessinger @goetzgaycken @noemina

Core/include/Acts/EventData/SubspaceHelpers.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+namespace Acts {
+
+template <std::size_t kFullSize>
+class SubspaceHelper {
+ public:
+  explicit SubspaceHelper(std::span<const std::uint8_t> indices)
+      : SubspaceHelper(indices, indices.size()) {}
+
+  SubspaceHelper(std::span<const std::uint8_t> indices, std::size_t size)
```
paulgessinger, line 28: Is this so you can clip the input span?

Core/include/Acts/Geometry/ConeVolumeBounds.hpp
```diff
@@ -34,7 +35,7 @@ class PlanarBounds;
 class ConeVolumeBounds : public VolumeBounds {
  public:
   /// @enum BoundValues for readability
-  enum BoundValues : unsigned int {
+  enum BoundValues : std::uint8_t {
```
paulgessinger, line 38: Is this connected to the projector change? I believe we don't store this often so the memory size shouldn't really matter should it?

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
```diff
@@ -146,22 +148,21 @@ class FixedSizeMeasurement {
   /// subspace are used for the computation.
   ParametersVector residuals(const FullParametersVector& reference) const {
     ParametersVector res = ParametersVector::Zero();
-    Acts::detail::calculateResiduals(static_cast<indices_t>(kSize),
-                                     m_subspace.indices(), reference, m_params,
-                                     res);
+    Acts::detail::calculateResiduals(static_cast<indices_t>(kSize), m_indices,
```
paulgessinger, line 151: Maybe this should be something like `subspaceIndices`?

paulgessinger (commented):
Seems like a good idea overall if this is cheaper to compute. 

The other enums changing underlying type is unrelated, right?

Core/include/Acts/EventData/SubspaceHelpers.hpp
andiwand: Yeah exactly - maybe I change this again now that I see it. I think it is okay to ask the user to provide the correct size.

Core/include/Acts/Geometry/ConeVolumeBounds.hpp
andiwand: I put these mostly as a reminder. I don't intent to change those in this PR because that might be breaking.

Do you think these should generally change to be lower size? As you say this will not really matter but I feel like if the box can be smaller it should be smaller.

andiwand:
This seems to be ~17% faster out of the box

image

There might be a bit more potential because I did not change the calibrator yet

Core/include/Acts/EventData/TrackStateProxy.hpp
```diff
@@ -656,9 +689,36 @@ class TrackStateProxy {
...
+    return component<ProjectorMapping, hashString("projector")>();
+  }
+
+  void setProjector(const ProjectorMapping& proj)
+    requires(!ReadOnly)
+  {
+    component<ProjectorMapping, hashString("projector")>() = proj;
+  }
+
+  void setProjector(std::span<const BoundIndices> proj)
```
andiwand, line 712: I am generally not so happy with this interface. It does not convert from a different int type even if I use a template. Maybe we should template on the full type and expect a range like type? @paulgessinger

Core/include/Acts/EventData/TrackStateProxy.hpp
```diff
@@ -656,9 +664,37 @@ class TrackStateProxy {
...
+        {mapping.begin(), mapping.begin() + calibratedSize()});
+  }
+
+  void setProjector(const ProjectorMapping& proj) requires(!ReadOnly) {
+    component<ProjectorMapping, hashString("projector")>() = proj;
+  }
+
+  void setProjector(std::span<const BoundIndices> proj) requires(!ReadOnly) {
+    assert(has<hashString("projector")>());
+    assert(proj.size() == calibratedSize() && "Invalid projector size");
```
andiwand, line 691: I noticed that I just coupled `projector` and `calibrated` properties. Do you think we should generally do this @paulgessinger ? Or is there a case where a projector might exist but a measurement not?

Core/include/Acts/EventData/SubspaceHelpers.hpp
goetzgaycken: Maybe, I have a biased view, but I would say that it would be better if it was fixed sized i.e. a 
````
template <size_t N> ...(...array<uint8_t, N> ...)
```
that should allow for better optimizations. There should only be a couple of cases likely (N=1,2,3), and likely optimizations are better if these cases are handled explicitly. One could add the alternative to use a span, but there should be an option for a fixed size container.

Core/include/Acts/Geometry/ConeVolumeBounds.hpp
goetzgaycken: In some cases uint8_t handling is inefficient, but I have no idea whether the type of the enum has some actual consequences other than casts.

andiwand: if it would be `uint8_t` we could pull the type through into the `std::array` later but right now it is too big. with operations I expect `uint8_t` to be a bit slower but if it just sits there I would expect similar performance

Core/include/Acts/EventData/SubspaceHelpers.hpp
andiwand: that might be a good option yes. I think the projector is only set in paths where we know the measurement dimension

Core/include/Acts/EventData/TrackStateProxy.hpp
paulgessinger: What do you want to be able to construct it from?

Core/include/Acts/EventData/TrackStateProxy.hpp
paulgessinger: Good question. I guess the main concern would be that it's not obvious that you need to set them in a specific order or risk running into the assertion.

paulgessinger:
17% is wild. I guess it's because in the ODD chain we still run the projection quite a lot for calibrating every single measurement.

Core/include/Acts/EventData/TrackStateProxy.hpp
andiwand: I thought it would be nice to take any kind of integer sequence with runtime size. But maybe what @goetzgaycken suggested makes more sense https://github.com/acts-project/acts/pull/3453#discussion_r1695672293

So strictly taking `template<int N> setProjector(std::array<BoundIndices, N>)` or `std::uint8_t` and then storing it in the maximally sized `std::array` in the backend.

Core/include/Acts/EventData/SubspaceHelpers.hpp
goetzgaycken: or fixed size span

paulgessinger:
I'll have another close look at this tomorrow

andiwand:
It builds and passes the tests now @goetzgaycken @paulgessinger if you want to have another look.

One big open question is how to model the API intermediately before we make the breaking change in 37. Generally I would like to drop all the projector functions and types from the `TrackStateProxy`. The user can interact with the `SubspaceHelper` to get those or to project directly from an eigen type. Intermediately I would keep them an potentially flag them as deprecated. I guess what we need to add minimally
- setter for the subspace array
- getter might not be necessary?
- getter for the subspace helper

The other open question is if we want to couple the subspace directly to the calibrated size. I think they always go along?

goetzgaycken:
> One big open question is how to model the API intermediately before we make the breaking change in 37. Generally I would like to drop all the projector functions and types from the `TrackStateProxy`.

If You do not want to break the api, the original projector bitset etc. interface should stay in place and create the bitset and projector matrix from the index map, unless You know that these functions are not used outside the acts repository. 

The component "projector" might cause problems. If this component is not created and if the projector map is not created from this component, persistification and alternative back-ends might get broken.

> The user can interact with the `SubspaceHelper` to get those or to project directly from an eigen type. Intermediately I would keep them an potentially flag them as deprecated. I guess what we need to add minimally
> 
>     * setter for the subspace array
> 
>     * getter might not be necessary?
> 
>     * getter for the subspace helper
I think there should be a setter getter pair which takes, returns the same type. If the setter does not take a subspace helper a getter which returns a subspace helper would be optional in my opinion (not the other way round). I assume that the subspace helper can be constructed from the type returned by the getter and the operation is cheap (there would be the small inconvenience to spell out the subspace helper type without such a getter, Personally, I would favour a minimal api).

> The other open question is if we want to couple the subspace directly to the calibrated size. I think they always go along?
The projector projects into the measurement space, so the dimensions of the calibrated measurement and the projector map should be identical.

goetzgaycken (commented):
My main comment would be that be both projector maps  and calibrated sub space helper methods in  the interfaces, either one or the other. Then it would be good if recognizable types are used   at all places where such "maps" are passed  (which could be the subspace helper directly maybe with a different name other than Helper, or just aliases for arrays and or spans).

There are a few places where extra asserts would be good.

I would not insist on any changes, since I suppose many things will be reworked a bit if the original projector handling is removed.

Core/include/Acts/EventData/TrackStateProxy.hpp
```diff
@@ -667,9 +676,41 @@ class TrackStateProxy {
...
+    setFullProjectorMapping(fullMapping);
+  }
+
+  FullProjectorMapping fullProjectorMapping() const {
+    assert(has<hashString("projector")>());
+    return component<FullProjectorMapping, hashString("projector")>();
+  }
+
+  template <std::size_t measdim>
+  ProjectorMapping<measdim> projectorMapping() const {
```
paulgessinger, line 694: I guess eventually we'd want to cycle these back around to just be `projector`? Otherwise we could call them `measurementMappingFunction` or `calibratedMappingFunction` since that's what they're called in the KF literature.

Core/include/Acts/EventData/TrackStateProxy.hpp
```diff
@@ -667,9 +676,41 @@ class TrackStateProxy {
...
+  /// @deprecated Use setProjector(span) instead
+  //[[deprecated("use setProjector(span) instead")]]
   void setProjectorBitset(ProjectorBitset proj) requires(!ReadOnly) {
+    BoundMatrix projMatrix = bitsetToMatrix<BoundMatrix>(proj);
+    FullProjectorMapping fullMapping =
+        projectorToIndices<eBoundSize>(projMatrix);
+    setFullProjectorMapping(fullMapping);
+  }
+
+  FullProjectorMapping fullProjectorMapping() const {
```
paulgessinger, line 688: Just to confirm my understanding: *full* here means it's bound-sized, and the opposite is measurement sized, so dynamic and equal to `calibratedSize`?

Core/include/Acts/EventData/TrackStateProxy.hpp
```diff
@@ -776,6 +817,29 @@ class TrackStateProxy {
...
   }
 
+  VariableSubspaceHelper<eBoundSize> calibratedVariableSubspace() const {
+    FullProjectorMapping fullMapping = fullProjectorMapping();
+    return VariableSubspaceHelper<eBoundSize>(
+        std::span(fullMapping.begin(), fullMapping.begin() + calibratedSize()));
+  }
+
+  template <std::size_t measdim>
+  FixedSubspaceHelper<eBoundSize, measdim> calibratedFixedSubspace() const {
```
paulgessinger, line 827: Not important, but maybe we should stick to the naming of default being fixed and the opposite being called *Effective*?

Core/include/Acts/EventData/SubspaceHelpers.hpp
```diff
@@ -0,0 +1,239 @@
...
+  Expander expander() const {
+    Expander result = Expander::Zero();
+    for (auto [i, index] : enumerate(*this)) {
+      result(index, i) = 1;
+    }
+    return result;
+  }
+
+  template <std::size_t K, typename Derived>
+  ApplyLeftResult<K> applyLeftOf(
```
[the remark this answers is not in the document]
paulgessinger: Theoretically these could use operator overloads, but that might be dangerous in other respects.

Core/include/Acts/EventData/TrackStateProxy.hpp
```diff
@@ -776,6 +817,29 @@ class TrackStateProxy {
...
+    ProjectorMapping<measdim> mapping = projectorMapping();
+    return FixedSubspaceHelper<eBoundSize, measdim>(mapping);
+  }
+
+  template <std::size_t measdim, typename index_t>
+  void setCalibratedSubspace(
```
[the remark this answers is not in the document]
paulgessinger: I think one set of methods operates on the pure mapping structure, and the other one on this helper that interprets it.

I think I agree with @goetzgaycken that it might be a better idea to either only operate on the bare types and push using the helper for interpretation to the user, or only expose the helper to the user, and only deal with the raw type internally.

Core/src/TrackFinding/MeasurementSelector.cpp
```diff
@@ -47,17 +48,18 @@ double MeasurementSelector::calculateChi2(
 
         using ParametersVector = ActsVector<kMeasurementSize>;
 
-        // Take the projector (measurement mapping function)
-        const auto H =
-            projector.template topLeftCorner<kMeasurementSize, eBoundSize>()
-                .eval();
+        FixedSubspaceHelper<eBoundSize, kMeasurementSize> subspaceHelper(
+            std::span<std::uint8_t, kMeasurementSize>(
+                projector.begin(), projector.begin() + kMeasurementSize));
```
paulgessinger, line 53: I understand why this is here, but it looks a bit funky. Could we make this a factory function maybe?

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
```diff
@@ -75,19 +72,15 @@ class FixedSizeMeasurement {
   ///   of parameters and covariance.
   template <typename parameters_t, typename covariance_t>
   FixedSizeMeasurement(Acts::SourceLink source,
-                       const std::array<indices_t, kSize>& indices,
+                       const std::array<indices_t, kSize>& subsetIndices,
```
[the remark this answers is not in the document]
paulgessinger: There are typedefs in the `Types.hpp` header that could be used here.

paulgessinger (dismissed):

paulgessinger (dismissed):

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [a28acc306ccd11f064c81a68e8eb7f3cc7526c2c]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42533989) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7964833)!

PR #3189, opened 2024-05-15
refactor!: Move `Navigator` external surfaces into options

andiwand:
After https://github.com/acts-project/acts/pull/3181 and https://github.com/acts-project/acts/pull/3182 we can properly initialize the external surfaces for the `Navigator` via options without modifying the state while propagation.

blocked by:
- https://github.com/acts-project/acts/pull/3181
- https://github.com/acts-project/acts/pull/3182
- https://github.com/acts-project/acts/pull/3190

PR #3454, opened 2024-07-30
feat: Separate external and measurement surfaces

ssdetlab:
The PR fixes the absence of the external surface mechanism in the Gen2 navigation. 

As of right now the `KalmanFitter` and the `GlobalChiSquareFitter` won't compile with the `DetectorNavigator` because the latter is missing the `insertExternalSurface` method in its options. 

The PR introduces the following way of the external surface handling:

1. External surfaces can be accessed through the `Options` struct (as per #3189) to accommodate for the fitters' calls
2. `NavigationState` has constant access to the range of external surfaces within current `DetectorVolume`
3. The range from is being updated at each `updateCandidateSurfaces` call
4. `updateCandidates` from the Navigation State Updaters handles the `BoundaryTolerance`
5.  The range is being accessed in every `preStep` and `postStep` call for the surface status updating

Unit test that demonstrates the functionality is added.


Core/include/Acts/Navigation/DetectorNavigator.hpp
```diff
@@ -279,6 +297,16 @@ class DetectorNavigator {
...
 
+    BoundaryTolerance boundaryTolerance =
+        nState.surfaceCandidate().boundaryTolerance;
+    for (auto it = nState.externalSurfaceRange.first;
+         it != nState.externalSurfaceRange.second; it++) {
+      if (nextSurface->geometryId() == it->second) {
+        boundaryTolerance = BoundaryTolerance::Infinite();
+        break;
+      }
+    }
```
andiwand, line 308: I guess that could be cached from the `preStep` call somehow but not critical

Core/include/Acts/Navigation/DetectorNavigator.hpp
```diff
@@ -48,6 +50,14 @@ class DetectorNavigator {
   };
 
   struct Options : public NavigatorPlainOptions {
```
andiwand, line 52: ultimately we should ensure these interfaces with a concept

Core/include/Acts/Navigation/DetectorNavigator.hpp
```diff
@@ -400,6 +428,11 @@ class DetectorNavigator {
 
     auto& nState = state.navigation;
 
+    auto volumeID = nState.currentVolume->geometryId().volume();
+
+    nState.externalSurfaceRange =
+        nState.options.externalSurfaces.equal_range(volumeID);
```
andiwand, line 434: I think the idea of external surface is to also put them into the candidates list which I don't see here

andiwand (commented):
I wonder if the external surface could also be piped in with surface updators which would be more in the spirit of Gen2 navigation?

Left a few other comments

Core/include/Acts/Navigation/DetectorNavigator.hpp
ssdetlab: When the `updateNavigationState` call is made by the `currentVolume`, we already have access to all the surfaces that are of interest to us within next several steps, including the external ones. Then we just have to properly treat the `BoundaryTolerance` in the `updateCandidates` to account for some surfaces being external. This way the external surfaces end up in the candidate list after the `updateNavigationState` call.

ssdetlab:
After some thinking, the `volumeID`-based selection of the external surfaces does not work well when you're trying to process the real data. The standard pipeline will involve associating measurements with the surfaces through the `geometryId`s directly coupled to the surfaces. Introducing the volume-based association adds another layer of conventions that are tied to a very arbitrary volume concept. 

In the most recent commit I'm trying to solve the problem by estimating the distance to all the external surfaces and sorting them by path length. It has to be close in speed to the `volumeID`-based selection: we trade the log(N) lookups at each volume for the additional intersections and the N log(N) sorting. But, most importantly, the measurements are free from the volume id assignment conventions. 

To implement this I had to modify the `insertExternalSurface` interface a bit, so that it accepts the surface pointers.

Core/include/Acts/Navigation/DetectorNavigator.hpp
```diff
@@ -48,6 +50,14 @@ class DetectorNavigator {
   };
 
   struct Options : public NavigatorPlainOptions {
+    /// Externally provided surfaces - these are tried to be hit
+    ExternalSurfaces externalSurfaces = {};
+
+    void insertExternalSurface(const Surface* surface) {
+      externalSurfaces.push_back(
+          std::make_pair(std::numeric_limits<double>::max(), surface));
```
andiwand, line 58: isn't the first entry a long?

Core/include/Acts/Navigation/DetectorNavigator.hpp
```diff
@@ -158,6 +168,33 @@ class DetectorNavigator {
...
+      auto sIntersection = it->second->intersect(
+          state.geoContext, nState.position, nState.direction,
+          BoundaryTolerance::Infinite(), s_onSurfaceTolerance);
+
+      for (auto& si : sIntersection.split()) {
+        if (si.isValid()) {
+          it->first = si.pathLength();
+        }
+      }
+    }
```
andiwand, line 185: I would disencourage doing this in the init because directions change which makes some surfaces intersectable. especially cylinders don't play nice here

andiwand (commented):
I think the original idea about the external surfaces was to wire surfaces into the navigator which are not part of the tracking geometry. I don't see this in the code here which makes me suspect that they should appear from the surface updators somehow?

The extended bounds is then just added to make sure these surfaces are encountered during fitting.

If Gen2 is adding these surfaces another way into the surface stream we should just absorb the boundary check thingy which should be okay by looking at geo id or surface pointer comparison?

ssdetlab:
After some more thinking this is the current proposed solution for the external surface handling. 

As of right now the external surfaces are not treated in the codebase as such. They're expected to have properly assigned geometry Ids and they receive preferential treatment in the fitters/navigators in terms of the `BoundaryTolerance`. Essentially, they act right now as measurement surfaces, which is a concept proposed in the most recent commit. 

The PR treats external surfaces as just surfaces that are not a part of the ``TrackingGeometry`/`Detector` but that are to be added to the list of potential candidates in the navigation. 

The concept of a measurement surface is introduced. The measurement surface is expected to be a part of the geometry (i.e. it has to have a valid `GeometryIdentifier`) and is intersected with the infinite `BoundaryTolerance`. For the `Navigator` the process of selecting relevant measurement surfaces for the current step is left as is (i.e. based on the layer id). For the `DetectorNavigator` implementation of the backlink from `Surface` to the containing it `DetectorVolume` is introduced, inspired by the already existing backlinks across the codebase. This way the initially proposed implementation based on the geometry Id of the volume is possible without constraining the Ids of the sensitives with the hard to handle volume-based conventions. Potentially, if the Gen2 is to be persistified in some version of Acts, it has to be replaced with a method inside the `Surface`.

ssdetlab:
Closing to develop a more robust approah

PR #3458, opened 2024-07-30
perf: Drop `stateBuffer` from Core CKF

andiwand:
I noticed that we recreate this buffer for every `findTracks` call which seems to be a massive overhead at the moment. Not sure if we want to pass it in from outside or just use the given container to store these temporary states in. The user will copy out the track and clear the container anyways which keeps memory allocated and should leave enough space for consecutive calls.


andiwand:
This seems to be another ~17%
image
This is also orthogonal to https://github.com/acts-project/acts/pull/3453 and brings down the ttbar PU 200, Fatras, ODD from 1 s to 0.7 s on my machine

paulgessinger:
This was originally a performance improvement because it pushed candidates into the target track container which we did not reuse at the time. 

Great it's not needed anymore

andiwand:
> This was originally a performance improvement because it pushed candidates into the target track container which we did not reuse at the time.
>
> Can you check the memory consumption though? Before I added this, we were using 3-4x more memory in ttbar.

In principle we can just wire this container in. It will definitely safe some memory because with my change we will create temporary track states but never drop them until all the branches are complete.

Memory consumption is a good point. I suspect it goes up quite significantly while allocations will be lower.

andiwand:
@goetzgaycken do you think this will also improve the Athena performance? I guess we do not use the buffer so much there because we select the measurements before we create the states?

paulgessinger:
With selection before track state creation this certainly becomes less of an issue. IIRC this was also before the loop over seeds became external to the CKF, so this buffer was reused much more than it is now.

paulgessinger:
Also: with this change we somehow need to clarify that you absolutely need to copy out tracks from the container, because its size can be very large.

goetzgaycken:
> @goetzgaycken do you think this will also improve the Athena performance? I guess we do not use the buffer so much there because we select the measurements before we create the states?

The athena measurement selector does not use the state buffer, but not allocating the unused state buffer for each find track call (called for every used seed) will certainly help.

paulgessinger:
@goetzgaycken if it's unused, it should have no overhead because no allocations are done, no?

andiwand:
> > @goetzgaycken do you think this will also improve the Athena performance? I guess we do not use the buffer so much there because we select the measurements before we create the states?
> 
> The athena measurement selector does not use the state buffer, but not allocating the unused state buffer for each find track call (called for every used seed) will certainly help.

Ok makes sense then I now also observer about 30% CPU improvement including https://github.com/acts-project/acts/pull/3453. The calibrator still uses about 10% of the track finding which is still a mystery to me. I think it is related to the source link but I don't understand how. The code is very hot but at the same time it does not really do anything.

> Also: with this change we somehow need to clarify that you absolutely need to copy out tracks from the container, because its size can be very large.

I wonder if we should rather wire in a track state buffer or try to clear up the mess we make with the temporary track states. Or just leave it as it is now and instruct the user to copy the tracks out which potential memory waste.

paulgessinger:
It's not just about memory waste, we need to avoid prohibitively high memory consumption, which we were seeing especially in the past.

andiwand:
> @goetzgaycken if it's unused, it should have no overhead because no allocations are done, no?

In principle it should not really show up in the profile but it will definitely cost something because I guess the vector will allocate some initial size? If we do not intent to use it we should definitely get rid of it.

> It's not just about memory waste, we need to avoid prohibitively high memory consumption, which we were seeing especially in the past.

Without a branch aborter and high branching this can definitely happen. But I don't think that is what we want to optimize for.

paulgessinger:
Agreed it should go if we don't need it. 

Most vector implementations have zero initial capacity, so no allocation at all.

goetzgaycken:
> @goetzgaycken if it's unused, it should have no overhead because no allocations are done, no?
There is still this allocation, it seems:
```
r.stateBuffer = std::make_shared<TrackStateContainerBackend>();
```
I don't expect this to be significant, but we still process many seeds ...

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [e6b73b51e9f4a9f40f1c7aea7445a4f3b84e9d84]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42516722) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7962065)!

PR #3464, opened 2024-08-01
ci: Use pre-commit for CI lint

paulgessinger:
- Formatting C++ and Python files
- Check trailing whitespace
- Check consistent newlines
- Check YAML contents


paulgessinger:
I'm planning to follow this up with a bump of clang-format to v18, and an introduction of cmake file formatting.

Examples/Scripts/Digitization/error_parameterisation.py
```diff
@@ -237,8 +237,8 @@ def run_error_parametriation(
 </head>
 <body>
 <div class="wrapper">
-    <a href="{previous}">Previous volume</a> | 
-    <a href="../index.html">Back to index</a> | 
+    <a href="{previous}">Previous volume</a> |
+    <a href="../index.html">Back to index</a> |
```
andiwand, line 241: this might be an interesting case. I think the newline will capture the space?

cmake/pythia8307-cpp20.patch
andiwand, line 1: does that work on a patch file?

docs/core/misc/figures/AxisBoundaryTypes.svg
andiwand, line 1: 1. wtf github has useful image diff
2. should this affect svgs?

paulgessinger: SVGs are text files, and they can have trailing spaces, so I guess, why not? It shouldn't change the visual image at all I don't think.

cmake/pythia8307-cpp20.patch
paulgessinger: I'm not even sure this file should still be here anymore. I'll remove it. However, to your point, I don't know.

Examples/Scripts/Digitization/error_parameterisation.py
paulgessinger: I believe so, yes.

andiwand (dismissed):

paulgessinger:
@andiwand this had a conflict, which should be resolved now.

andiwand (approved):

PR #3466, opened 2024-08-01
perf: Optimize SourceLink creation, track state assignment

paulgessinger:
This PR adds a benchmark for some `SourceLink` operations, and includes some optimization of creation and assignment into track states.

- Benchmark `main` 
```
Creating source link
  20000 runs of 100000 iteration(s), 7248.0ms total, 357.7920+/-0.1250µs per run, 3.578+/-0.395ns per iteration
Copy construct source link
20000 runs of 100000 iteration(s), 2583.0ms total, 126.8750+/-0.0349µs per run, 1.269+/-0.110ns per iteration
Copy then move construct source link
20000 runs of 100000 iteration(s), 4265.5ms total, 210.3750+/-0.0655µs per run, 2.104+/-0.207ns per iteration
Optional assignment
20000 runs of 100000 iteration(s), 9581.2ms total, 475.5420+/-0.1245µs per run, 4.755+/-0.394ns per iteration
Create track state
20000 runs of 10000 iteration(s), 12651.4ms total, 174.1250+/-0.4740µs per run, 17.413+/-4.740ns per iteration
Assign source link to track state
20000 runs of 10000 iteration(s), 6752.2ms total, 275.4160+/-0.6607µs per run, 27.542+/-6.607ns per iteration
```

- Benchmark `feat`
```
Creating source link
  20000 runs of 100000 iteration(s), 1528.3ms total, 75.2500+/-0.0118µs per run, 0.752+/-0.037ns per iteration
Copy construct source link
20000 runs of 100000 iteration(s), 2568.1ms total, 126.6250+/-0.0190µs per run, 1.266+/-0.060ns per iteration
Copy then move construct source link
20000 runs of 100000 iteration(s), 4225.4ms total, 209.7500+/-0.0177µs per run, 2.098+/-0.056ns per iteration
Optional assignment
20000 runs of 100000 iteration(s), 1827.6ms total, 90.5830+/-0.0027µs per run, 0.906+/-0.009ns per iteration
Create track state
20000 runs of 10000 iteration(s), 12539.9ms total, 177.1670+/-0.5062µs per run, 17.717+/-5.062ns per iteration
Assign source link to track state
20000 runs of 10000 iteration(s), 4855.1ms total, 189.8750+/-0.6325µs per run, 18.988+/-6.325ns per iteration
```

(measurements on my laptop)

- Raw creation time goes from about 3.6ns to about 0.8ns (about a 5x speedup)
- Move assignment of an empty optional from a source link goes from 4.8ns to 0.9ns (including a raw source link construction of of 0.8ns)
- Assignment of a source link into a track state (which holds an empty optional) goes from 27.5ns to 19ns (40% speedup), which includes track state creation (17.7ns which is unchanged by the optimizations in this PR).


paulgessinger:
@goetzgaycken @andiwand

Core/include/Acts/Utilities/Any.hpp
paulgessinger, line 1: Changes are not strictly needed here, I can remove them if preferred.

Tests/Benchmarks/SourceLinkBenchmark.cpp
```diff
@@ -0,0 +1,138 @@
...
+class BenchmarkSourceLink final {
+ public:
+  using Index = std::uint32_t;
+
+  /// Construct from geometry identifier and index.
+  constexpr BenchmarkSourceLink(Acts::GeometryIdentifier gid, Index idx)
+      : m_geometryId(gid), m_index(idx) {}
+
+  // Construct an invalid source link. Must be default constructible to
+  /// satisfy SourceLinkConcept.
```
paulgessinger, line 27:
```suggestion
```

Core/include/Acts/EventData/TrackStateProxy.hpp
```diff
@@ -678,7 +678,7 @@ class TrackStateProxy {
 
   /// Set an uncalibrated source link
   /// @param sourceLink The uncalibrated source link to set
-  void setUncalibratedSourceLink(SourceLink sourceLink) requires(!ReadOnly) {
+  void setUncalibratedSourceLink(SourceLink&& sourceLink) requires(!ReadOnly) {
```
paulgessinger, line 681: I can also achieve this with a const reference, which might be a bit easier on clients (controls where the copy happens)

goetzgaycken:
why is creation now faster than copy ?

Core/include/Acts/EventData/TrackStateProxy.hpp
andiwand: I would prefer value or reference over rvalue

paulgessinger:
@goetzgaycken I think it's because the creation needs to do less. The construction into the local buffer is known at compile time, so no pointer dereferencing is needed at all. Any type of copy or move operation needs to consult the type-specific handler pointer to figure out what to do, even if it's a trivial memcpy in the end.

Core/include/Acts/EventData/TrackStateProxy.hpp
paulgessinger: I'd have thought value + `move` would give ~ same performance as revalue ref, but it's 20-25% slower.

andiwand: that is unexpected - I wonder if the optimizer is tripping somehow

paulgessinger: Maybe. I'm not sure I'd expect the compiler to completely eliminate the move construction in all cases. With full on rvalue references, we basically are guaranteed to only do a single move construction into the backing vector, and nothing else in the call chain.

goetzgaycken: In principle move also does work on the source where copy only does work on the destination. When looking at the handler the move methods have more lines than the copy methods, so likely for simple types move adds overhead.

andiwand (dismissed):
let's get this in

acts-project-service:
## :red_circle: Athena integration test results [50bc40bf9c656d20aa311bfacf8189472395e23e]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42054857) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7887589)!

paulgessinger:
Sonar is happy now, and I've pushed an update which should make the Athena build happy.

I'll add another PR to make the API breaking changes in v37.

andiwand (approved):

paulgessinger:
Let's see if the docs build and format is happy now.

andiwand (approved):

PR #3467, opened 2024-08-01
feat(geometry): Allow operator<< on surfaces without detector elements

paulgessinger:
We previously disallowed this, but this is only necessary if we have a detector element. Arguably, specifically for printing it can sometimes be inconvenient to have to pass around a geometry context just for this.

### DIFF
--- a/Core/src/Surfaces/Surface.cpp
+++ b/Core/src/Surfaces/Surface.cpp
@@ -357,3 +357,16 @@ void Acts::Surface::assignSurfaceMaterial(
 void Acts::Surface::associateLayer(const Acts::Layer& lay) {
   m_associatedLayer = (&lay);
 }
+
+std::ostream& Acts::operator<<(std::ostream& os, const Acts::Surface& srf) {
+  if (srf.associatedDetectorElement() != nullptr) {
+    throw std::runtime_error(
+        "Cannot print Surface with associated DetectorElement without geometry "
+        "context");
+  }
+  // Using a default context is ONLY safe, if the surface does not have an
+  // associated detector element, which we ensure right above.
+  Acts::GeometryContext defaultContext;
+  os << srf.toStream(defaultContext);
+  return os;
+}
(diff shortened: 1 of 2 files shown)


paulgessinger:
This is convenient, but might not be 100% clean in all cases.

Thoughts @andiwand, @benjaminhuth, @asalzburger ?

Core/src/Surfaces/Surface.cpp
```diff
@@ -357,3 +357,16 @@ void Acts::Surface::assignSurfaceMaterial(
...
+std::ostream& Acts::operator<<(std::ostream& os, const Acts::Surface& srf) {
+  if (srf.associatedDetectorElement() != nullptr) {
+    throw std::runtime_error(
+        "Cannot print Surface with associated DetectorElement without geometry "
+        "context");
+  }
+  // Using a default context is ONLY safe, if the surface does not have an
+  // associated detector element, which we ensure right above.
+  Acts::GeometryContext defaultContext;
+  os << srf.toStream(defaultContext);
```
andiwand, line 370: I wonder if there should also be a `toStream` overload without context which then drives the exception also allowing for overrides later

paulgessinger: This requires implementing it on all derived surface types. I'm not sure that's needed, and they would likely also call each other internally to avoid copying code. Does that make sense?

andiwand: I thought we could default it to your behavior here and give the surface the opportunity to specialize but maybe this does not have any use case anyways

andiwand:
I guess this could be surprising using the `<<` for any surface but the exception should communicate this well so no objections form my side

Core/src/Surfaces/Surface.cpp
paulgessinger: I see your point. I guess the only reason using `toStream` over `operator<<` is that with `toStream` you can in fact pass a geometry context.

paulgessinger:
@andiwand That's exactly correct. I'm debating myself if the trade-off of confusion is worth it.

PR #3472, opened 2024-08-02
fix: Avoid segfaults in bin adjustments

benjaminhuth:
Previously, only the surface type was checked, and then a bounds type assumed. Then, the result of a `dynamic_cast` was used unchecked.
This changes this to ensure correcteness by checking the `dynamic_cast`s.


Core/include/Acts/Utilities/BinAdjustment.hpp
```diff
@@ -240,26 +240,28 @@ static inline BinUtility adjustBinUtility(const BinUtility& bu,
                                           const Surface& surface,
                                           const GeometryContext& gctx) {
   // The surface type is a cylinder
-  if (surface.type() == Surface::Cylinder) {
+  if (surface.type() == Surface::Cylinder &&
+      surface.bounds().type() == SurfaceBounds::eCylinder) {
     // Cast to Cylinder bounds and return
-    auto cBounds = dynamic_cast<const CylinderBounds*>(&(surface.bounds()));
+    auto cBounds = static_cast<const CylinderBounds*>(&(surface.bounds()));
```
andiwand, line 246: is that safe with virtual inheritance? or only as long as we do no diamond stuff?

andiwand (dismissed):

andiwand:
> Why not just switch from the type check to checking the dynamic cast instead? That seems the safest approach to me.

Right checking the result of the `dynamic_cast` would be the other way. I guess with having the `type` field and RTTI we always end up doing both mixed

benjaminhuth:
Yeah actually for some reason I haven't considered `dynamic_cast` only here, but it would be the cleanest here because its not a hot path... I'll change it to `dynamic_cast` only...

andiwand (approved):

acts-project-service:
## :red_circle: Athena integration test results [0b9ed368c93ea752ed1751e7fe7095af6ce61090]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/41970139) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7872892)!

PR #3474, opened 2024-08-02
fix: Really handle holes after measurement selection in Core CKF

andiwand:
Looks like I lost the actual objective in https://github.com/acts-project/acts/pull/3413 which did not resolve the issue.

After the refactoring of the measurement selector it is possible that we end up with no measurements after selection. That case was not handled by the Core CKF yet.

https://github.com/acts-project/acts/pull/3413 was an attempt to fix the problem but the hole handling was shifted to the wrong place. We want to check for a hole after `trackStateCandidateCreator` and before `processNewTrackStates`.


Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -927,23 +926,27 @@ class CombinatorialKalmanFilter {
...
       unsigned int nBranchesOnSurface = 0;
-      unsigned int nStoppedBranches = 0;
 
       auto rootBranch = result.activeBranches.back();
 
-      // Build the new branches by forking the root branch
+      // Build the new branches by forking the root branch. Reverse the order to
+      // process the best candidate first
       CkfTypes::BranchVector<TrackProxy> newBranches;
-      for (auto [i, tipIndex] : enumerate(newTrackStateList)) {
```
goetzgaycken, line 944: Not quite sure I understand the reason of this change. 
- the assembly created by the two version should bemostly identical 
- newTrackStateList should be sorted by chi2 (increasing order)

goetzgaycken (commented):
Looks fine to me. To be sure the relevant changes are that the hole count was increased for the preceding state not the newly added state and the branch intended for newTrackStateList.empty() could never be entered.  Both these issues are fixed by this PR. Or is there something else ? 

Otherwise I am only wondering about the newTrackStateList, in principle it should be ordered by chi2, if not than there might be an incompatibility between the default trackStateCandidateCreator and the new one in athena.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
andiwand: reason being that we store the chi2 in increasing order but the branches are processed in the inverse order so back first. to process the best branch first I reversed the order of the measurements

goetzgaycken: Just to answer myself, if I am not mistaken the new branches are "popped" from the top. Thus, adding the "best" state last would lead to the corresponding branch to be processed first.

goetzgaycken (approved):

acts-project-service:
## :red_circle: Athena integration test results [95da6aa371b891933f12735cee1d152befc9af7d]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42091417) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7893153)!

PR #3475, opened 2024-08-02
feat: Max chi2 for outliers in Core `MeasurementSelector`

andiwand:
- Add `chi2CutOffOutlier` which allows to cut for holes
- Refactor cut selection
- Refactor return logic

This should allow testing https://github.com/acts-project/acts/pull/3474 properly in ACTS standalone

blocked by
- https://github.com/acts-project/acts/pull/3474


Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -36,8 +36,10 @@ namespace Acts {
 struct MeasurementSelectorCuts {
   /// bins in |eta| to specify variable selections
   std::vector<double> etaBins{};
-  /// Maximum local chi2 contribution.
+  /// Maximum local chi2 contribution to classify as outlier.
   std::vector<double> chi2CutOff{15};
```
goetzgaycken, line 40: In the most likely case that the vector only contains one element this does not matter. Otherwise double precision for the chi2 cut feels over precise and wastes cache.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -745,64 +744,23 @@ class CombinatorialKalmanFilter {
         const CkfTypes::BranchVector<TrackIndexType>& newTrackStateList =
             *tsRes;
 
-        Result<std::pair<unsigned int, unsigned int>> procRes =
```
goetzgaycken, line 748: This and many other changes belong to #3474, I suppose, If the inclusion of these changes is intended than this is fine.

Core/include/Acts/TrackFinding/MeasurementSelector.ipp
```diff
@@ -28,24 +28,17 @@ MeasurementSelector::select(
 
   // Get geoID of this surface
   auto geoID = candidates.front().referenceSurface().geometryId();
+  // Get the eta of the first track state
+  const auto eta =
+      std::atanh(std::cos(candidates.front().predicted()[eBoundTheta]));
```
goetzgaycken, line 33: Is there no utility function for eta ?

Core/src/TrackFinding/MeasurementSelector.cpp
```diff
@@ -64,4 +65,38 @@ double MeasurementSelector::calculateChi2(
       });
 }
 
+MeasurementSelector::Cuts MeasurementSelector::getCutsByEta(
+    const MeasurementSelectorCuts& config, double eta) {
+  const double etaAbs = std::abs(eta);
+  const auto it =
+      std::lower_bound(config.etaBins.begin(), config.etaBins.end(), etaAbs);
```
goetzgaycken, line 72: I think lower bound does a binary search, which is slower for n ~<10. Not that it matters much. But the original code seems simpler.

Core/src/TrackFinding/MeasurementSelector.cpp
```diff
@@ -64,4 +65,38 @@ double MeasurementSelector::calculateChi2(
...
+    return vec.at(std::clamp(idx, std::size_t{0}, vec.size() - 1));
+  };
+
+  const double chi2CutOffMeasurement = select(config.chi2CutOff, bin);
+  const double chi2CutOffOutlier = config.chi2CutOffOutlier.empty()
+                                       ? std::numeric_limits<double>::infinity()
+                                       : select(config.chi2CutOffOutlier, bin);
+  const std::size_t numMeasurementsCutOff =
+      select(config.numMeasurementsCutOff, bin);
+  return {chi2CutOffMeasurement, chi2CutOffOutlier, numMeasurementsCutOff};
```
goetzgaycken, line 85: To me it would make more sense to store the cuts directly in such a struct, rather than recreating the struct for each measurement selection.

goetzgaycken (commented):
Seems fine to me. Only the changes concerning the cut search do not seem to be an improvement.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
andiwand: Right I put this in here to check physmon

Core/include/Acts/TrackFinding/MeasurementSelector.ipp
andiwand: I fear only for vectors. I can check again

Core/src/TrackFinding/MeasurementSelector.cpp
andiwand: I agree but that would be a breaking change

Core/include/Acts/TrackFinding/MeasurementSelector.ipp
andiwand: looks like there is no such function

andiwand:
This shows similar performance after modifying the cuts a bit
image

goetzgaycken (approved):
Looks fine.
Some nit-picks:
- check comments concerning chi2 cuts
- I think it would  make more sense to ensure that chi2, numMeasurement cuts all have one value per etaBin, rather than clamping bin to the allowed range (only 0, or [0...n_eta_bins) would make sense anyways, everything else would be a bug.)

andiwand:
I tried to improve the cuts a bit and check on their sizes. I think we cannot go much further for now as this would be breaking otherwise.

goetzgaycken (approved):
Looks fine to me. 

The only complaint I would have is that the track selection now has something like 20 possibly eta dependent cuts, all are double or size_t where none of the values really requires such a precision / bit width. It is not a lot of data but still needs to be loaded or stored in the cache.  But, that is maybe something to consider for a future update.

goetzgaycken (approved):
Still all fine.

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -125,6 +125,7 @@
         "nMeasurementsMin",
         "maxHoles",
         "maxOutliers",
+        "maxHolesAndOutliers",
```
timadye, line 128: need to update the defaults below:
`defaults=[(None, None)] * 7 + [None] * 7,`

timadye: otherwise I get a Python error with full_chain_itk.py (TrackSelectorConfig missing 1 required positional argument: 'loc0').

timadye:
I didn't yet check whether the variant access is a result of this PR, or is also there in `main`.

timadye:
OK, the `bad_variant_access` is also there in `main`, so shouldn't hold up this PR. I'll check some more, and if necessary submit an issue. For this PR, there's still the Python error.

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -37,8 +37,10 @@ namespace Acts {
 struct MeasurementSelectorCuts {
   /// bins in |eta| to specify variable selections
   std::vector<double> etaBins{};
-  /// Maximum local chi2 contribution.
+  /// Maximum local chi2 contribution to classify as measurement.
   std::vector<double> chi2CutOff{15};
+  /// Maximum local chi2 contribution to classify as outlier.
+  std::vector<double> chi2CutOffOutlier{25};
```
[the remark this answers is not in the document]
andiwand: good point I thought I kept this non breaking at some point but somehow a value ended up here. I will rever this

Core/include/Acts/TrackFinding/MeasurementSelector.ipp
timadye, line 33: The previous code only calculated `eta` if there were multiple eta bins, and so far, I think we have only used cuts that are not eta-dependent. This is a potential pessimisation.
I just realised that in Athena, when we harmonised all the eta binning, we picked up multiple eta bins also for the `MeasurementSelector`, so that might need reoptimising. That won't help if here we calculate eta even if it isn't needed.

andiwand: I can special case this again

Core/src/TrackFinding/MeasurementSelector.cpp
```diff
@@ -67,4 +91,41 @@ double MeasurementSelector::calculateChi2(
...
+      }
+    }
+  }
+
+  const double chi2CutOffMeasurement = config.chi2CutOff.at(bin);
+  const double chi2CutOffOutlier = config.chi2CutOffOutlier.empty()
+                                       ? std::numeric_limits<double>::infinity()
+                                       : config.chi2CutOffOutlier.at(bin);
+  const std::size_t numMeasurementsCutOff =
+      config.numMeasurementsCutOff.at(bin);
```
timadye, line 113: For each vector, it used to be possible that `bin >= cut.size()`, because we did not have to specify the cut for all eta bins. The convention we had before (and is used in Athena) was that it used the the last-bin cut for all subsequent bins. This is convenient to allow to specify a single element for cuts that are not eta-dependent, even if we have eta bins for the other cuts.

I see that now `validateCuts` checks that all bins are specified, so it will not crash here. Rather, I guess, `validateCuts` will fail for configurations that previously worked.

Isn't it easier to limit the `bin` index for each `vector`. This could be done with a simple lambda. Unfortunately, a lambda can't handle the empty vector case, because of the different no-cut values for `numMeasurementsCutOff`.

andiwand: Right this is what I had in a previous commit but @goetzgaycken and I agreed on simplifying it. I will try to make this non breaking again

timadye:
```
auto lim = [](const auto& vec, std::size_t bin) { return bin<vec.size() ? vec[bin] : vec.back(); };
```

timadye: Sorry I missed @goetzgaycken's comment before. I could have pushed back a bit at the time. Quite apart from the incompatibility with previous behaviour, I think it is useful to have this flexibility in the configuration.

timadye: Not that I am saying you should necessarily use this, but I was pleased to find the general limit lambda:
```
    auto lim = [](const auto& vec, std::size_t bin) {
        return vec.empty()
                   ? std::numeric_limits<std::remove_reference_t<decltype(vec[0])>>::max()
               : bin < vec.size() ? vec[bin]
                                  : vec.back();
    };
```
(returns max `double`, rather than `inf`, but that's OK I think.)

Core/src/TrackFinding/MeasurementSelector.cpp
```diff
@@ -15,17 +15,41 @@
...
+
+  if (cuts.chi2CutOff.size() != cuts.numMeasurementsCutOff.size()) {
+    throw std::invalid_argument(
+        "chi2CutOff and numMeasurementsCutOff must have the same size");
+  }
+
+  if (cuts.chi2CutOff.size() != cuts.etaBins.size() + 1) {
+    throw std::invalid_argument(
+        "chi2CutOff must have one more element than etaBins");
+  }
```
timadye, line 51: I guess we no longer want all these checks: just that `chi2CutOff` and `numMeasurementsCutOff` aren't empty, since those aren't allowed to be empty in `getCutsByTheta`.

Core/src/TrackFinding/MeasurementSelector.cpp
```diff
@@ -67,4 +91,46 @@ double MeasurementSelector::calculateChi2(
...
+    return bin < vec.size() ? vec[bin] : vec.back();
+  };
+
+  const double chi2CutOffMeasurement = getBinOrBack(config.chi2CutOff, bin);
+  const double chi2CutOffOutlier =
+      config.chi2CutOffOutlier.empty()
+          ? std::numeric_limits<double>::infinity()
+          : getBinOrBack(config.chi2CutOffOutlier, bin);
+  const std::size_t numMeasurementsCutOff =
+      config.numMeasurementsCutOff.at(bin);
```
timadye, line 118:
```suggestion
      getBinOrBack(config.numMeasurementsCutOff, bin);
```

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -93,11 +95,15 @@ class MeasurementSelector {
...
-      const std::vector<cut_value_t>& cuts, const Logger& logger);
+  struct Cuts {
+    double chi2Measurement{};
+    double chi2Outlier{};
+    std::size_t numMeasurements{};
+  };
+
+  static Cuts getCutsByTheta(const MeasurementSelectorCuts& config,
+                             double theta);
+  Result<Cuts> getCuts(const GeometryIdentifier& geoID, double theta) const;
```
timadye, line 106: I wonder if these would be useful for user `MeasurementSelector`s (@goetzgaycken), now they support a separate outlier cut. It would be good to harmonise where possible.

Maybe best left to a separate PR, but worth considering soon.

Core/src/TrackFinding/MeasurementSelector.cpp
```diff
@@ -64,4 +66,45 @@ double MeasurementSelector::calculateChi2(
...
+    const GeometryIdentifier& geoID, double theta) const {
+  // Find the appropriate cuts
+  auto cuts = m_config.find(geoID);
+  if (cuts == m_config.end()) {
+    // for now we consider missing cuts an unrecoverable error
+    // TODO consider other options e.g. do not add measurements at all (not
+    // even as outliers)
+    return CombinatorialKalmanFilterError::MeasurementSelectionFailed;
+  }
+  assert(!cuts->chi2CutOff.empty());
```
timadye, line 106:
```suggestion
```
no longer needed, I think.

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
andiwand: I agree this is just an intermediate step to get an outlier chi2 threshold that we can test in ACTS. The next step would be to generalize the measurement selector which requires breaking changes

Core/src/TrackFinding/MeasurementSelector.cpp
timadye: this is a `private` function (at least for the moment), so I don't think a change would be breaking.

However, I think RVO ensures that this will fill the caller's `MeasurementSelector::Cuts` variable. That may be complicated when filling into a `Result<>` (I can never remember the rules, so would need to check with godbolt), but I don't think we can easily do better with a different API.

goetzgaycken (dismissed):
I do not see anything wrong.

I think I remarked that before and it is not really very important but there is nothing that prevents an inconsistency between the per eta bin cuts and the eta bins. There are only the three cases which make sense: no cut value, one cut value, and one cut value per eta bin. In the latest version the cut value will be the last one or the one matching the eta bin. In case the cut values and eta bins are inconsistent a kind of arbitrary cut value is silently chosen.

Btw. since the cuts are now requested by theta, it would make sense to actually bin by theta which would avoid the computation of eta.

timadye:
The code looks fine, but it rejected my approval because of a more recent update. Unfortunately it also rejected all my performance measurements in the same comment. I'll try to reproduce now...

I tested this PR (just before the last update, 8c640ae77) with 100 ttbar_pu200 events with `full_chain_itk.py`. That script uses the default &chi;&sup2; cuts, so this will include the new outlier cut (&chi;&sup2;<25). Comparing against `main` (7c272316b), the `TrackFindingAlgorithm` is 11% slower than `main`, but still 10% faster than before (30/07/2024, `main` 68c92a61f). The global efficiency is very slightly better (99.0&rarr;99.2%), and there are some other changes in performance:
trackeff_vs_eta
nHoles_vs_eta
nOutliers_vs_eta
fakerate_vs_eta
These changes look reasonable, and though I don't understand why it is slower converting outliers to holes.

andiwand:
@timadye I think the increase and decrease of the CPU performance is caused by https://github.com/acts-project/acts/pull/3453 and https://github.com/acts-project/acts/pull/3527

timadye:
I included the latest change with the internal cuts (b997e9b27) and ran again. That update gave a 7% speedup, so we are now only 4% slower than `main` (7c272316b). There is no significant change in the performance from the update (see previous plots).

Core/include/Acts/TrackFinding/MeasurementSelector.hpp
```diff
@@ -93,11 +95,25 @@ class MeasurementSelector {
          bool& isOutlier, const Logger& logger) const;
 
  private:
-  template <typename traj_t, typename cut_value_t>
-  static cut_value_t getCut(
-      const typename traj_t::TrackStateProxy& trackState,
-      const Acts::MeasurementSelector::Config::Iterator selector,
-      const std::vector<cut_value_t>& cuts, const Logger& logger);
+  struct InternalCutBin {
+    double maxAbsTheta{};
```
timadye, line 99:
```suggestion
    double maxTheta{};
```
&theta; goes from 0 to &pi; [[1](https://acts.readthedocs.io/en/latest/tracking.html#track-parametrization)].

Core/src/TrackFinding/MeasurementSelector.cpp
```diff
@@ -10,19 +10,58 @@
...
+    using Value = std::remove_reference_t<decltype(vec[0])>;
+    static constexpr Value max = std::numeric_limits<Value>::max();
+    return vec.empty() ? max : (bin < vec.size() ? vec[bin] : vec.back());
+  };
+
+  for (std::size_t bin = 0; bin < config.etaBins.size() + 1; ++bin) {
+    InternalCutBin cuts;
+    cuts.maxAbsTheta = bin < config.etaBins.size()
+                           ? std::acos(std::tanh(config.etaBins[bin]))
+                           : std::numeric_limits<double>::max();
```
timadye, line 55:
```suggestion
                           : std::numeric_limits<double>::infinity();
```
we only had to use `max()` to get the generic lambda to work.

timadye:
> @timadye I think the increase and decrease of the CPU performance is caused by #3453 and #3527

I compared `main` before and after #3527 (which I understand just reverts #3453): it was just 1% slower after, and no change in performance.

andiwand:
> > @timadye I think the increase and decrease of the CPU performance is caused by #3453 and #3527
> 
> I compared `main` before and after #3527 (which I understand just reverts #3453): it was just 1% slower after, and no change in performance.

That is odd - I measured #3453 to make 16% difference

Core/src/TrackFinding/MeasurementSelector.cpp
```diff
@@ -64,4 +103,32 @@ double MeasurementSelector::calculateChi2(
...
+MeasurementSelector::Cuts MeasurementSelector::getCutsByTheta(
+    const InternalCutBins& config, double theta) {
+  const double absTheta = std::abs(theta);
+
+  auto it = std::find_if(config.begin(), config.end(),
+                         [absTheta](const InternalCutBin& cuts) {
+                           return absTheta < cuts.maxAbsTheta;
+                         });
+  assert(it != config.end());
+  std::size_t bin = std::distance(config.begin(), it);
```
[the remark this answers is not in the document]
andiwand: But that would switch one `if` with another `if` no? Anyways I would rather keep the code simple and readable without additional branches

Core/src/TrackFinding/MeasurementSelector.cpp
```diff
@@ -105,11 +111,12 @@ double MeasurementSelector::calculateChi2(
...
-  const double absTheta = std::abs(theta);
-
   auto it = std::find_if(config.begin(), config.end(),
-                         [absTheta](const InternalCutBin& cuts) {
-                           return absTheta < cuts.maxAbsTheta;
+                         [theta](const InternalCutBin& cuts) {
+                           // since theta is in [0, pi] and we have a symmetric
+                           // cut in eta, we can just look at the positive half
+                           // of the Z axis
+                           return std::min(theta, M_PI - theta) < cuts.maxTheta;
```
timadye, line 119: Good catch. You can move the `std::min(theta, M_PI - theta)` outside the algorithm.

Have you tested the eta-dependent cuts? I realise my tests in `full_chain_itk.py` just have one cut for all eta - and it is the currently same in Athena.

andiwand: just tested it manually with code changes in Python and it seems to work

timadye (dismissed):

Core/src/TrackFinding/MeasurementSelector.cpp
```diff
@@ -64,4 +109,34 @@ double MeasurementSelector::calculateChi2(
...
+
+  auto it = std::find_if(config.begin(), config.end(),
+                         [constrainedTheta](const InternalCutBin& cuts) {
+                           return constrainedTheta < cuts.maxTheta;
+                         });
+  assert(it != config.end());
+  std::size_t bin = std::distance(config.begin(), it);
+
+  return {config[bin].maxNumMeasurements, config[bin].maxChi2Measurement,
+          config[bin].maxChi2Outlier};
```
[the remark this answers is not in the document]
andiwand: Oh I overlooked that. Much nicer

timadye (approved):

acts-project-service:
## :white_check_mark: Athena integration test results [3e934e4babe37d3ca456cb14eb14a8b435bb620c]

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704971)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704970)||
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704969)||
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704968)||
|:green_circle: | [test_ActsCheckObjectCountsCachedWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704967)||
|:green_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704966)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704965)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704964)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704963)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704962)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704961)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704960)||
|:green_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704959)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704958)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704957)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704956)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704955)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704954)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704953)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704952)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704951)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704950)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704949)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704948)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704947)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704946)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704945)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704944)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704943)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704942)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704941)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704940)||
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42704939)||

PR #3268, opened 2024-06-07
feat: GeoShapes to DetectorVolumes and GeoMaterial Converters

Matthewharri:
This PR introduces a converter to convert geoShapes to DetectorVolumes. Some of the shapes added are very simple converters, i.e. simple bounding boxes which will probably be changed later on. I also bump the GeoModel version from 4.6.0 to 6.0.0 since Athena is using a much newer version. I also change (with the help of @paulgessinger) how cmake looks for the system installation of GeoModel.

I also include a way to convert `GeoMaterial` to `Acts::Material`

@junggjo9 

blocked by
- https://github.com/acts-project/acts/pull/3476

## Summary by CodeRabbit

- **New Features**
	- Introduced new functions for converting GeoModel shapes into detector volumes and for converting GeoMaterial objects into Acts materials.
	- Added support for multiple geometric shapes in the conversion process, enhancing compatibility with the Acts framework.

- **Bug Fixes**
	- Improved error handling for unsupported geometric shapes during conversion.

- **Documentation**
	- New header files added with detailed function declarations and usage instructions.

PR #3135, opened 2024-04-22
feat: Add CMake presets

andiwand:
Adds CMake presets (https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) to collect common configuration and build settings.

The current list of presets:
- `common` as a base for all the other presets. This enables everything necessary for running the ODD full chain examples with Fatras. It sets the cpp standard to 20, the generator to ninja and enables ccache.
- `dev` uses `RelWithDebInfo` and enables `CMAKE_EXPORT_COMPILE_COMMANDS` and `ACTS_FORCE_ASSERTIONS`
- `perf` uses `Release` sets a few compiler flags to keep frame pointers and enables `ACTS_BUILD_BENCHMARKS`
- `ci-common` is the base for all CI presets. It enables the usual flags for what to build in the CI, uses `Release`, `-Werror` and enables `ACTS_FORCE_ASSERTIONS` and `ACTS_ENABLE_LOG_FAILURE_THRESHOLD`
- `github-ci` will not change anything for now
- `gitlab-ci` will switch to `RelWithDebInfo` to allow FPE monitoring

While doing so I also
- removed the `ACTS_BUILD_EVERYTHING` CMake variable because the preset can be used to collect the "default" flags
- removed the `ACTS_BUILD_EXAMPLES` CMake option for a similar reason. I kept the cmake variable because it is used to toggle things internally in our CMake

PR #3117, opened 2024-04-18
feat: covfie plugin

fredevb:
Added functions for creating fields in covfie from magnetic fields in acts. I.e.:

- Acts::CovfiePlugin::covfieField(interpolatedMagneticField)
- Acts::CovfiePlugin::covfieField(constantBField)

The following method can be used for more a more general conversion by sampling an acts magnetic field provider (which creates a linearly interpolated covfie field):
- Acts::CovfiePlugin::covfieField(magneticFieldProvider, cache, nBins, min, max)

Added tests for conversion of InterpolatedMagneticField, ConstantBField, and SolenoidBField.
Added Python bindings using pybind11 and a simple test.

Furthermore:
- Added --exclude "*build/_deps/*" to clang-tidy (CI)
- Updated cmake options ACTS_BUILD_EVERYTHING to include ACTS_BUILD_PLUGIN_TRACCC (which includes the covfie plugin)

PR #3476, opened 2024-08-05
chore: Bump minimum GeoModel version to 6.3.0

andiwand:
Bumps GeoModel version and deals with the necessary changes. This is a subset of https://github.com/acts-project/acts/pull/3268 which isolates the version change and tackles the CI errors faced there.


cmake/ActsExternSources.cmake
```diff
@@ -31,7 +31,7 @@ set( ACTS_FRNN_SOURCE
 mark_as_advanced( ACTS_FRNN_SOURCE )
 
 set( ACTS_GEOMODEL_SOURCE
-   "GIT_REPOSITORY;https://gitlab.cern.ch/GeoModelDev/GeoModel;GIT_TAG;4.6.0;" CACHE STRING "Source to take GeoModel from")
+   "GIT_REPOSITORY;https://gitlab.cern.ch/GeoModelDev/GeoModel;GIT_TAG;${_acts_geomodel_version};" CACHE STRING "Source to take GeoModel from")
```
stephenswat, line 34: Question: I am operating on Fred's #3117 and he seems to have insisted on removing the trailing semicolon in this source string; do you know if this is a potential problem or if the semicolon should remain?

andiwand: happy to try dropping it

andiwand:
@krasznaa CMake seems stable now if you want to have a look. I have to check the unit test

thirdparty/GeoModel/CMakeLists.txt
```diff
@@ -7,20 +7,21 @@
...
+set(GEOMODEL_VERSION "${_acts_geomodel_version}")
+set(GEOMODEL_SETUP_JSON OFF CACHE BOOL "Skip setting up json completely")
+set(GEOMODEL_BUILD_GEOMODELG4 "${ACTS_BUILD_PLUGIN_GEANT4}" CACHE BOOL "Build the Geant4 plugin")
 
 # Declare where to get geomodel from.
 FetchContent_Declare(geomodel ${ACTS_GEOMODEL_SOURCE}
     PATCH_COMMAND git am ${CMAKE_CURRENT_SOURCE_DIR}/0001-Add-option-to-skip-setting-up-json-completely.patch)
 
+# Disable warnings flagged as errors
+set_property(DIRECTORY APPEND PROPERTY COMPILE_OPTIONS -Wno-error)
```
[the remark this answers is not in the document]
andiwand: I think moving https://github.com/acts-project/acts/blob/main/CMakeLists.txt#L175 will not really help because we set `-Werror` in the CI right now https://github.com/acts-project/acts/blob/0b9ed368c93ea752ed1751e7fe7095af6ce61090/.github/workflows/builds.yml#L100

In my experience coupling different CMake projects always results in these kinds of problems.

I also don't like turning flags off which could just be absent in the first place but I do not see a good way around this.
1. Keep it as it is, which should only modify the flags for the current directory as far as I understood?
2. Add another cmake options to wire in flags which are then only used for ACTS and not the dependencies. This again makes me feel like we are working around the issue of coupling different CMake projects.

What do you think @krasznaa ?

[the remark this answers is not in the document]
paulgessinger: While I'm not a huge fan of the directory scoping, the benefit is that it's uniformly applied and can't be forgotten (which it WILL be if it can be). Similarly, `-DCMAKE_CXX_FLAGS=-Werror` is there in the CI because I absolutely do not want even a single warning in `main`, from dependencies or otherwise.

OTOH, I agree we probably shouldn't apply our build flags to internal builds of the dependencies. In any case, we would need to work around possible issues in dependency headers with `SYSTEM` includes. 

I think this issue shows again that we should refrain from including third-party code in our build directly with `FetchContent`, unless we basically know it's compatible with our build (like `nlohmann_json` and `pybind11` are, for example).

So I personally see two solutions that are different from the ones you mentioned already:

1. Remove GeoModel from `thirdparty` and rely on `find_package` (my preference)
2. Patch GeoModel to not produce warnings with our compile flags.

andiwand:
More than 10 files need patching so I am giving up on that. This is blocked until we have an external build of GeoModel in the CI

.github/workflows/builds.yml
```diff
@@ -352,7 +349,8 @@ jobs:
           -DCMAKE_PREFIX_PATH="${{ env.DEPENDENCY_DIR }}"
           -DPython_EXECUTABLE=${{ env.DEPENDENCY_DIR }}/bin/python3
           -DCMAKE_INSTALL_PREFIX="${{ env.INSTALL_DIR }}"
-          -DACTS_BUILD_EXAMPLES_PYTHON_BINDINGS=OFF
+          -DACTS_BUILD_PLUGIN_GEOMODEL=ON
+          -DACTS_BUILD_EXAMPLES_PYTHON_BINDINGS=ON
```
paulgessinger, line 353: I don't know why this was off, but I think I'd prefer if we continue testing this build on macOS as well.

andiwand: I think this should be enabled with the preset?

paulgessinger: I don't believe it is. The Linux jobs also enable this manually I believe.

AJPfleger (approved):

.github/workflows/builds.yml
andiwand: for GitHub Actions it should be `ON` by default https://github.com/acts-project/acts/blob/1d0a641dfa0a1f4ff3fb898df7c3b3d1642c18d3/CMakePresets.json#L19C18-L19C53

I guess it was switched `OFF` because we didn't built that before https://github.com/acts-project/acts/pull/3135 either

acts-project-service:
## :red_circle: Athena integration test results [1d0a641dfa0a1f4ff3fb898df7c3b3d1642c18d3]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42338374) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7933556)!

PR #3477, opened 2024-08-05
ci: Disable clang-tidy for external dependencies

stephenswat:
When using CMake's FetchContent mechanism, the source code for those dependencies ends up in the `build/_deps/` directory, which was previously subject to analysis by clang-tidy, which can lead to errors rooted in external projects. This commit disables clang-tidy for that directory, resolving the issue.

This commit comes from #3117.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -45,6 +45,7 @@ clang_tidy:
         clang-tidy/clang-tidy.json
         --exclude "*thirdparty*"
         --exclude "*ActsPodioEdm*"
+        --exclude "*build/_deps/*"
 
     # Check the combined report against the defined limits
     - CI/clang_tidy/check_clang_tidy.py --report clang-tidy/clang-tidy.json --config CI/clang_tidy/limits.yml


andiwand (approved):

andiwand (approved):
LGTM - the other way would be to tweak the thirdparty dependencies to be handled as an external

andiwand:
I think the system flag is too new for what we require as a minimum version. But I think it would already be enough to build the dependencies externally instead of basically doing `FetchContent` which will do `add_subdirectory`. Then we directly couple two CMake projects

stephenswat:
Yeah, FetchContent is a crime against programming; but don't we use it for other dependencies already?

andiwand:
Yeah sorry to open up that can here. No need to change anything on this PR. Just one more example why `FetchContent` is awful as I also rediscovered in the other PR as well

acts-project-service:
## :red_circle: Athena integration test results [9d520f26a2d487ea4b32358f3e94d7de17955b52]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42023839) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7882579)!

paulgessinger:
I like `FetchContent` much better than any alternative I've seen so far. 

I'd argue the problem isn't `FetchContent` but the complexity of our web of dependencies.

andiwand:
What I think is problematic about it is that we share variables transparently with another project. Calling external build with the flags we actually want to transmit would be much better in my opinion. If we want to download stuff or not is a different story. To me it just seems like we are using a build system as a package manager which it is not really built for

PR #3480, opened 2024-08-05
ci: Enable traccc and friends in CI builds

stephenswat:
This commit adds traccc, detray, and covfie plugins to the CMake CI preset so these plugins should automatically be tested in CI jobs.


stephenswat:
Closing as superceded by #3509.

PR #3481, opened 2024-08-06
fix: Return `nullptr` if outside tracking geometry in `TrackingGeometry::lowestTrackingVolume`

andiwand:
Currently `TrackingGeometry::lowestTrackingVolume` will always return at least the world volume even though the given position is not inside the world volume. In this PR I change the behavior to return `nullptr` in case we are outside the tracking geometry.

Same is done for `TrackingVolume::lowestTrackingVolume`. The rest are necessary fixes and downstream changes to make this work.

blocked by
- https://github.com/acts-project/acts/pull/3581


andiwand:
this also fails the gx2f unit tests 🙃 
```
09:08:42    Navigator      VERBOSE   No Volume | Initialization.
09:08:42    Navigator      VERBOSE   No Volume | Current surface set to start surface undefined
09:08:42    Navigator      VERBOSE   No Volume | Slow start initialization through search.
09:08:42    Navigator      VERBOSE   No Volume | Starting from position (750.0221, -749.9779, -0.1743) and direction (-0.2816, 0.9595, 0.0001)
unknown location(0): fatal error: in "Gx2fTest/UpdatePushedToNewVolume": memory access violation at address: 0x1a8: no mapping at fault address
```
I guess we are pushing out of the world volume @AJPfleger

Tests/UnitTests/Core/TrackFitting/Gx2fTests.cpp
```diff
@@ -291,16 +291,14 @@ std::shared_ptr<const TrackingGeometry> makeToyDetectorYdirection(
...
   volumeConfig2.position = {volumeConfig2.length.x(),
                             volumeConfig2.length.y() / 2, 0.};
   volumeConfig2.name = "AdditionalVolume";
   volumeConfig2.binningDimension = Acts::BinningValue::binY;
 
   // Outer volume - Build TrackingGeometry configuration and fill
   CuboidVolumeBuilder::Config config;
-  config.length = {4 * halfSizeSurface, (nSurfaces + 1) * 1_m,
-                   2 * halfSizeSurface};
+  config.length = {10_m, 10_m, 10_m};
```
AJPfleger, line 301: I think we should make the size variable. If we would call `makeToyDetectorYdirection(...)` with `nSurfaces > 10`, not all of them would be covered.

andiwand: I reverted this

AJPfleger (dismissed):

acts-project-service:
## :red_circle: Athena integration test results [40ad6793c243c55820f755f66c643b115be6ff0d]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42074695) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/7890374)!

andiwand:
Looks like Athena is also relying on the faulty behavior

paulgessinger:
```
21:08:22 ActsCombinedTrackToTrackPa...ActsExtrapolationTool  ERROR Got error during propagation:SurfaceError:1. Returning empty step vector.
```

I think we use the extrapolation tool in some spots to to pre-checks, I don't think this is called from any of the core algorithms themselves. @goetzgaycken @timadye

goetzgaycken:
> ```
> 21:08:22 ActsCombinedTrackToTrackPa...ActsExtrapolationTool  ERROR Got error during propagation:SurfaceError:1. Returning empty step vector.
> ```
> 
> I think we use the extrapolation tool in some spots to to pre-checks, I don't think this is called from any of the core algorithms themselves. @goetzgaycken @timadye

The particle creator for acts tracks uses the acts extrapolation tool to check for expected innermost pixel hits.

paulgessinger:
@goetzgaycken So apparently sometimes it starts outside the tracking volume with this extrapolation. I guess that could become an explicit check before calling the tool?

goetzgaycken:
> @goetzgaycken So apparently sometimes it starts outside the tracking volume with this extrapolation. I guess that could become an explicit check before calling the tool?

As far as I can see the extrapolation start is the perigee surface. Needs to be looked at in more detail.

AJPfleger (approved):

PR #3482, opened 2024-08-06
fix!: Use `ActsScalar` in binning tools

stephenswat:
The current binning tools use floats everywhere, which is prone to conversion warnings. This commit converts that code to use the `ActsScalar` type alias instead.


cmake/ActsCompilerOptions.cmake
```diff
@@ -4,7 +4,7 @@ if (NOT CMAKE_BUILD_TYPE)
   message(STATUS "Setting default build type: ${CMAKE_BUILD_TYPE}")
 endif()
 
-set(cxx_flags "-Wall -Wextra -Wpedantic -Wshadow -Wno-unused-local-typedefs")
+set(cxx_flags "-Wall -Wextra -Wpedantic -Wshadow -Wno-unused-local-typedefs -fconcepts-diagnostics-depth=10")
```
paulgessinger, line 7: Unrelated change?

paulgessinger (commented):
Looks good overall, waiting for v37 to merge this.

cmake/ActsCompilerOptions.cmake
stephenswat: Whoooops.

stephenswat: Fixed!

andiwand:
always interesting to see what triggers the 3 ttbar events to change

AJPfleger:
I was looking at the physmon output. Some of the plots have random spikes. The outliers in `CKF | trackfinding | ttbar with 200 pileup | default seeding` change a lot.
image

The use of `double` instead of `float` seems correct, but I think we should investigate the underlying cause first before getting this in. What do you think?

stephenswat:
Huh, that's peculiar. :thinking:

AJPfleger:
Regarding the IF, I have no strong opinion.

It's annoying for the low stats, but maybe we should re-run with better stats. I will check the logs, how expensive the individual tests are, but if we are lucky we could upscale them at a cost of an extra minute or so :3

paulgessinger:
Why this affects the seeding is not entirely clear to me. 

I guess it could affect Fatras and then subsequently reconstruction.

andiwand:
> Why this affects the seeding is not entirely clear to me.
> 
> I guess it could affect Fatras and then subsequently reconstruction.

Isn't
- `Core/include/Acts/Utilities/BinUtility.hpp`
- `Core/include/Acts/Utilities/BinningData.hpp`
something that is used in seeding?

> The current physmon tests take around 11-18 minutes. In the `trackfinding_ttbar_pu200` we spend around 75-135 seconds. I think, we could double the number of events from 3 to 6, but everything else would take too much time. I tried it with 30 events and it takes 850s on github. As expected, we scale here linearly.

I think we would need 100-1000 events so I would rather keep it as it is for now if we cannot reach that

paulgessinger:
I was under the impression it uses the Grid for the internal binning, but I could be wrong.

CarloVarni:
In the seeding we use the `Acts::Grid`, but I'm not familiar with neither of `Core/include/Acts/Utilities/BinUtility.hpp` or `Core/include/Acts/Utilities/BinningData.hpp` and I don't know if they may be use internally somewhere

stephenswat:
Invalidated by #3873.

AJPfleger:
we could still have this with `double` instead of the `float`s, no?

PR #3490, opened 2024-08-06
ci: Upgrade doxygen version on RTD

paulgessinger:
This is an attempt to upgrade the Doxygen version on ReadTheDocs.


andiwand (approved):

paulgessinger:
This doesn't seem to work unfortunately.

PR #3491, opened 2024-08-07
draft: Gauss vertex seeding with time

CENcll:
draft


Core/include/Acts/Vertexing/GaussianTrackDensity.hpp
```diff
@@ -34,16 +34,20 @@ class GaussianTrackDensity {
     /// @param c2_ Quadratic coefficient in exponent
     /// @param lowerBound_ The lower bound
     /// @param upperBound_ The upper bound
+    /// @param time_ Time component 
     TrackEntry(double z_, double c0_, double c1_, double c2_,
-               double lowerBound_, double upperBound_)
+               double lowerBound_, double upperBound_,double time_)
```
andiwand, line 39: To me it looks a bit like we would need more coefficients by adding time since there are `z` coefficients

Core/include/Acts/Vertexing/GaussianTrackDensity.hpp
```diff
@@ -59,11 +63,12 @@ class GaussianTrackDensity {
...
   /// @brief The Config struct
   struct Config {
-    Config(double d0Sig = 3.5, double z0Sig = 12.)
+    Config(double d0Sig = 3.5, double z0Sig = 12.,double tScale = 1.)
         : d0MaxSignificance(d0Sig),
           z0MaxSignificance(z0Sig),
           d0SignificanceCut(d0Sig * d0Sig),
-          z0SignificanceCut(z0Sig * z0Sig) {}
+          z0SignificanceCut(z0Sig * z0Sig),
+          timeScale(tScale) {}
```
andiwand, line 71: I think `tScale` should be called `t0Sig`

Core/include/Acts/Vertexing/GaussianTrackDensity.hpp
```diff
@@ -116,7 +124,7 @@ class GaussianTrackDensity {
   ///
   /// @return Pair of position of global maximum and Gaussian width
   Result<std::optional<std::pair<double, double>>> globalMaximumWithWidth(
-      State& state, const std::vector<InputTrack>& trackList) const;
+      State& state, const std::vector<InputTrack>& trackList,double time) const; //do we need time here?
```
andiwand, line 127: ad comment: I don't think `time` is necessary here. But we will need to change the return type to capture `z` and `t`. Similar to this https://github.com/acts-project/acts/blob/00d84eefbb66ee9d63865e7beb72aaf22689d622/Core/include/Acts/Vertexing/AdaptiveGridTrackDensity.hpp#L102-L111

Core/include/Acts/Vertexing/GaussianTrackDensity.hpp
```diff
@@ -125,7 +133,7 @@ class GaussianTrackDensity {
   ///
   /// @return z position of the global maximum
   Result<std::optional<double>> globalMaximum(
-      State& state, const std::vector<InputTrack>& trackList) const;
+      State& state, const std::vector<InputTrack>& trackList,double time) const;
```
andiwand, line 136: Same as above - I think we need a different return type and `time` is part of the input tracks

Core/include/Acts/Vertexing/GaussianTrackDensity.hpp
```diff
@@ -146,7 +154,7 @@ class GaussianTrackDensity {
   ///
   /// @return Track density, first and second derivatives
   std::tuple<double, double, double> trackDensityAndDerivatives(State& state,
-                                                                double z) const;
+                                                                double z,double time) const;
```
andiwand, line 157: I think we should switch to an `Eigen` type here but not crucial. Apart from that we will need a matrix or a bigger array to store the new derivatives. Right now we have `density, d(density)/dz, d^2(density)/dz^2` but we will also need the `t` terms and the mixed term with `dzdt`

Core/include/Acts/Vertexing/GaussianTrackDensity.hpp
```diff
@@ -171,11 +179,13 @@ class GaussianTrackDensity {
...
   double stepSize(double y, double dy, double ddy) const;
 
+
+
   // Helper class to evaluate and store track density at specific position
   class GaussianTrackDensityStore {
    public:
     // Initialise at the z coordinate at which the density is to be evaluated
-    GaussianTrackDensityStore(double z_coordinate) : m_z(z_coordinate) {}
+    GaussianTrackDensityStore(double z_coordinate, double t_coordinate, double time_scale) : m_z(z_coordinate), m_time(t_coordinate), m_timeScale(time_scale) {}
```
andiwand, line 188: To me it looks like `time_scale` should not be necessary

PR #3493, opened 2024-08-07
feat: allow bare Surface pointers to be written to JSON

asalzburger:
This PR rennames `geomodel.py` to `geometry_geomodel.py` and adds the necessary `json` update to write out the sensitive surfaces as a hierarchy map.

## Release Notes

- **New Features**
  - Introduced functionality to write sensitive detector maps to JSON files.
  - Enhanced command-line options for outputting geometry data in JSON format.
  - Added a new method for converting surface pointers to JSON.

- **Bug Fixes**
  - Improved handling of JSON serialization for `Surface` objects.

- **Tests**
  - Added a new test case for validating the serialization of `CylinderSurface` objects using plain pointer references.


paulgessinger:
@andiwand can you update the PR title to describe the change? What does it add JSON writing functionality to?

Plugins/Json/src/SurfaceJsonConverter.cpp
```diff
@@ -42,6 +42,11 @@ void Acts::to_json(nlohmann::json& j, const Acts::Surface& surface) {
   j = SurfaceJsonConverter::toJson(gctx, surface);
 }
 
+void Acts::to_json(nlohmann::json& j, const Acts::Surface* surface) {
+  Acts::GeometryContext gctx;
+  j = SurfaceJsonConverter::toJson(gctx, *surface);
```
paulgessinger, line 47: This needs to check if the surface does not have an associated detector element, otherwise, this is invalid.

asalzburger: Mmm, we have the same functionality for `std::shared_ptr<Surface>` - this just mirrors it for plain pointers, shall I change that then as well?

I guess we could throw and exception? That the default context is used is already pointed out in the comments.

This is needed because the `GeometryIdHierarchyMap` converter relies on the `nlohmann::json` bare `to_json` implementation.

asalzburger:
> @asalzburger can you update the PR title to describe the change? What does it add JSON writing functionality to?

Sure, sorry - this was the last commit message - which didn't make much sense either.

Plugins/Json/src/SurfaceJsonConverter.cpp
paulgessinger: Didn't we add a way to serialize the surface with an explicit context? The problem with a default constructed context is that it's up to the detector element how it's handled. 

Sort of fundamentally, surfaces with detector elements can't really be serialized by writing out the transform, since it's only a snapshot. For the podio Track edm, I explicitly use the identifier of the surface if it has one, so it can be tied back to the detector element.

PR #3499, opened 2024-08-08
feat: GeoModelDetectorObjectFactory for flexible conversion of GeoModel objects to Acts Surfaces and Volumes

Berggren-Jonas:
This PR adds a `GeoModelDetectorObjectFacoty` that converts a variety of GeoModel elements to Acts elements in a flexible manner.

- For the interface with the database and the .obj conversion I added a `geomodel_vol.py`. It takes a number of optional parameters to select what and how to convert.

`python3 geomodel_vol.py -i ATLAS-R3-MUONTEST_v3.db -q Pixel SCT  --output-obj -o id -ml Silicon` converts the `FullPhysVol` elements of the ID to `Acts::SensitiveSurface`
converted_id00

`python3 geomodel_vol.py -i ATLAS-R3-MUONTEST_v3.db -q Muon --output-obj -o muons -ml RPCgas Aluminium --convert-boundingboxes MDT -n RpcGasGap Tube BML1_RPC11_1_3_1_1_2 BML1_MDT03_1_3_1` converts the gas gaps of the specified RPC to `Acts::SensitivSurface` elements. In addition it converts the `FullPhysVol` of the specified MDT to `Acts::Experimental::DetectorVolume` containing the tubes as `Acts::Surface` elements.
rpc_mdt00

- In addition I added a simple unit test that checks the conversion of a GeoBox containing a Tube and a an other GeoBox as subvolumes.
"@wip ready for review"


Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorObjectFactory.hpp
```diff
@@ -0,0 +1,78 @@
...
+    // /// List for names to match
+    std::vector<std::string> nameList;
+    /// List for materials to match
+    std::vector<std::string> materialList;
+
+    /// boolean flag to build subvolumes
+    bool convertSubVolumes = false;
+
+    /// flag to build the desired bounding Boxes
+    std::string convertBox;
```
dimitra97, line 43: we want this to be just  a string? I would think it as a vector of strings?

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorObjectFactory.hpp
```diff
@@ -0,0 +1,78 @@
...
+  void construct(Cache& cache, const GeometryContext& gctx,
+                 const GeoModelTree& geoModelTree, const Options& options);
+
+  void convertSensitive(PVConstLink geoPV, const Acts::Transform3& transform,
+                        std::vector<GeoModelSensitiveSurface>& sensitives);
+
+  std::vector<GeoChildNodeWithTrf> findAllSubVolumes(PVConstLink geoPV);
+
+  bool convertBox(std::string name);
+  bool matches(const std::string& name, PVConstLink physvol);
```
dimitra97, line 66: can we have this function as lamda as it was before?

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorObjectFactory.hpp
```diff
@@ -0,0 +1,78 @@
...
+  };
+
+  GeoModelDetectorObjectFactory(
+      const Config& cfg,
+      std::unique_ptr<const Logger> mlogger = getDefaultLogger(
+          "GeoModelDetectorObjectFactory", Acts::Logging::WARNING));
+
+  void construct(Cache& cache, const GeometryContext& gctx,
+                 const GeoModelTree& geoModelTree, const Options& options);
+
```
dimitra97, line 59: I guess from this function and below, they can be private member functions

paulgessinger:
Hey @Berggren-Jonas, this has picked up a small conflict now. Can you resolve it?
I'll look at the changeset later today!

Plugins/GeoModel/src/GeoModelDetectorObjectFactory.cpp
```diff
@@ -0,0 +1,215 @@
...
+}
+void Acts::GeoModelDetectorObjectFactory::convertFpv(
+    std::string name, GeoFullPhysVol *fpv, Cache &cache,
+    const GeometryContext &gctx) {
+  PVConstLink physVol{fpv};
+
+  // get children
+  std::vector<GeoChildNodeWithTrf> subvolumes =
+      getChildrenWithRef(physVol, false);
+  if (subvolumes.size() > 0) {
```
dimitra97, line 153: ok so this means that we convert always the children as long as they exist right? Is this working for all the cases? (MDTs, RPCs etc and also the ID) Because I was thinking of also checking with the matching function for the children..

Plugins/GeoModel/src/GeoModelDetectorObjectFactory.cpp
```diff
@@ -0,0 +1,215 @@
...
+    std::vector<GeoModelSensitiveSurface> sensitives;
+
+    for (auto surface : surfaces) {
+      const Transform3 &transform =
+          fpv->getAbsoluteTransform() * surface.transform;
+      convertSensitive(surface.volume, transform, sensitives);
+    }
+    cache.sensitiveSurfaces.insert(cache.sensitiveSurfaces.end(),
+                                   sensitives.begin(), sensitives.end());
+    if (convertBox(name)) {
```
dimitra97, line 165: cool!!

dimitra97 (commented):
Happy to see this PR here! As I see it sits on top of mine and Matthew's that we did for the Muons. 
I have added some comments.
 Also, one general question from my side: We keep bot the GeomodelSurfaceFactory and GeomodelObjectFactory and the GeomodelToDetectorVolume? It seems to me that these factories share the same functionality and I remember discussing with you that maybe the surface factory would be obsolent, right? Let's see which approach we should follow

Examples/Python/src/Obj.cpp
```diff
@@ -47,6 +48,47 @@ void addObj(Context& ctx) {
...
+
+              for (const auto& volume : Volumes) {
+                view3D.drawDetectorVolume(obj, *volume, viewContext,
+                                          Acts::Transform3::Identity(),
+                                          sConfig);
+              }
+              obj.write(fileName);
+            });
+    mex.def("writeVolumesSurfacesObj",
+            [](const std::vector<std::shared_ptr<Surface>>& surfaces,
```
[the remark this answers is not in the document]
Berggren-Jonas: I think this is resolved by caching the surfaces even if they are contained by the volumes. My guess as to why this is necessary, is that something in the obj conversion works by reference and somehow relies on the initial memory address of the surface. But ether way while converting all MDTs with the tubes at once it doesn't seem to make a difference memory wise whether or not I cache the surfaces.

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorObjectFactory.hpp
Berggren-Jonas: I fixed it

Examples/Scripts/Python/ATLAS-R3-MUONTEST_v3.db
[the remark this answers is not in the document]
Berggren-Jonas: I took it out

CI/setup_cvmfs_lcg105.sh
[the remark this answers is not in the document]
Berggren-Jonas: I forgot to delete the GeoModelDetectorSurfaceFactory files. This one is indeed obsolete. The DetObjectFactory depends of ToDetVol to call the right converter so we should keep that.

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorObjectFactory.hpp
Berggren-Jonas: I'm not sure since we call from both the `construct` and the `findAllSubVolumes` member functions. I could pass it as an argument to the `convertFpv` to have it then passed down to the `findAllSubVolumes`but that seems unnecessarily complicated to me. Or is there a  benefit in defining it as it was in the surfaceFactory?

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorObjectFactory.hpp
Berggren-Jonas: What functions do you mean? `construct` has to be public since this is doing the conversion that is being requested from outside.

Plugins/GeoModel/src/GeoModelDetectorObjectFactory.cpp
Berggren-Jonas: The routine for finding the children is called whenever they exist, but if none of the children names and materials match the config.nameList and config.materialList then nothing will be converted.

Plugins/GeoModel/src/GeoModelDetectorObjectFactory.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+void Acts::GeoModelDetectorObjectFactory::convertSensitive(
+    PVConstLink geoPV, const Acts::Transform3 &transform,
+    std::vector<GeoModelSensitiveSurface> &sensitives) {
+  const GeoLogVol *logVol = geoPV->getLogVol();
+  const GeoShape *shape = logVol->getShape();
+  int shapeId = shape->typeID();
+  std::string name = logVol->getName();
+  std::shared_ptr<const Acts::IGeoShapeConverter> converter =
+      Acts::GeoShapesConverters(shapeId);
```
paulgessinger, line 96:
```suggestion
  const Acts::IGeoShapeConverter& converter =
      Acts::geoShapesConverters(shapeId);
```

Plugins/GeoModel/src/GeoModelDetectorObjectFactory.cpp
```diff
@@ -0,0 +1,219 @@
...
+    sensitives.push_back(converted.value());
+    const auto &[el, sf] = converted.value();
+    auto ss = std::get<1>(converted.value());
+
+    ACTS_VERBOSE("(successfully converted: "
+                 << name << " / " << recType(*shape) << " / "
+                 << logVol->getMaterial()->getName() << ")");
+
+    if (!el || !sf) {
+      throw std::runtime_error("The Detector Element or the Surface is nllptr");
```
paulgessinger, line 114:
```suggestion
      throw std::runtime_error("The Detector Element or the Surface is nullptr");
```

Plugins/GeoModel/src/GeoModelDetectorObjectFactory.cpp
```diff
@@ -0,0 +1,219 @@
...
+    // convert fpvs to surfaces
+    const Transform3 &transform = fpv->getAbsoluteTransform();
+    convertSensitive(fpv, transform, cache.sensitiveSurfaces);
+  }
+}
+// lambda to determine if object fits query
+bool Acts::GeoModelDetectorObjectFactory::matches(const std::string &name,
+                                                  PVConstLink physvol) {
+  if (m_cfg.nameList.empty() && m_cfg.materialList.empty()) {
+    return true;
```
paulgessinger, line 191: Shouldn't this be `false`? If there's no matching, we take everything?

Plugins/GeoModel/src/GeoModelToDetectorVolume.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include <GeoModelKernel/GeoTrd.h>
+#include <GeoModelKernel/GeoTube.h>
+#include <GeoModelKernel/GeoTubs.h>
+
+namespace Acts {
+namespace GeoModel {
+std::shared_ptr<Experimental::DetectorVolume> convertVolume(
+    const GeometryContext& context, const GeoShape* shape,
+    const std::string& name, const GeoTrf::Transform3D transform,
+    std::vector<GeoModelSensitiveSurface> sensitives) {
```
paulgessinger, line 36: Is the copy intentional here?

Plugins/GeoModel/src/GeoModelToDetectorVolume.cpp
```diff
@@ -0,0 +1,186 @@
...
+  // dummy volume for conversion with surfaces
+  std::vector<std::shared_ptr<Acts::Experimental::DetectorVolume>> a;
+
+  // type conversion from GeoModelSensitiveSurface to Surface
+  std::vector<std::shared_ptr<Surface>> sensSurfaces(sensitives.size());
+  std::transform(sensitives.begin(), sensitives.end(), sensSurfaces.begin(),
+                 [](const std::tuple<std::shared_ptr<GeoModelDetectorElement>,
+                                     std::shared_ptr<Surface>>& t) {
+                   return std::get<1>(t);
+                 });
```
paulgessinger, line 46: This dance seems awkward (see my other comment on why we use pairs here.)

Plugins/GeoModel/src/GeoModelToDetectorVolume.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include <GeoModelKernel/GeoShapeSubtraction.h>
+#include <GeoModelKernel/GeoShapeUnion.h>
+#include <GeoModelKernel/GeoSimplePolygonBrep.h>
+#include <GeoModelKernel/GeoTrd.h>
+#include <GeoModelKernel/GeoTube.h>
+#include <GeoModelKernel/GeoTubs.h>
+
+namespace Acts {
+namespace GeoModel {
+std::shared_ptr<Experimental::DetectorVolume> convertVolume(
```
paulgessinger, line 33: Is it in-scope to factorize these functions to produce the `VolumeBounds` separately from the detector volumes? 
Ideally, and looking forward to Gen3, I think the best approach would be for these functions to produce standard `Acts::Volume`s, instead of `DetectorVolumes`. 
The Portal generation and navigation delegates are hard-coded anyway, so the conversion from `Acts::Volume` to `DetectorVolume` should be easy. This would allow this code to continue to be useful, rather than being superceded in the near future.

Tests/UnitTests/Core/EventData/CMakeLists.txt
```diff
@@ -7,8 +7,8 @@ add_unittest(MultiComponentBoundTrackParameters MultiComponentBoundTrackParamete
 add_unittest(MultiTrajectory MultiTrajectoryTests.cpp)
 add_unittest(TransformHelpers TransformHelpersTests.cpp)
 add_unittest(CorrectedTransformFreeToBound CorrectedTransformFreeToBoundTests.cpp)
-add_unittest(Track TrackTests.cpp)
-target_sources(ActsUnitTestTrack PUBLIC TrackTestsExtra.cpp)
+#add_unittest(Track TrackTests.cpp)
+#target_sources(ActsUnitTestTrack PUBLIC TrackTestsExtra.cpp)
```
paulgessinger, line 11:
```suggestion
add_unittest(Track TrackTests.cpp)
target_sources(ActsUnitTestTrack PUBLIC TrackTestsExtra.cpp)
```

Tests/UnitTests/Plugins/GeoModel/GeoDetectorObjectTest.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+  // create pars for conversion
+  Acts::GeoModelDetectorObjectFactory::Config gmConfig;
+  gmConfig.convertBox = {"LogVolumeXY"};
+  Acts::GeometryContext gContext;
+  Acts::GeoModelDetectorObjectFactory::Cache gmCache;
+
+  // create factory instance
+  Acts::GeoModelDetectorObjectFactory factory =
+      Acts::GeoModelDetectorObjectFactory(gmConfig);
```
paulgessinger, line 69:
```suggestion
  Acts::GeoModelDetectorObjectFactory factory(gmConfig);
```

paulgessinger (commented):
General comment: in cases where we don't need to take ownership, I don't think it's necessary to pass around `PVConstLink`s and related smart pointers. I guess the main reason is that we want to use it to retain ownership of the PVs in the detector elements?

CMakeLists.txt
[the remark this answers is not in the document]
Berggren-Jonas: done

CI/setup_cvmfs_lcg105.sh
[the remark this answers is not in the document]
Berggren-Jonas: on my local version it isn't, so I don't know why that would be

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorElement.hpp
```diff
@@ -15,7 +15,9 @@
 
 #include <memory>
 
-class GeoFullPhysVol;
+#include <GeoModelKernel/GeoFullPhysVol.h>
```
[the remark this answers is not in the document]
Berggren-Jonas: For the PVConstlink to work apparently

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorObjectFactory.hpp
```diff
@@ -0,0 +1,78 @@
...
+struct GeoModelTree;
+class Surface;
+namespace Acts {
+class GeoModelDetectorObjectFactory {
+ public:
+  using GeoModelBoundingBox = std::shared_ptr<Experimental::DetectorVolume>;
+
+  struct Options {
+    std::vector<std::string> queries = {};
+  };
```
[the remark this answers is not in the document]
Berggren-Jonas: I took that from the SurfaceFctory Dimitra did. I was also wondering why we need a struct for that. She said it is for some extensibility reason. Maybe you can add some information on that @dimitra97 ? Otherwise I am fine simplifying it.

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorObjectFactory.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+    /// boolean flag to build subvolumes
+    bool convertSubVolumes = false;
+
+    /// flag to build the desired bounding Boxes
+    std::vector<std::string> convertBox;
+  };
+  struct Cache {
+    /// The created detector elements and their surfaces
+    std::vector<GeoModelSensitiveSurface> sensitiveSurfaces;
```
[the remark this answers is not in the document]
Berggren-Jonas: I think the goal is to convert the FullPhysVols to Experimental::DetectorVolume containing the surfaces where ever bounding boxes are needed. Otherwise I considered the subvolumes being converted to SensitiveSurface as more of an artifact of the converter then anything else. So as far as I can tell there is no need to use SensitiveSurface, I just did it because this is the return type of the surface converter

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorObjectFactory.hpp
```diff
@@ -0,0 +1,78 @@
...
+class GeoModelDetectorObjectFactory {
+ public:
+  using GeoModelBoundingBox = std::shared_ptr<Experimental::DetectorVolume>;
+
+  struct Options {
+    std::vector<std::string> queries = {};
+  };
+  struct Config {
+    // /// List for names to match
+    std::vector<std::string> nameList;
```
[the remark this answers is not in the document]
Berggren-Jonas: What level of detail do you have in mind?

Plugins/GeoModel/src/GeoModelDetectorObjectFactory.cpp
Berggren-Jonas: The idea is that if nothing is passed we assume that everything should be converted so I think this is correct.

Plugins/GeoModel/src/GeoModelToDetectorVolume.cpp
Berggren-Jonas: what do you mean by copy?

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelConverters.hpp
```diff
@@ -68,4 +69,25 @@ using GeoUnionDoubleTrdConverter =
     detail::GenericGeoShapeConverter<GeoShapeUnion,
                                      detail::GeoUnionDoubleTrdConverter>;
 
+/// @brief The map that maps the converters with the shapes
+
+inline std::shared_ptr<const IGeoShapeConverter> GeoShapesConverters(
```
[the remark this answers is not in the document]
Berggren-Jonas: I made the code change but  I don't really get what you are saying with the rest.

Plugins/GeoModel/include/Acts/Plugins/GeoModel/detail/GenericGeoShapeConverter.hpp
```diff
@@ -20,15 +20,14 @@ namespace Acts::detail {
 template <typename Shape, typename Converter>
 struct GenericGeoShapeConverter : public IGeoShapeConverter {
   Acts::Result<Acts::GeoModelSensitiveSurface> toSensitiveSurface(
-      const GeoFullPhysVol& geoFPV) const override {
+      PVConstLink geoPV, const Transform3& transform) const override {
     // Retrieve logical volume and absolute transform
-    const GeoLogVol* logVol = geoFPV.getLogVol();
-    const Transform3& transform = geoFPV.getAbsoluteTransform(nullptr);
+    const GeoLogVol* logVol = geoPV->getLogVol();
```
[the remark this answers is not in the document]
Berggren-Jonas: What do you mean by checked?

Plugins/GeoModel/src/GeoModelToDetectorVolume.cpp
Berggren-Jonas: The (verry convenient) way in which I did the conversion is to pass the GeoShape to the ToDetecotorVolume which passes it to a different converter. Should I then change all of that to get Acts::Volume as return? Can/should Acts::Volume also contain the surfaces?

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorObjectFactory.hpp
dimitra97: Yes I mean the others below the construct

Examples/Python/src/Obj.cpp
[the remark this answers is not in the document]
Berggren-Jonas: Talking about it with @paulgessinger we concluded that this is a deep rooted "issue" which is not really worth addressing right now since it doesn't appear to cost much (if anything).

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorObjectFactory.hpp
Berggren-Jonas: The `convertSensitive` and `converFpv` also need to be public for the unit test or for anyone wanting to convert stuff without going through a database. The rest can be private i think even though I am not sure of the advantage this would give.

Plugins/GeoModel/src/GeoModelToDetectorVolume.cpp
Berggren-Jonas: I would prefer to do that in a future refactor pr

CMakeLists.txt
paulgessinger, line 1: I think these changes need to be reverted.

CI/setup_cvmfs_lcg105.sh
paulgessinger, line 1: This file is still present somehow.

paulgessinger (approved):

andiwand:
I think there are a bunch of `clang-tidy` failures on main after this one @Berggren-Jonas @paulgessinger

paulgessinger:
@andiwand my bad, fixed in separate PR.

PR #3501, opened 2024-08-09
feat: Gen3 geometry Portals

paulgessinger:
Part of #3502

This PR implements *Portals*. A portal connects two or more neighboring volumes. Each volume has a set of portals that describes which volumes lie behind the portal in that direction. Portals use associated portal links to perform lookups of target volumes. Each portal has two links, and a corresponding surface. One link is associated with the direction along the surface's normal vector, and one with the opposite direction.

Portals can be **fused** and **merged**. 

**Fusing** is the combination of two portal linkson the same logical surfaces. The actual surface instances can be different, as long as they are geometrically equivalent (within numerical precistion). The resulting portal will have one portal along the shared surface's normal vector, and one opposite that vector.

```
   portal1   portal2
     +---+   +---+
     |   |   |   |
     |   |   |   |
<----+   | + |   +---->
     |   |   |   |
     |   |   |   |
     +---+   +---+
```

The input portals need to have compatible link loadout, e.g. one portal needs to have the *along normal* slot filled, and the other one needs to have the *opposite normal* slot filled. If
portals share a filled slot, the function throws an exception.

**Merging** is the complementary operation to the fusing of portals. To be able to merge portals, the surfaces of their associated links need to be *mergeable*, and the portal links need to be compatible. This means that both portals need to have a link along the portal surface normal, opposite the normal, or both. If the equipped links are opposite relative to one another (e.g. one along one opposite), the function will throw an exception.

```
           ^                     ^
           |                     |
    portal1|              portal2|
   +-------+-------+     +-------+-------+
   |               |  +  |               |
   +-------+-------+     +-------+-------+
           |                     |
           |                     |
           v                     v
```

This is a destructive operation on both portals, their links will be moved to produce merged links, which can fail if the portal links are not compatible


Core/include/Acts/Geometry/Portal.hpp
```diff
@@ -0,0 +1,234 @@
...
+  /// Constructor that takes a geometry context and an rvalue reference to a
+  /// helper config struct from above. This pattern allows you to use designated
+  /// initializers to construct this object like:
+  /// ```cpp
+  /// Portal{gctx, {.oppositeNormal = {cyl1, *vol1}}};
+  /// Portal{gctx, {.alongNormal = {cyl2, *vol2}}};
+  /// ```
+  /// @param gctx The geometry context
+  /// @param config The configuration struct
+  Portal(const GeometryContext& gctx, Config&& config);
```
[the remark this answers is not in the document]
andiwand: Alternatively we could have a boolean and/or factory methods to express the different construction ways explicitly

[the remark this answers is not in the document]
paulgessinger: This is definitely expressible in a different way, like for example the other constructor which takes two finished links. I was just thinking that this mechanism here, where you have two optional `kwarg` style arguments looks *nice* with designated initializers.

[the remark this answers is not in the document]
benjaminhuth: This designated initializer thing should also work with normal const lvalue reference instead of rvalue referance, is there a specific reason you chose rvalue reference here?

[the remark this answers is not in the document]
paulgessinger: My idea was to use the `struct` only for arguments to this function, `kwargs`-style. The rvalue ref sort of encourages this I think, but I can also change it of course.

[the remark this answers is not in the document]
benjaminhuth: Ah I see... I mean one can of course always `std::move` the struct into the function if one does'nt like the kwargs-style...
But on the other hand its kind of our convention to have the config struct with const lvalue ref. I definitvely like the `kwargs` approach, but I'm not sure if we should break the convention for that... But up to you.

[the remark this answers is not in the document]
andiwand: I see - to me taking two `Link` params seems a bit simpler and does not trade much in "expressiveness" for modern editors. But I have no strong opinion for or against this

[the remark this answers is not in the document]
paulgessinger: Renamed `Config` here to `Arguments` to underline the purpose.

Core/include/Acts/Geometry/Portal.hpp
```diff
@@ -0,0 +1,234 @@
...
+  /// is needed because we allow fusing portals with surfaces that are
+  /// equivalent but one of them has material while the other does not. The
+  /// normal surface comparison would determine these surfaces as not
+  /// equivalent.
+  /// @param gctx The geometry context
+  /// @param a The first surface
+  /// @param b The second surface
+  /// @return True if the surfaces are equivalent
+  static bool isSameSurface(const GeometryContext& gctx, const Surface& a,
+                            const Surface& b);
```
andiwand, line 226: I wonder if this should live in the `Surface` class / header. Potentially such an equality check could be useful to other components

Core/include/Acts/Geometry/Portal.hpp
```diff
@@ -0,0 +1,234 @@
...
+  Portal(Direction direction, std::shared_ptr<RegularSurface> surface,
+         TrackingVolume& volume);
+
+  /// Constructor for a portal from two links. One of the links can be
+  /// `nullptr`, but at least one of them needs to be set. If both are set, they
+  /// need to be valid compatible links that can be fused.
+  /// @param gctx The geometry context
+  /// @param alongNormal The link along the normal of the surface
+  /// @param oppositeNormal The link opposite to the normal of the
+  Portal(const GeometryContext& gctx,
```
andiwand, line 74: It's a bit confusing that we have to wire the geometry context in here but I think it is the right thing to do for now

Core/include/Acts/Geometry/Portal.hpp
```diff
@@ -0,0 +1,234 @@
...
+    struct Link {
+      Link() = default;
+      /// Constructor from a surface and a volume
+      Link(std::shared_ptr<RegularSurface> surface, TrackingVolume& volume)
+          : m_surface(std::move(surface)), m_volume(&volume) {}
+
+      /// The associated surface
+      std::shared_ptr<RegularSurface> m_surface = nullptr;
+      /// The associated volume
+      TrackingVolume* m_volume = nullptr;
```
andiwand, line 91: nitpick: for plain public struct members we usually do `surface` instead of `m_surface`

Core/src/Geometry/Portal.cpp
```diff
@@ -0,0 +1,306 @@
...
+    return m_oppositeNormal.get();
+  }
+}
+
+Result<const TrackingVolume*> Portal::resolveVolume(
+    const GeometryContext& gctx, const Vector3& position,
+    const Vector3& direction) const {
+  assert(m_surface != nullptr);
+  const Vector3 normal = m_surface->normal(gctx, position);
+  Direction side = Direction::fromScalar(normal.dot(direction));
```
andiwand, line 139: I think this asserts on the parameter not being `0` - is that wanted here?

Core/src/Geometry/Portal.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+  if (a.type() != b.type()) {
+    return false;
+  }
+
+  if (a.bounds() != b.bounds()) {
+    return false;
+  }
+
+  if (!a.transform(gctx).isApprox(b.transform(gctx), 1e-9)) {
```
andiwand, line 299: I wonder if we should rather use `==` or a common global epsilon. Do you have a case in mind where the transform might be numerically altered compared to the other one?

andiwand (dismissed):
A couple of comments and questions - nothing critical

Core/include/Acts/Geometry/Portal.hpp
paulgessinger: We do have an equality operator, but that also checks if both surfaces have material. In this case, I explicitly want to allow the case where only one surface has material. In fact, the class will refuse fusing portals that both have material already.

Core/include/Acts/Geometry/Portal.hpp
paulgessinger: Yeah, it's a consequence of how we decided to do the transform lookups further down. With that policy in place, I think this is the only way to do it.

Core/src/Geometry/Portal.cpp
paulgessinger: Do you mean if `direction` and  the `normal` are perpendicular? I guess that's indeed an issue, we have no idea which volume to look up in that case.

Core/src/Geometry/Portal.cpp
paulgessinger: I don't think we should use `==`, because if volumes are being connected that come from size harmonization, I'm not willing to guarantee that they have **exactly** the same bounds and hence boundary surfaces.

I can make the `10e-9` a named variable, if you think that's more clear.

andiwand: I think it is ok as it is. I just wondered if we want to come up with a global constant for this at some point in case we hard code different magic numbers for the same purpose in different places. Similar to `s_onSurfaceTolerance`

andiwand (dismissed):
LGTM 👍

Core/src/Geometry/Portal.cpp
paulgessinger: Changed to that.

Core/src/Geometry/Portal.cpp
paulgessinger: I added `s_transformEquivalentTolerance` to `Algebra.hpp` in this PR for this.

andiwand (approved):

asalzburger (approved):
Only got to read this now, very clean change, I like it!

acts-project-service:
## :red_circle: Athena integration test results [45692302448dc93af3cb53385256a9af5f92a69c]
Please investigate the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8022440)!

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891356)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891355)||
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891354)||
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891353)||
|:red_circle: | [test_ActsCheckObjectCountsCachedWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891352)||
|:red_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891351)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891350)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891349)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891348)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891347)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891346)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891345)||
|:green_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891344)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891343)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891342)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891341)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891340)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891339)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891338)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891337)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891336)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891335)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891334)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891333)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891332)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891331)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891330)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891329)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891328)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891327)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891326)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891325)||
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42891324)||

PR #3507, opened 2024-08-14
fix: Require TBB to be found by cmake

benjaminhuth:
Previously, the configuration could trun through even if TBB is not present on the system.


andiwand (dismissed):

timadye:
Just to put Paul's point another way... why do we need to requite TBB, when ACTS works fine (single-threaded) without it? (Or at least it used to, last time I checked.)

benjaminhuth:
Okay, sure, but compilation fails in that case I think (at least that was what I experienced). Shouldn't we use some more explicit logic here like having an option such as `ACTS_EXAMPLES_NO_TBB` or something like that?

paulgessinger:
@benjaminhuth We do have `ACTS_USE_EXAMPLES_TBB`, we just fall back to `OFF` if we don't find TBB. I'm ok to make this explicit only without fallback.

andiwand:
I agree, if `ACTS_USE_EXAMPLES_TBB` is on we should definitely use it without a fallback. Not finding it is an error case IMO

timadye:
But mostly `ACTS_USE_EXAMPLES_TBB` is on by default, rather than requested by the user with an explicit `-DACTS_USE_EXAMPLES_TBB=ON`. In this case, it is nice to have the fallback to `OFF` to make building ACTS easier for users that just want get something working without having to install `tbb` or find another CMake option. Is this possible to fall back to `OFF` iif it isn't specified?

The question remains why it didn't compile for you. That may be a more important issue to solve.

benjaminhuth:
I think you're right @timadye: I should try to reproduce the compile failure.

Anyhow: I think if `ACTS_USE_EXAMPLES_TBB=ON` and TBB is not found, we should fail the cmake configuration.

andiwand:
Maybe I am missing something but this is what is actually change in the PR, right? expect for a more descriptive message

paulgessinger:
@andiwand Almost, but it's using the `TBB_FOUND` variable in a way that doesn't make sense anymore without the fallback. I think it's maybe a 10 line refactor to clean it up.

benjaminhuth:
I found the source of the compile errore: The track finding algorithm unconditionally uses `tbb::combinable` because of the memory statistics. Should we hide this behind a preprocessor define @paulgessinger ?

https://github.com/acts-project/acts/blob/a28acc306ccd11f064c81a68e8eb7f3cc7526c2c/Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/TrackFindingAlgorithm.hpp#L44

paulgessinger:
@benjaminhuth True, that was me, and I forgot about it to be honest. I don't have a strong opinion what the best solution is. I would probably lean towards avoiding making the single threaded use case require workarounds in many places, and therefore switch back to hard-requiring TBB?

timadye:
I think it is a pity to add a required dependency without good reason. It seems this one should be quite easy to `#ifndef ACTS_EXAMPLES_NO_TBB` around the `#include` and `m_memoryStatistics` definition and use (2 places). I think that just means we don't get memory stats in the case without tbb.

A more complete solution would be to make a dummy class like `tbbWrap.hpp` does, but I don't think that's necessary in this case.

paulgessinger:
@timadye It's required dependency for our example framework, not the core lib, but I see your point.

I don't know, your call @benjaminhuth

benjaminhuth:
I think that these build errors were not observered for such a long time also indicates the option is not widely used.

timadye:
OK, I guess I can't argue with that.

Do these direct uses of `tbb` conflict with `tbbWrap`'s other function of bypassing `tbb` setup when running in a single-threaded mode? That definitely was useful to simplify debugging.

benjaminhuth:
> Do these direct uses of `tbb` conflict with `tbbWrap`'s other function of bypassing `tbb` setup when running in a single-threaded mode? That definitely was useful to simplify debugging.

I think this should stay untouched, since I only remove the macro-logic in case tbb is not present, but not the wrapper objects.

paulgessinger:
@benjaminhuth  `getting_started.md` needs an update I believe. I can override the coverage failure afterwards.

paulgessinger (approved):
I'm ok with this.

timadye (approved):
lgtm

acts-project-service:
## :white_check_mark: Athena integration test results [e2789b66fa67fe419b86d74ef2d3d86562c8949b]

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45358125)||
|:green_circle: | [run_art_test: test_run4_acts_vertex_PU200](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45358124)||
|:green_circle: | [run_art_test: test_run4_ttbar_PU200](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45358123)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45358122)||
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45358121)||

PR #3519, opened 2024-08-20
feat: GeoModel plugin changes for ITk building.

benjaminhuth:
This adds some funcitonality to the GeoModelPlugin:
- A very simple Module splitter based on the annulus bound radii
- A subclass of `GeoModelDetectorElement` that holds ITk specific information
- Bindings to python


Plugins/GeoModel/src/GeoModelDetectorElementITk.cpp
```diff
@@ -0,0 +1,51 @@
...
+    auto bounds = std::make_shared<bounds_t>(
+        dynamic_cast<const bounds_t &>(srf->bounds()));
+
+    auto itkEl = std::make_shared<GeoModelDetectorElementITk>(
+        detEl->physicalVolume(), nullptr, detEl->transform(gctx),
+        detEl->thickness(), hardware, barrelEndcap, layerWheel, etaModule,
+        phiModule, side);
+    auto surface = Surface::makeShared<surface_t>(bounds, *itkEl.get());
+    itkEl->attachSurface(surface);
+    itkEl->setDatabaseEntryName(detEl->databaseEntryName());
```
paulgessinger, line 34: This is a static function, can't you just directly set `m_surface` and `m_databaseEntryName` without having getters and setters?

Plugins/GeoModel/include/Acts/Plugins/GeoModel/SplitModulesByRadius.hpp
```diff
@@ -0,0 +1,43 @@
...
+  ModuleByRadiusSplitter(
+      const std::map<std::string, std::vector<double>> &designs,
+      double tolerance = 1.e-3,
+      Acts::Logging::Level level = Acts::Logging::INFO)
+      : m_designs(designs),
+        m_tolerance(tolerance),
+        m_logger(getDefaultLogger("ModuleByRadiusSplitter", level)) {}
+
+  std::optional<std::vector<std::pair<
+      std::shared_ptr<Surface>, std::shared_ptr<GeoModelDetectorElement>>>>
```
paulgessinger, line 32: This is a **very** complicated return type.

Plugins/GeoModel/src/SplitModulesByRadius.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include "Acts/Surfaces/DiscSurface.hpp"
+#include "Acts/Utilities/Logger.hpp"
+
+#include <iostream>
+
+namespace Acts {
+
+std::optional<std::vector<std::pair<std::shared_ptr<Surface>,
+                                    std::shared_ptr<GeoModelDetectorElement>>>>
+ModuleByRadiusSplitter::split(
```
paulgessinger, line 21: Does it make sense to add a test for this function?

Plugins/GeoModel/src/SplitModulesByRadius.cpp
```diff
@@ -0,0 +1,84 @@
...
+      result->emplace_back(newSurface, newDetElement);
+    }
+
+    return result;
+  }
+
+  ACTS_WARNING("Could not split '" << detElement->databaseEntryName()
+                                   << "' (rmin: " << annulusBounds->rMin()
+                                   << ", rmax: " << annulusBounds->rMax()
+                                   << ")");
```
paulgessinger, line 80: I guess this might be safer as an exception?

Plugins/GeoModel/src/GeoModelDetectorElementITk.cpp
```diff
@@ -0,0 +1,51 @@
...
+    return std::make_pair(itkEl, surface);
```
paulgessinger, line 35: Nitpick:
```suggestion
    return std::pair(itkEl, surface);
```

Plugins/GeoModel/include/Acts/Plugins/GeoModel/SplitModulesByRadius.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include "Acts/Plugins/GeoModel/GeoModelDetectorElement.hpp"
+#include "Acts/Utilities/Logger.hpp"
+
+#include <map>
+#include <string>
+#include <vector>
+
+namespace Acts {
+
+class ModuleByRadiusSplitter {
```
paulgessinger, line 20: Is this class used anywhere other than the python bindings?

paulgessinger (commented):
The database name is separate from the volume name?

Plugins/GeoModel/include/Acts/Plugins/GeoModel/SplitModulesByRadius.hpp
benjaminhuth: I could break up the "design pattern" and only return a vector of `DetectorElements`, and add accessors for the surface to the bindings.

Plugins/GeoModel/src/GeoModelDetectorElementITk.cpp
```diff
@@ -0,0 +1,51 @@
...
+    std::shared_ptr<Surface> srf, const GeometryContext &gctx, int hardware,
+    int barrelEndcap, int layerWheel, int etaModule, int phiModule, int side) {
+  auto helper = [&]<typename surface_t, typename bounds_t>() {
+    auto bounds = std::make_shared<bounds_t>(
+        dynamic_cast<const bounds_t &>(srf->bounds()));
+
+    auto itkEl = std::make_shared<GeoModelDetectorElementITk>(
+        detEl->physicalVolume(), nullptr, detEl->transform(gctx),
+        detEl->thickness(), hardware, barrelEndcap, layerWheel, etaModule,
+        phiModule, side);
```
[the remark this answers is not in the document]
benjaminhuth: Hmm I'm unsure what you mean here... something like using the base class constructor, and explicitly `setsHardware`, `setBarrelEndcap`...?

Plugins/GeoModel/src/GeoModelDetectorElementITk.cpp
benjaminhuth: You mean why these setters exist at all, or why they are called here? In this specific place I also could set them directly, would that make it more clear in your opinion?

Plugins/GeoModel/src/SplitModulesByRadius.cpp
benjaminhuth: Yeah actually it would, I wanted to do this anyways after sonarcloud complained.

Plugins/GeoModel/src/SplitModulesByRadius.cpp
benjaminhuth: Hmm I thought in case there is no matching split pattern, we can fallback to use the unsplit surface (similar as if now `AnnulusBounds` were found.

Plugins/GeoModel/include/Acts/Plugins/GeoModel/SplitModulesByRadius.hpp
benjaminhuth: Nope. Do you think I should implement it directly in the bindings/in python? I though about it, but it seemed to me there would be a lot of code binding necessary for this...

paulgessinger: Not necessarily, no. I was just wondering about this. Should this be called in a unit test at least somewhere?

Plugins/GeoModel/src/SplitModulesByRadius.cpp
paulgessinger: I see. Generally, WARNINGs / silent errors are hard to figure out so that's why I'm wondering if this should be one.

Plugins/GeoModel/src/GeoModelDetectorElementITk.cpp
paulgessinger: Exactly, since this method has access to the private member variables anyway.

benjaminhuth: Actually I can't, because these members are private to the base class and not protected...

Plugins/GeoModel/src/GeoModelDetectorElementITk.cpp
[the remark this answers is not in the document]
benjaminhuth: A got it.

Plugins/GeoModel/src/GeoModelDetectorElementITk.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "Acts/Surfaces/AnnulusBounds.hpp"
+#include "Acts/Surfaces/DiscSurface.hpp"
+#include "Acts/Surfaces/PlaneSurface.hpp"
+#include "Acts/Surfaces/RectangleBounds.hpp"
+
+namespace Acts {
+
+std::tuple<std::shared_ptr<Acts::GeoModelDetectorElementITk>,
+           std::shared_ptr<Acts::Surface>>
+Acts::GeoModelDetectorElementITk::convertFromGeomodel(
```
paulgessinger, line 20: Could this be a plain old constructor?

Plugins/GeoModel/src/GeoModelModuleSplitter.cpp
```diff
@@ -0,0 +1,86 @@
...
+  auto annulusBounds = dynamic_cast<const AnnulusBounds *>(&surface->bounds());
+
+  if (annulusBounds == nullptr) {
+    ACTS_DEBUG("Not annulus bounds");
+    return std::nullopt;
+  }
+
+  std::optional<std::vector<DetectorElementPtr>> result;
+
+  for (auto [patternName, radii] : m_splitPatterns) {
```
paulgessinger, line 34:
```suggestion
  for (const auto& [patternName, radii] : m_splitPatterns) {
```

Plugins/GeoModel/src/GeoModelModuleSplitter.cpp
```diff
@@ -0,0 +1,86 @@
...
+    radii.back() = annulusBounds->rMax();
+
+    ACTS_DEBUG("Accept pattern '" << patternName << "' for element '"
+                                  << detElement->databaseEntryName());
+
+    result.emplace();
+    result->reserve(radii.size() - 1);
+    const auto origValues = annulusBounds->values();
+
+    for (auto i = 0ul; i < radii.size() - 1; ++i) {
```
paulgessinger, line 52:
```suggestion
    for (std::size_t i = 0ul; i < radii.size() - 1; ++i) {
```

Plugins/GeoModel/src/GeoModelModuleSplitter.cpp
```diff
@@ -0,0 +1,86 @@
...
+      values[AnnulusBounds::eMaxR] = radii[i + 1];
+
+      ACTS_VERBOSE("Make new annulus bounds: " << [&]() {
+        std::stringstream ss;
+        for (auto v : values) {
+          ss << v << " ";
+        }
+        return ss.str();
+      }());
+      auto bounds = std::make_shared<AnnulusBounds>(values);
```
paulgessinger, line 65: Can you construct the bounds using the constructor that does not take the *bound value* array? Should take the same arguments.

Plugins/GeoModel/src/GeoModelModuleSplitter.cpp
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    return result;
+  }
+
+  ACTS_WARNING("Could not split '" << detElement->databaseEntryName()
+                                   << "' (rmin: " << annulusBounds->rMin()
+                                   << ", rmax: " << annulusBounds->rMax()
+                                   << ")");
+  return result;
```
paulgessinger, line 83: I think this is the `null` case. Does this make sense as a warning? What are the valid cases? If it's valid, it shouldn't be a WARNING, if it's invalid, it should be an error.

Could the return value just contain the original detector element in case it wasn't split, if that's not an error condition?

paulgessinger (commented):
I see that the use of `createDetectorElement` is modeled after the `GeoModelDetectorElement`, but I don't get the need for a factory function.

Plugins/GeoModel/src/GeoModelModuleSplitter.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "Acts/Surfaces/AnnulusBounds.hpp"
+#include "Acts/Surfaces/DiscSurface.hpp"
+#include "Acts/Utilities/Logger.hpp"
+
+#include <iostream>
+
+namespace Acts {
+
+auto GeoModelModuleSplitter::split(DetectorElementPtr detElement,
```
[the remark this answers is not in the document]
benjaminhuth: There is a trailing return type

Plugins/GeoModel/src/GeoModelModuleSplitter.cpp
benjaminhuth: why?

Plugins/GeoModel/src/GeoModelDetectorElementITk.cpp
benjaminhuth: I would like to have a function that makes a transform (detEl, surface) -> (detEl, surface), while in the end everything is correctly connected. This is not possible with a plain constructor.

Plugins/GeoModel/src/GeoModelModuleSplitter.cpp
paulgessinger: Absolutely, if it's an error, you can probably throw an exception and not return anything at all, right?

Plugins/GeoModel/src/GeoModelDetectorElementITk.cpp
paulgessinger: If you need the surface to be returned externally, that's true. But do you? It might make sense to assemble the pair at the call site. The detector element contains the surface anyway, so the pair return type is redundant in some sense.

Plugins/GeoModel/src/GeoModelModuleSplitter.cpp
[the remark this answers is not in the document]
paulgessinger: And you need that because of the nested typedef? Ok. Consider either removing the typedef as per my other comment or explicitly typing out the full type name on the left side of the signature.

Plugins/GeoModel/src/GeoModelDetectorElementITk.cpp
benjaminhuth: I don't see why it makes sense to assemble the pair on the call site (which might be on python).
The fact that the types are redundant in is true, but that is the current pattern in the plugin, and it would require additional bindings etc. to make it work on the python side. I wouldn't want to touch this in that PR. Can we just leave it like this for now?

benjaminhuth:
> The database name is separate from the volume name?

Yes, it is. The volume name usually is only "ECSensor1" or so...

Plugins/GeoModel/src/GeoModelDetectorElementITk.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+    itkEl->m_hardware = hardware;
+    itkEl->m_barrelEndcap = barrelEndcap;
+    itkEl->m_layerWheel = layerWheel;
+    itkEl->m_etaModule = etaModule;
+    itkEl->m_etaModule = phiModule;
+    itkEl->m_side = side;
+
+    itkEl->attachSurface(surface);
+    itkEl->setDatabaseEntryName(detEl->databaseEntryName());
```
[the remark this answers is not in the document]
benjaminhuth: The surface I cannot, since it is created after the constructor call. The database string I could, but I would actually leave it like that in this case if its okay.

Plugins/GeoModel/src/GeoModelModuleSplitter.cpp
benjaminhuth: We in fact need a copy here.

andiwand (dismissed):
LGTM 👍

@paulgessinger I see two open discussions. Should we follow up on them here or on future PRs? As far as I understand there are a couple of things we may have to revisit in that Plugin.

To me it looks like one general confusion comes from the fact that we can construct `Surface` from a `DetectorElement` but I think in practice we always construct the surface first and then assign a detector element to it.

paulgessinger:
@andiwand ok to proceed for me.

andiwand:
@benjaminhuth is in the process of proposing changes to `Surface` and `DetectorElement` so can converge on one construction path. In that process we can also rework the construction in the Plugins

paulgessinger:
> To me it looks like one general confusion comes from the fact that we can construct `Surface` from a `DetectorElement` but I think in practice we always construct the surface first and then assign a detector element to it.

In ATLAS I believe I do it the other way round, the detector element constructs the surface.

andiwand:
I have not a good overview over our construction code but I feel like we convert to surfaces from some native format and then we "convert" from surface to detector element. If that is the case I would drop the surface construction from detector element and only keep the register function. In that case we can still construct the surface from within the detector element but break the seemingly circular dependency during construction and the confusion which way to go.

paulgessinger:
I still think the detector element should own the surface, however.

benjaminhuth:
@paulgessinger @andiwand  See #3574 for a proposal to change the interface of detector elment

andiwand (dismissed):

andiwand (dismissed):

andiwand (approved):

andiwand:
blocked by clang-tidy failures

AJPfleger (approved):

benjaminhuth:
I'd like to hold this for a moment, I'm currently working with that change and maybe there are some little adjustements that are needed still...

PR #3529, opened 2024-08-21
refactor: Rework projector

andiwand:
Originally https://github.com/acts-project/acts/pull/3453

---

Reworks the projector storage in the track EDM and the operations on vectors matrices. The bitset in the track EDM is replaced by a simple sequence of `std::uint8_t` which provide the mapping `measurement index -> bound index`. A helper is provided, `SubspaceHelper`, which collects convenience functions that convert to the legacy format and to matrix form and also do operations on vectors and matrices e.g. `projectVector` and `projectMatrix`.

The benefit of this change is that the mapping and storage of the mapping becomes easier to grasp and also computationally less expensive. Instead of relying on eigen matrix operations we map directly to the desired object.

A similar change in Athena improved the overall track finding CPU time by 30%.

In favor of `SubspaceHelper` I removed `FixedSizeSubspace` and `VariableSizeSubspace` and made the necessary changes in the Examples.

In principle these changes should be non-breaking. The switch in API should be scheduled for v37.


acts-project-service:
## :red_circle: Athena integration test results [f01d28a8150008d1c911c365b9dee41210bc3116]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/42749779) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8000918)!

andiwand:
Athena seems to violate some of our subspace index assumptions
```
MultiTrajectoryBasic_test.exe: /builds/acts/acts-athena-ci/acts-install/include/Acts/EventData/SubspaceHelpers.hpp:160: Acts::VariableSubspaceHelper<FullSize, index_t>::VariableSubspaceHelper(const OtherContainer&) [with OtherContainer = std::span<unsigned char>; long unsigned int FullSize = 6; index_t = unsigned char]: Assertion `checkSubspaceIndices(indices, kFullSize, indices.size()) && "Invalid indices"' failed.
 nopost.sh> ERROR: Test MultiTrajectoryBasic_test failed with exit code: 134
MultiTrajectoryBasic_test.exe: /builds/acts/acts-athena-ci/acts-install/include/Acts/EventData/SubspaceHelpers.hpp:160: Acts::VariableSubspaceHelper<FullSize, index_t>::VariableSubspaceHelper(const OtherContainer&) [with OtherContainer = std::span<unsigned char>; long unsigned int FullSize = 6; index_t = unsigned char]: Assertion `checkSubspaceIndices(indices, kFullSize, indices.size()) && "Invalid indices"' failed.
 noerror.sh> Please check /builds/acts/acts-athena-ci/athena-build/Tracking/Acts/ActsEvent/CMakeFiles/unitTestRun/MultiTrajectoryBasic_test.log
 ```

**Comment by andiwand** [comment]
`test_ActsBenchmarkWithSpot` fails somehow. I hope that disappears with the other PRs in the pipeline. Holding this back until then

File: Core/include/Acts/EventData/SubspaceHelpers.hpp [hunk 1]
```diff
@@ -0,0 +1,326 @@
...
+#include "Acts/Utilities/Enumerate.hpp"
+
+#include <bitset>
+#include <span>
+
+#include <boost/container/static_vector.hpp>
+
+namespace Acts {
+
+/// Check subspace indices for consistency
```
benjaminhuth, line 24: Maybe a short sentence what means consistent here exactly, at least to me it is not super obvious.

Core/include/Acts/EventData/SubspaceHelpers.hpp
```diff
@@ -0,0 +1,326 @@
...
+
+/// Serialize subspace indices to a single 64 bit integer
+///
+/// @tparam FullSize the full size of the subspace
+///
+/// @param indices the subspace indices
+///
+/// @return the serialized subspace indices
+template <std::size_t FullSize>
+inline static SerializedSubspaceIndices serializeSubspaceIndices(
```
benjaminhuth, line 64: Now where I used the `Acts::MultiIndex` yesterday, I wonder if that would not be a good tool for implementing that...

Core/include/Acts/EventData/SubspaceHelpers.hpp
```diff
@@ -0,0 +1,326 @@
...
+  requires(FullSize <= 8)
+{
+  SubspaceIndices<FullSize> result;
+  for (std::size_t i = 0; i < FullSize; ++i) {
+    result[i] = static_cast<std::uint8_t>(serialized >> (i * 8));
+  }
+  return result;
+}
+
+namespace detail {
```
benjaminhuth, line 94: Hmm I wonder if this file should go in `detail`, and if the above functions are actually public API?

Core/include/Acts/EventData/SubspaceHelpers.hpp
andiwand: Could be but I don't really see a benefit in that. Do you have an example one how you think the implementation would get easier / more readable?

Core/include/Acts/EventData/SubspaceHelpers.hpp
andiwand: IMO everything here is public API because it is used in the track EDM as return value or as conversion helper

Core/include/Acts/EventData/SubspaceHelpers.hpp
benjaminhuth: Yeah I though it would rely on a already established and tested tool for such bit shifting buisness. But yeah not a strong preference.

Core/include/Acts/EventData/SubspaceHelpers.hpp
benjaminhuth: okay so the `namespace detail` should go away?

benjaminhuth: ah sorry I didn't realize the namespace detail ends directly below that class

andiwand: the `detail` only captures the base class for the CRTP. I think this is fine to be a detail

PR #3573, opened 2024-08-28
refactor!: Fuse `Actor` and `Aborter`

andiwand:
Fuses Actors and Aborters into a single object with seperate `act` and `check` calls. The motivation for this is that Aborters cannot hold any state by their own currently but need an Actor to point to. To simplify this I propose to couple these two concepts.

At the same time I reworked the call mechanism for the actor and aborter functions which concepts and a simpler dispatch mechanism.

Class names changed a bit from `ActionList` to `ActorList` for example.


asalzburger:
Yes, why not - we actually had `actor, observer, aborter` in the original design I suppose, and I think that was over-designed.

The only comment I have here is that the ordering of the actors gets even more important, but that's not a bad thing - on the contrary, I would say.

andiwand:
The way I combined them right now preserves the ordering of the actor/aborter. I just combined the `act` and `check` function into one object. The calls are still happening as before, first `act` then `check`, in sequence.

Potentially we could also fuse the two functions since it is really just the return value which is different but then we have to be more careful about the ordering like you said @asalzburger 

cc @paulgessinger @benjaminhuth

Core/include/Acts/Propagator/ActorList.hpp
```diff
@@ -0,0 +1,130 @@
...
+  /// @tparam stepper_t Type of the stepper
+  /// @tparam navigator_t Type of the navigator
+  ///
+  /// @param [in,out] state is the state object from the propagator
+  /// @param [in] stepper Stepper used for the propagation
+  /// @param [in] navigator Navigator used for the propagation
+  /// @param [in] args are the arguments to be passed to the aborters
+  template <typename propagator_state_t, typename stepper_t,
+            typename navigator_t, typename... Args>
+  bool check(propagator_state_t& state, const stepper_t& stepper,
```
benjaminhuth, line 122: I think I would lean slightly towards naming this function actually `abort`, because `check` is very generic...

andiwand: I am also not quite happy with `act` and even less with `check`. `abort` sounded to me like we would call for an abort. Maybe `checkAbort`?

benjaminhuth: `checkAbort` sounds good

Core/include/Acts/Propagator/ActorConcepts.hpp
andiwand, line 1: please roast my attempt on concepts @stephenswat

Tests/UnitTests/Core/Propagator/DirectNavigatorTests.cpp
```diff
@@ -107,19 +106,18 @@ void runTest(const rpropagator_t& rprop, const dpropagator_t& dprop, double pT,
   using EndOfWorld = EndOfWorldReached;
 
   // Action list and abort list
-  using RefereceActionList = ActionList<MaterialInteractor, SurfaceCollector<>>;
-  using ReferenceAbortList = AbortList<EndOfWorld>;
+  using RefereceActorList =
```
stephenswat, line 109: Can we fix this spelling error?

Core/include/Acts/Propagator/detail/actor_list_implementation.hpp
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  template <typename actor_t, typename propagator_state_t, typename stepper_t,
+            typename navigator_t, typename... Args>
+  static bool check(const actor_t& actor, propagator_state_t& state,
+                    const stepper_t& stepper, const navigator_t& navigator,
+                    Args&&... args) {
+    static_assert(
+        Actor<actor_t, propagator_state_t, stepper_t, navigator_t, Args...>,
+        "The actor does not fulfill the Actor concept");
```
stephenswat, line 50: I'd suggest moving this into a requires clause function.

stephenswat (changes requested):
Looks very solid; minor changes only.

stephenswat (approved):

Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -501,6 +501,17 @@ class KalmanFitter {
...
 
+    template <typename propagator_state_t, typename stepper_t,
+              typename navigator_t>
+    bool checkAbort(propagator_state_t& /*state*/, const stepper_t& /*stepper*/,
+                    const navigator_t& /*navigator*/, const result_type& result,
+                    const Logger& /*logger*/) const {
+      if (!result.result.ok() || result.finished) {
+        return true;
+      }
+      return false;
```
stephenswat, line 512: Nit:

```suggestion
      return (!result.result.ok() || result.finished);
```

Core/include/Acts/Propagator/ActorConcepts.hpp
```diff
@@ -0,0 +1,105 @@
...
+template <typename actor_t, typename propagator_state_t, typename stepper_t,
+          typename navigator_t, typename... Args>
+concept ActorHasActWithResult =
+    requires(const actor_t& a, propagator_state_t& state,
+             const stepper_t& stepper, const navigator_t& navigator,
+             typename actor_t::result_type& result, Args&&... args) {
+      {
+        a.act(state, stepper, navigator, result, std::forward<Args>(args)...)
+      } -> std::same_as<void>;
+    };
```
stephenswat, line 29: I think it would make sense to require `ActorHasResult` here:

```suggestion
concept ActorHasActWithResult = ActorHasResultr<T> &&
    requires(const actor_t& a, propagator_state_t& state,
             const stepper_t& stepper, const navigator_t& navigator,
             typename actor_t::result_type& result, Args&&... args) {
      {
        a.act(state, stepper, navigator, result, std::forward<Args>(args)...)
      } -> std::same_as<void>;
    };
```

Core/include/Acts/Propagator/ActorConcepts.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+template <typename actor_t, typename propagator_state_t, typename stepper_t,
+          typename navigator_t, typename... Args>
+concept ActorHasActWithoutResult = requires(
+    const actor_t& a, propagator_state_t& state, const stepper_t& stepper,
+    const navigator_t& navigator, Args&&... args) {
+  {
+    a.act(state, stepper, navigator, std::forward<Args>(args)...)
+  } -> std::same_as<void>;
+};
```
stephenswat, line 39: Similarly:

```suggestion
concept ActorHasActWithoutResult = !ActorHasResult<T> && requires(
    const actor_t& a, propagator_state_t& state, const stepper_t& stepper,
    const navigator_t& navigator, Args&&... args) {
  {
    a.act(state, stepper, navigator, std::forward<Args>(args)...)
  } -> std::same_as<void>;
};
```

Core/include/Acts/Propagator/ActorConcepts.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+template <typename actor_t, typename propagator_state_t, typename stepper_t,
+          typename navigator_t, typename... Args>
+concept ActorHasAct =
+    (!ActorHasResult<actor_t> &&
+     ActorHasActWithoutResult<actor_t, propagator_state_t, stepper_t,
+                              navigator_t, Args...>) ||
+    (ActorHasResult<actor_t> &&
+     ActorHasActWithResult<actor_t, propagator_state_t, stepper_t, navigator_t,
+                           Args...>);
```
stephenswat, line 49: Then this simplifies:

```suggestion
concept ActorHasAct = ActorHasActWithResult<T> || ActorHasActWithoutResult<T>;
```

Core/include/Acts/Propagator/ActorConcepts.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+template <typename actor_t, typename propagator_state_t, typename stepper_t,
+          typename navigator_t, typename... Args>
+concept ActorHasAbort =
+    (!ActorHasResult<actor_t> &&
+     ActorHasAbortWithoutResult<actor_t, propagator_state_t, stepper_t,
+                                navigator_t, Args...>) ||
+    (ActorHasResult<actor_t> &&
+     ActorHasAbortWithResult<actor_t, propagator_state_t, stepper_t,
+                             navigator_t, Args...>);
```
stephenswat, line 80: Same as my comment above.

Core/include/Acts/Propagator/ActorConcepts.hpp
```diff
@@ -0,0 +1,105 @@
...
+    (!ActorHasResult<actor_t> &&
+     ActorHasActWithoutResult<actor_t, propagator_state_t, stepper_t,
+                              navigator_t, Args...> &&
+     ActorHasAbortWithoutResult<actor_t, propagator_state_t, stepper_t,
+                                navigator_t, Args...>) ||
+    (ActorHasResult<actor_t> &&
+     ActorHasActWithResult<actor_t, propagator_state_t, stepper_t, navigator_t,
+                           Args...> &&
+     ActorHasAbortWithResult<actor_t, propagator_state_t, stepper_t,
+                             navigator_t, Args...>);
```
stephenswat, line 94: This then simplifies to:

```suggestion
template <typename actor_t, typename propagator_state_t, typename stepper_t,
          typename navigator_t, typename... Args>
concept ActorHasActAndAbort = ActorHasAct<T> && ActorHasAbort<T>;
```

Core/include/Acts/Propagator/ActorConcepts.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+template <typename actor_t, typename propagator_state_t, typename stepper_t,
+          typename navigator_t, typename... Args>
+concept Actor =
+    ActorHasAct<actor_t, propagator_state_t, stepper_t, navigator_t, Args...> ||
+    ActorHasAbort<actor_t, propagator_state_t, stepper_t, navigator_t,
+                  Args...> ||
+    ActorHasActAndAbort<actor_t, propagator_state_t, stepper_t, navigator_t,
+                        Args...>;
```
stephenswat, line 103: This then also simplifies:

```suggestion
concept Actor =
    ActorHasAct<actor_t, propagator_state_t, stepper_t, navigator_t, Args...> ||
    ActorHasAbort<actor_t, propagator_state_t, stepper_t, navigator_t,
                  Args...>
```

Core/include/Acts/Propagator/ActorList.hpp
```diff
@@ -0,0 +1,158 @@
...
+      "all actors must be default constructible");
+  static_assert(
+      detail::all_of_v<std::is_copy_constructible<actors_t>::value...>,
+      "all actors must be copy constructible");
+  static_assert(
+      detail::all_of_v<std::is_move_constructible<actors_t>::value...>,
+      "all actors must be move constructible");
+
+  static_assert(!detail::has_duplicates_v<actors_t...>,
+                "same action type specified several times");
```
stephenswat, line 42: Can we put these into a requires clause on the class?

stephenswat (commented):
Some comments on how this might be improved, but I am not 100% sure it's equivalent.

andiwand:
@stephenswat I was wondering if the `Actor` should just flag what it will provide and only then we couple to the functions instead of deducing it from if the function exists or not. What do you think? Since this is breaking anyways we might just make this more explicit

andiwand:
I left the flags out and applied your comments @stephenswat. I think that makes it way easier to just check for methods and not combinations

stephenswat (approved):

Core/include/Acts/Propagator/ActorConcepts.hpp
```diff
@@ -0,0 +1,87 @@
...
+template <typename actor_t>
+concept ActorHasResult = requires { typename actor_t::result_type; };
+
+template <typename actor_t, typename propagator_state_t, typename stepper_t,
+          typename navigator_t, typename... Args>
+concept ActorHasActWithoutResult = requires(
+    const actor_t& a, propagator_state_t& state, const stepper_t& stepper,
+    const navigator_t& navigator, Args&&... args) {
+  {
+    a.act(state, stepper, navigator, std::forward<Args>(args)...)
```
stephenswat, line 26: SonarCloud has a good point; this should be `std::move` instead (also for 3 other occurrences in this file).

stephenswat (changes requested):

Core/include/Acts/Propagator/ActorConcepts.hpp
andiwand: Just for me to understand: Isn't `Args&&` a "forward reference"?

andiwand: Also isn't it kind of weird that it complains about that on the concept?

paulgessinger: I think it's only a universal reference if it's a function argument.

stephenswat: I'll give a bit of a verbose description of this; you are probably already familiar with most of this but just for posterity, because I think it can be a bit of a confusing topic. Might also be helpful for other people.

So first we will need to really understand universal references. A universal reference looks like this:

```cpp
template<typename T>
void my_function(T && t);
```

Why is this a universal reference? Because it can take both lvalue references and rvalue references due to something known as the C++ reference collapsing rules. Let's take a look at what happens when we call `my_function` with an lvalue reference:

```cpp
std::vector<int> t;
my_function(t);
```

In this case, `T` will resolve to be `std::vector<int> &`, and so the type of `t` will be `std::vector<int> &&&`. The reference collapsing rules refine that `T&&&` collapses to `T&`, so `t` has the type `T&`. It's an lvalue reference inside `my_function`, and `my_function` is able to use it as such.

Now let's call the same function with an rvalue reference instead:

```cpp
std::vector<int> t;
my_function(std::move(t));
```

In this case, `T` resolves to `std::vector<int> &&` so `t` has type `std::vector<int> &&&&` which resolves to `std::vector<int> &&`. The magic of universal references is that one templated function can support both lvalue and rvalue references. But we need to be careful. Let's imagine we want to use that reference inside of `my_function`:

```cpp
template<typename T>
void my_function(T && t) {
    std::vector<int>(t);
}
```

In the case where `T` is `std::vector<int> &`, this works as normal. It is passed as an lvalue reference, so you get the expensive copy operation. But what if we pass an rvalue reference to `my_function`? Inside of `my_function`, `t` is an lvalue, so even though it was moved into `my_function`, we still have to use the expensive copy operation instead of the much more efficient move operation. So let's "fix" `my_function`:

```cpp
template<typename T>
void my_function(T && t) {
    std::vector<int>(std::move(t));
}
```

This now works well if `my_function` is passed an rvalue, as it will give us the efficient copy constructor. But we've now introduced a massive bug in the code! If we pass `my_function` an lvalue, we are moving a reference which might put the vector into an invalid state. Remember, the caller of `my_function` still has a reference to that vector which `my_function` has now inadvertently moved. This could lead to a segmentation fault or worse.

So this is where `std::forward` comes in. It will cast `t` to `T &&` if it was passed as an rvalue, and to `T &` if it was an lvalue. That allows us to write `my_function` properly:

```cpp
template<typename T>
void my_function(T && t) {
    std::vector<int>(std::forward<T>(t));
}
```

The magic of this is that if `my_function` is passed an lvalue reference, it is forwarded as an lvalue and we get the safe, slow copy constructor. If it is passed an rvalue, it will move `t` instead. We know that this is safe because `t` was originally moved into `my_function`! This is what is called perfect forwarding.

So, we know that using `std::move` instead of `std::forward` is very dangerous. But now let's consider the following code:

```cpp
template<typename T>
struct my_struct {
    void my_function(T && t);
};
```

This situation is fundamentally different, because `T` is determined on the struct level rather than on the function level. In other words, `T` is not a universal reference, because `T` is already fixed to be either `U`, `U &`, or `U &&`. You could invoke `my_struct::my_function` in whatever way you want, and `T` will always be the same, e.g.:

```cpp
my_struct<std::vector<int>&> s;
std::vector<int> v;
s.my_function(v); // valid, passed as lvalue reference.
s.my_function(std::move(v)); // valid, but still an lvalue reference, because `T` is bound to `s` and is not a template of the function call
```

In this case, `std::forward` makes no sense. I don't think it's _harmful_ per se, but it doesn't do anything.

And this is why SonarCloud is complaining. The code has a template type on the outside struct (or, in this case, the outside concept) so it is not a universal reference.

stephenswat (approved):
LGTM. :+1:

Core/include/Acts/Propagator/ActorConcepts.hpp
stephenswat: The story for concepts is a bit more complicated, because when you add a `requires` clause with arguments, those arguments are never actually instantiated, so you can't actually have a `requires` clause with template arguments:

```
template<typename T>
requires(T && t) { ... };
```

This doesn't make sense, because you don't actually "call" this required clause anywhere, so there is never any template deduction. For concepts, you can only specify templates on the concept itself, and you have to explicitly provide those types. Again, no deduction is ever involved. So there is no real possibility of having universal references here.

Core/include/Acts/Propagator/detail/actor_list_implementation.hpp
```diff
@@ -0,0 +1,104 @@
...
+  static bool checkAbort(const std::tuple<actors_t...>& actor_tuple,
+                         propagator_state_t& state, const stepper_t& stepper,
+                         const navigator_t& navigator, Args&&... args) {
+    return std::apply(
+        [&](const actors_t&... actor) {
+          return (actor_caller::checkAbort(actor, state, stepper, navigator,
+                                           std::forward<Args>(args)...) ||
+                  ...);
+        },
+        actor_tuple);
```
paulgessinger, line 100: I think (might be wrong) previously, we would not call subsequent aborters when the first one indicated an abort condition. This I don't think does that anymore. Do we want that? Do we care?

Core/include/Acts/Utilities/detail/MPL/type_collector.hpp
paulgessinger, line 1: Spurious changes?

paulgessinger (approved):

Core/include/Acts/Propagator/detail/actor_list_implementation.hpp
paulgessinger: I missed the `||` in the fold expression. You're right.

acts-project-service:
## :white_check_mark: Athena integration test results [24735fba4a8fdcf248a94421d1ad56684023583e]

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901231)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901230) | [:paperclip:](https://acts-athena-outputs.web.cern.ch/test_data18_13TeV_1000evt/pulls/3573/)|
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901229) | [:paperclip:](https://acts-athena-outputs.web.cern.ch/test_ttbarPU40_reco/pulls/3573/)|
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901228)||
|:green_circle: | [test_ActsCheckObjectCountsCachedWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901227)||
|:green_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901226)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901225)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901224)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901223)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901222)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901221)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901220)||
|:green_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901219)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901218)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901217)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901216)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901215)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901214)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901213)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901212)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901211)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901210)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901209)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901208)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901207)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901206)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901205)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901204)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901203)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901202)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901201)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901200)||
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43901199)||

PR #3574, opened 2024-08-29
refactor: Detector Element refactoring proposal

benjaminhuth:
After several discussions with @asalzburger and @paulgessinger, I thought about a redesign of the `DetectorElementBase` with @andiwand. This should make clear the ownership model and order of construction, avoiding constructions like currently implemented.

Core ideas are:
- The `DetectorElementBase` has shared ownership to a surface
- Through `friend`ship with `Surface` it soley controls the association of a surface to a detector element.
- Surface constructors with `DetectorElementBase` are removed.  

**Note:** This is not meant to compile at the moment, it rather should be a point to discuss the interface on concrete code.

### DIFF
--- a/Core/include/Acts/Geometry/DetectorElementBase.hpp
+++ b/Core/include/Acts/Geometry/DetectorElementBase.hpp
@@ -31,7 +31,9 @@ class Surface;
 
 class DetectorElementBase {
  public:
-  DetectorElementBase() = default;
+  /// Construct a detector element. It will have ownership of the surface
+  DetectorElementBase(std::shared_ptr<Surface>&& surface)
+      : m_surface(surface) {}
   virtual ~DetectorElementBase() = default;
 
   /// Return the transform for the Element proxy mechanism
--- a/Core/include/Acts/Geometry/DetectorElementBase.hpp
+++ b/Core/include/Acts/Geometry/DetectorElementBase.hpp
@@ -10,6 +10,7 @@
 
 #include "Acts/Definitions/Algebra.hpp"
 #include "Acts/Geometry/GeometryContext.hpp"
+#include "Acts/Surfaces/Surface.hpp"
 
 #include <memory>
 #include <vector>
@@ -32,9 +33,16 @@ class Surface;
 class DetectorElementBase {
  public:
   /// Construct a detector element. It will have ownership of the surface
-  DetectorElementBase(std::shared_ptr<Surface>&& surface)
-      : m_surface(surface) {}
-  virtual ~DetectorElementBase() = default;
+  DetectorElementBase(std::shared_ptr<Surface>&& surface) : m_surface(surface) {
+    if (m_surface->m_associatedDetElement != nullptr) {
+      throw std::runtime_error("Detector Element already associated");
+    }
+
+    m_surface->m_associatedDetElement = this;
+  }
+  virtual ~DetectorElementBase() {
+    m_surface->m_associatedDetElement = nullptr;
+  }
 
   /// Return the transform for the Element proxy mechanism
   ///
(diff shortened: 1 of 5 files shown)


Core/include/Acts/Geometry/DetectorElementBase.hpp
```diff
@@ -31,23 +32,35 @@ class Surface;
 
 class DetectorElementBase {
  public:
-  DetectorElementBase() = default;
-  virtual ~DetectorElementBase() = default;
+  /// Construct a detector element. It will have ownership of the surface
+  DetectorElementBase(std::shared_ptr<Surface>&& surface) : m_surface(surface) {
```
andiwand, line 36:
```suggestion
  explicit DetectorElementBase(std::shared_ptr<Surface>&& surface) : m_surface(surface) {
```

also
1. does the rvalue give us any benefit? looks like we could just use a value
2. this is a breaking change right? should we schedule that for v37?
3. should we make assumptions about the ownership? require that the shared pointer is not used anywhere else or enforce that with a `unique_ptr`?

Core/include/Acts/Geometry/DetectorElementBase.hpp
```diff
@@ -31,23 +32,35 @@ class Surface;
...
+    if (m_surface->m_associatedDetElement != nullptr) {
+      throw std::runtime_error("Detector Element already associated");
+    }
+
+    m_surface->m_associatedDetElement = this;
+  }
+  virtual ~DetectorElementBase() {
+    m_surface->m_associatedDetElement = nullptr;
+  }
```
andiwand, line 45: I wonder if this is strictly necessary. Also could that be a problem with thread safety?

Core/include/Acts/Geometry/DetectorElementBase.hpp
benjaminhuth: 1. Its more a hint to the caller, but not strictly necessary
2. I would think that is definitively breaking.
3. So the issue I encountered with `unique_ptr` is that `std::shared_ptr<...>::unique` is actually removed from C++20 because of thread safety issues... We could assume that geometry construction is always done in a single thread however.

Core/include/Acts/Geometry/DetectorElementBase.hpp
asalzburger: Yes, and no, actually `DetectorElement` should live longer than even the `MultiTrajectory`, but it is hard to control.

Core/include/Acts/Geometry/DetectorElementBase.hpp
```diff
@@ -31,23 +32,35 @@ class Surface;
...
   /// @param gctx The current geometry context object, e.g. alignment
   virtual const Transform3& transform(const GeometryContext& gctx) const = 0;
 
   /// Return surface representation - const return pattern
-  virtual const Surface& surface() const = 0;
+  virtual const Surface& surface() const { return *m_surface; };
 
   /// Non-const return pattern
-  virtual Surface& surface() = 0;
+  virtual Surface& surface() { return *m_surface; };
```
asalzburger, line 56: This I like, makes it clear where the surface comes from.

benjaminhuth:
Hmm because we have these livetime issues: What is actually the reason, that the Surface cannot own its detector element? like this:

```c++
class Surface {
    // ...
    const DetectorElementBase *associatedDetectorElement();
	void associateDetectorElement(std::unique_ptr<DetectorElementBase> &&el);
  private:
	std::unique_ptr<DetectorElementBase> m_detectorElement;
};
```

Because that would also solve our livetime issues, right?

paulgessinger:
With most construction code now using detector element factories, is this still needed @benjaminhuth @asalzburger @andiwand ?

benjaminhuth:
Not a priority right now

benjaminhuth:
Not a priority right now

PR #3420, opened 2024-07-20
feat: Implement modified Bryson–Frazier (mBF) smoother as alternative to `GainMatrixSmoother`

andiwand:
Adding an implementation of the [modified Bryson–Frazier (mBF)](https://en.wikipedia.org/wiki/Kalman_filter#Modified_Bryson%E2%80%93Frazier_smoother) which can be used as a replacement for the existing `GainMatrixSmoother`. The benefit of mBF is that we do not require any matrix inverses in principle. The current implementations recomputes $S^{-1}$ which could be cached by the forward filtering but in principle this should be already more optional than the inverse of the bound covariance in the `GainMatrixSmoother`. mBF seems to require more computation apart from the inverse so lets see if we gain anything.

Thanks again @pbutti for going through the math with me.

## Performance
`GainMatrixSmoother / MbfSmoother` = 2.56 speedup without caching

In out current track finding workloads this does not really change anything because we spend about 1% of time in the smoothing.

PR #3582, opened 2024-08-30
perf: Use MBF smoother by default in `TrackHelpers` in Examples track finding

andiwand:
Making the MBF smoother default in the Examples track finding after https://github.com/acts-project/acts/pull/3420


paulgessinger:
Did this have a significant impact on performance?

andiwand:
> Did this have a significant impact on performance?

I measured a speedup of `2.56` which does not really effect the overall track finding performance because we only spend about 1% of the time in the smoothing step. I measured that with ODD and Fatras so it might have a bigger impact with full sim and ITk. Since we do not loose anything be using this I thought it makes sense to use it as a default.

Interestingly it seems to shift some outlier tracks which is why physmon fails. Some of the pT outliers are shuffled around but I think those where outliers in the first place.

paulgessinger:
My only argument against this is that it's nice to have the straight up 1:1 Frühwirth smoothing as the default, since it's readily available in literature.

andiwand:
> My only argument against this is that it's nice to have the straight up 1:1 Frühwirth smoothing as the default, since it's readily available in literature.

Good point - maybe it makes more sense to use it per default in the Examples then?

paulgessinger (commented):
No objections. Would we be testing / monitoring the gain matrix smoother anywhere? Might be nice to avoid decay.

PR #3595, opened 2024-09-05
refactor: Volume holds bounds as mutable

paulgessinger:
This PR changes the internal storage of `VolumeBounds` inside `Volume` to be mutable. This allows transitive mutable access (which we need in a few cases).

To prevent accidental mutable use of the contained bounds object (which can be shared with other objects), I'm moving the shared pointer member to be private, and allow access only via setters and getters. This way, in a const function of a derived class, only const access to the volume bounds is possible.


paulgessinger:
cc @Berggren-Jonas

stephenswat (approved):
LGMT. :+1:

acts-project-service:
## :red_circle: Athena integration test results [2798d98247ce73c3055ff38f54d586cd7f184ea5]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43269755) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8082905)!

PR #3625, opened 2024-09-18
refactor: Remove Sequencer dataflow override

paulgessinger:
This option does more harm than it does good, so I propose we remove it altogether.


asalzburger:
> This option does more harm than it does good, so I propose we remove it altogether.

You want to always check the dataflow then?

Both @benjaminhuth and I had problems with the check, i.e. we then need to get to the bottom of it why this could potentially fail.

paulgessinger:
@asalzburger Exactly. If you have a reproducer, it would be great to investigate why it fails.

andiwand (approved):
I also believe the check is fine and it is a good protection mechanism against inconsistent data types on the whiteboard of our Examples Framework. If there are problems with that mechanism it is highly likely that it fails due to inconsistent compilation units which might be cause by build caches like ccache or dirty build folders. For more details see the discussion here https://mattermost.web.cern.ch/acts/pl/teonnaeyzjf1xqzgcz9nuyobzy

I would generally argue against these kinds of workarounds against protection mechanisms as this looks like we are hacking them and distrusting them on purpose. If people experience problems with it, it is most likely a symptom of a deeper issue and the actual issue should be treated instead of the symptom. No one will stop one from hacking it locally, this is simply one line change to "make it work" but I would urge to keep the main repo clean of it.

benjaminhuth:
So I agree that if the check fails, this is a sign that something in the build that is used is wrong, so removing the option alltogether is also fine...

But yeah, as @paulgessinger said, I see myself already hacking this in cases this happens and rebuild of the whole acts+dependencies is not an option immediatly...

paulgessinger:
@benjaminhuth If we understand why it happens we could try to identify which parts need to be rebuild. The mismatch that @asalzburger reported this morning, `flat_map` is a header-only part of boost. That would indicate to me that it's not a dependency issue, but a build consistency issue, possibly `ccache`.

It's *possible* that the check is faulty, in which case it should be improved. I can't reproduce this issue at this moment, so I can't debug it myself, which is why I'd love if you or @asalzburger could help debugging this.

asalzburger:
Let's merge this one then.

asalzburger (approved):

acts-project-service:
## :red_circle: Athena integration test results [9ee3a10813e1ecf00d6fbe36e4872dbb33791536]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43893650) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8194518)!

PR #3627, opened 2024-09-18
feat: Add measurement emplace functions

stephenswat:
Recently, the measurement EDM was replaced by a memory-optimal bulk storage, but this makes it a bit harder to add new measurements to a collection. This commit adds two new emplace functions which variadically allow users to add measurements, similar to how they might use `std::vector::emplace_back`.


Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
```diff
@@ -282,19 +289,56 @@ class MeasurementProxyBase {
...
+  ///
+  template <typename OtherDerived>
+  void construct(const OtherDerived& other)
+    requires(!ReadOnly) && requires {
+      { other.size() } -> std::integral;
+      { other.sourceLink() };
+      { other.subspaceIndexVector() };
+      { other.parameters() };
+      { other.covariance() };
+    }
```
andiwand, line 324: 1. should we create a named concept for this?
2. also what kind of object can be fed into this?

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
```diff
@@ -485,4 +529,25 @@ class VariableMeasurementProxy
...
+
+template <std::size_t Size, typename... Args>
+MeasurementContainer::FixedProxy<Size>
+MeasurementContainer::emplaceFixedMeasurement(Args&&... args) {
+  FixedProxy<Size> meas = makeMeasurement<Size>();
+
+  meas.construct(std::forward<Args>(args)...);
+
+  return meas;
+}
```
andiwand, line 551: I wonder if it might be easier to inline define these functions

andiwand (commented):
Looks nice 👍 Two more requests
1. Could you add tests for this interface?
2. Would you be up to change the existing measurements creation in the Examples to the new interface?

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
stephenswat: I don't think that's possible, because the actual types for `FixedProxy` and `VariableProxy` are defined only _after_ the class definition, so you run into instantiations of undefined templates. :(

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
stephenswat: Basically just any measurement proxy.

stephenswat:
I went and:

1. Added tests for these new methods;
2. Converted all instances of `makeMeasurement` to these new methods;
3. Added a named measurement concept.

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
```diff
@@ -117,6 +119,12 @@ class MeasurementContainer {
     return getMeasurement<Size>(addMeasurement(Size));
   }
 
+  template <typename... Args>
+  VariableProxy emplaceVariableMeasurement(std::uint8_t size, Args&&... args);
+
+  template <std::size_t Size, typename... Args>
+  FixedProxy<Size> emplaceFixedMeasurement(Args&&... args);
```
andiwand, line 126: If we use concrete parameter types here we could overload an `emplaceFixedMeasurement` function for both cases. Nothing critical but that might align more with the rest of the API

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
```diff
@@ -282,19 +290,49 @@ class MeasurementProxyBase {
...
+      { this->self().setSubspaceIndices(std::forward<Subspace>(subspace)) };
+      { this->self().parameters() = parameters };
+      { this->self().covariance() = covariance };
+    }
+  {
+    setSourceLink(source_link);
+    self().setSubspaceIndices(std::forward<Subspace>(subspace));
+    self().parameters() = parameters;
+    self().covariance() = covariance;
+  }
```
andiwand, line 311: I wonder if `construct` is a good name for the function as it does not really construct something but rather fill all the information? Maybe `fill` makes more sense? Again not a real issue

andiwand (approved):
Nice! Two small remarks but nothing crucial

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
stephenswat: Good idea, I like "fill".

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
andiwand: What I mean is that there could be a single function called `emplace` or `emplaceMeasurement` and one has the overload with the size template for fixed size proxy returns. But I think that only works if you type out the parameters otherwise C++ will not let us overload it.

stephenswat: Ah right! Let me see if I can get that to work.

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
stephenswat: Done.

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
stephenswat: Wow, that went surprisingly easily.

andiwand (approved):
nice! thank you!

andiwand (approved):

acts-project-service:
## :red_circle: Athena integration test results [a129eebc1c617d7988da249f3929ff0dfb4394d5]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43984435) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8207882)!

PR #3638, opened 2024-09-23
docs: Add minimal Spack spec for ACTS core

stephenswat:
The current Spack documentation recommends a Spack spec which can be used to build ACTS and all its plugins; this is useful but, as we encountered over the weekend, also takes a lot of build time. As such, I wanted to add additional instructions for a simplified Spack build which only allows you to build the core ACTS library.

### DIFF
--- a/docs/misc/spack.md
+++ b/docs/misc/spack.md
@@ -33,22 +33,30 @@ the environment every time you open a new shell.
 With the Spack environment activated, you are ready to install ACTS'
 dependencies. You can either do this manually, or you can rely on Spack's
 definition of how to do this. To use the Spack definition, execute the
-following command (tuned for ACTS version 36.1.0):
+following command (tuned for ACTS version 36.1.0; make sure you update the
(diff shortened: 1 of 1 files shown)


AJPfleger (approved):
Looks good :)

On a side note - I am not super happy that the "build-all" command is a manual list of all (measured today) modules. Does spack provide any framework where we could define flags (maybe the same we use for the cmake presets). Then we could update the "build-all" command in the background and don't need to send around a new command, whenever we add/remove/rename any module.

acts-project-service:
## :red_circle: Athena integration test results [ee0b2f447e1657c5ee3e0f3f9c87a755c41dddbc]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43880981) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8192424)!

AJPfleger:
Broken Athena build is unrelated.

PR #3608, opened 2024-09-12
feat: add Detray surface grid converter

asalzburger:
This PR adds the surface grid converters drafted by @EleniXoch in #3607 and adds it to the overall `DetrayConverter`.

First tests show that the files written out contain the new surface grids.

Including the surface grids into the navigation reduces the time for the CPU navigation by a factor of  5.5 with identical results.

PR #3641, opened 2024-09-24
build: Mark Detray as a system dependency

stephenswat:
This change should silence warnings in the Detray source code, resolving the issue in #3608.

### DIFF
--- a/thirdparty/detray/CMakeLists.txt
+++ b/thirdparty/detray/CMakeLists.txt
@@ -15,7 +15,11 @@ message(STATUS "Building Detray as part of the Acts project")
 set(DETRAY_VERSION "v${_acts_detray_version}")
 
 # Declare where to get Detray from.
-FetchContent_Declare(Detray ${ACTS_DETRAY_SOURCE})
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.25.0)
+    FetchContent_Declare(Detray ${ACTS_DETRAY_SOURCE} SYSTEM)
+else()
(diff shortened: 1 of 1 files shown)


thirdparty/detray/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ message(STATUS "Building Detray as part of the Acts project")
 set(DETRAY_VERSION "v${_acts_detray_version}")
 
 # Declare where to get Detray from.
-FetchContent_Declare(Detray ${ACTS_DETRAY_SOURCE})
+FetchContent_Declare(Detray ${ACTS_DETRAY_SOURCE} SYSTEM)
```
andiwand, line 18: I feat this requires `CMake >= 3.25` while we require `3.14` right now https://github.com/acts-project/acts/blob/17b7b92f58ce143a745fb2b671cf3d5785fa2676/CMakeLists.txt#L1

stephenswat: Ah okay, I can add a case-statement for this.

stephenswat: Okay I added the check, but sadly the failing CI runs version 3.22, not 3.25. :cry:

andiwand: I think there are ways to work around it by creating a new target and declaring it as `SYSTEM` manually. We had a similar discussion with GeoModel update a month ago which was then resolved by pushing it to pre-built dependencies. But I guess that is not an option here...

PR #3648, opened 2024-09-25
fix: Reject material states when no measurements are found yet in Core CKF

andiwand:
Currently CKF tracks can have material states before any measurement was found. This is not true for outliers and holes. The logic is reworked so material, outliers and holes are handled in the same way if we have not found any measurements yet.

This is an attempt to also understand the output changes of https://github.com/acts-project/acts/pull/3391 better


andiwand:
@paulgessinger can we override sonarcloud here?

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [4a62f680413aa70af041174015d6c6a672dd95c6]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/43986956) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8208192)!

PR #3422, opened 2024-07-22
refactor: Rework initial `qOverP` sigma in Examples

andiwand:
- Use relative `pt` resolution as parameter. Get initial `qOverP` sigma by combining relative `pt` resolution and `theta` sigma
- Apply to `ParticleSmearing`
- Wire to python
- Disable smearing for truth smeared seeding to allow track finding encounter measurements

PR #3665, opened 2024-09-30
fix: Correct initial `q/p` covariance term in `TrackParamsEstimationAlgorithm` in Examples

andiwand:
I seem to have made a mistake calculating the covariance term on the first try in https://github.com/acts-project/acts/pull/3422. This is corrected here


asalzburger (dismissed):

asalzburger:
Changes in output are expected, I approve the change (as it seems correct), so let's wait for the physmon results.

asalzburger (approved):

Examples/Algorithms/TrackFinding/src/TrackParamsEstimationAlgorithm.cpp
```diff
@@ -56,7 +56,7 @@ Acts::BoundSquareMatrix makeInitialCovariance(
 
       // theta contribution
       variance +=
-          varianceTheta * std::pow(params[Acts::eBoundQOverP] *
+          varianceTheta * std::pow(params[Acts::eBoundQOverP] /
```
AJPfleger, line 59: We probably have different problems, when this happens, but shouldn't we protect us here against a division by zero?

andiwand: I would leave it up to the user to provide valid bound parameters

paulgessinger: Assertion maybe?

andiwand: Sure - just note that this would apply to many many more places

andiwand: this fits better in https://github.com/acts-project/acts/pull/3683

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/ParticleSmearing.cpp
```diff
@@ -143,7 +143,7 @@ ActsExamples::ProcessCode ActsExamples::ParticleSmearing::execute(
 
             // theta contribution
             variance += varianceTheta *
-                        std::pow(params[Acts::eBoundQOverP] *
+                        std::pow(params[Acts::eBoundQOverP] /
```
AJPfleger, line 146: Here we are the user and create a `theta` that might be `0`:
```
const auto [newPhi, newTheta] = Acts::detail::normalizePhiTheta(
    phi + sigmaPhi * stdNormal(rng), theta + sigmaTheta * stdNormal(rng));
```

Examples/Algorithms/TrackFinding/src/TrackParamsEstimationAlgorithm.cpp
AJPfleger: I am fine with having this division unchecked, because the function can not be called accidentally from outside.

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/ParticleSmearing.cpp
andiwand: This is outside my concern for this PR

AJPfleger: I started a new PR to fix this:
- https://github.com/acts-project/acts/pull/3687

AJPfleger (approved):
I think, we should also be concerned when introducing operations that might result in FPEs or invalid operations.

The divisions here are at least defined with ±∞ and NaN, and variances with those values probably handled well afterwards.

asalzburger (approved):

acts-project-service:
## :red_circle: Athena integration test results [a42ce7c116cea43d0f43b385171c14b5c63f698c]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/44290262) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8254639)!

PR #3676, opened 2024-10-01
refactor: Removed maxHits condition from ScoreBasedAmbiguitySolver

Ragansu:
The `maxHits` setting was used in cuts and `ambiscore`. For the `ambiscore` this was essential 

But `maxHits`  often set low for the `ambiscore` to work well. This would mean there will be removal of tracks in high eta region due to cuts. Now the `maxHits`  is removed from cuts hence solving the issue. Since we have optional cuts users can add it back with independent value if needed. 

Ragansu Chakkappai


Corentin-Allaire (approved):
Looks good for me.

PR #3675, opened 2024-10-01
feat(geo): TrackingVolume gets surface storage

paulgessinger:
Part of #3502 

This PR adds storage for surfaces (with shared ownership) inside tracking volumes.

Blocked by:
- https://github.com/acts-project/acts/pull/3673

Issue #3502, opened 2024-08-09
Gen3 geometry


paulgessinger:
Umbrella issue for the project to unify Gen1 (layer-based) and Gen2 (initial layer-less) geometry paradigms into a Gen3 (final) geometry.

See [this presentation](https://indico.cern.ch/event/1367908/contributions/5856261/attachments/2816681/4917402/2024-03-11_geometry-strategy_v1.pdf) for details.

## Associated PRs:

- https://github.com/acts-project/acts/pull/3500

## Initial strategy

```cpp

class Node {
    
    virtual std::shared_ptr<Volume> build();
    
    virtual std::vector<Portal> connect(
        const std::shared_ptr<TrackingVolume>& volume);
    
};
```

- Stack Volumes: `CylinderVolumeStack`, `CuboidVolumeStack`
    - Associated `CylinderContainerNode` and `CuboidContainerNode`
- `Volume::assignVolumeBounds()` becomes `virtual`
    - Overridden by stacks, used to resize child volumes
- Portal fusing + Portal stitching
- Special nodes called explicitly `CylinderLayer` etc?
    - Could also just be functions on `Node` for building purposes

```cpp
class CylinderContainerNode : public Node {
    ...
};
```

Two passes:

1. Recursively call `build()` on all children
    - Can build internal structure as configured
    - Can attach navigation delegate as configured
    - Returns a *representing volume* that can be a straight `TrackingVolume` or a *composite volume* like `CylinderVolumeStack`
    - Resizing `CylinderVolumeStack` resizes it's components, can create gaps on the outside if configured
    - Container nodes synchronize the representing volumes they receive, which propagates down the tree
    
2. Recursively call `connect()` with the current parent volume (starting from the root volume) as an argument
    - Regular nodes:
        - Register their `TrackingVolume` with the parent for ownership
        - Create a vector of `Portal` indexed by the common enum
        - Register the portals with their `TrackingVolume` for ownership and navigation
    - Containers nodes:
        - Call `connect()` on their children, but pass through their own parent `TrackingVolume`
        - No *real* volume is created for containers
        - They collect the portals, perform internal portal fusing + external portal stitching, then return the stitched portals
        - No portals / volumes need to be resized at this point
        - Portals are overwritten on child volumes using `PortalHandle`, internal registration remains valid, only outer registrations is delegated up the tree
    - When children have completed `connect()`, child `TrackingVolume`s have been registered: can create volume-local acceleration structure for finding
    
[...]

PR #3681, opened 2024-10-02
refactor(geo): Geometry visualization update

paulgessinger:
Blocked by:
- https://github.com/acts-project/acts/pull/3675

Related to #3502

---

The most significant changes include adding visualization methods to `TrackingGeometry` and `TrackingVolume`, updating the `Surface` class, and enhancing the `IVisualization3D` interface. Additionally, the `ObjVisualization3D` class has been refactored, and new default view configurations have been defined.

### Visualization Enhancements:
* [`TrackingGeometry.hpp`](diffhunk://#diff-cc497a4ee5032615db90ef065e9466279a0fde38fbd201840a32d783d3d6ec4aR148-R158): Added a `visualize` method to enable visualization of tracking geometry, including substructures. (`[Core/include/Acts/Geometry/TrackingGeometry.hppR148-R158](diffhunk://#diff-cc497a4ee5032615db90ef065e9466279a0fde38fbd201840a32d783d3d6ec4aR148-R158)`)
* [`TrackingVolume.hpp`](diffhunk://#diff-835fb549fb77cdaa632a4e2131c8dc17ad4973ab3f1600a4608582b706a3f68eR326-R344): Added methods to visualize tracking volumes and manage surfaces, including `visualize`, `surfaces`, and `addSurface`. (`[[1]](diffhunk://#diff-835fb549fb77cdaa632a4e2131c8dc17ad4973ab3f1600a4608582b706a3f68eR326-R344)`, `[[2]](diffhunk://#diff-835fb549fb77cdaa632a4e2131c8dc17ad4973ab3f1600a4608582b706a3f68eR477-R487)`, `[[3]](diffhunk://#diff-835fb549fb77cdaa632a4e2131c8dc17ad4973ab3f1600a4608582b706a3f68eR550)`)

### Interface Updates:
* [`IVisualization3D.hpp`](diffhunk://#diff-c56ccee98d351daf6ac595cd265d30716a594ceb74030138411595aca66fdc39R75-R80): Introduced a destructor and a new `object` method to start a new object context. (`[Core/include/Acts/Visualization/IVisualization3D.hppR75-R80](diffhunk://#diff-c56ccee98d351daf6ac595cd265d30716a594ceb74030138411595aca66fdc39R75-R80)`)
* [`ObjVisualization3D.hpp`](diffhunk://#diff-fab69ace861a3e5d8fc173e8627732777a5fdd52703e9e162a14db563ea56a2aL29-R32): Refactored to use `double` as the value type and added support for object contexts. (`[[1]](diffhunk://#diff-fab69ace861a3e5d8fc173e8627732777a5fdd52703e9e162a14db563ea56a2aL29-R32)`, `[[2]](diffhunk://#diff-fab69ace861a3e5d8fc173e8627732777a5fdd52703e9e162a14db563ea56a2aR76-R102)`)

### Configuration and Defaults:
* [`ViewConfig.hpp`](diffhunk://#diff-e403e98244e75d6b7d32864417c37619de6becee73f53415b733073f655f0929R114-R117): Defined new default view configurations for surfaces, portals, volumes, and other elements. (`[[1]](diffhunk://#diff-e403e98244e75d6b7d32864417c37619de6becee73f53415b733073f655f0929R114-R117)`, `[[2]](diffhunk://#diff-e403e98244e75d6b7d32864417c37619de6becee73f53415b733073f655f0929R139-R146)`)

### Codebase Cleanup:
* Removed obsolete and redundant code from `ObjVisualization3D.ipp`. (`[Core/include/Acts/Visualization/detail/ObjVisualization3D.ippL1-L179](diffhunk://#diff-29bbb6e8cfcb388c51d1e88237dd0c7d582762d7e05a0ec12278bdc336b0aaa1L1-L179)`)


Core/include/Acts/Visualization/ViewConfig.hpp
```diff
@@ -49,17 +49,31 @@ struct Color {
...
+    for (char c : hex) {
+      value = (value << 4) + hexCharToInt(c);
+    }
+    return value;
+  };
+
+ public:
+  /// Constructor from hex string. The expected format is `#RRGGBB`
+  /// @param hex The hex string
+  constexpr Color(std::string_view hex) {
```
paulgessinger, line 76: I intentionally made this non `explicit`, because it makes code in GCC more verbose, but nor sonar complains about this.

Examples/Python/src/Output.cpp
```diff
@@ -130,10 +131,17 @@ void addOutput(Context& ctx) {
         .def(py::init<>())
         .def(py::init<int, int, int>())
         .def(py::init<double, double, double>())
-        .def(py::init<std::string>())
+        .def(py::init<std::string_view>())
         .def_readonly("rgb", &Color::rgb);
   }
 
+  py::class_<IVisualization3D>(m, "IVisualization3D")
+      .def("write", [](const IVisualization3D& self, const py::object& arg) {
```
paulgessinger, line 139: This actually prevents materials from working. I need to change this back to path.

Core/include/Acts/Geometry/TrackingGeometry.hpp
```diff
@@ -145,6 +145,17 @@ class TrackingGeometry {
   const std::unordered_map<GeometryIdentifier, const Surface*>&
   geoIdSurfaceMap() const;
 
+  /// Visualize a tracking geometry including substructure
```
benjaminhuth, line 148: Would it be also an option to not couple the `TrackingGeometry` directly to the visualization interface, but rather do it as for surfaces, layers, etc. in `GeometryView3D`? Or are there reasons to do this directly here? (same for `TrackingVolume`)

Core/include/Acts/Visualization/PlyVisualization3D.hpp
```diff
@@ -60,6 +60,10 @@ class PlyVisualization3D : public IVisualization3D {
   /// @copydoc Acts::IVisualization3D::clear()
   void clear() final;
 
+  void object(const std::string& /*name*/) final {
```
benjaminhuth, line 63: Should this throw a `std::runtime_error`?

Core/include/Acts/Visualization/IVisualization3D.hpp
```diff
@@ -72,6 +72,12 @@ class IVisualization3D {
   /// Remove all contents of this helper
   ///
   virtual void clear() = 0;
+
+  virtual ~IVisualization3D() = default;
+
+  /// Start a new object context
+  /// @param name The name of the object
+  virtual void object(const std::string& name) = 0;
```
benjaminhuth, line 80: Hmm could we find a more descriptive name like `newNamedContext(...)` or so?

Core/include/Acts/Visualization/ViewConfig.hpp
```diff
@@ -49,17 +49,31 @@ struct Color {
   constexpr Color(double r, double g, double b)
       : Color{std::array<double, 3>{r, g, b}} {}
 
+ private:
+  constexpr static int hexToInt(std::string_view hex) {
```
benjaminhuth, line 53: Hmm this should be `constexpr`? because `std::stoi(hex, 16)` should also work I think otherwise...

benjaminhuth (commented):
Looks good, I have a few comments.

Core/include/Acts/Geometry/TrackingGeometry.hpp
paulgessinger: My (original) idea is that `IVisualization3D` is the interface abstraction, and it abstracts the concrete output format. In my opinion, it's fine to have the interface of the geometry objects couple to that. It also (I think) simplifies where this logic lives.

The idea is to remove `GeometryView3D` eventually.

I plan to follow this same pattern for JSON conversion (which is a bit confusing right now).

Core/include/Acts/Visualization/PlyVisualization3D.hpp
paulgessinger: It could be, but I was thinking that since PLY still produces a valid output file even when objects are being used, it would make sense to have this be silent.

Core/include/Acts/Visualization/IVisualization3D.hpp
paulgessinger: I was going off of the naming in OBJ, where it's called `object`. I can change the name if you want, but until we have a second (real) format (PLY isn't really used much) I would keep the level of abstraction somewhat close to OBJ. What do you think?

Core/include/Acts/Visualization/ViewConfig.hpp
paulgessinger: Yeah, I'd like this to be `constexpr` if possible. It's not critical, but I think it's nice having the strings (potentially) be converted at compile-time.

benjaminhuth: I'm always hesitant a bit when we reimplement such basic things for not performance critical code, but don't have a very strong opinion on it...

Core/include/Acts/Geometry/TrackingGeometry.hpp
benjaminhuth: Hmm so I think from my personal taste I would prefer the other way around, because it keeps classes small and focussed, but this is also fine of course.

Core/include/Acts/Visualization/IVisualization3D.hpp
benjaminhuth: fine for me

Core/include/Acts/Visualization/PlyVisualization3D.hpp
benjaminhuth: We don't have a logger in the class, right?

paulgessinger: Currently not. We could add it, but outside of this so far there hasn't been much need for output messages.

Core/include/Acts/Geometry/TrackingGeometry.hpp
paulgessinger: I'm in any case planning to push this further to allow (eventually) full serialization and visualization of volume-internal navigation policies. I think the current approach with lots of downcasting has shown this to be very complex, which I'd like to solve by doing it the other way round.

Core/include/Acts/Visualization/ViewConfig.hpp
paulgessinger: This is an internal function, so in the absence of a `constexpr` library function, I think the fallout from reimplementing is not that large.

paulgessinger:
Can you approve @benjaminhuth?

benjaminhuth (approved):

PR #3688, opened 2024-10-03
refactor!: Do not use geometry extent during seeding

CarloVarni:
In the seeding we only need the radius range of the space points (and only when the range is computed by the code and not provided by the user). As of now, when filling the grid we waste a lot of time to fill the `Acts::Extent` object (for all the possible `Acts::BinningValue` even if we don't need them all).

I have refactored the code so that we do not do this expensive operation. The user will have to do it by them selves or provide a validity search window via JO.

No changes expected from this PR


CarloVarni:
don't mind sonarclound. The issues would require extensive changes. I'd like to decouple that fro this PR @paulgessinger

Examples/Algorithms/TrackFinding/src/SeedingAlgorithm.cpp
```diff
@@ -243,16 +242,32 @@ ActsExamples::ProcessCode ActsExamples::SeedingAlgorithm::execute(
...
+      continue;
+    }
+    const auto* firstEl = coll.front();
+    const auto* lastEl = coll.back();
+    if (firstEl->radius() < minRange) {
+      minRange = firstEl->radius();
+    }
+    if (lastEl->radius() > maxRange) {
+      maxRange = lastEl->radius();
+    }
```
paulgessinger, line 269: Nitpick, this could be
```suggestion
    minRange = std::min(firstEl->radius(), minRange);
    maxRange = std::max(lastEl->radius(), maxRange);
```

I believe.

Examples/Algorithms/TrackFinding/src/SeedingAlgorithmHashing.cpp
```diff
@@ -269,15 +268,32 @@ ActsExamples::ProcessCode ActsExamples::SeedingAlgorithmHashing::execute(
...
+        continue;
+      }
+      const auto* firstEl = coll.front();
+      const auto* lastEl = coll.back();
+      if (firstEl->radius() < minRange) {
+        minRange = firstEl->radius();
+      }
+      if (lastEl->radius() > maxRange) {
+        maxRange = lastEl->radius();
+      }
```
paulgessinger, line 295: Same applies here, but again, not important.

paulgessinger (approved):
LGTM

Examples/Algorithms/TrackFinding/src/SeedingAlgorithm.cpp
CarloVarni: good point. Done

AJPfleger (dismissed):

CarloVarni:
I had to switch to `float` for the computation (also downstream expects `floats`)

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [81e129bc52d420a85b7e8f381c865affd951024b]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/44317773) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8259075)!

PR #3693, opened 2024-10-04
fix: Copy whole trackstate did not copy calibrated

benjaminhuth:
Know your defaults...


andiwand (dismissed):

benjaminhuth:
Yeah I would agree... Any thoughs @paulgessinger ?

benjaminhuth:
@paulgessinger I added a test case that only checks the copy of the calibrated field of two track states, that also confirmed the bug we found. The test is very specific to this error, is it okay like that?

paulgessinger (approved):

andiwand (approved):

paulgessinger (approved):

PR #3698, opened 2024-10-07
refactor: Do not insert space points if not inside grid boundaries

CarloVarni:
Changes:
- Calling `grid.isInside` to check if space point falls within the boundaries
- Check min and max values are correct (i.e. min < max)
- Additional changes to check phi is within `[-std::numbers::pi`, `std::numbers::pi`]
- Move config and options values to `double` (also because units are `double`s). Only partial since moving them all requires change in other parts of the code (will follow up in another PR)
- Use `std::numbers::pi` instead of `M_PI`

In the example, we go back to the previous computation for the `rMiddleSPRange`. Not clear why the `std::floor` was introduced in the first place, going back to that computation while we figure out why


Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -412,8 +413,8 @@
                 maxSharedTracksPerMeasurement=2,
                 pTMax=1400,
                 pTMin=0.5,
-                phiMax=3.14,
-                phiMin=-3.14,
```
andiwand, line 416: lol

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -89,7 +90,7 @@
         "maxPhiBins",
         "phi",  # (min,max)
     ],
-    defaults=[None] * 6 + [(None, None)] * 1,
+    defaults=[None] * 6 + [(-math.pi, math.pi)],
```
andiwand, line 93: Usually we use `None` here to default to the C++ options. Any reason to default it twice?

CarloVarni: right, this can indeed go back to `None`. It's a remnant of my testings

Core/include/Acts/Seeding/detail/CylindricalSpacePointGrid.hpp
```diff
@@ -56,9 +57,9 @@ struct CylindricalSpacePointGridConfig {
   // maximum impact parameter in mm
   float impactMax = 0 * Acts::UnitConstants::mm;
   // minimum phi value for phiAxis construction
-  float phiMin = -M_PI;
+  double phiMin = -std::numbers::pi;
```
CarloVarni, line 60: @andiwand @paulgessinger concerning this. What is your opinion? `double` or `float`?
We can have as well `std::numbers::pi_v<float>` I believe.

I'd eventually go to `double` since `mm` and units are `double` ... but moving everything to `double` will be quite an extensive change in the end.

andiwand: If we really do all the calculations here with `float` I would stick with that type to avoid runtime conversion

Examples/Python/python/acts/examples/reconstruction.py
CarloVarni: done

Core/include/Acts/Seeding/detail/CylindricalSpacePointGrid.hpp
CarloVarni: done

andiwand (approved):

PR #3702, opened 2024-10-09
fix: Implement `DirectNavigator` direction handling

andiwand:
Our current implementation of the `DirectNavigator` does not handle the propagation direction which is essential if we for example reverse direction to smooth track in the Kalman filter.

Apart from that I added a search mechanism in the initialization for the starting surface.

---

This pull request to `Core/include/Acts/Propagator/DirectNavigator.hpp` includes changes to improve the navigation logic in the `DirectNavigator` class. The most important changes include adding new utility methods for surface navigation, updating the surface index handling, and enhancing logging for better debugging.

Enhancements to navigation logic:

* Added utility methods `navSurface`, `nextSurface`, `endOfSurfaces`, and `remainingSurfaces` to the `DirectNavigator` class to streamline surface navigation. (`Core/include/Acts/Propagator/DirectNavigator.hpp`) [[1]](diffhunk://#diff-56afaecc4b563cfe416f8615f35dd5ac21f706869dd2c164dd326087b3edd655R77-R100) [[2]](diffhunk://#diff-56afaecc4b563cfe416f8615f35dd5ac21f706869dd2c164dd326087b3edd655L138-R192) [[3]](diffhunk://#diff-56afaecc4b563cfe416f8615f35dd5ac21f706869dd2c164dd326087b3edd655L166-R215) [[4]](diffhunk://#diff-56afaecc4b563cfe416f8615f35dd5ac21f706869dd2c164dd326087b3edd655L186-R229) [[5]](diffhunk://#diff-56afaecc4b563cfe416f8615f35dd5ac21f706869dd2c164dd326087b3edd655L227-R279) [[6]](diffhunk://#diff-56afaecc4b563cfe416f8615f35dd5ac21f706869dd2c164dd326087b3edd655L255-R299)

Updates to surface index handling:

* Changed the type of `surfaceIndex` from `std::size_t` to `int` to support bidirectional navigation. (`Core/include/Acts/Propagator/DirectNavigator.hpp`)

Logging improvements:

* Enhanced logging in the `initialize` method to provide more detailed information about the start surface and surface sequence. (`Core/include/Acts/Propagator/DirectNavigator.hpp`)


Core/include/Acts/Propagator/DirectNavigator.hpp
```diff
@@ -63,7 +65,7 @@ class DirectNavigator {
     Options options;
 
     /// Index of the next surface to try
-    std::size_t surfaceIndex = 0;
+    int surfaceIndex = 0;
```
benjaminhuth, line 68: Why no `std::size_t`?

Core/include/Acts/Propagator/DirectNavigator.hpp
```diff
@@ -135,17 +166,41 @@ class DirectNavigator {
...
     }
 
     // We set the current surface to the start surface
     state.navigation.currentSurface = state.navigation.options.startSurface;
-    if (state.navigation.currentSurface) {
+    if (state.navigation.currentSurface != nullptr) {
       ACTS_VERBOSE("Current surface set to start surface "
                    << state.navigation.currentSurface->geometryId());
+    } else {
+      ACTS_ERROR("No start surface provided.");
```
benjaminhuth, line 180: Hmm this does error, but continues? What does happen without start surface?

benjaminhuth (commented):
Looks good, just a few comments!

Core/include/Acts/Propagator/DirectNavigator.hpp
andiwand: I need a signed type because the index can now go below `0`

Core/include/Acts/Propagator/DirectNavigator.hpp
andiwand: I guess I could also throw an exception but I felt like `ACTS_ERROR` is what we usually do in the propagation. I think we will just end up running towards the next surface in the stream

benjaminhuth: Hmm I just wonder if we e.g. initialize should return `Result<void>` instead of `void`, to make this properly fail the propagation. But this might be out of the scope of that PR...

andiwand: Yeah I think we do not really forward any hard errors to the propagator at the moment but this is something we can improve on. I think I can put this check into the `makeState` which would cause the failure to pop up earlier and there an exception seems to make more sense.

benjaminhuth: Yeah I think we should be able to return errors, but this would be a larger refactor indeed. Seems a good solution to check it in the `makeState`. Then we could replace the error case with an `assert`?

benjaminhuth (approved):
Looks good!

andiwand:
I would hold this back until https://github.com/acts-project/acts/pull/3733 is in so we can check the physmon

benjaminhuth (approved):

acts-project-service:
## :red_circle: Athena integration test results [7436f407c06b00f785d4fd652d8662c7efdea029]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/44775709) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8331718)!

PR #3703, opened 2024-10-09
fix: `material_recoding.py` conditional loading of GeoModel

paulgessinger:
The previous method of a nested `import` does not actually work.


benjaminhuth (approved):
Good solution!

Examples/Scripts/Python/material_recording.py
```diff
@@ -114,8 +120,6 @@ def main():
             acts.examples.geant4.GdmlDetectorConstructionFactory(args.input)
         )
     elif args.input.endswith(".sqlite") or args.input.endswith(".db"):
-        import acts.examples.geant4.geomodel
```
andiwand, line 117: out of curiosity: why did this not work?

PR #3710, opened 2024-10-10
fix: Make charge smearing optional in digi config

stephenswat:
Currently, the JSON reading for digitization configurations unconditionally reads the charge smearing data, but this need not necessarily be present; the geometric config for the ODD, for example, does not have this data. As a result, parser was failing to parse the geometric ODD digitization config. This commit makes the reading conditional and adds tests to ensure the ODD config reading remains functional.


paulgessinger:
No idea why the Actsvg download fails.

stephenswat:
Let's see if this is transient.

paulgessinger (approved):

PR #3728, opened 2024-10-12
docs: Add comment on computation of impact parameter from the doublet

CarloVarni:
One of the criterion used in doublet finding relies on computing the minimal distance between the origin and the straight line connecting the two points of the doublet. This distance is compared against a maximum impact parameter (`impactMax`) provided by the user. 

Let's consider for instance the middle-top doublet search. 
Geometry

In the attached image:
- `A` is the origin
- `B` is the middle space point, thus `AB = rM` (radius of middle space point) 
- `C` is the top space point

What the code does is to compute `BD` (`xNewFrame` in the code) and `CD` (`yNewFrame` in the code)  and then a simple geometric similarity to compute the impact parameter. 

The code currently does: `Im / CD = AB / BD`, where `Im` is the impact parameter we are trying to compute.  Thus `Im = (AB * CD) / BD <= impactMax`. Unfortunately `Im` is equal to `AF`, while we really want to compute `AE`. The proper computation would be `Im / CD = AB / BC`, which leads to `Im = (AB * CD) / BC <= impactMax`. In this case `Im = AE`.

However, `BC` has to be computed from `CD` and `BD`. In order to avoid square roots I'm using square quantities. 

It is also possible that the code assumes that the `alpha` angle is too small to make a difference and thus uses an approximation... But there are no comments in the code about this

### DIFF
--- a/Core/include/Acts/Seeding/SeedFinder.ipp
+++ b/Core/include/Acts/Seeding/SeedFinder.ipp
@@ -369,10 +369,13 @@ SeedFinder<external_spacepoint_t, grid_t, platform_t>::getCompatibleDoublets(
       // We check the interaction point by evaluating the minimal distance
       // between the origin and the straight line connecting the two points in
       // the doublets. Using a geometric similarity, the Im is given by
-      // yNewFrame * rM / deltaR <= m_config.impactMax We use squared quantities
-      // to avoid square roots
-      if (yNewFrame * yNewFrame * rM * rM <=
-          m_config.impactMax * m_config.impactMax * deltaR2) {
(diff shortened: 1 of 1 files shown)


andiwand (approved):
LGTM 👍

acts-project-service:
## :red_circle: Athena integration test results [02b779b2b467739f4cf67d857e442029decf8cee]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/44763708) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8330165)!

Issue #3738, opened 2024-10-16
GSF throws warnings regularly with ODD


andiwand:
I see the following warning quite often with the GSF and the ODD:

```
Encountered 47 cases where x/X0 exceeds the range of the Bethe-Heitler-Approximation. The maximum x/X0 encountered was 2.0238. Enable DEBUG output for more information.
```

To avoid CI failures I had to reduce the number of events we run in `test_examples` with https://github.com/acts-project/acts/pull/3733.

@benjaminhuth mentioned that we may have to re-parametrize the energy loss polynom or to understand why we encounter such thick material in the ODD.

andiwand:
https://github.com/acts-project/acts/pull/3739 improves the situation by clamping `X/X0` to the valid range of the approximation.

Still open
- [ ] investigate where and why we exceed the valid input range of the approximation
- [ ] define a procedure to train/retrain the approximation
- [ ] retrain the approximation for the ODD if necessary

AJPfleger:
devmeeting: The problem is not ODD specific and can also be seen in ATHENA on the ITk.

andiwand:
Duplicate of https://github.com/acts-project/acts/issues/1748

PR #3739, opened 2024-10-16
feat: Make it possible to turn off warnings in GSF

benjaminhuth:
This adds an option that should disable the warnings appearing in the GSF in a meaningful way...

Worksaround but does not solve #3738


andiwand (dismissed):
Makes sense from my side. I think this is best we can do for now 👍

Examples/Python/src/TrackFitting.cpp
```diff
@@ -106,12 +107,17 @@ void addTrackFitting(Context& ctx) {
...
-        .def_static("makeDefault",
-                    []() { return Acts::makeDefaultBetheHeitlerApprox(); });
+        .def_static(
+            "loadFromFiles", &ActsExamples::BetheHeitlerApprox::loadFromFiles,
+            "lowParametersPath"_a, "highParametersPath"_a, "lowLimit"_a = 0.1,
+            "highLimit"_a = 0.2, "clampToRange"_a = false)
+        .def_static(
+            "makeDefault",
+            [](bool clampToRange) {
+              return Acts::makeDefaultBetheHeitlerApprox(clampToRange);
```
paulgessinger, line 117: Just a question: why bind this as a scoped static function instead of a free function like it is in C++? Or alternatively, why not make the free function a scoped static like in Python?

benjaminhuth: The main reason is, that the C++ class `Acts::AtlasBetheHeitlerApproximation` is templated, and for the python binding a specific choice of template parameters is choosen. So in contrary to the C++ case it makes sense to have it as a static factory as part of the class.
But yeah, the differences between C++ and binding are also not optimal.
Given that the CI is green now I would like to keep it for now...

andiwand (approved):

acts-project-service:
## :red_circle: Athena integration test results [41066dc3e90520693fe52baafe494a46e62169db]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/44857320) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8344331)!

PR #3750, opened 2024-10-17
refactor: Add helper for Transform inversion

paulgessinger:


andiwand:
Nice!

I wonder if we should make a custom transform object in the future which only carries translation and rotation where we can also put this shortcut into place.

andiwand:
at least in my usual tests with https://github.com/andiwand/cern-scripts/tree/main/acts/perf/ckf it seems not to have a measurable impact
image
image

PR #3756, opened 2024-10-17
feat: Validate track parameters

andiwand:
Adds two free functions which allow checking `BoundVector`s and `FreeVectors`s for valid track parameters. This is then used as `assert`s in the track parameter EDM.

blocked by
- https://github.com/acts-project/acts/pull/3777


andiwand:
https://github.com/acts-project/acts/pull/3756/commits/0a02484e43c7683ca69cd49eb12ddf8151fc47f8 contains your feedback @paulgessinger

andiwand:
This fires basically everywhere. It would take quite some time to fix every single problem. I think we can just add the functionality and not assert on it for now. The question is if we really want to assert on it. $\phi $ being out of range usually does not hurt. With $\theta$ it gets trickier.

andiwand:
This is a bit frustrating: we have a bunch of invalid track parameters in the code which are generated randomly in different ways. I think it would be good to have one mechanism of generating random track parameters which has to be parameterized before we can continue with this PR.

Core/include/Acts/Utilities/AngleHelpers.hpp
```diff
@@ -9,16 +9,51 @@
...
+  static constexpr Scalar maxTheta = std::numbers::pi_v<Scalar>;
+
+  static constexpr Scalar minEta = -std::numeric_limits<Scalar>::infinity();
+  static constexpr Scalar maxEta = std::numeric_limits<Scalar>::infinity();
+};
+
+template <>
+struct EtaThetaConversionTraits<float> {
+  static constexpr float minTheta = 0;
+  static constexpr float maxTheta = std::numbers::pi_v<float> - 1e-6f;
```
timadye, line 28: Why are these all specialised? I guess this is still experimental...

How about this:
```
template <Float>
struct EtaThetaConversionTraits<Float> {
  static constexpr Float minTheta = std::numeric_limits<Float>::min();
  static constexpr Float maxTheta = std::numbers::pi_v<Float> - std::numeric_limits<Float>::epsilon();
  static constexpr Float minEta = -std::numeric_limits<Float>::infinity();
  static constexpr Float maxEta = std::numeric_limits<Float>::infinity();
};
```

I checked `maxTheta<float>` gives &eta;=-16.39916 and `maxTheta<double>` gives &eta;=-37.33185619326892, which look reasonable (as does `long double`). I don't think it is possible to find a value of `maxTheta` that gives &eta;=`-inf`, but at least we can ensure it is not `nan`.

For `minTheta`, I think `0` (&eta;=`inf`) or `std::numeric_limits<Float>::min()` (&eta;=88.02969) are both reasonable.

Note that `epsilon()` is only strictly correct at 1.0, but it seems to do something reasonable for &pi;.

You could also do something clever with `concept`s for better error messages if ever used for non-FP types, if you want.

Core/include/Acts/Utilities/AngleHelpers.hpp
```diff
@@ -29,6 +64,13 @@ Scalar etaFromTheta(Scalar theta) {
 /// @return the polar angle in radian towards the z-axis.
 template <typename Scalar>
 Scalar thetaFromEta(Scalar eta) {
+  if (eta < EtaThetaConversionTraits<Scalar>::minEta) {
+    return std::numbers::pi_v<Scalar>;
+  }
+  if (eta > EtaThetaConversionTraits<Scalar>::maxEta) {
+    return 0;
+  }
+
```
timadye, line 73:
```suggestion
```
I don't think these checks do anything, or are necessary. The `thetaFromEta()` function works over the full range.

However, if you want to ensure that the returned &theta; value can be converted back with `etaFromTheta()` to non-`nan`, we would have to try something else.

Core/include/Acts/Utilities/AngleHelpers.hpp
andiwand: I was experimenting with it and quite surprised that almost all the values worked out of the box. Usually `float` and `double` behave a little different so I am not sure if the `epsilon` will work. The idea was to capture the "last" value which works before the operations would throw FPEs. For `float` I found this boundary while for all others it worked out of the box.

Haven't checked if it throws FPEs yet.

Core/include/Acts/Utilities/AngleHelpers.hpp
andiwand: I didn't check for FPEs yet but otherwise it seemed to work yes

Core/include/Acts/Utilities/AngleHelpers.hpp
timadye: I have been thinking about this... One problem is that these details of the FP rounding at the limits of precision are quite fragile. They could change with compiler or library version, or CPU type (x86_64, ARM, or even GPU) or version. I was testing with gcc godbolt on x86_64. Maybe you were using clang on macOS ARM?

Whatever. I don't know much about forward physics, but I don't think people would be using pseudorapidity in the very forward region. If we limit &eta; anywhere &NestedGreaterGreater;4, we should be fine. Can I propose something like
```
static constexpr Float thetaFromEtaNoChecked(Float eta) { return 2 * std::atan(std::exp(-eta)); }
static constexpr Float maxEta = 10;  // arbitrary large eta
static constexpr Float minEta = -maxEta;
static constexpr Float minTheta = thetaFromEtaNoChecked(maxEta);
static constexpr Float maxTheta = thetaFromEtaNoChecked(minEta);
```
and then include the checks in `thetaFromEta()` and `etaFromTheta()` substituting the `minTheta` etc if outside the limits.

One problem with this simple definition is that `std::atan` and `std::exp` aren't defined `constexpr` until (probably) C++26. gcc doesn't complain, but clang does. For the moment, I think it's fine to put in
```
static constexpr Float minTheta = static_cast<Float>(9.0799859462585550026e-05L);  // 2*std::atan(std::exp(maxEta))`
static constexpr Float maxTheta = static_cast<Float>(3.1415018537303306529L);  // 2*std::atan(std::exp(minEta))`
```
This won't be exact (what is?), but since they are limits, I don't think it really needs to be.

andiwand: I think the round trip will not produce the same numbers anyways so we do not need to be too precise with the boundaries. I thought it would be nice to extend the range based on the floating point type since it might actually reach further with double but maybe you are right and there is no use case in converting these shallow angles anyways and one can just cut at 10.

andiwand:
Since the angle helper changes are quite a stretch I pulled the changes here @timadye https://github.com/acts-project/acts/pull/3788

paulgessinger (approved):

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [9c0bc4719c10c079235c40b4c6dcffd10c14a147]
Please investigate the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8408416)!

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262221)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262220)||
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262219)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262218)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262217)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262216)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262215)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262214)||
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262213)||
|:green_circle: | [test_ActsCheckObjectCountsCachedWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262212)||
|:green_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262211)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262210)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262209)||
|:red_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262208)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262207)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262206)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262205)||
|:green_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262204)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262203)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262202)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262201)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262200)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262199)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262198)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262197)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262196)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262195)||
|:red_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262194)||
|:red_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262193)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262192)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262191)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262190)||
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45262189)||

PR #3757, opened 2024-10-17
feat: ProxyAccessor constructor can now be constexpr

timadye:
Since C++20 supports `constexpr std::string`, we can make `ProxyAccessorBase(const std::string& _key)` constructor `constexpr`.


timadye:
> Makes sense. There's one unit test failure on macOS related to an unused variable.

So the warning is [here](https://github.com/acts-project/acts/blob/main/Tests/UnitTests/Core/EventData/TrackContainerComplianceTests.cpp#L77):
```
/Users/runner/work/acts/acts/build/Tests/UnitTests/Core/EventData/TrackContainerComplianceTests.cpp:77:36: error: unused variable 'caccNMeasuements' [-Werror,-Wunused-variable]
  ConstProxyAccessor<unsigned int> caccNMeasuements("nMeasurements");
```
Indeed, that variable is not used, because it is just used to check that you can't modify through a `ConstProxyAccessor`. I don't understand why this warning wasn't always there, and only appears now this constructor is `constexpr`.

That aside, do you have any ideas on how to fix this?

paulgessinger:
In principle there should be a fix for this already. Maybe it's not merged yet

timadye:
@AJPfleger I think this is waiting for your approval, if you do approve. Thanks!

AJPfleger (approved):

acts-project-service:
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45695015) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8468781)!

PR #3759, opened 2024-10-18
ci: Enable standard library assertions in debug builds

benjaminhuth:
Should add assertions that check preconditions of calls to standardlibrary functions.

## Summary by CodeRabbit

- **Chores**
	- Enhanced configuration for compiler flags to conditionally enable assertions.
	- Addressed compilation issues with GCC 13 by undefining specific preprocessor directives.
	- Maintained existing logic for compiler type checks without altering public entity signatures.
- **Bug Fixes**
	- Adjusted the logic for generating random directions to ensure correct bounds for angles.
- **Documentation**
	- Improved comments for clarity on the calculations in the random direction generation function.


paulgessinger:
Maybe we should put this behind an option and enable it explicitly in the CI? That way we wouldn't silently modify local Debug builds.

Also, is there an equivalent for clang / libc++?

cmake/ActsCompilerOptions.cmake
```diff
@@ -13,6 +13,18 @@ set(cxx_flags
...
 )
 
+# Add assertions to standard libraries for debug builds
+if(
+    (CMAKE_BUILD_TYPE STREQUAL RelWithDebInfo)
+    OR (CMAKE_BUILD_TYPE STREQUAL Debug)
+)
+    if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+        set(cxx_flags "${cxx_flags} -D_GLIBCXX_ASSERTIONS")
+    elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
paulgessinger, line 23: I don't think clang implies libc++. I *think* on many linux distros clang compiles with libstdc++.

benjaminhuth:
Hmm I added it also for clang... but actually, I do not check which stdlib is used. Maybe we should add both macros unconditionally of the `CMAKE_COMPILER_ID`.

Regarding the option, you mean to avoid rebuild after this PR?

cmake/ActsCompilerOptions.cmake
benjaminhuth: right, I just realized this. As written below, I would add both flags unconditionally, because they should not hurt if the other stdlib is used I guess...

paulgessinger:
@benjaminhuth I don't think we should unexpectedly enable any assertions outside of the CI, even in `Debug` builds. But if you gate this with `ACTS_FORCE_ASSERTIONS`, which we turn on in the CI, we could still get the assertions there.

benjaminhuth:
Okay yeah that makes sense, `ACTS_FORCE_ASSERTIONS` seems a good place to do this I haven't though of.

benjaminhuth:
But just out of interest @paulgessinger: To me every standard library assertion would be a violated precondition, so what would be the issue with enabling them for `RelWithDebMode`?

GCC documentation even recommendes them for production builds actually (https://gcc.gnu.org/wiki/LibstdcxxDebugMode)

paulgessinger:
@benjaminhuth I don't see it being recommended. It's **HUGE** change to enable e.g. bounds checking quietly. ATLAS builds with `RelWithDebInfo`, so we'd force any compilation unit compiled in Athena to have these assertions on. That'd be super disruptive.

benjaminhuth:
Yeah I agree that its a huge change, but would be a good change in my opinion. But yeah, I wasn't aware that by default acts is build in RelWithDebInfo for production builds in ATLAS. Thats of course a different story then...

paulgessinger:
> Yeah I agree that its a huge change, but would be a good change in my opinion.

Sure, but there's no way we can decide this as a library.

andiwand:
Also this does not come for free right? I guess the standard lib functions will do extra checks which are skipped otherwise. So it is a bit like with our assertions while ours might be more expensive

paulgessinger:
@andiwand I believe these are reasonably light checks, like bounds checks or smart pointer `nullptr` checks, which a reasonable branch predictor will probably get right most of the time (at least that's the justification for this in Rust land). So @benjaminhuth is probably right that they're not terrible for performance and are likely safe for production use.

paulgessinger:
Seems like the detray conversion code does some out-of-bounds `std::array` access, which ostensibly with these flags are checked at compile-time.

```
/usr/include/c++/13/bits/stl_algobase.h:437:30: error: 'void* __builtin_memcpy(void*, const void*, long unsigned int)' forming offset [16, 17] is out of the bounds [0, 16] of object '<anonymous>' with type 'const double [2]' [-Werror=array-bounds=]
  437 |             __builtin_memmove(__result, __first, sizeof(_Tp) * _Num);
      |             ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/__w/acts/acts/Plugins/Json/src/DetrayJsonHelper.cpp: In function 'std::tuple<unsigned int, std::vector<double, std::allocator<double> > > Acts::DetrayJsonHelper::maskFromBounds(const Acts::SurfaceBounds&, bool)':
/__w/acts/acts/Plugins/Json/src/DetrayJsonHelper.cpp:46:47: note: '<anonymous>' declared here
   46 |         boundaries = {bValues[0u], bValues[1u]};
      |                                               ^
cc1plus: all warnings being treated as errors
```

benjaminhuth:
I'll go into this later/next week

andiwand:
> @andiwand I believe these are reasonably light checks, like bounds checks or smart pointer `nullptr` checks, which a reasonable branch predictor will probably get right most of the time (at least that's the justification for this in Rust land). So @benjaminhuth is probably right that they're not terrible for performance and are likely safe for production use.

Right but ultimately this is something which should be measured before being deployed anywhere

paulgessinger:
@andiwand And despite all of this, it should be off-by-default.

Plugins/Json/src/DetrayJsonHelper.cpp
```diff
@@ -6,6 +6,13 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+// For whatever reason, this compilation unit does not compile
+// with those assertions and GCC 13. For now just disable the
+// flags here.
+#ifdef _GLIBCXX_ASSERTIONS
+#undef _GLIBCXX_ASSERTIONS
+#endif
```
paulgessinger, line 14: Should this be restricted to gcc13 only then? Does this show up in gcc14 as well?

Core/include/Acts/EventData/detail/GenerateParameters.hpp
```diff
@@ -106,15 +106,16 @@ inline std::pair<double, double> generateBoundDirection(
   // since we want to draw the direction uniform on the unit sphere, we must
   // draw from cos(theta) instead of theta. see e.g.
   // https://mathworld.wolfram.com/SpherePointPicking.html
-  double cosThetaMin = std::cos(options.thetaMin);
+  // Get cosThetaMin from thetaMax and vice versa
+  double cosThetaMin = std::cos(options.thetaMax);
```
paulgessinger, line 110: Can you elaborate?

cmake/ActsCompilerOptions.cmake
```diff
@@ -13,6 +13,11 @@ set(cxx_flags
     "-Wall -Wextra -Wpedantic -Wshadow -Wzero-as-null-pointer-constant -Wold-style-cast"
 )
 
+# Add assertions to standard libraries
+if(ACTS_FORCE_ASSERTIONS)
+    set(cxx_flags "${cxx_flags} -D_GLIBCXX_ASSERTIONS -D_LIBCPP_DEBUG")
```
paulgessinger, line 18: Is it safe to apply both flags? Should this be gated based on the stdlib version being user or is that overkill?

paulgessinger (dismissed):
couple small comments, looks good to me otherwise.

Core/include/Acts/EventData/detail/GenerateParameters.hpp
benjaminhuth: cos [0,pi] from should be monotonically decreasing, so the lower theta, the higher cos theta. therefore we must flip those for putting them into the distribution

benjaminhuth: should I put this into a comment as well?

cmake/ActsCompilerOptions.cmake
benjaminhuth: Hmm doubt there are issues if the macro for GLIBCXX is present in libc++ and vise versa. I would keep it simple until this makes problems.

Plugins/Json/src/DetrayJsonHelper.cpp
benjaminhuth: Hmm locally tested with GCC 11 and GCC 14, no issues observed there. So we could restrict this to GCC 13

Core/include/Acts/EventData/detail/GenerateParameters.hpp
paulgessinger: Ah, this is only for the distribution. Yeah maybe a comment would help.

paulgessinger (approved):

Core/include/Acts/EventData/detail/GenerateParameters.hpp
andiwand, line 1: Can we move these changes to a fix PR? I feel like we would burry them in a very unrelated change here

acts-project-service:
## :red_circle: Athena integration test results [595f2e67788ad608381ac4c75db5377865d0f2dd]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/48075872) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/9586743)!

PR #3760, opened 2024-10-18
feat(nav): Gen3 Navigation policies and factory

paulgessinger:
Terminology:

- "Navigation delegate": the function that is registered with a tracking volume. In principle, this can be anything
- "Navigation policy": the object that is registered with the tracking volume. It contains a method that is connected to the "navigation delegate". It has extra methods
- "Navigation policy factory": To delay construction of the actual policy object **until after** the volume is fully sized and has all of its internal structure registered, the blueprint tree only contains *navigation policy factories*. This is configurable. During construction, the factory is applied to volumes, and produces a policy that is registered with the volume. This is called "navigation policy factory" from a conceptual point of view.
- "MultiNavigationPolicy": chains together multiple policies in a sort of composition. You can have one policy that only deals with portals, one for sensitive and one for passive surfaces, for example.
- To make this less annoying to construct, as you would have to manage the factory, I'm adding a concrete class `NavigationPolicyFactory`. Its job is to make defining a factory that produces a "MultiNavigationPolicy" easy, like:
  ```cpp
  using namespace Acts;
  using SurfaceArrayNavigationPolicy::LayerType;
  SurfaceArrayNavigationPolicy::Config config{
    .layerType = LayerType::Cylinder,
    .bins = {10, 10}
  };
  auto factory = NavigationPolicyFactory::make()
    .add<TryAllPortalNavigationPolicy>()
    .add<SurfaceArrayNavigationPolicy>(config);
  ```
  or in python:
  ```python
  policy = (
    acts.NavigationPolicyFactory.make()
      .add(acts.TryAllPortalNavigationPolicy)
      .add(acts.TryAllSurfaceNavigationPolicy)
      .add(
        acts.SurfaceArrayNavigationPolicy,
        acts.SurfaceArrayNavigationPolicy.Config(
            layerType=acts.SurfaceArrayNavigationPolicy.LayerType.Plane,
            bins=(10, 10),
        ),
      )
  )
  ```

Part of #3502


paulgessinger:
@asalzburger I'm making some changes to the `NavigationStream` here to make it support the Gen3 portals. I'm keeping the Gen2 portals in, but I'm prefixing them with "Gen2" to clarify.

Core/include/Acts/Geometry/NavigationPolicyFactory.hpp
```diff
@@ -0,0 +1,220 @@
...
+///  .add<NavigationPolicy2>(/*no args*/)
+///  .asUniquePtr();
+/// ```
+class NavigationPolicyFactory {
+ public:
+  virtual ~NavigationPolicyFactory() = default;
+
+  // This needs to be listed here, but the return type cannot be spelled out
+  // yet.
+  static auto make();
```
andiwand, line 44: alternatively this could be a free function but I guess there is no real advantage

Core/include/Acts/Navigation/NavigationDelegate.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Acts {
+
+class NavigationStream;
+class Logger;
+
+/// Struct that serves as the argument to the navigation delegate.
+/// It is not supposed to be used as an lvalue.
+struct NavigationArguments {
+  NavigationStream& main;
```
andiwand, line 23: this object will not be copy or movable, but I guess we don't need that?

Core/include/Acts/Navigation/MultiNavigationPolicy.hpp
```diff
@@ -0,0 +1,77 @@
...
+/// `NavigationPolicyConcept`, which means that they have a conventional
+/// `updateState` method.
+///
+/// Internally, this uses a delegate chain factory to produce an unrolled
+/// delegate chain.
+///
+/// @tparam Policies The navigation policies to be combined
+template <NavigationPolicyConcept... Policies>
+  requires(sizeof...(Policies) > 0)
+class MultiNavigationPolicy final : public MultiNavigationPolicyBase {
```
andiwand, line 29: Why is this chaining necessary again?

Core/include/Acts/Navigation/SurfaceArrayNavigationPolicy.hpp
```diff
@@ -0,0 +1,77 @@
...
+  /// @param volume The *layer volume* to construct the surface array from
+  /// @param logger A logging instance
+  /// @param config The configuration for the surface array
+  explicit SurfaceArrayNavigationPolicy(const GeometryContext& gctx,
+                                        const TrackingVolume& volume,
+                                        const Logger& logger, Config config);
+
+  /// Update the navigation state from the surface array
+  /// @param args The navigation arguments
+  void updateState(const NavigationArguments& args) const;
```
andiwand, line 47: this interface look a bit suspicious to me because it is called `updateState` and we only provide a `const &`. I wonder if it would be better to separate the arguments a little like: `Stream &, const NavigationArguments &, Logger &`

andiwand: I also think this entry point is way to generic. `updateState` could mean anything: did we enter the volume? did we just finish a step? I would rather argue for concrete interface for each of these conditions.

andiwand (commented):
Some general comments and questions about the logic from my side

Core/include/Acts/Navigation/TryAllNavigationPolicies.hpp
```diff
@@ -0,0 +1,72 @@
...
+namespace Acts {
+
+class TrackingVolume;
+class GeometryContext;
+class Logger;
+
+/// Policy which adds **all** portals of the volume to the main navigation
+/// stream
+class TryAllPortalNavigationPolicy final : public INavigationPolicy {
+ public:
```
andiwand, line 22: couldn't this be merged with `TryAllSurfaceNavigationPolicy` and configurable if we insert only portals or only sensitives?

Core/include/Acts/Geometry/NavigationPolicyFactory.hpp
paulgessinger: Right, seems equivalent.

Core/include/Acts/Navigation/NavigationDelegate.hpp
paulgessinger: I'm thinking of this like a parameter pack, so that we don't have to put everything into the signature of the navigation call itself.

Core/include/Acts/Navigation/MultiNavigationPolicy.hpp
paulgessinger: To be able to combine e.g. a portal lookup with an unrelated sensor lookup. 

The alternative is writing a mega policy that can do both, or use separate user defined composition or class inheritance. 

The idea with the chain here is to provide a pattern to support this, rather than having to implement this on for each policy you might want to combine. 

In practice, I imagine using the try all portal policy in combination with a layer like policy most of the time, but I'd like to avoid the latter having to also implement portal lookup.

Core/include/Acts/Navigation/SurfaceArrayNavigationPolicy.hpp
paulgessinger: Again, `NavigationArguments` is effectively a keyword argument set, in my mind. I can split it up if that's cleaner, but I was thinking that this allows changing the interface without having to change it in many places, e.g. if we want to add an argument. `Logger` might be a special case, but otherwise I don't think there's a conceptual difference. 

The const ref is to enable passing an rvalue to the call from the outside. It doesn't prevent you from mutating mutable refs inside the structure, e.g. the navigation stream. The position and direction are const in this way however. 

I don't have an opinion on the name. 

Concerning interface and other methods: this one is really just meant as the _main_ navigation callback. I would separate this from an _on enter_ or _on exit_ call, and not try to pipe that through with argument switching. This one would provide navigation candidates, and initiate renavigation inside a volume if needed. We'd have to see if we end up only calling it after volume enter, or more often (after every step?) to enable renavigation. 

Other methods I would just add to the `INavigationPolicy` as virtual functions, and have `MultiNavigationPolicy` broadcast them to its children (which is again a reason why I think this level of composition makes sense).

Core/include/Acts/Navigation/TryAllNavigationPolicies.hpp
paulgessinger: Could to, yes

Core/include/Acts/Navigation/MultiNavigationPolicy.hpp
andiwand: I just see the use case rather narrow and the interface and code burden rather big. Instead of chaining one could have an additional navigation policy in the surface array lookup for example to delegate what to do about the portals.

My main concern is that we shape the interface of the initialize candidates call so it can be chained which results in a gen2 like navigation delegate which can practically access and mutate anything.

paulgessinger: > Instead of chaining one could have an additional navigation policy in the surface array lookup for example to delegate what to do about the portals.

But that requires each policy to foresee remaining strategies, or implement another type erasure mechanism, i.e. by using another delegate or a nested class. 

I think having this as a system, rather than leaving it up to each policy would allow us to keep the policies simpler by making them do only one thing.

Core/include/Acts/Navigation/SurfaceArrayNavigationPolicy.hpp
paulgessinger: We also lose the ability to reuse memory if we return by value. I don't think returning by value is actually an option at all. 

Also the renavigation will eventually require modifying candidates that already exist, I think. 

I agree with you main point that we shouldn't pipe through the whole state to blanket modification, but this not that, I don't think. I specifically called it `Arguments` to make clear that we pick and choose exactly what we pass in here as explicit arguments, which happen to be contained in a struct. 

Technically, right now, I agree that this is on entry. My thinking is that once we have the volume local state, where a policy can decide not to act on subsequent calls by flipping a bool in the state, we could make the navigation call this more often, with possibly no action being taken. 

For right now, `initializeCandidates` sounds good to me, but as said above we might revisit that in the future.

paulgessinger: > As long as this is the case I would not generalize the interface.

Just to clarify; that's not the intention, this function is not supposed to be the general and only entry point. Ultimately, I imagine this will be the "hot call" that updates the candidates for example at each step. At that point, "on entry" would be a separate function. 

At this time, this function serves as the "on entry" function, but it will not end up doing both at the same time.

Core/include/Acts/Navigation/MultiNavigationPolicy.hpp
paulgessinger: Also let me clarify that the chaining imo does not contradict the goal of a well defined interface I think.

The only complexity it adds imo is that indeed returning by value is tricky. 

With narrow purpose function calls, we can design the interface to be clear on how the multiple results can be accumulated. Like an output vector, that the policies push back into. 

I agree that we want to avoid just passing through complex state and allow each policy to mutate arbitrarily.

andiwand: I think my main problem with the chaining is that I only see a single use case and it looks like we overgeneralize this similar to what is done in the propagation. We allow for N arbitrary actors but in reality we only ever use one because they would only conflict with each other and it is just not useful to have more than one.

I think we only ever 1. provide surface candidates and/or 2. provide portal candidates and we have a common way of doing this individually so we want to compose that.

Core/include/Acts/Navigation/SurfaceArrayNavigationPolicy.hpp
andiwand: > We also lose the ability to reuse memory if we return by value. I don't think returning by value is actually an option at all. Also, I thought the plan was to use the navigation streams that @asalzburger added? I don't think that can work with returned candidates.

Right, I see the problem with the return by value but that does not stop us mentioning the output parameter explicitly in the function definition.

> Also the renavigation will eventually require modifying candidates that already exist, I think.

I don't think that makes sense. Re-navigation == re-initialization which makes all our lives much easier. We kill the candidates vector and create a new one.

---

I think my general problem with the gen2 navigation and this proposal is that we push navigation logic into the delegates but this is not what I would do. I would push the candidate access logic to the delegate and let the navigator do the rest. This gives us a clean separation of who is responsible of what and avoids potential conflicts and re-doing of work.

IMO the navigation policy provides a plain set of candidates which includes surfaces and portals which are intersected, filtered and sorted by the navigator. In this case we do not even have to run intersections in the navigation delegate and we can a plain vector of unintersected candidates the different policies can push to.

In the future, if we want to have re-navigation logic we can provide a way to validate the current position and direction against a provided cone/whatsoever which is checked by the navigator and it will issue a renavigation if necessary.

paulgessinger: > Re-navigation == re-initialization which makes all our lives much easier. We kill the candidates vector and create a new one.

I think that should be done inside the policy, not by the navigator. Otherwise we go back to Gen1 where the global navigator which implements the entire navigation logic for all geometric configurations. The policy is then literally only the acceleration structure, but we lose the flexibility in the process. 

> In the future, if we want to have re-navigation logic we can provide a way to validate the current position and direction against a provided cone/whatsoever which is checked by the navigator and it will issue a renavigation if necessary.

Like above, I think the cone should be internal to the policy. Only the policy knows the validity cone, where and when to create and restore it etc. 

If we only do top level renavigation from the navigator down, then we're completely back at the Gen1 model, with the caveat that we can't really implement muon system, TRT, other navigation solutions.

Core/include/Acts/Navigation/MultiNavigationPolicy.hpp
paulgessinger: Right but then what if there are many portals? Try all? What if I have a layer binning and many portals? Do I now implement four variations:

- layer lookup + smart Portal lookup
- layer lookup + try all portals
- try all sensors + smart Portal lookup
- try all sensors + try all portals

What if I add another policy as an experiment? Do I get to reimplement the other stuff as well? Sure, we can architect each strategy as a thing the combined policy call into, but that's exactly what the chaining does, in a structured and systematic way.

Core/include/Acts/Navigation/SurfaceArrayNavigationPolicy.hpp
andiwand: I don't think we should put navigation code on the geometry. We can allow the user to inject code into the `Navigator` but I would not do that through the geometry. If the user needs different navigation it might be better to just write a custom navigator.

I don't see the navigation for muons or TRT being different to anything else we do. There are surfaces and there are portals. We go through them in a sequence which is well defined and can be implemented in a single place. How to trigger re-navigation can still be specialized through the policy.

Sorry but I just don't see that level of customization being useful. This sounds like we put another Actor chain into the propagation but now it lives inside the navigator and has only partly access to the propagation. But maybe I am just missing the use case you and @asalzburger have in mind.

Core/include/Acts/Navigation/MultiNavigationPolicy.hpp
paulgessinger: > I think we only ever 1. provide surface candidates and/or 2. provide portal candidates and we have a common way of doing this individually so we want to compose that.

What I'm proposing here **is that composition** in a structured way. But you don't have to invent your own interface every time, or make lots duplicates or lists of possible configurations, you just plug it in the chain and that's it.

Core/include/Acts/Navigation/SurfaceArrayNavigationPolicy.hpp
paulgessinger: > I don't think we should put navigation code on the geometry. We can allow the user to inject code into the Navigator but I would not do that through the geometry. If the user needs different navigation it might be better to just write a custom navigator.

Really? I think this was the original reason to depart from Gen1 navigation. That we want to put the surface lookup and things like ray navigation **in the geometry**, because it depends on the geometry. 

> I don't see the navigation for muons or TRT being different to anything else we do. There are surfaces and there are portals. We go through them in a sequence which is well defined and can be implemented in a single place. 

It's not, the acceleration structure and lookup pattern fully depends on the geometry. 

> How to trigger re-navigation can still be specialized through the policy.

But the policy needs to then be able to take action. 

> Sorry but I just don't see that level of customization being useful. This sounds like we put another Actor chain into the propagation but now it lives inside the navigator and has only partly access to the propagation. But maybe I am just missing the use case you and @asalzburger have in mind.


I don't see a way we can implement TRT, MS or other non silicon tracker geometries with a god navigator. We tried in Gen1, it became impossibly complicated, we try again. 

If we need to list out all navigation modes (from layers to everything else) we just go all the way back to square one.

paulgessinger:
After discussion with @andiwand, I refactored the interface of the function contained in the `NavigationDelegate` to be called `initializeCandidates` in most cases, and I pull out the navigation stream and the logger as separate arguments to clarify the function's purpose.

I'm also wrapping the navigation stream into a separate wrapper object called `AppendOnlyNavigationStream` that only exposes methods to append to the navigation stream, thereby guaranteeing that the navigation policies will not be able to arbitrarily manipulate the candidate vector.

This is to be complemented (in a future PR) with a method like `isValid` where the navigation policies can return whether their candidates can still be considered valid, as well as dedicated function for volume entry and exit.

andiwand (approved):
I think that makes a lot of sense!

Lets get this in!

paulgessinger:
CI is green now @andiwand

andiwand (approved):

andiwand (approved):

acts-project-service:
## :red_circle: Athena integration test results [c34520f1a19ad56006495dc1a96d492d4f813e7b]
Please investigate the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8381407)!

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089291)||
|:green_circle: | [run_art_test: test_data18_13TeV_1000evt](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089290)||
|:green_circle: | [run_art_test: test_ttbarPU40_reco](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089289)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089288)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089287)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089286)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089285)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089284)||
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089283)||
|:green_circle: | [test_ActsCheckObjectCountsCachedWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089282)||
|:green_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089281)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089280)||
|:red_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089279)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089278)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089277)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089276)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089275)||
|:green_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089274)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089273)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089272)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089271)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089270)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089269)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089268)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089267)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089266)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089265)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089264)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089263)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089262)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089261)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089260)||
|:green_circle: | [run_unit_tests](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/45089259)||

PR #3765, opened 2024-10-21
feat: Add standalone digitization coordinates converter

ntadej:
This adds simple python bindings to the test framework to convert between global and local coordinates.


paulgessinger (commented):
Could you add a simple unit test either in C++ or even better in pytest to just run this code? That's just to make sure it doesn't spontaneously break.

ntadej:
> Could you add a simple unit test either in C++ or even better in pytest to just run this code? That's just to make sure it doesn't spontaneously break.

Of course, which test file would fit it the best?

ntadej:
I added the test to `test_detectors.py`.

It seems that linux_ubuntu_2204_clang failure is unrelated but let me know if this is not the case. Otherwise the MR seems to be ready.

paulgessinger (approved):

PR #3768, opened 2024-10-21
fix: Make `DataHandle` type check more robust

andiwand:
It seems that the current implementation is prone to symbol hiding. I worked around this by inserting another non-templated layer in the class hierarchy which can do the type check more reliably.

The origin of the problem could be `-fvisibility=hidden` being toggled on all of my `ActsPythonBindings` compilation commands. It might be that this is another case where this flag leaks in via CMake from another library. I also did not want to change the visibility of the symbol manually as this would require some macros to be compiler agnostic.

Some resources
- https://www.qt.io/blog/quality-assurance/one-way-dynamic_cast-across-library-boundaries-can-fail-and-how-to-fix-it
- https://developers.redhat.com/articles/2021/10/27/compiler-option-hidden-visibility-and-weak-symbol-walk-bar#

### DIFF
--- a/Examples/Framework/CMakeLists.txt
+++ b/Examples/Framework/CMakeLists.txt
@@ -13,6 +13,7 @@ add_library(
     src/Framework/WhiteBoard.cpp
     src/Framework/RandomNumbers.cpp
     src/Framework/Sequencer.cpp
+    src/Framework/DataHandle.cpp
     src/Utilities/EventDataTransforms.cpp
     src/Utilities/Paths.cpp
     src/Utilities/Options.cpp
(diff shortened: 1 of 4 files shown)


paulgessinger:
I think you didn't commit the `DataHandle.cpp`. 

This now works because the base classes are compiled in the same shared library and the template classes remain in the separate libs, but that doesn't matter anymore?

Does the hiding leak from the Python binding lib, or is it just that in the Python lib the symbols are hidden?

paulgessinger (dismissed):

paulgessinger (approved):

PR #3795, opened 2024-10-30
refactor: Combine initial and final particles in Examples

andiwand:
The split of initial and final state of particles became more and more of a burden to carry. Here I propose to combine the information into the particle object by carrying different states: initial, current, final.

I opted to have them default constructed and not as `std::optional` since we rely on construction of the state after object construction so there is no clean way of initializing initial or final.


paulgessinger:
I'm not sure if I completely understand this, but aren't these separate things? Not all initial state particles make it to the final state at the end of simulation and conversely not all final state particles correspond to any particle in the generator. 

I always found the separation into separate collections of the same type relatively clean.

paulgessinger:
Ah, I think I see what you mean now. This is separate from the split of particles from event generation and simulation, but rather is about the split between particles in the simulation when they get created vs their final state in the simulation.

Ok, that makes more sense to me to combine.

andiwand:
Right I would not touch the separation of generated and simulated particles. I think that really makes sense. But initial and final trapped me multiple times. They got out of sync, you sometimes need to wire both, you sometimes need to write both to disk. I think it is really one thing and it was only separated because of convenience. Brining it together should solve an array of issues.

andiwand:
This got a bit out of hand quickly... the current use of `Particle` makes it quite clear that we deal with a single particle in a single current state. This makes a lot of sense during simulation.

@paulgessinger and I discussed that it makes more sense to push the syncing towards the Examples. I will create another draft for this.

PR #3800, opened 2024-10-31
feat!: Estimate parameters on surface also if bottom SP is not on surface

andiwand:
The interface of
```c++
template <typename spacepoint_iterator_t>
std::optional<BoundVector> estimateTrackParamsFromSeed(
    const GeometryContext& gctx, spacepoint_iterator_t spBegin,
    spacepoint_iterator_t spEnd, const Surface& surface, const Vector3& bField,
    ActsScalar bFieldMin, const Acts::Logger& logger = getDummyLogger()) {
```
suggests that the parameters will be estimated on any surface. The doc is making this clear but we also had multiple cases when this would be useful.

Generally a spacepoint does not need to be on surface. For example for combined strip space points this is not the case.

For this reason I propose to replace this function with
```c++
template <std::ranges::range spacepoint_range_t>
Result<BoundVector> estimateTrackParamsFromSeedAtSurface(
    const GeometryContext& gctx, const MagneticFieldContext& mctx,
    const Surface& surface, spacepoint_range_t spRange,
    const std::shared_ptr<const MagneticFieldProvider>& bField,
    const std::shared_ptr<const BasePropagator>& propagator = nullptr,
    const Acts::Logger& logger = getDummyLogger()) {
```
which makes extrapolates the parameters to the `surface` and the extrapolation is explicit given the `propagator` parameter. The propagator will be default constructed if none is provided (default).

## Summary by CodeRabbit

- **New Features**
	- Introduced new methods for estimating track parameters at a specified surface with enhanced error handling.
	- Added new classes and an enumeration for improved context management and error reporting.
  
- **Bug Fixes**
	- Enhanced error handling for invalid inputs in track parameter estimation methods.

- **Documentation**
	- Revised comments and documentation for clarity on new and existing methods.

- **Refactor**
	- Streamlined magnetic field handling by removing unnecessary checks and caches in estimation algorithms.


andiwand:
cc @paulgessinger @benjaminhuth

benjaminhuth:
would be very useful! :+1:

Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp
```diff
@@ -249,34 +258,79 @@ std::optional<BoundVector> estimateTrackParamsFromSeed(
...
+          "The surface does not intersect with the origin and estimated "
+          "direction.");
+      return std::nullopt;
+    }
+
+    Direction propagatorDirection =
+        Direction::fromScalarZeroAsPositive(surfaceIntersection.pathLength());
+
+    Propagator propagator(
+        EigenStepper<>(std::make_shared<ConstantBField>(bField)),
```
benjaminhuth, line 294: So I'm not a huge fan of allocating this shared magnetic field for all seeds...

benjaminhuth: What about the idea returning the free parameters and leaving the propagation to the caller?

Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp
```diff
@@ -249,34 +258,79 @@ std::optional<BoundVector> estimateTrackParamsFromSeed(
...
+    MagneticFieldContext mctx;
+    auto propagatorOptions =
+        typename decltype(propagator)::template Options<>(gctx, mctx);
+    propagatorOptions.direction = propagatorDirection;
+
+    auto result =
+        propagator.propagate(estimatedParams, surface, propagatorOptions);
+
+    if (!result.ok()) {
+      ACTS_INFO("The propagation failed.");
```
benjaminhuth, line 305: Hmm I don't think that should be INFO level

Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp
andiwand: I think we should either do the propagation in this function or remove the function as it is quite confusing with the surface as parameter. I agree with you that the magnetic field should rather be passed in. Better would even be to push the propagator in so the user can decide on which propagator will be used.

This is along the lines what @paulgessinger suggested. But if we add a propagator argument we will always require it since we cannot default construct it because of the field.

benjaminhuth: Hmm so what I don't really like from an interface/design perspective, that this function can get silently very expensive with the propagation... 
I think the best solution would be to remove the surface from the interface, return curvilinear parameters, and provide a helper function for the propagation.

andiwand: I fear curvilinear will not really solve the problem because we use the parameters as starting parameters and then people might miss the measurement on the first surface. Curvilinear are also practically free parameters and we now have a function for that.

In that case I would maybe just drop the function and as you say provide a helper

benjaminhuth: Ah right I forgot that we have the free parameters function now. Yeah, then this is the way to go I think.

andiwand:
@paulgessinger I implemented what we discussed. cc @benjaminhuth

benjaminhuth (dismissed):
Looks good!

Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp
```diff
@@ -50,15 +62,36 @@ FreeVector estimateTrackParamsFromSeed(const Vector3& sp0, const Vector3& sp1,
...
+/// @param bField the magnetic field provider
+/// @param propagator is the propagator used in case the bottom spacepoint is not on the surface
+/// @param logger the logger to be used
+///
+/// @return bound parameters result
+Result<BoundVector> estimateTrackParamsFromSeedAtSurface(
+    const GeometryContext& gctx, const MagneticFieldContext& mctx,
+    const Surface& surface, const Vector3& sp0, const Vector3& sp1,
+    const Vector3& sp2,
+    const std::shared_ptr<const MagneticFieldProvider>& bField,
```
paulgessinger, line 87: You need this as shared to be able to construct the propagator, right?

paulgessinger (approved):
Looks good to me overall.

Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp
andiwand: Yes the stepper really wants have a shared pointer. I thought it might be good to pass in the shared pointer as a reference so we avoid a copy in case the user also provides a propagator.

We will pay the overhead on the propagator construction on every call if it is constructed by us though. I guess it is not huge but in principle avoidable by factorizing this function in estimate, transform, propagate.

Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp
```diff
@@ -50,15 +62,36 @@ FreeVector estimateTrackParamsFromSeed(const Vector3& sp0, const Vector3& sp1,
...
+    const std::shared_ptr<const BasePropagator>& propagator = nullptr,
```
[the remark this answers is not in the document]
andiwand: Good point!

pbutti (approved):

andiwand:
Looking at https://github.com/acts-project/acts/pull/3800#discussion_r1858260286 what is still not 100% clear to me is if we should really provide such a high level function hiding the potential cost of creating multiple propagator instances which copy shared pointers and needs the creation of a magnetic field cache.

@benjaminhuth pointed out that it might be good enough to have the `FreeVector` version and from there the user can transform the parameters and extrapolate if necessary.

I am sort of in the middle. I think this interface captures what the user whats to do but it does not have 0 cost under the hood if you call it many times.

Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp
```diff
@@ -99,56 +132,55 @@ FreeVector estimateTrackParamsFromSeed(spacepoint_range_t spRange,
...
-  params[eBoundLoc1] = bottomLocalPos.y();
-  params[eBoundTime] = sp0->t().value_or(0);
 
-  return Result<BoundVector>::success(params);
+  Result<BoundVector> paramsResult = estimateTrackParamsFromSeedAtSurface(
+      gctx, mctx, surface, spPositions[0], spPositions[1], spPositions[2],
+      bField, propagator, logger);
+  if (paramsResult.ok()) {
+    paramsResult.value()[eBoundTime] = spTimes[0].value_or(0);
+  }
```
andiwand, line 182: time needs to flow through the propagation

PR #3742, opened 2024-10-16
refactor: Rework particle selection in Examples

andiwand:
- remove `TruthSeedRanges` as this a duplication of `ParticleSelector `
- enhance `ParticleSelector` by also input/output final particle selections which are necessary to access the number of hits
- raise some pT cuts to the usual 1 GeV
- handle downstream changes

blocked by
- https://github.com/acts-project/acts/pull/3744

PR #3808, opened 2024-11-01
refactor: Remove `TruthSeedSelector` from Examples

andiwand:
After removing `TruthSeedRanges` with https://github.com/acts-project/acts/pull/3742 we can also safely remove `TruthSeedSelector` which does very much the same as the `ParticleSelector` by now.


paulgessinger (approved):
❤️

PR #3812, opened 2024-11-04
feat: adding G4 stepping

asalzburger:
This adds the possibility to use Geant4 as a propagation reference without the need of the SimHitConversion algorithm


Examples/Algorithms/Geant4/src/SensitiveSteppingAction.cpp
```diff
@@ -187,9 +204,59 @@ void ActsExamples::SensitiveSteppingAction::UserSteppingAction(
...
+    // Create a new step for the step logging
+    Acts::detail::Step pStep = stepFromG4Step(step);
+    pStep.geoID = geoId;
+    pStep.surface = surface != nullptr ? surface->getSharedPtr() : nullptr;
+    // Check if last step was on same surface
+    if (!pSummary.steps.empty() && pSummary.steps.back().geoID == geoId &&
+        pSummary.steps.back().surface != nullptr) {
+      auto& lastStep = pSummary.steps.back();
+      lastStep.stepSize = Acts::ConstrainedStep(currentStepLength);
+      lastStep.position = 0.5 * (pStep.position + lastStep.position);
```
andiwand, line 250: same here

Examples/Algorithms/Geant4/src/SensitiveSteppingAction.cpp
```diff
@@ -53,46 +55,59 @@ BOOST_DESCRIBE_ENUM(G4TrackStatus, fAlive, fStopButAlive, fStopAndKill,
...
+                         preStepMomentum, postStepMomentum, index);
 }
+
+Acts::detail::Step stepFromG4Step(const G4Step* step) {
+  Acts::detail::Step pStep;
+  auto [preStepPosition, preStepMomentum, postStepPosition, postStepMomentum] =
+      kinematicsOfStep(step);
+
+  pStep.navDir = Acts::Direction::Forward;
+  pStep.position = 0.5 * (preStepPosition + postStepPosition).block<3, 1>(0, 0);
```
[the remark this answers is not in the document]
asalzburger: Well, this is a bit different I fear with Geant4:
- a step in Geant4 is not a Step in Acts, the transport is done differently from the G4Stepping, hence it's more accurate to take the mid-point of a Geant4 step as the actual step equivalent ?

Examples/Algorithms/Geant4/src/SensitiveSteppingAction.cpp
asalzburger: See my comment, I don't actually know what exactly we should do if we wanted to also get the range kutta steps, I will investigate

asalzburger: I am checking with John Apostolakis from Geant4 so we understand exactly what a Geant4 step is and how we can deduce our step definition from it.

asalzburger:
Screenshot 2024-11-05 at 15 25 09

A bit of follow-up: as one can see the mid-point implementation creates steps exactly at hypothetical cylinders, independent from the Runge-Kutta-Stepping

andiwand:
I thought the `pre` and `post step` is pretty much what we have before and after the step? I imagine the next `pre step` will be the previous `post step`.

In ACTS I think we don't store the previous position (`pre step`) and always work with the current position (`post step`).

I wonder how `pre` and `post step` would look like in this picture. I imagine that the post step position would also be on surface?

After thinking about it I beliefe what G4 calls `post step` is what we store in the stepping logger as it runs after the stepper. The distance is then simply the path length of this last step which we can approximate with the distance between `pre` and `post step` in G4 or they might have some path length quantity we can use.

andiwand:
I also thought about the `detail` in `Acts::detail::Step`. I think even the step logger should not be in detail. Having them coupled in a file seems not to bad while `Acts::Step` seems a bit too general purpose. Maybe `Acts::PostStepInfo`?

Right the mapping from volume based to surface based is where we will ultimately clash with G4. I wonder if we can push these necessary conversion to the very end of the processing? Keeping the stepping info decoupled from any kind of geo/navigation info but later relate them if necessary. But I don't know yet what the use case is.

andiwand (approved):

asalzburger:
> I also thought about the `detail` in `Acts::detail::Step`. I think even the step logger should not be in detail. Having them coupled in a file seems not to bad while `Acts::Step` seems a bit too general purpose. Maybe `Acts::PostStepInfo`?
> 
> Right the mapping from volume based to surface based is where we will ultimately clash with G4. I wonder if we can push these necessary conversion to the very end of the processing? Keeping the stepping info decoupled from any kind of geo/navigation info but later relate them if necessary. But I don't know yet what the use case is.

I will make a PR for that, let me resolve the conflicts first here.

Examples/Algorithms/Geant4/src/SensitiveSteppingAction.cpp
```diff
@@ -185,9 +202,57 @@ void SensitiveSteppingAction::UserSteppingAction(const G4Step* step) {
...
+      double zVtx = track->GetVertexPosition().z();
+      double xDirVtx = track->GetVertexMomentumDirection().x();
+      double yDirVtx = track->GetVertexMomentumDirection().y();
+      double zDirVtx = track->GetVertexMomentumDirection().z();
+      double absMomentum = track->GetMomentum().mag();
+
+      PropagationSummary iSummary(Acts::CurvilinearTrackParameters(
+          Acts::Vector4(xVtx, yVtx, zVtx, 0.),
+          Acts::Vector3(xDirVtx, yDirVtx, zDirVtx), absCharge / absMomentum,
+          std::nullopt, Acts::ParticleHypothesis::pion()));
```
andiwand, line 223: do we have to worry about units here?

asalzburger: Good question, will look at it.

asalzburger: indeed this was missing.

asalzburger:
Good to go.

Examples/Algorithms/Geant4/src/SensitiveSteppingAction.cpp
andiwand: don't we have to use conversions similar to the ones above to get the right quantities for the propagation summary?

```c++
static constexpr double convertLength = Acts::UnitConstants::mm / CLHEP::mm;
static constexpr double convertEnergy = Acts::UnitConstants::GeV / CLHEP::GeV;
static constexpr double convertTime = Acts::UnitConstants::ns / CLHEP::ns;
```

andiwand (approved):
maybe I am missing something

Examples/Algorithms/Geant4/src/SensitiveSteppingAction.cpp
asalzburger: I missed the energy .. the other one is a direction  ... will fix.

asalzburger: Updated.

andiwand (approved):
Let's get this in!

asalzburger:
Now should be fine.

acts-project-service:
## :red_circle: Athena integration test results [af6a7b79b903c3a48f80eddf0f905ce98087f662]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/47508836) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/9285166)!

PR #3816, opened 2024-11-06
feat(util): Add `overloaded` helper

paulgessinger:
This is useful in combination with `std::visit`. With this, you can do something like 

```cpp
  struct A {};
  std::variant<int, double, A> var;

  var = 42;

  std::visit(overloaded{
                 [](int) { BOOST_CHECK(true); },
                 [](double) { BOOST_CHECK(false); },
                 [](A) { BOOST_CHECK(false); },
             },
             var);
```

which can be convenient.

I'm using this in https://github.com/acts-project/acts/pull/3869


paulgessinger:
I have a usage in the upcoming blueprint PR.

paulgessinger:
@AJPfleger Objections?

AJPfleger:
looks good, could you just link the blueprint-PR?

paulgessinger:
@AJPfleger That's not ready yet

paulgessinger:
@AJPfleger can you have another look now?

andiwand (approved):

PR #3785, opened 2024-10-25
fix: protect TrackSelector against eta=nan

timadye:
Protect against η=`nan` in `Acts::TrackSelector::isValidTrack()`, which would `throw std::invalid_argument{"Eta is outside the abs eta bin edges"}` exception. This can come about if somehow the track θ<0 or θ>π. We should try to prevent these out-of-range values, but having them shouldn't raise an exception.

PR #3825, opened 2024-11-07
refactor!: move source link creation, track state creation, measurement selection into a single unit outside of the CKF

goetzgaycken:
Merge retrieval of source link ranges, measurement selection and
track state creation into one unit which the CKF interacts with
via a single delegate. The delegates for measurement selection
track state creation and calibration are removed from the CKF
options/extensions.
    
The original building blocks (SourceLink accessor, measurement selector,
and track state creator, or the combined measurement selector and track
state creator) can still be used, however require to setup an additional
helper object which combines these independent components. The
algorithmic code is unchanged.

As a prerequisite the track states are converted to bound states also on
empty sensitive surfaces.

This increases the computational effort slightly for empty, sensitive
surfaces, since the computation of bound states is slightly more
demanding than the computation of curvilinear states, but for complex
events like HL-LHC events the degradation was not measurable.

--- END COMMIT MESSAGE ---

## Release Notes

- **New Features**
  - Introduced `TrackStateCreator` for enhanced track state management.
  - Added `CombinatorialKalmanFilterExtensions` for improved filter behavior customization.

- **Improvements**
  - Simplified method signatures in `CombinatorialKalmanFilterOptions`, reducing complexity.
  - Enhanced modularity in `TrackFindingAlgorithm` and related components.
  - Improved clarity and maintainability in the test framework.
  - Streamlined connections for measurement selection and calibration within the track finding process.

- **Bug Fixes**
  - Streamlined error handling and measurement selection processes.

- **Documentation**
  - Updated documentation generator to include new structures for better tracking.


goetzgaycken:
- rebase to main
- resolved merge conflicts with: fix: protect TrackSelector against eta=nan (#3785)

goetzgaycken:
changes d3a565e29b216f8ed7c6f69c89edb09c35a84a68 to b44b8fe89abae78002c327ff462b54ed7828be29:
- fix doxygen template parameter tag

goetzgaycken:
Why should there be brackets arround slAccessorDelegate [TrackFindingAlgorithm.cpp:365](https://github.com/acts-project/acts/blob/b44b8fe89abae78002c327ff462b54ed7828be29/Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp#L364-367) (macos build complains):
```
364 | TrackStateCandidateCreatorType trackStateCreator{
365 |    slAccessorDelegate,
366 |    Acts::makeDelegate<&MeasurementCalibratorAdapter::calibrate>(calibrator),
367 |    Acts::makeDelegate<&MeasurementSelector::select>(measSel)};
```

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
```diff
@@ -0,0 +1,415 @@
...
+      if (it != slBegin) {
+        // not the first TrackState, only need uncalibrated and calibrated
+        mask = PM::Calibrated;
+      }
+
+      ACTS_VERBOSE("Create temp track state with mask: " << mask);
+      // CAREFUL! This trackstate has a previous index that is not in this
+      // MultiTrajectory Visiting brackwards from this track state will
+      // fail!
+      auto ts = bufferTrajectory.makeTrackState(mask, prevTip);
```
paulgessinger, line 228: The only remaining reason for this `bufferTrajectory` is to allow the use of the previous measurement selector after calibration, which requires track states, right?

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
```diff
@@ -0,0 +1,415 @@
...
+  /// @param candidates Measurement track state candidates
+  static Result<std::pair<typename std::vector<TrackStateProxy>::iterator,
+                          typename std::vector<TrackStateProxy>::iterator>>
+  voidMeasurementSelector(typename std::vector<TrackStateProxy>& candidates,
+                          bool& /*isOutlier*/, const Logger& /*logger*/) {
+    return std::pair{candidates.begin(), candidates.end()};
+  };
+};
+
+/// @brief Base class for the ComposableTrackStateCreator which simply delegates its tasks
```
paulgessinger, line 356: This is now derived, not base anymore, isn't it?

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
```diff
@@ -0,0 +1,415 @@
...
+      Acts::Result<CkfTypes::BranchVector<TrackIndexType>>;
+  using TrackStateContainerBackend =
+      typename track_container_t::TrackStateContainerBackend;
+  using TrackProxy = typename track_container_t::TrackProxy;
+  using TrackStateProxy = typename track_container_t::TrackStateProxy;
+  using BoundState = std::tuple<BoundTrackParameters, BoundMatrix, double>;
+
+  SourceLinkAccessor sourceLinkAccessor;
+
+  const track_state_creator_derived_t& derived() const {
```
paulgessinger, line 91: This could be private maybe.

Tests/UnitTests/Core/TrackFinding/CombinatorialKalmanFilterTests.cpp
```diff
@@ -370,8 +395,11 @@ BOOST_AUTO_TEST_CASE(ZeroFieldBackward) {
 
   Fixture::TestSourceLinkAccessor slAccessor;
   slAccessor.container = &f.sourceLinks;
-  options.sourceLinkAccessor.connect<&Fixture::TestSourceLinkAccessor::range>(
-      &slAccessor);
+
+  auto trackStateCreator = makeTrackStateCreator(slAccessor, f.measSel);
```
paulgessinger, line 399: Is this used anywhere?

paulgessinger:
Couple more thoughts on this:

I think going with a single level of composition would be clearer. Currently there's the base composable class that connects to the previous `SourceLinkAccessor`, and then a second layer which accepts the other original delegates. 

A single *composable* struct that by itself has the three (?) original delegates including the `SourceLinkAccessor` would be sufficient if we drop the compatibility with the intermediate *track state creator* style delegate. 

What do we need this second layer for after this change? I think having one layer of adapter to the fully separate makes sense, but the surface area of the *track state creator* seems small enough to me to remove that altogether.

goetzgaycken:
It  is only needed for the transition period. In my opinion it would be better to first introduce the new api and then merge on the athena side the track state creator and source link accessor. The amount of code duplication when introducing different helper objects for the two cases is small, though the simplification is also not significant. Therefore I  would rather not duplicate the code, and rather remove  the unnecessary code and layers later when they are not needed anymore. But I do not have a strong opinion.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -357,15 +352,33 @@ ProcessCode TrackFindingAlgorithm::execute(const AlgorithmContext& ctx) const {
...
+  MeasurementSelector measSel{
+      Acts::MeasurementSelector(m_cfg.measurementSelectorCfg)};
+
+  using TrackStateCandidateCreatorType =
+      Acts::TrackStateCandidateCreatorImpl<IndexSourceLinkAccessor::Iterator,
+                                           TrackContainer>;
+  TrackStateCandidateCreatorType trackStateCreator{
+      slAccessorDelegate,
+      Acts::makeDelegate<&MeasurementCalibratorAdapter::calibrate>(calibrator),
+      Acts::makeDelegate<&MeasurementSelector::select>(measSel)};
```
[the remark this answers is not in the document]
goetzgaycken: You create a not properly initialized object, and the compiler would not tell You if You missed something.

andiwand:
The question for me would be, do we need a transition period? The change is breaking either way no? Touching this twice means we have to go through two breaking cycles.

goetzgaycken:
> The question for me would be, do we need a transition period? The change is breaking either way no? Touching this twice means we have to go through two breaking cycles.

It is certainly possible to merge at the same time on the athena side the source link accessor and track state creator. I don't have any strong argument to motivate why it is better to break this into smaller steps.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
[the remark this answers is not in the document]
paulgessinger: Since these are supposed to be runtime configurable, at least optionally, I don't see this as an issue. What we do everywhere else is either introduce default functions (which you do) than can error on invocation, or when pushing the object to a consumer it can check if all of the delegates are connected.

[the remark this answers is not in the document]
goetzgaycken: I don't have a strong opinion. But if it is preferred I can remove these helper methods and the starting point would be an uninitialized object and it is up to the user to remember what needs to be connected.

Though I don't see it as an improvement to move something that could be caught during compile time to the runtime. But of course You are right that delegates could be fully configurable thus it often is a runtime problem anyways.

It also seems to simplify client code. There are a few cases like 
```
Delgate delegate;
delegate.connect<&func>(obj):
Options option(delegate);
```
which just becomes 
```
Options option(makeDelegate<&func>(obj));
```
Of course the first version could just be 
```
Options option;
option.delegate.connect<&func>(obj):
```
But then the delegate has to get a default value also in cases in which there is no meaningful default value, and in which the default value is only short lived.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -357,15 +352,33 @@ ProcessCode TrackFindingAlgorithm::execute(const AlgorithmContext& ctx) const {
...
   firstOptions.targetSurface = m_cfg.reverseSearch ? pSurface.get() : nullptr;
 
   TrackFinderOptions secondOptions(ctx.geoContext, ctx.magFieldContext,
-                                   ctx.calibContext, slAccessorDelegate,
-                                   extensions, secondPropOptions);
+                                   ctx.calibContext, extensions,
+                                   secondPropOptions);
   secondOptions.targetSurface = m_cfg.reverseSearch ? nullptr : pSurface.get();
   secondOptions.skipPrePropagationUpdate = true;
+  secondOptions.extendOrBranchTrajectory =
```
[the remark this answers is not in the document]
goetzgaycken: At the moment the extensions struct would only contain a delegate for the updater and branchStopper. Both are optional, aren't they? And in theory could be identical for the entire job. In practice e.g. the branchStopper might not be stateless (e.g. information whether it is the in-side-out or out-side-in pass). The extendOrBranchTrajectory  has to know at least something about the current event, so cannot be stateless (not considering some thread local global variables ... )

It is not fully clear to me what the role of the extensions should be or become. At the moment the extensions seem to be a bit of an arbitrary subset of the options. But if it makes sense to collect all delegates in one struct and everything which is not a delegate stays in the parent struct then certainly the extendOrBranchTrajectory should be moved to the extensions.

There also was the question about the naming. Is this a good name ? Or is it misleading ? It does add either one or more new track states to the current branch (which still need some post processing), so it should do what the name suggests.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -1098,22 +818,13 @@ class CombinatorialKalmanFilter {
...
-        typename CombinatorialKalmanFilterOptions<
-            SourceLinkIterator, track_container_t>::TrackStateCandidateCreator;
+    using TrajectoryExtender = typename CombinatorialKalmanFilterOptions<
+        track_container_t>::TrajectoryExtender;
 
     /// the stateCandidator to be used
     /// @note will be set to a default trackStateCandidateCreator or the one
     //        provided via the extension
-    TrackStateCandidateCreator trackStateCandidateCreator;
+    TrajectoryExtender extendOrBranchTrajectory;
```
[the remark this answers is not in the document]
goetzgaycken: At the moment none of the members in options or extensions has the "m_" prefix. In that sense adding an "m_" prefix would make it inconsistent.

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
```diff
@@ -0,0 +1,415 @@
...
+  /// "filtered" data.
+  Result<CkfTypes::BranchVector<TrackIndexType>> extendTrajectoryOntoSurface(
+      const GeometryContext& gctx, const CalibrationContext& calibrationContext,
+      [[maybe_unused]] const Surface& surface, const BoundState& boundState,
+      TrackIndexType prevTip, TrackStateContainerBackend& bufferTrajectory,
+      std::vector<TrackStateProxy>& trackStateCandidates,
+      TrackStateContainerBackend& trajectory, const Logger& logger) const {
+    TrackStatesResult tsRes = TrackStatesResult::success({});
+    using SourceLinkRange = decltype(sourceLinkAccessor(surface));
+    std::optional<SourceLinkRange> slRange = sourceLinkAccessor(surface);
```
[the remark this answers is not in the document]
goetzgaycken: That is copy&pasted from the original code in the CKF the optional lost its purpose. So should be removed. the sourceLinkAccessor returns a range not an optional.

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
goetzgaycken: exactly.  Thanks!

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
goetzgaycken: Not sure whether that is something that needs to be hidden from the public, but it is presumably also not really necessary  to keep it public:

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
```diff
@@ -0,0 +1,415 @@
...
+///   by a source link range, calibrate the measurements and fill the
+///   the calibrated data of these track states using a dedicated calibrator
+/// - The measurement selection is delegated to a dedicated measurement
+/// selector.
+/// - Finally add branches to the given trajectory for the selected, temporary
+///   track states. The track states of these branches still lack the filtered
+//    data which is to be filled by the next stage e.g. the
+//    CombinatorialKalmanFilter.
+template <typename source_link_iterator_t, typename track_container_t>
+struct TrackStateCandidateCreatorImpl
```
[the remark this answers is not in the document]
goetzgaycken: Maybe the class should be renamed to fit the delegate name. In the original implementation in which the hierarchy was inverted this was the base class, and then this name made more sense.

Tests/UnitTests/Core/TrackFinding/CombinatorialKalmanFilterTests.cpp
goetzgaycken: Yes, that is the delegate for extendOrBranchTrajectory. Maybe I should rename it ?

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
```diff
@@ -0,0 +1,415 @@
...
+inline auto makeDelegate(const instance_t& instance) {
+  return detail::makeDelegate<memberfunc_ptr>(memberfunc_ptr, instance);
+}
+
+/// @brief convenience method to create and connect a delegate.
+/// @tparam func_ptr a pointer to a function
+template <auto func_ptr>
+inline auto makeDelegate() {
+  return detail::makeDelegate<func_ptr>(func_ptr);
+}
```
andiwand, line 67: This seems a bit generic for a header called `ComposableTrackStateCreator.hpp`. Either this should be a utility placed in the delegate header or dropped in favor of doing this by hand.

I think the deduction should also work without a function wrapper?

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
andiwand: IMO it is always a good idea to keep the public surface as small as possible. I don't think this is useful to the user?

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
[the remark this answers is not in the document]
paulgessinger: I would prefer this, yes. There's also the added benefit that explicitly writing the delegate being connected avoids any ordering confusion (although this is a compile error due to mismatching types).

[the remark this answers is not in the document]
goetzgaycken: I don't understand the argument about ordering confusion, either way You cannot make a "connected" delegate to a non existing object or function by accident, or at least I do not see how one snippet is more likely for errors than the other, except for uninitialized member delegates which are caught at runtime in one case and at compile time in the other. 

But I see You prefer for some reason unused default values and or more boiler plate code.

goetzgaycken:
changes from  2cc2ecc to dbfba96 :
- attempt to solve unresolved doxygen reference.

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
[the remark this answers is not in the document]
goetzgaycken: I did remove the optional

goetzgaycken:
changes from 2fa962c to 9493657:
- rebase to main of 2024-11-28_T1014

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
```diff
@@ -0,0 +1,415 @@
...
+template <typename source_link_iterator_t>
+using SourceLinkAccessorDelegate =
+    Delegate<std::pair<source_link_iterator_t, source_link_iterator_t>(
+        const Surface&)>;
+
+/// expected max number of track states that are expected to be added by
+/// stateCandidateCreator
+/// @note if the number of states exceeds this number dynamic memory allocation will occur.
+///       the number is chosen to yield a container size of 64 bytes.
+static constexpr std::size_t s_maxBranchesPerSurface = 10;
```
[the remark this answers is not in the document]
goetzgaycken: Resolved, the definition is now in the CombinatorialKalmanFilterExtensions.hpp which defines the delegate which needs this definition.

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
goetzgaycken: I removed these helper functions, although I think that they would have been useful

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
goetzgaycken: done

goetzgaycken:
Open discussion points:
- should the new extension member be prefixed with "m_" although non of the others are prefixed in this way ? Extend this PR and add the prefix for all members ?
- is the name "extendOrBranchTrajectory" a good name for the delegate ? Or should it be called trackStateCreator i.e the delegate this is kind of replacing (together with the sourceLinkAccessor)
- Remove TrackStateCreatorDelegate which provides a helper which allows to use  the previous track state create interface currently used by athena ? This would require to directly implement the new interface on the athena side.
- Rename TrackStateCandidateCreatorImpl ? To what ?
- Remove the buffer trajectory etc. from the new interface ? This would require that the implementation of the ComposableTrackStateCreator  (TrackStateCandidateCreatorImpl) would have to provide this i.e. would gain this as a mutable member.

goetzgaycken:
changes from 9493657 to abedde4:
- added missing include file, which only was included indirectly before.

andiwand:
> Open discussion points:

my 2 cents

> * should the new extension member be prefixed with "m_" although non of the others are prefixed in this way ? Extend this PR and add the prefix for all members ?

The lose guideline here is that for `struct`s we don't use `m_` but only for classes.

> * is the name "extendOrBranchTrajectory" a good name for the delegate ? Or should it be called trackStateCreator i.e the delegate this is kind of replacing (together with the sourceLinkAccessor)

I would be in favor of keeping `trackStateCreator`

> * Remove TrackStateCreatorDelegate which provides a helper which allows to use  the previous track state create interface currently used by athena ? This would require to directly implement the new interface on the athena side. The idea is that e.g. the minimal changes to support the new interface would be simpler: ([cern-gitlab:f8ac37ab3f](https://gitlab.cern.ch/goetz/athena/-/commit/f8ac37ab3f73ac693b7fbdf6b31ad350af4091c1), to directly support the new interface more changes are needed:  [cern-gitlab:171095fc26](https://gitlab.cern.ch/goetz/athena/-/commit/171095fc26884755d67131979580d300cfa68fd5)

To me it seems easier to make one sharp cut and break the API directly without a transition phase. I think Atlas is the only experiment using this delegate so it is really up to you IMO. What is easier for you?

> * Rename TrackStateCandidateCreatorImpl ? To what ?

I wondered if we remove the compatibility layer for the existing `TrackStateCreator` if we could/should collapse `ComposableTrackStateCreator` and `TrackStateCandidateCreatorImpl` into one class?

> * Remove the buffer trajectory etc. from the new interface ? This would require that the implementation of the ComposableTrackStateCreator  (TrackStateCandidateCreatorImpl) would have to provide this i.e. would gain this as a mutable member.

I think this would be good to remove. In case we drop the existing `TrackStateCreator` this would come for free?

goetzgaycken:
> > * should the new extension member be prefixed with "m_" although non of the others are prefixed in this way ? Extend this PR and add the prefix for all members ?
> 
> The lose guideline here is that for `struct`s we don't use `m_` but only for classes.
> 

advantage "m_"  prefix: 
- easy to see what is a member variable and what not without needing an IDE
disadvantage "m_" prefix:
- the CKF extensions is a struct so would be against the convention
- delegate calls don't look like function calls but have the "m_" prefix
I presumably would not the prefix because there wasn't one.

> > * is the name "extendOrBranchTrajectory" a good name for the delegate ? Or should it be called trackStateCreator i.e the delegate this is kind of replacing (together with the sourceLinkAccessor)
> 
> I would be in favor of keeping `trackStateCreator`

ok.

> > * Remove TrackStateCreatorDelegate which provides a helper which allows to use  the previous track state create interface currently used by athena ? This would require to directly implement the new interface on the athena side. The idea is that e.g. the minimal changes to support the new interface would be simpler: ([cern-gitlab:f8ac37ab3f](https://gitlab.cern.ch/goetz/athena/-/commit/f8ac37ab3f73ac693b7fbdf6b31ad350af4091c1), to directly support the new interface more changes are needed:  [cern-gitlab:171095fc26](https://gitlab.cern.ch/goetz/athena/-/commit/171095fc26884755d67131979580d300cfa68fd5)
> 
> To me it seems easier to make one sharp cut and break the API directly without a transition phase. I think Atlas is the only experiment using this delegate so it is really up to you IMO. What is easier for you?

Certainly reduces the amount of new code.

> > * Rename TrackStateCandidateCreatorImpl ? To what ?
> 
> I wondered if we remove the compatibility layer for the existing `TrackStateCreator` if we could/should collapse `ComposableTrackStateCreator` and `TrackStateCandidateCreatorImpl` into one class?

Indeed, if there is only a single implementation of ComposableTrackStateCreator there would not be a need to split off a derived class.

> > * Remove the buffer trajectory etc. from the new interface ? This would require that the implementation of the ComposableTrackStateCreator  (TrackStateCandidateCreatorImpl) would have to provide this i.e. would gain this as a mutable member.
> 
> I think this would be good to remove. In case we drop the existing `TrackStateCreator` this would come for free?

The Acts measurement selector needs the buffer trajectory, so it would be moved from the CKF result  and become a member of the track state creator helper class (TrackStateCandidateCreatorImpl). Interfaces become simpler. The TrackStateCandidateCreatorImpl would get a mutable state, but already has the source link accessor, so has an event state anyways.

goetzgaycken:
Changes from abedde4 to df55dfb:
- removed support for the previous track state creator interface
- replaced composable track state creator by its single implementation, and renamed file
- renamed delegate etc. to trackStateCreate
- removed buffer trajectory, and track state candidate  argument and, moved vector for candidate track state proxies to the helper class (mutable member)

Updated changes an the athena side can be found here [cern-gitlab:main_support_for_new_extendOrBranchTrajectory_delegate)](https://gitlab.cern.ch/goetz/athena/-/commits/main_support_for_new_extendOrBranchTrajectory_delegate)

goetzgaycken:
changes  df55dfb to 9261956:
- fixed spelling in comment

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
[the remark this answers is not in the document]
goetzgaycken: renamed to trackStateCreator

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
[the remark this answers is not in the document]
goetzgaycken: renamed to TrackStateCreator

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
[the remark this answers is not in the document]
goetzgaycken: did not add "m_" prefix:
- it is a struct
- contains only type definitions and delegates; without "m_" prefix invoking delegates look like function calls.

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
goetzgaycken: - Removed buffer trajectory argument (in the only use case, the buffer trajectory and the main trajectory had been the same).
- Moved the buffer for newly created candidate track state proxies to the TrackStateCreator, where it is now a mutable member.

Tests/UnitTests/Core/TrackFinding/CombinatorialKalmanFilterTests.cpp
goetzgaycken: name now agrees with the delegate name.

goetzgaycken:
Changes from 9261956 to cae4ee3:
- added missing include
- forgot to propagate name change to parts of the documentation

goetzgaycken:
Still an open question:

is the mutable member in the TrackStateCreator for the track state candidate proxies a good idea ?

 In case track finding is executed in parallel each thread (or even coroutine) would need its own version of the TrackStateCreator. It may sound worse than it is, since the trajectory cannot be extended concurrently, and the TrackStateCreator contains also the source link accessor which must have some kind of event state. 
 
Though, the latter are not unsolvable, the trajectory is passed as argument to the track finder, and the source link accessor could get the event state from thread local storage. 

But, on the other hand is good to have an argument in the trackStateCreator delegate and a buffer in CKF Result which has no obvious purpose except for the one implementation of the trackStateCreator which is currently in Acts ?  

Maybe it is better to delay the removal until there is an implementation in Acts which does not need this buffer (?)

andiwand:
I think you are right - this kind of hidden thread local state is what we usually try to avoid in ACTS so I agree with you that it might be best for now to carry it through. I was hoping to get rid of it already but forgot about this side effect.

goetzgaycken:
changes  cae4ee3 to 56d0051:
- moved the buffer for candidate track states back to the CKF results and pass it as argument to trackStateCreator.

corresponding update on the athena side is here:  [main_support_for_new_extendOrBranchTrajectory_delegate](cern-gitlab:](https://gitlab.cern.ch/goetz/athena/-/commits/main_support_for_new_extendOrBranchTrajectory_delegate)

goetzgaycken:
I don't have any further ideas. Undrafting.

Core/include/Acts/TrackFinding/ComposableTrackStateCreator.hpp
goetzgaycken: moved temporary buffer for track state candidates back to the CKF results, which is now passed again by argument to the TrackStateCreator

goetzgaycken:
A bit misleading summary by coder abitai, 
- it correctly understood that it was about restructuring of parts of the CKF,
- it correctly understood that  method signatures are simplified,
- it correctly "understood" that the molecularity improved i.e. better decoupling of CKF and source links, track state creation.
- But what is listed under "bug fixes" and "new features" is incorrect.

goetzgaycken:
changes 56d0051 to dca3186:
- adjust ActsExamplesTrackFindingExaTrkX for new track state creator interface

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -39,97 +40,12 @@
...
-using BranchVector = boost::container::small_vector<T, s_maxBranchesPerSurface>;
-
-}  // namespace CkfTypes
-
 /// Combined options for the combinatorial Kalman filter.
 ///
 /// @tparam source_link_iterator_t Type of the source link iterator
 /// @tparam track_container_t Type of the track container
-template <typename source_link_iterator_t, typename track_container_t>
+template <typename track_container_t>
```
andiwand, line 47: down to one template argument 🎉

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -309,190 +186,13 @@ class CombinatorialKalmanFilter {
...
-
   /// @brief Propagator Actor plugin for the CombinatorialKalmanFilter
   ///
-  /// @tparam source_link_accessor_t The type of source link accessor
   /// @tparam parameters_t The type of parameters used for "local" parameters.
   ///
   /// The CombinatorialKalmanFilter Actor does not rely on the measurements to
   /// be sorted along the track.
-  template <typename source_link_accessor_t, typename parameters_t>
+  template <typename parameters_t>
```
andiwand, line 195: I wonder if we could drop that? But that is something for another PR

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -1161,21 +829,17 @@ class CombinatorialKalmanFilter {
   ///
   /// @return a container of track finding result for all the initial track
   /// parameters
-  template <typename source_link_iterator_t, typename start_parameters_t,
+  template <typename start_parameters_t,
```
andiwand, line 832: These two template args also look like relics. We can clean that up later

Core/include/Acts/TrackFinding/CombinatorialKalmanFilterExtensions.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include <vector>
+
+namespace Acts {
+
+/// expected max number of track states that are expected to be added by
+/// stateCandidateCreator
+/// @note if the number of states exceeds this number dynamic memory allocation will occur.
+///       the number is chosen to yield a container size of 64 bytes.
+static constexpr std::size_t s_maxBranchesPerSurface = 10;
```
andiwand, line 22: Since we are breaking the API, do you think it makes sense to move this into a CKF related namespace?

Core/include/Acts/TrackFinding/TrackStateCreator.hpp
```diff
@@ -0,0 +1,294 @@
...
+///   selector.
+/// - Finally add branches to the given trajectory for the selected, temporary
+///   track states. The track states of these branches still lack the filtered
+///    data which is to be filled by the next stage e.g. the
+///    CombinatorialKalmanFilter.
+/// All track states, the temporary track states and track states for selected
+/// measurements, are created in the given trajectory. The resulting container
+/// may become big. Thus, it is advisable to copy selected tracks and their
+/// track states to a separate container after each track finding step.
+
```
andiwand, line 34:
```suggestion
///
```
nitpick

andiwand (dismissed):
This is great!

I left a few minor comments.

Since this is breaking we can only merge this in ~2 weeks. But I don't expect it to conflict with other CKF developments.

goetzgaycken:
changes 7d5548c to 3639a5a:
- ... and the 3rd attempt to fix the delegate connections for the ExaTrk proto track algorithm

goetzgaycken:
changes 3639a5a to 335974e:
- fix setup of track finder options of ExaTrk proto track finding

Core/include/Acts/TrackFinding/CombinatorialKalmanFilterExtensions.hpp
goetzgaycken: We can move this to the CfkTypes namespace or even move this and the CfkTypes to the extensions.

Tests/UnitTests/Core/TrackFinding/CombinatorialKalmanFilterTests.cpp
```diff
@@ -289,14 +284,40 @@ struct Fixture {
...
+                              TrackContainer>;
+  TrackStateCreatorType trackStateCreator;
+  trackStateCreator.sourceLinkAccessor
+      .template connect<&source_link_accessor_t::range>(&slAccessor);
+  trackStateCreator.calibrator.template connect<
+      &testSourceLinkCalibrator<TrackStateContainerBackend>>();
+  trackStateCreator.measurementSelector.template connect<
+      &Acts::MeasurementSelector::select<TrackStateContainerBackend>>(&measSel);
+  return trackStateCreator;
+}
```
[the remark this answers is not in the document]
goetzgaycken: Where does master coder abitai get this from ?

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -318,27 +319,34 @@ ProcessCode TrackFindingAlgorithm::execute(const AlgorithmContext& ctx) const {
...
 
   using Extensions = Acts::CombinatorialKalmanFilterExtensions<TrackContainer>;
 
   BranchStopper branchStopper(m_cfg);
+  MeasurementSelector measSel{
+      Acts::MeasurementSelector(m_cfg.measurementSelectorCfg)};
+
+  IndexSourceLinkAccessor slAccessor;
+  slAccessor.container = &measurements.orderedIndices();
+
```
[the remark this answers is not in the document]
goetzgaycken: Dear mater coder abitai, I think an empty measurement range is acceptable. Anyway has nothing to do with this PR.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -318,27 +319,34 @@
...
   extensions.branchStopper.connect<&BranchStopper::operator()>(&branchStopper);
-
-  IndexSourceLinkAccessor slAccessor;
-  slAccessor.container = &measurements.orderedIndices();
-  Acts::SourceLinkAccessorDelegate<IndexSourceLinkAccessor::Iterator>
-      slAccessorDelegate;
-  slAccessorDelegate.connect<&IndexSourceLinkAccessor::range>(&slAccessor);
+  extensions.createTrackStates
+      .template connect<&TrackStateCreatorType ::createTrackStates>(
+          &trackStateCreator);
```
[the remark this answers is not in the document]
goetzgaycken: Dear master coder abitai, I think connect returns void .

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -318,27 +319,34 @@
...
+  using TrackStateCreatorType =
+      Acts::TrackStateCreator<IndexSourceLinkAccessor::Iterator,
+                              TrackContainer>;
+  TrackStateCreatorType trackStateCreator;
+  trackStateCreator.sourceLinkAccessor
+      .template connect<&IndexSourceLinkAccessor::range>(&slAccessor);
+  trackStateCreator.calibrator
+      .template connect<&MeasurementCalibratorAdapter::calibrate>(&calibrator);
+  trackStateCreator.measurementSelector
+      .template connect<&MeasurementSelector::select>(&measSel);
```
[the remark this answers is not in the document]
goetzgaycken: still returns void

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
goetzgaycken: I actually wonder whether this can be different from BoundTrackParameters

andiwand: Right I think this can only be `BoundTrackParameters` since it should not be coupled to the start parameters. Even for the start parameters I would argue that only `BoundTrackParameters` make sense for now

Core/include/Acts/TrackFinding/CombinatorialKalmanFilterExtensions.hpp
goetzgaycken: Just experimenting with this. The  s_maxBranchesPerSurface  can certainly be moved into a namespace. I said one can move everything into the CKF extensions. The disadvantage would be that the full type name becomes a template of the track container and is very long. Thus, aliases would be used where ever these types appear. These aliases makes it often tedious to search and/or understand the actual definition of a type.  Anyway I would not have a strong opinion.

Tests/UnitTests/Core/TrackFinding/CombinatorialKalmanFilterTests.cpp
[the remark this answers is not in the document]
goetzgaycken: Dear master coder abitai, unfortunately connect returns void. So, Your error handing is not possible.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilterExtensions.hpp
andiwand: Yes I agree we should not put it behind one of the templated classes. Potentially `CkfTypes` makes sense but I don't have a strong preference. I just thought `s_maxBranchesPerSurface` is a bit generic for the `Acts` namespace.

goetzgaycken:
Changes from  335974e to 175a588:
- rebase to main of 2024-12-04_T1314
- propagated Juan Miguel Carceller's spelling fix to the TrackStateCreator 
- moved definition of s_maxBranchesPerSurface into namespace CkfTypes

goetzgaycken:
Changes from 175a588 to 5dd2547:
- added the "lost the /// instead of an empty line" fix

Core/include/Acts/TrackFinding/CombinatorialKalmanFilterExtensions.hpp
goetzgaycken: moved s_maxBranchesPerSurface into CkfTypes

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -768,17 +454,17 @@ class CombinatorialKalmanFilter {
...
+      if (isSensitive) {
+        // extend trajectory with measurements associated to the current surface
+        // which may create extra trajectory branches if more than one
+        // measurement is selected.
+        tsRes = m_extensions.createTrackStates(
             state.geoContext, *calibrationContextPtr, *surface, boundState,
-            slBegin, slEnd, prevTip, *result.trackStates,
-            result.trackStateCandidates, *result.trackStates, logger());
+            prevTip, result.trackStateCandidates, *result.trackStates,
+            logger());
```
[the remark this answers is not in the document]
goetzgaycken: Dear master coder abitai,
The  proposed debug message is useless since the cache is expected to be empty before calling createTrackStates. But the proposal for the error message makes indeed sense.

goetzgaycken:
Changes from 5dd2547 to 65948cf:
- improved error message as suggested by coderabbitai

Changes from [65948cf](https://github.com/acts-project/acts/commit/65948cf3cbd7c960fe4debdb6d851da168475f58) to [dbd9d5e](https://github.com/acts-project/acts/commit/dbd9d5ea71593a0e0981affa2d8712c30bc27c97):
- rebase to main of 2024-12-10_T2146

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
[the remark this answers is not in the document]
goetzgaycken: Applied improvement to error message.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
goetzgaycken: Something for a different PR

goetzgaycken:
changes:
[dbd9d5e](https://github.com/acts-project/acts/commit/dbd9d5ea71593a0e0981affa2d8712c30bc27c97) to [b8e7d2b](https://github.com/acts-project/acts/commit/b8e7d2ba8937ce533c989f1144c0e709b8ae7aeb):
- white space fixes

goetzgaycken:
The corresponding necessary changes for athena can be found in [gitlab-cern:main_support_for_new_extendOrBranchTrajectory_delegate](https://gitlab.cern.ch/goetz/athena/-/commits/main_support_for_new_extendOrBranchTrajectory_delegate)

paulgessinger:
Let's make sure to get this in for v39.

Core/include/Acts/TrackFinding/TrackStateCreator.hpp
```diff
@@ -0,0 +1,294 @@
...
+    for (auto it = slBegin; it != slEnd; ++it) {
+      // get the source link
+      const auto sourceLink = *it;
+
+      // prepare the track state
+      PM mask = PM::Predicted | PM::Jacobian | PM::Calibrated;
+      if (it != slBegin) {
+        // not the first TrackState, only need uncalibrated and calibrated
+        mask = PM::Calibrated;
+      }
```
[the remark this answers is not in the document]
goetzgaycken: it is intended to overwrite the mask.

andiwand (approved):

PR #3836, opened 2024-11-09
refactor: Reworked `HitSelector` in Examples

andiwand:
The motivations is that one can select hits to constrain the reconstruction to a subset of them. This can be especially useful for studies which look at different pile-up. Pile-up vertices have consecutive primary vertex IDs so one can simulate for example pile-up 200 and then reconstruct only a subset of them.

Bocked by
- https://github.com/acts-project/acts/pull/3837

## Summary by CodeRabbit

- **New Features**
	- Introduced the `HitSelector` class for enhanced hit selection based on various criteria such as position, time, energy loss, and primary vertex ID.
	- Added new parameters to the `HitSelector` algorithm, allowing for more comprehensive filtering of hits.

- **Bug Fixes**
	- Improved validation logic in the `ParticleSelector` class to include primary vertex ID checks.

- **Documentation**
	- Updated header files to reflect the new structure and functionality of the `HitSelector`.
  
- **Chores**
	- Removed outdated `HitSelector` files from the utilities library to streamline the codebase.

PR #3838, opened 2024-11-10
refactor: Relate measurements to truth in Examples

andiwand:
Relates our Examples measurements to truth hits and particles which makes the additional maps on the side unnecessary. This simplifies the modification of measurements collections as the truth relations do not need to be updated as well.

One potential problem is that we combine reconstruction and truth information in a single place and people might mix it up for reconstruction workflows without truth. However this is our Examples framework and not an experiment framework so the simplification outweighs the potential risk of misuse for experiments IMO.

Useful for https://github.com/acts-project/acts/pull/3836


andiwand:
After discussions with @benjaminhuth we agreed that we would continue relying on maps for the mapping for now. I would like to pull out some changes from this PR which are not related to the feature change. Closing

PR #3849, opened 2024-11-12
refactor: Stable RNG for smeared digitization in Examples

andiwand:
Our smeared digitization will create very different random numbers if the simulations changes. This can be unexpected for pile-up studies where you would think your measurements from the hard scatter stay the same and you just get more pile-up hits.

I am trying to improve the situation by creating stable random number streams for each particle on a sensitive. To avoid correlations for multiple hits of a single particle on a single surface (should only happen for loopers) I introduced an index which is also part of the seed for the rng.


andiwand:
Tested this locally and works fine

andiwand:
cc @benjaminhuth @paulgessinger @asalzburger 

I don't understand how this changes physmon so much? Did we have some weird correlations before? Or do we now?

benjaminhuth:
One thing I'm not sure is if the the properties of any / this particular RNG are similar in situations like
```
rng = rng_from_seed(seed)
seq1 = [ rng() for  i in range(N) ]
# vs
seq2 = [ rng_from_seed(i)() for i in range(N) ]
```
Because effectively we have a measurement unique RNG, that is only called once or twice, right?

Another thing one could try to improve the quality of the seed is maybe using `std::seed_seq`, instead of just summing up the ids.

Examples/Algorithms/Digitization/src/DigitizationAlgorithm.cpp
```diff
@@ -206,6 +207,19 @@ ActsExamples::ProcessCode ActsExamples::DigitizationAlgorithm::execute(
...
+            const auto hitIndex =
+                particleOnModuleHitCount[simHit.particleId()]++;
+
+            if (hitIndex > 0) {
+              ACTS_DEBUG("more than one hit for particle "
+                         << simHit.particleId() << " on module " << moduleGeoId
+                         << " - still forwarding it to the digitizer");
+            }
+
+            auto hitSeed = eventSeed + moduleGeoId.value() +
```
paulgessinger, line 219: Are we concerned about overflow with `value` possibly having very large values? Also, do we need the `auto` or can we be more specific here?

andiwand:
looks like `std::seed_seq` only does 32 bits

Examples/Algorithms/Digitization/src/DigitizationAlgorithm.cpp
andiwand: Our seeds are `std::uint64_t` so they should gracefully wrap around. I didn't want to spell out the specific type but maybe I can typedef it

PR #3856, opened 2024-11-13
fix: Particle selector works on measurements, not simhits

benjaminhuth:
The particle selection did work on the simhit count, but usually we rather desire to select on the actual number of measurements in the detector, which can be different.

--- END COMMIT MESSAGE ---

### DIFF
--- a/Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/ParticleSelector.cpp
+++ b/Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/ParticleSelector.cpp
@@ -28,6 +28,12 @@ ActsExamples::ParticleSelector::ParticleSelector(const Config& config,
   }
 
   m_inputParticles.initialize(m_cfg.inputParticles);
+
+  Config defaults;
+  if (m_cfg.measurementsMin > defaults.measurementsMin ||
+      m_cfg.measurementsMax < defaults.measurementsMax) {
+    m_inputMeasPartMap.initialize(m_cfg.inputMeasurementParticlesMap);
+  }
   m_outputParticles.initialize(m_cfg.outputParticles);
 
   ACTS_DEBUG("selection particle rho [" << m_cfg.rhoMin << "," << m_cfg.rhoMax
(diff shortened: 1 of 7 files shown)


Examples/Io/Json/src/TestSurfaceJsonRoundtrip.cpp
paulgessinger, line 1: This seems unrelated?

paulgessinger:
I'm not sure I understand when this makes a difference. Is if if we implement some kind of module inefficiency? When is the number of measurements per particle different from the number of truth hits?

benjaminhuth:
Yeah so I think simulated hit and digitized measurement are different concepts, right? E.g., geometric clustering could filter out clusters with a threshold, we can drop hits based on energy deposit, there could be split clusters...

Examples/Io/Json/src/TestSurfaceJsonRoundtrip.cpp
benjaminhuth: ah that went in accidentally

paulgessinger:
@benjaminhuth Right, that's what I was thinking of with module inefficiency. I guess a minimum charge threshold could have an affect like you say.

Split clusters don't currently exist I don't think, and in that case we go from multiple hits to a single merged cluster that is then split again later using reco information only. In case of merged clusters, the cluster would be associated with both source particles. 

I think with how you implement this here, that should be ok, as the measurement is mapped back to each simhits particles, so would survive the map reversal you add here.

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/ParticleSelector.cpp
```diff
@@ -56,6 +62,21 @@ ActsExamples::ProcessCode ActsExamples::ParticleSelector::execute(
...
+    const auto& measPartMap = m_inputMeasPartMap(ctx);
+    using InvPair = std::pair<ActsFatras::Barcode, Index>;
+    std::vector<InvPair> v(measPartMap.size());
+    std::ranges::transform(measPartMap, v.begin(), [](const auto& p) {
+      return InvPair{p.second, p.first};
+    });
+    std::ranges::sort(v, std::less{}, &InvPair::first);
+    partMeasMap.emplace(boost::container::ordered_range_t{}, v.begin(),
+                        v.end());
+  }
```
andiwand, line 78: I thought this inverted map is somewhere generated already and put on the whiteboard?

Also, isn't there a utility to do this kind of transformation? I can also try to look it up

andiwand:
Thanks for following up on this! I overlooked that hits are not enough for this

PR #3859, opened 2024-11-14
feat: Random number channel for Examples

andiwand:
Utility to create independent random number channels which allows to have stable streams for different purposes. For example smearing bound parameters will create different results depending on which parameters are smeared and in which order. Having different channels avoids this.


andiwand:
Superseded by https://github.com/acts-project/acts/pull/3861

PR #3862, opened 2024-11-15
refactor!: Use `double` for material

andiwand:
There is no good reason to use `float` for material properties and calculations, we just end up doing a bunch of `float` `double` conversions. The only good reason for `float` is to keep the storage low which is still the case with this PR.

## Release Notes

- **New Features**
  - Added new constant `kAvogadro` for Avogadro's number to enhance calculations involving moles.
  - Introduced static methods for predefined material parameters for vacuum conditions.
  - Added a new test case for validating material mapping functionality in both rz and xyz coordinate systems.

- **Improvements**
  - Updated various methods and parameters across multiple classes to use `double` instead of `float`, enhancing precision in calculations related to material properties and interactions.
  - Improved documentation comments for clarity on physical constants.

- **Bug Fixes**
  - Adjusted tests to ensure that comparisons are based on absolute differences for material properties, enhancing test accuracy.


andiwand:
image
image

AJPfleger:
I started to work on the same topic in
- https://github.com/acts-project/acts/pull/3609

In a first attempt, I changed all instances as well but it messed up the tests. To understand, where the problem comes from, I believe it is necessary, to do it in smaller steps. (What I tried above, but still too course.)

Of course the problem lies in some rounding, since a lot of calculations are done in `double` but then stored as `float`. We need to understand which steps cause the changes in the output, since it is very likely, that we are doing some comparisons between `float`s which lead to different behaviour when looking at more digits. Therefore, we probably need to reintroduce some comparison operators or understand if we actually need to do those comparisons.

andiwand:
I already adopted the tests here. I don't think we should change any `==` operators. Changing them breaks the interface. I would rather remove them as they are not very useful anyways.

AJPfleger:
Most of the unit tests are not an issue - that we need to compare to a double instead of a float is straight forward. But the physmon changes are complicated. Have you had time to analyse them?

Just a few examples:
Here we have shifts:
image
image
image

Here a lot of stuff going on:
image

We don't need to change the `==`-operator, we could also change the place where it is used incorrectly. But I don't know if there is an actual case, where we use it, since I didn't get that far in my analysis yet :( Before changing anything, we need to understand, why it isfailing.

andiwand:
I don't think it is worth looking into these physmon changes in detail as there is no structure to it. They all look consistent to low stat fluctuations. We change how material interactions are numerically represented so such changes in physmon are expected.

I can check how easy it is to remove the `==` operators.

paulgessinger:
@AJPfleger @andiwand are we happy with this? If so, the references need to be updated, and then we can get this in!

AJPfleger:
I am not happy with this, since we don't understand what causes the changes. The material was (supposedly) tested with `float`. Just changing to `double` adds more uncertainty to the model.

> They all look consistent to low stat fluctuations

Then we need to repeat it with larger statistics.

> physmon changes in detail as there is no structure to it

There is a structure in the changes. We have shifts in numerous plots. If they are not significant due to the statistics, we need larger statistics.

andiwand:
Well in this case we can close the PR then

PR #3877, opened 2024-11-20
feat: Helpers to calculate residuals, chi2 and unbiased states

andiwand:
This came up a couple of times and it really should be in core so I added helpers to
- calculate residuals and their covariance for predicted, filtered and smoothed states
- calculate chi2 for these
- calculate unbiased parameters and covariance

For now I added them to `TrackHelpers.hpp` but potentially this could also live in the `TrackStateProxy`.

## Summary by CodeRabbit

- **New Features**
	- Introduced new helper functions for track state calculations, including methods for computing predicted, filtered, and smoothed residuals and chi-squared values.
	- Added functionality to calculate unbiased track parameters and their covariance.

- **Bug Fixes**
	- Simplified the logic for calculating unbiased track parameters, improving maintainability and reducing complexity.

- **Documentation**
	- Updated method signatures and descriptions to reflect changes in functionality.


andiwand:
cc @pbutti @paulgessinger

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -493,6 +497,136 @@ void trimTrack(track_proxy_t track, bool trimHoles, bool trimOutliers,
...
+    track_state_proxy_t trackState) {
+  // calculate the unbiased track parameters (i.e. fitted track parameters with
+  // this measurement removed) using Eq.(12a)-Eq.(12c) of NIMA 262, 444 (1987)
+
+  return visit_measurement(trackState.calibratedSize(), [&](auto N) {
+    constexpr int measdim = decltype(N)::value;
+
+    auto H =
+        trackState.projector().template topLeftCorner<measdim, eBoundSize>();
+    auto s = trackState.smoothed();
```
pbutti, line 619: Maybe add check that trackState has smoothed and has projector otherwise throw exception to avoid users calling this on wrong state type? Maybe just check that trackState is measurement type?

andiwand: @paulgessinger do you have an opinion on this? should we already throw in `smoothed()` or `calibratedSize`? or should this always be handled downstream?

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -493,6 +497,136 @@ void trimTrack(track_proxy_t track, bool trimHoles, bool trimOutliers,
...
+std::pair<ActsVector<nMeasurementDim>, ActsSquareMatrix<nMeasurementDim>>
+calculateFilteredResidual(track_state_proxy_t trackState) {
+  using MeasurementVector = ActsVector<nMeasurementDim>;
+  using MeasurementMatrix = ActsSquareMatrix<nMeasurementDim>;
+
+  auto subspaceHelper =
+      trackState.template fixedBoundSubspaceHelper<nMeasurementDim>();
+
+  MeasurementVector measurement = trackState.calibrated();
+  MeasurementMatrix measurementCovariance = trackState.calibratedCovariance();
```
paulgessinger, line 535: This takes copies of the vector and matrix. Do we need these?

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -493,6 +497,136 @@ void trimTrack(track_proxy_t track, bool trimHoles, bool trimOutliers,
...
+std::pair<ActsVector<nMeasurementDim>, ActsSquareMatrix<nMeasurementDim>>
+calculateSmoothedResidual(track_state_proxy_t trackState) {
+  using MeasurementVector = ActsVector<nMeasurementDim>;
+  using MeasurementMatrix = ActsSquareMatrix<nMeasurementDim>;
+
+  auto subspaceHelper =
+      trackState.template fixedBoundSubspaceHelper<nMeasurementDim>();
+
+  MeasurementVector measurement = trackState.calibrated();
+  MeasurementMatrix measurementCovariance = trackState.calibratedCovariance();
```
paulgessinger, line 559: Also copies here.

Core/include/Acts/Utilities/TrackHelpers.hpp
paulgessinger: `smoothed` and `calibratedSize` only check as assertions, not explicitly. 

There exists `hasSmoothed`, which could be checked, but that only checks if the allocation is present or not. Equivalently, I believe you can check if `calibratedSize()` is `kInvalid` to check if a measurement is allocated.

Probably the extra check isn't to expensive to just do unconditionally but I never tried this out.

paulgessinger (commented):
Good idea. The new public functions would need at least doxygen comments.

pbutti (dismissed):

paulgessinger:
Sonar has a couple of valid points on the usage of `auto` parameters in lambdas where the parameter is then actually used.

andiwand:
> Sonar has a couple of valid points on the usage of `auto` parameters in lambdas where the parameter is then actually used.

@paulgessinger I think this is the code I copied from other places to do the `visit_measurement` unroll magic. Do you think we should change this coherently in all places to use `template` instead of `auto`?

paulgessinger:
@andiwand This only complains about cases where the template parameter is subsequently accessed. Indeed, this only became possible in C++20 with explicit template parameters to lambdas.

Ultimately, we should migrate to it, but I personally think updating this incrementally is fine.

Core/include/Acts/Utilities/TrackHelpers.hpp
```diff
@@ -493,6 +497,216 @@ void trimTrack(track_proxy_t track, bool trimHoles, bool trimOutliers,
...
+    throw std::invalid_argument("track state has no calibrated parameters");
+  }
+
+  return visit_measurement(
+      trackState.calibratedSize(),
+      [&]<std::size_t measdim>(std::integral_constant<std::size_t, measdim>) {
+        auto [residual, residualCovariance] =
+            calculatePredictedResidual<measdim>(trackState);
+
+        return residual.transpose() * residualCovariance.inverse() * residual;
```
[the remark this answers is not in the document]
andiwand: I think this is something we could generally do better in ACTS. can you open an issue about this @coderabbitai?

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [93d878140c530a157ba61b2a3a0f61caaf58da3a]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/46550127) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8597661)!

PR #3886, opened 2024-11-21
feat: Use RootTracksummaryWriter without truth information

benjaminhuth:
Since some people start to do tests within the examples framework with real data, we should extend our algorithms/writers to not require these information. This PR does the relevant changes for the `RootTracksummaryWriter`. 

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **New Features**
	- Enhanced flexibility in handling optional input parameters for the `RootTrackSummaryWriter`, allowing the program to operate without mandatory input collections.
	- Improved robustness in data handling processes, ensuring the program continues executing even with incomplete data.

- **Bug Fixes**
	- Updated error handling to prevent exceptions when input collections are not populated.

- **Tests**
	- Introduced new tests to validate the functionality of the `RootTrackSummaryWriter` and integrated it into existing testing frameworks for comprehensive coverage.


Examples/Io/Root/src/RootTrackSummaryWriter.cpp
```diff
@@ -244,6 +249,9 @@ ProcessCode RootTrackSummaryWriter::writeT(const AlgorithmContext& ctx,
     m_nSharedHits.push_back(track.nSharedHits());
     m_chi2Sum.push_back(track.chi2());
     m_NDF.push_back(track.nDoF());
+
+    // IDs are stored as double (as the vector of vector of int is not known
+    // to ROOT)
```
andiwand, line 254: I think this is not the case anymore?

Examples/Io/Root/src/RootTrackSummaryWriter.cpp
```diff
@@ -221,8 +220,14 @@ ProcessCode RootTrackSummaryWriter::finalize() {
...
-  const auto& particles = m_inputParticles(ctx);
-  const auto& trackParticleMatching = m_inputTrackParticleMatching(ctx);
+  std::optional<SimParticleContainer> particles;
+  std::optional<TrackParticleMatching> trackParticleMatching;
+
+  if (m_inputParticles.isInitialized() &&
+      m_inputTrackParticleMatching.isInitialized()) {
+    particles = m_inputParticles(ctx);
+    trackParticleMatching = m_inputTrackParticleMatching(ctx);
+  }
```
andiwand, line 230: note that this will do a copy

Examples/Io/Root/src/RootTrackSummaryWriter.cpp
```diff
@@ -131,18 +107,41 @@ RootTrackSummaryWriter::RootTrackSummaryWriter(
...
-  m_outputTree->Branch("res_eT_fit", &m_res_eT_fit);
-  m_outputTree->Branch("pull_eLOC0_fit", &m_pull_eLOC0_fit);
-  m_outputTree->Branch("pull_eLOC1_fit", &m_pull_eLOC1_fit);
-  m_outputTree->Branch("pull_ePHI_fit", &m_pull_ePHI_fit);
-  m_outputTree->Branch("pull_eTHETA_fit", &m_pull_eTHETA_fit);
-  m_outputTree->Branch("pull_eQOP_fit", &m_pull_eQOP_fit);
-  m_outputTree->Branch("pull_eT_fit", &m_pull_eT_fit);
+
+  if (m_inputParticles.isInitialized() &&
+      m_inputTrackParticleMatching.isInitialized()) {
```
andiwand, line 112: I am always a bit unsure if we should make columns optional. This can make reading a file complicated but I am also not sure if reading really is a use case

benjaminhuth: what do you mean by "reading" here?

Examples/Io/Root/src/RootTrackSummaryWriter.cpp
benjaminhuth: Hmm should we use raw pointer instead?

Examples/Io/Root/src/RootTrackSummaryWriter.cpp
paulgessinger: I think "when is what actually being read, needed, non-empty".

paulgessinger: I think figuring out which parts are set in which scenarios when reading the source code. The more optional parts you have, the trickier it gets.

andiwand:
Overall I wonder if it might be easier to just have empty truth information in case these entries in the whiteboard are not present. I think not having any truth should act similarly to not matching to truth?

I would also like to avoid special casing in all directions so we can make data workflows work as this seems really not the goal of our design.

Examples/Io/Root/src/RootTrackSummaryWriter.cpp
```diff
@@ -221,8 +217,14 @@ ProcessCode RootTrackSummaryWriter::finalize() {
...
-  const auto& particles = m_inputParticles(ctx);
-  const auto& trackParticleMatching = m_inputTrackParticleMatching(ctx);
+  const SimParticleContainer* particles = nullptr;
+  const TrackParticleMatching* trackParticleMatching = nullptr;
+
+  if (m_inputParticles.isInitialized() &&
+      m_inputTrackParticleMatching.isInitialized()) {
+    particles = &m_inputParticles(ctx);
+    trackParticleMatching = &m_inputTrackParticleMatching(ctx);
+  }
```
andiwand, line 227: alternatively you could use references and if `m_inputParticles` is not preset for example let it point to a static empty one.

something like

```suggestion
  static SimParticleContainer emptyParticles;

  const SimParticleContainer particles = m_inputParticles.isInitialized() ? m_inputParticles(ctx) : emptyParticles;
```

Examples/Io/Root/src/RootTrackSummaryWriter.cpp
andiwand, line 1: shouldn't most of the diff go away with this approach?

maybe I am missing something

Examples/Io/Root/src/RootTrackSummaryWriter.cpp
paulgessinger: @andiwand your diff would now do a copy, you'd need to add the `&`.

Examples/Io/Root/src/RootTrackSummaryWriter.cpp
```diff
@@ -99,6 +94,20 @@ RootTrackSummaryWriter::RootTrackSummaryWriter(
...
+  m_outputTree->Branch("ePHI_fit", &m_ePHI_fit);
+  m_outputTree->Branch("eTHETA_fit", &m_eTHETA_fit);
+  m_outputTree->Branch("eQOP_fit", &m_eQOP_fit);
+  m_outputTree->Branch("eT_fit", &m_eT_fit);
+  m_outputTree->Branch("err_eLOC0_fit", &m_err_eLOC0_fit);
+  m_outputTree->Branch("err_eLOC1_fit", &m_err_eLOC1_fit);
+  m_outputTree->Branch("err_ePHI_fit", &m_err_ePHI_fit);
+  m_outputTree->Branch("err_eTHETA_fit", &m_err_eTHETA_fit);
+  m_outputTree->Branch("err_eQOP_fit", &m_err_eQOP_fit);
+  m_outputTree->Branch("err_eT_fit", &m_err_eT_fit);
```
paulgessinger, line 109: This probably doesn't need to move now?

Examples/Io/Root/src/RootTrackSummaryWriter.cpp
```diff
@@ -276,120 +277,6 @@ ProcessCode RootTrackSummaryWriter::writeT(const AlgorithmContext& ctx,
...
-    m_t_pz.push_back(t_pz);
-    m_t_theta.push_back(t_theta);
-    m_t_phi.push_back(t_phi);
-    m_t_eta.push_back(t_eta);
-    m_t_p.push_back(t_p);
-    m_t_pT.push_back(t_pT);
-    m_t_d0.push_back(t_d0);
-    m_t_z0.push_back(t_z0);
-
     // Initialize the fitted track parameters info
```
paulgessinger, line 280: This is also only moved now?

Examples/Io/Root/src/RootTrackSummaryWriter.cpp
benjaminhuth: I reordered some code, I guess thats what the diff is about.

Examples/Io/Root/src/RootTrackSummaryWriter.cpp
andiwand: Thanks I corrected the diff

andiwand (approved):
Looks good thank you! 🚀

acts-project-service:
## :red_circle: Athena integration test results [832b3271c473963d629ffd426ff4d96af2bd425f]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/46546929) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/8597170)!

PR #3891, opened 2024-11-22
add misalignment decoration for DD4hep detector

Zequn-SUN:
- Add misalignment decoration for odd. In this version, a file `odd-misalignment-matrix.json` is required. The json file includes the identifiers of the misaligned DD4hepDetectorElements, as well as six parameters: the translation parameters in the x, y, and z directions (unit: cm), and the rotation angles along the x, y, and z axes (unit: degree). In the decorator, `DD4hepGeometryContext` is used to store the misaligned transform matrix.
- A follow-up PR will be made including the following development:
- Create an interface to control whether the detector is misaligned. 
- Integrate the functionality to generate misalignment parameters?

### DIFF
--- a/Examples/Detectors/DD4hepDetector/CMakeLists.txt
+++ b/Examples/Detectors/DD4hepDetector/CMakeLists.txt
@@ -3,6 +3,7 @@ add_library(
     SHARED
     src/DD4hepDetector.cpp
     src/DD4hepGeometryService.cpp
+    src/DD4hepAlignmentDecorator.cpp
 )
 
 target_include_directories(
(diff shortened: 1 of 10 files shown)


andiwand:
where do we stand with this one @asalzburger ? you mentioned that you can already misalign all the different examples detectors

paulgessinger:
@andiwand I think this is partially overlapping with @asalzburger's work to make the DD4hep construction use a factory to produce it's detector element objects.

asalzburger:
My work is based on Zequn's work, I simply generalised his first approach to all detectors. So I think we can close this one.

PR #3905, opened 2024-11-25
build: Update `FindFilesystem.cmake`

### LINKED ISSUE #3900: Compile acts/master on my mac
I have tried to compile on my mac the latest acts master branch.

It turned out that it was not able to compile due to the find_package(FileSystem REQUIRED)

. What i did has been to change in CMakeLists.txt, adding filesystem from boost 

```
            COMPONENTS program_options unit_test_framework filesystem

        find_package(Boost ${_acts_boost_version} REQUIRED COMPONENTS filesystem)
```

and fully comment out the 
```
# find_package(Filesystem REQUIRED)

```

With those 2 changes the code compile on my mac, after having installed with ```brew``` both ```eigen``` and ```boost```

Mac os version : 

```
$|=>clang --version
Apple clang version 15.0.0 (clang-1500.1.0.2.5)
Target: arm64-apple-darwin23.3.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Apple M3 Pro, MacOS 14.3 (23D56)

paulgessinger:
Vector-of-bool's `FindFilesystem.cmake` was updated at some point to include and explicit `CMAKE_CXX_STANDARD=17`. We've observed occasionally that on macOS, the find module would fail because it would default to C++14, where `std::filesystem` is not available (and I guess the experimental implementation has since been removed).

This PR updates the `FindFilesystem.cmake` to include this.

Fixes #3900

### DIFF
--- a/cmake/FindFilesystem.cmake
+++ b/cmake/FindFilesystem.cmake
@@ -103,6 +103,8 @@ if(TARGET std::filesystem)
     return()
 endif()
 
+cmake_minimum_required(VERSION 3.10)
+
 include(CMakePushCheckState)
 include(CheckIncludeFileCXX)
 
@@ -123,6 +125,9 @@ endif()
 cmake_push_check_state()
 
 set(CMAKE_REQUIRED_QUIET ${Filesystem_FIND_QUIETLY})
+
+# All of our tests required C++17 or later
+set(CMAKE_CXX_STANDARD 17)
 
 # Normalize and check the component list we were given
 set(want_components ${Filesystem_FIND_COMPONENTS})


paulgessinger:
Updated to reset the original value of `CMAKE_CXX_STANDARD` or unsetting if it was not set before.

cmake/FindFilesystem.cmake
```diff
@@ -124,6 +126,13 @@ cmake_push_check_state()
 
 set(CMAKE_REQUIRED_QUIET ${Filesystem_FIND_QUIETLY})
 
+# All of our tests required C++17 or later
+if(DEFINED CMAKE_CXX_STANDARD)
+    set(_prior_cmake_cxx_standard ${CMAKE_CXX_STANDARD})
+endif()
+
+set(CMAKE_CXX_STANDARD 17)
```
stephenswat, line 134: Is there a reason to use C++17 here if ACTS as a whole requires C++20 these days?

stephenswat (approved):
Happy with this; one non-blocking comment.

cmake/FindFilesystem.cmake
paulgessinger: Not really, vector-of-bool's implentation uses this, I guess because that's when `std::filesystem` entered the standard.

PR #3915, opened 2024-11-27
test: Use true start params for fitters in Examples Physmon

andiwand:
In order to factor out the effects of inaccurate starting parameters and potential navigation issues I propose to disable the start parameters smearing for the fitters in physmon. I also made the sigmas explicit.

### DIFF
--- a/CI/physmon/workflows/physmon_trackfinding_1muon.py
+++ b/CI/physmon/workflows/physmon_trackfinding_1muon.py
@@ -91,19 +91,6 @@ def run_ckf_tracking(label, seeding):
             s,
             setup.trackingGeometry,
             setup.field,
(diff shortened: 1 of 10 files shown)


paulgessinger:
Wouldn't this have consequences on the kind of pulls that we can see in the refitting? Or is this sufficiently covered by inflated uncertainties of the initial parameters?

andiwand:
> Wouldn't this have consequences on the kind of pulls that we can see in the refitting? Or is this sufficiently covered by inflated uncertainties of the initial parameters?

My thought was that we will only see the uncertainties of the measurements and cut out the initial parameter uncertainty which comes on top. Some of the residuals and pulls seems to get better but I also see weird effects in the endcaps. I have to take another look.

I think what we should rather do is start from the first measurement and not from the beamline which should also fix the problem of not reaching sensitives.

PR #3846, opened 2024-11-11
fix: oneAPI 2025 Fixes, main branch (2024.11.11.)

krasznaa:
After @CarloVarni's update in https://github.com/acts-project/traccc/pull/765, I became unable to build [traccc](https://github.com/acts-project/traccc) with [oneAPI 2025.0.0](https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-toolkit-release-notes.html). 😦 

The issue being solved here is of the type:

```
[159/737] Building CXX object Core/CMakeFiles/ActsCore.dir/src/TrackFitting/GlobalChiSquareFitter.cpp.o
FAILED: Core/CMakeFiles/ActsCore.dir/src/TrackFitting/GlobalChiSquareFitter.cpp.o
/software/intel/oneapi-2025.0.0/compiler/2025.0/bin/compiler/clang++  --gcc-toolchain=/software/gcc/13.3.0/x86_64-el9 -DActsCore_EXPORTS -I/home/krasznaa/ATLAS/projects/acts/acts/Core/include -I/home/krasznaa/ATLAS/projects/acts/build-oneapi/Core -isystem /home/krasznaa/ATLAS/projects/acts/acts/cmake/assert_include -isystem /software/boost/1.83.0/x86_64-el9-gcc11-opt/include -isystem /home/krasznaa/ATLAS/projects/traccc/build-vanilla/_deps/eigen3-src -Wall -Wextra -Wpedantic -Wshadow -Wzero-as-null-pointer-constant -Wold-style-cast  -O2 -g -DNDEBUG -std=gnu++20 -fPIC -MD -MT Core/CMakeFiles/ActsCore.dir/src/TrackFitting/GlobalChiSquareFitter.cpp.o -MF Core/CMakeFiles/ActsCore.dir/src/TrackFitting/GlobalChiSquareFitter.cpp.o.d -o Core/CMakeFiles/ActsCore.dir/src/TrackFitting/GlobalChiSquareFitter.cpp.o -c /home/krasznaa/ATLAS/projects/acts/acts/Core/src/TrackFitting/GlobalChiSquareFitter.cpp
In file included from /home/krasznaa/ATLAS/projects/acts/acts/Core/src/TrackFitting/GlobalChiSquareFitter.cpp:9:
[...]

PR #3938, opened 2024-12-03
fix: oneAPI 2025 Fixes, main branch (2024.12.04.)

krasznaa:
This is a twin of #3846. Unfortunately between that PR going in and [v38.0.0](https://github.com/acts-project/acts/releases/tag/v38.0.0) being made, some more errors of the same kind went in. This is meant to fix those.

--- END COMMIT MESSAGE ---

@paulgessinger, could we add a CI build with the latest clang version? That would pick up on these issues as well. We don't need oneAPI itself for catching these. :thinking:

## Summary by CodeRabbit

- **Bug Fixes**
	- Simplified method invocation syntax by removing the `template` keyword from `allocateCalibrated` and `allocateCalibrated_impl` calls, enhancing readability without altering functionality.
- **Tests**
	- Updated test cases to reflect the syntax change in method calls while preserving existing validation logic and error handling.

### DIFF
--- a/Core/include/Acts/EventData/MultiTrajectory.hpp
+++ b/Core/include/Acts/EventData/MultiTrajectory.hpp
@@ -696,7 +696,7 @@ class MultiTrajectory {
 
     visit_measurement(measdim, [this, istate]<std::size_t DIM>(
                                    std::integral_constant<std::size_t, DIM>) {
-      self().template allocateCalibrated_impl(
+      self().allocateCalibrated_impl(
           istate, ActsVector<DIM>{ActsVector<DIM>::Zero()},
           ActsSquareMatrix<DIM>{ActsSquareMatrix<DIM>::Zero()});
     });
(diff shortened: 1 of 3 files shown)


AJPfleger (approved):
looks, basically the same as in the last one.

Having this in the CI might be useful. Do we need a full build for this? Ideally we could combine it with testing, but this problem we would already see at compile time, no?

krasznaa:
Adding a Clang build would of course increase the CPU time of the CI. :thinking:

At the same time see https://github.com/acts-project/machines/pull/129 for that discussion. :wink:

acts-project-service:
## :red_circle: Athena integration test results [3d58bcafcc7157e763947b580dab1276bce25de3]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/47449916) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/9255332)!

PR #3943, opened 2024-12-04
ci: Add extra build job with clang 19

stephenswat:
clang 19 introduces new warnings (related to the use of the template keyword) which seem to be firing a lot in ACTS. Because OneAPI 2025 is built on clang 19, this is causing us problems when trying to compile ACTS as a dependency of traccc. To alleviate this issue I would like to introduce a clang 19-based build test so we can precent new warnings from sneaking into the ACTS codebase.

## Summary by CodeRabbit

- **Chores**
	- Updated GitHub Actions workflow configuration
	- Added new performance monitoring job
	- Updated container image for Ubuntu build
	- Enhanced build matrix with new Ubuntu version support


stephenswat:
This is now ready, but probably conflicts with #3980. @paulgessinger would you prefer if we shift to Spack first and then add more CI jobs afterwards?

paulgessinger:
@stephenswat Evidently, this runs with dependencies built using a different clang version. In #3980 I have dependencies from clang 19 as well. Maybe we can align this PR with #3980 so that it "only" changes where the deps come from, and not the overall job structure? (Haven't looked at this in detail yet)

paulgessinger:
We're now picking up a clang19 build through LCG 107. Do you prefer having a separate build on ubuntu + clang19 as well?

stephenswat:
No, I don't think that would be necessary. Thanks for the update, I'll close this!

PR #3951, opened 2024-12-05
refactor!: Always return a magnetic field vector

andiwand:
This is a proposal to switch the `Result<Vector3>` with just `Vector3`. The reason being that this should never fail and we end up doing a lot of error handling in the stepping code which seems not optimal for performance and also hurts readability.

At the same time I changed the interface and name of `getFieldAndGradient` which seems not to be used so maybe we should remove it?

We could take one step further and put `noexcept` on the `getField` functions to really enforce that these functions are meant to always return.

Also contains the usual include cleanups.


andiwand:
cc @paulgessinger @timadye @asalzburger

andiwand:
Discussed this with @paulgessinger - the main counterargument is that this can silence errors. When somebody wrongly defines the bounds of the magnetic field and we start stepping outside we get an explicit error right now. After the change we are forced to give some default value, and if the magnetic field is incorrectly scaled or placed the user will observe the default field value instead of getting an error.

IMO this sort of error case we have everywhere, in the geometry or material mapping for example. If the user provides wrong data we will usually return wrong outputs if it is not easily detectable.

The main idea here is not to silence these errors but to make the interface of the magnetic field provider simpler which directly influences downstream code to also become simpler. The interface is changed to always observe a value magnetic field, which is the physical reality. In case providers are bound to some subspace we need a fallback value which has to be chose explicitly by the user.

paulgessinger:
Just to state it again, I think having the interpolated B field return some fallback value outside of the interpolation domain is a mistake.

As an alternative would be to `std::abort` if we're out of bounds, instead of making this a handle-able error. That seems more user hostile to me but better than the fallback.

I don't think the interpolated B field values should have a fallback value.

Core/include/Acts/MagneticField/InterpolatedBFieldMap.hpp
```diff
@@ -200,10 +190,6 @@ class InterpolatedBFieldMap : public InterpolatedMagneticField {
     std::array<Vector3, nCorners> neighbors;
     const auto& cornerIndices = m_cfg.grid.closestPointsIndices(gridPosition);
 
-    if (!isInsideLocal(gridPosition)) {
-      return MagneticFieldError::OutOfBounds;
-    }
```
paulgessinger, line 205: I think this should be an exception or terminate the program.

andiwand: Right I just patched this up for now so it compiles. I wonder if the overflow bins can be used for this purpose?

andiwand:
@paulgessinger isn't this basically what we have underflow/overflow bins for?

Core/include/Acts/MagneticField/InterpolatedBFieldMap.hpp
```diff
@@ -278,47 +264,35 @@ class InterpolatedBFieldMap : public InterpolatedMagneticField {
...
-
-    return Result<Vector3>::success(
-        m_cfg.transformBField(m_cfg.grid.interpolate(gridPosition), position));
-  }
-
-  Vector3 getFieldUnchecked(const Vector3& position) const final {
+  Vector3 getField(const Vector3& position) const {
     const auto gridPosition = m_cfg.transformPos(position);
     return m_cfg.transformBField(m_cfg.grid.interpolate(gridPosition),
                                  position);
```
paulgessinger, line 270: I think relying on the over/underflow behavior of the grid here is dangerous. I think the chance that the user will intentionally and correctly initialize this is very small, with the overwhelming majority of cases being silent random magnetic field values.

andiwand: but that overflow bins are set can be asserted on when constructing the field no?

benjaminhuth:
I do not have a strong opinion on that, but I also would be generally hesitant in removing the error path, also because it is a virtual interface and we do not really know what kind of errors clients might want to raise in their use case.

andiwand:
I measure a very faint performance difference in the stepping benchmark. `perf1` is main, `perf2` is this branch
```
% ~/cern/build/acts/acts/perf1/bin/ActsBenchmarkEigenStepper
14:55:00    EigenStepper   INFO      propagating 20000 tracks with pT = 1GeV in a 2T B-field
14:55:02    EigenStepper   INFO      Execution stats: 20000 runs of 1 iteration(s), 319.5ms total, 15.9580+/-0.0032µs per run, 15958.000+/-3.219ns per iteration
14:55:02    EigenStepper   INFO      average path length = 5000mm
14:55:02    EigenStepper   INFO      average number of steps = 144
14:55:02    EigenStepper   INFO      step efficiency = 0.966443
% ~/cern/build/acts/acts/perf2/bin/ActsBenchmarkEigenStepper
14:55:04    EigenStepper   INFO      propagating 20000 tracks with pT = 1GeV in a 2T B-field
14:55:06    EigenStepper   INFO      Execution stats: 20000 runs of 1 iteration(s), 314.1ms total, 15.6660+/-0.0043µs per run, 15666.000+/-4.288ns per iteration
14:55:06    EigenStepper   INFO      average path length = 5000mm
14:55:06    EigenStepper   INFO      average number of steps = 144
14:55:06    EigenStepper   INFO      step efficiency = 0.966443
```

But the values seem to be compatible with noise level which is expected I think. We definitely end up with less instructions but seems not to matter for this CPU.

My usual full chain performance measure seems unaffected.

andiwand:
> I do not have a strong opinion on that, but I also would be generally hesitant in removing the error path, also because it is a virtual interface and we do not really know what kind of errors clients might want to raise in their use case.

I would not think of this as removing error paths but just requiring that the magnetic field is defined everywhere. How this is achieved is up to the implementation. I believe while querying the field there should never be an error or an exception because this is something that should be in memory and accessible at any point in time.

The only case we have where the field is not defined everywhere right now is the interpolated implementation and don't think that this is a good reason to force special casing on all the other components. This can be handled with overflow bins or a fallback value which has to be user supplied.

Core/include/Acts/MagneticField/InterpolatedBFieldMap.hpp
timadye: I don't think we should be second-guessing the users: if the service can return a well-defined B-field, then return that. We have a mechanism for under/overflow, so we should apply it, not give an error. I think it is reasonable to give an error if the under/overflow B-field are not filled. Certainly they should not be "random magnetic field values".

Is there a problem with the client code if we return `B=(0,0,0)`? If not, I personally think that would be a reasonable default even for unspecified bins. If outside the range of the magnet, we have zero (or residual) field. I realise that may be a minority view, so it's fine to require that all bins be specified.

stephenswat:
I have to say I rather like the way it is implemented now; exception are evil, default values are foot-guns. But I also understand that the amount of boilerplate needed to handle this is quite excessive. I think this is the result of us lacking two things right now. At its core, the `Result` type borrows from the way Rust handles errors, which is good™ exactly because it has these two features.

The first is some macro-based error handling. In Rust one might write

```rust
let Vector3 r = bfield.getField({12.0, -5.0, 32.0})?
```

Which translates to something like this in C++: 

```cxx
Vector3 r;

{
    Result<Vector3> tmp = bfield.getField({12.0, -5.0, 32.0});
    if (tmp.ok()) {
        r = *tmp;
    } else {
        return Result::failure();
    }
}
```

Which is exactly the type of boilerplate we're trying to remove here. Now, we all know the C++ macro system is unhygienic and terrible, but we could consider having a macro that does something similar like this.

The second thing a series of three operators on `Result`, which are as follows:

1. The `or` operator, which returns either the value in the `Result` or some default value. This way, `field.getValue(...).or({0., 0., 0.})` can elegantly model a default value of no magnetic field. This exists in `std::optional` as `std::optional::value_or` and `std::optional::or_else`.
2. The `map` operator, which applies a function to the `Result` or leaves it in an error state if no okay result exists. This is known as `fmap` in `Functor` in Haskell, and was added to `std::optional` as `std::optional::transform` in C++23 under P0798R8 which is slightly incorrectly named "monadic operators" because functors are not monads, but we also need:
3. The `bind` operator, which is so important that the syntax used to denote this operator, `>>=`, is literally the Haskell logo. Classically, bind has the signature `m a -> (a -> m b) -> m b` and is more powerful than `map` because it is capable of peeling away the outer monadic layer automatically. This is added to `std::optional` as `std::optional::and_then`.

These operators can be easily added to `Result` and should make them a lot easier to use, stripping away a lot of the boilerplate. For the remainder of the boilerplate, perhaps we should add a macro for handling `Result` values? That way we can enjoy the flexibility of having this nice way to return errors while not having to deal with as much boilerplate.

andiwand:
To provide another example: Athena also seems to always provide a field vector https://gitlab.cern.ch/atlas/athena/-/blob/main/Tracking/Acts/ActsGeometry/ActsGeometry/ATLASMagneticFieldWrapper.h#L38-55

osbornjd:
> Just to state it again, I think having the interpolated B field return some fallback value outside of the interpolation domain is a mistake.

I would echo Paul on this, I agree that this is a mistake. It has the potential to allow effectively silent treatment of a situation that should not happen. As an explicit example, right now in production we have tracks that have poorly estimated track parameters because they are not corrected for misalignments/distortions etc. Trying to fit these tracks returns some error message that the field is out of bounds, and with the current proposition this would just be fit with some default field value which may be actually wrong (for example, if the track really points out of our default ~constant field volume and so it bends by some non constant field value but we try to fit it with the constant field value, that will actually be wrong). 

I recognize the value in trying to remove boiler plate code but I think removing this and just silently continuing on would lead to potential confusion on what the KF is actually doing behind the scenes.

andiwand:
> > Just to state it again, I think having the interpolated B field return some fallback value outside of the interpolation domain is a mistake.
> 
> I would echo Paul on this, I agree that this is a mistake. It has the potential to allow effectively silent treatment of a situation that should not happen. As an explicit example, right now in production we have tracks that have poorly estimated track parameters because they are not corrected for misalignments/distortions etc. Trying to fit these tracks returns some error message that the field is out of bounds, and with the current proposition this would just be fit with some default field value which may be actually wrong (for example, if the track really points out of our default ~constant field volume and so it bends by some non constant field value but we try to fit it with the constant field value, that will actually be wrong).
> 
> I recognize the value in trying to remove boiler plate code but I think removing this and just silently continuing on would lead to potential confusion on what the KF is actually doing behind the scenes.

But I think this is exactly what this error is not supposed to do. The reason you see an error there is that the navigation fails and you leave the tracking geometry volume and end up in an area the magnetic field is not defined anymore. So we fail at a very late state with an error that has nothing to do with the original cause.

Even without this error, the propagation would reach the maximum number of steps at some point also causing an error which actually gives you a deeper insight of what is going on.

paulgessinger:
@stephenswat I personally like the idea of tackling the boilerplate, rather than the value guarantees.

andiwand:
> @stephenswat I personally like the idea of tackling the boilerplate, rather than the value guarantees.

I don't think these changes change anything about the boilerplate but most likely increase it than decrease it. We always have the situation that we want the field and work with the field. If we do not get the field for a weird reason we lazy error out at each occasion. There is nothing else we can do.

paulgessinger:
@andiwand 
> There is no case we can recover from not having a field.

This I believe underlines my point about not returning a default value in the interpolated B field. I don't think there's a way not to have a default value (i.e. error out), except for maybe a configuration the field class to throw an exception or abort.

In case where there's just no correct field value to be returned, imo returning an error like we do is the one and only correct thing to do. I think we just have to pay the price for this (imo) fact.

Core/include/Acts/MagneticField/InterpolatedBFieldMap.hpp
paulgessinger: There's no way to opt out of the under/overflow behavior if we remove the error mechanism.

andiwand:
But we can also solve this problem by requiring to have a valid field value at every point in space if this is not a case we can handle downstream anyways.

As I suggested before, the interpolated implementation can return a field value by querying the overflow bin. That the overflow bin is set can be a hard requirement on construction. We are not hiding anything under the rug. The physical appropriate value is 0 as there is no magnetic field outside the detector. But this can be chosen by the user. We are not setting it to 0 anywhere with that suggestion.

paulgessinger:
@andiwand I don't believe there's a way to *unset* the under/overflow bins. We could use `nan` as a sentinel for this, but then we need extra checks on this, otherwise we just propagate `nan`s.

> The physical appropriate value is 0 as there is no magnetic field outside the detector. But this can be chosen by the user. We are not setting it to 0 anywhere with that suggestion.

I disagree with this. The only physical appropriate value outside of the detector is *no value*. If the user wants to define a field outside of their detector, they are free to provide a magnetic field instance that covers whatever query domain they desire.

paulgessinger:
@andiwand 

> I don't think these changes change anything about the boilerplate but most likely increase it than decrease it.

I think you can achieve 80% of the diff of this PR with two macros similar to what @stephenswat was suggesting before.

osbornjd:
> Fundamentally we don't have a field outside of the interpolation domain. We would be choosing to return something instead, which is wrong, but it silences our errors.

> In case where there's just no correct field value to be returned, imo returning an error like we do is the one and only correct thing to do. I think we just have to pay the price for this (imo) fact.

I would agree with this statement also. Returning something where the user provided nothing seems like a dangerous path to go down, and feels like it could cause more confusion than it prevents.

andiwand:
> I think you can achieve 80% of the diff of this PR with two macros similar to what @stephenswat was suggesting before.

Ah I missed that part with the macro. I am not a fan of macros but in this case it might make sense. We have this kind of error propagation everywhere in our code and it seems quite verbose. Also choosing variable names becomes difficult in these situations.

I can see that making sense if we want to adopt it as a general solution for dealing with `Result`.

> @andiwand I don't believe there's a way to _unset_ the under/overflow bins. We could use `nan` as a sentinel for this, but then we need extra checks on this, otherwise we just propagate `nan`s.

But that sounds more like limitation of the Grid implementation no?

> This I believe underlines my point about not returning a default value in the interpolated B field. Fundamentally _we don't have a field_ outside of the interpolation domain. We would be choosing to return _something_ instead, which is wrong, but it silences our errors.
> 
> In case where there's just no correct field value to be returned, imo returning an error like we do is the one and only correct thing to do. I think we just have to pay the price for this (imo) fact.
> 
> This is not true. The B field can be defined correctly inside the interpolation domain, it cannot be defined outside.
We can patch it and suppress the error outside the interpolation domain by returning something arbitrary. I don't think that's the same thing.

But I feel like we chose the interface given one very specific case here. Potentially two different kinds of interfaces would make sense: a partial magnetic field and a physical magnetic field. Each one serving the purpose of the different implementation needs. But I can see that this might overcomplicate things.

benjaminhuth:
> But I feel like we chose the interface given one very specific case here.

I would agree with @andiwand, that there are to different concepts of a field that are discussed here. But the interface we have is actually is very generic, and allows to represent all cases...

stephenswat:
> > 
> 
> I would agree with @andiwand, that there are two different concepts of a field that are discussed here. But the interface we have is actually is very generic, and allows to represent all cases...

I agree with this. The question in my mind then becomes: how do we want to express this difference between magnetic fields with finite domains versus fields with infinite domains. I see three options:

1. We split the virtual interface into two types, `MagneticFieldProvider` and `UnboundedMagneticFieldProvider` (names TBD) and ask for either one of these depending on the needs of the call site.
2. We extend the magnetic field provider with a new method `getFieldForRealTrustMeBro` (name also TBD) and let the call site choose the function they want depending on the need.
3. The call site expresses the default value explicitly, e.g. using `field.getField(x, y, z).value_or(0, 0, 0)`.

benjaminhuth:
Hmm tbh, options 1 and 2 are not practical, because I'm not sure how we would define the need of the call site. 
The call site often is also in Core (e.g., the stepper, the parameter estimation, ...). So here, we would probably again choose the more generic option, not saving any boilerplate. Or we would choose the unbounded case for those cases, and then we could just merge this PR...

andiwand:
I tried the macro approach and I fear that is a dead end. You can find some example code here https://godbolt.org/z/TT8c1fxfn.

@stephenswat provided me a snipped similar to this
```c++
#define GetValueOrReturn(x) ({ \
if (!x.ok()) return Result<std::string>::failure(x.error()); \
x.value(); \
})
```

which can be used somewhat like this

```c++
std::string result = GetValueOrReturn(Result<std::string>(""));
```

There seem to be two major problems with this
- The control flow is hidden inside the macro ⚡💣 💥 💀 
- This seems to require some GNU C++ extensions ⚡💣 💥 💀 

Ultimately this leaves us with two options IMO for the overall boilerplate withe `Acts::Result`
- buy it, as we do now
- use exceptions instead, only switch to `Acts::Result` if there is a measurable difference

To me, C++ has a way of handling recoverable "errors" within the language and since we do not use `noexcept` widely we pay for it anyways right now. So in the most cases exception should come without additional costs and where it does we can enforce `noexcept` and use `Acts::Result`.

For the specific case here with the field this would mean that some implementations might throw which then bubbles up the propagation stack and we could still avoid `Acts::Result`. Or as initially proposed we get rid of the error case by demanding the field is defined everywhere, enforcing this even further with a `noexcept` on the interface. For a clean separation we could have two magnetic field interfaces as discussed above. But agree that this might be too much and the generic case is that it errors / throws even though this hurts the calling code in 99.9% without impact.

paulgessinger:
@andiwand what do think about this? I was not in favor of this, but we can re-litigate this if you still want to push it. Otherwise, should we close?

andiwand:
I will close this for now - no time to look at this right now and there was no consens

PR #3957, opened 2024-12-05
feat: Add monadic functions to `Result`

stephenswat:
In order to make the `Result` type easier to use, this commit adds three new functions:

 * `Result::value_or` allows the user to obtain the value or a provided default value.
 * `Result::transform` models functorial mapping, allowing users to modify values inside results.
 * `Result::and_then` models monadic binding, allowing users to build complex chains of actions on results.

Implemented are lvalue and rvalue versions of these functions as well as tests.

image

## Summary by CodeRabbit

- **New Features**
	- Enhanced `Result` class with new methods for improved error handling and value transformation:
		- `value_or`: Retrieve valid value or default.
		- `transform`: Apply a function to the valid value.
		- `and_then`: Chain functions based on the result's validity.

- **Tests**
	- Added comprehensive test cases to validate the new functionalities of the `Result` class, ensuring reliability and correctness:
		- `ValueOrResult`: Validates the `value_or` method.
		- `TransformResult`: Assesses the `transform` method.
		- `AndThenResult`: Evaluates the `and_then` method.


Core/include/Acts/Utilities/Result.hpp
```diff
@@ -172,6 +175,144 @@ class Result {
...
 
+  /// Retrieves the valid value from the result object, or returns a default
+  /// value if no valid value exists.
+  ///
+  /// @param[in] v The default value to use if no valid value exists.
+  /// @note This is the lvalue version.
+  /// @note This function always returns by value.
+  /// @return Either the valid value, or the given substitute.
+  template <typename U>
+  T value_or(U&& v) const&
```
andiwand, line 186: this will copy `value()` if `ok()`?

Core/include/Acts/Utilities/Result.hpp
```diff
@@ -172,6 +175,144 @@ class Result {
...
+  /// @param[in] callable The transformation function to apply.
+  /// @note This is the lvalue version.
+  /// @note This functions is `fmap` on the functor in `A` of `Result<A, E>`.
+  /// @return The modified valid value if exists, or an error otherwise.
+  template <typename C>
+  auto transform(C&& callable) const&
+    requires std::invocable<C, const T&>
+  {
+    using R =
+        Result<std::decay_t<decltype(std::declval<C>()(std::declval<T>()))>, E>;
```
andiwand, line 228: this hurts my brain - but I guess there is no way around it?

benjaminhuth: maybe lambdas in unevaluated contexts could help here, but I don't think its that bad

Core/include/Acts/Utilities/Result.hpp
stephenswat: Yes, I think that makes sense because it's the lvalue version; do we cannot move the value out.

andiwand: but in principle `value()` could allow you to return a `const &` right? but I guess the supplied default makes this impossible?

benjaminhuth: Would an overload taking `U` as (const) lvalue ref and returning everything as (const) lvalue ref make sense?

stephenswat: I think I managed to make this work with some conditional typing. Take a look!

Core/include/Acts/Utilities/Result.hpp
stephenswat: This is now two smaller typedefs, hopefully aiding readability.

andiwand (approved):
LGTM 👍

stephenswat:
What on earth is this CI failure? :thinking:

andiwand:
never seen this before. maybe @benjaminhuth @paulgessinger ?

stephenswat:
Should be solved by #3967.

acts-project-service:
## :red_circle: Athena integration test results [9067679f3908b08e8ce55c5c61d75fde0b9061a9]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/47795377) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/9449137)!

PR #3983, opened 2024-12-13
refactor!: Remove `MagneticFieldProvider::getFieldGradient`

andiwand:
Remove `MagneticFieldProvider::getFieldGradient` after discussions as it is not used anywhere.

---

previously:

- Remove output param
- Return `Result<std::pair<Vector3, SquareMatrix3>>` instead of `Result<Vector3>`
- Actually implement this function correctly

pulled out of https://github.com/acts-project/acts/pull/3951

## Release Notes

- **Refactor**
  - Removed `getFieldGradient` method from multiple magnetic field classes, streamlining their interfaces.
  - Updated destructor implementation in `MagneticFieldProvider` for simplification.

- **Chores**
  - Cleaned up unused header file inclusions.
  - Removed unimplemented gradient calculation methods.

The changes primarily focus on enhancing the clarity and usability of the magnetic field-related code by eliminating incomplete functionality.


andiwand:
alternatively we could drop this function as it is not used anywhere

paulgessinger:
@andiwand I'm wondering if dropping might be the better idea for now.

PR #3986, opened 2024-12-14
refactor!: Revise stepper update functions and constrained step types

andiwand:
- remove default parameters from stepper update functions
- remove `release` flag from update functions. this is an important part of our propagation which should not be put under the rug
- add `ConstrainedStep::Type` param so it can be used for other purposes than just the navigation slot
- rename `ConstrainedStep::Type`s
  - `actor` -> `navigator` since we only use it for navigation
  - `aborter` -> `actor` since we only use it in the actors (aborters have been merged into actors)

## Release Notes

- **New Features**
	- Enhanced handling of step sizes and surface status updates across various classes, including `AtlasStepper`, `EigenStepper`, and `StraightLineStepper`, now incorporating a new parameter for step size type.
	- Introduction of a `getStepSize` method for better access to step sizes based on type.

- **Bug Fixes**
	- Adjusted method calls to ensure consistent use of `ConstrainedStep::navigator` instead of `ConstrainedStep::actor`, improving accuracy in step size management.

- **Documentation**
	- Updated comments and documentation across various components to enhance clarity regarding functionality and usage.

- **Tests**
	- Comprehensive updates to unit tests to reflect changes in method signatures and ensure correct functionality of the stepper classes with the new parameters.


Core/include/Acts/Propagator/detail/SteppingHelper.hpp
```diff
@@ -28,61 +26,53 @@ namespace Acts::detail {
...
+/// @param stype [in] The step size type to be set
+/// @param logger [in] A @c Logger instance
 template <typename stepper_t>
 Acts::IntersectionStatus updateSingleSurfaceStatus(
     const stepper_t& stepper, typename stepper_t::State& state,
-    const Surface& surface, std::uint8_t index, Direction navDir,
+    const Surface& surface, std::uint8_t index, Direction direction,
     const BoundaryTolerance& boundaryTolerance, double surfaceTolerance,
-    const Logger& logger) {
+    ConstrainedStep::Type stype, const Logger& logger) {
```
paulgessinger, line 45: I suppose there's no point in defaulting this? What values does it take de-factor?

Core/include/Acts/Propagator/ConstrainedStep.hpp
```diff
@@ -44,10 +42,10 @@ namespace Acts {
...
  public:
   /// the types of constraints
-  /// from actor    - this would be a typical navigation step
-  /// from aborter  - this would be a target condition
-  /// from user     - this is user given for what reason ever
-  enum Type : int { actor = 0, aborter = 1, user = 2 };
+  /// from navigator - this would be a navigation step
+  /// from actor     - this would be an actor condition
+  /// from user      - this is user given for what reason ever
+  enum Type : int { navigator = 0, actor = 1, user = 2 };
```
paulgessinger, line 48: Consider using the opportunity to make this an `enum class`.

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -622,7 +622,7 @@ class CombinatorialKalmanFilter {
                 boundParams.referenceSurface().getSharedPtr());
           }
 
-          stepper.releaseStepSize(state.stepping, ConstrainedStep::actor);
+          stepper.releaseStepSize(state.stepping, ConstrainedStep::navigator);
```
paulgessinger, line 625: So this I find a bit weird. Not saying that it made a ton of sense before, but in this case, we're unsetting the navigation constraint from the actor. This is a consequence of the renaming, so I understand this doesn't actually change the behavior (the navigator was setting the actor constraint before, if I'm not mistaken.).

Not sure what to do about this though.

paulgessinger (approved):

Core/include/Acts/Propagator/ConstrainedStep.hpp
andiwand: That is a good point. I think we use the unscoped enum like this `ConstrainedStep::navigator` right now which would change to `ConstrainedStep::Type::navigator` and might be a bit more verbose. We can also think about making this upper case.

What do you think? I would push this to a follow-up PR since we are in a breaking cycle anyways

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
andiwand: I would argue that this is a consequence of running multiple propagations inside a single propagation. We also reset the stepper and navigator state which is equally confusing given we are in an actor IMO

Core/include/Acts/Propagator/ConstrainedStep.hpp
paulgessinger: You can introduce a 

```cpp
struct ConstrainedStep {
...
using enum Type;
...
};
```

I think as of C++20, which would give the same symbols as before while still having the benefits of `enum class`.

With that, renaming to upper case is orthogonal to this PR.

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [567db0c727c2d8f9ffb7e803f4db715919d36142]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/49123829) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/9839391)!

PR #3988, opened 2024-12-16
feat: Marking Edgeholes during track finding

pbutti:
Added track state type for edge holes. 
Added marking of track states as edge holes in CKF Actor with 2mm tolerance (to be configured/changed)
Currently option is propagated via propagator plain options (to be changed). 
Tried to add nEdgeHoles to VectorTrackContainer but fails. 

--- END COMMIT MESSAGE ---

@andiwand @paulgessinger 


Core/include/Acts/Surfaces/AnnulusBounds.hpp
```diff
@@ -209,6 +209,11 @@ class AnnulusBounds : public DiscBounds {
 
   /// Private helper method
   double squaredNorm(const Vector2& v, const SquareMatrix2& weight) const;
+
+private:
+  bool inside(const Vector2& lposition, double tolR,
+              double tolPhiR,
+              double scale) const;
```
andiwand, line 216: I think this can be replaced by a proper implementation of the inside check given a tolerance

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
```diff
@@ -823,7 +828,23 @@ class CombinatorialKalmanFilter {
...
+            if (!keepEdgeHoles) {
+              currentBranch.nHoles()++;
+            } 
+            else {
+              auto typeFlags = currentState.typeFlags();
+              typeFlags.reset(TrackStateFlag::HoleFlag);
+              typeFlags.set(TrackStateFlag::EdgeHoleFlag);
+              //currentBranch.nEdgeHoles()++;
+            }
+           }
```
andiwand, line 847: IMO this is an experiment choice and it should not be part of the Core CKF. This can be put into the track state creator to put the user in control

Core/src/Surfaces/RectangleBounds.cpp
```diff
@@ -16,6 +16,15 @@
...
     const Acts::BoundaryTolerance& boundaryTolerance) const {
+  if (boundaryTolerance.hasAbsoluteCartesian()) {
+     
+     const BoundaryTolerance::AbsoluteCartesian& tolerance = boundaryTolerance.asAbsoluteCartesian();
+     return    lposition(0,0) > (m_min.x()-tolerance.tolerance0)
+            && lposition(0,0) < (m_max.x()+tolerance.tolerance0)
+            && lposition(1,0) > (m_min.y()-tolerance.tolerance1)
+            && lposition(1,0) < (m_max.y()+tolerance.tolerance1);
+  }
+
```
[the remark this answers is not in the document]
pbutti: @goetzgaycken Can you comment on this?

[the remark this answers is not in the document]
goetzgaycken: It was just easier to implement the negative tolerance, this way then trying to add it to the more generic implementation.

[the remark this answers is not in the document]
paulgessinger: Doesn't `insideAlignedBox` do exactly the same thing as the first thing before going into anything else?

Core/include/Acts/Surfaces/AnnulusBounds.hpp
goetzgaycken: passing the "scale" parameter avoids a division. But anyway the way it is implemented duplicates code. So, as mentioned at a different place. This was the easiest way to implement it and was anyway more meant to be a proof of concept. And so far we only see that the concept does not work all that great ;-)

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -35,7 +35,7 @@ namespace Acts {
 template <typename object_t>
 struct NavigationOptions {
   /// The boundary check directive
-  BoundaryTolerance boundaryTolerance = BoundaryTolerance::None();
+   BoundaryTolerance boundaryTolerance = BoundaryTolerance::AbsoluteCartesian{2*UnitConstants::mm,2*UnitConstants::mm};
```
goetzgaycken: This is why the modules get bigger.  However, without handling of multiple modules passing the intersection, this is I suppose counterproductive.

Core/src/Surfaces/RectangleBounds.cpp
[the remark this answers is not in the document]
goetzgaycken: - I have to admit that I do not understand how the methods handles the case that the point is inside the polygon. Will computeClosestPointOnPolygon return the point itself if it is inside the polygon ? If not than I would not understand how the tolerance works .
- anyway currently tolerance.isTolerated() does not handle a negative tolerance, and the passed distance does not carry any information whether the point is inside or outside unless "inside" will always yield a zero distance.
- far more complicated than this simple set of 4 comparisons.

Core/include/Acts/Propagator/Navigator.hpp
pbutti: Just to clarify and resolve, this is not in this current PR. I've removed this from the navigator

Core/include/Acts/TrackFinding/CombinatorialKalmanFilter.hpp
goetzgaycken: > IMO this is an experiment choice and it should not be part of the Core CKF. This can be put into the track state creator to put the user in control

Not sure whether there is a universally accepted definition of what a hole is.  Is it really a how if the edge is within the uncertainties of the trajectory ? It is a definition, but I guess one can argue either way, whether it should be counted as a hole or not. Sure an ad hoc  2mm tolerance is not universal.

paulgessinger:
@pbutti @goetzgaycken @andiwand what's the plan for this? Is still relevant?

andiwand:
I think this should be based on https://github.com/acts-project/acts/pull/3990 - as soon as I can come back to this

andiwand:
I think this can be purely done on the experiment side now? We will have to check when we get back to this, and in case close this

AJPfleger:
@pbutti @goetzgaycken @andiwand Is this still relevant?

andiwand:
closing due to inactivity

PR #4012, opened 2025-01-09
feat: Add support for CUDA streams in GNN plugin

benjaminhuth:
Uses the torch framework (as it is required anyways currently) as a source of cuda streams. Extends the interface of the some components to use the streams.

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **New Features**
	- Introduced `ExecutionContext` structure to manage device and stream information across ExaTrkX plugin components.
	- Enhanced execution context handling for graph construction, edge classification, and track building processes.

- **Refactor**
	- Updated method signatures in multiple classes to use `ExecutionContext` instead of direct `torch::Device` parameters.
	- Improved device and stream management in ExaTrkX pipeline components.

- **Chores**
	- Standardized execution context management across ExaTrkX plugin implementation.


Plugins/ExaTrkX/include/Acts/Plugins/ExaTrkX/Stages.hpp
```diff
@@ -13,13 +13,20 @@
...
 namespace Acts {
 
 /// Error that is thrown if no edges are found
 struct NoEdgesError : std::exception {};
 
+/// Capture the context of the execution
+struct ExecutionContext {
+  torch::Device device{torch::kCPU};
+  std::optional<c10::cuda::CUDAStream> stream;
+};
```
stephenswat, line 28: What is the meaning here if the device is set to CPU but the CUDA stream is set to something? Or is that an invalid state? What about if the device is set to a GPU but the stream is unset? Does it then use the default stream?

Plugins/ExaTrkX/src/BoostTrackBuilding.cpp
```diff
@@ -48,7 +48,7 @@ namespace Acts {
 
 std::vector<std::vector<int>> BoostTrackBuilding::operator()(
     std::any /*nodes*/, std::any edges, std::any weights,
-    std::vector<int>& spacepointIDs, torch::Device) {
+    std::vector<int>& spacepointIDs, const ExecutionContext& execContext) {
```
stephenswat, line 51: Is it on purpose that you give no default value here?

Plugins/ExaTrkX/src/ExaTrkXPipeline.cpp
```diff
@@ -39,11 +39,18 @@ std::vector<std::vector<int>> ExaTrkXPipeline::run(
     std::vector<float> &features, const std::vector<std::uint64_t> &moduleIds,
     std::vector<int> &spacepointIDs, const ExaTrkXHook &hook,
     ExaTrkXTiming *timing) const {
+  ExecutionContext ctx;
+  ctx.device = m_graphConstructor->device();
+#ifndef EXATRKX_CPUONLY
+  if (ctx.device.type() == torch::kCUDA) {
+    ctx.stream = c10::cuda::getStreamFromPool(ctx.device.index());
+  }
```
stephenswat, line 47: I would suggest adding an assert for the case where `EXATRKX_CPUONLY` is set but the device type is not set to CPU.

Plugins/ExaTrkX/src/OnnxEdgeClassifier.cpp
```diff
@@ -90,7 +90,8 @@ std::ostream &operator<<(std::ostream &os, Ort::Value &v) {
 
 std::tuple<std::any, std::any, std::any, std::any>
 OnnxEdgeClassifier::operator()(std::any inputNodes, std::any inputEdges,
-                               std::any inEdgeFeatures, torch::Device) {
+                               std::any inEdgeFeatures,
+                               const ExecutionContext &execContext) {
   auto torchDevice = torch::kCPU;
```
stephenswat, line 95: Should this not depend on the execContext?

Plugins/ExaTrkX/src/TorchEdgeClassifier.cpp
```diff
@@ -67,17 +67,21 @@ TorchEdgeClassifier::~TorchEdgeClassifier() {}
...
+                                std::any inEdgeFeatures,
+                                const ExecutionContext& execContext) {
+  const auto& device = execContext.device;
+  decltype(std::chrono::high_resolution_clock::now()) t0, t1, t2, t3, t4;
   t0 = std::chrono::high_resolution_clock::now();
   ACTS_DEBUG("Start edge classification, use " << device);
   c10::InferenceMode guard(true);
 
   // add a protection to avoid calling for kCPU
 #ifndef ACTS_EXATRKX_CPUONLY
```
stephenswat, line 79: Before this was called `EXATRKX_CPUONLY`, now it's `ACTS_EXATRKX_CPUONLY`. Is that intentional?

Plugins/ExaTrkX/include/Acts/Plugins/ExaTrkX/Stages.hpp
benjaminhuth: Yeah so this definitively can be in an inivalid state. I wanted to keep it lightweight for now, but of course could add constructors to ensure validity of the state.
what happens if the stream is not set but we are on GPU is decided in the modules I would say.

stephenswat: Okay!

Plugins/ExaTrkX/include/Acts/Plugins/ExaTrkX/BoostTrackBuilding.hpp
```diff
@@ -25,7 +25,7 @@ class BoostTrackBuilding final : public Acts::TrackBuildingBase {
   std::vector<std::vector<int>> operator()(
       std::any nodes, std::any edges, std::any edge_weights,
       std::vector<int> &spacepointIDs,
-      torch::Device device = torch::Device(torch::kCPU)) override;
+      const ExecutionContext &execContext = {}) override;
```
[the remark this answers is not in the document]
benjaminhuth: I will remove this inconsitency later

PR #4027, opened 2025-01-13
fix: Don't reset component state in `MultiStepperSurfaceReached`

andiwand:
After facing another problem with the GSF in https://github.com/acts-project/acts/pull/3449 this might resolve it. The problem seems to be that the navigator and the surface reached aborter are fighting over the surface reached status which can lead to removing all components if the target surface is not reached and the maximum step trials for the multi stepper are reached.

This might introduce other problems so waiting for physmon report.

blocked by
- https://github.com/acts-project/acts/pull/4025


andiwand:
This seems to trigger other problems

benjaminhuth (approved):
Looks good to me! Maybe a adept the PR title for the current set of changes?

PR #4030, opened 2025-01-14
refactor: `ProtoLayer` can (optionally) hold mutable surfaces

paulgessinger:
In the Gen3 building context, using mutable pointers makes more sense since we store as much as possible internally as mutable.

This is implemented in a API backwards compatible way, because the Gen1 building assumes the opposite.

PLEASE DESCRIBE YOUR CHANGES.
DO NOT USE @-MENTIONS HERE!

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **Refactor**
	- Restructured `ProtoLayer` class to improve modularity and flexibility.
	- Introduced a template-based approach for handling const and non-const surface pointers.
	- Moved core functionality into a base class with enhanced parameter calculation methods.

- **New Features**
	- Added support for both const and mutable surface operations.
	- Implemented more flexible surface handling in geometry components.

- **Tests**
	- Updated unit tests to support const and non-const `ProtoLayer` scenarios.
	- Enhanced test coverage for different surface pointer types.


Core/src/Geometry/ProtoLayer.cpp
```diff
@@ -16,89 +16,59 @@
 using Acts::VectorHelpers::perp;
 using Acts::VectorHelpers::phi;
 
-namespace Acts {
+namespace Acts::detail {
```
asalzburger, line 19: At some point we need to clarify what we put in details and what we don't. Here it makes a lot of sense though.

asalzburger (approved):
Makes sense to me.

asalzburger:
There's one SonarCloud issue also:

`Remove these redundant parentheses.`

Core/src/Geometry/ProtoLayer.cpp
paulgessinger: Fair. In this case, I'm using it to indicate it shouldn't be used directly.

paulgessinger:
Updated branch and fixed sonar issue.

Can you go again @asalzburger?

asalzburger (approved):

acts-project-service:
## :red_circle: Athena integration test results [7b2b9360dfdac717726cd88a66d1d000c0aa726e]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/50856261) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/10316474)!

PR #4036, opened 2025-01-17
feat: Allow propagator initialization to fail

andiwand:
This change allows the propagator initialization to fail which can happen if for example pre conditions of the navigator are not met. To achieve this the `initialize` function in the propagator was elevated to be part of the public interface.

I think it would make sense to use this `initialize` function to also initialize the stepper and to remove that from `makeState`. That way the user of the propagator can reuse the propagation state after `initialize`.

Breakdown of the changes
- add return type to `Navigator::initialize`
- elevate `Propagator::initialize` to public API
- don't call `Propagator::initialize` in `Propagator::makeState`
- align propagator functions
- `namespace Acts {` in `Propagator.ipp`

## Summary by CodeRabbit

- **New Features**
	- Introduced structured status returns for improved error reporting, enabling clearer success and failure feedback.
- **Refactor**
	- Streamlined state initialization and overall API interfaces across navigation, propagation, and track fitting modules.
	- Removed redundant parameters to simplify state management.
- **Bug Fixes**
	- Enhanced error handling in initialization processes across various components to ensure robustness.
- **Chores**
	- Updated build configurations and dependency management to support the enhanced error handling mechanisms.


andiwand:
https://github.com/acts-project/acts/pull/4036/commits/7e4563da29a7fad537d39f46265ce95f1555f7ef comes with some measurable slowdowns

image

image

andiwand:
With the previous implementation no slowdowns are observed

Core/include/Acts/Propagator/DirectNavigator.hpp
```diff
@@ -171,9 +172,9 @@ class DirectNavigator {
   /// @param position The start position
   /// @param direction The start direction
   /// @param propagationDirection The propagation direction
-  void initialize(State& state, const Vector3& position,
-                  const Vector3& direction,
-                  Direction propagationDirection) const {
+  Result<void> initialize(State& state, const Vector3& position,
```
benjaminhuth, line 175: I would suggest using `nodiscard` for all initialize calls, since it is dangerous that we silently ignore the return value here, especially, because it was a `void` function before.

andiwand: one more reason why exceptions would be better

benjaminhuth: actually true. Now where I think of it, the whole argument of "it forces me to handle errors" breaks down in case of `void` functions...

benjaminhuth: But this would be circumvented if `no_discard` would be the default of course...

andiwand:
Didn't measure any significant change in CPU timings for fatras or CKF

PR #4039, opened 2025-01-20
refactor!: Decouple stepping from navigation

andiwand:
Very similar to https://github.com/acts-project/acts/pull/3449 this decouples the stepping from the navigation. Instead of passing the whole propagation state into stepping function we can just pass in the stepping state. This cuts all dependencies from the navigation and allows to use and test steppers independent from a propagation state and navigation.


acts-project-service:
## :red_circle: Athena integration test results [7c2993d64222a0832ad42529b43ce89d6a317230]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/49615134) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/9930990)!

paulgessinger:
I'll look at this tomorrow!

Core/include/Acts/Propagator/Propagator.ipp
```diff
@@ -258,11 +252,6 @@ auto Acts::Propagator<S, N>::makeState(
   StateType state{eOptions, m_stepper.makeState(eOptions.stepping),
                   m_navigator.makeState(eOptions.navigation)};
```
paulgessinger, line 254: This can't happen anymore? I believe the idea was mainly to give a better error message.

Core/include/Acts/Propagator/Propagator.ipp
```diff
@@ -258,11 +252,6 @@ auto Acts::Propagator<S, N>::makeState(
...
 
-  static_assert(
-      detail::propagator_stepper_compatible_with<S, StateType, N>,
-      "Step method of the Stepper is not compatible with the propagator "
-      "state");
-
```
paulgessinger, line 265: This can't happen anymore? I believe the idea was mainly to give a better error message. I see that the outer state is now not passed into the stepper anymore, but the stepping state could still be inconsistent I guess? In any case, the `static_assert` would have to move up. I guess with the state being a simpler type in the stepper functions, the error message might be good enough on it's own though.

Core/src/Propagator/SympyStepper.cpp
```diff
@@ -77,6 +77,12 @@ SympyStepper::boundState(
       state.pathAccumulated, freeToBoundCorrection);
 }
 
+bool SympyStepper::prepareCurvilinearState(State& state) const {
+  // TODO implement like in EigenStepper
```
paulgessinger, line 81: TODO?

Core/src/Propagator/SympyStepper.cpp
```diff
@@ -120,8 +126,16 @@ void SympyStepper::transportCovarianceToBound(
       freeToBoundCorrection);
 }
 
+Result<double> SympyStepper::step(State& state, Direction propDir,
+                                  const IVolumeMaterial* material) const {
+  (void)material;
```
paulgessinger, line 131: Nitpick:

```suggestion
  static_cast<void>(material);
```

for new code, ideally.

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -235,9 +235,8 @@ struct GsfActor {
     }
 
     for (auto cmp : stepper.componentIterable(state.stepping)) {
-      auto singleState = cmp.singleState(state);
-      cmp.singleStepper(stepper).transportCovarianceToBound(
-          singleState.stepping, surface);
+      cmp.singleStepper(stepper).transportCovarianceToBound(cmp.state(),
```
paulgessinger, line 238: I think this previously created a copy, and now it passes a reference. I'm assuming this is safe?

Core/include/Acts/Propagator/AtlasStepper.hpp
```diff
@@ -1149,32 +1148,36 @@ class AtlasStepper {
...
   ///
   /// @param state is the provided stepper state (caller keeps thread locality)
-  template <typename propagator_state_t, typename navigator_t>
-  Result<double> step(propagator_state_t& state,
-                      const navigator_t& /*navigator*/) const {
+  /// @param propDir is the direction of propagation
+  /// @param material is the material properties
+  /// @return the result of the step
+  Result<double> step(State& state, Direction propDir,
+                      const IVolumeMaterial* material) const {
```
paulgessinger, line 1155: Is there a good place do document that this is allowed to be `nullptr`?

Core/include/Acts/Propagator/VoidNavigator.hpp
```diff
@@ -56,6 +58,14 @@ class VoidNavigator {
     return nullptr;
   }
 
+  const TrackingVolume* currentVolume(const State& /*state*/) const {
+    return nullptr;
+  }
+
+  const IVolumeMaterial* currentVolumeMaterial(const State& /*state*/) const {
+    return nullptr;
+  }
```
paulgessinger, line 67: My impression was that the steppers stop using these directly. I guess they're still used in the Propagator?

Core/include/Acts/Propagator/Propagator.ipp
andiwand: maybe I was a bit too fast to remove this. the stepper is templated on the propagator so this can only be incompatible if you create a state from a different propagator with a different stepper and then pass this state to the first propagator right? is this something we have to worry about?

Core/src/Propagator/SympyStepper.cpp
andiwand: I realized that this is not implemented in the sympy stepper yet. I will try to follow up with this in another PR

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
andiwand: `transportCovarianceToBound` should only depend on a const reference

Core/include/Acts/Propagator/AtlasStepper.hpp
andiwand: hm I think this is a common problem with our templated interfaces. I guess I can only copy it to all instances

Core/include/Acts/Propagator/VoidNavigator.hpp
andiwand: right I now depend on this in the propagator

paulgessinger (approved):

PR #4040, opened 2025-01-21
feat: Adding cuboid volume stack

ssdetlab:
Adding `CuboidVolumeStack`. The overall logic is similar to `CylinderVolumeStack` -- a set of aligned volumes is formed into a stack, the stack volumes are adjusted in size and the connection between the volumes is insured via resizing/construction of the gap volumes. 

The main difference with `CylinderVolumeStack` is in the fact that cuboid volumes don't really have a "special" direction that can fix the stacking conventions, e.g. specific treatment of `AxisZ` and `AxisR` stacking. For this reason a level of generalization is applied to the stacking direction handling, including handling of arbitrary (in global coordinate system) merging directions.      

## Release Notes

- **New Features**
	- Introduced `CuboidVolumeStack` class for managing stacks of cuboid volumes.
	- Added `VolumeAttachmentStrategy` and `VolumeResizeStrategy` for flexible volume management.
	- Enhanced `CuboidVolumeBounds` with a new constructor for flexible initialization.

- **Improvements**
	- Generalized strategy handling across various volume types.
	- Improved error checking and initialization for volume bounds.
	- Expanded geometry module with new strategy classes.

- **Testing**
	- Added comprehensive unit tests for new volume management features.
	- Enhanced test coverage for volume initialization and error handling.

- **Documentation**
	- Updated documentation generation to support new classes and strategies.


Core/src/Geometry/CuboidVolumeStack.cpp
```diff
@@ -0,0 +1,893 @@
...
+                                     AttachmentStrategy strategy,
+                                     ResizeStrategy resizeStrategy,
+                                     const Logger& logger)
+    : Volume(initialVolume(volumes)),
+      m_dir(direction),
+      m_resizeStrategy(resizeStrategy),
+      m_volumes(volumes) {
+  auto dirIdx = static_cast<std::size_t>(m_dir);
+  m_dirOrth1 = static_cast<AxisDirection>((dirIdx + 1) % 3);
+  m_dirOrth2 = static_cast<AxisDirection>((dirIdx + 2) % 3);
```
paulgessinger, line 99: I see what this is trying to accomplish. I'm not entirely sure, if we should rely on the internal numbering inside `AxisDirection` to perform this rotation. Could this be spelled out with regular control flow like a `switch`? It shouldn't be a lot of cases, should it?

Core/src/Geometry/CuboidVolumeStack.cpp
```diff
@@ -0,0 +1,893 @@
...
+    m_dir = AxisDirection::AxisZ;
+  } else {
+    throw std::invalid_argument("CuboidVolumeStack: Invalid axis direction");
+  }
+
+  auto dirIdx = static_cast<std::size_t>(m_dir);
+  m_dirOrth1 = static_cast<AxisDirection>((dirIdx + 1) % 3);
+  m_dirOrth2 = static_cast<AxisDirection>((dirIdx + 2) % 3);
+
+  initializeOuterVolume(strategy, logger);
```
paulgessinger, line 130: You could put the determination of the direction from the `Vector3` into a (static) function, and call the other constructor directly from this one, something like:

```cpp
CuboidVolumeStack::CuboidVolumeStack(std::vector<Volume*>& volumes,
                                     Vector3 direction,
                                     AttachmentStrategy strategy,
                                     ResizeStrategy resizeStrategy,
                                     const Logger& logger)
    : CuboidVolumeStack(volumes, 
                        directionFromVector(direction), 
                        strategy, resizeStrategy, logger) {}
```

Core/src/Geometry/CuboidVolumeStack.cpp
```diff
@@ -0,0 +1,893 @@
...
+    Volume::update(std::make_shared<CuboidVolumeBounds>(*bounds), std::nullopt,
+                   logger);
+    ACTS_VERBOSE("Transform is now: " << m_transform.matrix());
+    return;
+  }
+
+  ACTS_VERBOSE("Checking volume alignment");
+  checkVolumeAlignment(volumeTuples, logger);
+
+  auto dirIdx = static_cast<std::size_t>(m_dir);
```
paulgessinger, line 195: This currently works out (which might be what @asalzburger originally intended), but I'm a bit worried to always assume this. I'd probably just do this with a `switch` to be sure.

Core/src/Geometry/CuboidVolumeStack.cpp
```diff
@@ -0,0 +1,893 @@
...
+  ACTS_VERBOSE("Sorting by volume " << axisDirectionName(m_dir) << " position");
+  std::ranges::sort(volumeTuples, {}, [&dirIdx](const auto& v) {
```
paulgessinger, line 197:
```suggestion
  std::ranges::sort(volumeTuples, {}, [dirIdx](const auto& v) {
```

Core/src/Geometry/CuboidVolumeStack.cpp
```diff
@@ -0,0 +1,893 @@
...
+  }
+
+  ACTS_DEBUG("*** Volume configuration after final " << axisDirectionName(m_dir)
+                                                     << " sorting:");
+  printVolumeSequence(volumeTuples, logger, Acts::Logging::DEBUG);
+
+  double min = volumeTuples.front().min(m_dir);
+  double max = volumeTuples.back().max(m_dir);
+
+  double mid = (min + max) / 2.0;
```
paulgessinger, line 248: I guess sonar will again suggest using `std::midpoint`.

Core/src/Geometry/CuboidVolumeStack.cpp
```diff
@@ -0,0 +1,893 @@
...
+  double hl = (max - min) / 2.0;
+
+  Translation3 translation(Vector3::Unit(dirIdx) * mid);
+  m_transform = m_groupTransform * translation;
+  std::array<double, 3> bounds;
+  bounds.at(dirIdx) = hl;
+  bounds.at(static_cast<std::size_t>(m_dirOrth1)) = hl1;
+  bounds.at(static_cast<std::size_t>(m_dirOrth2)) = hl2;
+
+  Volume::update(std::make_shared<CuboidVolumeBounds>(bounds), std::nullopt,
```
paulgessinger, line 258: Could you use the `CuboidVolumeBounds` constructor that uses named arguments? We're trying to phase out directly constructing from the "bound values" vectors / arrays, since this is tied to the internal representation.

Core/src/Geometry/CuboidVolumeStack.cpp
```diff
@@ -0,0 +1,893 @@
...
+  }
+}
+
+std::vector<CuboidVolumeStack::VolumeTuple>
+CuboidVolumeStack::checkOverlapAndAttach(
+    std::vector<VolumeTuple>& volumes,
+    CuboidVolumeStack::AttachmentStrategy strategy, const Logger& logger) {
+  // Preconditions: volumes are sorted along stacking direction
+  auto dirIdx = static_cast<std::size_t>(m_dir);
+  auto dirBoundIdx = static_cast<CuboidVolumeBounds::BoundValues>(m_dir);
```
paulgessinger, line 300: This should not be assumed. Access to these numbers should not rely on the internal integer value of these enums. The fact that they're the same should be considered *coincidence*.

Core/src/Geometry/CuboidVolumeStack.cpp
```diff
@@ -0,0 +1,893 @@
...
+          ACTS_VERBOSE("  - Gap width: " << gapWidth);
+          ACTS_VERBOSE("  - New bounds for first volume: ["
+                       << (aMidNew - aHlNew) << " <- " << aMidNew << " -> "
+                       << (aMidNew + aHlNew) << "]");
+
+          assert(std::abs(a.min(m_dir) - (aMidNew - aHlNew)) < 1e-9 &&
+                 "Volume shrunk");
+          assert(aHlNew >= a.halfLength(m_dir) && "Volume shrunk");
+
+          Translation3 translationA(Vector3::Unit(dirIdx) * aMidNew);
```
paulgessinger, line 378: This should only be done if `dirIdx` is ensured to be `0, 1, 2` explicitly. You can do this, if at the entry boundary this index is derived in a safe way.

Core/src/Geometry/CuboidVolumeStack.cpp
```diff
@@ -0,0 +1,893 @@
...
+    logger().log(lvl, ss.str());
+  }
+}
+
+void CuboidVolumeStack::checkVolumeAlignment(
+    const std::vector<VolumeTuple>& volumes, const Logger& logger) const {
+  std::size_t n = 0;
+  auto dirIdx = static_cast<std::size_t>(m_dir);
+  auto dirOrth1Idx = static_cast<std::size_t>(m_dirOrth1);
+  auto dirOrth2Idx = static_cast<std::size_t>(m_dirOrth2);
```
paulgessinger, line 476: Same comment applies here. I think the best approach is to have a (hidden) function that converts between these in a safe way with a `switch`, or maybe a member method that returns the orthogonal axes as native Eigen axes (`0, 1, 2`).

Core/src/Geometry/CuboidVolumeStack.cpp
```diff
@@ -0,0 +1,893 @@
...
+}
+
+std::pair<double, double> CuboidVolumeStack::synchronizeBounds(
+    std::vector<VolumeTuple>& volumes, const Logger& logger) {
+  auto boundDirOrth1 = static_cast<CuboidVolumeBounds::BoundValues>(m_dirOrth1);
+  auto boundDirOrth2 = static_cast<CuboidVolumeBounds::BoundValues>(m_dirOrth2);
+
+  const double maxHl1 =
+      std::max_element(volumes.begin(), volumes.end(),
+                       [&boundDirOrth1](const auto& a, const auto& b) {
```
paulgessinger, line 524:
```suggestion
                       [boundDirOrth1](const auto& a, const auto& b) {
```

Not a huge difference, but the `&` is not necessary here, because it's not mutated, and the size of the integer being captured is the same as the pointer.

Core/src/Geometry/CuboidVolumeStack.cpp
```diff
@@ -0,0 +1,893 @@
...
+
+  ACTS_VERBOSE("*** Initial volume configuration:");
+  printVolumeSequence(volumeTuples, logger, Acts::Logging::DEBUG);
+  for (const auto& dir : {m_dirOrth1, m_dirOrth2}) {
+    if (!same(newVolume.min(dir), oldVolume.min(dir)) ||
+        !same(newVolume.max(dir), oldVolume.max(dir))) {
+      ACTS_VERBOSE("Resize all volumes to new " << axisDirectionName(dir)
+                                                << " bounds");
+      for (auto& volume : volumeTuples) {
+        volume.set({{static_cast<CuboidVolumeBounds::BoundValues>(dir),
```
paulgessinger, line 661: Conversion helper could be used here.

Core/src/Geometry/CuboidVolumeStack.cpp
```diff
@@ -0,0 +1,893 @@
...
+  }
+
+  if (same(newVolume.halfLength(m_dir), oldVolume.halfLength(m_dir))) {
+    ACTS_VERBOSE("Halflength " << axisDirectionName(m_dir) << "is the same, no "
+                               << axisDirectionName(m_dir) << "resize needed");
+  } else {
+    auto dirIdx = static_cast<std::size_t>(m_dir);
+    auto dirOrth1Idx = static_cast<std::size_t>(m_dirOrth1);
+    auto dirOrth2Idx = static_cast<std::size_t>(m_dirOrth2);
+    auto boundDirIdx = static_cast<CuboidVolumeBounds::BoundValues>(m_dir);
```
paulgessinger, line 681: And here.

Core/include/Acts/Geometry/detail/VolumeAttachmentStrategy.hpp
paulgessinger, line 1: Since these are (or will be) part of the public API, I think they shouldn't be in a `detail` folder.

Examples/Python/src/Geometry.cpp
```diff
@@ -294,16 +296,15 @@ void addGeometry(Context& ctx) {
...
+        .value("Second", VolumeAttachmentStrategy::Second)
+        .value("Midpoint", VolumeAttachmentStrategy::Midpoint);
 
-    py::enum_<CylinderVolumeStack::ResizeStrategy>(cylStack, "ResizeStrategy")
-        .value("Gap", CylinderVolumeStack::ResizeStrategy::Gap)
-        .value("Expand", CylinderVolumeStack::ResizeStrategy::Expand);
+    py::enum_<VolumeResizeStrategy>(cylStack, "ResizeStrategy")
+        .value("Gap", VolumeResizeStrategy::Gap)
+        .value("Expand", VolumeResizeStrategy::Expand);
   }
```
paulgessinger, line 308: I think they should also not be nested in the `CylinderVolumeStack` class on the python side, but rather be top-level members of the `acts` package.

paulgessinger (commented):
- `VolumeAttachmentStrategy` and `VolumeResizeStrategy`: good choice of names!
- I'd move them outside of the `detail` folder.
- Python bindings for them should move out of the `CylinderVolumeStack`

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [9f58ce1a589cd3fd665c5d72310e4884dfc96e7a]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/49818037) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/9976464)!

PR #4026, opened 2025-01-13
build: CMake find directives, policy grouping

paulgessinger:
This changes:

- The way Python is found through `CMP0094`, which sets `Python_STRATEGY=LOCATION`:
  > LOCATION: Stops lookup as soon as a version satisfying version constraints is founded. This is the default if policy [CMP0094](https://cmake.org/cmake/help/latest/policy/CMP0094.html#policy:CMP0094) is set to NEW.
- Sets `CMAKE_FIND_FRAMEWORK=LAST` which instructs CMake to consider macOS *frameworks* after all other sources, which is e.g. relevant for Python (see above)
- Groups all policy decisions into one block in the main `CMakeLists.txt`.

--- END COMMIT MESSAGE ---

Any further description goes here, @-mentions are ok here!

- Use a *conventional commits* prefix: [quick summary](https://www.conventionalcommits.org/en/v1.0.0/#summary)
  - We mostly use `feat`, `fix`, `refactor`, `docs`, `chore` and `build` types.
- A milestone will be assigned by one of the maintainers

## Summary by CodeRabbit

- **Configuration Updates**
	- Enhanced CMake project configuration with new policy settings
	- Updated framework discovery settings for continuous integration builds

PR #4042, opened 2025-01-21
build: find Python after DD4hep to find the correct one

jmcarcell:
Find Python after DD4hep to make sure the correct one is found.

In cases when there are multiple versions around and the one from the system is greater than the one we want to use `find_package(Python)` will find that one and then DD4hep will complain that it is different from the one used to build ROOT and fail the CMake configuring. This is the case when building the Key4hep stack on Ubuntu 24.04 with Spack, where Python from the system is 3.12 but the one provided in the stack is 3.11 since not everything builds (at least not easily) with 3.12.

## Summary by CodeRabbit

- **Chores**
	- Reorganized CMake configuration by relocating the Python package finding directive.


paulgessinger:
@jmcarcell So I actually put in this specific `find_package` call before the DD4hep `find_package` call to make sure we don't trigger DD4hep's python lookup logic, because that was causing issues for us.

IIRC the problem I had is that without doing this manually first, the python finding in DD4hep will ignore `Python_EXECUTABLE` overrides, and just pick a random python that it happens to see. I haven't run into this recently, so it's possible this workaround is not needed anymore.

I changed the python finding policy recently in https://github.com/acts-project/acts/pull/4026, but I haven't tested yet how this interacts with DD4hep. In principle, with #4026, I set the policy for ACTS so that CMake / `FindPython` will pick the first python that it finds that fulfills the version requirements. I think this is the right thing to do, as that means as long as e.g. the target python is in your prefix path or path, it will pick that one as long as it's recent enough. In the spack build, it will find that one.

I'm generally ok with the change as long as the CI and later on the Spack CI run with this, although I'd probably remove the extra `find_package` python call altogethe.

paulgessinger:
Actually, on second thought: 
I do think we might have to keep our `find_package(Python)` before DD4hep to make sure our find policy ([CMP0094](https://cmake.org/cmake/help/latest/policy/CMP0094.html#policy:CMP0094)) is applied instead of DD4hep's. 

The issue that I see is that DD4hep's config just includes the `DD4hepBuild.cmake`, which calls `find_package` itself. At that point, I'm not entirely sure whose find policy is effective, and it might be DD4hep's. In that case, we'll need to pre-resolve the correct python and then have DD4hep pick that up.

PR #4055, opened 2025-01-28
refactor: 

pbutti:
This PR proposes the explicit usage of Isometry instead of Affine in the Acts Algebra Definitions. 
This should imply an optimized inversion of the transforms (transpose of the linear part instead of general inversion). Additionally it should make it clear that current transformations in ACTS are isometries (no scaling, no sheers). 

- I just tested that ODD runs
- I disabled beveled disc bounds for the moment. That's the only place where the scaling is used
- I rewrote the initialization of transforms using Eigen::Isometry3d around rotations. This doesn't check that the rotation matrix is orthogonal and with det=1 (but we weren't doing it anyway). 

I haven't seen sensible improvement in processing time (just checked the full ODD processing time on 1k events) but I still propose the change as it makes clear we are using isometric transformations.

### DIFF
--- a/Core/src/Seeding/EstimateTrackParamsFromSeed.cpp
+++ b/Core/src/Seeding/EstimateTrackParamsFromSeed.cpp
@@ -30,11 +30,13 @@ Acts::FreeVector Acts::estimateTrackParamsFromSeed(const Vector3& sp0,
   rotation.col(0) = newXAxis;
   rotation.col(1) = newYAxis;
   rotation.col(2) = newZAxis;
+
   // The center of the new frame is at the bottom space point
   Translation3 trans(sp0);
-  // The transform which constructs the new frame
-  Transform3 transform(trans * rotation);
 
+  // The transform which constructs the new frame
+  Transform3 transform(trans * Eigen::Isometry3d(rotation));
+  
   // The coordinate of the middle and top space point in the new frame
   Vector3 local1 = transform.inverse() * sp1;
   Vector3 local2 = transform.inverse() * sp2;
--- a/Plugins/Geant4/src/Geant4Converters.cpp
+++ b/Plugins/Geant4/src/Geant4Converters.cpp
@@ -252,9 +252,11 @@ Acts::Geant4ShapeConverter::planarBounds(const G4VSolid& g4Solid) {
 namespace {
 Acts::Transform3 axesOriented(const Acts::Transform3& toGlobalOriginal,
                               const std::array<int, 2u>& axes) {
-  auto originalRotation = toGlobalOriginal.rotation();
-  auto colX = originalRotation.col(std::abs(axes[0u]));
-  auto colY = originalRotation.col(std::abs(axes[1u]));
+  //auto originalRotation = toGlobalOriginal.rotation();
+  auto originalRotation = toGlobalOriginal.linear();
+  //make a copy
+  Acts::Vector3 colX = originalRotation.col(std::abs(axes[0u])).eval();
+  Acts::Vector3 colY = originalRotation.col(std::abs(axes[1u])).eval();
   colX *= std::copysign(1, axes[0u]);
   colY *= std::copysign(1, axes[1u]);
   Acts::Vector3 colZ = colX.cross(colY);
--- a/Core/include/Acts/Definitions/Algebra.hpp
+++ b/Core/include/Acts/Definitions/Algebra.hpp
@@ -87,7 +87,7 @@ using AngleAxis3 = Eigen::AngleAxis<double>;
 // - 2d affine compact stored as 2x3 matrix
 // - 3d affine stored as 4x4 matrix
 using Transform2 = Eigen::Transform<double, 2, Eigen::AffineCompact>;
-using Transform3 = Eigen::Transform<double, 3, Eigen::Affine>;
+using Transform3 = Eigen::Transform<double, 3, Eigen::Isometry>;
 
 constexpr double s_transformEquivalentTolerance = 1e-9;
 
--- a/Core/src/Geometry/CylinderVolumeBounds.cpp
+++ b/Core/src/Geometry/CylinderVolumeBounds.cpp
@@ -97,18 +97,20 @@ std::vector<OrientedSurface> CylinderVolumeBounds::orientedSurfaces(
   double bevelMaxZ = get(eBevelMaxZ);
   Transform3 transMinZ, transMaxZ;
   if (bevelMinZ != 0.) {
-    double sy = 1 - 1 / std::cos(bevelMinZ);
-    transMinZ = transform * vMinZ *
-                Eigen::AngleAxisd(-bevelMinZ, Eigen::Vector3d(1., 0., 0.)) *
-                Eigen::Scaling(1., 1. + sy, 1.);
+    //double sy = 1 - 1 / std::cos(bevelMinZ);
+    //transMinZ = transform * vMinZ *
+    //            Eigen::AngleAxisd(-bevelMinZ, Eigen::Vector3d(1., 0., 0.)) *
+    //            Eigen::Scaling(1., 1. + sy, 1.);
+    throw std::invalid_argument("Bevel Min Z != 0 is not supported");
   } else {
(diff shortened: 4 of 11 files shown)


pbutti:
Tagging: @paulgessinger @andiwand

andiwand:
I think this is a breaking change?

paulgessinger:
Output changes only I think. Will you look into those @pbutti?

andiwand:
I am surprised by the physmon change. I think I have seen a similar change in a different PR. Potentially it goes away with merging main? Or this is just a fluctuation with a single bin entry

paulgessinger:
@andiwand let's find out.

Core/src/Seeding/EstimateTrackParamsFromSeed.cpp
```diff
@@ -30,10 +30,12 @@ Acts::FreeVector Acts::estimateTrackParamsFromSeed(const Vector3& sp0,
   rotation.col(0) = newXAxis;
   rotation.col(1) = newYAxis;
   rotation.col(2) = newZAxis;
+
   // The center of the new frame is at the bottom space point
   Translation3 trans(sp0);
+
   // The transform which constructs the new frame
-  Transform3 transform(trans * rotation);
+  Transform3 transform(trans * Eigen::Isometry3d(rotation));
```
paulgessinger, line 38: This doesn't work without the explicit `Isometry3d`?

Plugins/Geant4/src/Geant4Converters.cpp
```diff
@@ -319,9 +319,11 @@ Acts::Geant4ShapeConverter::planarBounds(const G4VSolid& g4Solid) {
 namespace {
 Acts::Transform3 axesOriented(const Acts::Transform3& toGlobalOriginal,
                               const std::array<int, 2u>& axes) {
-  auto originalRotation = toGlobalOriginal.rotation();
-  auto colX = originalRotation.col(std::abs(axes[0u]));
-  auto colY = originalRotation.col(std::abs(axes[1u]));
+  // auto originalRotation = toGlobalOriginal.rotation();
```
paulgessinger, line 322:
```suggestion
```

Also, this shouldn't matter with the change, no?

Plugins/GeoModel/src/GeoModelDetectorObjectFactory.cpp
```diff
@@ -162,8 +162,18 @@ void Acts::GeoModelDetectorObjectFactory::convertFpv(
   std::vector<GeoModelSensitiveSurface> sensitives;
 
   for (const auto &surface : surfaces) {
-    const Transform3 &transform =
-        fpv->getAbsoluteTransform() * surface.transform;
+    const Transform3 &absTrans =
+        Eigen::Isometry3d(fpv->getAbsoluteTransform().matrix());
```
paulgessinger, line 166: This doesn't work:
```suggestion
    Transform3 absTrans{fpv->getAbsoluteTransform().matrix()};
```

Also, maybe you can remove the const refs? I think this is safe due to lifetime extension, but it could just be an lvalue honestly.

Plugins/GeoModel/src/GeoModelDetectorObjectFactory.cpp
```diff
@@ -162,8 +162,18 @@ void Acts::GeoModelDetectorObjectFactory::convertFpv(
...
+    const Transform3 &surfTrf = Eigen::Isometry3d(surface.transform.matrix());
```
paulgessinger, line 167:
```suggestion
    Transform &surfTrf = Eigen::Isometry3d(surface.transform.matrix());
```

Plugins/GeoModel/src/GeoModelDetectorObjectFactory.cpp
```diff
@@ -162,8 +162,18 @@ void Acts::GeoModelDetectorObjectFactory::convertFpv(
...
+
+    if (!isIsometry(absTrans) || !isIsometry(surfTrf)) {
+      throw std::runtime_error(
+          "GeoModelDetectorObjectFactory::ERROR Transformation is not a valid "
+          "isometry!");
+    }
+
+    const Transform3 &transform = absTrans * surfTrf;
```
paulgessinger, line 175:
```suggestion
    Transform3 transform = absTrans * surfTrf;
```

Plugins/GeoModel/src/GeoModelToDetectorVolume.cpp
```diff
@@ -126,12 +126,25 @@ Volume convertVolume(const Transform3& trf, const GeoShape& shape) {
         dynamic_cast<const GeoShapeShift*>(&shape);
     const GeoShape* shapeOp = shiftShape->getOp();
     newTrf = trf * shiftShape->getX();
-    return convertVolume(newTrf, *shapeOp);
+    // This assumes, without checking, that newTrf is an isometry
+    Transform3 newtrf = Eigen::Isometry3d(newTrf.matrix());
```
paulgessinger, line 130: Also here, this won't work?

```suggestion
    Transform3 newtrf{newTrf.matrix()};
```

paulgessinger (commented):
- Did we understand where the output changes come from? Does physmon look ok?
- Can you check if removing the redundant type constructions could be removed?
- Can you avoid using const refs when a temporary is created?

Core/src/Seeding/EstimateTrackParamsFromSeed.cpp
pbutti: Without the explicit call to Eigen::Isometry3d(rotation) the result of `trans * rotation` will be interpreted as an affine transformation because a priori it cannot say that the rotation doesn't contain scale and sheer. 
Maybe I could wrap Eigen::Isometry3d(trans*rotation) if you prefer.

Plugins/GeoModel/src/GeoModelDetectorObjectFactory.cpp
pbutti: I can remove the const ref, that's fine. 
But I think I still need to wrap fpv->getAbsoluteTransform().matrix() inside Eigen::Isometry3d. I can check.

Plugins/Geant4/src/Geant4Converters.cpp
pbutti: Shouldn't matter but I kept linear. Removed the comment.

Plugins/GeoModel/src/GeoModelDetectorObjectFactory.cpp
pbutti: Changed this block in 
```
Transform3 absTrans{fpv->getAbsoluteTransform().matrix()};
Transform3 surfTrf{surface.transform.matrix()};
```

Plugins/GeoModel/src/GeoModelDetectorObjectFactory.cpp
pbutti: Your suggestion should work. I've changed as you suggested, see next comment.

Plugins/GeoModel/src/GeoModelToDetectorVolume.cpp
pbutti: Your suggestion works. Updated the code accordingly.

pbutti:
Did we understand where the output changes come from? Does physmon look ok?

>>> The Physmon changes were due to a single bin in the resd0vsEta in the GSF. Actually before such bin was empty and now it's filled and that what caused the mismatch in the plot. Looking at latest monitoring, seems more healthy now. 

Can you check if removing the redundant type constructions could be removed?
>>> I checked and I could remove it in most of the cases apart in the EstimateTrackParameters, where I need to say that that rotation is an isometry. I don't find it too bad to place it there if you don't mind. 


Can you avoid using const refs when a temporary is created?
>>> I applied all your suggestions and removed the const refs where you suggested. Are there other places where you want me to act on?

pbutti:
@paulgessinger @andiwand 
Anything missing for this one still?

andiwand:
closing due to inactivity

PR #4061, opened 2025-01-30
refactor!: Merge curvilinear track params into bound track params

andiwand:
After our last dev meeting we decided that the curvilinear track params are really just bound track params which are constructed in a special way. This PR implements the suggested change and introduces a factory function on the bound track params to construct curvilinear params.

This allows to simplify the propagator interface in the future. The plan is to only accept bound track params as an input and produce bound track params as an output.

--- END COMMIT MESSAGE ---

@osbornjd @paulgessinger @asalzburger 

@benjaminhuth this also touches the multi stepper params


acts-project-service:
## :red_circle: Athena integration test results [f9d52362e4fed552b6c62160d6d35bb6a93496ff]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/52186623) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/10936277)!

Core/include/Acts/EventData/GenericBoundTrackParameters.hpp
```diff
@@ -41,6 +42,76 @@ class GenericBoundTrackParameters {
...
+
+  /// Construct from four-position, angles, and qOverP.
+  ///
+  /// @param pos4 Track position/time four-vector
+  /// @param phi Transverse track direction angle
+  /// @param theta Longitudinal track direction angle
+  /// @param qOverP Charge over momentum
+  /// @param cov Curvilinear bound parameters covariance matrix
+  /// @param particleHypothesis Particle hypothesis
+  static GenericBoundTrackParameters makeCurvilinear(
```
paulgessinger, line 103:
```suggestion
  static GenericBoundTrackParameters createCurvilinear(
```

Also should this be a separate factory, or do we just require the user to manually call `makeDirectionFromPhiTheta`?

Core/include/Acts/EventData/MultiComponentTrackParameters.hpp
```diff
@@ -67,11 +66,55 @@ class MultiComponentBoundTrackParameters {
   }
 
  public:
+  using ConstructionTuple = std::tuple<double, Acts::Vector4, Acts::Vector3,
```
paulgessinger, line 69: Struct with named members? Can still use structured bindings.

Core/include/Acts/EventData/MultiComponentTrackParameters.hpp
```diff
@@ -67,11 +66,55 @@ class MultiComponentBoundTrackParameters {
...
+                                       double, CovarianceMatrix>;
+
+  /// We need this helper function in order to construct the base class properly
+  static MultiComponentBoundTrackParameters makeCurvilinear(
+      const std::vector<ConstructionTuple>& curvi,
+      ParticleHypothesis particleHypothesis) {
+    // TODO where to get a geometry context here
+    Acts::GeometryContext gctx{};
```
paulgessinger, line 77: This is only needed for the interface with the `planeSurface` that comes out of `CurvilinearSurface` right? I've had this issue a couple of times in the past, and the conclusion of various discussions was to still pipe it through the call-chain, even if you can prove locally that you can never be passing this to a surface that has a detector element attached. We've also discussed the possibility of just making it a runtime error if you pass an empty context and there is a detector element behind the surface.

In absence of this, I think this should be part of the function arguments.

Core/include/Acts/EventData/MultiComponentTrackParameters.hpp
```diff
@@ -67,11 +66,55 @@ class MultiComponentBoundTrackParameters {
   }
 
  public:
+  using ConstructionTuple = std::tuple<double, Acts::Vector4, Acts::Vector3,
+                                       double, CovarianceMatrix>;
+
+  /// We need this helper function in order to construct the base class properly
+  static MultiComponentBoundTrackParameters makeCurvilinear(
+      const std::vector<ConstructionTuple>& curvi,
```
paulgessinger, line 74:
```suggestion
      const std::vector<ConstructionTuple>& states,
```
?

Core/include/Acts/EventData/MultiComponentTrackParameters.hpp
```diff
@@ -67,11 +66,55 @@ class MultiComponentBoundTrackParameters {
...
+
+    // Construct and average surface
+    Acts::Vector3 avgPos = Acts::Vector3::Zero();
+    Acts::Vector3 avgDir = Acts::Vector3::Zero();
+    for (const auto& [w, pos4, dir, qop, cov] : curvi) {
+      avgPos += w * pos4.template segment<3>(0);
+      avgDir += w * dir;
+    }
+
+    auto s = CurvilinearSurface(avgPos, avgDir).planeSurface();
```
paulgessinger, line 87: Pedantically, I'd prefer the type here instead of `auto`. It's not clear what comes out of this.

Core/include/Acts/EventData/TrackParameters.hpp
paulgessinger, line 1: Will we still be able to tell if a bound parameter object is a curvilinar parameter set?

Core/include/Acts/EventData/GenericBoundTrackParameters.hpp
```diff
@@ -41,6 +42,76 @@ class GenericBoundTrackParameters {
...
+  /// @param dir Track direction three-vector; normalization is ignored.
+  /// @param qOverP Charge over momentum
+  /// @param cov Curvilinear bound parameters covariance matrix
+  /// @param particleHypothesis Particle hypothesis
+  static GenericBoundTrackParameters makeCurvilinear(
+      const Vector4& pos4, const Vector3& dir, double qOverP,
+      std::optional<CovarianceMatrix> cov,
+      ParticleHypothesis particleHypothesis) {
+    return GenericBoundTrackParameters(
+        CurvilinearSurface(pos4.segment<3>(ePos0), dir).surface(),
```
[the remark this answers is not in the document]
andiwand: yes exactly. just moved it

Core/include/Acts/EventData/GenericBoundTrackParameters.hpp
andiwand: I think we can drop it. I just moved it from the previous implementation

Core/include/Acts/EventData/MultiComponentTrackParameters.hpp
andiwand: I moved this up from `MultiComponentCurvilinearTrackParameters`. I believe the multi stepper infrastructure has quite some `using` here and there. In that sense it is consistent right now. Potentially something we want to follow up on in general.

Core/include/Acts/EventData/MultiComponentTrackParameters.hpp
andiwand: I agree - this was there before my changes. I can check how easy it is to change.

Core/include/Acts/EventData/TrackParameters.hpp
andiwand: No this is not possible. TBH I cannot think of any use of this information. A curvilinear parameter is a bound parameter and it was like this before already just in a weaker form.

Core/include/Acts/EventData/GenericBoundTrackParameters.hpp
andiwand: actually I went back to this because providing `phi, theta` will not need any conversion for the track params. without that unit tests fail

Core/include/Acts/EventData/MultiComponentTrackParameters.hpp
paulgessinger: ok

paulgessinger (approved):

PR #4076, opened 2025-02-10
refactor: Delete non-compile tests

paulgessinger:
This seemed like a good idea at the time, but the complexity seems not to be worth it.

--- END COMMIT MESSAGE ---


asalzburger (approved):
Good from my side.

PR #4078, opened 2025-02-10
feat: Remove std::isinf check on configuration parameter and a sqrt computation

pbutti:
This PR proposes to
- remove the isinf check on a configuration parameter which is fixed. It has been shown to waste time  when repeatedly checked on a fixed parameter. Propose to just do a simpler >0 check instead
- get rid of a sqrt computation if we move sinTheta evaluation (only used if detailedMeasurements are asked for) in the proper constexpr check. 

I don't expect much of a gain, but better than nothing... 

## Summary by CodeRabbit

- **Refactor**
	- Enhanced candidate filtering logic by simplifying measurement condition handling.
	- Improved threshold evaluations for scattering parameters to deliver more consistent and reliable outcomes.


Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -771,7 +773,7 @@ SeedFinder<external_spacepoint_t, grid_t, platform_t>::filterCandidates(
       // convert p(T) to p scaling by sin^2(theta) AND scale by 1/sin^4(theta)
       // from rad to deltaCotTheta
       float p2scatterSigma = iHelixDiameter2 * sigmaSquaredPtDependent;
-      if (!std::isinf(m_config.maxPtScattering)) {
```
paulgessinger, line 774: Could we just check this in the `SeedFinder` constructor? There's a number of `isnan` checks there already?

Maybe `inf` is used as a sentinel value to switch this cut off, but we could maybe also use an `std::optional` to signal that the cut is unset.

andiwand: did you try using `std::isfinite`?

pbutti: I did not try, but in principle `std::isfinite` will check against `NaN` and not infinity, so I would naively say that might be the same effect.

andiwand: I cannot reproduce this in a micro benchmark https://quick-bench.com/q/-bf3LqAZSlpdQ56i958SVbKBCsI

I would also think that the branch predictor would get this always right since the config never changes after construction. Of course the CPU would still have to evaluate if the branch was correctly taken but compared to what else we do in the seeding I have a hard time to imagine why this would have an impact.

Potentially the profiler accumulated different costs on this line or just "randomly" associated it there.

What is the difference in runtime you measured with this change?

pbutti: Probably I might be just fooling myself and perhaps the profiler accumulated different costs as you suggests.  
The expected improvement in runtime is negligible and probably within noise. I just saw it popping up in the vtune (see screenshot) with variable percentage (0.1-0.3%) depending on config and thought might as well change it with a simpler check. 

I agree that I am surprised too to see it popping up: being a fixed value the processor shouldn't do mispredictions. 
Agree that it's super minor compared to other things in this code. 

Screenshot 2025-02-18 at 21 24 21

paulgessinger:
@andiwand @pbutti what's the plan for this?

andiwand:
I think we should take another look as soon as we have the newly ported seeding. Other than that in my experience these impacts where rather part of branch prediction

andiwand:
I removed this check with
- https://github.com/acts-project/acts/pull/4533
- https://github.com/acts-project/acts/pull/4535

PR #4083, opened 2025-02-11
build: Add cmake helper function for header compilation

paulgessinger:
This PR adds a CMake function that will

- Take all headers given explicitly either by glob or explicitly
- The first argument is a sort of scope target
- The function creates a new target with the suffix `_HEADERS`
- It loops over all input headers, creates a `.cpp` file which includes that header, and adds it to the `_HEADERS` target for compilation
- The target is excluded from the "default" build to not slow it down.

This will help static analysis tools like clang-tidy, because the `compile_commands.json` will be populated.

This PR also removes a leftover `ACTS_BUILD_NONCOMPILE_TESTS` that was unused.

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **New Features**
  - Integrated an automated header compilation process into various libraries, streamlining the build workflow.
  - Enhanced public interfaces to support improved covariance handling and advanced binning data options for more robust data processing.

- **Bug Fixes**
  - Addressed issues related to header file discovery and inclusion for multiple libraries.

- **Refactor / Chores**
  - Removed deprecated simulated hit conversion functionality and updated linkage configurations, contributing to a cleaner and more efficient framework.
  - Updated CMake configurations to improve header file management and library dependencies.
  - Added a new CMake option for generating targets to compile header files, enhancing configuration capabilities.
  - Removed the `ACTS_BUILD_NONCOMPILE_TESTS` option from the configuration settings.


Core/CMakeLists.txt
```diff
@@ -111,3 +111,9 @@ add_subdirectory(src/Utilities)
 add_subdirectory(src/Vertexing)
 add_subdirectory(src/Visualization)
 add_subdirectory(src/AmbiguityResolution)
+
+# Help with compiler flags discovery
+include(ActsFunctions)
```
andiwand, line 116: out of curiosity: why is this necessary?

cmake/ActsFunctions.cmake
andiwand, line 1: this is only included in `Core/CMakeLists.txt`?

andiwand (dismissed):
👍

Core/CMakeLists.txt
paulgessinger: I think it needs to be included somewhere once for the function to be available. I guess this location works because it's evaluated early on. I'll move it out to the top level `CMakeLists.txt` however.

paulgessinger:
Added the ability to turn this off via an option.

CMakeLists.txt
```diff
@@ -114,7 +114,6 @@ else()
   set(_default_examples_unit_tests OFF)
 endif()
 option(ACTS_BUILD_EXAMPLES_UNITTESTS "Build unit tests" ${_default_examples_unit_tests}) # default: ACTS_BUILD_UNITTESTS AND ACTS_BUILD_EXAMPLES
-option(ACTS_BUILD_NONCOMPILE_TESTS "Build tests that check build failure invariants" OFF)
```
benjaminhuth, line 117: unrelated?

cmake/ActsFunctions.cmake
```diff
@@ -0,0 +1,78 @@
+# This function adds a helper target called ${target}_HEADERS which has
+# generated source files that include each hedaer that was given one-by-one.
+# The generated target links against the one given in `target` and therefore
+# has access to all includes.
+# The generated target is not included in the default build, and is only meant
+# to help tools like clangd discover compiler flags.
+function(acts_compile_headers target)
+    set(options "")
+    set(oneValueArgs GLOB)
```
benjaminhuth, line 9: Maybe I missed the section where `GLOB` was not used as an argument. But if in fact `GLOB` is always used, couldn't we get rid of quite a bit of argument parsing code here?

paulgessinger: The argument parsing is largely done by `cmake_parse_args`, and I'd keep using this to have the keyword `GLOB` in front of the actual glob pattern. The other argument to my mind doesn't add much complexity, but I can remove it and always require the glob if you prefer.

CMakeLists.txt
paulgessinger: Yeah that's a leftover from a previous PR. I can break it out if you prefer, but the option is already unused now so it seemed like an easy fix on the side.

cmake/ActsFunctions.cmake
benjaminhuth: Hmm so if I would write the code, I would probably strip `GLOB` and the other argument away, but I'm good also with just removing the second argument. If needed, this can be re-added in the futuer

CMakeLists.txt
benjaminhuth: Hmm just mention it in the commit message then maybe.

cmake/ActsFunctions.cmake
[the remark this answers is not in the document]
paulgessinger: We don't have it in other CMake files so far.

cmake/ActsFunctions.cmake
paulgessinger: removed the other arguments and kept only GLOB.

paulgessinger:
Done, can you have another look @benjaminhuth?

benjaminhuth (approved):
LGTM

PR #4090, opened 2025-02-14
feat: Allow compiling examples without ROOT

benjaminhuth:
First step towards `pip install acts`?

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **New Features**
  - Introduced a configurable build option that enables advanced ROOT I/O components.
  - Enhanced Python integration with new input and output capabilities when ROOT I/O is activated.

- **Refactor**
  - Updated module inclusion to conditionally incorporate ROOT functionalities, modernizing input handling and streamlining dependencies.

- **Documentation**
  - Revised build instructions to document the new option for enabling ROOT I/O–based modules.


Examples/Python/src/Material.cpp
```diff
@@ -25,12 +25,15 @@
 #include "Acts/Plugins/Python/Utilities.hpp"
 #include "Acts/Utilities/Logger.hpp"
 #include "ActsExamples/Framework/ProcessCode.hpp"
-#include "ActsExamples/Io/Root/RootMaterialDecorator.hpp"
 #include "ActsExamples/MaterialMapping/CoreMaterialMapping.hpp"
 #include "ActsExamples/MaterialMapping/MappingMaterialDecorator.hpp"
 #include "ActsExamples/MaterialMapping/MaterialMapping.hpp"
 #include "ActsExamples/MaterialMapping/MaterialValidation.hpp"
 
+#ifdef ACTS_EXAMPLES_WITH_ROOT
```
paulgessinger, line 33: I think we should try to find a way to do this without a preprocessor define. Can we handle this in CMake somehow?
This would probably need to move to the ROOT IO folder.

benjaminhuth: Yeah probably we should put this to `RootInput.cpp` and avoid the preprocessor `ifdef`.

CMakePresets.json
```diff
@@ -17,6 +17,7 @@
                 "ACTS_BUILD_PLUGIN_LEGACY": "ON",
                 "ACTS_BUILD_EXAMPLES_DD4HEP": "ON",
                 "ACTS_BUILD_EXAMPLES_PYTHON_BINDINGS": "ON",
+                "ACTS_BUILD_EXAMPLES_ROOT": "ON",
```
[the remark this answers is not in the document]
benjaminhuth: The convention is, that in the CmakeLists.txt, all options are off. Therefore, this aligns with our design.

paulgessinger:
@benjaminhuth I will be adding a hard HepMC3 dependency in the Examples. We'll have to think about how that would work in this context.

Examples/Io/Root/CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ target_link_libraries(
         Threads::Threads
     PRIVATE ROOT::Core ROOT::Hist ROOT::Tree
 )
+target_compile_definitions(ActsExamplesIoRoot PUBLIC ACTS_EXAMPLES_WITH_ROOT)
```
paulgessinger, line 49: Do we need this? I don't see where it's used, and usually public defines like this can be a bit painful.

Examples/Python/src/RootInput.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include <pybind11/stl.h>
+
+namespace py = pybind11;
+using namespace pybind11::literals;
+
+using namespace ActsExamples;
+
+namespace Acts::Python {
+
+void addRootInput(Context& ctx) {
```
paulgessinger, line 29: I'm wondering if we should put these into a separate sub-package `root` now. I think @asalzburger has been thinking about this recently.

CMakeLists.txt
```diff
@@ -103,6 +103,7 @@ option(ACTS_BUILD_EXAMPLES_HEPMC3 "Build HepMC3-based code in the examples" OFF)
 option(ACTS_BUILD_EXAMPLES_HASHING "Build Hashing-based code in the examples" OFF)
 option(ACTS_BUILD_EXAMPLES_PYTHIA8 "Build Pythia8-based code in the examples" OFF)
 option(ACTS_BUILD_EXAMPLES_PYTHON_BINDINGS "Build python bindings for the examples" OFF)
+option(ACTS_BUILD_EXAMPLES_ROOT "Build modules based on ROOT I/O" OFF)
```
paulgessinger, line 106: I saw that you had a discussion with the Robot already about this, but I'm wondering if we shouldn't turn this on by default actually, since it aligns with the *previous* configuration.

.github/workflows/builds.yml
```diff
@@ -111,6 +111,22 @@ jobs:
...
 
+      # We configure and build without ROOT in the same job to garantuee that we can use the ccache cache
+      - name: Configure (without ROOT)
+        run: >
+          ccache -z &&
+          rm -rf build &&
+          cmake -B build -S .
+          --preset=github-ci-no-root
+          -DCMAKE_INSTALL_PREFIX="${INSTALL_DIR}"
+          -DPython_EXECUTABLE=$(which python3)
```
paulgessinger, line 122: Maybe we can also disable the unit tests for this config? That shouldn't make a difference for what we're trying to test here, but I think we spend the majority of the time in this step on linking the unit test executables, since that's not sped up by ccache.

CMakeLists.txt
benjaminhuth: Can do it, I don't have a strong opinion on this...

.github/workflows/builds.yml
benjaminhuth: I actually found a bug because unit test where switched on (we have some examples unit test using root)

Examples/Python/src/RootInput.cpp
benjaminhuth: Hmm I think is is definitively a good idea, to abstract the whole `TTree` reading/writing a bit and package it, but this goes beyond the scope of this PR I think...

.github/workflows/builds.yml
paulgessinger: But that should be caught in the first run, shouldn't it?

paulgessinger (approved):

PR #4121, opened 2025-03-04
fix: allow to compile without examples.

goetzgaycken:
When compiling without examples the ActsExamplesFramework library does not exist, but it is also not needed.

--- END COMMIT MESSAGE ---


goetzgaycken:
There is likely a more elegant way to achieve the same which does not require to duplicate the cmake command

paulgessinger:
@goetzgaycken @stephenswat does this still apply?

andiwand:
@goetzgaycken can you have a look at the CI failure?

andiwand (approved):

PR #4124, opened 2025-03-10
feat: Gen3 geometry identifier assignment

paulgessinger:
--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **New Features**
	- Introduced a new API to add and configure geometry identifier nodes with method chaining.
	- Enabled custom sorting of tracking volumes via a user-provided comparator in both C++ and Python interfaces.

- **Refactor / Enhancements**
	- Updated method documentation for clarity regarding geometry identifier creation.
	- Improved error messages for duplicate volume identifiers.

- **Documentation**
	- Revised and restored documentation for the geometry identifier functionality.

### DIFF
--- a/Core/include/Acts/Geometry/BlueprintNode.hpp
+++ b/Core/include/Acts/Geometry/BlueprintNode.hpp
@@ -35,6 +35,7 @@ namespace Experimental {
 class MaterialDesignatorBlueprintNode;
 class StaticBlueprintNode;
 class LayerBlueprintNode;
+class GeometryIdentifierBlueprintNode;
 
 class CylinderContainerBlueprintNode;
 class CuboidContainerBlueprintNode;
@@ -230,6 +231,13 @@ class BlueprintNode {
   LayerBlueprintNode& addLayer(
       const std::string& name,
       const std::function<void(LayerBlueprintNode& layer)>& callback = {});
+
+  /// Convenience method for creating a @ref Acts::GeometryIdentifierBlueprintNode.
+  /// @param name The name of the geometry identifier node.
+  /// @param callback An optional callback that receives the node as an argument
+  GeometryIdentifierBlueprintNode& withGeometryIdentifier(
+      const std::function<void(
+          GeometryIdentifierBlueprintNode& geometryIdentifier)>& callback = {});
 
   /// @}
 
--- a/Core/include/Acts/Geometry/BlueprintNode.hpp
+++ b/Core/include/Acts/Geometry/BlueprintNode.hpp
@@ -232,8 +232,7 @@ class BlueprintNode {
       const std::string& name,
       const std::function<void(LayerBlueprintNode& layer)>& callback = {});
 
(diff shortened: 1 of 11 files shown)


Core/include/Acts/Geometry/BlueprintNode.hpp
```diff
@@ -231,6 +232,13 @@ class BlueprintNode {
       const std::string& name,
       const std::function<void(LayerBlueprintNode& layer)>& callback = {});
 
+  /// Convenience method for creating a @ref Acts::GeometryIdentifierBlueprintNode.
+  /// @param name The name of the geometry identifier node.
+  /// @param callback An optional callback that receives the node as an argument
+  GeometryIdentifierBlueprintNode& withGeometryIdentifier(
+      const std::function<void(
+          GeometryIdentifierBlueprintNode& geometryIdentifier)>& callback = {});
```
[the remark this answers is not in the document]
paulgessinger: Good catch!

Core/include/Acts/Geometry/GeometryIdentifierBlueprintNode.hpp
```diff
@@ -0,0 +1,110 @@
...
+  /// @note Requires exactly one child node
+  Volume& build(const BlueprintOptions& options, const GeometryContext& gctx,
+                const Logger& logger = Acts::getDummyLogger()) override;
+
+  /// @brief Connect portals in the volume hierarchy
+  /// @param options Blueprint build options
+  /// @param gctx The geometry context
+  /// @param logger The logger instance
+  /// @return Reference to the connected portal shell
+  PortalShellBase& connect(
```
asalzburger, line 55: I missed this: why does the `GeometryIdentifierBlueprintNode` have a connect ? Is this just fulfilling the API?

Core/src/Geometry/GeometryIdentifierBlueprintNode.cpp
```diff
@@ -0,0 +1,238 @@
...
+    // Skip volumes that were already in the parent before the subtree
+    // was processed
+    if (previous.contains(&v)) {
+      continue;
+    }
+
+    volumes.push_back(&v);
+  }
+
+  if (m_impl->m_sortBy) {
```
asalzburger, line 184: How is this implicitly converted into `bool`?

asalzburger (approved):
I would have two questions for clarification, which would be good if they could be answered.

Core/src/Geometry/GeometryIdentifierBlueprintNode.cpp
paulgessinger: `std::function` declares an [`operator bool`](https://en.cppreference.com/w/cpp/utility/functional/function/operator_bool), and I believe it's actually the only way to check if it contains a value.

Core/include/Acts/Geometry/GeometryIdentifierBlueprintNode.hpp
paulgessinger: Yes, that's exactly the reason!

PR #4128, opened 2025-03-11
ci: Set `errexit`, `nounset` in dependency script

stephenswat:
This commit adds the `-e` and `-u` flags to the CI dependency downloading shell script. I recently noticed that any error in some of these scripts is silently ignored, which causes the CI to fail later and with worse errors.

## Summary by CodeRabbit

- **Chores**
  - Enhanced internal script reliability with improved error handling for quicker and clearer feedback during operations.

### DIFF
--- a/CI/dependencies.sh
+++ b/CI/dependencies.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 
 function run() {
-    set -x
+    set -xeu
     "$@"
-    { set +x;  } 2> /dev/null
+    { set +xeu;  } 2> /dev/null
 }
 
 function set_env {


paulgessinger (dismissed):

stephenswat:
@paulgessinger seems like RTD is having some struggles. :thinking:

paulgessinger:
Right, I don't understand the error tbh.

paulgessinger:
@stephenswat there seem to be a few CI build failure now.

stephenswat:
I shrank the scope of this a lot to just the dependency downloading script. This was because we are sourcing a lot of these scripts, which makes the shell options propagate outside of the script.

paulgessinger:
Remaining errors look like they could be actual issues.

paulgessinger:
We can revisit this after #3980, where I'm changing the setup script anyway

PR #4125, opened 2025-03-10
chore: Update traccc to v0.20.0

stephenswat:
Building on work by @asalzburger in #4068, this commit updates traccc to v0.20.0; in the process, it also updates detray to v0.89.0, vecmem to v1.14.0, and covfie to v0.12.1.

--- END COMMIT MESSAGE ---

Closes #4068.

## Summary by CodeRabbit

- **Chores**
	- Updated dependency versions and source integrity checks.
	- Revised build configurations to include additional components.
- **Refactor**
	- Enhanced internal propagation and data conversion processes.
	- Updated Python detector bindings for streamlined performance.
- **New Features**
	- Added flexible numeric type support in field processing via a new field wrapper.
- **Tests**
	- Refined field conversion validations to ensure precise value checks.

PR #4130, opened 2025-03-13
ci: Replace clang-14 job with clang-19

stephenswat:
This commit replaces the clang-14 CI job based on Ubuntu 22.04 with a job using clang-19 based on Ubuntu 24.04. This is because clang-14 has a compiler bug that causes it to break on #4125.


AJPfleger (approved):
Happily accepted in the dev-meeting :)

PR #4133, opened 2025-03-13
fix: Improve and fix some interaction code

andiwand:
- check numbers and units
- minor code shuffles
- use `computeEnergyLossBethe` for mode approximation

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected variable naming and refined conditional checks to ensure more reliable energy calculations.
  - Updated unit representations and scaling factors for improved computational accuracy.

- **Refactor**
  - Transitioned the underlying energy loss computation method for enhanced precision.
  - Streamlined calculation logic to provide more consistent performance.

- **Tests**
  - Updated hash values for various test files to reflect recent changes in their content.


Core/src/Material/Interactions.cpp
```diff
@@ -114,19 +115,19 @@ inline float logDeriveEpsilon(float qOverP, const RelativisticQuantities& rq) {
...
-  /// Uses RPP2018 eq. 33.6 which is only valid for high energies.
+  // Uses RPP2018 eq. 33.6 which is only valid for high energies.
   // only relevant for very high ernergies; use arbitrary cutoff
   if (rq.betaGamma < 10.0f) {
     return 0.0f;
   }
   // pre-factor according to RPP2019 table 33.1
   const float plasmaEnergy =
-      PlasmaEnergyScale * std::sqrt(1000.f * molarElectronDensity);
+      PlasmaEnergyScale * std::sqrt(molarElectronDensity / (1 / 1_cm3));
```
benjaminhuth, line 128: I guess this is on purpose not `molarElectronDensity * 1_cm3`?

Core/src/Material/Interactions.cpp
```diff
@@ -462,7 +464,7 @@ float Acts::computeEnergyLossMode(const MaterialSlab& slab, PdgParticle absPdg,
                                   float m, float qOverP, float absQ) {
   // see ATL-SOFT-PUB-2008-003 section 3 for the relative fractions
   // TODO this is inconsistent with the text of the note
-  return 0.9f * computeEnergyLossLandau(slab, m, qOverP, absQ) +
+  return 0.9f * computeEnergyLossBethe(slab, m, qOverP, absQ) +
```
benjaminhuth, line 467: I don't have all these equations in mind currently, but I think it is worth a comment why this function is chosen to approximate this. Same for below...

benjaminhuth (commented):
Hmm it seems like the CI is still failing...

Core/src/Material/Interactions.cpp
andiwand: I think my thought here was that `molarElectronDensity` is expected to be in $cm^{-3}$ here. The way we usually convert these units is then to divide it by our native unit constants. Does that make sense to you? Happy to change it

Core/src/Material/Interactions.cpp
andiwand: It actually comes from the references note `ATL-SOFT-PUB-2008-003` https://cds.cern.ch/record/1114577/files/soft-pub-2008-003.pdf

benjaminhuth: Okay, so it was just wrong before?

benjaminhuth (approved):

benjaminhuth (dismissed):

benjaminhuth (approved):

acts-project-service:
## :red_circle: Athena integration test results [c738d9f6fafa0c4a44b37c7f1d0f8e9ecc6a0a10]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/53286588) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/11125890)!

PR #4134, opened 2025-03-17
fix: Increase the surfaceCandidateIndex in the Detector Navigator an re initialize navigation state in case of unreachable portal

dimitra97:
This PR can be also discussed, but

1. It fixes the Detector Navigator to increase the surface Candidate Index for the next Target
2. It takes care of cases where the nextTarget is a portal and it is not reachable anymore after the update from the stepper. We had a case in the tracking geometry of the MuonSpectrometer where the portal was unreachable after being updated from the stepper. I was trying to fix that and I think there are some approaches (not sure if I picked the most efficient). 

The options I thought that we had are the following:
- Let the propagator just continue until it reaches the pathLength limit, but this means it wouldn't try to intersect the other surfaces after this point, it just goes ahead.
- Try to change the boundaryTolerance for the surface candidates/portals. I tried to changed it to` AbsoluteBounds(1.1)
` and during the stepping updates the surface candidate was still reachable (as I understood in the update intersection from the stepper it uses the boundaryTolerance from the surface candidate which is `None` for a portal). However, the way that the boundaryTolerance can be changed it may not be really easy:
I thought either in the` initializeSurfaceCandidates ` where we force it to be `None `or when we fill in the `surfaceCandidates fillers `by also passing a parameter for the `boundaryTolerance `- but this is touching more central code maybe.

- The way I am trying to resolve it (after also discussing with andiwand ) is in the propagator by checking if we are in the case where the nextTarget is a portal and unreachable to re initialize the navigation state with the current stepping parameters. The problem here is that maybe it could be expensive - not sure how often it could happen
If you think which approach is better or propose another one

## Summary by CodeRabbit

- **Refactor**
  - Enhanced navigation logic for smoother transitions between surfaces and portals.
  - Improved target processing to ensure a more consistent and reliable navigation experience.


Core/include/Acts/Propagator/Propagator.ipp
andiwand, line 1: I believe these changes should not affect the `Propagator`, since this was not necessary for `Navigator` either

Core/include/Acts/Propagator/NavigationTarget.hpp
```diff
@@ -27,15 +27,17 @@ struct NavigationTarget {
...
   std::uint8_t surfaceIntersectionIndex = 0;
   BoundaryTolerance boundaryTolerance = BoundaryTolerance::None();
+  bool isPortal = false;
 
   static NavigationTarget None() { return NavigationTarget(); }
 
   NavigationTarget(const Surface& surface_,
                    std::uint8_t surfaceIntersectionIndex_,
-                   BoundaryTolerance boundaryTolerance_)
+                   BoundaryTolerance boundaryTolerance_, bool isPortal_ = false)
```
andiwand, line 36: I think a default parameter should not be necessary

Core/include/Acts/Propagator/NavigationTarget.hpp
andiwand, line 1: I wonder if we can push these changes into the `DetectorNavigator`. But potentially it can make sense to add them here as well

Core/include/Acts/Propagator/NavigationTarget.hpp
dimitra97: Actually I think this is not needed at all anymore! I will remove it

Core/include/Acts/Propagator/Propagator.ipp
```diff
@@ -54,6 +54,8 @@ Acts::Result<void> Acts::Propagator<S, N>::propagate(
           ConstrainedStep::Type::Navigator, logger());
       if (preStepSurfaceStatus == IntersectionStatus::reachable ||
           preStepSurfaceStatus == IntersectionStatus::onSurface) {
+        ACTS_VERBOSE("Found valid target surface after " << i << " attempts.");
+        ACTS_VERBOSE("Target is " << nextTarget.surface->geometryId());
```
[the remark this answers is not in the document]
dimitra97: No, sorry I forgot.

Core/include/Acts/Navigation/DetectorNavigator.hpp
```diff
@@ -174,7 +183,8 @@ class DetectorNavigator {
                                     state.surfaceCandidateIndex)
                                 << " out of " << state.surfaceCandidates.size()
                                 << " surfaces remain to try.");
-    // Take the surface
+
+    // Take the surface and mark it as a portal or a surface
```
andiwand, line 187:
```suggestion
    // Take the surface
```
nitpick: looks like this is not true anymore?

andiwand (dismissed):
LGTM 👍 one minor comment below

Nice that you pulled this off @dimitra97 👏 

This solution should work fine as long as the targeted portal is at the end of the candidates list. Potentially this creates problems with cylinders where you can exit the volume before walking through all the candidates. But I would not worry about this too much right now.

andiwand (approved):

acts-project-service:
## :red_circle: Athena integration test results [cb393184cb0571368e2aa2867d363ba14c7137f0]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/53295712) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/11127226)!

PR #4141, opened 2025-03-20
build: Make codegen respect system library flag

stephenswat:
This commit makes the codegen respect the `ACTS_USE_SYSTEM_LIBS` flag, allowing users to build ACTS in a fully offline environment if they have the dependencies already installed. To this end, I've added an `OFFLINE` flag to the codegen CMake function which will tweak the invocation of `uv` such that it will try to use system packages, and will refuse to install any new packages.

--- END COMMIT MESSAGE ---

Pinging @wdconinc as an interested party; see https://github.com/spack/spack/pull/49531.


Fatras/scripts/generate_particle_data_table.py
```diff
@@ -1,11 +1,5 @@
 #!/usr/bin/env python3
 
-# /// script
-# dependencies = [
-#   "particle==0.24.0",
-# ]
-# ///
```
paulgessinger, line 7: I thought it was nice having this written in the script.

cmake/ActsCodegen.cmake
```diff
@@ -84,7 +84,12 @@ function(acts_code_generation)
     )
 
     if(NOT DEFINED ARGS_PYTHON_VERSION)
-        set(ARGS_PYTHON_VERSION "3.13")
+        if(ARGS_OFFLINE)
+            find_package(Python REQUIRED COMPONENTS Interpreter)
+            set(ARGS_PYTHON_VERSION ${Python_EXECUTABLE})
```
paulgessinger, line 89: At the very least this should lock down the exact python version (3.13 right now) to ensure compatibility. 

The goal was to avoid any kind of breakage due to user setups and this was part of that.

paulgessinger (commented):
I was trying to avoid the need to manage the environment for the user.  

If we can't control the build to lock down Python versions and exact package versions, then I think we might want to abandon this approach entirely. 

I'm not sure I want to debug users Python environments, which is the reason I wanted to force separation from the system environment in all cases.

Core/src/Definitions/CMakeLists.txt
```diff
@@ -2,10 +2,20 @@ target_sources(ActsCore PRIVATE Common.cpp Direction.cpp ParticleData.cpp)
 
 include(ActsCodegen)
 
+if(ACTS_USE_SYSTEM_LIBS)
+    set(_codegen_offline_flags "OFFLINE")
```
paulgessinger, line 6: Is there a way to roll the detection into the `acts_code_generation` function, so we don't have to duplicate this in a number of places?

andiwand:
this has picked up conflicts @stephenswat - how do we proceed here? cc @paulgessinger

stephenswat:
I'll pick this up this week.

stephenswat:
Took another look at this, but I don't think it's possible to implement this anymore. uv is simply not designed to run offline or with system packages.

paulgessinger:
@stephenswat right, we might want to just skip uv in the offline case. We could delegate the installation and package version management to the user and just use whatever is there (without guarantees)

PR #4172, opened 2025-03-31
feat: Scoped timer utility

paulgessinger:
This adds a RAII utility to measure the timing in a block like this:

```cpp
{
  ScopedTimer timer{"Description", logger(), Logging::DEBUG};
  // Do some expensive work here
}
```

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **New Features**
	- Introduced an automated execution timer that measures and logs the duration of key processes, enhancing performance monitoring and diagnostics.


Core/include/Acts/Utilities/ScopedTimer.hpp
```diff
@@ -0,0 +1,70 @@
...
+///
+class ScopedTimer {
+ public:
+  /// @brief Construct a new Scoped Timer
+  ///
+  /// @param name Identifier for the timed block
+  /// @param lvl Logging level for the timing output
+  /// @param logger Logger instance to use for output
+  explicit ScopedTimer(const std::string& name,
+                       Logging::Level lvl = Logging::Level::INFO,
```
benjaminhuth, line 41: It could in some cases be more convenient to have the arguments ordered the other way around, because to actual get any output, you'll have to override the third argument, and by doing so always the second...
But this is very nitpicky of course.

PR #4181, opened 2025-04-01
feat: Use 80-bit high-precision floating point

stephenswat:
The current ACTS code uses 64-bit IEEE 754 floating point numbers virtually everywhere, but our CPUs have had fully functional 80-bit FPUs all along. That means that we are using 16 bits (20%) fewer bits than possible, and therefore we are not getting the precision we could really use in track finding. After all, if you're going to build a multi-billion CHF accelerator and experiment, why not use the highest possible precision?

This commit changes all the 64-bit SSE floats by more precise 80-bit x87 floating point numbers, which will increase our precision and have no CPU performance impact at all.


asalzburger:
Looking forward to the CI results ...

stephenswat:
> Looking forward to the CI results ...

Looks good to me. :+1:

PR #4184, opened 2025-04-02
refactor: remove useless asserts

CarloVarni:
These asserts are not usefull. The SpacePointData object is constructed by the SpacePointContainer and the size of the vectors, as well as the number of space point proxies, are always consistent. Thus, there is no way these asserts can complain.

\cc @pbutti


paulgessinger (commented):
Hm. They protect against a bug in the sizing or lookup implementation. Do the asserts hurt?

PR #4186, opened 2025-04-03
feat: Add an averaging timer

paulgessinger:
This timer is a variation on the `ScopedTimer`: instead of recording it's own lifetime duration, it can collect samples (think: durations of loops) and will print the number of samples, their total time, mean and standard deviation when it's destroyed.

```cpp
{
  Acts::AveragingScopedTimer timer{"Measure", logger()};
  for(std::size_t i=0;i<1000;i++) {
    auto sample = timer.sample()
    // Do expensive stuff here
  }
  // `timer` prints output here now
}

```

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **New Features**
  - Introduced an enhanced performance monitoring tool that aggregates timing samples and displays statistical insights such as average and variability.

- **Chores**
  - Updated internal build configurations to support the new performance measurement components.


stephenswat (approved):
LGTM!

PR #4210, opened 2025-04-12
build: Conditionally enable ROOT for Examples only

paulgessinger:
This PR changes the handling of the `ACTS_BUILD_EXAMPLES_ROOT` flag, and other `*_EXAMPLES_*` flags to be only evaluated if the examples are configured to be built at all.

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **Chores**
  - Adjusted the build configuration so that components relying on ROOT I/O are now disabled by default.
- **Documentation**
  - Updated the getting started guide to reflect the new default setting for these components.


benjaminhuth (approved):
LGTM

PR #4212, opened 2025-04-13
feat: Add blue print from geomodel test for the muon spectrometer

dimitra97:
This PR introduces the first effort for the migration to gen3 for the Muon Spectrometer in roder to be able to tests things in ACTS standalone
- Change the functionality in the geomodel factory in order to return also `Volume` instead of only `DetectorVolume` objects for gen3 geometry use cases
- Add a test for the conversion from geomodel to gen3 scheme with the blueprint to perform the tests with the navigaiton models.


Tests/UnitTests/Plugins/GeoModel/GeoModelMuonMockupBluePrintTests.cpp
```diff
@@ -0,0 +1,236 @@
...
+    std::unique_ptr<Acts::TrackingVolume> chamberVolume =
+        std::make_unique<Acts::TrackingVolume>(
+            volTrf, chamberBounds,
+            "Chamber_" + name + "_" + std::to_string(sector + 1));
+
+    // add the multilayer volumes
+    chamberVolume->addVolume(std::make_unique<TrackingVolume>(
+        *vol1, "Chamber_" + name + "_" + std::to_string(sector + 1) + "_MDT1"));
+    chamberVolume->addVolume(std::make_unique<TrackingVolume>(
+        *vol2, "Chamber_" + name + "_" + std::to_string(sector + 1) + "_MDT2"));
```
paulgessinger, line 127: What you're doing here is essentially fully assembling the cuboids which contain the individual multilayers, correct?

This is one way to do it, especially given the fact that there's currently no way to make a node that produces an *auto-sized* wrapping volume. I think down the road, I'll want to change this to have a "wrapper cuboid" node type that produces a cuboid that self-sizes to fit around the children that it receives.

That way, it would not be necessary to calculate the correct sizing for the chamber volume individually here.

What do you think @dimitra97?

Tests/UnitTests/Plugins/GeoModel/GeoModelMuonMockupBluePrintTests.cpp
```diff
@@ -0,0 +1,236 @@
...
+    for (const auto& surface : sensitiveSurfaces) {
+      auto sname = std::get<0>(surface)->databaseEntryName();
+
+      // Skip MDT surfaces and surfaces not in the current sector
+      if (sname.find("MDT") != std::string::npos ||
+          sname.find("_1_" + std::to_string(sector + 1)) == std::string::npos) {
+        continue;
+      }
+
+      chamberVolume->addSurface(std::get<1>(surface));
```
paulgessinger, line 139: The surfaces go directly into the `chamberVolume`, and not into the corresponding `MDT` volume?

Tests/UnitTests/Plugins/GeoModel/GeoModelMuonMockupBluePrintTests.cpp
```diff
@@ -0,0 +1,236 @@
...
+BOOST_AUTO_TEST_SUITE(GeoModelMuonMockupBluePrintTests)
+
+BOOST_AUTO_TEST_CASE(MockupMuonGen3Geometry) {
+  // Read the geomodel file to create the mockup geometry
+
+  const GeometryContext context;
+  auto logger = Acts::getDefaultLogger("UnitTests", Acts::Logging::VERBOSE);
+
+  Acts::GeoModelTree gmTree =
+      readFromDb("/eos/user/c/cimuonsw/GeometryFiles/DimitrasMockup.db");
```
paulgessinger, line 185: Clearly ok for testing, but we'll have to think of a solution for this when we want to merge this.

Tests/UnitTests/Plugins/GeoModel/GeoModelMuonMockupBluePrintTests.cpp
```diff
@@ -0,0 +1,236 @@
...
+  auto& cyl =
+      root.addCylinderContainer("CylinderContainer", AxisDirection::AxisR);
+  cyl.setAttachmentStrategy(VolumeAttachmentStrategy::Gap);
+
+  // Add the volumes to the blueprint
+  cyl.addChild(std::move(innerBarrel));
+  cyl.addChild(std::move(middleBarrel));
+  cyl.addChild(std::move(outerBarrel));
+
+  auto trackingGeometry = root.construct({}, context, *logger);
```
paulgessinger, line 233: You could add at least a basic check that the total number of volumes and the total number of surfaces is the expected one.

Tests/UnitTests/Plugins/GeoModel/GeoModelMuonMockupBluePrintTests.cpp
dimitra97: I agree. Could it be to add also the file ?

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelDetectorObjectFactory.hpp
```diff
@@ -49,8 +52,15 @@ class GeoModelDetectorObjectFactory {
   struct Cache {
     // The created detector elements and their surfaces
     std::vector<GeoModelSensitiveSurface> sensitiveSurfaces;
-    // The created representation of bounding box
+
+    // The created representation of bounding boxes as a pair of the Volume and
+    // DetectorVolume
+    // for gen1/3 and gen2 geometry constructions
```
AJPfleger, line 58:
```suggestion
    // DetectorVolume for gen1/3 and gen2 geometry constructions
```

dimitra97:
I think I am closing it because the whole work will be supported by: https://github.com/acts-project/acts/pull/4276

PR #4215, opened 2025-04-14
refactor(Examples): Sequencer runs events more in order

paulgessinger:
We were using tbb's blocked range, which partitions the input event
range into chunks and then has a number of threads process events in
these chunks. Since our input data only depends on the event number and
our tasks are so big that inter-event memory caching likely has
negligible impact, we don't benefit from this.

OTOH, if our events were processed less out of range, so more
sequentially, we could benefit when we're trying to serialize events for
e.g. IO.

This commit changes the parallel for loop in the Sequencer to not use
blocked ranges but just pick individual elements. 
Each thread then selects the next event by incrementing an atomic
event index integer. This effectively represents a queue, but since our
"workloads" are fully determined and take nothing as input other than
the event number, we don't need an actual queue.

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **New Features**
	- Improved concurrent event processing enhances performance and stability.
	- Updated logging now provides clearer tracing of parallel task execution.

### DIFF
--- a/Core/src/Navigation/SurfaceArrayNavigationPolicy.cpp
+++ b/Core/src/Navigation/SurfaceArrayNavigationPolicy.cpp
@@ -37,6 +37,13 @@ SurfaceArrayNavigationPolicy::SurfaceArrayNavigationPolicy(
     surfaces.push_back(surface.getSharedPtr());
   }
 
+  ACTS_VERBOSE("Number of surfaces passed to the surface array creation: "
+               << surfaces.size());
+  if (surfaces.empty()) {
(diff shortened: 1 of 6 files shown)


acts-project-service:
## :red_circle: Athena integration test results [88ed2f259b9fc01f6b74b95ab52f27fa9621d969]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/54693141) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/11775876)!

PR #4216, opened 2025-04-14
feat: Add `initialSigmaQoverPt` to `estimateTrackParamCovariance`

andiwand:
Currently we offset the initial $\sigma(q/p)$ with a constant which is causing problems at high eta. The reason for this is that $q/p$ is much smaller at high eta with constant $p_T$. It would be better to offset with a constant $\sigma(q/p_T)$ as this would scale correctly for $q/p$ and represents the intrinsic resolution limit for a cylindrical detector for a given magnetic field.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new parameter to explicitly set the initial uncertainty for charge over transverse momentum (Q/pT) in track parameter estimation and seeding configuration.
  - Python and C++ interfaces updated to allow configuration of this new uncertainty parameter.

- **Improvements**
  - Enhanced documentation and configuration options for initial momentum uncertainty, now distinguishing between relative transverse momentum and Q/pT uncertainties.
  - Updated example scripts and workflows to utilize the new parameter for more precise control over seeding uncertainties.

- **Tests**
  - Updated test data to reflect changes in track parameter estimation behavior.


pbutti:
I don't see major issues from the CI plots

CI/physmon/workflows/physmon_trackfinding_1muon.py
```diff
@@ -128,10 +128,11 @@ def run_ckf_tracking(label, seeding):
                 1 * u.mm,
                 1 * u.degree,
                 1 * u.degree,
-                0.1 * u.e / u.GeV,
+                0 * u.e / u.GeV,
```
pbutti, line 131: This means that initial Q/p sigma is 0 if I understand correctly the logic, right?

Core/src/Seeding/EstimateTrackParamsFromSeed.cpp
```diff
@@ -107,10 +107,14 @@ Acts::BoundMatrix Acts::estimateTrackParamCovariance(
       // note that we rely on the fact that sigma theta is already computed
       double varianceTheta = result(eBoundTheta, eBoundTheta);
 
-      // transverse momentum contribution
+      // contribution from sigma(q/pt)
+      variance += std::pow(
+          config.initialSigmaQoverPt * std::sin(params[eBoundTheta]), 2);
+
+      // contribution from sigma(pt)/pt
       variance += std::pow(config.initialSigmaPtRel * params[eBoundQOverP], 2);
```
pbutti, line 115: Sorry for the naive question, but can you explain why are you still adding this term if you have a non 0 initial sigmaQoverPt here? 
if I write q/p as (q/pT)*sin(theta), then 
sigma(q/p)^2 = |sin(theta)|^2*sigma(q/pT)^2 + |(q/p)*1/tan(theta)|^2*var(theta). Why there is an additional pT term?

andiwand: we still want to keep the error contribution from $\sigma(p_T)/p_T$ which scales with the initial $q/p$. the additional term with $\sigma(q/p_T)$ basically only sets a minimal error for high momentum.

andiwand: I think it does not hurt to add both. I do not see an immediate use case but the user is in control of these values and can choose how to set them. Potentially it might be useful to parameterize with all three values?

pbutti: I find it a bit confusing having an initial q/p and an initial q/pT, honestly. If you pass both != 0, then you basically double count the error. It doesn't necessarily harm, but perhaps one could add some comments / make it clear in the config or in the header. 

If you pass q/pT = 0, then we are in the original behavior, if q/p=0 instead, you use the new formulation. if both q/pT and q/p !=0 then one inflates the original error in the end, I guess.

andiwand: I will try to clarify in the documentation 👍 

Right, if you set both != 0 you get a mixture. At the end you just have more free params at hand to model something. I thought it might be useful and it is less confusing without a p vs pT toggle

pbutti (approved):

PR #4219, opened 2025-04-16
feat: Navigation with Gen3 geometry

paulgessinger:
- https://github.com/acts-project/acts/pull/4349
- https://github.com/acts-project/acts/pull/4350
- https://github.com/acts-project/acts/pull/4351

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **New Features**
  - Navigation now supports multiple geometry versions, enabling distinct navigation behaviors for different geometry generations.
  - Navigation arguments include new options to control whether portals and surfaces are considered during navigation.

- **Bug Fixes**
  - Improved candidate initialization logic to respect both configuration settings and user-specified navigation preferences.

- **Refactor**
  - Updated internal data structures to include portal information in boundary intersections and adjusted related logic for consistency.

- **Chores**
  - Enhanced logging to reflect geometry version during navigation initialization.


paulgessinger:
This is now ready for review I believe. @andiwand could you have a look when you get a chance and then we discuss this?

Core/include/Acts/Navigation/NavigationDelegate.hpp
```diff
@@ -25,10 +25,14 @@ struct NavigationArguments {
...
 
   BoundaryTolerance tolerance = BoundaryTolerance::None();
+
+  bool wantsPortals = true;
+  bool wantsSurfaces = true;
 };
 
 /// Central alias for the navigation delegate. This type is owning to support
 /// (type-erased) navigation delegate chains (i.e. multiple policies).
+// @TODO: Add geometry context to navigation delegate signature
```
andiwand, line 35: is this breaking and should be done in a follow up?

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -224,7 +230,12 @@ class Navigator {
   explicit Navigator(Config cfg,
                      std::shared_ptr<const Logger> _logger =
                          getDefaultLogger("Navigator", Logging::Level::INFO))
-      : m_cfg{std::move(cfg)}, m_logger{std::move(_logger)} {}
+      : m_cfg{std::move(cfg)}, m_logger{std::move(_logger)} {
+    if (m_cfg.trackingGeometry == nullptr) {
+      throw std::invalid_argument("Navigator: No tracking geometry provided.");
+    }
```
andiwand, line 236: Looks like this is a case we handle right now
https://github.com/acts-project/acts/blob/e98b1686390b75a4a24775069f7f70949d441556/Core/include/Acts/Propagator/Navigator.hpp#L739-L742
I think it makes sense to push this out. Can you remove the special case handling below?

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -279,11 +290,29 @@ class Navigator {
...
+        default:
+          throw std::runtime_error("Unknown geometry version.");
+      }
+    };
+    ACTS_VERBOSE(volInfo(state) << "Geometry version is: "
+                                << printGeometryVersion(m_geometryVersion));
+
     state.reset();
 
+    state.stream.candidates().reserve(50);
```
andiwand, line 311: magic number - could be pushed into config / options

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -443,8 +476,8 @@ class Navigator {
         }
         if (state.navBoundaryIndex.value() < state.navBoundaries.size()) {
           ACTS_VERBOSE(volInfo(state) << "Target set to next boundary.");
-          return NavigationTarget(*state.navBoundary().first.object(),
-                                  state.navBoundary().first.index(),
+          return NavigationTarget(*std::get<0>(state.navBoundary()).object(),
+                                  std::get<0>(state.navBoundary()).index(),
```
andiwand: potentially it is time to have a struct for this?

Tests/UnitTests/Core/Propagator/MultiStepperTests.cpp
```diff
@@ -831,8 +832,16 @@ template <typename multi_stepper_t>
...
+  auto bounds = std::make_shared<Acts::CylinderVolumeBounds>(10, 20, 20);
+  auto tvol = std::make_shared<TrackingVolume>(Transform3::Identity(), bounds,
+                                               "Undefined");
+
+  auto tgeo = std::make_shared<Acts::TrackingGeometry>(tvol);
+
   Propagator<multi_stepper_t, Navigator> propagator(
-      std::move(multi_stepper), Navigator{Navigator::Config{}});
+      std::move(multi_stepper),
+      Navigator{Navigator::Config{.trackingGeometry = tgeo}});
```
andiwand: can this test live with a `VoidNavigator`? looks like this was effectively the case before?

andiwand (commented):
Makes sense - left a few comments / suggestions / questions

Core/include/Acts/Propagator/Navigator.hpp
paulgessinger: Ah, I actually ran into just this, and added a check to the constructor. I agree that this navigator should just always require a tracking geometry to work.

Core/include/Acts/Navigation/NavigationDelegate.hpp
paulgessinger: It is breaking, I think, and I didn't want to do it here before I've settled on the actual signature, because it might change down the road again when the navigation is finalized.

Core/include/Acts/Propagator/Navigator.hpp
paulgessinger: It's more or less completely arbitrary. I'm leaning towards not making this user facing (yet), but I'll add a comment that this is arbitrary.

Core/include/Acts/Propagator/Navigator.hpp
paulgessinger: Can do, although ultimately this will become a pair again, when the `BoundarySurface` class goes away.

PR #4251, opened 2025-05-07
build: Require HepMC version 3.2.5

stephenswat:
The HepMC code in ACTS requires some functions (like `vertices_size`) that are available only in HepMC 3.2.5 onward, but the CMake configuration suggests that 3.2.1 is permissible. This commit updates the version requirement to the correct 3.2.5 minimum.

## Summary by CodeRabbit

- **Chores**
	- Updated the minimum required version of the HepMC3 dependency to 3.2.5 for example builds.


paulgessinger (approved):

paulgessinger:
Closing this in favor of https://github.com/acts-project/acts/pull/4323/, which sets it to 3.2.4 and actually removes the use of the unavailable functions.

PR #4255, opened 2025-05-08
build: Allow to toggle setup of traccc subprojects

benjaminhuth:
This allows to turn off the setup of `algebra-plugins`, `vecmem`, detray`, and `covfie` in ACTS and delegate this to `traccc`.
Also includes a convenience macro to shorten the code.

--- END COMMIT MESSAGE ---

@stephenswat @krasznaa


CMakeLists.txt
```diff
@@ -467,32 +471,31 @@ if(ACTS_BUILD_PLUGIN_GEANT4)
...
-        find_package(algebra-plugins ${_acts_algebraplugins_version} REQUIRED)
-    else()
-        add_subdirectory(thirdparty/algebra-plugins)
-    endif()
-
-    if(ACTS_USE_SYSTEM_DETRAY)
-        find_package(detray ${_acts_detray_version} REQUIRED CONFIG)
+# Convenience macro to simplify the check if we want to setup, and if
+# so, if we want to find it in the system or build internally
+macro(conditional_setup _var _package _path)
```
krasznaa, line 476: Not sure if this is already done consistently for all the other functions/macros, but let's use an `acts_` prefix for this macro. (Unfortunately functions/macros enter a global namespace with CMake, so conventions like this are pretty necessary.)

CMakeLists.txt
```diff
@@ -467,32 +471,31 @@ if(ACTS_BUILD_PLUGIN_GEANT4)
...
+    if(ACTS_SETUP_${_var})
+        if(ACTS_USE_SYSTEM_${_var})
+            message(VERBOSE "Use system ${_package}")
```
krasznaa, line 479: Careful with all these `VERBOSE` messages! These have been introduced in CMake 3.15.

https://cmake.org/cmake/help/latest/command/message.html#general-messages

But Acts only seems to require CMake 3.14 at the moment.

https://github.com/acts-project/acts/blob/main/CMakeLists.txt#L1

I have to admit I didn't even know that `VERBOSE` messages existed. And I wonder what the use-case really is. Since debugging CMake can be done without the help of debugging messages. (CMake can be put in a trace mode where you can check exactly what's happening.)

So on first order I'd just print the meaningful / useful stuff as `STATUS`, and not bother about any `VERBOSE` messages.

CMakeLists.txt
```diff
@@ -467,32 +471,31 @@ if(ACTS_BUILD_PLUGIN_GEANT4)
...
 
-    if(ACTS_USE_SYSTEM_VECMEM)
-        find_package(vecmem ${_acts_vecmem_version} REQUIRED)
-    else()
-        add_subdirectory(thirdparty/vecmem)
-        # Make the "VecMem language code" available for the whole project.
+if(ACTS_BUILD_PLUGIN_TRACCC)
+    conditional_setup(ALGEBRAPLUGINS algebra-plugins thirdparty/algebra-plugins _acts_algebraplugins_version REQUIRED)
+    conditional_setup(DETRAY detray thirdparty/detray _acts_detray_version REQUIRED CONFIG)
+    conditional_setup(VECMEM vecmem thirdparty/vecmem _acts_vecmem_version REQUIRED)
```
krasznaa, line 493: Don't hide this behind `ACTS_BUILD_PLUGIN_TRACCC`!  Move the setup of the externals back to the top level.

In general the setup of externals, and the decision about which plugins are built, are independent of each other. By combining the logic like this, you'll just invite trouble down the road.

Granted, the old SYCL plugin (which used to use vecmem) has been removed some time ago. But we could absolutely have more elaborate relationships between externals and plugins than the one proposed here.)

krasznaa (changes requested):
I didn't check, but you may need to update the presets a little as well after these flag changes.

https://github.com/acts-project/acts/blob/main/CMakePresets.json

CMakeLists.txt
benjaminhuth: good point!

CMakeLists.txt
benjaminhuth: yeah we can remove this... Good to know, I didn't new about trace mode actually

CMakeLists.txt
benjaminhuth: Yeah so I just took this from the previous state. So the plugins where set up only within `ACTS_BUILD_PLUGIN_TRACCC` before...
So currently, the whole cmake is set up to build either everything from the traccc-universe or nothing, like in `Plugins/CMakeLists.txt`:
```
add_component_if(Detray PluginDetray ACTS_BUILD_PLUGIN_TRACCC)
add_component_if(Covfie PluginCovfie ACTS_BUILD_PLUGIN_TRACCC)
```
So they are all triggered by `ACTS_BUILD_PLUGIN_TRACCC`.

I think I would not change/generalize this in that PR, since it seems this granularity is not needed right now. Maybe we can disentangle this later?

krasznaa: You're right. I didn't check, the current logic is indeed like this. :thinking: I believe I may have complained about this also when it was first set up.

Okay, let's keep the current logic then. But I still think that the logic in this file should be reduced in the future, and valid configurations could be provided to the users through presets. Unfortunately there's no golden path here, as this many flags will never be possible to set up such that the user couldn't set them to an invalid combination. We should (I believe) just make sure that technically as many combinations can be expressed using the flags as possible.

benjaminhuth:
@krasznaa I adopted your feedback. I think the presets are fine, since I set the `ACTS_SETUP_<package>` variables to `ON` by default, so the current behaviour should be preserved with this PR.
If you are fine, can you approve the PR?

CMakeLists.txt
paulgessinger: Just chiming in here that I think we should stick to the all-or-nothing configuration to reduce complexity. 

In my opinion, both

- whether we use traccc et al at all
- if we use fetch content or find package 

Should be behind a single flag each. We should not introduce additional granularity and I advocated for not having individual plugins and flags for the sub projects.

paulgessinger:
I don't fully understand what the purpose of this is. Under which circumstances to we want to delegate this to traccc? If we eventually want to remove traccc, these options become redundant anyway, don't they?

benjaminhuth:
The reason I want that in is that we currently cannot delegate the setup of `detray`, `covfie`, ... to `traccc` right now. Without this, one always needs to manually figure out what are the correct dependencies, and as we know in the dependency-hell of all this projects this is not nice...

krasznaa (approved):
I'm on board with this setup. :+1:

Don't know what the CI complaint about the documentation is though. :confused:

PR #4256, opened 2025-05-08
feat: Gen3 Migration of MultiWireVolume Builder class and Navigation Policy for the Muons

dimitra97:
- Implement the` MultiWireVolumeBuilder` for MDT's multilayers structures (based on Gen2 implementation which is gonna be discarded).
- Implement `MultiLayerNavigationPolicy` for MultiLayer Structures and attaches the policy to the tracking volume of the `MultiWireVolumeBuilder`
- Include test for gen3

general comments: It is based on some Gen2 related interfaces (e.g the IndexedSurfacesNavigation from IInternalNavigation - can be migrated soon as well though). Also Needed to make the Policy templated for the grid so I made type alias for static assert tests (open to hear other suggestions as well). 

@junggjo9 @noemina @paulgessinger 

## Summary by CodeRabbit

- **New Features**
  - Introduced support for building multiwire tracking volumes, such as wire chambers, with a new volume builder.
  - Added a multilayer navigation policy for efficient navigation through indexed grids of surfaces in tracking volumes.

- **Bug Fixes**
  - Corrected surface orientation and positioning for trapezoidal volume bounds, improving geometry accuracy.

- **Tests**
  - Enhanced and expanded unit tests for multiwire navigation, including new test cases and reusable surface generation utilities.


Core/include/Acts/Geometry/MultiWireVolumeBuilder.hpp
```diff
@@ -0,0 +1,66 @@
...
+    std::string name = "undefined";
+
+    // The surfaces to be wrapped from the tracking volume
+    std::vector<std::shared_ptr<Surface>> mlSurfaces = {};
+
+    /// The transform of the tracking volume
+    Transform3 transform = Transform3::Identity();
+
+    /// The bounds of the tracking volume
+    std::vector<double> bounds = {};
```
junggjo9, line 36: Are these the 3D bounds? Why not using the volume bound object?

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
```diff
@@ -0,0 +1,178 @@
...
+#include "Acts/Utilities/Grid.hpp"
+
+namespace Acts {
+
+/// Concept for the indexed grid type
+template <typename T>
+concept IndexGridConcept = requires(T igrid) {
+  igrid.grid;
+  igrid.transform;
+  igrid.extractor;
```
junggjo9, line 26: Are these methods or data types? If data types can you specify what they should return?

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
```diff
@@ -0,0 +1,178 @@
...
+      surfaces.push_back(surface.getSharedPtr());
+    }
+
+    Acts::Experimental::detail::CenterReferenceGenerator rGenerator;
+    Acts::Experimental::detail::IndexedGridFiller filler{config.binExpansion};
+    filler.fill(gctx, m_indexedGrid, surfaces, rGenerator, {});
+
+    // Define step size and number of steps
+    m_step = std::sqrt(std::pow(m_indexedGrid.grid.binWidth()[0], 2) +
+                       std::pow(m_indexedGrid.grid.binWidth()[1], 2));
```
junggjo9, line 74: Propose to use  - https://github.com/acts-project/acts/blob/main/Core/include/Acts/Utilities/MathHelpers.hpp#L25-L28

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
```diff
@@ -0,0 +1,178 @@
...
+  /// @param logger The logger
+  void initializeCandidates(const NavigationArguments& args,
+                            AppendOnlyNavigationStream& stream,
+                            const Logger& logger) const {
+    ACTS_VERBOSE(
+        "MultiLayerNavigationPolicy Candidates initialization for volume"
+        << m_volume.volumeName());
+    const Transform3& transform = m_volume.transform();
+    const Acts::Vector3 locPosition = transform * args.position;
+    const Acts::Vector3 locDirection = transform.linear() * args.direction;
```
junggjo9, line 90: Isn't it that the transform of a volume is usually a local -> global one? If so how can that be a local position?

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+    std::vector<Vector3> path =
+        generatePath(locPosition, locDirection, m_step, m_nSteps);
+
+    const auto& surfaces = m_volume.surfaces();
+    std::vector<const Surface*> surfCandidates = {};
```
junggjo9, line 96: I'd call a reserve after to avoid multiple deallocations

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+    for (const auto& pos : path) {
+      std::vector<const Surface*> eSurfaces;
```
junggjo9, line 99: Move that out of the loop and reserve as much memory as you've surfaces  and empty the vector at the beginning of each call?

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
```diff
@@ -0,0 +1,178 @@
...
+      std::vector<std::size_t> indices = m_indexedGrid.grid.atPosition(pos);
+      eSurfaces.reserve(indices.size());
+
+      std::for_each(indices.begin(), indices.end(),
+                    [&](const auto& i) { eSurfaces.push_back(&surfaces[i]); });
+
```
junggjo9, line 105: or even better you might want to use std::ranges::transform(indices, std::back_inserter(surfCandidates),[]);

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
```diff
@@ -0,0 +1,178 @@
...
+  std::vector<Vector3> generatePath(const Vector3& startPosition,
+                                    const Vector3& direction, double stepSize,
+                                    std::size_t numberOfSteps) const {
+    std::vector<Vector3> path;
+    path.reserve(numberOfSteps);
+    for (std::size_t i = 0; i < numberOfSteps; ++i) {
+      path.push_back(startPosition + i * stepSize * direction);
+    }
+    return path;
+  }
```
junggjo9, line 142: Hmm okay, you're going multiple straightline steps through the chamber without updating the direction. I'm wondering whether it wouldn't be easier to just intersect the tube planes and from each tube intersection you could count the number of tube pitches you're away from the first surface, add another +- candidate and you're done? This way you also spare the duplicate surface check because you know that each step will just guide you into a single layer.

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
andiwand: `perp` would also be an option https://github.com/acts-project/acts/blob/d19a6d75ad5df23f45aaa776b0129137ba941d38/Core/include/Acts/Utilities/VectorHelpers.hpp#L58-L77

Core/include/Acts/Geometry/MultiWireVolumeBuilder.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include "Acts/Utilities/Logger.hpp"
+#include "Acts/Utilities/ProtoAxis.hpp"
+
+namespace Acts {
+
+/// @class MultiWireVolumeBuilder
+
+///@brief A class to build multiwire tracking volumes (e.g wire chambers)
+
+class MultiWireVolumeBuilder {
```
andiwand, line 22:
```suggestion
/// @class MultiWireVolumeBuilder
/// @brief A class to build multiwire tracking volumes (e.g wire chambers)
class MultiWireVolumeBuilder {
```

Core/include/Acts/Geometry/MultiWireVolumeBuilder.hpp
andiwand, line 1: Naive question: Does this need to be in Core? Is the reason that the same class can be used in Athena?

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
```diff
@@ -0,0 +1,168 @@
...
+  /// @param logger The logger
+  void initializeCandidates(const NavigationArguments& args,
+                            AppendOnlyNavigationStream& stream,
+                            const Logger& logger) const {
+    ACTS_VERBOSE(
+        "MultiLayerNavigationPolicy Candidates initialization for volume"
+        << m_volume.volumeName());
+    const Transform3& itransform = m_volume.transform().inverse();
+    const Acts::Vector3 locPosition = itransform * args.position;
+    const Acts::Vector3 locDirection = itransform * args.direction;
```
andiwand, line 89: nitpick: `Acts::` is redundant and should be dropped

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
```diff
@@ -0,0 +1,168 @@
...
+ private:
+  // The tracking volume
+  const TrackingVolume& m_volume;
+
+  // The grid that holds the indexed surfaces
+  indexed_grid m_indexedGrid;
+};
+
+// Make alias for the static assert check for the Navigation Policy
+using gridType =
```
andiwand, line 159:
```suggestion
// Make alias for the static assert check for the Navigation Policy
using GridType =
```
naming convention

but more importantly: this goes into the `Acts` namespace right? `GridType` seems to be very generic for a specific header like `MultiLayerNavigationPolicy.hpp`

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
```diff
@@ -0,0 +1,168 @@
...
+    Grid<std::vector<std::size_t>,
+         Axis<AxisType::Equidistant, Acts::AxisBoundaryType::Bound>,
+         Axis<AxisType::Equidistant, Acts::AxisBoundaryType::Bound>>;
+using indexedUpdatorType = Experimental::IndexedSurfacesNavigation<gridType>;
```
andiwand, line 163:
```suggestion
using IndexedUpdatorType = Experimental::IndexedSurfacesNavigation<gridType>;
```
forwarding `Experimental::` to `Acts::` looks a bit scary

Core/src/Geometry/MultiWireVolumeBuilder.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include "Acts/Geometry/TrapezoidPortalShell.hpp"
+#include "Acts/Geometry/TrapezoidVolumeBounds.hpp"
+#include "Acts/Navigation/InternalNavigation.hpp"
+#include "Acts/Navigation/MultiLayerNavigationPolicy.hpp"
+#include "Acts/Navigation/TryAllNavigationPolicy.hpp"
+#include "Acts/Surfaces/SurfaceArray.hpp"
+#include "Acts/Utilities/Helpers.hpp"
+#include "Acts/Utilities/StringHelpers.hpp"
+
+// Constructor
```
andiwand, line 22:
```suggestion
```

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+    auto maxXIndex = m_indexedGrid.grid.numLocalBins()[0];
+    auto maxYIndex = m_indexedGrid.grid.numLocalBins()[1];
+
+    for (std::size_t i = 0; i < maxYIndex; i++) {
+      auto v1 = m_indexedGrid.grid.lowerLeftBinEdge({1, i + 1});
+      auto v2 = m_indexedGrid.grid.upperRightBinEdge({maxXIndex, i + 1});
+
+      auto intersection = detail::IntersectionHelper2D::intersectSegment(
+          Vector2(v1[0], v1[1]), Vector2(v2[0], v2[1]),
```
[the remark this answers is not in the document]
dimitra97: this is an std::array

Core/include/Acts/Geometry/MultiWireVolumeBuilder.hpp
dimitra97: The plan is to use it in Athena for MDTs construction (as we did for gen2)

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
junggjo9:
```suggestion
    const Vector3 locPosition = itransform * args.position;
    const Vector3 locDirection = itransform * args.direction;
```

Core/include/Acts/Geometry/MultiWireVolumeBuilder.hpp
```diff
@@ -0,0 +1,64 @@
...
+  /// @param config The configuration struct
+  /// @param logger The logger instance for screen output
+  explicit MultiWireVolumeBuilder(
+      const Config& config,
+      std::unique_ptr<const Acts::Logger> logger = Acts::getDefaultLogger(
+          "MultiWireVolumeBuilder", Acts::Logging::VERBOSE));
+
+  /// Destructor
+  ~MultiWireVolumeBuilder() = default;
+
```
andiwand, line 50:
```suggestion
```

can be dropped in favor of implicit default destructor

Core/src/Geometry/MultiWireVolumeBuilder.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include "Acts/Geometry/TrapezoidPortalShell.hpp"
+#include "Acts/Geometry/TrapezoidVolumeBounds.hpp"
+#include "Acts/Navigation/InternalNavigation.hpp"
+#include "Acts/Navigation/MultiLayerNavigationPolicy.hpp"
+#include "Acts/Navigation/TryAllNavigationPolicy.hpp"
+#include "Acts/Surfaces/SurfaceArray.hpp"
+#include "Acts/Utilities/Helpers.hpp"
+#include "Acts/Utilities/StringHelpers.hpp"
+
+Acts::Experimental::MultiWireVolumeBuilder::MultiWireVolumeBuilder(
```
andiwand, line 22: if you do `namespace Acts::Examples {...}` like in the header you can safe a lot of characters by dropping all `Acts::` and `Acts::Experimental::`

andiwand (commented):
left two minor comments

PR #4262, opened 2025-05-12
fix: Link to https version of DD4hep website

benjaminhuth:
The http version doesn't work at least for me and seems to cause issues in the CI.

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **Documentation**
  - Updated the DD4hep dependency link in the prerequisites section to use a secure URL.


benjaminhuth:
Okay, so this seems to be broken as well, I would maybe just wait a bit if they fix it...

PR #4263, opened 2025-05-13
refactor: Cleanup some seeding code

andiwand:
- utilize `Acts` namespace
- cleanup includes

PR #4264, opened 2025-05-13
refactor: Move `SeedFilterConfig` to `SeedFilter.hpp`

andiwand:
This PR moves the `SeedFilterConfig` to `SeedFilter.hpp`. Algorithm configs are tightly coupled to the algorithm and are part of the algorithm header by conversion. Usually they are a sub-struct within the algorithm class but since there is a template on the `SeedFinder` it is a bit inconvenient.

blocked by
- https://github.com/acts-project/acts/pull/4263


andiwand:
Will be handled differently

PR #4266, opened 2025-05-13
feat: Set geometryIdentifier for portals in the shell construction and add check in the blueprint

dimitra97:
This PR adds the portal Ids for the portals when the shell is being constructed for the particular volume and during the blueprint construction it checks if the portals have geoIds set or not.
This allows us to set manually the portals ids from the particular volume's geoId before the construction of the blueprint is happening - which we do for the Atlas Muon Spectrometer

### DIFF
--- a/Core/src/Geometry/Blueprint.cpp
+++ b/Core/src/Geometry/Blueprint.cpp
@@ -267,6 +267,9 @@ std::unique_ptr<TrackingGeometry> Blueprint::construct(
           ACTS_ERROR(prefix() << "No current volume found");
           throw std::logic_error("No current volume found");
         }
+        if (portal.surface().geometryId() != GeometryIdentifier{}) {
+          return;
+        }
 
         iportal += 1;
         auto id = currentVolume->geometryId().withBoundary(iportal);
--- a/Core/src/Geometry/TrapezoidPortalShell.cpp
+++ b/Core/src/Geometry/TrapezoidPortalShell.cpp
@@ -79,6 +79,8 @@ void SingleTrapezoidPortalShell::applyToVolume() {
         throw std::runtime_error{"Invalid portal found in shell at " +
                                  ss.str()};
       }
+      portal->surface().assignGeometryId(
+          m_volume->geometryId().withBoundary(i + 1));
       m_volume->addPortal(portal);
     }
   }


paulgessinger:
The goal of this is to make the portal ID assignment consistent based on the shell? Can you elaborate in which case the automatic numbering causes issues? Is it because of interference with child volumes?

dimitra97:
> The goal of this is to make the portal ID assignment consistent based on the shell? Can you elaborate in which case the automatic numbering causes issues? Is it because of interference with child volumes?

So yes, we wanted to group some trapezoidal volumes (static nodes - representing some chambers in the barrel) as children of a single cylinder (static ndoe again) and the portals of these trapezoids were assigned during the blueprint construction using the geoId of the cyllinder while we wantes to use the one from the trapezoids

paulgessinger:
@dimitra97 I see. In that case, I suppose the auto-numbering needs to be aware that it needs to check existing IDs before assigning additional ones: e.g. if there's already a pre-numbered portal with `id=1` the auto-counting can't give that portal id to any other portal, otherwise there will be clashes.

dimitra97:
> @dimitra97 I see. In that case, I suppose the auto-numbering needs to be aware that it needs to check existing IDs before assigning additional ones: e.g. if there's already a pre-numbered portal with `id=1` the auto-counting can't give that portal id to any other portal, otherwise there will be clashes.
> 
> The other possible issue I see is that the volume ID **might not be known** at that time. In this case, you might already have a valid volume ID, but this is not always true I think. To fix this, I think the auto-numbering should check if the portal ID specifically is non-zero (i.e. already set) rather than checking if the entire geo id is zero, and then reapply that fixed portal ID on top of volume id at that point in time.
> 
> Let me know if you see any issues with this!

I see your point for the possible reassignment of the same ID in the portals, but I am thinking if this is happening on top of the volumes' IDs (having a consecutive numbering of the volume's portals), should this still be an issue? Because then that would mean that also the volumes would be clashed, right?

Also for the second possible issue, I am not sure I understand. You mean it is better to check if the portals have a `boundary` geoID assigned? And if it is not the case then apply the geoId with the boundary on top of the current volume's ID? 
As I understood it the thing is that in the auto-numbering it gets the portals and assigns to them IDs before reaching their volume. So the current volume at that stage is the wrong one

Tests/UnitTests/Core/Geometry/BlueprintTests.cpp
```diff
@@ -1096,6 +1097,42 @@ BOOST_AUTO_TEST_CASE(LayerCenterOfGravity) {
...
+        trf, tbounds, "child" + std::to_string(i));
+    auto childNode = std::make_shared<StaticBlueprintNode>(std::move(childVol));
+    parentNode->addChild(std::move(childNode));
+  }
+
+  cubcontainer.addChild(std::move(parentNode));
+
+  auto trackingGeometry = root.construct({}, gctx, *logger);
+  std::cout << "tracking geometry built" << std::endl;
+  std::cin.ignore();
```
junggjo9, line 1133:
```suggestion
```

Tests/UnitTests/Core/Geometry/BlueprintTests.cpp
```diff
@@ -1096,6 +1097,42 @@ BOOST_AUTO_TEST_CASE(LayerCenterOfGravity) {
...
+    auto childVol = std::make_unique<TrackingVolume>(
+        trf, tbounds, "child" + std::to_string(i));
+    auto childNode = std::make_shared<StaticBlueprintNode>(std::move(childVol));
+    parentNode->addChild(std::move(childNode));
+  }
+
+  cubcontainer.addChild(std::move(parentNode));
+
+  auto trackingGeometry = root.construct({}, gctx, *logger);
+  std::cout << "tracking geometry built" << std::endl;
```
junggjo9, line 1132:
```suggestion
```
Or use the boost equivalent to print messages

paulgessinger:
@dimitra97 is the unit test "ready" as in: it fails a s expected?

dimitra97:
> @dimitra97 is the unit test "ready" as in: it fails a s expected?

yes by removing the assignment in the `TrapezoidPortalShell `creation and let the auto numbering it fails with 50 volumes as children (assigning portalIds for their portals + the portal Ids for the world it exceeds because it only uses the` volId = 1` as already observed). I am putting also the log

```
 VERBOSE   TryAllNavigationPolicy created for volume child
08:41:56    UnitTests      DEBUG     [Root]: Number of unused volume IDs: 255
08:41:56    UnitTests      DEBUG     [Root]: Assigning volume IDs for remaining volumes
08:41:56    UnitTests      DEBUG     [Root]: Assigning volume ID vol=1 for World
08:41:56    UnitTests      VERBOSE   [Root]: Assigning portal ID: vol=1|bnd=1
08:41:56    UnitTests      VERBOSE   [Root]: Assigning portal ID: vol=1|bnd=2
08:41:56    UnitTests      VERBOSE   [Root]: Assigning portal ID: vol=1|bnd=3
08:41:56    UnitTests      VERBOSE   [Root]: Assigning portal ID: vol=1|bnd=4
08:41:56    UnitTests      VERBOSE   [Root]: Assigning portal ID: vol=1|bnd=5
08:41:56    UnitTests      VERBOSE   [Root]: Assigning portal ID: vol=1|bnd=6
08:41:56    UnitTests      VERBOSE   [Root]: Assigning portal ID: vol=1|bnd=7
08:41:56    UnitTests      VERBOSE   [Root]: Assigning portal ID: vol=1|bnd=8
08:41:56    UnitTests      VERBOSE   [Root]: Assigning portal ID: vol=1|bnd=9
08:41:56    UnitTests      VERBOSE   [Root]: Assigning portal ID: vol=1|bnd=10
08:41:56    UnitTests      VERBOSE   [Root]: Assigning portal ID: vol=1|bnd=11
08:41:56    UnitTests      VERBOSE   [Root]: Assigning portal ID: vol=1|bnd=12
08:41:56    UnitTests      DEBUG     [Root]: Assigning volume ID vol=2 for child
08:41:56    UnitTests      VERBOSE   Checking identifier for volume(World): vol=1
08:41:56    UnitTests      VERBOSE   Inserted volume(World) ID: vol=1 pointing at 0x20f4540
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=1
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=1 pointing at 0x20e5608
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=1
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=1 pointing at 0x20e5608
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=2
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=2 pointing at 0x20ebd28
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=2
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=2 pointing at 0x20ebd28
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=3
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=3 pointing at 0x20cf2a8
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=3
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=3 pointing at 0x20cf2a8
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=4
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=4 pointing at 0x20eb588
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=4
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=4 pointing at 0x20eb588
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=5
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=5 pointing at 0x20f3e68
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=5
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=5 pointing at 0x20f3e68
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=6
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=6 pointing at 0x20e4cd8
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=6
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=6 pointing at 0x20e4cd8
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=7
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=7 pointing at 0x20e4e28
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=7
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=7 pointing at 0x20e4e28
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=8
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=8 pointing at 0x20e4108
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=8
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=8 pointing at 0x20e4108
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=9
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=9 pointing at 0x20e37d8
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=9
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=9 pointing at 0x20e37d8
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=10
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=10 pointing at 0x20e3928
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=10
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=10 pointing at 0x20e3928
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=11
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=11 pointing at 0x20e3a78
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=11
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=11 pointing at 0x20e3a78
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=12
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=12 pointing at 0x20e3bc8
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=12
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=12 pointing at 0x20e3bc8
08:41:56    UnitTests      VERBOSE   Checking identifier for volume(child): vol=2
08:41:56    UnitTests      VERBOSE   Inserted volume(child) ID: vol=2 pointing at 0x20eaee0
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=7
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=7 pointing at 0x20e4e28
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=7
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=7 pointing at 0x20e4e28
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=8
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=8 pointing at 0x20e4108
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=8
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=8 pointing at 0x20e4108
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=9
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=9 pointing at 0x20e37d8
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=9
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=9 pointing at 0x20e37d8
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=10
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=10 pointing at 0x20e3928
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=10
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=10 pointing at 0x20e3928
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=11
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=11 pointing at 0x20e3a78
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=11
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=11 pointing at 0x20e3a78
08:41:56    UnitTests      VERBOSE   Checking identifier for portal: vol=1|bnd=12
08:41:56    UnitTests      VERBOSE   Inserted portal ID: vol=1|bnd=12 pointing at 0x20e3bc8
08:41:56    UnitTests      VERBOSE   Checking identifier for surface: vol=1|bnd=12
08:41:56    UnitTests      VERBOSE   Inserted surface ID: vol=1|bnd=12 pointing at 0x20e3bc8
08:41:56    UnitTests      DEBUG     TrackingGeometry created with 2 volumes and 0 sensitive surfaces
08:41:56    UnitTests      INFO      [Root]: Building tracking geometry from blueprint tree
08:41:56    UnitTests      INFO      [Root]: Building tracking geometry from blueprint tree
08:41:56    UnitTests      INFO      [Root]: Building tracking geometry from blueprint tree
08:41:56    UnitTests      INFO      [Root]: Building tracking geometry from blueprint tree
08:41:56    UnitTests      INFO      [Root]: Building tracking geometry from blueprint tree
unknown location(0): [4;31;49mfatal error: in "Geometry/BlueprintNodeTest/GeometryIdnetifiersForPortals": std::invalid_argument: Value 256 exceeds maximum value 255 for this field[0;39;49m
/home/damperia/ACTS_Hackathon/acts/Tests/UnitTests/Core/Geometry/BlueprintTests.cpp(1100): [1;36;49mlast checkpoint: "GeometryIdnetifiersForPortals" test entry[0;39;49m
```

Core/include/Acts/Geometry/TrackingGeometryVisitor.hpp
```diff
@@ -20,15 +20,39 @@ class Layer;
...
+      : m_visitInDepth(visitInDepth) {}
+
+  /// @brief indicate the order of visiting
+  /// @note default is outermost --> innermost volume visiting
+  bool visitInDepth() { return m_visitInDepth; }
+
+ private:
+  /// Flag to indicate if the visitor should follow from the outermost to the
+  /// innermost volume depth
+  bool m_visitInDepth;
```
junggjo9:
```suggestion
  bool m_visitInDepth{true};
```

Core/include/Acts/Geometry/TrackingGeometryVisitor.hpp
```diff
@@ -20,15 +20,39 @@ class Layer;
 
 template <typename T>
 class BoundarySurfaceT;
+/// @brief Base class for tracking geometry visitors
+///
+/// This class provides a common interface for both const and mutable visitors
+/// to the tracking geometry hierarchy. It allows decide on the visiting order
```
junggjo9, line 26: Maybe a short sentencewhat vistiInDepth means ;)

Core/include/Acts/Geometry/TrackingGeometryVisitor.hpp
```diff
@@ -64,8 +88,12 @@ class TrackingGeometryVisitor {
 /// This visitor allows for non-const access to traverse and modify the
 /// tracking geometry components. It's used for operations like geometry
 /// construction, material decoration, or geometry ID assignment.
-class TrackingGeometryMutableVisitor {
+class TrackingGeometryMutableVisitor : public ITrackingGeometryVisitor {
  public:
+  /// @brief Constructor
+  explicit TrackingGeometryMutableVisitor(bool visitInDepth = true)
+      : ITrackingGeometryVisitor(visitInDepth) {}
```
junggjo9: dunno whether this is complaint with the coding guidelines in acts but I would just do
```` 
using ITrackingGeometryVisitor::ITrackingGeometryVisitor;
```
if the constructor does not do anything on top of the parent one

Core/src/Geometry/Blueprint.cpp
```diff
@@ -28,6 +28,70 @@ const std::string s_rootName = "Root";
...
+        m_logger(logger) {}
+
+  void visitVolume(TrackingVolume &volume) override {
+    m_iportal = 0;
+    m_isensitive = 0;
+
+    auto id = volume.geometryId();
+
+    if (id == GeometryIdentifier{}) {
+      auto it = std::ranges::find(m_volumesById, nullptr);
```
junggjo9, line 50: Why using a fat array if the job can be also done with a simple maximum counter?

Core/src/Geometry/Blueprint.cpp
```diff
@@ -28,6 +28,70 @@ const std::string s_rootName = "Root";
...
+      const Logger &logger,
+      std::array<const TrackingVolume *, GeometryIdentifier::getMaxVolume()>
+          &volumesById)
+      : TrackingGeometryMutableVisitor(false),
+        m_volumesById(volumesById),
+        m_logger(logger) {}
+
+  void visitVolume(TrackingVolume &volume) override {
+    m_iportal = 0;
+    m_isensitive = 0;
```
junggjo9: I think these can be function local instead of class members, no?

Core/src/Geometry/Blueprint.cpp
```diff
@@ -28,6 +28,68 @@ const std::string s_rootName = "Root";
...
+        m_logger(logger) {}
+
+  void visitVolume(TrackingVolume &volume) override {
+    GeometryIdentifier::Value iportal = 0;
+    GeometryIdentifier::Value isensitive = 0;
+
+    auto id = volume.geometryId();
+
+    if (id == GeometryIdentifier{}) {
+      auto it = std::ranges::find(m_volumesIds, GeometryIdentifier{});
```
junggjo9: Confusione.  The VolumeIds are set in the constructor and we're using the visitor to assign new volume Ids to the freshly made volumes.  In this line you're asking for the next unused ID in the sequence of VolumeIds and if nothing is available, the code just shows the finger and that's it. So the only thing what the user can do by settting the volumeIds externally is to restrict the range of allowed volumeIds. Are you giving him the means into his hands in order to shoot into his own foot?

Core/src/Geometry/Blueprint.cpp
```diff
@@ -28,6 +28,68 @@ const std::string s_rootName = "Root";
...
+      isensitive += 1;
+      auto surfaceId = id.withSensitive(isensitive);
+      ACTS_DEBUG("Assigning surface ID: " << surfaceId);
+      surface.assignGeometryId(surfaceId);
+    }
+  }
+
+ private:
+  std::array<GeometryIdentifier, GeometryIdentifier::getMaxVolume()>
+      &m_volumesIds;
```
junggjo9: Really just a reference?

Core/src/Geometry/Blueprint.cpp
```diff
@@ -231,63 +293,22 @@ std::unique_ptr<TrackingGeometry> Blueprint::construct(
...
-        surface.assignGeometryId(id);
-      }});
+  std::array<GeometryIdentifier, GeometryIdentifier::getMaxVolume()>
+      usedGeoIds{};
+  usedGeoIds.fill(GeometryIdentifier{});
+
+  std::ranges::for_each(volumesById,
+                        [&usedGeoIds](const TrackingVolume *volume) {
+                          if (volume != nullptr) {
+                            auto id = volume->geometryId();
```
junggjo9: What's the other mechanism which assigns the Id to the volume?

Core/src/Geometry/TrackingVolume.cpp
```diff
@@ -715,12 +715,22 @@ void TrackingVolume::apply(TrackingGeometryVisitor& visitor) const {
...
 
 void Acts::TrackingVolume::apply(TrackingGeometryMutableVisitor& visitor) {
+  // if the visitor is configured for inner--->outer volume visiting we visit
+  // the children first
+  if (!visitor.visitInDepth()) {
+    for (auto& volume : volumes()) {
+      volume.apply(visitor);
+    }
+  }
+
```
junggjo9, line 725: Shouldn't line 786ff be protected by the contrary statement?

dimitra97: yes could be in order to avoid another loop over the volumes, but it won't do anything actually

Core/src/Geometry/Blueprint.cpp
dimitra97: I am not sure I get the point but if the volumeIds are set externally, the visitor will try to assign volumeIds for the volumes that they do not have the ids assigned by making sure that the already set ids wont be used again because we cannot have duplicated Ids.

Core/src/Geometry/TrackingVolume.cpp
junggjo9: That depends on the visitor. If you've one moving bikini bottom somewhere else. It's moved twice

Core/src/Geometry/Blueprint.cpp
junggjo9: If that's an option by the API, okay then the array makes sense and the rest of the comments obsoletes.

Core/include/Acts/Geometry/TrackingGeometryVisitor.hpp
```diff
@@ -20,16 +20,41 @@ class Layer;
...
+/// This class provides a common interface for both const and mutable visitors
+/// to the tracking geometry hierarchy. It allows decide on the visiting order
+/// based on the visitInDepth flag. If true, the visiting happens from the
+/// outermost volume and goes deeper to the volumes into the hierarchy.
+class ITrackingGeometryVisitor {
+ public:
+  virtual ~ITrackingGeometryVisitor() = default;
+
+  explicit ITrackingGeometryVisitor(bool visitInDepth = true)
+      : m_visitInDepth(visitInDepth) {}
```
paulgessinger:
```suggestion
      : m_visitDepthFirst(visitDepthFirst) {}
```

I would personally prefer `visitDepthFirst` because as far as I know this is an established term.

Core/include/Acts/Geometry/TrackingGeometryVisitor.hpp
```diff
@@ -20,16 +20,41 @@ class Layer;
...
 class BoundarySurfaceT;
+/// @brief Base class for tracking geometry visitors
+///
+/// This class provides a common interface for both const and mutable visitors
+/// to the tracking geometry hierarchy. It allows decide on the visiting order
+/// based on the visitInDepth flag. If true, the visiting happens from the
+/// outermost volume and goes deeper to the volumes into the hierarchy.
+class ITrackingGeometryVisitor {
+ public:
+  virtual ~ITrackingGeometryVisitor() = default;
```
paulgessinger, line 31:
```suggestion
  virtual ~ITrackingGeometryVisitor() = 0;
```

If you default this, it's not an abstract class and you can instantiate it. If you make the destructor pure virtual, this changes (which I think is what we want).

Core/include/Acts/Geometry/TrackingGeometryVisitor.hpp
```diff
@@ -20,16 +20,41 @@ class Layer;
...
+
+  explicit ITrackingGeometryVisitor(bool visitInDepth = true)
```
paulgessinger: Maybe the default should be `false`, because that's the original order?

Core/src/Geometry/Blueprint.cpp
```diff
@@ -231,63 +293,22 @@ std::unique_ptr<TrackingGeometry> Blueprint::construct(
...
+                            if (id != GeometryIdentifier{}) {
+                              usedGeoIds[id.volume() - 1] = id;
```
paulgessinger:
```suggestion
                              usedGeoIds.at(id.volume() - 1) = id;
```

Core/src/Geometry/Blueprint.cpp
```diff
@@ -28,6 +28,68 @@ const std::string s_rootName = "Root";
...
+    GeometryIdentifier::Value isensitive = 0;
+
+    auto id = volume.geometryId();
+
+    if (id == GeometryIdentifier{}) {
+      auto it = std::ranges::find(m_volumesIds, GeometryIdentifier{});
+      if (it == m_volumesIds.end()) {
+        ACTS_ERROR("No free volume IDs left, all " << m_volumesIds.size()
+                                                   << " are used");
+        throw std::logic_error("No free volume IDs left");
```
paulgessinger, line 55:
```suggestion
        // @TODO: Maybe link to documentation about this
        throw std::logic_error("No free volume IDs left");
```

Core/src/Geometry/Blueprint.cpp
```diff
@@ -231,63 +293,22 @@ std::unique_ptr<TrackingGeometry> Blueprint::construct(
...
-        auto id = currentVolume->geometryId().withSensitive(isensitive);
-        ACTS_VERBOSE(prefix() << "Assigning surface ID: " << id);
-        surface.assignGeometryId(id);
-      }});
+  std::array<GeometryIdentifier, GeometryIdentifier::getMaxVolume()>
+      usedGeoIds{};
+  usedGeoIds.fill(GeometryIdentifier{});
+
+  std::ranges::for_each(volumesById,
+                        [&usedGeoIds](const TrackingVolume *volume) {
```
paulgessinger: Nitpick: in this case, I think a regular for loop is more readable than `for_each`.

Core/src/Geometry/Blueprint.cpp
```diff
@@ -231,63 +293,22 @@ std::unique_ptr<TrackingGeometry> Blueprint::construct(
...
-          return;
-        }
-
-        isensitive += 1;
-        auto id = currentVolume->geometryId().withSensitive(isensitive);
-        ACTS_VERBOSE(prefix() << "Assigning surface ID: " << id);
-        surface.assignGeometryId(id);
-      }});
+  std::array<GeometryIdentifier, GeometryIdentifier::getMaxVolume()>
+      usedGeoIds{};
```
paulgessinger: Would this not still work with the existing `volumesById`? What I was doing before was looking for the first entry with a `nullptr` set, where now you're looking for the first entry with a `0` geometry id?

dimitra97: I changed it to the previous logic, using the tracking volume array

Core/include/Acts/Geometry/TrackingGeometryVisitor.hpp
```diff
@@ -64,9 +89,12 @@ class TrackingGeometryVisitor {
...
 /// tracking geometry components. It's used for operations like geometry
 /// construction, material decoration, or geometry ID assignment.
-class TrackingGeometryMutableVisitor {
+class TrackingGeometryMutableVisitor : public ITrackingGeometryVisitor {
  public:
-  virtual ~TrackingGeometryMutableVisitor();
+  /// @brief Constructor
+  using ITrackingGeometryVisitor::ITrackingGeometryVisitor;
+
+  ~TrackingGeometryMutableVisitor() override = 0;
```
paulgessinger:
```suggestion
  ~TrackingGeometryMutableVisitor() override;
```

Core/include/Acts/Geometry/TrackingGeometryVisitor.hpp
```diff
@@ -20,16 +20,41 @@ class Layer;
...
 /// visualization, validation, or collecting information about the geometry
 /// structure.
-class TrackingGeometryVisitor {
+class TrackingGeometryVisitor : public ITrackingGeometryVisitor {
  public:
-  virtual ~TrackingGeometryVisitor() = 0;
+  /// @brief Constructor from base class
+  using ITrackingGeometryVisitor::ITrackingGeometryVisitor;
+
+  ~TrackingGeometryVisitor() override = 0;
```
paulgessinger:
```suggestion
  ~TrackingGeometryVisitor() override;
```

Core/src/Geometry/TrackingGeometryVisitor.cpp
```diff
@@ -12,6 +12,7 @@ namespace Acts {
 
 TrackingGeometryVisitor::~TrackingGeometryVisitor() = default;
 TrackingGeometryMutableVisitor::~TrackingGeometryMutableVisitor() = default;
+ITrackingGeometryVisitor::~ITrackingGeometryVisitor() = default;
```
paulgessinger:
```suggestion
```

paulgessinger (commented):
Thanks @dimitra97, I have one more point on the destructor: I think we'd want to leave the `ITrackingGeometryVisitor::~ITrackingGeometryVisitor` destructor pure virtual, and keep the ones from the derived classes implemented and overriden.

The reason is that I would prefer if 

```cpp
ITrackingGeometryVisitor visitor; // is a compilation error
```

dimitra97:
> Thanks @dimitra97, I have one more point on the destructor: I think we'd want to leave the `ITrackingGeometryVisitor::~ITrackingGeometryVisitor` destructor pure virtual, and keep the ones from the derived classes implemented and overriden.
> 
> The reason is that I would prefer if
> 
> ```c++
> ITrackingGeometryVisitor visitor; // is a compilation error
> ```

ok after also discussing it offline, maybe let's keep it like that for now because I am getting linkin errors otherwise. 
@paulgessinger  Let me know if you have any other objections/ comments. Thanks for looking into it :)

PR #4268, opened 2025-05-14
refactor: Improved binding of LayerCreator

benjaminhuth:
* Constructor with log level
* named arguments

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **New Features**
  - Enhanced Python interface for geometry layer creation, allowing configuration of logging level during object creation.
  - Improved Python method signatures for layer creation, now supporting named parameters for easier and clearer usage.


Core/include/Acts/Geometry/LayerCreator.hpp
```diff
@@ -65,6 +65,8 @@ class LayerCreator {
                         std::unique_ptr<const Logger> logger =
                             getDefaultLogger("LayerCreator", Logging::INFO));
 
+  LayerCreator(LayerCreator&&) = default;
+
   /// Destructor
   ~LayerCreator() = default;
```
andiwand, line 71: alternatively you can just drop the explicit default destructor. then the compiler will create move constructor and move assignment by itself. right now you would be missing the move assignment

benjaminhuth: You mean dropping the `explicit`?

andiwand:
```suggestion
```

I think this way you should get implicit move constructor and assignment

benjaminhuth: A I misread de/constructor,  that makes sense

andiwand (dismissed):

paulgessinger:
@benjaminhuth can you fix clang-format?

Examples/Python/src/GeometryBuildingGen1.cpp
```diff
@@ -36,20 +36,29 @@ void addGeometryBuildingGen1(Context &ctx) {
...
+                "gctx"_a, "surfaces"_a, "binsPhi"_a, "binsZ"_a)
+            .def(
+                "discLayer",
+                [](const Acts::LayerCreator &self, const GeometryContext &gctx,
+                   SurfacePtrVector surfaces, std::size_t binsR,
+                   std::size_t binsPhi) {
+                  return self.discLayer(gctx, std::move(surfaces), binsR,
+                                        binsPhi);
+                },
+                "gctx"_a, "surfaces"_a, "binsR"_a, "binsPhi"_a);
```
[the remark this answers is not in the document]
benjaminhuth: Hmm there are multiple overloads of those...

paulgessinger (approved):

PR #4283, opened 2025-05-20
fix: Make sure CalibrationContext gets copied to actor.

scott-snyder:
In the overload of KalmanFitter::fit taking a vector of Surface*, make sure that CalibrationContext (and freeToBoundCorrection) gets copied from the options to the actor.  This was done in the previous overload but was apparently forgotten here.

If we don't do this, we'll end up initializing a reference to the CalibrationContext, dereferencing the null pointer.

While we were apparently never actually doing anything with this null reference (and thus not crashing), this does provoke a warning from the undefined behavior sanitizer in debug builds, which causes test failures in FPGATrackSimConfTools:

```
/afs/cern.ch/work/s/ssnyder/acts/acts/Core/include/Acts/TrackFitting/KalmanFitter.hpp:616:14: runtime error: reference binding to null pointer of type 'const struct CalibrationContext'
...
```

However, this has generally been hidden by the fact that in the CI in dbg builds, the tests tend to time out before they get to this point (and there are also other assertion failures ahead of it...).

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved configuration handling in the track fitting process to ensure additional options are correctly applied during fitting.


asalzburger (approved):
Good catch, indeed.

andiwand:
@scott-snyder you received an invite which allows the `CI Bridge` to run - can you accept it so we can merge this?

PR #4284, opened 2025-05-21
feat: Add few new math helper functions

junggjo9:
- Introduce new helper functions to calculate factorial or the sum of N integers
- Introduce binomial coefficient calculation function
- Introduce new helper function to calculate a direction vector from the tangents of the x to the z & from the tangent of the y to the z axis
- Implementation of the Legendre & Chebychev polynomials avoiding the useage of the recursion formulas. Other implementations, I've seen make explicit use of the recursion formula which is kind of slow if higher order polynomial order functions are evaluated frequently. This implementation calcualtes the coefficients of the first 10 orders & their associated derivatives at compile time.
- Introduce helper functions to intersect lines & planes, also to calculate the signed distance between two lines

--- END COMMIT MESSAGE ---

@dimitra97


Core/include/Acts/Utilities/UnitVectors.hpp
```diff
@@ -46,23 +47,32 @@ inline Eigen::Matrix<T, 3, 1> makeDirectionFromPhiEta(T phi, T eta) {
...
+}
 
 /// Construct a phi and theta angle from a direction vector.
 ///
 /// @param unitDir 3D vector indicating a direction
 ///
 template <typename T>
 inline Eigen::Matrix<T, 2, 1> makePhiThetaFromDirection(
-    Eigen::Matrix<T, 3, 1> unitDir) {
-  unitDir.normalize();
```
andiwand, line 65: I agree that the `normalize` should not be done here. We have to make sure this does not break anything

Core/include/Acts/Utilities/UnitVectors.hpp
```diff
@@ -46,23 +47,32 @@ inline Eigen::Matrix<T, 3, 1> makeDirectionFromPhiEta(T phi, T eta) {
...
+      csPhi.cs * csTheta.sn,
+      csPhi.sn * csTheta.sn,
+      csTheta.cs,
   };
 }
+/// @brief Construct a normalized direction vector from the tangents of the
+///        x-axis to the z-axis and of the y-axis to the z-axis
+///
+/// @param tanAlpha: Tangent of the x-axis to the z-axis
+/// @param tanBeta: Tangent of the y-axis to the z-axis
```
andiwand, line 62:
```suggestion
/// @param tanAlpha Tangent of the x-axis to the z-axis
/// @param tanBeta Tangent of the y-axis to the z-axis
```

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -27,4 +27,26 @@ constexpr auto fastHypot(T... args) {
...
 
+/// @brief Helper struct to calculate sin/cos in one go
+template <typename T>
+struct sincos {
+  /// @brief Standard constructor taking an angle
+  /// @param alpha: Angle to consider
+  sincos(const T alpha) : cs{std::cos(alpha)}, sn{std::sin(alpha)} {}
+  T cs{0.};
+  T sn{0.};
+};
```
andiwand, line 38: I do not see a big benefit of this. I would see a benefit if this had a faster way to compute both values.

Also this should rather be called `SinCos` or be a free function returning a `std::pair`

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -27,4 +27,26 @@ constexpr auto fastHypot(T... args) {
...
+
+
+/// @brief Calculates the sum of 1 + 2 + 3+ ... + N using the
+///        Gaussian sum formula
+/// @param N: Upper limit of the sum
+template<typename T,
+        typename std::enable_if_t<std::is_integral_v<T>, bool> = true>
+constexpr T sumUp (const T N) {
```
andiwand, line 46:
```suggestion
constexpr T prefixSum(const T N) {
```

Core/include/Acts/Utilities/ArrayHelpers.h
```diff
@@ -0,0 +1,21 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include <array>
+
+namespace Acts{
+    /** @brief creates an array of type T and size N and assigns all elements to the parsed 
+     *         default value def_val.
+     *         This basically allows for inline construction and initialization of class-member arrays
+     *  @param def_val: Default value to assign */
+    template<typename T, std::size_t N> constexpr std::array<T, N> make_array(const T& def_val){
```
andiwand, line 16:
```suggestion
    template<typename T, std::size_t N> constexpr std::array<T, N> filledArray(const T& def_val){
```

Core/include/Acts/Surfaces/detail/LineHelper.hpp
```diff
@@ -0,0 +1,112 @@
...
+      lineSurfTrf2.translation(), lineSurfTrf2.linear().col(2));
+}
+/** @brief Intersect a line in 3D space with a plane represented by the Hesse-Normal form
+ *  @param linePos: Arbitrary point on the line to intersect
+ *  @param lineDir: Direction of the line to intersect
+ *  @param planeNorm: Normal vector of the plane
+ *  @param offSet: Offset to move the plane along the normal vector */
+inline Intersection3D planeIntersect(const Vector3& linePos,
+                                     const Vector3& lineDir,
+                                     const Vector3& planeNorm,
```
andiwand, line 70: This definitely exists already in ACTS

https://github.com/acts-project/acts/blob/1cfe965fe0889070ae34c61075ce9e82b3928f39/Core/include/Acts/Surfaces/detail/PlanarHelper.hpp#L17-L26

Core/include/Acts/Surfaces/detail/LineHelper.hpp
```diff
@@ -0,0 +1,112 @@
...
+ *         closest approach of the second line to the first line.
+ *  @param linePosA: Arbitrary point on the first line
+ *  @param lineDirA: Direction of the first line (Unit-length)
+ *  @param linePosB: Arbitrary point on the second line
+ *  @param lineDirB: Direction of the second line (Unit-length) */
+template <unsigned N>
+inline Intersection<N> lineIntersect(const ActsVector<N>& linePosA,
+                                     const ActsVector<N>& lineDirA,
+                                     const ActsVector<N>& linePosB,
+                                     const ActsVector<N>& lineDirB) {
```
andiwand, line 27: I am not sure if we have this already. Potentially not as general

I think Eigen types should be generally received with `const Eigen::MatrixBase<T> &`

Core/include/Acts/Utilities/MathHelpers.hpp
junggjo9: I copy pasted the simple part from athena (https://gitlab.cern.ch/atlas/athena/-/blob/main/Control/CxxUtils/CxxUtils/sincos.h). There's more fancy stuff in the preprocessor macro which I don't know is worth to do as well. But sure, I see the point ;)

Core/include/Acts/Surfaces/detail/LineHelper.hpp
junggjo9: You mean templating it twice?

Core/include/Acts/Surfaces/detail/LineHelper.hpp
junggjo9: yeah I saw that but I don't remember why I choose to propose the new one... Let's talk later

Core/include/Acts/Surfaces/detail/LineHelper.hpp
andiwand: in the most general case it would require 4 templates like here https://github.com/acts-project/acts/blob/1cfe965fe0889070ae34c61075ce9e82b3928f39/Tests/UnitTests/Core/Utilities/UnitVectorsTests.cpp#L171-L174

Core/include/Acts/Surfaces/detail/LineHelper.hpp
andiwand: hm at which position do you assume the origin of the plane? generally the origin and the `planeNorm` make up the transform

Core/include/Acts/Utilities/UnitVectors.hpp
benjaminhuth: I guess a `assert(std::abs(unitDir.norm() - 1) < 1e-8);` or so would do it, right?

Core/include/Acts/Utilities/MathHelpers.hpp
benjaminhuth: Just out of interest: Could `std::sqrt(1 - sin(x)*sin(x))` be cheaper then evaluating sin and cos?

Core/include/Acts/Utilities/UnitVectors.hpp
junggjo9: But it does not need to be normalized in order to calculate the angles

benjaminhuth: Ah I think you're right. But then we should change the name of the parameter.

Core/include/Acts/Surfaces/detail/LineHelper.hpp
```diff
@@ -0,0 +1,113 @@
...
+/// @brief Calculates the signed distance between two lines in 3D space
+/// @param linePosA: Arbitrary point on the first line
+/// @param lineDirA: Direction of the first line (Unit-length)
+/// @param linePosB: Arbitrary point on the second line
+/// @param lineDirB: Direction of the second line (Unit-length)
+inline double signedDistance(const Vector3& linePosA, const Vector3& lineDirA,
+                             const Vector3& linePosB, const Vector3& lineDirB) {
+  /// Project the first direction onto the second & renormalize to a unit vector
+  const double dirDots = lineDirA.dot(lineDirB);
+  const Vector3 AminusB = linePosA - linePosB;
```
andiwand:
```suggestion
  const Vector3 aMinusB = linePosA - linePosB;
```

Core/include/Acts/Utilities/ArrayHelpers.hpp
```diff
@@ -0,0 +1,24 @@
...
+#pragma once
+
+#include <array>
+
+namespace Acts {
+///  @brief creates an array of type T and size N and assigns all elements to the parsed
+///         default value def_val. This basically allows for inline construction
+///         and initialization of class-member arrays
+///  @param def_val: Default value to assign
+template <typename T, std::size_t N>
```
andiwand, line 18: would allow type deduction of `T` but ordering is then inconsistent with `std::array`
```suggestion
template <std::size_t N, typename T>
```

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -12,9 +12,43 @@
 
 namespace Acts {
 
+/// @brief Calculates the ordinary power of the number x.
+/// @param x: Number to take the power from
+/// @param p: Power to take
+template <typename T, typename P>
+constexpr T pow(T x, P p)
+  requires(std::is_integral_v<P>)
```
andiwand:
```suggestion
/// @brief Calculates the ordinary power of the number x.
/// @param x: Number to take the power from
/// @param p: Power to take
template <typename T, typename P>
constexpr T pow(T x, P p)
  requires(std::is_integral_v<P>)
```
- I think `typename P` can be replaced by something like `std::integral P`? then the `requires` can be dropped
- for integer exponents this is usually done with https://en.wikipedia.org/wiki/Exponentiation_by_squaring - not sure if relevant

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -12,9 +12,43 @@
...
+  using unsigned_p = std::make_unsigned_t<P>;
+  constexpr unsigned_p four = 4;
+  return pow(x, four) * pow(x, static_cast<unsigned_p>(p) - four);
+}
+
+/// @brief Returns the square of the passed number
 template <typename T>
 constexpr auto square(T x) {
-  return x * x;
+  return pow(x, 2u);
```
andiwand: IMO this is easy enough to be left as `x * x` - no need to call another function

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -27,4 +61,35 @@ constexpr auto fastHypot(T... args) {
   return std::sqrt(hypotSquare(args...));
 }
 
+/// @brief Calculates the sum of 1 + 2 + 3+ ... + N using the
+///        Gaussian sum formula
+/// @param N: Number until which the sum runs
+template <typename T>
+  requires(std::is_integral_v<T>)
```
andiwand: same as above

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -27,4 +61,35 @@ constexpr auto fastHypot(T... args) {
...
+constexpr T sumUp(const T N) {
```
andiwand: `sumUp` sounds very generic - maybe `integrerPrefixSum`?

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -27,4 +61,35 @@ constexpr auto fastHypot(T... args) {
...
+  return N * (N + 1) / 2;
+}
+/// @brief Calculates the factorial of a number
+///        N!= N*(N-1)....*3*2*1
+/// @param upperN: Upper factor until which the factorial is calculated
+/// @param lowerN: Optional argument to remove the first factors from the calculation
+template <typename T>
+  requires(std::is_integral_v<T>)
```
andiwand: same as above

Core/include/Acts/Utilities/detail/Polynomials.hpp
```diff
@@ -0,0 +1,263 @@
...
+  for (unsigned k = 0u; 2u * k + d <= n; ++k) {
+    result += coeffUn(n, k) * pow(x, n - 2u * k - d) *
+              (d > 0u ? factorial(n - 2u * k, n - 2u * k - d + 1u) : 1u);
+  }
+  return result;
+}
+}  // namespace Chebychev
+
+/// @brief Helper macros to setup the evaluation of the n-th orthogonal
+///        polynomial and of its derivatives. The  coefficients of the
```
andiwand, line 196:
```suggestion
///        polynomial and of its derivatives. The coefficients of the
```

Core/include/Acts/Utilities/detail/Polynomials.hpp
```diff
@@ -0,0 +1,263 @@
...
+///        of the n-th polynomial and of the first two derivatives
+///        are precalculated at compile time.  For higher order derivatives
```
andiwand:
```suggestion
///        are precalculated at compile time. For higher order derivatives
```

Core/include/Acts/Utilities/detail/Polynomials.hpp
```diff
@@ -0,0 +1,263 @@
...
+    result += coeffUn(n, k) * pow(x, n - 2u * k - d) *
+              (d > 0u ? factorial(n - 2u * k, n - 2u * k - d + 1u) : 1u);
+  }
+  return result;
+}
+}  // namespace Chebychev
+
+/// @brief Helper macros to setup the evaluation of the n-th orthogonal
+///        polynomial and of its derivatives. The  coefficients of the
+///        of the n-th polynomial and of the first two derivatives
```
andiwand:
```suggestion
///        n-th polynomial and of the first two derivatives
```

andiwand (commented):
couple of comments from my side

some of this is going in `Utilities` and increases API - would be good to get more eyes on this - pinging @benjaminhuth @paulgessinger

Core/include/Acts/Utilities/MathHelpers.hpp
junggjo9: Indeed

Core/include/Acts/Utilities/MathHelpers.hpp
junggjo9: Changed

Core/include/Acts/Utilities/MathHelpers.hpp
junggjo9: Changed

Core/include/Acts/Utilities/MathHelpers.hpp
junggjo9: Nice feature of the concepts

Core/include/Acts/Utilities/UnitVectors.hpp
```diff
@@ -9,6 +9,7 @@
 #pragma once
 
 #include "Acts/Definitions/Algebra.hpp"
+#include "Acts/Utilities/MathHelpers.hpp"
```
junggjo9:
```suggestion
```

Core/include/Acts/Surfaces/detail/LineHelper.hpp
```diff
@@ -0,0 +1,111 @@
...
+///        first line.
+/// @param linePosA: Arbitrary point on the first line
+/// @param lineDirA: Direction of the first line (Unit-length)
+/// @param linePosB: Arbitrary point on the second line
+/// @param lineDirB: Direction of the second line (Unit-length)
+template <int N>
+inline Intersection<N> lineIntersect(const ActsVector<N>& linePosA,
+                                     const ActsVector<N>& lineDirA,
+                                     const ActsVector<N>& linePosB,
+                                     const ActsVector<N>& lineDirB) {
```
paulgessinger:
```suggestion
                                     const ActsVector<N>& lineDirB) requires(N>=2) {
```

Core/include/Acts/Surfaces/detail/LineHelper.hpp
```diff
@@ -0,0 +1,111 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+#include "Acts/Definitions/Algebra.hpp"
+#include "Acts/Utilities/Intersection.hpp"
+#include "Acts/Utilities/MathHelpers.hpp"
+
+namespace Acts::LineHelper {
```
andiwand:
```suggestion
namespace Acts::detail::LineHelper {
```

Core/include/Acts/Utilities/ArrayHelpers.hpp
```diff
@@ -0,0 +1,24 @@
...
+constexpr std::array<T, N> filledArray(const T& def_val) {
+  std::array<T, N> arr{};
+  arr.fill(def_val);
+  return arr;
+}
```
paulgessinger:
```suggestion
constexpr std::array<T, N> filledArray(const T& defVal) {
  std::array<T, N> arr{};
  arr.fill(defVal);
  return arr;
}
```

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -12,6 +12,50 @@
 
 namespace Acts {
 
+/// @brief Returns the absolute of a number
+///        (Can be removed for c++ 23)
+template <typename T>
+constexpr T abs(const T n) {
```
paulgessinger, line 18: I'm not sure I understand this one. What does this add over `std::abs`? Are you expecting to use this with types that are not supported in `std::abs`?

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -12,6 +12,50 @@
...
+      return -n;
+    }
+  }
+  return n;
+}
+
+/// @brief Calculates the ordinary power of the number x.
+/// @param x: Number to take the power from
+/// @param p: Power to take
+template <typename T, std::integral P>
```
paulgessinger, line 30: Is the goal for this to be constexpr and? If the exponents aren't to large why not a simpler recursive function like this:

```cpp
template <typename T, std::integral P>
constexpr T pow(T base, P exp) {
    return exp == 0 ? 1 : base * pow_constexpr(base, exp - 1);
}
```

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -27,4 +71,33 @@ constexpr auto fastHypot(T... args) {
   return std::sqrt(hypotSquare(args...));
 }
 
+/// @brief Calculates the sum of 1 + 2 + 3+ ... + N using the
+///        Gaussian sum formula
+/// @param N: Number until which the sum runs
+template <std::integral T>
+constexpr T sumAllNumUpTo(const T N) {
```
paulgessinger: The function name is longer than the expression?

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -27,4 +71,33 @@ constexpr auto fastHypot(T... args) {
...
+  return N * (N + 1) / 2;
+}
+/// @brief Calculates the factorial of a number
+///        N!= N*(N-1)....*3*2*1
+/// @param upperN: Upper factor until which the factorial is calculated
+/// @param lowerN: Optional argument to remove the first factors from the calculation
+template <std::integral T>
+constexpr T factorial(const T upperN, const T lowerN = 1) {
```
paulgessinger, line 71: This could also have a shorter recursive version depending on how large the numbers are:

```cpp
template <typename T>
constexpr T factorial_constexpr(T n) {
    return (n <= 1) ? 1 : n * factorial_constexpr(n - 1);
}
```

although this lacks the lower N bound, although that could probably be added as well.

paulgessinger (commented):
No objections for functions added to the `detail` namespace. If the ones added to `MathHelpers.hpp` don't have immediate wider use-cases, I'd prefer to push them to detail as well.

Another consideration, again depending on how widely they would be used, is to add tham to a private header colocated with the file that uses them, although this precludes the tests (which I think are good).

I'm not sure if the intersect methods aren't at least partially available in the surface classes. If we add more fundamental helper methods like this, I think the surface classes should then be refactored to use them. I don't think we should have two separate line and plane intersection functions in the code, to avoid confusion.

Core/include/Acts/Utilities/MathHelpers.hpp
junggjo9: It's nothing else than the ordinary std::abs. I had to reimplement it because I used it in a `constexpr` function. GCC does not have any problems, but the compile for mac was doing mimimimimi, this expression is not constexpr. you cannot use it... With c++23, abs becomes constexpr https://en.cppreference.com/w/cpp/numeric/math/abs

Core/include/Acts/Utilities/MathHelpers.hpp
junggjo9: The goal is to have a constexpr pow function. Indeed that would be a much shorter solution, I thought that recursion might lead to more opened stacks and x*x or x*x*x will be faster. Honestly, it's just gut feeling why I implemented the first four powers by hand and the  rest in recursive manner

Core/include/Acts/Utilities/MathHelpers.hpp
junggjo9: my first Idea was `sumUp`.  Or maybe sumUpToN?

Core/include/Acts/Utilities/MathHelpers.hpp
junggjo9: Indeed one could do it recursively as well. I was not so sure about performance loss from the number of stacks vs. shortness.

Core/include/Acts/Surfaces/detail/LineHelper.hpp
junggjo9: done

paulgessinger:
CI broken until https://github.com/acts-project/acts/pull/4445 is in.

Core/include/Acts/Utilities/MathHelpers.hpp
paulgessinger: Fair enough. If we don't see any issues, I'd argue shorter is better, we can change if we want.

acts-project-service:
## :white_check_mark: Athena integration test results [28e24ce418039c0d1cdc76d44436f7559703c63e]

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962850)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962849)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962848)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962847)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962846)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962845)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962844)||
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962843)||
|:green_circle: | [test_ActsCheckObjectCountsWorkflowHgtd](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962842)||
|:green_circle: | [test_ActsCheckObjectCountsCachedWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962841)||
|:green_circle: | [test_ActsCheckObjectCountsLegacyWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962840)||
|:green_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962839)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962838)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962837)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962836)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962835)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962834)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962833)||
|:green_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962832)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962831)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962830)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962829)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962828)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962827)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962826)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962825)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962824)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962823)||
|:yellow_circle: | [test_ActsGx2fRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962822)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962821)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962820)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962819)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58962818)||

PR #4297, opened 2025-05-26
feat: add eic-shell CI build

veprbl:
This adds a build of Acts in eic-shell container, which then also builds EICrecon main branch against it. This is to actively detect breaking changes.

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **Chores**
  - Added a new automated build process for enhanced integration and testing of key components.


paulgessinger:
Thanks @veprbl, this is pretty nice! I'll have a closer look later today.

andiwand:
This way the CI will fail if a change is breaking no? So in principle we cannot merge any breaking changes before your main branch is already patched against the changes?

Or will it fail and be ignored by our merge bots?

.github/workflows/builds.yml
```diff
@@ -414,3 +414,67 @@ jobs:
...
+            popd
+
+            pushd EICrecon
+            export LD_LIBRARY_PATH="$CMAKE_INSTALL_PREFIX/lib:$LD_LIBRARY_PATH"
+            ccache -z && \
+            env CXXFLAGS="-I$CMAKE_INSTALL_PREFIX/include" \
+                LDFLAGS="-L$CMAKE_INSTALL_PREFIX/lib" \
+            cmake -B build -S . \
+              -GNinja \
+              -DActs_ROOT=$CMAKE_INSTALL_PREFIX
```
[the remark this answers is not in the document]
veprbl: That's why I set `CXXFLAGS` above. I expect "EDM4hep" should be ignored because of the option ordering or because its path is in `-isystem`. `Acts_ROOT` here is mostly for CMake.

veprbl:
> This way the CI will fail if a change is breaking no? So in principle we cannot merge any breaking changes before your main branch is already patched against the changes?
> 
> Or will it fail and be ignored by our merge bots?

It looks like it should trigger a requirement
https://github.com/acts-project/acts/blob/935c883e70490337cbaaf52a649439172da4b3d2/.merge-sentinel.yml#L20
but that can be avoided with a rename (might require to spare an extra yml file to go from "Build / eic-shell" to, say "External / eic-shell"). A separate question is if this produces a red x. There is no way for an GitHub Actions job to result in warning, but we could be always green and just post a commit status or a PR comment.

.github/workflows/external.yml
```diff
@@ -0,0 +1,109 @@
...
+
+      - name: Build
+        id: build
+        uses: eic/run-cvmfs-osg-eic-shell@main
+        continue-on-error: true
+        with:
+          platform-release: "eic_xl:nightly"
+          run: |
+            pushd acts
+            echo 99.9.9 > version_number
```
paulgessinger, line 57: This would default to `9.9.9` otherwise. We could actually change the default version on `main` at this point.

.github/workflows/external.yml
```diff
@@ -0,0 +1,109 @@
...
+            ccache -z && \
+            export CMAKE_INSTALL_PREFIX=$HOME/prefix
+            export CMAKE_CXX_COMPILER_LAUNCHER=ccache
+            cmake -B build -S . \
+              -GNinja \
+              -DACTS_BUILD_EXAMPLES=ON \
+              -DACTS_BUILD_FATRAS=ON \
+              -DACTS_BUILD_PLUGIN_DD4HEP=ON
+            cmake --build build
+            cmake --install build
```
paulgessinger, line 67:
```suggestion
            cmake --install build > install.log
```

.github/workflows/external.yml
```diff
@@ -0,0 +1,109 @@
...
+        if: ${{ steps.build.outcome == 'failure' && github.event_name == 'pull_request' }}
+        uses: peter-evans/create-or-update-comment@v4
+        with:
+          comment-id: ${{ steps.fc.outputs.comment-id }}
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            Downstream build failure
+
+             - eic-shell (cc @eic/epic-acts-devs)
+          edit-mode: replace
```
paulgessinger, line 102: Seems like this won't get triggered in the PR context, and I think PR jobs also won't have permission to post. 

But what we can do instead is have a workflow trigger when this build job completes using the [`workflow_run`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows?versionId=free-pro-team%40latest&productId=actions&restPage=writing-workflows%2Cchoosing-when-your-workflow-runs%2Cevents-that-trigger-workflows#workflow_run) trigger. 

That requires a few extra steps like saving the PR number as an artifact and checking the other workflows output.

veprbl: Right. I could modify report.yml to download this too and concatenate it to the report. There is an option go with `pull_request_target` too, if we don't expect the action to have changes done to it often.

.github/workflows/external.yml
veprbl: Is this for upload or we might discard output altogether?

paulgessinger: I was thinking suppression. It's a lot of output and only very rarely does it have relevant information. We could add it to an artifact so it's accessible if need be I suppose.

paulgessinger:
@veprbl should we try this again now? The conflict comes only from `version_number`, which I changed on main.

.github/workflows/builds.yml
```diff
@@ -414,3 +414,95 @@ jobs:
         run: >
           PATH="${{ env.DEPENDENCY_DIR }}/bin:$PATH"
           && ./build-downstream/bin/ShowActsVersion
+
+  external_eic-shell:
```
veprbl, line 415: I had to get rid of external.yml, because we need to have error message artifact and physmon artifact available at the same workflow_run.

.github/workflows/builds.yml
```diff
@@ -414,3 +414,95 @@ jobs:
...
+
+            ccache -s
+
+      - name: Create comment body
+        if: ${{ steps.build.outcome == 'failure' && github.event_name == 'pull_request' }}
+        run: |
+          cat >comment-50-external-failure.md <<EOF
+          # ❗️: Downstream build failure
+
+           - eic-shell (cc @veprbl)
```
veprbl: Pings would not go through to a team from eic org for me (https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#mentioning-people-and-teams). We could setup a team of responders in acts-project namespace, or just start with my name for now.

.github/workflows/report.yml
```diff
@@ -78,6 +84,15 @@ jobs:
         with:
           python-version: '3.12'
 
+      - uses: actions/download-artifact@v4
+        if: steps.dl-af.outputs.comment-artifacts-found
```
veprbl, line 88: There is no way to gracefully handle missing artifacts, we have to do a lookup to see if those are present in the first place.

.github/workflows/report.yml
```diff
@@ -91,6 +106,7 @@ jobs:
           else
             echo "🟥 summary not found!" >> comment.md
           fi
+          cat comment-*.md >> comment.md || true
```
veprbl, line 109: In the end, I'm not even relying #4361. For this version, posting any additional notices requires physmon artifact to be ready first. We could change physmon implementation to upload its comment as a separate artifact (assuming you don't need to pin it in the herald app).

.github/workflows/builds.yml
paulgessinger: I can set up a team if your prefer that. Might be nicer to manage rather than having it hard-coded in the config, but i'll defer this decision to you.

veprbl: Hi, let's add an ePIC team. Let me know what name you choose, so I can update the PR.

paulgessinger: Sorry for the delay, it's this: https://github.com/orgs/acts-project/teams/epic-contacts

veprbl:
```suggestion
           - eic-shell (cc @acts-project/epic-contacts)
```

paulgessinger (approved):
Let's go 🚀

PR #4317, opened 2025-06-03
feat: Add ROOT IO support to HepMC3

paulgessinger:
--- END COMMIT MESSAGE ---

### DIFF
--- a/Examples/Io/HepMC3/src/HepMC3Writer.cpp
+++ b/Examples/Io/HepMC3/src/HepMC3Writer.cpp
@@ -26,6 +26,10 @@
 #include <HepMC3/WriterGZ.h>
 #endif
 
+#ifdef HEPMC3_ROOT_SUPPORT
+#include <HepMC3/WriterRootTree.h>
+#endif
+
 #include <boost/algorithm/string/join.hpp>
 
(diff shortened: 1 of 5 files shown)


Examples/Io/HepMC3/src/HepMC3Writer.cpp
```diff
@@ -26,6 +26,10 @@
 #include <HepMC3/WriterGZ.h>
 #endif
 
+#ifdef HEPMC3_ROOT_SUPPORT
```
asalzburger: Shall we pre-fix this with ACTS_HEPMC3 ... to make sure it's an ACTS-local definition?

asalzburger (approved):
Approve, with one quesiton/suggestion - but more a nitpick, actually,

Examples/Io/HepMC3/src/HepMC3Writer.cpp
paulgessinger: That's a fair point. I'll change this.

paulgessinger: @asalzburger updated now, can you approve again?

PR #4318, opened 2025-06-04
perf: Optimize sort and value range checks in `SeedFinder`

andiwand:
Some small optimizations I found during working on the space point / seed EDM profiling.
- `|` is faster than `||` for value range checks as the overall result is more predictable for the CPU (`||` will cause two branches which are basically 50/50)
- sorting on $cot \theta$ is faster if we copy the values into a flat vector first as this is more optimal for the cache

## Summary by CodeRabbit

- **Performance Improvements**
  - Enhanced performance in seed finding by optimizing conditional checks for better branch prediction.
  - Improved memory efficiency by updating internal indexing to use a smaller integer type.
- **No User-Facing Changes**
  - No changes to the application's interface or visible features.


andiwand:
on my macbook with arm

image

on the threadripper

image

Core/include/Acts/Seeding/SeedFilter.ipp
```diff
@@ -55,9 +55,15 @@ void SeedFilter<external_spacepoint_t>::filterSeeds_2SpFixed(
...
+    // intentionally using `|` after profiling. faster due to better branch
+    // prediction
     seedConfRange =
-        (bottomSp.z() > m_cfg.centralSeedConfirmationRange.zMaxSeedConf ||
-         bottomSp.z() < m_cfg.centralSeedConfirmationRange.zMinSeedConf)
+        static_cast<bool>(
+            static_cast<int>(bottomSp.z() >
+                             m_cfg.centralSeedConfirmationRange.zMaxSeedConf) |
+            static_cast<int>(bottomSp.z() <
+                             m_cfg.centralSeedConfirmationRange.zMinSeedConf))
```
paulgessinger: I suppose this could be wrapped in a macro if we wanted to give this a name and improve readability, but that also has significant downsides.

benjaminhuth: I mean we don't need a macro, we can also do a lambda, right?
```c++
auto outOfRangeBinary = [](auto v, auto max, auto min) -> bool {
	return static_cast<bool>(static_cast<int>(v > max) | static_cast<int>(z < min));
};

File: Core/include/Acts/Seeding/SeedFinder.ipp [hunk 2]
```diff
@@ -255,6 +255,14 @@ SeedFinder<external_spacepoint_t, grid_t, platform_t>::getCompatibleDoublets(
   float deltaR = 0.;
   float deltaZ = 0.;
 
+  const auto outsideRangeCheck = [](const float value, const float min,
+                                    const float max) -> bool {
+    // intentionally using `|` after profiling. faster due to better branch
+    // prediction
+    return static_cast<bool>(static_cast<int>(value < min) |
```
asalzburger, line 262: Is this a pattern we use more often - or is this really bound to that case?

asalzburger (approved):
Approved, but one question about generalization to other cases.

Core/include/Acts/Seeding/SeedFinder.ipp
andiwand: I believe this only makes sense in very hot paths in the code and after confirmed by profiling

PR #4322, opened 2025-06-05
feat: Count and print stats for `SeedFinder`

andiwand:
Adds stat counting for the explored doublet and triplet combinatorics and cuts counters for the triplet combinatorics. This is useful to investigate the performance of the triplet seeding.

I do not expect any meaningful runtime performance impacts, but will measure to be sure.


andiwand:
to my surprise, I measure a 3% CPU impact by the counters in the tight loop for the triplets. need to reconsider

image

andiwand:
closing this for now due to priorities

PR #4324, opened 2025-06-05
feat: `BroadTripletSeedFinder` using the new space point and seed EDM

andiwand:
After getting the new space point and seed EDM in with https://github.com/acts-project/acts/pull/4341, we can move the seeding algorithm to take and return this new EDM. Apart from this, the new implementation of the seeding is more generic and can likely be applied to the orthogonal seeding too. This also gets rid of the templating by externalizing the grid access.

**Summary**
- `DoubletSeedFinder` was extracted from `SeedFinder` and only deals with doublets. This should be rather generic and applicable for different seeding strategies, potentially also GBTS. There are multiple frontends to find doublets right now to deal with collections that are sorted or not, as this is required by the `SeedFilter`.
- `BroadTripletSeedFinder` is practically the replacement of `SeedFinder`. It also has two frontends to deal with sorted and unsorted space points. The grid is not part of this interface anymore. It will directly work on collections of space points. I choose the naming "broad" because this algorithm is tuned for dealing with many space points and therefore lots of combinatorics and requires a filter to get rid of those in a second step. Config/options have been refined.
- `BroadTripletSeedFilter` was extracted from `SeedFilter` and stayed very much the same. This filter is tightly coupled to the finder and will likely not be useful to the user by itself. Potentially we should hide it in the future. Config/options have been refined.
- `CylindricalSpacePointGrid2` is very much the same as the one before (`CylindricalSpacePointGrid`). It relies on the new EDM and the config/options have been refined.
- `GridTripletSeedingAlgorithm` uses the new classes above to implement the seeding in Examples. Config/options have been refined.
- Physmon has been moved to use the new seeding implementation.
- Examples EDMs have not been adapted to use the new EDM in the whiteboard yet.

These changes have been tested in Athena and show similar CPU performance and the same physics performance.

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

* **New Features**
  * Introduced a new grid-based triplet seeding algorithm for track finding, offering configurable geometric and kinematic cuts, binning strategies, and advanced seed filtering.
  * Added user-facing configuration options for magnetic field, spatial binning, seed quality confirmation, and space point selection.
  * Provided Python and C++ interfaces for the new seeding algorithm, enabling integration into reconstruction workflows and scripts.

* **Bug Fixes**
  * Updated default seeding algorithm in multiple workflows and scripts to use the new grid triplet seeding method, ensuring consistent behavior across reconstruction chains.

* **Documentation**
  * Enhanced configuration documentation for the new seeding algorithm and updated script argument descriptions to reflect the new default.

* **Chores**
  * Updated build configurations to include new algorithm sources and ensure proper integration with existing libraries and Python bindings.

### DIFF
--- a/Core/include/Acts/Geometry/TrackingVolume.hpp
+++ b/Core/include/Acts/Geometry/TrackingVolume.hpp
@@ -79,9 +79,17 @@ using LayerMultiIntersection =
 
 /// BoundarySurface of a volume
 using BoundarySurface = BoundarySurfaceT<TrackingVolume>;
+
 /// Intersection with a @c BoundarySurface
-using BoundaryIntersection =
-    std::pair<SurfaceIntersection, const BoundarySurface*>;
+/// @note This struct is currently split between a gen1 boundary surface
+///       and a gen3 portal but only one of them will be set. This will go away
+///       once the gen 1 geometry is removed.
+struct BoundaryIntersection {
+  SurfaceIntersection intersection;
+  const BoundarySurface* boundarySurface;
+  const Portal* portal;
+};
+
 /// Multi-intersection with a @c BoundarySurface
 using BoundaryMultiIntersection =
     std::pair<SurfaceMultiIntersection, const BoundarySurface*>;
--- a/Core/include/Acts/Navigation/NavigationDelegate.hpp
+++ b/Core/include/Acts/Navigation/NavigationDelegate.hpp
@@ -25,10 +25,14 @@ struct NavigationArguments {
   Vector3 direction;
 
   BoundaryTolerance tolerance = BoundaryTolerance::None();
+
+  bool wantsPortals = true;
+  bool wantsSurfaces = true;
 };
 
 /// Central alias for the navigation delegate. This type is owning to support
 /// (type-erased) navigation delegate chains (i.e. multiple policies).
+// @TODO: Add geometry context to navigation delegate signature
 using NavigationDelegate = OwningDelegate<void(
     const NavigationArguments&, AppendOnlyNavigationStream&, const Logger&)>;
 
--- a/Core/include/Acts/Propagator/Navigator.hpp
+++ b/Core/include/Acts/Propagator/Navigator.hpp
@@ -8,10 +8,12 @@
 
 #pragma once
 
+#include "Acts/Definitions/Tolerance.hpp"
 #include "Acts/Geometry/GeometryIdentifier.hpp"
 #include "Acts/Geometry/Layer.hpp"
 #include "Acts/Geometry/TrackingGeometry.hpp"
 #include "Acts/Geometry/TrackingVolume.hpp"
+#include "Acts/Navigation/NavigationStream.hpp"
 #include "Acts/Propagator/NavigationTarget.hpp"
 #include "Acts/Propagator/NavigatorError.hpp"
 #include "Acts/Propagator/NavigatorOptions.hpp"
@@ -91,6 +93,8 @@ class Navigator {
       boost::container::small_vector<BoundaryIntersection, 4>;
 
   using ExternalSurfaces = std::multimap<std::uint64_t, GeometryIdentifier>;
+
+  using GeometryVersion = TrackingGeometry::GeometryVersion;
 
   /// The navigation stage
   enum struct Stage : int {
@@ -190,6 +194,8 @@ class Navigator {
 
     NavigatorStatistics statistics;
 
+    NavigationStream stream;
+
     void resetAfterLayerSwitch() {
       navSurfaces.clear();
       navSurfaceIndex.reset();
@@ -224,7 +230,12 @@ class Navigator {
   explicit Navigator(Config cfg,
                      std::shared_ptr<const Logger> _logger =
                          getDefaultLogger("Navigator", Logging::Level::INFO))
-      : m_cfg{std::move(cfg)}, m_logger{std::move(_logger)} {}
+      : m_cfg{std::move(cfg)}, m_logger{std::move(_logger)} {
+    if (m_cfg.trackingGeometry == nullptr) {
+      throw std::invalid_argument("Navigator: No tracking geometry provided.");
+    }
+    m_geometryVersion = m_cfg.trackingGeometry->geometryVersion();
+  }
 
   State makeState(const Options& options) const {
     State state(options);
@@ -279,10 +290,30 @@ class Navigator {
 
     ACTS_VERBOSE(volInfo(state) << "Initialization.");
 
+    auto printGeometryVersion = [](auto ver) {
+      using enum TrackingGeometry::GeometryVersion;
+      switch (ver) {
+        case Gen1:
+          return "Gen1";
+        case Gen3:
+          return "Gen3";
+        default:
+          throw std::runtime_error("Unknown geometry version.");
+      }
+    };
+    ACTS_VERBOSE(volInfo(state) << "Geometry version is: "
+                                << printGeometryVersion(m_geometryVersion));
+
     state.reset();
+
+    // Empirical pre-allocation of candidates for the next navigation iteration.
+    // @TODO: Make this user configurable through the configuration
+    state.stream.candidates().reserve(50);
 
     state.startSurface = state.options.startSurface;
     state.targetSurface = state.options.targetSurface;
+
+    // @TODO: Implement fast initialization with Gen3. This requires the volume lookup to work properly
 
     // Fast Navigation initialization for start condition:
     // - short-cut through object association, saves navigation in the
@@ -409,7 +440,11 @@ class Navigator {
         } else {
           // This was the last surface, switch to layers
           ACTS_VERBOSE(volInfo(state) << "Target layers.");
-          state.navigationStage = Stage::layerTarget;
+          if (m_geometryVersion == GeometryVersion::Gen1) {
+            state.navigationStage = Stage::layerTarget;
+          } else {
+            state.navigationStage = Stage::boundaryTarget;
+          }
         }
       }
 
@@ -443,8 +478,8 @@ class Navigator {
         }
         if (state.navBoundaryIndex.value() < state.navBoundaries.size()) {
           ACTS_VERBOSE(volInfo(state) << "Target set to next boundary.");
-          return NavigationTarget(*state.navBoundary().first.object(),
-                                  state.navBoundary().first.index(),
+          return NavigationTarget(*state.navBoundary().intersection.object(),
+                                  state.navBoundary().intersection.index(),
                                   BoundaryTolerance::None());
         } else {
           // This was the last boundary, we have to leave the volume somehow,
@@ -532,6 +567,9 @@ class Navigator {
 
     state.currentSurface = &surface;
 
+    ACTS_VERBOSE(volInfo(state)
+                 << "Current surface: " << state.currentSurface->geometryId());
+
     if (state.navigationStage == Stage::surfaceTarget &&
         state.navSurface().object() == &surface) {
       ACTS_VERBOSE(volInfo(state) << "Handling surface status.");
@@ -554,20 +592,40 @@ class Navigator {
     }
 
     if (state.navigationStage == Stage::boundaryTarget &&
-        state.navBoundary().first.object() == &surface) {
+        state.navBoundary().intersection.object() == &surface) {
       ACTS_VERBOSE(volInfo(state) << "Handling boundary status.");
 
-      // Switch to the next volume using the boundary
-      const BoundarySurface* boundary = state.navBoundary().second;
-      state.currentVolume = boundary->attachedVolume(state.options.geoContext,
-                                                     position, direction);
+      if (m_geometryVersion == GeometryVersion::Gen1) {
+        // Switch to the next volume using the boundary
+        const BoundarySurface* boundary = state.navBoundary().boundarySurface;
+        assert(boundary != nullptr && "Retrieved boundary surface is nullptr");
+        state.currentVolume = boundary->attachedVolume(state.options.geoContext,
+                                                       position, direction);
+      } else {
+        const Portal* portal = state.navBoundary().portal;
+        assert(portal != nullptr && "Retrieved portal is nullptr");
+        auto res = portal->resolveVolume(state.options.geoContext, position,
+                                         direction);
+        if (!res.ok()) {
+          ACTS_ERROR(volInfo(state)
+                     << "Failed to resolve volume through portal: "
+                     << res.error().message());
+          return;
+        }
+
+        state.currentVolume = res.value();
+      }
 
       // partial reset
       state.resetAfterVolumeSwitch();
 
       if (state.currentVolume != nullptr) {
         ACTS_VERBOSE(volInfo(state) << "Volume updated.");
-        state.navigationStage = Stage::layerTarget;
+        if (m_geometryVersion == GeometryVersion::Gen1) {
+          state.navigationStage = Stage::layerTarget;
+        } else {
+          state.navigationStage = Stage::surfaceTarget;
+        }
       } else {
         ACTS_VERBOSE(volInfo(state)
                      << "No more volume to progress to, stopping navigation.");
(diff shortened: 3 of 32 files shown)


Core/include/Acts/Seeding2/detail/CandidatesForMiddleSp2.hpp
```diff
@@ -0,0 +1,191 @@
...
+  /// @param indices The collection
+  /// @param currentSize The current number of element stored in the collection. The function will
+  /// diminish this value by 1
+  void pop(std::vector<std::size_t>& indices, std::size_t& currentSize);
+
+  /// @brief Return the weight for a candidate
+  /// @param indices The collection in which the element is stored
+  /// @param n Index of the element in the collection
+  /// @returns The weight of the candidate
+  float weight(const std::vector<std::size_t>& indices, std::size_t n) const;
```
paulgessinger, line 160: What do the indices refer to, i.e. what to they index into? SPs? Just from the comments it's not entirely clear.

Core/include/Acts/Seeding2/detail/CandidatesForMiddleSp2.hpp
```diff
@@ -0,0 +1,191 @@
...
+  /// all the selection criteria
+  /// @param indices The collection
+  /// @param n Current number of stored elements in the collection
+  /// @param nMax The maximum number of elements that can be stored in the collection
+  /// @param element The element that must be added to the collection
+  void addToCollection(std::vector<std::size_t>& indices, std::size_t& n,
+                       const std::size_t nMax,
+                       const TripletCandidate2& element);
+
+ private:
```
paulgessinger: Already in `private:`.

Core/src/Seeding2/BroadTripletSeedFilter.cpp
```diff
@@ -0,0 +1,328 @@
...
+namespace Acts::Experimental {
+
+namespace {
+
+float getBestSeedQuality(
+    const SpacePointContainer2& spacePoints,
+    const std::unordered_map<SpacePointIndex2, float>& bestSeedQualityMap,
+    SpacePointIndex2 sp) {
+  if (spacePoints.hasExtraColumns(SpacePointKnownExtraColumn::CopyFromIndex)) {
+    sp = spacePoints.at(sp).extra(spacePoints.copyFromIndexColumn());
```
paulgessinger, line 27: So this redirects to some other SP? When is this used, I'm not sure I understand it.

Core/src/Seeding2/BroadTripletSeedFilter.cpp
```diff
@@ -0,0 +1,328 @@
...
+    if (it != bestSeedQualityMap.end()) {
+      it->second = std::max(quality, it->second);
+    } else {
+      bestSeedQualityMap.emplace(sp, quality);
+    }
+  };
+
+  set(top);
+  set(middle);
+  set(bottom);
```
paulgessinger: Not important, but could be a for loop I guess.

Core/src/Seeding2/BroadTripletSeedFilter.cpp
```diff
@@ -0,0 +1,328 @@
...
+  // seed confirmation
+  SeedConfirmationRangeConfig seedConfRange;
+  std::size_t nTopSeedConf = 0;
+  if (m_cfg.seedConfirmation) {
+    // check if bottom SP is in the central or forward region
+    seedConfRange =
+        (spB.z() > m_cfg.centralSeedConfirmationRange.zMaxSeedConf ||
+         spB.z() < m_cfg.centralSeedConfirmationRange.zMinSeedConf)
+            ? m_cfg.forwardSeedConfirmationRange
+            : m_cfg.centralSeedConfirmationRange;
```
paulgessinger: Nitpick: could write full-fledged if/else for legibility.

Core/src/Seeding2/BroadTripletSeedFilter.cpp
```diff
@@ -0,0 +1,328 @@
...
+  };
+
+  std::size_t beginCompTopIndex = 0;
+  // loop over top SPs and other compatible top SP candidates
+  for (const std::size_t topSpIndex : cache.topSpIndexVec) {
+    auto topSp = topSpVec[topSpIndex];
+    auto spT = spacePoints.at(topSp);
+
+    // if two compatible seeds with high distance in r are found, compatible
+    // seeds span 5 layers
```
paulgessinger: Does this statement not depend on the detector?

Core/include/Acts/Seeding2/BroadTripletSeedFinder.hpp
```diff
@@ -0,0 +1,256 @@
...
+    DerivedTripletCuts derive(float bFieldInZ) const;
+  };
+
+  struct DerivedTripletCuts : public TripletCuts {
+    float highland = 0;
+    float pTPerHelixRadius = std::numeric_limits<float>::quiet_NaN();
+    float minHelixDiameter2 = std::numeric_limits<float>::quiet_NaN();
+    float sigmapT2perRadius = std::numeric_limits<float>::quiet_NaN();
+    float multipleScattering2 = std::numeric_limits<float>::quiet_NaN();
+  };
```
paulgessinger, line 93: This (and the structs below) could probably be private, right?

Core/src/Seeding2/BroadTripletSeedFinder.cpp
```diff
@@ -0,0 +1,808 @@
...
+  return true;
+}
+
+}  // namespace
+
+BroadTripletSeedFinder::DerivedTripletCuts
+BroadTripletSeedFinder::TripletCuts::derive(float bFieldInZ) const {
+  DerivedTripletCuts result;
+
+  static_cast<TripletCuts&>(result) = *this;
```
paulgessinger, line 76: Is this safe?

Core/src/Seeding2/BroadTripletSeedFinder.cpp
```diff
@@ -0,0 +1,808 @@
...
+
+  // no top SP found -> cannot form any triplet
+  if (cache.topDoublets.empty()) {
+    ACTS_VERBOSE("No compatible Tops, returning");
+    return;
+  }
+
+  // apply cut on the number of top SP if seedConfirmation is true
+  float rMaxSeedConf = 0;
+  if (filter.config().seedConfirmation) {
```
paulgessinger, line 140: Configuration between finder and filter are coupled?

Core/src/Seeding2/BroadTripletSeedFinder.cpp
```diff
@@ -0,0 +1,808 @@
...
+    // check if middle SP is in the central or forward region
+    SeedConfirmationRangeConfig seedConfRange =
+        (spM.z() > filter.config().centralSeedConfirmationRange.zMaxSeedConf ||
+         spM.z() < filter.config().centralSeedConfirmationRange.zMinSeedConf)
+            ? filter.config().forwardSeedConfirmationRange
+            : filter.config().centralSeedConfirmationRange;
```
paulgessinger: Consider replacing ternary with if/else

Core/src/Seeding2/BroadTripletSeedFinder.cpp
```diff
@@ -0,0 +1,808 @@
...
+  std::ranges::transform(
+      bottomSpGroups, std::back_inserter(cache.bottomSpOffsets),
+      [&](const std::span<const SpacePointIndex2>& bottomSps) {
+        return std::ranges::lower_bound(
+                   bottomSps, firstMiddleSpR - bottomCuts.deltaRMax, {},
+                   [&](const SpacePointIndex2& spIndex) {
+                     auto sp = spacePoints.at(spIndex);
+                     return sp.r();
+                   }) -
+               bottomSps.begin();
```
paulgessinger: Purely for readibility, maybe:
```suggestion
        auto low = std::ranges::lower_bound(
                   bottomSps, firstMiddleSpR - bottomCuts.deltaRMax, {},
                   [&](const SpacePointIndex2& spIndex) {
                     auto sp = spacePoints.at(spIndex);
                     return sp.r();
                   });
        return low - bottomSps.begin();
```

Core/src/Seeding2/BroadTripletSeedFinder.cpp
```diff
@@ -0,0 +1,808 @@
...
+    BroadTripletSeedFilter::Cache& filterCache,
+    const SpacePointContainer2& spacePoints, const ConstSpacePointProxy2& spM,
+    const DoubletSeedFinder::DoubletsForMiddleSp& bottomDoublets,
+    const DoubletSeedFinder::DoubletsForMiddleSp& topDoublets,
+    TripletTopCandidates& tripletTopCandidates,
+    CandidatesForMiddleSp2& candidatesCollector) {
+  const float rM = spM.r();
+  const float cosPhiM = spM.x() / rM;
+  const float sinPhiM = spM.y() / rM;
+  // TODO use some reasonable defaults
```
paulgessinger: To be followed up?

Core/src/Seeding2/CylindricalSpacePointGrid2.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+  // vector that will store the edges of the bins of z
+  std::vector<double> zValues{};
+
+  // If zBinEdges is not defined, calculate the edges as zMin + bin * zBinSize
+  if (m_cfg.zBinEdges.empty()) {
+    // TODO: can probably be optimized using smaller z bins
+    // and returning (multiple) neighbors only in one z-direction for forward
+    // seeds
+    // FIXME: zBinSize must include scattering
```
paulgessinger, line 118: Follow up item?

Core/src/Seeding2/CylindricalSpacePointGrid2.cpp
```diff
@@ -0,0 +1,218 @@
...
+    // consecutive phi bins in the seed making step.
+    // Each individual bin should be approximately a fraction (depending on this
+    // number) of the maximum expected azimutal deflection.
+
+    // set protection for large number of bins, by default it is large
+    phiBins = std::min(phiBins, m_cfg.maxPhiBins);
+  }
+
+  Axis<AxisType::Equidistant, AxisBoundaryType::Closed> phiAxis(
+      m_cfg.phiMin, m_cfg.phiMax, phiBins);
```
paulgessinger: No change needed, just wanted to point it out:


```suggestion
  Axis phiAxis(AxisClosed, m_cfg.phiMin, m_cfg.phiMax, phiBins);
```

Core/src/Seeding2/CylindricalSpacePointGrid2.cpp
```diff
@@ -0,0 +1,218 @@
...
+  if (m_cfg.rBinEdges.empty()) {
+    rValues = {m_cfg.rMin, m_cfg.rMax};
+  } else {
+    rValues.insert(rValues.end(), m_cfg.rBinEdges.begin(),
+                   m_cfg.rBinEdges.end());
+  }
+
+  Axis<AxisType::Variable, AxisBoundaryType::Open> zAxis(std::move(zValues));
+  Axis<AxisType::Variable, AxisBoundaryType::Open> rAxis(std::move(rValues));
+
```
paulgessinger:
```suggestion
  Axis zAxis(AxisOpen, std::move(zValues));
  Axis rAxis(AxisOpen, std::move(rValues));

```

Core/include/Acts/Seeding2/DoubletSeedFinder.hpp
```diff
@@ -0,0 +1,174 @@
...
+
+#include "Acts/EventData/SpacePointContainer2.hpp"
+#include "Acts/Seeding/SeedFinderUtils.hpp"
+#include "Acts/Utilities/Delegate.hpp"
+
+#include <vector>
+
+namespace Acts::Experimental {
+
+class DoubletSeedFinder {
```
paulgessinger, line 20: This only has static methods, so is effectively a namespace, or an abstract class? I don't have a better idea how to model this, just wanted to point it out.

Core/src/Seeding2/DoubletSeedFinder.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+  // equivalent to impactMax / (rM * rM);
+  float vIPAbs = impactMax * middleSpInfo.uIP2;
+
+  float deltaR = 0.;
+  float deltaZ = 0.;
+
+  const auto outsideRangeCheck = [](const float value, const float min,
+                                    const float max) {
+    // intentionally using `|` after profiling. faster due to better branch
```
paulgessinger, line 67: 🤡

Core/src/Seeding2/detail/CandidatesForMiddleSp2.cpp
```diff
@@ -0,0 +1,290 @@
...
+    // In the first two cases we stop, since we are already in the correct
+    // position
+
+    // if there is no left child, that also means no right child is present.
+    // We do nothing
+    if (!exists(leftChild, actualSize)) {
+      break;
+    }
+
+    // At least one of the child is present. Left child for sure, right child we
```
paulgessinger:
```suggestion
    // At least one of the children is present. Left child for sure, right child we
```

Core/src/Seeding2/detail/CandidatesForMiddleSp2.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+  // rely on the fact that m_indices* are both min heap trees
+  // Sorting comes naturally by popping elements one by one and
+  // placing this element at the end of the output vector
+  while (m_nHigh != 0 || m_nLow != 0) {
+    // no entries in collection high, we attach the entire low collection
+    if (m_nHigh == 0) {
+      std::size_t idx = m_nLow;
+      for (std::size_t i(0); i < idx; i++) {
+        output[outIdx--] = m_storage[m_indicesLow[0]];
```
paulgessinger: Split up decrement operator and `operator[]` to clarify. Same below.

Core/src/Seeding2/detail/CandidatesForMiddleSp2.cpp
```diff
@@ -0,0 +1,290 @@
...
+  std::size_t outIdx = output.size() - 1;
+
+  // rely on the fact that m_indices* are both min heap trees
+  // Sorting comes naturally by popping elements one by one and
+  // placing this element at the end of the output vector
+  while (m_nHigh != 0 || m_nLow != 0) {
+    // no entries in collection high, we attach the entire low collection
+    if (m_nHigh == 0) {
+      std::size_t idx = m_nLow;
+      for (std::size_t i(0); i < idx; i++) {
```
paulgessinger:
```suggestion
      for (std::size_t i = 0; i < idx; i++) {
```
?

paulgessinger (commented):
Is this covered by any of the python level tests and/or physmon?

Core/src/Seeding2/detail/CandidatesForMiddleSp2.cpp
andiwand: yeah I agree this code is not great - just copy pasted this. I will apply your suggestion

Core/include/Acts/Seeding2/detail/CandidatesForMiddleSp2.hpp
andiwand: I think this points into `m_storage`

Core/src/Seeding2/BroadTripletSeedFilter.cpp
andiwand: My idea here was that before running the seeding, the calling code could copy all relevant space points into a local cache so they are dense in memory and not randomly distributed. But the seed filter depends on the original index so I have to keep an original index

Core/src/Seeding2/BroadTripletSeedFilter.cpp
andiwand: removing the comment as it seems out of place and not relevant

Core/src/Seeding2/BroadTripletSeedFinder.cpp
andiwand: to my best knowledge this should be safe yes

Core/include/Acts/Seeding2/BroadTripletSeedFinder.hpp
andiwand: I think in principle these are all user facing

Core/src/Seeding2/BroadTripletSeedFinder.cpp
andiwand: yes the finder behaves differently depending on the configuration of the filter. I don't think this is pretty but duplicating these props seems even worse

at the end of the day these two classes are completely coupled and inseparable. so I would argue it is okay

Core/src/Seeding2/BroadTripletSeedFinder.cpp
andiwand: yeah I still do not know how we use this in Athena

Core/src/Seeding2/CylindricalSpacePointGrid2.cpp
andiwand: this was a copy paste. personally I do not know what to do with this

Issue #2346, opened 2023-08-04
FPE: Bethe-Heithler gamma_distribution


paulgessinger:
https://github.com/acts-project/acts/blob/b5babcb04bb610b55a02a753b180e5d3893ca972/Fatras/include/ActsFatras/Physics/ElectroMagnetic/BetheHeitler.hpp#L64-L65

Also discussed in the past on [Mattermost](https://mattermost.web.cern.ch/acts/pl/k69o4hp5wjr53c4ayy6rouxmkr)

PR #4339, opened 2025-06-12
fix: no underflow in gamma distribution

pbalek:
fix underflow in gamma distribution

--- END COMMIT MESSAGE ---

In Bethe-Heitler description, the `std::gamma_distribution` might throw `FPE_UND` (underflow), when it tries to calculate `std::pow(small_number, large_number)`. The result is zero and both Bethe and Heitler are fine with it, but we get `FPE_UND` along the way.

This PR overpasses the problem. If the parameter `alpha` is greater than 1 (unlikely in this scenario), everything proceeds as before; there is no risk of underflow. If it's less than 1, then the code proceeds as in [the actual implementation](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/random.tcc#L2339) of the `std::gamma_distribution`:
 1. [use](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/random.tcc#L2344) `alpha+1` and get a random number from that gamma distribution
 2. [multiply](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/random.tcc#L2393) this number by `pow(..., 1/alpha)`

However, we first check if the result of the second step will be above floating point minimum. If not, zero is returned as it would anyway, only this time without `FPE_UND`; if it is large enough, we proceed with the actual calculation of `pow(...)`.

As long as we use `libstdc++` standard libraries, the modified code uses the same amount of calls to the random number engine (and with the same usage), and thus the produced random numbers sequence is the same as before. If we use `libcxx` libraries, the overall  distribution is still the same, however the random number sequence will be different.

Closes issue #2346 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved numerical stability when simulating electron energy loss, especially for cases with small shape parameters, reducing the risk of calculation errors.

- **Refactor**
  - Enhanced internal handling of random sampling for more robust and accurate results during simulations.


andiwand:
I wonder if this should be rather dealt with FPE masking

andiwand:
converting to draft for now to take some congestion from the gitlab ci

pbalek:
Hi @andiwand 

> I wonder if this should be rather dealt with FPE masking

Do you mean masking of `u = uPlus1 * std::pow(u2, 1 / alpha);` or masking of the original code (`u = gDist(generator`)?

The problem is that masking in the original code didn't work for me, I got FPE from `std::gamma_distribution` and couldn't get rid of it.

Fatras/include/ActsFatras/Physics/ElectroMagnetic/BetheHeitler.hpp
```diff
@@ -55,16 +56,35 @@
...
+      u = gDist(generator);
+    } else {
+      std::gamma_distribution<double> gDistPlus1(alpha + 1., 1.);
+      // Get a random number using alpha+1
+      const auto uPlus1 = gDistPlus1(generator);
+      double u2;
+      do {
+        u2 = uDist(generator);
+      } while (u2 == 0.);
+      // Check if this would cause underflow
```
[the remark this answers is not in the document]
pbalek: No no, the number of calls is different in the `std::gamma_distribution` implementation as well.

andiwand:
there is a way to mark multiple lines of code like for example here
https://github.com/acts-project/acts/blob/766b51652f75d58a6cfcc1d4f1569edcbf2a76d3/Core/include/Acts/Visualization/Interpolation3D.hpp#L63-L68

apart from that, I usually run my workflows without FPE handling because it is quite annoying. For the CI it makes a lot of sense, but if it stops because there is an FPE in event 37128 locally, I really don't mind.
https://github.com/acts-project/acts/blob/766b51652f75d58a6cfcc1d4f1569edcbf2a76d3/Examples/Framework/include/ActsExamples/Framework/Sequencer.hpp#L85

pbalek:
Hi @andiwand,
I wasn't able to silence the FPE with fpeMaskBegin/fpeMaskEnd.
However, setting `trackFpes=False` works - thank you! 
Thus, if there's no need for this PR, I guess we can close it?

andiwand:
> Hi @andiwand, I wasn't able to silence the FPE with fpeMaskBegin/fpeMaskEnd. However, setting `trackFpes=False` works - thank you! Thus, if there's no need for this PR, I guess we can close it?

I am surprised that the masking does not work - potentially something we should investigate at some point.

For me personally it does not make a lot of sense to protect against FPEs in all possible places, because it is just too many. I would say our common solution right now is to only protect against FPEs when they happen a lot. At least for now I could not remember seeing this FPE often. So I would be inclined to leave it as it for now, but that is only my opinion.

pbalek:
> I am surprised that the masking does not work - potentially something we should investigate at some point.

I put the beginning before `std::gamma_distribution<double> gDist(...)` and end after `const auto z = std::exp(-u);`. Maybe it's not silenced because the FPE is raised inside the inner function and not in this one? That would be quite counter-intuitive for me...

> For me personally it does not make a lot of sense to protect against FPEs in all possible places, because it is just too many. I would say our common solution right now is to only protect against FPEs when they happen a lot. At least for now I could not remember seeing this FPE often. So I would be inclined to leave it as it for now, but that is only my opinion.

It happens once or twice almost every event (in Pb+Pb collision). So there are not hundreds of them per event but it's virtually impossible for me to run a job and not to see it.

andiwand:
> I put the beginning before `std::gamma_distribution<double> gDist(...)` and end after `const auto z = std::exp(-u);`. Maybe it's not silenced because the FPE is raised inside the inner function and not in this one? That would be quite counter-intuitive for me...

the FPE masking should work even if you use it outside the function the FPE happens. we use this quite extensively because most FPEs happen inside `std::` math functions.

> It happens once or twice almost every event (in Pb+Pb collision). So there are not hundreds of them per event but it's virtually impossible for me to run a job and not to see it.

ok that sounds a bit much. I will try to think about it a bit more... potentially it makes sense to build a custom FPE safe gamma distribution? just to lift this functionality and make it reusable

Fatras/include/ActsFatras/Physics/ElectroMagnetic/BetheHeitler.hpp
```diff
@@ -55,16 +56,35 @@ struct BetheHeitler {
...
                                      Particle &particle) const {
+    std::uniform_real_distribution<double> uDist(0., 1.);
+
     // Take a random gamma-distributed value - depending on t/X0
-    std::gamma_distribution<double> gDist(
-        slab.thicknessInX0() / std::numbers::ln2, 1.);
+    const double alpha = slab.thicknessInX0() / std::numbers::ln2;
+    double u = 0.;
+    if (alpha > 1.) {
+      std::gamma_distribution<double> gDist(alpha, 1.);
```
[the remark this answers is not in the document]
pbalek: Hmm, right you are, I will zero return, if alpha equal or less than zero shall be.

pbalek:
Hi @andiwand, 
> potentially it makes sense to build a custom FPE safe gamma distribution? just to lift this functionality and make it reusable

That's a good idea. I moved the relevant part to a separate function. I took `Fatras/include/ActsFatras/Utilities/LandauDistribution.hpp` as a template and created `GammaDistribution.hpp`. I also made a unit test to check we'll get exactly the same results from `GammaDistribution` as from `std::gamma_distribution<double>`

Does it look good?

Fatras/include/ActsFatras/Utilities/GammaDistribution.hpp
```diff
@@ -0,0 +1,120 @@
...
+    } else {
+      std::gamma_distribution<double> gDistPlus1(params.alpha + 1.,
+                                                 params.theta);
+      // Get a random number using alpha+1
+      const auto uPlus1 = gDistPlus1(generator);
+      std::uniform_real_distribution<double> uDist(0., 1.);
+      double u2 = 0.;
+      do {
+        u2 = uDist(generator);
+      } while (u2 == 0.);
```
andiwand:
```suggestion
      // draw from uniform (0,1)
      std::uniform_real_distribution<double> uDist(std::numeric_limits<double>::min(), 1.);
      double u2 = uDist(generator);
```
would this also do it?

Fatras/include/ActsFatras/Utilities/GammaDistribution.hpp
```diff
@@ -0,0 +1,120 @@
...
+  }
+  /// Generate a random number from the given gamma distribution.
+  template <typename Generator>
+  result_type operator()(Generator &generator, const param_type &params) {
+    if (params.alpha >= 1.) {
+      std::gamma_distribution<double> gDist(params.alpha, params.theta);
+      return gDist(generator);
+    } else if (params.alpha <= 0.) {
+      return 0.;
+    } else {
```
andiwand:
```suggestion
    }
```
matter of taste - you could remove `else` and the `{}` block below which gets rid of some indentation

Fatras/include/ActsFatras/Utilities/GammaDistribution.hpp
pbalek: There's a small chance that `u2` will be zero. However, it's possible to avoid initialization with 0, and initialize with a random number and then use `while(...) {...}`

Fatras/include/ActsFatras/Utilities/GammaDistribution.hpp
pbalek: 👍

acts-project-service:
## :white_check_mark: Athena integration test results [1ad6abf5520c0ad19defad102e12114cedf72be7]

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627296)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627295)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627294)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627293)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627292)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627291)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627290)||
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627289)||
|:green_circle: | [test_ActsCheckObjectCountsWorkflowHgtd](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627288)||
|:green_circle: | [test_ActsCheckObjectCountsCachedWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627287)||
|:green_circle: | [test_ActsCheckObjectCountsLegacyWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627286)||
|:green_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627285)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627284)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627283)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627282)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627281)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627280)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627279)||
|:green_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627278)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627277)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627276)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627275)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627274)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627273)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627272)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627271)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627270)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627269)||
|:green_circle: | [test_ActsGx2fRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627268)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627267)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627266)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627265)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/57627264)||

PR #4341, opened 2025-06-12
feat: A new space point and seed container

andiwand:
This adds a new, experimental space point and seed container. The idea is to make them easy to use and extensible, without any templating. This means that users will have to copy space points or seeds in the future if they have a different EDM on their side. My argument for this is that the EDM is part of the algorithms interface and to guarantee anything about computational performance of the algorithm, the EDM should be fixed. Apart from that this reduces templating and indirections.

Current limitations
- space points are identified by a single index. if there are multiple space point containers, they need to be combined if both sets are relevant for later reconstruction steps
- mixed types of space points may have suboptimal memory usage. for example strip space points have additional information which is not present for pixel seeds. we either waste memory by adding a dense column or buy runtime overhead by resolving which SP has what information
- seed container does not refer to or hold the space point container. the user has to keep the space point container around and pass it in when necessary


andiwand:
I would be happy for feedback on this @paulgessinger @CarloVarni @timadye @noemina @LuisFelipeCoelho @asalzburger

Core/include/Acts/EventData/SeedContainer2.hpp
```diff
@@ -0,0 +1,356 @@
...
+  /// associated with it.
+  /// @return The number of space points in the seed.
+  [[nodiscard]] std::size_t size() const {
+    return m_container->spacePointIndices(m_index).size();
+  }
+  /// Checks if the seed is empty, i.e., has no space points associated with it.
+  /// @return True if the seed is empty, false otherwise.
+  [[nodiscard]]
+  bool empty() const {
+    return size() == 0;
```
asalzburger: Why not calling `return m_container->spacePointIndices(m_index).empty()` - I think this is recommended.

Core/include/Acts/EventData/SeedContainer2.hpp
```diff
@@ -0,0 +1,356 @@
...
+  }
+  /// Const access to the quality of the seed.
+  /// @return The quality of the seed.
+  float quality() const { return m_container->quality(m_index); }
+  /// Const access to the vertex Z coordinate of the seed.
+  /// @return The vertex Z coordinate of the seed.
+  float vertexZ() const { return m_container->vertexZ(m_index); }
+
+  class SpacePointIterator {
+   public:
```
asalzburger: Is this sufficiently implemented that we can call `for (const auto& sp : spacePointContainer)` ?

Core/include/Acts/EventData/SeedContainer2.hpp
```diff
@@ -0,0 +1,356 @@
...
+  std::size_t size() const { return m_entries.size(); }
+  /// Checks if the seed container is empty.
+  /// @return True if the container is empty, false otherwise.
+  bool empty() const { return size() == 0; }
+
+  /// Reserves space for the given number of seeds.
+  /// @param size The number of seeds to reserve space for.
+  void reserve(std::size_t size) {
+    m_entries.reserve(size);
+    m_spacePoints.reserve(size * 3);
```
asalzburger: Can we make the avgerage/expected number of space points per seed  a parameter we give to `reserve( .., std::size_t seedSize = 3u)` and default it to three?

Core/include/Acts/EventData/SeedContainer2.hpp
```diff
@@ -0,0 +1,356 @@
...
+
+  class SpacePointRange {
+   public:
+    SpacePointRange(const SpacePointContainer2 &spacePointContainer,
+                    std::span<const std::size_t> spacePointIndices)
+        : m_spacePointContainer{&spacePointContainer},
+          m_spacePointIndices{spacePointIndices} {}
+
+    std::size_t size() const { return m_spacePointIndices.size(); }
+    bool empty() const { return size() == 0; }
```
asalzburger: Same as above, at least `SonarCloud` always complains.

Core/include/Acts/EventData/SpacePointContainer2.hpp
```diff
@@ -0,0 +1,673 @@
...
+  [[nodiscard]] bool empty() const { return size() == 0; }
+
+  /// Reserves space for the given number of space points.
+  /// This will reserve space for the source links and the extra columns as
+  /// well.
+  /// @param size The number of space points to reserve space for.
+  void reserve(std::size_t size) {
+    m_entries.reserve(size);
+    m_xyz.reserve(size * 3);
+    m_sourceLinks.reserve(size);
```
asalzburger: Does this design force more than one source link for the space point? We will need that for the strips.

Core/include/Acts/EventData/SpacePointContainer2.hpp
```diff
@@ -0,0 +1,673 @@
...
+  /// @return A const reference to the y coordinate of the space point.
+  float y(IndexType index) const { return m_xyz[index * 3 + 1]; }
+  /// Const access to the z coordinate of the space point at the given index.
+  /// @param index The index of the space point.
+  /// @return A const reference to the z coordinate of the space point.
+  float z(IndexType index) const { return m_xyz[index * 3 + 2]; }
+
+  /// Additional dense column of data that can be added to the space point
+  /// container. The column is indexed by the space point index.
+  template <typename T>
```
asalzburger, line 466: I have not fully understood what the benefit of this wrt `std::vector<ValueType>` alone is?

Core/include/Acts/EventData/SpacePointContainer2.hpp
```diff
@@ -0,0 +1,673 @@
...
+      ++m_index;
+      return *this;
+    }
+    Iterator operator++(int) {
+      Iterator tmp(*this);
+      ++(*this);
+      return tmp;
+    }
+
+    bool operator==(const Iterator &other) const {
```
asalzburger: This checks not only if the objects are identical in shape and type, they also need to point to the same container - if this is intended, please specify so in a comment.

asalzburger (changes requested):

Core/include/Acts/EventData/SpacePointContainer2.hpp
andiwand: I think returning `std::vector<ValueType>` would give the user too much control over the extra column. For instance they could modify the size and elements without us knowing. This way we narrow the interface to what we believe is really necessary

Core/include/Acts/EventData/SpacePointContainer2.hpp
andiwand: Right - I have to check what STL does about this. I would think mixing iterators from different containers is a user error but it is good to stay in line with common practice here

Core/include/Acts/EventData/SpacePointContainer2.hpp
andiwand: The current implementation will allow basically any number of source links. We should use a lower limit of 1 I suppose. There is also no upper limit for now. I would leave this up to the user

Core/include/Acts/EventData/SeedContainer2.hpp
andiwand: Right this and `begin`/`end` allows for using range based for loops

Core/include/Acts/EventData/SeedContainer2.hpp
andiwand: I think in this case sonar cloud is happy because it is the implementation of `empty` and not the usage of `size() == 0` vs `empty()`. To me the current implementation seems easiest but happy to change if you have concerns

asalzburger: ok, fine then.

Core/include/Acts/EventData/SeedContainer2.hpp
asalzburger: Not picky here.

Core/include/Acts/EventData/SpacePointContainer2.hpp
asalzburger: Ok, understood.

Core/include/Acts/EventData/SpacePointContainer2.hpp
asalzburger: Understood, makes sense.

asalzburger:
I'd be very interested in the your ATHENA results here.

andiwand:
> I'd be very interested in the your ATHENA results here.

Unfortunately, until now I only measured slow downs. I believe this is due to the algorithmic changes I made to simplify the interface. I will walk them back and measure again. Apparently the Athena ITk workload is very different compared to the ACTS ODD workload.

Core/include/Acts/EventData/SpacePointContainer2.hpp
andiwand: STL only checks the index

Core/include/Acts/EventData/SeedContainer2.hpp
```diff
@@ -0,0 +1,386 @@
...
+  /// @return A span of space point indices associated with the seed at the given
+  ///         index.
+  std::span<SpacePointIndex2> spacePointIndices(IndexType index) {
+    assert(index < m_entries.size() && "Index out of bounds");
+    return std::span<SpacePointIndex2>(
+        m_spacePoints.data() + m_entries[index].spacePointOffset,
+        m_spacePoints.data() + m_entries[index].spacePointOffset +
+            m_entries[index].seedSize);
+  }
+  /// Mutable access to the quality of the seed at the given index.
```
paulgessinger: NIT:

```suggestion
  }
  
  /// Mutable access to the quality of the seed at the given index.
```

Core/include/Acts/EventData/SeedContainer2.hpp
```diff
@@ -0,0 +1,386 @@
...
+      ++(*this);
+      return tmp;
+    }
+
+    bool operator==(const SeedIterator &other) const {
+      return m_index == other.m_index && m_container == other.m_container;
+    }
+    bool operator!=(const SeedIterator &other) const {
+      return !(*this == other);
+    }
```
paulgessinger: hidden friends?

Core/include/Acts/EventData/SpacePointContainer2.hpp
```diff
@@ -0,0 +1,965 @@
...
+/// links. Individual space points are addressed via index. A proxy object
+/// simplifies the handling.
+class SpacePointContainer2 {
+ public:
+  using IndexType = SpacePointIndex2;
+  using IndexRangeType = SpacePointIndexRange2;
+  using MutableProxyType = MutableSpacePointProxy2;
+  using ConstProxyType = ConstSpacePointProxy2;
+
+  enum KnownExtraColumn : std::uint32_t {
```
paulgessinger: I think I know why you're using a bare enum, but consider using an `enum class` with [explicit operators](https://github.com/acts-project/acts/blob/95264c89db9c31d4bda519484f29112c973d81a9/Core/include/Acts/Utilities/EnumBitwiseOperators.hpp) instead, or make the enum private if possible.

Core/include/Acts/EventData/SpacePointContainer2.hpp
```diff
@@ -0,0 +1,965 @@
...
+            TopStripCenter
+  };
+
+  /// Constructs and empty space point container.
+  SpacePointContainer2() = default;
+
+  /// Constructs a copy of the given space point container.
+  /// The extra columns are copied as well.
+  /// @param other The space point container to copy.
+  SpacePointContainer2(const SpacePointContainer2 &other)
```
paulgessinger: Much of the implementation in this class could potentially go into a .cpp file, I imagine.

Core/include/Acts/EventData/SpacePointContainer2.hpp
```diff
@@ -0,0 +1,965 @@
...
+  /// Reserves space for the given number of space points.
+  /// This will reserve space for the source links and the extra columns as
+  /// well.
+  /// @param size The number of space points to reserve space for.
+  /// @param averageSourceLinks The average number of source links per space point.
+  void reserve(std::size_t size, float averageSourceLinks = 1) {
+    m_xyz.reserve(size * 3);
+    m_sourceLinkOffsets.reserve(size);
+    m_sourceLinkCounts.reserve(size);
+    m_sourceLinks.reserve(static_cast<std::size_t>(size * averageSourceLinks));
```
paulgessinger: The other columns are handled because they will also be stored in `m_extraColumns`, right?

Core/include/Acts/EventData/SpacePointContainer2.hpp
```diff
@@ -0,0 +1,965 @@
...
+
+    std::unique_ptr<ColumnHolderBase> copy() const final {
+      return std::make_unique<ExtraColumnHolder<T>>(*this);
+    }
+    ProxyType proxy() const { return ProxyType(m_data); }
+
+    void reserve(std::size_t size) final { m_data.reserve(size); }
+    void clear() final { m_data.clear(); }
+    void resize(std::size_t size) final { m_data.resize(size); }
+    void emplace_back() final { m_data.emplace_back(); }
```
paulgessinger: I think the `final` here is redundant because the class is alreay `final`. Sonar might flag this and suggest `override`.

Core/include/Acts/EventData/SpacePointContainer2.hpp
```diff
@@ -0,0 +1,965 @@
...
+  std::vector<float> m_xyz;
+  std::vector<std::size_t> m_sourceLinkOffsets;
+  std::vector<std::uint8_t> m_sourceLinkCounts;
+  std::vector<SourceLink> m_sourceLinks;
+
+  // cylindrical coordinates
+  std::optional<ExtraColumnHolder<float>> m_rColumn;
+  std::optional<ExtraColumnHolder<float>> m_phiColumn;
+  // time information
+  std::optional<ExtraColumnHolder<std::optional<float>>> m_timeColumn;
```
paulgessinger: Double optional? Probably not necessary, but you could encode the *unset* time with a sentinel `float` value.

Core/include/Acts/EventData/SpacePointContainer2.hpp
```diff
@@ -0,0 +1,965 @@
...
+      throw std::runtime_error("Extra column not found: " + name);
+    }
+    auto &holder = dynamic_cast<Holder &>(*it->second);
+    return holder.proxy();
+  }
+};
+
+/// A proxy class for accessing individual space points.
+template <bool read_only>
+class SpacePointProxy2 {
```
paulgessinger: Consider splitting this out into a separate file. I didn't do this in the beginning for the track edm, but I think it's beneficial when looking up the API.

Core/include/Acts/EventData/SpacePointContainer2.hpp
andiwand: What I wanted to capture here is that not all space points will have time information even if some of them do. This is what we also have in the ODD. A sentinel value would also do it but I guess the only meaningful would be `NaN`?

Core/include/Acts/EventData/SpacePointContainer2.hpp
andiwand: True - I think the accessors should stay in the header to allow inlining. But the copy constructor can be moved for example 👍

Core/include/Acts/EventData/SeedContainer2.hpp
andiwand: I thought this is only relevant for the binary case?

for tracks we seem to do the same here

https://github.com/acts-project/acts/blob/95264c89db9c31d4bda519484f29112c973d81a9/Core/include/Acts/Utilities/Iterator.hpp#L122-L126

Core/include/Acts/EventData/SpacePointContainer2.hpp
andiwand: right

Core/include/Acts/EventData/SpacePointContainer2.hpp
andiwand: Implemented this now but it is a bit hacky. Since the container and the proxy are co-dependent it requires to use an `ipp` file and the correct include order.

Core/include/Acts/EventData/SpacePointContainer2.hpp
andiwand: used nan now

asalzburger (approved):
Ok from my side

andiwand:
compile error should be gone @asalzburger @paulgessinger 

weird clang 19 bug ...

asalzburger (approved):

PR #4362, opened 2025-06-24
chore: Update Geomodel version

junggjo9:
Update GeoModel to the lateset 6.14 release

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

- **Chores**
  - Updated the minimum required version of the GeoModel dependency to 6.13.0.


paulgessinger:
Does this need to be the absolute minimum version that we're compatible with? Is there anything that fundamentally breaks if it's below?

junggjo9:
No, there're no breaking API changes between 6.14.0 and 6.3.0.

paulgessinger:
Then we probably don't need to bump the minimum required version at all, I suppose?

PR #4368, opened 2025-06-30
refactor: GeoModelTree gets its own list of published fullPhysVol

junggjo9:
- Introduce helper struct into the GeoModelTree which manages the list of full physical volumes. If the geometry is built from the DB nothing changes basically. However, in case of a transient in-memory built, the full phyiscal volumes can be published here
- Construct the tree from the GMDBManager as this is the primary class to query infromation from the SQLite file
--- END COMMIT MESSAGE ---

Tagging @paulgessinger 

## Summary by CodeRabbit

* **New Features**
  * Added thread-safe management and retrieval of geometry volumes by subsystem, improving reliability in concurrent environments.
  * Introduced explicit volume publishing and caching for more efficient access to geometry data.

* **Documentation**
  * Clarified documentation for tracking geometry builder requirements.

* **Bug Fixes**
  * Improved error handling and messaging for missing components in geometry management.

* **Chores**
  * Updated build configuration to include new source files.


Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelTree.hpp
```diff
@@ -9,17 +9,85 @@
...
-
+/// @brief Holder struct to manage a GeoModel world. It holds the pointer to the
+///        root volume of the world and provides additional links to dedicated
+///        volumes inside the tree representing the sensors and chambers in a
+///        detector.
 struct GeoModelTree {
-  std::shared_ptr<GeoModelIO::ReadGeoModel> geoReader = nullptr;
-  PVConstLink worldVolume = nullptr;
-  std::string worldVolumeName = "World";
+  using FPVLink = GeoIntrusivePtr<GeoFullPhysVol>;
```
paulgessinger: Nitpick: conventionally: `FpvLink`.

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelTree.hpp
```diff
@@ -9,17 +9,85 @@
...
+  using FPVConstLink = GeoIntrusivePtr<const GeoFullPhysVol>;
+  /// @brief Helper struct to manage the list of published full physical volumes.
+  ///        These volumes have unique name tags and seed the construction of
+  ///        tracking volumes. The list of published full physical volumes are
+  ///        the name tags associated with a particular node in the GeoModel
+  ///        tree. The lists are usually separated by subdetector and are
+  ///        shipped together with the SQLite database. The latter is querried,
```
paulgessinger:
```suggestion
  ///        shipped together with the SQLite database. The latter is queried,
```

Plugins/GeoModel/include/Acts/Plugins/GeoModel/GeoModelTree.hpp
```diff
@@ -9,17 +9,85 @@
...
+      return m_reader;
+    }
+
+   private:
+    /// @brief Pointer to the ReadGeoModel instance
+    std::shared_ptr<GeoModelIO::ReadGeoModel> m_reader{};
+    /// @brief Map of all published full phyical volumes per subsytstem
+    mutable PublisherMap_t m_publishedVols{};
+    /// @brief Mutex for thread-safe access of the published volumes
+    mutable std::mutex m_mutex{};
```
paulgessinger: Does this have to be `mutable` and mutexed? I imagine the geometry construction only runs in a single thread almost all of the time, so you could just work with a single mutable instance that does not have to have internal mutable members. Or is this used (mutated) during event processing as well?

junggjo9: Ideally, one would know the list of all published systems in advance, but unfortunately, that's not the case. The published voluems are lazily populated and given that GeomodelTree is in many applications const, the publisher is it as well and that's kind of a workaround

paulgessinger:
@junggjo9 good to go? I see the mutability was changed as discussed, did you want to change anything else?

junggjo9:
Fine to go

PR #4374, opened 2025-07-01
test: `SympyStepper` const field integration test

andiwand:
Similar to the `EigenStepper` integration test, which compares the semi-analytical covariance transport against the ridder approach, this adds testing for the `SympyStepper`


andiwand:
closing for now as this is more cumbersome than expected

PR #4378, opened 2025-07-01
ci: Remove Ubuntu22.04

paulgessinger:
Remove Ubuntu 22.04 jobs from CI coverage.

--- END COMMIT MESSAGE ---

## Summary by CodeRabbit

* **Chores**
  * Updated the build and test environments to use a newer Ubuntu 24.04 base image.


paulgessinger:
This still needs a new find module for onnx. The decision was not to patch the one that's in the binary releases but handle it on our side.

PR #4386, opened 2025-07-03
refactor!: Replace `ObjectIntersection<T>` with `SurfaceIntersection`

andiwand:
`ObjectIntersection<T>` is effectively only used with `T = Surface` and seems too generic for this specific use case. As such the template was removed and the class was renamed to `SurfaceIntersection`.

Additionally the interface was changed to take and return references over pointers while still storing a pointer to support trivial copy and move.

--- END COMMIT MESSAGE ---


Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -190,104 +189,103 @@ class ObjectIntersection {
     return m_intersection.status();
   }
 
-  /// Returns the object that has been intersected
-  constexpr const object_t* object() const { return m_object; }
+  /// Returns the surface that has been intersected
+  constexpr const Surface& surface() const { return *m_surface; }
```
paulgessinger, line 206: I *think* this is the only actual API breaking change? We could add an extra `[[deprecated]]` method `object()` for compatibility and call it non-breaking. What do you think?

andiwand: makes sense! but also `ObjectIntersection<T>` would go away - not sure if anybody is using this but removing the type feels breaking to me.

except we play the card `Utilities` is not part of the versioning 😄

paulgessinger: I mean even if it's technically breaking, limiting the breakage still seems beneficial to me. We could also have a template `ObjectIntersection<T>` that inherits from the new `SurfaceIntersection` and deprecate that whole type if we want to overdo things.

andiwand: ah I see - makes sense. we can add it back in and deprecate it.

adding `ObjectIntersection<T>` back in and have it convertible from `SurfaceIntersection` would make it non breaking I guess. depends a bit how long we need to wait for a new major release I suppose

acts-project-service:
## :red_circle: Athena integration test results [40e02d18085c845feae61a3589b4e2eebc2f3fa7]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/59285577) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12520015)!

paulgessinger (approved):
Lgtm

PR #4387, opened 2025-07-04
chore: moved material decorator

asalzburger:
This PR makes the ROOT based material decorator accessible through the `Root` plugin instead of the `Examples/Io/Root` directory and hence allows clients to use it w/o relying on the examples to be built.

--- END COMMIT MESSAGE ---


asalzburger:
Superseded.

PR #3989, opened 2024-12-16
fix: Prefer external surface candidates in Navigation

andiwand:
External surfaces get infinite bounds during navigation and should always be preferred over any other surface in the geometry. This is not the case right now as a surface which is flagged as external might live on the same plane as a standard surface and we might still end up targeting the standard one first.

In this PR I propose to sort surfaces not only by path length but also by the fact if they are external or not. This will prioritize external surfaces over standard ones.

This is important during refitting with for example the KF. In case the estimated trajectory aims for a neighboring surface we would observe holes instead of the actual measurements.

PR #4389, opened 2025-07-04
refactor: Remove overlapping surfaces during navigation

andiwand:
Overlapping surfaces are not supported by the current navigation and propagation model. This change explicitly resolves surface overlaps during resolving compatible surfaces.

--- END COMMIT MESSAGE ---

### DIFF
--- a/Core/include/Acts/Clusterization/Clusterization.hpp
+++ b/Core/include/Acts/Clusterization/Clusterization.hpp
@@ -23,14 +23,14 @@ concept HasRetrievableRowInfo = requires(Cell cell) {
   { getCellRow(cell) } -> std::same_as<int>;
 };
 
-template <typename Cell>
-concept HasRetrievableLabelInfo = requires(Cell cell) {
-  { getCellLabel(cell) } -> std::same_as<int&>;
-};
-
 template <typename Cell, typename Cluster>
 concept CanAcceptCell = requires(Cell cell, Cluster cluster) {
   { clusterAddCell(cluster, cell) } -> std::same_as<void>;
+};
+
+template <typename Cluster>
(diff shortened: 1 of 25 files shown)


Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -686,6 +686,16 @@ class Navigator {
           state.options.geoContext, position, direction, navOpts);
       std::ranges::sort(state.navSurfaces,
                         SurfaceIntersection::pathLengthOrder);
+      // For now we implicitly remove overlapping surfaces.
+      // For track finding it might be useful to discover overlapping surfaces
+      // and check for compatible measurements. This is under investigation
+      // and might be implemented in the future.
+      auto toBeRemoved = std::ranges::unique(
```
asalzburger, line 716: mmmm, we have no control if we remove the external or the non-external surface in this way, right?

andiwand: right, this would only be available after https://github.com/acts-project/acts/pull/3989

but I would argue in any kind of overlap situation we should only use the first one and discard all the other ones

after https://github.com/acts-project/acts/pull/3989 the external surface will always be the first in the list, and therefore kept

asalzburger (approved):

acts-project-service:
## :red_circle: Athena integration test results [cfe068afd60c8ef59cce7c2860293b77b8741c70]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58482500) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12393160)!

PR #4395, opened 2025-07-07
feat: Json I/O for `Seeding2` configs

andiwand:
Adds Json I/O for `Seeding2` option and config structs. The plan is to dump data and config with these utilities and test / benchmark individual parts of the seeding offline / outside a full reconstruction chain.

--- END COMMIT MESSAGE ---


Core/include/Acts/Utilities/GridBinFinder.hpp
```diff
@@ -49,6 +53,11 @@ class GridBinFinder {
     storeValue(std::forward<args>(vals)...);
   }
 
+  explicit GridBinFinder(const std::array<stored_values_t, DIM>& values)
```
paulgessinger: Could this be private and the corresponding converter function made a friend?

Plugins/Json/src/Seeding2ConfigJsonConverter.cpp
```diff
@@ -0,0 +1,270 @@
...
+  for (const auto& value : f.values()) {
+    if (std::holds_alternative<int>(value)) {
+      values.push_back(std::get<int>(value));
+    } else if (std::holds_alternative<std::pair<int, int>>(value)) {
+      values.push_back(std::get<std::pair<int, int>>(value));
+    } else if (std::holds_alternative<std::vector<std::pair<int, int>>>(
+                   value)) {
+      values.push_back(std::get<std::vector<std::pair<int, int>>>(value));
+    }
+  }
```
paulgessinger: Is this not easier with `std::visit`?

Core/include/Acts/Utilities/GridBinFinder.hpp
andiwand: I guess we would need to forward declare json types then?

To me this construction seems less troublesome then the other one, so I think the new constructor is beneficial. I guess I could somehow use the other constructor as well but it seems unnecessarily complicated

paulgessinger: I'm only worried about proliferated use of opaque constructors in other code. We've had this happen in the past, that's why I'm cautious.

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [3fe3a7024f29c43f1081a48407658c0e1b84b321]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58506077) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12396104)!

PR #4399, opened 2025-07-08
feat: Allow dropping extra columns in `SpacePointContainer2`

andiwand:
- adds `dropExtraColumns` to allow for dropping extra known columns
- adds `dropExtraColumn` to allow for dropping extra named columns
- fix some `noexcept`
- simplify `createExtraColumns` implementation a bit

--- END COMMIT MESSAGE ---


Core/include/Acts/EventData/SpacePointContainer2.hpp
```diff
@@ -605,26 +611,28 @@ class SpacePointContainer2 {
   /// @return A reference to the newly created column.
   /// @throws std::runtime_error if a column with the same name already exists.
   template <typename T>
-  SpacePointExtraColumnProxy<T> createExtraColumn(
-      const std::string &name) noexcept {
+  SpacePointExtraColumnProxy<T> createExtraColumn(const std::string &name) {
```
paulgessinger, line 614: `noexcept` not applicable? Does it throw if the column already exists?

Core/src/EventData/SpacePointContainer2.cpp
```diff
@@ -148,60 +148,109 @@ void SpacePointContainer2::createExtraColumns(
...
   m_knownExtraColumns = m_knownExtraColumns | columns;
+
+  initializeExtraColumns();
+}
+
+void SpacePointContainer2::dropExtraColumns(
+    SpacePointKnownExtraColumn columns) noexcept {
+  using enum SpacePointKnownExtraColumn;
+
+  if ((columns & R) != None) {
```
paulgessinger: Not only in the diff here, but you could use [`ACTS_CHECK_BIT`](https://github.com/acts-project/acts/blob/main/Core/include/Acts/Utilities/Helpers.hpp#L21) for these expressions to make it more legible. I always have a hard time parsing these binary ops at a glance.

```suggestion
  if (ACTS_CHECK_BIT(columns, R)) {
```

The logic is slightly different, instead of `!= 0` the macro compares with the mask itself, but I think if should give the same result.

Core/include/Acts/EventData/SpacePointContainer2.hpp
andiwand: right this version actually throws if `name` is already booked

paulgessinger (approved):

PR #4404, opened 2025-07-08
refactor: Couple of improvements to `SpacePointContainer2`

andiwand:
- Rearrange code
- All `SpacePointContainer2` columns become optional
- Drops `*extra*` from naming
- Direct access to column data
- Subset space point iterator
- Column `zip`ped iterator range
- Add `ConstSpacePointColumnProxy<T>` and `MutableSpacePointColumnProxy<T>`

--- END COMMIT MESSAGE ---


paulgessinger:
I also see some changes to the way that the source links are stored?

Core/include/Acts/EventData/SpacePointContainer2.hpp
```diff
@@ -686,126 +708,317 @@ class SpacePointContainer2 {
...
-  ConstRange range(const IndexRangeType &range) const noexcept {
+  /// Creates a range of space points from the given index range.
+  /// @param range The index range to create the range from.
+  /// @return A const range of space points.
+  ConstRange range(const IndexRange &range) const noexcept {
     return ConstRange(*this, range);
   }
 
+  template <bool read_only>
+  class Subset {
```
paulgessinger, line 747: I'm not sure I follow how this is different from the range above.

andiwand: subset uses a collection of indices while range uses only a lower and upper bound index

Core/include/Acts/EventData/SpacePointContainer2.hpp
```diff
@@ -28,55 +30,66 @@ namespace Acts::Experimental {
...
+  }
 
  private:
-  const ContainerType *m_data{};
+  const SpacePointContainer2 *m_container{};
+  const Column *m_column{};
 
-  ContainerType &data() { return const_cast<ContainerType &>(*m_data); }
-  const ContainerType &data() const { return *m_data; }
+  Column &column() { return const_cast<Column &>(*m_column); }
```
paulgessinger: Hm, any way we can avoid the `const_cast` here? Can't you change `m_column` to `Column*`? The const getter should be able to cast to const, and this getter should be able to return as is.

paulgessinger:
@andiwand could you have a quick look at the sonar issues? I think there are a couple of legitimate ones in there.

Core/include/Acts/EventData/SpacePointContainer2.hpp
andiwand: pushing this up creates even more issues because the proxy is often created in a const context

andiwand: I removed the public data accessors again which should make this safe again. `const_cast` is still there because I do not see a better solution

paulgessinger: I'm not sure I follow why this is needed. I think this would be the first reintroduction of a const cast in the code aside from the geometry where we never got rid of it. We've been pushing back against this, so if there's any other way I think it's worth exploring.

andiwand: happy to change it if you have an alternative in mind

andiwand:
Added `ConstSpacePointColumnProxy<T>` and `MutableSpacePointColumnProxy<T>` to propagate the `const`ness

Core/include/Acts/EventData/SpacePointColumnProxy2.hpp
```diff
@@ -0,0 +1,97 @@
...
+  Span data() const noexcept { return Span(column().data(), column().size()); }
+
+  /// Returns a reference to the column entry at the given index.
+  /// If the index is out of range, an exception is thrown.
+  /// @param index The index of the space point to access.
+  /// @return A reference to the column entry at the given index.
+  /// @throws std::out_of_range if the index is out of range.
+  Reference at(Index index) const {
+    if (index >= column().size()) {
+      throw std::out_of_range("Index out of range");
```
junggjo9:
```suggestion
      throw std::out_of_range("Index "+std::to_string(index)+ " out of range [0-"+std::to_string(column().size()) + ")" );
```
Then the user has a bit of a clue ;)

Core/include/Acts/EventData/SpacePointColumnProxy2.hpp
```diff
@@ -0,0 +1,97 @@
...
+template <typename T, bool read_only>
+class SpacePointColumnProxy {
+ public:
+  constexpr static bool ReadOnly = read_only;
+  using Index = SpacePointIndex2;
+  using Value = const_if_t<ReadOnly, T>;
+  using Container = const_if_t<ReadOnly, SpacePointContainer2>;
+  using Column = const_if_t<ReadOnly, std::vector<T>>;
+  using Span = std::span<Value>;
+  using Reference = Value &;
```
junggjo9: Small question, the c++ standard uses reference_type, in case that thingy is passed through typetraits. and so on. Also value_type... I guess that's not intended to be possible, right?

Core/include/Acts/EventData/SpacePointColumnProxy2.hpp
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+ private:
+  Container *m_container{};
+  Column *m_column{};
+
+  Column &column() const noexcept
+    requires(!ReadOnly)
+  {
+    return *m_column;
```
junggjo9, line 135:
```suggestion
    assert(m_column);
    return *m_column;
```
If it's possible at all to have a nullptr to the column?

junggjo9 (commented):
One minor comment and one question out of curiosty.

Core/include/Acts/EventData/SpacePointColumnProxy2.hpp
andiwand: good point - I believe this should not be compatible with any STL function. at least I would not know which STL tool would be useful here

Core/include/Acts/EventData/SpacePointColumnProxy2.hpp
andiwand: *should* be protected by the constructor
```c++
  SpacePointColumnProxy(Container &container, Column &column)
      : m_container{&container}, m_column(&column) {}
```
as long as nobody "dereferences" a `nullptr`. but we could check it in the constructor if you think this is useful!

Core/include/Acts/EventData/SpacePointColumnProxy2.hpp
andiwand: yeah good point - let me do that

Core/include/Acts/EventData/SpacePointProxy2.hpp
```diff
@@ -67,120 +76,117 @@ class SpacePointProxy2 {
...
     return m_container->sourceLinks(m_index);
   }
   /// Mutable access to the x coordinate of the space point.
   /// @return A mutable reference to the x coordinate of the space point.
-  float &x() noexcept
+  float &x() const noexcept
     requires(!ReadOnly)
   {
     return m_container->x(m_index);
   }
```
paulgessinger, line 98: I won't ask you to complicate this any further, so feel free to disregard, but this means if you have a const-ref to a mutable proxy, you can mutate the contents, right?

andiwand: yeah I was a bit confused what is the right thing to do here... it felt odd to have a `const` and non-`const` version because it should rather be decided by `ReadOnly`. but if I only do it by `ReadOnly` and use `require(!ReadOnly)`, the signature of the `const` and non-`const` access become the same.

I think in the track edm we use the `const` context of the proxy, right?

Core/include/Acts/EventData/SeedContainer2.hpp
```diff
@@ -156,7 +156,7 @@ class SeedContainer2 {
     IndexType m_index{};
 
     friend bool operator==(const SeedIterator &a, const SeedIterator &b) {
-      return a.m_index == b.m_index && a.m_container == b.m_container;
+      return a.m_index == b.m_index;
```
paulgessinger, line 159: We're assuming comparing iterators from different containers won't be an issue?

paulgessinger (approved):

Core/include/Acts/EventData/SeedContainer2.hpp
andiwand: I would argue that this is the callers responsibility. I believe STL does the same - even though that should not mean we have to do it. But I think it would be more reasonable to throw or assert in this case because someone is doing something terribly wrong...

acts-project-service:
## :white_check_mark: Athena integration test results [e76b5bc4769cc3397469206f03436acd51f44796]

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798924)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798923)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798922)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798921)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798920)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798919)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798918)||
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798917)||
|:green_circle: | [test_ActsCheckObjectCountsWorkflowHgtd](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798916)||
|:green_circle: | [test_ActsCheckObjectCountsCachedWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798915)||
|:green_circle: | [test_ActsCheckObjectCountsLegacyWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798914)||
|:green_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798913)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798912)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798911)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798910)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798909)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798908)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798907)||
|:green_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798906)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798905)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798904)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798903)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798902)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798901)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798900)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798899)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798898)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798897)||
|:yellow_circle: | [test_ActsGx2fRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798896)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798895)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798894)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798893)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58798892)||

PR #4415, opened 2025-07-10
build: Add compile check for std::format

paulgessinger:
We want to be able to use `std::format`. It's not supported below GCC12. Instead of doing a compiler version check, I'm adding here a feature check that tries to compile a file with `std::format`. If it fails, you'll get an error message at CMake time, instead of later on at compile time.

This required https://github.com/acts-project/acts/pull/4378 and will fail CI until then because we still have GCC11 builds.


paulgessinger:
@andiwand @CarloVarni do we call this breaking?

andiwand:
to me `std::format` looks like a cpp20 feature. since we require cpp20 I would argue it is not breaking even if some compiler see it differently

paulgessinger:
@andiwand technically you're right. What I'm trying to avoid here is a confusing error in gcc11 and down that will happily take the c++20 flag but will just not have format. 

I could do a compiler version check to give a better message, but this feature based check seems more robust.

andiwand (approved):

acts-project-service:
## :red_circle: Athena integration test results [52f80a47fb3c7fb5091659a748f1e7edae58a45e]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/58674800) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12421163)!

PR #4400, opened 2025-07-08
feat: move ROOT surface material map I/O to plugins

asalzburger:
This PR moves the I/O for material map reading/writing into `Plugins/Root` and removes the duplicated code from `RootMaterialWriter` and `RootMaterialDecorator`.

It also harmonizes the usage/definition of `SurfaceMaterialMaps`. `VolumeMaterialMaps`, and `DetectorMaterialMaps` consistently by declaring them once.

It is backward compatible, and introduces a new, common ROOT I/O material backend that can be shared between all material classes (homongeneous, binned, grid, indexed).

This PR also allows UnitTests to be added for Material writing, the first ones are done with it, and it removes roughly 100 SonarCloud issues, mainly related to how we call and use ROOT.

This will be followed by a PRs doing the same for Volume material and updated Grid based surface material.

The reading gis kept backward-compatible.

--- END COMMIT MESSAGE ---

Any further description goes here, @-mentions are ok here!

- Use a *conventional commits* prefix: [quick summary](https://www.conventionalcommits.org/en/v1.0.0/#summary)
  - We mostly use `feat`, `fix`, `refactor`, `docs`, `chore` and `build` types.
- A milestone will be assigned by one of the maintainers

PR #4416, opened 2025-07-11
feat: add unittest for ROOT based material track IO

asalzburger:
Since The MaterialTrack ROOT I/O is now in Plugins, one can define Unit Tests, voila.

Sits on top of #4400 

--- END COMMIT MESSAGE ---


asalzburger:
superseded.

PR #4132, opened 2025-03-13
feat: Implement volume material propagation with `SympyStepper`

andiwand:
Implements volume material propagation with proper continuous energy loss, energy loss uncertainty and multiple scattering. This required piping an additional covariance term through `CovarianceEngine` and the steppers which is modeled as a `std::optional` for now. The `EigenStepper` was not changed for now and could be adapted similarly to the `SympyStepper`. `MaterialAccumulator` was added which is a tool to handle increasing material during the stepping. The `SympyStepper` gained another generated stepping function which deals with volume material. It automatically switches to this function when volume material is detected.

--- END COMMIT MESSAGE ---

PR #4419, opened 2025-07-12
chore: Bump codegen python package version

andiwand:
After changing the `codegen` package in https://github.com/acts-project/acts/pull/4132 I forgot to raise the version which resulted in downstream build problems for users. This PR bumps the version

--- END COMMIT MESSAGE ---


codegen/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "codegen"
-version = "0.1.0"
+version = "0.2.0"
```
paulgessinger, line 3: Ah, I didn't appreciate that this was the problem. My bad. I'm note sure there's a way around this. The codegen installs this package because we're using it from a number of scripts, and I didn't see a better way to make it importable.

paulgessinger (approved):

PR #4435, opened 2025-07-15
chore: introduce `ActsPython` 

asalzburger:
This PR is the first of a proposed set of streamlining and (re-)ordering the python bindings.

It introduces a dedicated Python Directory, following the remaining structure of the repository:

| *Directory* | *Namespace* |
| `Core` | `Acts` |
| `Plugins` | `Acts` (we could consider `ActsPlugins`   |
| `Alignment` | `ActsAlignment` |
| `Examples` | `ActsExamples` |
| `Fatras` |  `ActsFatras` |
| `Python` | `ActsPython` |

Within the `Python` directory there will be:

* `Utilities` which replaces (with this PR) `Examples/Python/Utilities`
* `Core` which will hold the python bindings for `Core` classes and functionality, and will be accessible as `acts.ClassName`
* `Plugins` which will hold the python bindings for the `Plugins` and will be accessed with `acts.<Plugin>.`
* `Examples` which hold everything from the `Examples` package and will be accessible as `acts.examples.*`

The PR already clears some inconsistent usage of `namespace` and will allow to unit test some of the core functionality.

--- END COMMIT MESSAGE ---

@paulgessinger @andiwand - I would really like your insight / comments on this !

Proposed PRs:

```mermaid
graph TD
    A[ PR #4435 : Introduce ActsPython ] --> AO[ PR #4440: Introduce ActsPython/Module ]
    AO --> B[ PR #4448: Introduce Python/Core ]
    B --> C[ Move Plugin binding to Python/Plugins ]
    C --> D[ Move Example binding to Python/Examples ]
    D --> E[ Move tests and remove Examples/Python ]
```


Examples/Python/src/Base.cpp
```diff
@@ -114,10 +114,10 @@ class PythonLogger {
 
  private:
   std::string m_name;
-  std::unique_ptr<const Logger> m_logger;
+  std::unique_ptr<const Acts::Logger> m_logger;
```
paulgessinger, line 117: Consider `using namespace Acts` which should be safe in `.cpp` files. Purely for conciseness.

Examples/Python/src/ModuleEntry.cpp
```diff
@@ -102,53 +101,53 @@ PYBIND11_MODULE(ActsPythonBindings, m) {
     mv.attr("commit_hash_short") = Acts::CommitHashShort;
   }
```
paulgessinger: You could do `using namespace ActsPython` inside this function for conciseness.

Python/Examples/include/ActsPython/Examples/Helpers.hpp
```diff
@@ -0,0 +1,20 @@
...
+#include "ActsExamples/Framework/AlgorithmContext.hpp"
+#include "ActsExamples/Framework/ProcessCode.hpp"
+
+namespace ActsPython::Concepts {
+template <typename T>
+concept has_write_method =
+    requires(T& t, const ActsExamples::AlgorithmContext& ctx) {
+      { t.write(ctx) } -> std::same_as<ActsExamples::ProcessCode>;
+    };
+}  // namespace ActsPython::Concepts
```
paulgessinger: In my opinion it's questionable if this needs to be in a separate file if it's only used in `Macros.hpp` in one location.

paulgessinger (commented):
I don't fully understand the structure yet. The folder structures:

- `Examples/Python` and
- `Python/Examples`

don't make sense to me. The overall python bindings target is still created in `Examples/Python/CMakeLists.txt`, and contains all `.cpp` source files, so why does it spill out into the other folder?

Maybe there's some larger purpose for this that I'm not getting yet.

Python/Examples/include/ActsPython/Examples/Helpers.hpp
asalzburger: I've put it into a different file, because this one has Framework dependencies (I can get around within the macros without dependency).

Examples/Python/src/Base.cpp
asalzburger: Indeed, though for the moment I wanted to have them explicitly to see where things come from.
The aim of this exercise is to split dependencies into the dedicated Python modules as well.

asalzburger:
> I don't fully understand the structure yet. The folder structures:
> 
> * `Examples/Python` and
> * `Python/Examples`
> 
> don't make sense to me. The overall python bindings target is still created in `Examples/Python/CMakeLists.txt`, and contains all `.cpp` source files, so why does it spill out into the other folder?
> 
> Maybe there's some larger purpose for this that I'm not getting yet.

Oh - `Examples/Python` will not exist anymore when this is done. But I can never do that in a single PR. This one is the first one and I propose to then, in a second step shuffle the Core related part over.

I have included a schematic of the proposed PRs to the description of this PR,

The final structure should be:

`Python` directory - contains everything for the python bindings + the pytests 
It has the following sub-directories:
 * `Utilities` : context and macros
 * `Module` : the master that binds in the module entries from the others
 * `Core`: python bindings for `Core` classes
 * `Plugins`: python bindings for `Plugins` stuff
 * `Fatras`: python bindings for Fatras
 * `Examples`: python bindings for everything in the Examples 
 * `Tests`: all py test related things

If E.g. Examples are not build, a stub is built and screen info indicates the setup.

You could then build Core + Plugins (e.g. Json) + Python bindings and work with Json material maps and manipulate them, without even building the Examples.

Examples/Python/src/ModuleEntry.cpp
asalzburger: Indeed, though this file will disappear, so I will do some cleanup, once everything is moved.

asalzburger: Done for the moment.

Python/Examples/include/ActsPython/Examples/Helpers.hpp
asalzburger: Moved back to `Macros.hpp` - which worked by forward declaring the Examples dependencies.

asalzburger:
Will be superseded.

PR #4447, opened 2025-07-17
refactor: Generalized template unrolling for `DoubletSeedFinder` implementation

andiwand:
Uses `Boost.Mp11` to unroll templates of the `DoubletSeedFinder` implementation. The doublet finding is quite hot and at the same time supposed to be flexible. With increasing flexibility we add more and more code to a hot function which will slow down other use cases. The proposed alternative here is to use N template switches which all get instantiated and dispatched during runtime. The implementation here decides on the appropriate function based on the config once on instantiation and then uses virtual dispatch into the appropriate one.

--- END COMMIT MESSAGE ---

PR #4450, opened 2025-07-18
refactor: Generalize sorting requirement for `BroadTripletSeedFinder`

andiwand:
With https://github.com/acts-project/acts/pull/4447 the sorting requirement of the `DoubletSeedFinder` becomes an option. The same can now be done for the `BroadTripletSeedFinder`. This generalizes the use of the `BroadTripletSeedFinder` and will be beneficial for the module map approach.

--- END COMMIT MESSAGE ---

blocked by
- https://github.com/acts-project/acts/pull/4447

### DIFF
--- a/Core/src/Seeding2/BroadTripletSeedFinder.cpp
+++ b/Core/src/Seeding2/BroadTripletSeedFinder.cpp
@@ -68,300 +68,30 @@ bool stripCoordinateCheck(float tolerance, const ConstSpacePointProxy2& sp,
   return true;
 }
 
-}  // namespace
-
-BroadTripletSeedFinder::DerivedTripletCuts::DerivedTripletCuts(
-    const TripletCuts& cuts, float bFieldInZ_)
-    : TripletCuts(cuts), bFieldInZ(bFieldInZ_) {
-  // similar to `theta0Highland` in `Core/src/Material/Interactions.cpp`
-  {
-    const double xOverX0 = radLengthPerSeed;
-    const double q2OverBeta2 = 1;  // q^2=1, beta^2~1
-    // RPP2018 eq. 33.15 (treats beta and q² consistently)
-    const double t = std::sqrt(xOverX0 * q2OverBeta2);
-    // log((x/X0) * (q²/beta²)) = log((sqrt(x/X0) * (q/beta))²)
-    //                          = 2 * log(sqrt(x/X0) * (q/beta))
-    highland =
-        static_cast<float>(13.6_MeV * t * (1.0 + 0.038 * 2 * std::log(t)));
-  }
-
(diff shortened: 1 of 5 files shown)


Core/src/Seeding2/BroadTripletSeedFinder.cpp
```diff
@@ -531,15 +277,28 @@ void BroadTripletSeedFinder::createTriplets(
...
+/// @param filter Triplet seed filter that defines the filtering criteria
+/// @param filterState State object that holds the state of the filter
+/// @param filterCache Cache object that holds memory used in SeedFilter
+/// @param spacePoints Space point container
+/// @param spM Space point candidate to be used as middle SP in a seed
+/// @param bottomDoublets Bottom doublets to be used for triplet creation
+/// @param topDoublets Top doublets to be used for triplet creation
+/// @param tripletTopCandidates Cache for triplet top candidates
+/// @param candidatesCollector Collector for candidates for middle space points
+void createStripTriplets(
```
paulgessinger, line 293: I this function local only to the `.cpp` file now? If so, documentation is optional, and maybe put it in an anonymous namespace?

Core/src/Seeding2/BroadTripletSeedFinder.cpp
```diff
@@ -775,4 +534,231 @@ void BroadTripletSeedFinder::createStripTriplets(
   }  // loop on bottoms
 }
 
+template <typename SpacePointCollections>
+void createSeedsFromGroupsImpl(
```
paulgessinger, line 538: Anonymous namespace?

paulgessinger, line 538: Function content is unchanged?

Core/src/Seeding2/BroadTripletSeedFinder.cpp
andiwand: should be in the anonymous namespace already

the doc is also a bit repetitive. I will think about dropping it in the next PRs

Core/src/Seeding2/BroadTripletSeedFinder.cpp
andiwand: should be in an anonymous namespace already

andiwand: Right, the function is the same except for the generalized input

acts-project-service:
## :red_circle: Athena integration test results [b426d731bbf8159002a6433f6ff2090aa537e775]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/59098698) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12492653)!

PR #4452, opened 2025-07-18
build!: Use function to consistently add ALIAS targets

paulgessinger:
This PR introduces a function `acts_add_library` that is used for all library targets. It forwards all arguments to `add_library`, but it prepends `Acts` to the library name that was given. In addition, it will create an ALIAS target `Acts::` so that we can use namespaced target names, which are generally safer.

--- END COMMIT MESSAGE ---


Alignment/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ target_include_directories(
         $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 
-target_link_libraries(ActsAlignment PUBLIC ActsCore)
+target_link_libraries(ActsAlignment PUBLIC Acts::Core)
 
 install(
```
krasznaa: Would it not make sense to put some of these common operations behind the `acts_add_library(...)` function as well? I know not everybody agrees, but I myself strongly believe that this greatly simplifies the maintenance of the R&D build configuration.

I guess you already looked, but have a look at `vecmem_add_library(...)`. (https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-functions.cmake) That performs a few different things that are done in the same way for the "installed vecmem libraries".

cmake/ActsFunctions.cmake
```diff
@@ -1,3 +1,22 @@
...
+    add_library(${full_target_name} ${ARGN})
+
+    # Create alias target with Acts:: namespace
+    add_library(Acts::${library_name} ALIAS ${full_target_name})
+
+    set_target_properties(
+        ${full_target_name}
+        PROPERTIES EXPORT_NAME Acts::${library_name}
+    )
+endfunction()
```
krasznaa, line 44: 🤔 I do think it would make sense to move all the library related `install(...)` commands in here.

krasznaa (commented):
As I said in the morning, I'm happy with this direction. I think the `vecmem::core`, `detray::core`, etc. formalism is serving us well in the R&D projects. I believe this formalism will make the Acts code a little nicer as well.

Alignment/CMakeLists.txt
paulgessinger: I'm on the fence with the `install` directives (see other comment), but I think in terms of other properties, dependencies, I think I'd rather keep the `acts_add_library` function as minimal as possible, even if it means duplicating cmake instructions in a number of places.

krasznaa: I'm only proposing to move the `install(...)` commands. Since those are really meant to be done in the same way for all of these libraries.

In vecmem we also generate export headers, and do some stuff with `SOVERSION`. But those are not relevant here...

paulgessinger: We use different target paths for the installed headers (`Acts`, `ActsExamples`) for the same reason we kept the namespaces separate (as discussed above), so I can't easily put them in the function without adding extra arguments

Tests/CommonHelpers/CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
 # get the path of the test data directory
 file(TO_NATIVE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../Data" acts_test_data_dir)
 
-add_library(
-    ActsTestsCommonHelpers
+acts_add_library(
+    TestsCommonHelpers
```
krasznaa: It may not make sense to use `acts_add_library(...)` for internal libraries. 🤔

In the R&D projects I just fall back on `add_library(...)` for such things.
  - https://github.com/acts-project/traccc/blob/main/examples/run/CMakeLists.txt#L11
  - https://github.com/acts-project/vecmem/blob/main/tests/CMakeLists.txt#L25
  - etc.

In this project the `EXPORT` statements are not actually hurtful (unlike how they would be in the R&D code), but still, it is probably wasteful to set up exports for libraries that never get exported.

Examples/Scripts/MaterialMapping/CMakeLists.txt
```diff
@@ -1,16 +1,11 @@
...
     ROOT::Tree
     ROOT::TreePlayer
     Boost::program_options
     nlohmann_json::nlohmann_json
 )
-
-install(
-    TARGETS ActsAnalysisMaterialComposition
-    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
-)
```
krasznaa: Did you mean to remove this? 😕

Examples/Io/Root/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
-add_library(
-    ActsExamplesIoRoot
+acts_add_library(
+    ExamplesIoRoot
```
krasznaa, line 2: This library was also not exported so far. It may be better to not bother it.

krasznaa (commented):
I think it brings up a deeper question on whether we want to export example libraries. 🤔 I don't know... But this PR may be opening some cans of worms with that...

I'm still very much on board, let's just be careful with the update. It's very easy to make subtle mistakes when changing so many CMake lines at the same time...

Examples/Io/Root/CMakeLists.txt
paulgessinger: Yeah true.

paulgessinger: Actually: wasn't it? It was running

```cmake
install(TARGETS ActsExamplesIoRoot LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
``` 

at the end of the file.

Examples/Scripts/MaterialMapping/CMakeLists.txt
paulgessinger: Don't think so. I'll put it back.

Tests/CommonHelpers/CMakeLists.txt
paulgessinger: You're probably right on this.

Examples/Io/Root/CMakeLists.txt
krasznaa: It is installed, but not exported as far as I can tell.

This may very well be a bug in the current configuration. So now making it possible to export it, is probably a good thing. In general I think this PR highlights that there are some inconsistencies currently with how some of the libraries are handled.

In the end, as long as a library is installed, it should also be possible to use in the CMake configuration of a child project. So it's good that an export is added for this example library... I guess. Since I'm still not sure if "example libraries" are meant to be used by client projects at all, in any shape or form. 🤔

paulgessinger: They're not supposed to, but in practice they are. I think in the end I'm ok with exporting them and making them consumable, with the assumption that people know that they're effectively on their own if they use them.

Tests/CommonHelpers/CMakeLists.txt
paulgessinger: Changed

andiwand (dismissed):
should we get this in for v43?

acts-project-service:
## :red_circle: Athena integration test results [2d24085aa7df43c39bf0fba382cad1c4efa6f011]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/59789818) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12596719)!

PR #4453, opened 2025-07-18
feat: RootTrackSummaryWriter can write multiple track collections

paulgessinger:
--- END COMMIT MESSAGE ---

### DIFF
--- a/Examples/Io/Root/include/ActsExamples/Io/Root/RootTrackSummaryWriter.hpp
+++ b/Examples/Io/Root/include/ActsExamples/Io/Root/RootTrackSummaryWriter.hpp
@@ -181,6 +181,7 @@ class RootTrackSummaryWriter final : public IWriter {
   std::vector<float> m_t_z0;
   /// Production radius of majority particle
   std::vector<float> m_t_prodR;
+  std::vector<unsigned int> m_t_pdg;
 
   /// If the track has fitted parameter
   std::vector<bool> m_hasFittedParams;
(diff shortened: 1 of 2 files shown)


andiwand:
Alternatives would be to schedule multiple writers or merge the tracks before writing - what is the use case here?

paulgessinger:
@andiwand the use case is that for the summer student project I'd like to be able to have the same particles reconstructed under different hypothesis and produce a single TTree. Alternatives would be to add an algorithm that can merge track containers I guess.

Would you prefer that? In the end I don't think the changes are overly disruptive, but I can also keep them private for the summer student project.

andiwand:
My main concern is that we add more and more targeted features to the writers which tend to break over time and mix with other features. Especially dealing with N inputs and outputs seems like something that can be avoided and generalized to reduce them to their main function.

Maybe a silly suggestion but would multiple writers and `hadd` also do it?

If you think this has a broader use beyond the project I am happy to get this in.

paulgessinger:
> My main concern is that we add more and more targeted features to the writers which tend to break over time and mix with other features. Especially dealing with N inputs and outputs seems like something that can be avoided and generalized to reduce them to their main function.

True. I guess the problem is we don't have a place to put these changes / features anywhere else. I think keeping them on private branches has the downside that some things are lost unintentionally. I don't really see a lot of harm in this case since the change is a generalization of the functionality we had before. 

I think a separate algorithm for merging track containers could alleviate your concerns maybe?

> Maybe a silly suggestion but would multiple writers and `hadd` also do it?

I think with hadd the tracks then end up in separate events, as the tree uses events as entries.

andiwand:
Ah true our TTrees are not easy to merge...

I agree with you. I believe a merging algorithm would be the most generic solution but we can also refactor this afterwards and go ahead with this one first.

PR #4468, opened 2025-07-22
ci: hotfix for eic-shell failures

veprbl:
Since recent container upgrade we are seeing build failures. This allows to bypass the erroneous process until we figure out a proper fix.

--- END COMMIT MESSAGE ---

cc @paulgessinger

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -463,6 +463,9 @@ jobs:
             echo "::endgroup::"
             popd
 
+            # workaround ccache build picking up Acts installed in the container and causing linker errors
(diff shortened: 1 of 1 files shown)


paulgessinger:
So this is really due to out of date cached object files? Interesting!

veprbl:
> So this is really due to out of date cached object files? Interesting!

No, it must have something to do with options getting re-ordered or otherwise corrupted along the way.

veprbl:
Okay, so with some educated guess, this has now a solution that keeps ccache enabled.

paulgessinger:
Excellent. Let's merge this then! Thanks for looking into this.

PR #4494, opened 2025-07-30
feat: Teach the Navigator to treat free surfaces

junggjo9:
- Allow the support of free surfaces by the navigator during the fit
- Prepare hook to handle the external surfaces in Gen3

--- END COMMIT MESSAGE ---

Tagging: @andiwand, @paulgessinger, @asalzburger

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -411,6 +411,14 @@ endmacro()
 # find packages explicitly for each component even if this means searching for
 # the same package twice. This avoids having complex if/else trees to sort out
 # when a particular package is actually needed.
+if(ACTS_BUILD_PYTHON_BINDINGS)
+    find_package(Python 3.8 REQUIRED COMPONENTS Interpreter Development.Module)
+    if(ACTS_USE_SYSTEM_PYBIND11)
+        find_package(pybind11 CONFIG REQUIRED)
+    else()
+        add_subdirectory(thirdparty/pybind11)
+    endif()
+endif()
 if(ACTS_BUILD_PLUGIN_ACTSVG)
     if(ACTS_USE_SYSTEM_ACTSVG)
         find_package(actsvg ${_acts_actsvg_version} REQUIRED CONFIG)
@@ -420,7 +428,7 @@ if(ACTS_BUILD_PLUGIN_ACTSVG)
 endif()
 if(ACTS_BUILD_PLUGIN_DD4HEP)
     # Explicitly find python so we can more easily override the version
-    find_package(Python 3.8 REQUIRED COMPONENTS Interpreter Development)
+    find_package(Python 3.8 REQUIRED COMPONENTS Interpreter Development.Module)
     find_package(
         DD4hep
         ${_acts_dd4hep_version}
@@ -605,14 +613,6 @@ if(ACTS_BUILD_EXAMPLES)
             COMPONENTS Core Geom Graf GenVector Hist Tree TreePlayer
         )
         check_root_compatibility()
-    endif()
-    if(ACTS_BUILD_PYTHON_BINDINGS)
-        find_package(Python 3.8 REQUIRED COMPONENTS Interpreter Development)
-        if(ACTS_USE_SYSTEM_PYBIND11)
-            find_package(pybind11 CONFIG REQUIRED)
-        else()
-            add_subdirectory(thirdparty/pybind11)
-        endif()
     endif()
     if(ACTS_BUILD_EXAMPLES_DD4HEP AND ACTS_BUILD_EXAMPLES_GEANT4)
         find_package(
--- a/Core/include/Acts/Geometry/DiamondPortalShell.hpp
+++ b/Core/include/Acts/Geometry/DiamondPortalShell.hpp
@@ -38,10 +38,6 @@ class DiamondPortalShell : public PortalShellBase {
 
   /// @copydoc PortalShellBase::fill
   void fill(TrackingVolume& volume) override;
-
-  /// @brief Get the transformation matrix for this polygon shape portal shell
-  /// @return Reference to the transformation matrix
-  virtual const Transform3& transform() const = 0;
 };
 // Output stream operator for the CuboidPortalShell::Face enum
 /// @param os The output stream
@@ -73,11 +69,6 @@ class SingleDiamondPortalShell : public DiamondPortalShell {
 
   /// @copydoc PortalShellBase::label
   std::string label() const override;
-
-  /// @copydoc DiamondPortalShell::transform
-  const Transform3& transform() const override {
-    return m_volume->transform();
-  };
 
  private:
   std::array<std::shared_ptr<Portal>, 8> m_portals;
--- a/Core/include/Acts/Geometry/TrapezoidPortalShell.hpp
+++ b/Core/include/Acts/Geometry/TrapezoidPortalShell.hpp
@@ -40,10 +40,6 @@ class TrapezoidPortalShell : public PortalShellBase {
 
   /// @copydoc PortalShellBase::fill
   void fill(TrackingVolume& volume) override;
-
-  /// @brief Get the transformation matrix for this trapezoid portal shell
-  /// @return Reference to the transformation matrix
-  virtual const Transform3& transform() const = 0;
 };
 
 /// Output stream operator for the TrapezoidPortalShell::Face enum
@@ -77,10 +73,6 @@ class SingleTrapezoidPortalShell : public TrapezoidPortalShell {
 
   /// @copydoc PortalShellBase::label
   std::string label() const override;
-
-  const Transform3& transform() const override {
-    return m_volume->transform();
-  };
 
  private:
   std::array<std::shared_ptr<Portal>, 6> m_portals{};
--- a/Core/include/Acts/Propagator/ActorConcepts.hpp
+++ b/Core/include/Acts/Propagator/ActorConcepts.hpp
@@ -30,7 +30,7 @@ concept ActorHasActWithoutResult = requires(
     const actor_t& a, propagator_state_t& state, const stepper_t& stepper,
     const navigator_t& navigator, Args&&... args) {
   {
-    a.act(state, stepper, navigator, std::move<Args>(args)...)
+    a.act(state, stepper, navigator, std::forward<Args>(args)...)
   } -> std::same_as<Result<void>>;
 };
 
@@ -42,7 +42,7 @@ concept ActorHasActWithResult =
              const stepper_t& stepper, const navigator_t& navigator,
              typename actor_t::result_type& result, Args&&... args) {
       {
-        a.act(state, stepper, navigator, result, std::move<Args>(args)...)
+        a.act(state, stepper, navigator, result, std::forward<Args>(args)...)
       } -> std::same_as<Result<void>>;
     };
 
@@ -54,7 +54,7 @@ concept ActorHasOldVoidInterface =
              const stepper_t& stepper, const navigator_t& navigator,
              Args&&... args) {
       {
-        a.act(state, stepper, navigator, std::move<Args>(args)...)
+        a.act(state, stepper, navigator, std::forward<Args>(args)...)
       } -> std::same_as<void>;
     } ||
     // Check with result parameter (for actors with result_type)
@@ -63,7 +63,7 @@ concept ActorHasOldVoidInterface =
               const stepper_t& stepper, const navigator_t& navigator,
               typename actor_t::result_type& result, Args&&... args) {
        {
-         a.act(state, stepper, navigator, result, std::move<Args>(args)...)
+         a.act(state, stepper, navigator, result, std::forward<Args>(args)...)
        } -> std::same_as<void>;
      });
 
@@ -81,7 +81,7 @@ concept ActorHasAbortWithoutResult = requires(
     const actor_t& a, propagator_state_t& state, const stepper_t& stepper,
     const navigator_t& navigator, Args&&... args) {
   {
-    a.checkAbort(state, stepper, navigator, std::move<Args>(args)...)
+    a.checkAbort(state, stepper, navigator, std::forward<Args>(args)...)
   } -> std::same_as<bool>;
 };
 
@@ -94,7 +94,7 @@ concept ActorHasAbortWithResult =
              typename actor_t::result_type& result, Args&&... args) {
       {
         a.checkAbort(state, stepper, navigator, result,
-                     std::move<Args>(args)...)
+                     std::forward<Args>(args)...)
       } -> std::same_as<bool>;
     };
 
--- a/Core/include/Acts/Propagator/Propagator.hpp
+++ b/Core/include/Acts/Propagator/Propagator.hpp
@@ -105,66 +105,27 @@ class Propagator final
   static_assert(StepperConcept<stepper_t>,
                 "Stepper does not fulfill stepper concept.");
 
- public:
-  /// Type of the stepper in use for public scope
-  using Stepper = stepper_t;
-
-  /// Type of the navigator in use for public scope
-  using Navigator = navigator_t;
-
-  /// Type of state object used by the propagation implementation
-  using StepperState = typename Stepper::State;
-
-  /// Typedef the navigator state
-  using NavigatorState = typename navigator_t::State;
-
-  /// Type alias for propagator state combining stepper and navigator states
-  template <typename propagator_options_t, typename... extension_state_t>
-  using State = PropagatorState<propagator_options_t, StepperState,
-                                NavigatorState, extension_state_t...>;
-
-  /// Type alias for stepper configuration options
-  using StepperOptions = typename stepper_t::Options;
-
-  /// Type alias for navigator configuration options
-  using NavigatorOptions = typename navigator_t::Options;
-
-  /// Type alias for propagator configuration options with actor list
-  template <typename actor_list_t = ActorList<>>
-  using Options =
-      PropagatorOptions<StepperOptions, NavigatorOptions, actor_list_t>;
-
-  /// Constructor from implementation object
-  ///
-  /// @param stepper The stepper implementation is moved to a private member
-  /// @param navigator The navigator implementation, moved to a private member
-  /// @param _logger a logger instance
-  explicit Propagator(stepper_t stepper, navigator_t navigator = navigator_t(),
-                      std::shared_ptr<const Logger> _logger =
-                          getDefaultLogger("Propagator", Acts::Logging::INFO))
-      : m_stepper(std::move(stepper)),
-        m_navigator(std::move(navigator)),
-        m_logger{std::move(_logger)} {}
-
- private:
   /// @brief Helper struct determining the state's type
   ///
   /// @tparam propagator_options_t Propagator options type
-  /// @tparam actor_list_t List of propagation action types
   ///
   /// This helper struct provides type definitions to extract the correct
   /// propagation state type from a given TrackParameter type and an
   /// ActorList.
   ///
-  template <typename propagator_options_t, typename actor_list_t>
+  template <typename propagator_options_t>
   struct state_type_helper {
+    using actor_list_t = typename propagator_options_t::actor_list_type;
+
     /// @brief Propagation state type for an arbitrary list of additional
     ///        propagation states
     ///
     /// @tparam args Parameter pack specifying additional propagation states
     ///
     template <typename... args>
-    using this_state_type = State<propagator_options_t, args...>;
+    using this_state_type =
+        PropagatorState<propagator_options_t, typename stepper_t::State,
+                        typename navigator_t::State, args...>;
 
     /// @brief Propagation result type derived from a given action list
     using type = typename actor_list_t::template result_type<this_state_type>;
@@ -172,15 +133,17 @@ class Propagator final
 
   /// @brief Helper struct determining the result's type
   ///
-  /// @tparam parameters_t Type of final track parameters
-  /// @tparam actor_list_t List of propagation action types
+  /// @tparam propagator_options_t Propagator options type
   ///
   /// This helper struct provides type definitions to extract the correct
   /// propagation result type from a given TrackParameter type and an
   /// ActorList.
   ///
-  template <typename parameters_t, typename actor_list_t>
+  template <typename propagator_options_t>
   struct result_type_helper {
+    using parameters_t = StepperBoundTrackParameters;
+    using actor_list_t = typename propagator_options_t::actor_list_type;
+
     /// @brief Propagation result type for an arbitrary list of additional
     ///        propagation results
     ///
@@ -194,24 +157,58 @@ class Propagator final
   };
 
  public:
-  /// @brief Short-hand type definition for propagation state derived from
-  ///        an action list
-  ///
-  /// @tparam actor_list_t List of propagation action types
-  ///
-  template <typename propagator_options_t, typename actor_list_t>
-  using actor_list_t_state_t =
-      typename state_type_helper<propagator_options_t, actor_list_t>::type;
+  /// Type of the stepper in use
+  using Stepper = stepper_t;
 
-  /// @brief Short-hand type definition for propagation result derived from
-  ///        an action list
+  /// Type of the navigator in use
+  using Navigator = navigator_t;
+
+  /// Type of the stepper options
+  using StepperOptions = typename Stepper::Options;
+
+  /// Type of the navigator options
+  using NavigatorOptions = typename Navigator::Options;
+
+  /// Type of the propagator options with actor list
+  template <typename actor_list_t = ActorList<>>
+  using Options =
+      PropagatorOptions<StepperOptions, NavigatorOptions, actor_list_t>;
+
+  /// Type of the stepper state
+  using StepperState = typename Stepper::State;
+
+  /// Type of the navigator state
+  using NavigatorState = typename Navigator::State;
+
+  /// Type of the propagation state derived from the propagator options
+  /// @tparam propagator_options_t Type of the propagator options
+  template <typename propagator_options_t>
+  using State = typename state_type_helper<propagator_options_t>::type;
+
+  /// Type of the propagation result derived from the propagator options
+  /// @tparam propagator_options_t Type of the propagator options
+  template <typename propagator_options_t>
+  using ResultType = typename result_type_helper<propagator_options_t>::type;
+
+  /// Constructor from implementation object
   ///
-  /// @tparam parameters_t Type of the final track parameters
-  /// @tparam actor_list_t List of propagation action types
-  ///
-  template <typename parameters_t, typename actor_list_t>
-  using actor_list_t_result_t =
-      typename result_type_helper<parameters_t, actor_list_t>::type;
+  /// @param stepper The stepper implementation is moved to a private member
+  /// @param navigator The navigator implementation, moved to a private member
+  /// @param _logger a logger instance
+  explicit Propagator(Stepper stepper, Navigator navigator = Navigator(),
+                      std::shared_ptr<const Logger> _logger =
+                          getDefaultLogger("Propagator", Acts::Logging::INFO))
+      : m_stepper(std::move(stepper)),
+        m_navigator(std::move(navigator)),
+        m_logger{std::move(_logger)} {}
+
+  /// Access to the stepper instance
+  /// @return Const reference to the stepper
+  const Stepper& stepper() const { return m_stepper; }
+
+  /// Access to the navigator instance
+  /// @return Const reference to the navigator
+  const Navigator& navigator() const { return m_navigator; }
 
   /// @brief Propagate track parameters
   ///
@@ -233,10 +230,9 @@ class Propagator final
   ///
   template <typename parameters_t, typename propagator_options_t,
             typename path_aborter_t = PathLimitReached>
-  Result<actor_list_t_result_t<StepperBoundTrackParameters,
-                               typename propagator_options_t::actor_list_type>>
-  propagate(const parameters_t& start, const propagator_options_t& options,
-            bool createFinalParameters = true) const;
+  Result<ResultType<propagator_options_t>> propagate(
+      const parameters_t& start, const propagator_options_t& options,
+      bool createFinalParameters = true) const;
 
   /// @brief Propagate track parameters - User method
   ///
@@ -259,10 +255,9 @@ class Propagator final
   template <typename parameters_t, typename propagator_options_t,
             typename target_aborter_t = SurfaceReached,
             typename path_aborter_t = PathLimitReached>
-  Result<actor_list_t_result_t<StepperBoundTrackParameters,
-                               typename propagator_options_t::actor_list_type>>
-  propagate(const parameters_t& start, const Surface& target,
-            const propagator_options_t& options) const;
+  Result<ResultType<propagator_options_t>> propagate(
+      const parameters_t& start, const Surface& target,
+      const propagator_options_t& options) const;
 
   /// @brief Builds the propagator state object
   ///
@@ -349,11 +344,9 @@ class Propagator final
   ///
   /// @return Propagation result
   template <typename propagator_state_t, typename propagator_options_t>
-  Result<actor_list_t_result_t<StepperBoundTrackParameters,
-                               typename propagator_options_t::actor_list_type>>
-  makeResult(propagator_state_t state, Result<void> result,
-             const propagator_options_t& options,
-             bool createFinalParameters) const;
+  Result<ResultType<propagator_options_t>> makeResult(
+      propagator_state_t state, Result<void> result,
+      const propagator_options_t& options, bool createFinalParameters) const;
 
   /// @brief Builds the propagator result object
   ///
@@ -372,32 +365,24 @@ class Propagator final
   ///
   /// @return Propagation result
   template <typename propagator_state_t, typename propagator_options_t>
-  Result<actor_list_t_result_t<StepperBoundTrackParameters,
-                               typename propagator_options_t::actor_list_type>>
-  makeResult(propagator_state_t state, Result<void> result,
-             const Surface& target, const propagator_options_t& options) const;
-
-  /// Access to the stepper instance
-  /// @return Const reference to the stepper
-  const stepper_t& stepper() const { return m_stepper; }
-
-  /// Access to the navigator instance
-  /// @return Const reference to the navigator
-  const navigator_t& navigator() const { return m_navigator; }
+  Result<ResultType<propagator_options_t>> makeResult(
+      propagator_state_t state, Result<void> result, const Surface& target,
+      const propagator_options_t& options) const;
 
  private:
-  const Logger& logger() const { return *m_logger; }
-
-  template <typename propagator_state_t, typename result_t>
-  void moveStateToResult(propagator_state_t& state, result_t& result) const;
-
   /// Implementation of propagation algorithm
-  stepper_t m_stepper;
+  Stepper m_stepper;
 
   /// Implementation of navigator
-  navigator_t m_navigator;
+  Navigator m_navigator;
 
   std::shared_ptr<const Logger> m_logger;
+
+  const Logger& logger() const { return *m_logger; }
+
+  template <typename propagator_state_t, typename propagator_result_t>
+  void moveStateToResult(propagator_state_t& state,
+                         propagator_result_t& result) const;
 };
 
 }  // namespace Acts
--- a/Core/include/Acts/Propagator/Propagator.ipp
+++ b/Core/include/Acts/Propagator/Propagator.ipp
@@ -11,7 +11,6 @@
 #include "Acts/Propagator/Propagator.hpp"
 
 #include "Acts/EventData/TrackParametersConcept.hpp"
-#include "Acts/Propagator/ActorList.hpp"
 #include "Acts/Propagator/ConstrainedStep.hpp"
 #include "Acts/Propagator/NavigationTarget.hpp"
 #include "Acts/Propagator/PropagatorError.hpp"
@@ -21,10 +20,11 @@
 
 #include <concepts>
 
+namespace Acts {
+
 template <typename S, typename N>
 template <typename propagator_state_t>
-Acts::Result<void> Acts::Propagator<S, N>::propagate(
-    propagator_state_t& state) const {
+Result<void> Propagator<S, N>::propagate(propagator_state_t& state) const {
   ACTS_VERBOSE("Entering propagation.");
 
   state.stage = PropagatorStage::prePropagation;
@@ -192,12 +192,10 @@ Acts::Result<void> Acts::Propagator<S, N>::propagate(
 template <typename S, typename N>
 template <typename parameters_t, typename propagator_options_t,
           typename path_aborter_t>
-auto Acts::Propagator<S, N>::propagate(const parameters_t& start,
-                                       const propagator_options_t& options,
-                                       bool createFinalParameters) const
-    -> Result<
-        actor_list_t_result_t<StepperBoundTrackParameters,
-                              typename propagator_options_t::actor_list_type>> {
+auto Propagator<S, N>::propagate(const parameters_t& start,
+                                 const propagator_options_t& options,
+                                 bool createFinalParameters) const
+    -> Result<ResultType<propagator_options_t>> {
   static_assert(std::copy_constructible<StepperBoundTrackParameters>,
                 "return track parameter type must be copy-constructible");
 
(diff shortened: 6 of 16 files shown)


Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -56,7 +56,7 @@ struct NavigationOptions {
   const object_t* endObject = nullptr;
 
   /// External surface identifier for which the boundary check is ignored
-  std::vector<GeometryIdentifier> externalSurfaces = {};
+  std::vector<const Surface*> externalSurfaces{};
```
andiwand: one thing we need to consider is if this should take ownership of the surfaces or if raw pointers are fine. if these surfaces are not part of the geometry, the user would have to take care of their lifetime right now

junggjo9:
Targets https://github.com/acts-project/acts/issues/4840

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -141,13 +140,13 @@ class Navigator {
     double farLimit = std::numeric_limits<double>::max();
 
     /// Externally provided surfaces - these are tried to be hit
-    ExternalSurfaces externalSurfaces = {};
+    boost::container::small_vector<const Surface*, 30> externalSurfaces{};
```
paulgessinger: I suppose `ExternalSurfaces` was also a small vector?

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -141,13 +140,13 @@ class Navigator {
...
 
     /// Insert an external surface to be considered during navigation
-    /// @param geoid Geometry identifier of the surface to insert
-    void insertExternalSurface(GeometryIdentifier geoid) {
-      externalSurfaces.insert(
-          std::pair<std::uint64_t, GeometryIdentifier>(geoid.layer(), geoid));
+    /// @param surface: Pointer to the surface to insert
+    void insertExternalSurface(const Surface* surface) {
```
paulgessinger: How about 

```suggestion
    void insertExternalSurface(const Surface& surface) {
```

? Pushes responsibility for having a valid surface to the caller.

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -796,6 +793,7 @@ class Navigator {
     NavigationArguments args;
     args.position = position;
     args.direction = direction;
+    args.desiredCandidates = state.options.externalSurfaces;
```
paulgessinger: *desired* vs *external*?

Core/include/Acts/TrackFitting/GlobalChiSquareFitter.hpp
```diff
@@ -1204,12 +1205,12 @@ class Gx2Fitter {
...
                               << " input measurements");
-    std::map<GeometryIdentifier, SourceLink> inputMeasurements;
+    std::unordered_map<const Surface*, SourceLink> inputMeasurements;
 
     for (; it != end; ++it) {
       SourceLink sl = *it;
-      auto geoId = gx2fOptions.extensions.surfaceAccessor(sl)->geometryId();
-      inputMeasurements.emplace(geoId, std::move(sl));
+      const Surface* surface = gx2fOptions.extensions.surfaceAccessor(sl);
+      inputMeasurements.emplace(surface, std::move(sl));
```
paulgessinger: Consider `try_exmplace` here.

Core/include/Acts/Propagator/Navigator.hpp
junggjo9: it was a map of layerId to a vector of GeometryIds.

Core/include/Acts/Propagator/Navigator.hpp
junggjo9:
```suggestion
    args.externalCandidates = state.options.externalSurfaces;
```

Core/include/Acts/Propagator/Navigator.hpp
andiwand: this might have quite some impact on navigations without external surfaces because the struct gros by `8*30` bytes. I think using a simple `std::vector` would be more practical

Core/include/Acts/Navigation/NavigationDelegate.hpp
```diff
@@ -12,20 +12,25 @@
...
   /// Current position in 3D space for navigation
-  Vector3 position{};
+  Vector3 position{Vector3::Zero()};
   /// Direction vector for navigation propagation
-  Vector3 direction{};
-
+  Vector3 direction{Vector3::Zero()};
+  /// List of candidate surfaces that should be appended to the navigation state
+  /// if possible
+  std::span<const Surface*, std::dynamic_extent> desiredCandidates{};
```
junggjo9:
```suggestion
  std::span<const Surface*, std::dynamic_extent> externalCandidates{};
```

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -173,25 +171,25 @@ class Navigator {
     /// the vector of navigation surfaces to work through
     NavigationSurfaces navSurfaces = {};
     /// the current surface index of the navigation state
-    std::optional<std::size_t> navSurfaceIndex;
+    std::optional<std::size_t> navSurfaceIndex{std::nullopt};
```
andiwand: personally I would remove all explicit initializers if they are non-trivial. but if initializing them makes it consistent I am fine with that too

Core/include/Acts/Navigation/NavigationDelegate.hpp
andiwand:
```suggestion
  std::span<const Surface*> desiredCandidates{};
```
isn't dynamic the default?

Core/include/Acts/Navigation/NavigationDelegate.hpp
```diff
@@ -12,20 +12,25 @@
...
 class AppendOnlyNavigationStream;
 class Logger;
+class Surface;
 
 /// Struct that serves as the argument to the navigation delegate.
 /// It is not supposed to be used as an lvalue.
 struct NavigationArguments {
   /// Current position in 3D space for navigation
-  Vector3 position{};
+  Vector3 position{Vector3::Zero()};
```
junggjo9:
```suggestion
  Vector3 position{};
```

Core/include/Acts/Navigation/NavigationDelegate.hpp
```diff
@@ -12,20 +12,25 @@
...
 #include "Acts/Utilities/Delegate.hpp"
 
+#include <span>
+
 namespace Acts {
 
 class NavigationStream;
 class AppendOnlyNavigationStream;
 class Logger;
+class Surface;
```
junggjo9:
```suggestion
```

Core/include/Acts/Navigation/NavigationDelegate.hpp
```diff
@@ -12,20 +12,25 @@
 #include "Acts/Surfaces/BoundaryTolerance.hpp"
 #include "Acts/Utilities/Delegate.hpp"
 
+#include <span>
+
```
junggjo9:
```suggestion
```

Core/include/Acts/Navigation/NavigationDelegate.hpp
```diff
@@ -12,20 +12,25 @@
...
 /// Struct that serves as the argument to the navigation delegate.
 /// It is not supposed to be used as an lvalue.
 struct NavigationArguments {
   /// Current position in 3D space for navigation
-  Vector3 position{};
+  Vector3 position{Vector3::Zero()};
   /// Direction vector for navigation propagation
-  Vector3 direction{};
-
+  Vector3 direction{Vector3::Zero()};
```
junggjo9:
```suggestion
  Vector3 direction{};
```

Core/include/Acts/Navigation/NavigationDelegate.hpp
```diff
@@ -12,20 +12,25 @@
...
+  /// List of candidate surfaces that should be appended to the navigation state
+  /// if possible
+  std::span<const Surface*, std::dynamic_extent> externalCandidates{};
```
junggjo9:
```suggestion
```

Core/include/Acts/Propagator/Navigator.hpp
junggjo9: I usually initialize them all inline, no matter whether they're trivial objects or not. I can remove the change if you want

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -796,6 +792,7 @@ class Navigator {
     NavigationArguments args;
     args.position = position;
     args.direction = direction;
+    args.externalCandidates = state.options.externalSurfaces;
```
junggjo9:
```suggestion
```

Core/include/Acts/Navigation/NavigationDelegate.hpp
junggjo9: true

Core/include/Acts/Propagator/Navigator.hpp
junggjo9:
```suggestion
    std::vector<const Surface*> externalSurfaces{};
```

Core/include/Acts/TrackFitting/GlobalChiSquareFitter.hpp
junggjo9: Done

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -796,9 +804,24 @@ class Navigator {
     NavigationArguments args;
     args.position = position;
     args.direction = direction;
+    args.externalSurfaces = state.options.externalSurfaces;
     state.currentVolume->initializeNavigationCandidates(args, appendOnly,
                                                         logger());
-
+    if (!state.options.freeSurfaces.empty()) {
```
junggjo9: This inserts the free surfaces to the stream. However, there's no limit on how far the insertion can go. For example, the user put a free surface at the beginning and at the end of the proto-track, then the navigator would pick up the first, the measurements of the first layer and then the last one

Core/include/Acts/Navigation/NavigationDelegate.hpp
```diff
@@ -12,20 +12,23 @@
...
 struct NavigationArguments {
   /// Current position in 3D space for navigation
-  Vector3 position{};
+  Vector3 position{Vector3::Zero()};
   /// Direction vector for navigation propagation
-  Vector3 direction{};
-
+  Vector3 direction{Vector3::Zero()};
+  /// List of surfaces that are requested by the user. The policy should
+  /// append the ones from the list which are managed by it
```
paulgessinger: Does this require support from the policy now? If not, do we need this in the `NavigationArguments` struct at all at this point?

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -173,27 +184,37 @@ class Navigator {
...
     // Navigation candidates(portals and surfaces together)
     /// the vector of navigation candidates to work through
     NavigationCandidates navCandidates = {};
     /// the current candidate index of the navigation state
-    std::optional<std::size_t> navCandidateIndex;
-
+    std::optional<std::size_t> navCandidateIndex{std::nullopt};
+
+    // Navigation on free surface level. Free surfaces are traversed
+    NavigationCandidates freeCandidates = {};
```
paulgessinger: So we need a separate candidate set? Is there a way to pre-populate the candidates object that we're about to hand over to the policy? That way, they would end up in the same candidate list, and you would have to do less book keeping later on, no?

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -730,21 +805,51 @@ class Navigator {
   /// @param direction The current direction
   NavigationTarget getNextTargetGen3(State& state, const Vector3& position,
                                      const Vector3& direction) const {
+    if (!state.freeCandidates.empty() &&
+        !state.freeCandidateIndex.has_value()) {
```
paulgessinger: I don't think we should have separate candidate sequences, to be honest

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -663,21 +710,49 @@ class Navigator {
...
+                                    << state.navSurfaces.size());
       }
-      if (state.navSurfaceIndex.value() < state.navSurfaces.size()) {
+      constexpr auto maxIdx = std::numeric_limits<std::size_t>::max();
+      const auto navIdx = state.navSurfaceIndex.value_or(maxIdx);
+      const auto freeIdx = state.freeCandidateIndex.value_or(maxIdx);
+      /// Check whether free surfaces and geometry surfaces are available at the
+      /// same time If the free surface is closer than the geometry surface
+      /// return the free
+      state.navSurfaceIsFree =
```
paulgessinger: This is used to then select the correct stream to take from. I think we should avoid this.

Core/include/Acts/Propagator/Navigator.hpp
junggjo9: In the latest implementation, the freeCandidates are merged into the corresponding surface or navigation stream

Core/include/Acts/Propagator/Navigator.hpp
junggjo9: That case is removed after merging

Core/include/Acts/Propagator/Navigator.hpp
junggjo9: Removed

Core/include/Acts/Navigation/NavigationDelegate.hpp
junggjo9: The policy knows best about the candidates

paulgessinger: Right, but the policy should ideally not have to be aware of the extra surface being passed in.

Core/include/Acts/Propagator/NavigationTarget.hpp
```diff
@@ -266,6 +274,10 @@ class NavigationTarget {
 
   /// Default constructor creating a none target
   constexpr NavigationTarget() = default;
+
+  /// @brief print method
+  /// @param ostr: Stream to which the object is printed
+  void print(std::ostream& ostr) const;
```
paulgessinger, line 280: Is the outstream operator not sufficient?

Core/include/Acts/Propagator/Navigator.hpp
paulgessinger: Ok, so this is now only used to hold the candidates after they've been copied over from the options struct?

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -812,16 +857,32 @@ class Navigator {
...
+                  std::max(freeFarLimit, candidate.intersection().pathLength());
+            }
+          });
+      std::ranges::copy_if(
+          state.freeCandidates, std::back_inserter(state.navCandidates),
+          [&](const NavigationTarget& candidate) {
+            return detail::checkPathLength(
+                candidate.intersection().pathLength(), state.options.nearLimit,
+                freeFarLimit, logger());
+          });
```
paulgessinger: So here you're essentially clipping the free candidates to the maximum path length to any portal, so the furthest away portal. like you suggested the other day. This happens every time I suppose? We never reduce the number of free surfaces by removing the ones that are no longer reachable?

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -841,7 +902,24 @@ class Navigator {
       logger().log(Logging::VERBOSE, os.str());
     }
   }
-
+  /// @brief  Update the intersections with the free surfaces of the navigation state.
+  ///         Selects the first free surface which is ahead of the track
+  /// @param state: The current navigation state
+  /// @param position: The current track position
+  /// @param direction: The current track direction
+  void updateFreeInterSections(State& state, const Vector3& position,
```
paulgessinger: Nitpick:

```suggestion
  void updateFreeIntersections(State& state, const Vector3& position,
```

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -870,22 +948,34 @@ class Navigator {
     navOpts.endObject = state.targetSurface;
     navOpts.nearLimit = state.options.nearLimit;
     navOpts.farLimit = state.options.farLimit;
-
+    /// setup the surface candidates requested by the user
```
paulgessinger:
```suggestion
    // setup the surface candidates requested by the user
```

Core/src/Navigation/MultiLayerNavigationPolicy.cpp
```diff
@@ -61,9 +63,25 @@ void MultiLayerNavigationPolicy::initializeCandidates(
...
+  }
+  std::ranges::for_each(surfaces, [&](const Surface& surf) {
+    if (addedIds.count(surf.geometryId()) == 0u &&
+        rangeContainsValue(args.externalSurfaces, surf.geometryId())) {
+      ACTS_VERBOSE(
+          "MultiLayerNavigationPolicy add externally requested candidate "
+          << surf.geometryId());
+      stream.addSurfaceCandidate(surf, args.tolerance);
+    }
+  });
```
paulgessinger: This bit I don't understand: I believe the navigation policy should not be involved in adding the external surfaces. Aren't these candidates added separately in any case?

Core/src/Propagator/NavigationTarget.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include "Acts/Geometry/BoundarySurfaceT.hpp"
+#include "Acts/Geometry/Layer.hpp"
+#include "Acts/Geometry/Portal.hpp"
+#include "Acts/Surfaces/Surface.hpp"
+
+namespace Acts {
+void NavigationTarget::print(std::ostream& ostr) const {
+  std::visit(
+      [&](const auto target) {
```
paulgessinger:
```suggestion
      [&](const auto& target) {
```

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -173,26 +184,29 @@ class Navigator {
...
     /// the vector of navigation candidates to work through
     NavigationCandidates navCandidates = {};
     /// the current candidate index of the navigation state
-    std::optional<std::size_t> navCandidateIndex;
-
+    std::optional<std::size_t> navCandidateIndex{std::nullopt};
+    // Navigation on free surface level. Free surfaces are traversed
+    NavigationCandidates freeCandidates = {};
+    /// Switch toggling whether the free or the navigation surface shall
+    /// be returned by navSurface
```
paulgessinger:
```suggestion
```

Core/include/Acts/Propagator/Navigator.hpp
junggjo9: indeed

Core/include/Acts/Propagator/NavigationTarget.hpp
junggjo9: Some stuff was forward declared, so I am using this mechanism to just push that towards the .cpp

Core/src/Navigation/MultiLayerNavigationPolicy.cpp
junggjo9: That's a very good point and my honest answer is dunno

Core/include/Acts/Propagator/Navigator.hpp
junggjo9: Implemented

Core/src/Propagator/NavigationTarget.cpp
```diff
@@ -0,0 +1,36 @@
...
+        } else if constexpr (std::is_same_v<Target_t, const BoundarySurface*>) {
+          const auto& surf{target->surfaceRepresentation()};
+          ostr << "Boundary " << surf.type() << " surface "
+               << surf.geometryId();
+        } else if constexpr (std::is_same_v<Target_t, const Portal*>) {
+          const auto& surf{target->surface()};
+          ostr << surf.type() << " portal " << surf.geometryId();
+        }
+      },
+      m_target);
```
andiwand, line 34: this could use the `overloaded` helper like here https://github.com/acts-project/acts/blob/5dc047cfd9d4ed6bd188c92e65863f27df9f0c77/Core/include/Acts/Geometry/TrackingVolume.hpp#L183-L188

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -254,19 +266,27 @@ class Navigator {
...
+      /// Ensure that the free surfaces become available again
+      std::ranges::for_each(freeCandidates, [&](NavigationTarget& target) {
+        if (target.pathLength() > options.nearLimit) {
+          return;
+        }
+        Intersection<3> newIsect{
+            target.position(), options.nearLimit + s_epsilon, target.status()};
+        target = NavigationTarget{newIsect, target.intersectionIndex(),
+                                  target.surface(), target.boundaryTolerance()};
+      });
```
andiwand: This looks quite sketchy to me. Why is this necessary?

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -387,6 +406,29 @@ class Navigator {
...
+      std::ranges::transform(
+          state.options.freeSurfaces, std::back_inserter(state.freeCandidates),
+          [&](const Surface* freeSurf) {
+            auto [intersection, intersectionIndex] =
+                freeSurf
+                    ->intersect(state.options.geoContext, position, direction,
+                                BoundaryTolerance::Infinite())
+                    .closestWithIndex();
+            return NavigationTarget{intersection, intersectionIndex, *freeSurf,
+                                    BoundaryTolerance::Infinite()};
```
andiwand: intersecting everything from the starting position seems unstable to me. I think the order should be rather supplied by the user and we stick to that. Similar to what is done in the `DirectNavigator`

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -665,19 +707,23 @@ class Navigator {
...
+                     << "Target set to next surface: " << state.navSurface());
         return state.navSurface();
-      } else {
-        // This was the last surface, switch to layers
-        ACTS_VERBOSE(volInfo(state) << "Target layers.");
-        state.navigationStage = Stage::layerTarget;
       }
+      // This was the last surface, switch to layers
+      ACTS_VERBOSE(volInfo(state) << "Target layers.");
+      state.navigationStage = Stage::layerTarget;
```
andiwand: won't this now always switch to layer navigation?

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -796,9 +849,26 @@ class Navigator {
...
     args.direction = direction;
+
     state.currentVolume->initializeNavigationCandidates(
         state.options.geoContext, args, appendOnly, logger());
-
+    /// Add the external surfaces from the volume as extra candidates
+    if (!state.options.externalSurfaces.empty()) {
+      const auto volId = state.currentVolume->geometryId().volume();
+
+      std::ranges::for_each(state.options.externalSurfaces,
```
andiwand: using `std::ranges::for_each` in these cases seems not beneficial to me - why not just use a ranged based for loop?

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -141,13 +140,25 @@ class Navigator {
...
+      externalSurfaces.push_back(geoId);
+    }
     /// Insert an external surface to be considered during navigation
-    /// @param geoid Geometry identifier of the surface to insert
-    void insertExternalSurface(GeometryIdentifier geoid) {
-      externalSurfaces.insert(
-          std::pair<std::uint64_t, GeometryIdentifier>(geoid.layer(), geoid));
+    /// @param surface: Reference to the surface to insert
+    void insertExternalSurface(const Surface& surface) {
+      /// All geometry surfaces
```
andiwand: I think we use `///` consistently for doxygen and comments should use `//`

occurs a couple of times below

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -812,16 +882,32 @@ class Navigator {
...
+                  std::max(freeFarLimit, candidate.intersection().pathLength());
+            }
+          });
+      std::ranges::copy_if(
+          state.freeCandidates, std::back_inserter(state.navCandidates),
+          [&](const NavigationTarget& candidate) {
+            return detail::checkPathLength(
+                candidate.intersection().pathLength(), state.options.nearLimit,
+                freeFarLimit, logger());
+          });
```
andiwand: I don't think I understand the logic here - can't we just put all of the free surfaces into the candidate vector?

Core/include/Acts/Propagator/Navigator.hpp
junggjo9: The free targets are continously updated as the propagator travels through the volumes. So if you reset the state back to the beginning you alo want to make the free candidates available again. The ones which have not yet been passed still have a pathLength() > 0. The ones which have been passed will fail the nearPathLimit and you push them slightly into the allowed ranges such that the updateFreeInterSections will pick them up again

Core/include/Acts/Propagator/Navigator.hpp
junggjo9: yeah, but you need an initial intersection to assess which free surfaces are to be considereed at the given navigation step.

Core/include/Acts/Propagator/Navigator.hpp
junggjo9:
```suggestion
      // All geometry surfaces
```

Core/include/Acts/Propagator/Navigator.hpp
junggjo9: The else block is not relly needed. if the surface index is within the navSurface range, then the surface is returned other wise the state is set to layer. So if the if statement already returns something from the function, why is an else after neccessary and it can be dropped to drop one layer of nestiness

Core/include/Acts/Propagator/Navigator.hpp
junggjo9: To be honest, I've never really undertood the advantage of std:;ranges::for_each over a conventional for loop. I just wanted to use something more seemlingy fancy

Core/include/Acts/Propagator/Navigator.hpp
junggjo9: As far as I understand, I would say no. If you've a geometry with 5 layers and you put a freesurface at the very beginning and another one at the very end, the existence of the last free surface would make the navigator to skip the layers in between, woudln't  it? The stepper is on the last surface from the layer and the next surface passed is then the free surface

Core/src/Propagator/NavigationTarget.cpp
junggjo9: can check it tomorrow

Core/include/Acts/Propagator/Navigator.hpp
andiwand: That is true. But I think that should either be solved by constraining surfaces to layers / volumes and get this information from the user or by having 2 surface streams that are checked at the same time for the next surface. The latter is also more robust against overlaps

junggjo9: that was in place but @paulgessinger had objections ;)

paulgessinger: I think having separate streams complicates things. We have a single stream for the portals, sensitives and passives, and in my opinion this simplified the code a lot compared with what we had before. I was also suggesting to constrain the free surfaces by volume association, either by doing a volume lookup of the `center` position, or by having the caller provide this information explicitly. I think that would work here? @junggjo9 @andiwand

Core/include/Acts/Propagator/Navigator.hpp
paulgessinger: After our discussion yesterday @junggjo9, is this actually still the case? If you always add the candidates (filtered by volume, see other comment) into the main candidate stream, the intersections should happen on volume entry, not on reset.

Core/include/Acts/Propagator/Navigator.hpp
paulgessinger: @junggjo9 and this is because you still want to clip by path length?

Core/include/Acts/Propagator/Navigator.hpp
paulgessinger: I think the logic flow is equivalent to before, because the `if` clause has a `return` in it.

Core/include/Acts/Propagator/Navigator.hpp
paulgessinger: It's beneficial if you have a function that you want to apply already. The pattern of `for_each` with an inline lambda seems more verbose than a for loop to me as well.

Core/include/Acts/Propagator/Navigator.hpp
andiwand: it depends what problem you want to solve.

if you want to assign free surfaces to a volume because they are strictly confined, we should add them to the surface stream. but then this should be done explicitly. the user should provide the surface and the volume. hiding the volume looking inside the navigator is a poor design imo because it does not communicate what is happening to the user.

the other kind of free surface is something like the perigee which overlaps with the whole tracking geometry. this can only be handled in a stable way if we have multiple surface streams because overlaps can result in re-ordering of surface candidates

Core/include/Acts/Propagator/Navigator.hpp
andiwand: oh yeah right - I overlooked this 👍

Core/include/Acts/Surfaces/SensitiveSurface.hpp
andiwand, line 1: This is a new pattern and in conflict with https://github.com/acts-project/acts/pull/4985

any opinions @paulgessinger @benjaminhuth @asalzburger ?

I also asked @junggjo9 if he could present this tomorrow during the ACTS dev weekly

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -138,14 +138,31 @@ class Navigator {
...
+    ///        for each surface that has not been marked as reached yet.
+    /// @param gctx: Current geometry context carrying the alignment information
+    /// @param currentVol: The current tracking volume in which the propagator resides
+    /// @param pos: Position of the track in global coordinates
+    /// @param dir: Direction vector of the track
+    /// @param surface: Free surface candidate to test
+    using FreeSurfaceSelctor_t = Delegate<bool(
+        const GeometryContext& gctx, const TrackingVolume& currentVol,
+        const Vector3& pos, const Vector3& dir, const Surface& candidate)>;
+    FreeSurfaceSelctor_t freeSurfaceSelector{};
```
andiwand, line 165:
```suggestion
    FreeSurfaceSelctor_t freeSurfaceSelector{};

```

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -138,14 +138,31 @@ class Navigator {
...
+    /// @param surface: The surface to add to the list
+    void insertExternalSurface(const Surface& surface) {
+      if (surface.geometryId() != GeometryIdentifier{}) {
+        externalSurfaces.push_back(surface.geometryId());
+      } else {
+        freeSurfaces.push_back(&surface);
+      }
     }
-
+    /// @brief Delegate to decide whether free surfaces are appended to the navigation
```
andiwand, line 153:
```suggestion

    /// @brief Delegate to decide whether free surfaces are appended to the navigation
```

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -263,6 +286,11 @@ class Navigator {
 
       navigationBreak = false;
       navigationStage = Stage::initial;
+      // Set the surface reached switches back to false
+      std::ranges::for_each(freeCandidates,
+                            [](std::pair<const Surface*, bool>& freeSurface) {
+                              freeSurface.second = false;
+                            });
```
andiwand, line 293:
```suggestion
      for (auto & [surface, reached] : freeCandidates) {
        reached = false;
      }
```

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -263,6 +286,11 @@ class Navigator {
 
       navigationBreak = false;
       navigationStage = Stage::initial;
+      // Set the surface reached switches back to false
```
andiwand, line 289:
```suggestion

      // Set the surface reached switches back to false
```

Core/include/Acts/Surfaces/SensitiveSurface.hpp
```diff
@@ -0,0 +1,56 @@
...
+  }
+  /// @brief implement the assignment operators
+  /// Assignment operator
+  ///
+  /// @param other is the source surface for copying
+  /// @return Reference to this surface for assignment chaining
+  SensitiveSurface& operator=(const Surface_t& other) {
+    static_cast<Surface_t&>(*this) = other;
+    return *this;
+  }
```
andiwand: are these necessary to fulfill the `SurfaceConcept`?

Core/src/Navigation/Navigator.cpp
```diff
@@ -16,8 +16,8 @@
 #include "Acts/Utilities/StringHelpers.hpp"
 
 #include <algorithm>
+#include <cassert>
 #include <sstream>
-
 namespace Acts {
```
andiwand, line 21:
```suggestion

namespace Acts {
```
I really need these new lines 😿

Core/src/Navigation/Navigator.cpp
```diff
@@ -457,7 +478,48 @@ void Navigator::resolveCandidates(State& state, const Vector3& position,
 
   ACTS_VERBOSE(volInfo(state) << "Found " << state.stream.candidates().size()
                               << " navigation candidates.");
-
+  if (!state.options.externalSurfaces.empty()) {
+    std::ranges::for_each(
```
andiwand: do you see an advantage over a regular range-based for loop? if not I would rather stick to them for consistency and I believe they are easier to digest

same below

Core/src/Navigation/Navigator.cpp
```diff
@@ -520,15 +594,13 @@ void Navigator::resolveSurfaces(State& state, const Vector3& position,
...
-         itSurface != externalSurfaceRange.second; itSurface++) {
-      navOpts.externalSurfaces.push_back(itSurface->second);
-    }
+    const auto layerId = layerSurface->geometryId().layer();
+    navOpts.externalSurfaces.reserve(state.options.externalSurfaces.size());
+    std::ranges::copy_if(state.options.externalSurfaces,
+                         std::back_inserter(navOpts.externalSurfaces),
+                         [&layerId](const GeometryIdentifier& id) {
+                           return id.layer() == layerId;
+                         });
```
andiwand: - `state.options.externalSurfaces.size()` might over-allocate quite a bit
- `copy_if` might be more involved than a simple range-based for loop

Core/include/Acts/Surfaces/SensitiveSurface.hpp
junggjo9: The template parameters are already restricted in a two-fold way. The template argument itself is required to fullfil the surface concept and the constructor only works if the class somehow inherits from the Surface base class which by itself already satisfies he concept. Anyhow the assignment is delegated to the direct parent

Core/include/Acts/Surfaces/SensitiveSurface.hpp
benjaminhuth: Indeed I would also fear that this could cause more confusion then it has benefits. Users might think they need to use `SenistiveSurface` if they want a sensitive surface, but in fact they should not (they should use detector elements in almost all cases)

I wonder if this is very workflow specific @junggjo9 ? Maybe this class would better live in the Athena code where it is used?

paulgessinger: I think I concur, that this makes the most sense. In the ATLAS software I think the detector element calls always receive the identifier of the surface in the method calls, which allows them to have one detector element potentially represent multipl surfaces, and I believe this is used to model e.g. straws, possibly also used in the muon system, but not sure.

Core/src/Navigation/Navigator.cpp
junggjo9: So no allocation beforehand?

Core/src/Navigation/Navigator.cpp
junggjo9: In principle, I've no preference over the other. Will change to for loop

Core/include/Acts/Surfaces/SensitiveSurface.hpp
junggjo9: File removed again...resolving

Core/include/Acts/Surfaces/SensitiveSurface.hpp
junggjo9: In the ATLAS software, DetectorElements are produced for each surface but managed but the associated readout element. Anyhow, the SensitiveSurface is off the table and removed. Resolving

acts-project-service:
## :red_circle: Athena integration test results [6a4a247b4038fd2bee2ea9d5907bc82825958ae8]
Please investigate the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13839776)!

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183167)||
|:red_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183166)||
|:red_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183165)||
|:red_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183164)||
|:red_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183163)||
|:yellow_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183162)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183161)||
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183160)||
|:red_circle: | [test_ActsCheckObjectCountsWorkflowHgtd](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183159)||
|:red_circle: | [test_ActsCheckObjectCountsCachedWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183158)||
|:red_circle: | [test_ActsCheckObjectCountsLegacyWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183157)||
|:red_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183156)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183155)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183154)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183153)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183152)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183151)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183150)||
|:red_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183149)||
|:red_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183148)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183147)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183146)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183145)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183144)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183143)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183142)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183141)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183140)||
|:green_circle: | [test_ActsGx2fRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183139)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183138)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183137)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183136)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68183135)||

PR #4496, opened 2025-07-31
fix: enlarge cylinder volume rmax for layers with displaced center

wdconinc:
This PR ensures that cylinder volumes fully enclose cylinder and radial bound layers with displaced centers. Without this fix, the `rmax` for the volume is (without envelope) equal to the `rmax` of the layer, even if the volume is not placed on axis. This leads to the layer sticking out of the cylinder.

It would be possible to modify the `rmin` as well, but it adds branching (three options for 'donut' radial/cylinder bounds: z axis through donut hole, z axis through donut itself, z axis outside donut) and it does not resolve any current buggy behavior (it just uses an `rmin` that could be smaller than it currently is). Therefore that is not included here.

Before (cross hairs are xy, looking along z, off-axis detector with its layer shown in blue, cylinder volume in dark yellow):
image

After (cross hairs are xy, looking along z, off-axis detector with its layer shown in blue, cylinder volume in dark yellow):
image

--- END COMMIT MESSAGE ---

### DIFF
--- a/Core/src/Geometry/CylinderVolumeBuilder.cpp
+++ b/Core/src/Geometry/CylinderVolumeBuilder.cpp
@@ -574,8 +574,8 @@ VolumeConfig CylinderVolumeBuilder::analyzeContent(
             CylinderBounds::eHalfLengthZ);
         lConfig.rMin =
             std::min(lConfig.rMin, rMinC - m_cfg.layerEnvelopeR.first);
-        lConfig.rMax =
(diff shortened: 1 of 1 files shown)


andiwand (approved):

acts-project-service:
## :red_circle: Athena integration test results [f812fb17537712f5d31df803ef6c052fea9a90b8]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/59597876) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12567119)!

PR #4499, opened 2025-07-31
refactor: Extend use of `copyFromIndex` in `Seeding2`

andiwand:
- use `SpacePointIndex2` over `std::size_t` for `copyFromIndexColumn`
- add `resolvedIndex` helper to ease discovery of proper index
- propagate `copyFromIndex`, if available, to seed

--- END COMMIT MESSAGE ---

### DIFF
--- a/Core/include/Acts/Geometry/SurfaceArrayCreator.hpp
+++ b/Core/include/Acts/Geometry/SurfaceArrayCreator.hpp
@@ -363,15 +363,17 @@ class SurfaceArrayCreator {
   /// @tparam bdtB AxisBoundaryType of axis B
   /// @tparam F1 type-deducted value of g2l lambda
   /// @tparam F2 type-deducted value of l2g lambda
-  /// @param globalToLocal transform callable
-  /// @param localToGlobal transform callable
+  /// @param type The surface type, this determines the local to global calculation
+  /// @param transform The transform to apply to the surface`
+  /// @param R the radius (interpretation depends on @p type)
+  /// @param Z the z position (interpretation depends on @p type)
   /// @param pAxisA ProtoAxis object for axis A
   /// @param pAxisB ProtoAxis object for axis B
-  template <AxisBoundaryType bdtA, AxisBoundaryType bdtB, typename F1,
-            typename F2>
+  template <AxisBoundaryType bdtA, AxisBoundaryType bdtB>
   static std::unique_ptr<SurfaceArray::ISurfaceGridLookup>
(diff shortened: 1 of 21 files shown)


Core/src/Seeding2/BroadTripletSeedFilter.cpp
```diff
@@ -287,34 +277,37 @@ void BroadTripletSeedFilter::filter1SpFixed(
       break;
     }
 
+    std::array<SpacePointIndex2, 3> triplet{spacePoints[bottom].resolvedIndex(),
```
benjaminhuth, line 280: Is this too hot to use the `.at()` functions?

Core/src/Seeding2/BroadTripletSeedFilter.cpp
```diff
@@ -287,34 +277,37 @@ void BroadTripletSeedFilter::filter1SpFixed(
...
-    setBestSeedQuality(spacePoints, state.bestSeedQualityMap, top, middle,
-                       bottom, bestSeedQuality);
+    setBestSeedQuality(state.bestSeedQualityMap, triplet[0], triplet[1],
+                       triplet[2], bestSeedQuality);
 
-    ACTS_VERBOSE("Adding seed: [b=" << bottom << ", m=" << middle << ", t="
-                                    << top << "], quality=" << bestSeedQuality
-                                    << ", vertexZ=" << zOrigin);
+    ACTS_VERBOSE("Adding seed: original indices=["
+                 << triplet[0] << ", " << triplet[1] << ", " << triplet[2]
```
benjaminhuth, line 304: I think in the log statement one could use `.at()`...

Core/src/Seeding2/BroadTripletSeedFilter.cpp
andiwand: for now I consistently used `[]` in the seeding code. I saw a tiny impact of using `at()`. I will reconsider this at a later stage

Core/src/Seeding2/BroadTripletSeedFilter.cpp
andiwand: right here it really does not matter. the main reason I use array accessors here is that `triplet` is defined above as `std::array<SpacePointIndex2, 3>` which should guarantee this to work in all cases

acts-project-service:
## :red_circle: Athena integration test results [7c1489247f463516d906b2d59399d2d8fe52f93f]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/59848711) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12604969)!

PR #4502, opened 2025-07-31
fix: propagate transform to ProtoLayer in DD4hep builder; use in SurfaceArrayCreator

wdconinc:
This PR ensures that off-axis DD4hep endcap sensitives are placed in a layer that takes their transform into account, rather than radial bounds that are determined with respect to the global z axis. The corresponding change in the surface array creator is that the transverse position must be similarly aware of the transform to end up with correct binning.

The change to `Core/src/Geometry/SurfaceArrayCreator.cpp` may affect other non-DD4hep geometry, but should only affect off-axis (or rotated) geometry. I am afraid these fall outside of the unit tests.

Before: (cross hairs are xy, looking along z, off-axis detector with its layer shown in blue, cylinder volume in dark yellow)
image

After: (cross hairs are xy, looking along z, off-axis detector with its layer shown in blue, cylinder volume in dark yellow)
image

--- END COMMIT MESSAGE ---


acts-project-service:
## :red_circle: Athena integration test results [05eb9c9ca448e56cef79ae8678e50e8b0895cb1f]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/59750445) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12590944)!

PR #4509, opened 2025-08-04
feat: Generalized `TripletSeedFinder`

andiwand:
A generalization after the implementation of the `BroadTripletSeedFinder` which should be viable for different applications.

--- END COMMIT MESSAGE ---

For now this depends on the output of the `DoubletSeedFinder` and is another building block for a full seeding algorithm. No filtering is involved for now so it has to be done by the caller.

If the additional (virtual) function calls have a noticeable performance penalty is still to be seen. There are a couple of ways to mitigate this if this is the case. One solution would be to pull in the loop over the bottom doublets. This then requires to call the filter from within  `TripletSeedFinder`. If this is still too slow we can expose the templated doublet and triplet seed finder functions to the experiment and build the seeding algorithm on top of this on the experiment side.


andiwand:
I measure similar performance on arm. I will check x86 as well

image

edit: threadripper

image

Core/include/Acts/Seeding2/ITripletSeedCuts.hpp
```diff
@@ -43,9 +43,8 @@ class ITripletSeedCuts {
 
   /// @param seedCandidates contains collection of seed candidates created for one middle
   /// space point in a std::tuple format
-  /// @return vector of seed candidates that pass the cut
   virtual void cutPerMiddleSp(
-      std::span<TripletCandidate2> seedCandidates) const = 0;
+      std::span<TripletCandidate2>& seedCandidates) const = 0;
```
paulgessinger, line 47: Does this make a difference? I imagine having 1 pointer instead of 1 pointer and one size is slightly, but is that all? Aside from that, how does this method work anyway? It does not return, does it modify the span? Is there an example of an implementation anywhere yet?

Core/include/Acts/Seeding2/TripletSeedFinder.hpp
```diff
@@ -0,0 +1,229 @@
...
+
+    /// Whether the input doublets are sorted by cotTheta
+    bool sortedByCotTheta = true;
+  };
+
+  struct DerivedConfig : public Config {
+    DerivedConfig(const Config& config, float bFieldInZ);
+
+    float bFieldInZ = std::numeric_limits<float>::quiet_NaN();
+    float highland = std::numeric_limits<float>::quiet_NaN();
```
paulgessinger, line 130: Default changes from 0 to `nan`, but I imagine these are set from the constructor anyway?

Core/include/Acts/Seeding2/TripletSeedFinder.hpp
```diff
@@ -0,0 +1,229 @@
...
+
+   protected:
+    DerivedConfig m_cfg;
+  };
+  template <bool useStripInfo, bool sortedInCotTheta>
+  class Impl;
+
+  static std::shared_ptr<ImplBase> makeImpl(const DerivedConfig& config);
+
+  std::shared_ptr<ImplBase> m_impl;
```
paulgessinger: `shared_ptr` over `unique_ptr`?

Core/include/Acts/Seeding2/TripletSeedFinder.hpp
```diff
@@ -0,0 +1,229 @@
...
+      const SpacePointContainer2& spacePoints, const ConstSpacePointProxy2& spM,
+      const DoubletsForMiddleSp::Proxy& bottomDoublet,
+      DoubletsForMiddleSp::Subset2& topDoublets,
+      TripletTopCandidates& tripletTopCandidates) const {
+    m_impl->createTripletTopCandidates(spacePoints, spM, bottomDoublet,
+                                       topDoublets, tripletTopCandidates);
+  }
+
+ private:
+  class ImplBase {
```
paulgessinger: Not important, but you could forward declare this if you're ok with moving the function calls into this class into the `.cpp` file (they're currently defined here in the header).

Core/include/Acts/Seeding2/TripletSeedFinder.hpp
```diff
@@ -0,0 +1,229 @@
...
+        const ConstSpacePointProxy2& spM,
+        const DoubletsForMiddleSp::Proxy& bottomDoublet,
+        DoubletsForMiddleSp::Subset2& topDoublets,
+        TripletTopCandidates& tripletTopCandidates) const = 0;
+
+   protected:
+    DerivedConfig m_cfg;
+  };
+  template <bool useStripInfo, bool sortedInCotTheta>
+  class Impl;
```
paulgessinger: Do you need to declare this here at all?

Core/include/Acts/Seeding2/TripletSeedFinder.hpp
```diff
@@ -0,0 +1,229 @@
...
+
+  static std::shared_ptr<ImplBase> makeImpl(const DerivedConfig& config);
```
paulgessinger: Also probablly not necessary to declare here, as it's static and doesn't access members.

paulgessinger (commented):
looks solid to me overall

Core/include/Acts/Seeding2/TripletSeedFinder.hpp
andiwand: right these are set on construction. I just thought `0` makes little sense if it ever ends up as a value. I would rather let everything fail in this case

Core/include/Acts/Seeding2/ITripletSeedCuts.hpp
andiwand: This was an oversight on my end. I think the idea is that this function can remove candidates. I modeled this here as a mutable span, so the user could shuffle the candidates and return a sub-span. I don't think we actually use this anywhere right now. Copied this over from the previous implementation

paulgessinger (approved):

acts-project-service:
## :red_circle: Athena integration test results [015aff0ab702786db8c45f069add97879f179922]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/60179057) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12657300)!

PR #4515, opened 2025-08-05
perf!: Speed-up hit access in HoughTransFormUtils

goblirsc:
Speed-up hit access in HoughTransFormUtils
--- END COMMIT MESSAGE ---

This MR removes a bottleneck in the `HoughTransformUtils` where access to the hits in a given cell would always perform a sorting operation. Speeds up use cases where hit duplication is guaranteed not to happen, such as ATLAS muon reco. 
Add a legacy `uniqueHitIds` method providing the old-style sorted hit access. 
CC @jojungge

### DIFF
--- a/Core/include/Acts/Seeding/HoughTransformUtils.hpp
+++ b/Core/include/Acts/Seeding/HoughTransformUtils.hpp
@@ -204,9 +204,9 @@ class HoughPlane {
   /// @param xBin: bin index in the first coordinate
   /// @param yBin: bin index in the second coordinate
   /// @return the list of identifiers of the hits for this cell
-  /// Can include duplicates if a hit was filled more than once 
-  std::span<const identifier_t,std::dynamic_extent> hitIds(std::size_t xBin,
-                                          std::size_t yBin) const {
+  /// Can include duplicates if a hit was filled more than once
+  std::span<const identifier_t, std::dynamic_extent> hitIds(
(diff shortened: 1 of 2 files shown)


andiwand:
I think this is a breaking change and should be flagged as `perf!: ...` in the title

does this require a patch in athena?

goblirsc:
> I think this is a breaking change and should be flagged as `perf!: ...` in the title
> 
> does this require a patch in athena?

yes, this requires a one-line patch. I can push this patch, it will cost a bit of performance but function for now, and then gain back the speed when an ACTS build with this change is released?

andiwand (approved):
👍 I can absorb the one-line fix on our canary

goblirsc:
> > I think this is a breaking change and should be flagged as `perf!: ...` in the title
> > does this require a patch in athena?
> 
> yes, this requires a one-line patch. I can push this patch, it will cost a bit of performance but function for now, and then gain back the speed when an ACTS build with this change is released?

-> https://gitlab.cern.ch/atlas/athena/-/merge_requests/81702

acts-project-service:
## :red_circle: Athena integration test results [ca2230215326a283f4f4fb284d4c24bc50d5b2bc]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/59809360) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12599209)!

PR #4532, opened 2025-08-08
perf: Fully SoA intermediate doublets in `SeedFinder`

andiwand:
The original Athena code uses SoA and I also measured this to be faster with ACTS

--- END COMMIT MESSAGE ---

Measurement incoming ...


andiwand:
In our examples this turns out to be slower. I will check with Athena again. If this turns out again faster there I would still go ahead with this.

image

andiwand:
speedup did not manifest again sadly - closing

PR #4330, opened 2025-06-06
feat: Global cluster position and improved spacepoint writers

benjaminhuth:
This adds a mechanism to compute global cluster positions, also for 1D clusters. This is also present in Athena, and e.g. use for the GNN workflow.
Also updates the Spacepoint and measurement writers to write out more information.

## Summary by CodeRabbit

- **New Features**
  - Added an option to compute and store the global position of clusters during digitization, available via both C++ and Python interfaces.
  - Measurement and spacepoint data now include global position (x, y, z) and time information in CSV and ROOT outputs.
  - Spacepoint outputs now support and record both single and paired measurement/geometry identifiers, as well as time values.

- **Bug Fixes**
  - Improved bounds checking when accessing cluster data to prevent out-of-range errors.

- **Documentation**
  - Extended data structures and output formats to reflect new global position and time fields.

PR #4544, opened 2025-08-12
feat: Write global cluster position in `RootMeasurementWriter`

andiwand:
Adds branches for global reconstructed cluster position to the `RootMeasurementWriter`. Note that these will likely only be meaningful for pixel sensors.

--- END COMMIT MESSAGE ---

@murnanedaniel @paulgessinger

### DIFF
--- a/Examples/Io/Root/src/RootMeasurementWriter.cpp
+++ b/Examples/Io/Root/src/RootMeasurementWriter.cpp
@@ -10,6 +10,8 @@
 
 #include "Acts/Definitions/TrackParametrization.hpp"
 #include "Acts/Geometry/GeometryContext.hpp"
+#include "Acts/Surfaces/RegularSurface.hpp"
+#include "Acts/Utilities/Logger.hpp"
 #include "ActsExamples/EventData/AverageSimHits.hpp"
 #include "ActsExamples/EventData/Index.hpp"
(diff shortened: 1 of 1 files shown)


paulgessinger (approved):

andiwand:
just realized that @benjaminhuth implemented something similar here https://github.com/acts-project/acts/pull/4330

paulgessinger:
Outputs change, but I think that's expected. @andiwand @benjaminhuth how to you want to proceed with this?

andiwand:
we converged to use benjamin's PR https://github.com/acts-project/acts/pull/4330

murnanedaniel:
> we converged to use benjamin's PR #4330

@andiwand Are there any major differences with Benjamin's version? I will switch across now, but just in case there are subtleties that I didn't notice.

andiwand:
> > we converged to use benjamin's PR #4330
> 
> @andiwand Are there any major differences with Benjamin's version? I will switch across now, but just in case there are subtleties that I didn't notice.

hey @murnanedaniel ! sorry I should have kept you in the loop - the change from benjamin should be equivalent. we improved the situation for 1D measurements. I think the branches are called a little bit differently.

so should be fine - let me know if you face any problems!

PR #4551, opened 2025-08-15
fix: Improve layer surface array lookup with representative intersection

andiwand:
Try to intersect the representative surface of the layer to get a better estimate of the position to lookup the surface grid. If the intersection is not successful this falls back to the original behavior.

This will have a computational overhead but should lead to better physics results. After tuning the ITk surface grid we started seeing missing innermost hits in the innermost barrel layer due to high inclination of the track.

--- END COMMIT MESSAGE ---

### DIFF
--- a/Core/src/Geometry/Layer.cpp
+++ b/Core/src/Geometry/Layer.cpp
@@ -192,9 +192,23 @@ Layer::compatibleSurfaces(const GeometryContext& gctx, const Vector3& position,
   // check the sensitive surfaces if you have some
   if (m_surfaceArray && (options.resolveMaterial || options.resolvePassive ||
                          options.resolveSensitive)) {
+    Vector3 lookupPosition = position;
+
+    // if possible use a position on the representative surface for the surface
+    // array lookup
+    if (SurfaceIntersection intersection =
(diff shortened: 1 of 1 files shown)


andiwand:
CPU performance looks fine

physmon looks affected due to different simulation results and low stats

image

paulgessinger:
Seems like this breaks the epic build

andiwand:
> Seems like this breaks the epic build

Is there any information somewhere why this is the case?

I also see `test_geomodel_G4` failing without any indication. Locally it passed for me

paulgessinger:
I think it broke already before. I'll try to see when exactly. Where is `test_geomodel_G4` failing, I don't see that?

andiwand:
This went away after retrying. I think this was just temporary

acts-project-service:
## :red_circle: Athena integration test results [f57871f4a873f7ea7d41cd80eb95f4117d9d9b77]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/60819148) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12761304)!

PR #4568, opened 2025-08-21
feat: Test temporary directory can be disabled

paulgessinger:
Sometimes it's useful to be able to put the output files in tests in the working directory, and not clean them up. With this PR, if `ACTS_NO_TEST_TEMP` is set, the temporary directory helper will just point at the local working directory, and not delete files on destruction.

--- END COMMIT MESSAGE ---


Tests/CommonHelpers/include/Acts/Tests/CommonHelpers/TemporaryDirectory.hpp
```diff
@@ -16,8 +16,28 @@
 namespace Acts::Test {
 
 class TemporaryDirectory {
+  static bool skipTemporary() {
+    static const bool flag = [] {
+      const char* envvar = std::getenv("ACTS_NO_TEST_TEMP");
+      if (envvar != nullptr) {
+        std::string val{envvar};
+        if (val != "") {
```
andiwand, line 24: from the pr description
> if `ACTS_NO_TEST_TEMP` is set

I thought an empty string would work too. but maybe that would be against env variables in general?

so in principle one would just put anything else than an empty string to flip the switch?

paulgessinger: Eh, good point I'm not quite sure. I can test this tomorrow.

asalzburger:
> Sometimes it's useful to be able to put the output files in tests in the working directory, and not clean them up. With this PR, if `ACTS_NO_TEST_TEMP` is set, the temporary directory helper will just point at the local working directory, and not delete files on destruction.
> 
> --- END COMMIT MESSAGE ---
> 
> Any further description goes here, @-mentions are ok here!
> 
> * Use a _conventional commits_ prefix: [quick summary](https://www.conventionalcommits.org/en/v1.0.0/#summary)
>   
>   * We mostly use `feat`, `fix`, `refactor`, `docs`, `chore` and `build` types.
> * A milestone will be assigned by one of the maintainers

This is really good, because sometimes you want to inspect the actual output files.

Tests/CommonHelpers/include/Acts/Tests/CommonHelpers/TemporaryDirectory.hpp
paulgessinger: Ok, so `ACTS_NO_TEST_TEMP= my_exe` results in an emptry string, and will not deactivate the functionality. @andiwand what would you expect to happen in this case?

andiwand: I think it makes sense - so the user is expected to just put anything there to deactivate it right?

acts-project-service:
## :red_circle: Athena integration test results [2743010676219f0aded56f282b5ad4bf096c93a6]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/60848145) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12765551)!

PR #4571, opened 2025-08-21
fix: Do not allow set Navigaiton Policy in the static blueprint node if the volume has already one assigned

dimitra97:
I also add a check in the `StaticBlueprintNode` finalize method in case the volume has already a navigation policy assigned before being attached to the static node

### DIFF
--- a/Core/src/Geometry/StaticBlueprintNode.cpp
+++ b/Core/src/Geometry/StaticBlueprintNode.cpp
@@ -111,7 +111,11 @@ void StaticBlueprintNode::finalize(const BlueprintOptions& options,
   if (m_navigationPolicyFactory) {
     policyFactory = m_navigationPolicyFactory.get();
   }
-  m_volume->setNavigationPolicy(policyFactory->build(gctx, *m_volume, logger));
+
+  if (!m_volume->navigationPolicy()) {
+    m_volume->setNavigationPolicy(
+        policyFactory->build(gctx, *m_volume, logger));
+  }
 
   parent.addVolume(std::move(m_volume));
 }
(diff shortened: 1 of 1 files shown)


paulgessinger (commented):
One issue with this is that the navigation policy should not be created before the volume sizing is final. This is only guaranteed to be the case at the end of the building procedure. Do you **have to** attach the policy already to the bare volumes, or could you add them to the static blueprint node, @dimitra97? If the sizing is already final (i.e. if the volumes are registered as contained children), this won't make a difference, but if not, it's more difficult to get it wrong.

dimitra97:
> One issue with this is that the navigation policy should not be created before the volume sizing is final. This is only guaranteed to be the case at the end of the building procedure. Do you **have to** attach the policy already to the bare volumes, or could you add them to the static blueprint node, @dimitra97?
> 
> If the sizing is already final (i.e. if the volumes are registered as contained children), this won't make a difference, but if not, it's more difficult to get it wrong.
> 
> I would actually propose the opposite; make the static blueprint node throw an exception if the associated volume already has a navigation policy assigned.

I see what you mean. The only reason I needed that is that I use the `MultiWireVolumeBuilder `  https://github.com/acts-project/acts/blob/main/Core/include/Acts/Geometry/MultiWireVolumeBuilder.hpp to construct the `TrackingVolume` where I attach also the navigation policy (fixed size volume - child of another volume) and then from this I construct the `StaticBlueprintNode`. One solution could be to refactor the Buidler to return the static blueprint node? Because otherwise during the blueprint construction the navigation policy is ignored and being set as default. Although I wanted to leave it more generic by returning a fixed size volume but in the case where is used in the blueprint it creates this issue.

paulgessinger:
@dimitra97 yeah I think returning the blueprint node would be a good solution. What do you think?

dimitra97:
> @dimitra97 yeah I think returning the blueprint node would be a good solution. What do you think?



> @dimitra97 yeah I think returning the blueprint node would be a good solution. What do you think?



> @dimitra97 yeah I think returning the blueprint node would be a good solution. What do you think?

Thanks Paul.  This or maybe I think I will leave it by returning the volume (in order not to affect much athena side) and attach the navigation policy extenrally when I create the blueprint node from this volume. I close this then

PR #4574, opened 2025-08-22
fix: Add tests, try to avoid UB in math helpers

paulgessinger:
There's a lot of very easily triggered UB in `MathHelpers.hpp`. This PR tries to avoid some of it, and cover it with tests.

--- END COMMIT MESSAGE ---

@junggjo9

### DIFF
--- a/Core/include/Acts/Utilities/MathHelpers.hpp
+++ b/Core/include/Acts/Utilities/MathHelpers.hpp
@@ -8,7 +8,9 @@
 
 #pragma once
 
+#include <algorithm>
 #include <cmath>
+#include <concepts>
 
 namespace Acts {
 
--- a/Core/include/Acts/Utilities/MathHelpers.hpp
+++ b/Core/include/Acts/Utilities/MathHelpers.hpp
@@ -20,6 +20,10 @@ template <typename T>
 constexpr T abs(const T n) {
   if constexpr (std::is_signed_v<T>) {
     if (n < 0) {
+      // Let's try to avoid UB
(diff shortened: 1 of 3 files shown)


Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -27,17 +33,10 @@ constexpr T abs(const T n) {
 /// @brief Calculates the ordinary power of the number x.
 /// @param x: Number to take the power from
 /// @param p: Power to take
-template <typename T, std::integral P>
+template <typename T, std::unsigned_integral P>
```
junggjo9, line 36: Do we really want to remove the possibility to have pow(x, -1)?

junggjo9: I remember, no we need to allow for negative exponents. For example, in the Chebychev polynomials https://github.com/acts-project/acts/blob/24d2caa3e072e0c01eb7ffc015efb75029bef618/Core/include/Acts/Utilities/detail/Polynomials.hpp#L129, this expression can have a negative exponent

paulgessinger: Right, but the way this is implemented, negative exponents are not safe, and can easily pop over into UB. I've found it to even segfault (probably due to stack overflow) with a negative exponent and a sufficiently small base.

paulgessinger: Concretely, I got it to segfault on `pow(1e-300, -1)`. At the very least, I would like to add a runtime check that the base is not too small.

PR #4563, opened 2025-08-20
refactor: Less allocs during clustering

CarloVarni:
This adds an alternative `createClusters ` method, while keeping the previous version for backward compatibility. The new alternative methods allows the user to provide the cluster collection and a data object that will store the label vector and additional quantities used during the clustering (thus avoiding several allocations).

When using the new alternative method I see an improvement of ~2 % in speed.

We still have a minimum of: 5  + (additional internal resizes of the disjoint set) heap allocations

PR #4584, opened 2025-08-27
refactor: Disjoint set implementation

CarloVarni:
Custom implementation of a disjoint set, to be used for clustering. From local tests, this improved timing by a few %. 

This needs https://github.com/acts-project/acts/pull/4563

### DIFF
--- a/Core/include/Acts/Clusterization/DisjointSet.hpp
+++ b/Core/include/Acts/Clusterization/DisjointSet.hpp
@@ -1,0 +1,112 @@
+// This file is part of the ACTS project.
+//
+// Copyright (C) 2016 CERN for the benefit of the ACTS project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+
+#include <vector>
+#include <boost/pending/disjoint_sets.hpp>
+#include <cstddef>
+#include <cstdint>
+#include <cstdlib>
+#include <new>
+#include <utility>
+
+namespace Acts::Ccl {
+using Label = std::size_t;
+constexpr Label NO_LABEL = 0;
+}  // namespace Acts::Ccl
+
+
+namespace Acts::Ccl {
+class DisjointSets {
+public:
+  explicit DisjointSets(std::size_t initial_size = 128)
+    : m_defaultCapacity(initial_size),
+      m_nextId(1) {
+    // index 0 is a dummy so that id == index
+    m_parent.reserve(m_defaultCapacity + 1);
+    m_rank.reserve(m_defaultCapacity + 1);
+    m_parent.push_back(0);
+    m_rank.push_back(0);
+  }
+  ~DisjointSets() = default;
+  
+  DisjointSets(const DisjointSets&) = default;
+  DisjointSets& operator=(const DisjointSets&) = default;
+
+  DisjointSets(DisjointSets&&) noexcept = default;
+  DisjointSets& operator=(DisjointSets&&) noexcept = default;
+
+  inline Acts::Ccl::Label makeSet() {
+    // If we have reach the maximum extend the capacity if the vectors
+    // and avoid the zero-filling from a resize
+    if (m_parent.size() == m_parent.capacity()) {
+      const std::size_t newCap = m_parent.capacity() * 2;
+      m_parent.reserve(newCap);
+      m_rank.reserve(newCap);
+    }
+
+    const std::size_t id = m_nextId++;
+    m_parent.push_back(id);
+    m_rank.push_back(0);
+    return id;
+  }
+
+  inline void unionSet(std::size_t x, std::size_t y) noexcept {
+    std::size_t rootX = findRoot(x);
+    std::size_t rootY = findRoot(y);
+    if (rootX == rootY) return;
+
+    const std::size_t rankRootX = m_rank[rootX];
+    const std::size_t rankRootY = m_rank[rootY];
+    if (rankRootY < rankRootX) {
+      m_parent[rootX] = rootY;
+    } else if (rankRootX < rankRootY) {
+      m_parent[rootY] = rootX;
+    } else {
+      m_parent[rootY] = rootX;
+      m_rank[rootX] = rankRootX + 1;
+    }
+  }
+
+  inline Acts::Ccl::Label findSet(std::size_t x) noexcept {
+    return findRoot(x);
+  }
+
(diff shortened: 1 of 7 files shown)


Core/include/Acts/Clusterization/DisjointSet.hpp
```diff
@@ -0,0 +1,107 @@
...
+    m_parent.push_back(0);
+    m_rank.push_back(0);
+  }
+  ~DisjointSets() = default;
+
+  DisjointSets(const DisjointSets&) = default;
+  DisjointSets& operator=(const DisjointSets&) = default;
+
+  DisjointSets(DisjointSets&&) noexcept = default;
+  DisjointSets& operator=(DisjointSets&&) noexcept = default;
```
andiwand: nitpick: in principle you can drop all of these - they will be implicitly defaulted

Core/include/Acts/Clusterization/DisjointSet.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+  inline Acts::Ccl::Label makeSet() {
+    // If we have reach the maximum extend the capacity if the vectors
+    // and avoid the zero-filling from a resize
```
andiwand:
```suggestion
    // If we have reached the maximum, extend the capacity and avoid the zero-filling from a resize
```
not sure if I got it right

Core/include/Acts/Clusterization/DisjointSet.hpp
```diff
@@ -0,0 +1,107 @@
...
+    if (m_parent.size() == m_parent.capacity()) {
+      const std::size_t newCap = m_parent.capacity() * 2;
+      m_parent.reserve(newCap);
+      m_rank.reserve(newCap);
+    }
```
andiwand, line 38: wouldn't the `push_back` auto-grow the vector with uninitialized memory as well?

Core/include/Acts/Clusterization/DisjointSet.hpp
```diff
@@ -0,0 +1,107 @@
...
+    m_parent.push_back(id);
+    m_rank.push_back(0);
+    return id;
+  }
+
+  inline void unionSet(std::size_t x, std::size_t y) noexcept {
+    std::size_t rootX = findRoot(x);
+    std::size_t rootY = findRoot(y);
+    if (rootX == rootY)
+      return;
```
andiwand:
```suggestion
    if (rootX == rootY) {
      return;
    }
```
the linter will flag this

Core/include/Acts/Clusterization/DisjointSet.hpp
```diff
@@ -0,0 +1,107 @@
...
+    m_rank.resize(1);
+  }
+
+ private:
+  inline std::size_t findRoot(std::size_t x) noexcept {
+    assert(x < m_parent.size());
+    while (true) {
+      std::size_t parent = m_parent[x];
+      if (parent == x)
+        return x;
```
andiwand:
```suggestion
      if (parent == x) {
        return x;
      }
```

Core/include/Acts/Clusterization/DisjointSet.hpp
```diff
@@ -0,0 +1,107 @@
...
+      // also check the grand-parent
+      std::size_t grandparent = m_parent[parent];
+      // halve path so that next search is faster
+      m_parent[x] = grandparent;
+      x = grandparent;
+      if (x == m_parent[x])
+        return x;
```
andiwand:
```suggestion
      if (x == m_parent[x]) {
        return x;
      }
```

Core/include/Acts/Clusterization/DisjointSet.hpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+
+ private:
+  std::size_t m_defaultCapacity{128};
+  std::size_t m_nextId{1};
+  std::vector<std::size_t> m_parent{};
+  std::vector<std::size_t> m_rank{};
```
andiwand:
```suggestion
  std::vector<std::size_t> m_parent;
  std::vector<std::size_t> m_rank;
```
nitpick: `{}` is not necessary since `std::vector` has a default constructor

Core/include/Acts/Clusterization/DisjointSet.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+#pragma once
+
+#include <cassert>
+#include <utility>
+#include <vector>
+
+namespace Acts::Ccl {
+using Label = std::size_t;
+constexpr Label NO_LABEL = 0;
```
andiwand:
```suggestion
static constexpr Label NO_LABEL = 0;
```

Core/include/Acts/Clusterization/DisjointSet.hpp
```diff
@@ -0,0 +1,107 @@
...
+    // and avoid the zero-filling from a resize
+    if (m_parent.size() == m_parent.capacity()) {
+      const std::size_t newCap = m_parent.capacity() * 2;
+      m_parent.reserve(newCap);
+      m_rank.reserve(newCap);
+    }
+
+    const std::size_t id = m_nextId++;
+    m_parent.push_back(id);
+    m_rank.push_back(0);
```
andiwand:
```suggestion
    const std::size_t id = m_nextId;
    ++m_nextId;
    m_parent.push_back(id);
    m_rank.push_back(0);
```
we try to avoid relying on the "result" of `++int` or `int++`

Core/include/Acts/Clusterization/DisjointSet.hpp
CarloVarni: in the previous version we were calling resize and filling all the entries with zeros. this is now avoided and we simply extends the cpacity.

Core/include/Acts/Clusterization/DisjointSet.hpp
CarloVarni: yes, but the compiler decides by how much (50% or 100%). here we impose *2 factor

Core/include/Acts/Clusterization/DisjointSet.hpp
andiwand: ok lets drop this

Core/include/Acts/Clusterization/DisjointSet.hpp
andiwand: I see - in this case we should add a comment why we do this instead of auto-grow

CarloVarni: added

andiwand:
closing due to inactivity

PR #4585, opened 2025-08-27
chore: Cleanup Kalman Fitter in Examples

andiwand:
- use mbf smoother
- reserve space in track and track state container
- clean namespaces
- clean includes
- remove implicit track state container copy during reverse filter

--- END COMMIT MESSAGE ---

I measure now ~ 49 ms for a truth tracking ttbar PU200 ODD workflow which is about 50 us per track on my M3 pro. On the treadripper I see about 100 us per track


Examples/Algorithms/TrackFitting/src/TrackFittingAlgorithm.cpp
```diff
@@ -159,9 +167,11 @@ ActsExamples::ProcessCode ActsExamples::TrackFittingAlgorithm::execute(
...
   }
 
-  std::stringstream ss;
-  trackStateContainer->statistics().toStream(ss);
-  ACTS_DEBUG(ss.str());
+  if (logger().doPrint(Acts::Logging::DEBUG)) {
+    std::stringstream ss;
+    trackStateContainer->statistics().toStream(ss);
+    ACTS_DEBUG(ss.str());
+  }
```
benjaminhuth, line 174:
```suggestion
  ACTS_DEBUG([](){
    std::stringstream ss;
    trackStateContainer->statistics().toStream(ss);
    return ss.str();
  }());
```

andiwand: I think `DEBUG` is preferable because `VERBOSE` will contain all the propagation verbose logs

to me the `if` looks easier to read but the lambda looks interesting. I did't think of that

andiwand:
gsf changed because kf prefit did because of smoother change

Examples/Algorithms/TrackFitting/src/TrackFittingAlgorithm.cpp
benjaminhuth: yeah not really important, you in principle can avoid the redundancy of `DEBUG` by this...

PR #4594, opened 2025-08-29
add neutral interaction list

Biblehome:
`feat`: add PointLikeProcess and StandardNeutralInteractionList and a FatrasPointLikeProcess unit test

### DIFF
--- a/Fatras/include/ActsFatras/Kernel/PointLikeProcess.hpp
+++ b/Fatras/include/ActsFatras/Kernel/PointLikeProcess.hpp
@@ -1,0 +1,87 @@
+// This file is part of the ACTS project.
+//
+// Copyright (C) 2016 CERN for the benefit of the ACTS project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+#include "Acts/Material/MaterialSlab.hpp"
+#include "ActsFatras/EventData/Particle.hpp"
+
+#include <limits>
+
+namespace ActsFatras {
+
+/// A continuous simulation process based on a physics model plus selectors.
+///
+/// @tparam physics_t is the physics model type
+/// @tparam input_particle_selector_t is the input particle selector
+/// @tparam output_particle_selector_t is the output particle selector
+/// @tparam child_particle_selector_t is the child particle selector
+///
+/// The physics model type **must** provide a call operator with the following
+/// two member functionss
+///
+///     // generate X0/L0 limits
+///     template <typename generator_t>
+///     std::pair<Scalar, Scalar>
+///     generatePathLimits(
+///         generator& rng,
+///         const Particle& particle) const
+///
+///     // run the process simulation
+///     template <typename generator_t>
+///     bool
+///     run(
+///         generator_t& rng,
+///         Particle& particle,
+///         std::vector<Particle>& generatedParticles) const
+///
+/// The return type of generatePathLimits() can be any `Container` with
+/// `Particle` elements.
+///
+/// The input selector defines whether the process is applied while the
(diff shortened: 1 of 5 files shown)


Fatras/include/ActsFatras/Kernel/PointLikeProcess.hpp
```diff
@@ -0,0 +1,87 @@
...
+/// `Particle` elements.
+///
+/// The input selector defines whether the process is applied while the
+/// output selector defines a break condition, i.e. whether to continue
+/// simulating the particle propagation. The child selector is used to
+/// filter the generated child particles.
+///
+/// @note The output and child particle selectors are identical unless the
+///       child particle selector is explicitly specified.
+template <typename physics_t, typename input_particle_selector_t,
```
paulgessinger: Since this is new, you could add the `PointLikeProcessConcept` from `InteractionList.hpp` as an actual type requirement. I don't think we have concepts for the other template parameters, though, but that could be very interesting to add as well at some point.

Biblehome: Ok, thanks, I'll add it.

Tests/UnitTests/Fatras/Kernel/PointLikeProcessTests.cpp
```diff
@@ -0,0 +1,164 @@
...
+//      generator_t & /*generator*/, const Acts::MaterialSlab & /*slab*/,
+//      ActsFatras::Particle & /*particle*/) const {
+//    // create daughter particles
+//    return {
+//        Particle().setAbsoluteMomentum(1_GeV),
+//        Particle().setAbsoluteMomentum(2_GeV),
+//        Particle().setAbsoluteMomentum(3_GeV),
+//        Particle().setAbsoluteMomentum(4_GeV),
+//    };
+//  }
```
andiwand: dead code?

Fatras/include/ActsFatras/Kernel/PointLikeProcess.hpp
```diff
@@ -0,0 +1,87 @@
...
+#pragma once
+
+#include "Acts/Material/MaterialSlab.hpp"
+#include "ActsFatras/EventData/Particle.hpp"
+
+#include <limits>
+
+namespace ActsFatras {
+
+/// A continuous simulation process based on a physics model plus selectors.
```
andiwand:
```suggestion
/// A point like simulation process based on a physics model plus selectors.
```

Fatras/include/ActsFatras/Kernel/PointLikeProcess.hpp
```diff
@@ -0,0 +1,87 @@
...
+  template <class generator_t>
+  std::pair<double, double> generatePathLimits(generator_t& generator, const Particle& particle) const
+  {
+    return physics.generatePathLimits(generator, particle);
+  }
+
+  template <class generator_t>
+  bool run(generator_t& rng, Particle& particle, std::vector<Particle>& generatedParticles) const
+  {
+    if(!selectInputParticle(particle)) return false;
```
andiwand:
```suggestion
    if (!selectInputParticle(particle)) {
      return false;
    }
```

Fatras/include/ActsFatras/Kernel/PointLikeProcess.hpp
Biblehome: It seems I need to include the whole InteractionList.hpp file here. Is it a better solution to move these concepts in a separate header?

paulgessinger:
Thanks @Biblehome ! Now the only thing missing the code formatting. Can you run `pre-commit` locally and push a fixed commit?

Biblehome:
> Thanks @Biblehome ! Now the only thing missing the code formatting. Can you run `pre-commit` locally and push a fixed commit?

OK.

paulgessinger (dismissed):

andiwand:
@paulgessinger we are missing CI permissions here

PR #4600, opened 2025-08-29
better barcodes

pbalek:
allowing larger values for barcode elements

--- END COMMIT MESSAGE ---

Barcode is a unique identifier of a particle. It used to be a single `std::uint64_t`, where 5 values were encoded via `MultiIndex`:
 * primary vertex ID
 * secondary vertex ID
 * particle ID
 * generation ID
 * sub-particle ID

This PR changes the single 64-bit number into 5 separate numbers, allowing larger range for each of them. If necessary, enlarging the range should be easy.

Along this change, modifications of writers and readers are necessary. Some of them can accommodate a vector instead of a number (such as ROOT files) but for some of them, it is necessary to save 5 numbers instead of 1 (such as CSV files).


Examples/Io/Obj/src/ObjSimHitWriter.cpp
```diff
@@ -89,12 +89,10 @@ ActsExamples::ProcessCode ActsExamples::ObjSimHitWriter::writeT(
       }
       ACTS_VERBOSE("Accepting: Hit above threshold: " << momentum);
 
-      if (particleHits.find(simHit.particleId().value()) ==
-          particleHits.end()) {
-        particleHits[simHit.particleId().value()] = {};
+      if (particleHits.find(simHit.particleId().hash()) == particleHits.end()) {
+        particleHits[simHit.particleId().hash()] = {};
```
paulgessinger: Is this needed if the `std::hash` function is correctly specialized?

Examples/Io/Root/src/RootAthenaDumpReader.cpp
paulgessinger, line 1: @benjaminhuth

Fatras/include/ActsFatras/EventData/Barcode.hpp
```diff
@@ -91,81 +94,166 @@ namespace ActsFatras {
...
+  constexpr ValueGen generation() const { return generationID; }
   /// Return the sub-particle identifier.
-  constexpr Value subParticle() const { return level(4); }
+  constexpr ValuePart subParticle() const { return subParticleID; }
+
+  std::vector<std::uint32_t> getData() const {
+    return {vertexPrimary(), vertexSecondary(), particle(), generation(),
+            subParticle()};
+  }
+  void setData(const std::vector<std::uint32_t>& data) {
```
paulgessinger: Nitpick: consider using an `std::span` and making the `assert` a check also in non-debug builds.

Fatras/include/ActsFatras/EventData/Barcode.hpp
```diff
@@ -91,81 +94,166 @@ namespace ActsFatras {
...
+
+  Acts::HashedString hash() const {
+    std::string str = std::to_string(vertexPrimary()) + "-" +
+                      std::to_string(vertexSecondary()) + "-" +
+                      std::to_string(particle()) + "-" +
+                      std::to_string(generation()) + "-" +
+                      std::to_string(subParticle());
+
+    return Acts::hashStringDynamic(str);
+  }
```
paulgessinger: This will be **very** expensive, I think. Maybe we could use boost's [`hash_combine`](https://www.boost.org/doc/libs/1_89_0/libs/container_hash/doc/html/hash.html#ref_hash_combine) of [`hash_range`](https://www.boost.org/doc/libs/1_89_0/libs/container_hash/doc/html/hash.html#ref_hash_range) for this?

Examples/Io/Root/include/ActsExamples/Io/Root/RootSimHitReader.hpp
```diff
@@ -87,16 +87,18 @@ class RootSimHitReader : public IReader {
...
       "tx",  "ty", "tz",      "tt",      "tpx",     "tpy",
       "tpz", "te", "deltapx", "deltapy", "deltapz", "deltae"};
-  constexpr static std::array<const char *, 2> m_uint64Keys = {"geometry_id",
-                                                               "particle_id"};
+  constexpr static std::array<const char *, 1> m_uint64Keys = {"geometry_id"};
   constexpr static std::array<const char *, 6> m_uint32Keys = {
       "event_id", "volume_id",   "boundary_id",
       "layer_id", "approach_id", "sensitive_id"};
   constexpr static std::array<const char *, 1> m_int32Keys = {"index"};
+  constexpr static std::array<const char *, 1> m_vec_uint32Keys = {"barcode"};
```
andiwand:
```suggestion
  constexpr static std::array<const char *, 1> m_vecUint32Keys = {"barcode"};
```

Examples/Io/Root/include/ActsExamples/Io/Root/RootSimHitReader.hpp
```diff
@@ -87,16 +87,18 @@ class RootSimHitReader : public IReader {
...
 
   std::unordered_map<std::string_view, float> m_floatColumns;
   std::unordered_map<std::string_view, std::uint32_t> m_uint32Columns;
   std::unordered_map<std::string_view, std::int32_t> m_int32Columns;
+  std::unordered_map<std::string_view, std::vector<std::uint32_t> *>
+      m_vec_uint32Columns;
```
andiwand:
```suggestion
      m_vecUint32Columns;
```

Examples/Io/Root/include/ActsExamples/Io/Root/RootSimHitWriter.hpp
```diff
@@ -77,7 +78,7 @@ class RootSimHitWriter final : public WriterT<SimHitContainer> {
   /// Hit surface identifier.
   std::uint64_t m_geometryId = 0;
   /// Event-unique particle identifier a.k.a. barcode.
-  std::uint64_t m_particleId = 0;
+  std::vector<std::uint32_t> m_barcode = {0, 0, 0, 0, 0};
```
andiwand: `std::array` does not work with root?

Fatras/include/ActsFatras/EventData/Barcode.hpp
```diff
@@ -91,81 +94,166 @@ namespace ActsFatras {
...
-class Barcode : public Acts::MultiIndex<std::uint64_t, 12, 12, 16, 8, 16> {
-  using Base = Acts::MultiIndex<std::uint64_t, 12, 12, 16, 8, 16>;
-
+class Barcode {
  public:
-  using Base::Base;
-  using Base::Value;
+  using ValueVtx = std::uint16_t;
+  using ValuePart = std::uint32_t;
+  using ValueGen = std::uint8_t;
```
andiwand:
```suggestion
  using VertexId = std::uint16_t;
  using ParticleId = std::uint32_t;
  using GenerationId = std::uint8_t;
```

andiwand: I think it would be good to have typedefs for each part, even if they end up having the type

Examples/Io/Root/src/VertexNTupleWriter.cpp
```diff
@@ -938,7 +938,7 @@ void VertexNTupleWriter::writeTrackInfo(
     } else {
       ACTS_VERBOSE("Track has no matching truth particle.");
 
-      innerTrkParticleId.push_back(-1);
+      innerTrkParticleId.push_back({0, 0, 0, 0, 0});
```
andiwand: Can you add a static factory function to `Barcode` for this? something like `Barcode::Invalid()` for example?

Fatras/include/ActsFatras/EventData/Barcode.hpp
```diff
@@ -91,81 +94,166 @@ namespace ActsFatras {
...
-  constexpr Value particle() const { return level(2); }
+  constexpr ValuePart particle() const { return particleID; }
   /// Return the generation identifier.
-  constexpr Value generation() const { return level(3); }
+  constexpr ValueGen generation() const { return generationID; }
   /// Return the sub-particle identifier.
-  constexpr Value subParticle() const { return level(4); }
+  constexpr ValuePart subParticle() const { return subParticleID; }
+
+  std::vector<std::uint32_t> getData() const {
```
andiwand:
```suggestion
  std::array<std::uint32_t, 5> asVector() const {
```
typedef might be good e.g.
```c++
using IdVector = std::array<std::uint32_t, 5>
```

Fatras/include/ActsFatras/EventData/Barcode.hpp
```diff
@@ -91,81 +94,166 @@ namespace ActsFatras {
...
+    return lhs.subParticle() < rhs.subParticle();
+  }
+
+  friend constexpr bool operator==(Barcode lhs, Barcode rhs) {
+    return lhs.vertexPrimary() == rhs.vertexPrimary() &&
+           lhs.vertexSecondary() == rhs.vertexSecondary() &&
+           lhs.particle() == rhs.particle() &&
+           lhs.generation() == rhs.generation() &&
+           lhs.subParticle() == rhs.subParticle();
+  }
```
andiwand: I think these could be implemented on top of a `std::array`

Fatras/include/ActsFatras/EventData/Barcode.hpp
```diff
@@ -91,81 +94,166 @@ namespace ActsFatras {
...
+  using ValueGen = std::uint8_t;
 
   // Construct an invalid barcode with all levels set to zero.
-  constexpr Barcode() : Base(Base::Zeros()) {}
+  constexpr Barcode()
+      : vertexPrimaryID(0u),
+        vertexSecondaryID(0u),
+        particleID(0u),
+        generationID(0u),
+        subParticleID(0u) {}
```
andiwand: I think it would be good to replace that with an explicit `Invalid()` factory function

andiwand (commented):
This is great - thank you for changing this @pbalek !

Examples/Io/Root/include/ActsExamples/Io/Root/RootSimHitWriter.hpp
pbalek: Hi @andiwand, unfortunately no. I tried using `std::array` but it's impossible to save vector of arrays to the root file. On the other hand, `std::vector` works smoothly.

Fatras/include/ActsFatras/EventData/Barcode.hpp
pbalek: Hi @paulgessinger, how would use `std::span`? From what I tried, it's possible to feed it by a vector of any length, although it's not really in the standard...
Also, what do you mean by "making the `assert` a check also in non-debug builds"? I thought `assert` is present in any build, is it not? If not, would `throw std::invalid_argument(..)` be better?

paulgessinger: @pbalek 

- `std::span` can by fixed or dynamic size. In this case, I would suggest a dynamically sized span. I think this is standardized behavior.
- `assert` is only executed in DEBUG builds (`#ifndef NDEBUG`) so I think throwing an exception like you suggested would be best.

Fatras/include/ActsFatras/EventData/Barcode.hpp
pbalek: indeed 👍

Fatras/include/ActsFatras/EventData/Barcode.hpp
pbalek: ah, so we can have `std::span` and `setData` will accept both `std::vector` and `std::array` 👍

Fatras/include/ActsFatras/EventData/Barcode.hpp
pbalek: it seems `boost::hash_value()` is exactly what is needed here

Fatras/include/ActsFatras/EventData/Barcode.hpp
pbalek: It's not possible to save `std::array` nor `std::vector<std::array>` to a root file, so I'd stick with `vector` here. However, a name `asVector` is nicely intuitive.

Examples/Io/Obj/src/ObjSimHitWriter.cpp
pbalek: if `particleHits` are defined as `std::unordered_map<ActsFatras::Barcode, ...>`, then `hash()` is indeed not needed

Fatras/include/ActsFatras/EventData/Barcode.hpp
pbalek: renamed and added types for each identifier 👍

Fatras/include/ActsFatras/EventData/Barcode.hpp
pbalek: ~I got rid of empty constructor and~ introduced `static constexpr Barcode Invalid()`

Examples/Io/Root/src/VertexNTupleWriter.cpp
pbalek: replaced by `ActsFatras::Barcode::Invalid().asVector()`

Fatras/include/ActsFatras/EventData/Barcode.hpp
pbalek: Ok, I think I was too quick with this one. I can't get rid of `Barcode()` completely because it's used in `std::unordered_map`. That can't be changed without writing some horrible workarounds.
So, I removed all mentions of `Barcode()` in the code. Now, the code uses either `Barcode(prim_vtx,sec_vtx,part,gen,subpart)` constructor, or `Barcode::Invalid()`. Constructor `Barcode()` still exists, though.

Examples/Framework/include/ActsExamples/EventData/SimParticle.hpp
```diff
@@ -203,24 +203,20 @@ struct CompareParticleId {
 };
 struct PrimaryVertexIdGetter {
   SimBarcode operator()(const SimParticleState& particle) const {
-    return SimBarcode(0u).setVertexPrimary(
-        particle.particleId().vertexPrimary());
+    return SimBarcode(particle.particleId().vertexPrimary(), 0, 0);
```
paulgessinger: This is a personal opinion, but I think I preferred the setters in this case because it gives names to the components being set. With the constructor arguments you need tooling support to be able to see which number is which.

Examples/Framework/include/ActsExamples/EventData/SimVertex.hpp
```diff
@@ -18,44 +18,48 @@ namespace ActsExamples {
...
+  constexpr SimVertexBarcode& setVertexSecondary(SecondaryVertexId id) {
     return m_id.setVertexSecondary(id), *this;
   }
   /// Set the particle identifier.
-  constexpr SimVertexBarcode& setGeneration(Value id) {
+  constexpr SimVertexBarcode& setGeneration(GenerationId id) {
     return m_id.setGeneration(id), *this;
   }
 
+  Acts::HashedString hash() const { return m_id.hash(); }
```
paulgessinger: That happens to work because the return type is a typedef to a `uint32`. I think that should be made explicit though.

Fatras/include/ActsFatras/EventData/Barcode.hpp
```diff
@@ -91,81 +95,159 @@ namespace ActsFatras {
...
+  constexpr ParticleId particle() const { return particleID; }
   /// Return the generation identifier.
-  constexpr Value generation() const { return level(3); }
+  constexpr GenerationId generation() const { return generationID; }
   /// Return the sub-particle identifier.
-  constexpr Value subParticle() const { return level(4); }
+  constexpr SubParticleId subParticle() const { return subParticleID; }
+
+  /// Export barcode as vector
+  constexpr std::vector<std::uint32_t> asVector() const {
```
paulgessinger, line 140: This is being called by the spaceship operator and the hash function, which means that it will allocate every single time a hash container will do any operation. 

At this point, can't the internal storage of the numbers just be an `std::array` with an enum being used to index into it? OTOH, with both a vector and an array we fully lose the ability to store a different number of bits for each component. 

To my mind the best solution would be to keep the separate members, avoid the vector creation entirely, handle hash using `hash_combine` and implement the spaceship operator by hand.

paulgessinger (commented):
One more aspect about the setters which I'm happy to discuss;

I changed the geometry identifiers away from mutable setters that return a reference to themselves over to const methods that will always produce an updated copy. I've found this to be cleaner and easier to reason about in terms of which instance of the id changes. 

Since this PR already touches the API, maybe it would make sense to also make this type of change at the same time?

Examples/Framework/include/ActsExamples/EventData/SimVertex.hpp
pbalek: Oh, I would prefer this didn't work, so I wouldn't missed it. Thanks, I fixed it.

Examples/Framework/include/ActsExamples/EventData/SimParticle.hpp
pbalek: Hi @paulgessinger, wouldn't this together with you comment above mean that code e.g. 
`auto p1 = Barcode().setVertexPrimary(1).setVertexSecondary(2).setGeneration(3);`
will make 4 different objects? Or will the compiler optimize this out, so that's not an issue?

paulgessinger: I guess you're referring to my top-level comment about the const methods?

The idea would be for the compiler to optimize this out, yes. But you're right of course that now that the barcode is not a single integer anymore, it might be less obvious that it will actually do that. 

With the mutable setters returning `*this` I would think that the object only gets created once and then the mutating functions are inlined.

Fatras/include/ActsFatras/EventData/Barcode.hpp
pbalek: I didn't realize that it's creating new and new vectors. I changed the `<=>` operator and `hash()` uses `boost::hash_combine`, so the function is called only from outside of the class.

Examples/Framework/include/ActsExamples/EventData/SimParticle.hpp
pbalek: I'm not sure if compiler can optimize it. I tried it with `std::cout` print outs in the constructor, and it was indeed called several times. However, the presence of the print out may prevent the optimization...

I also tried to measure the impact on timing but there's no measurable impact. 
Thus, if you think having `const` setters is better, I'll change them. Was it `GeometryIdentifier.hpp` that you had in mind? So I can change it in the same style (with "deprecated")?

paulgessinger: Anyway, let's stick to the current chaining now. Consider switching to the `set` methods over the constructor with multiple numbers, but I won't insist on this.

paulgessinger: I think the `std::cout` side-effect is guaranteed to happen regardless of optimizations. You're right that we can't guarantee its fully optimized out. This [Godbolt](https://godbolt.org/z/bnn1PWWv9) listing suggests if the function is inlinable, it would be combined into a single operation starting from `-O1`, which I would argue is acceptable.

pbalek: > This Godbolt listing suggests if the function is inlinable, it would be combined into a single operation starting from -O1, which I would argue is acceptable. 

Interesting, thank you for the link!

I changed the code to have `with...` everywhere. For most of the places, it sets 1 parameter of the time. For a few readers, `withData` sets all 5 parameters at once.

Examples/Algorithms/Geant4/src/ParticleTrackingAction.cpp
```diff
@@ -163,13 +163,8 @@ std::optional<SimBarcode> ParticleTrackingAction::makeParticleId(
...
   auto pid = eventStore().trackIdMapping.at(parentId).makeDescendant();
-
-  auto key = EventStore::BarcodeWithoutSubparticle::Zeros();
-  key.set(0, pid.vertexPrimary())
-      .set(1, pid.vertexSecondary())
-      .set(2, pid.particle())
-      .set(3, pid.generation());
-  pid.setSubParticle(++eventStore().subparticleMap[key]);
+  auto key = pid.withoutSubparticle();
+  pid = pid.withSubParticle(++eventStore().subparticleMap[key]);
```
paulgessinger: Nitpick: separate the increment and assignment. It's easy to confuse whether this increements the number before returning or after.

Examples/Framework/include/ActsExamples/EventData/SimVertex.hpp
paulgessinger: Sorry I'm still stumbling over this: should the return type here really be `HashedString`? The return type of `m_id.hash();` is `std::size_t`, right?

paulgessinger:
Since we're on a breaking cycle anyway, should we declare this breaking as well?

Examples/Algorithms/Geant4/src/ParticleTrackingAction.cpp
pbalek: Fixed!

Examples/Framework/include/ActsExamples/EventData/SimVertex.hpp
pbalek: `std::size_t` is indeed the only type that `hash()` should return (for barcode, of course). I fixed this and went through the rest of the code. I'm moderately confident it's not present anywhere else.

paulgessinger:
Looks good. Unfortunately there are some conflicts with the reference updates with another PR. Can you update the references again?

acts-project-service:
## :red_circle: Athena integration test results [e3e275ac46b7e8947f58b9bc03e6d5b185652cc0]
Please investigate the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12949308)!

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971962)||
|:red_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971961)||
|:red_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971960)||
|:red_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971959)||
|:red_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971958)||
|:yellow_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971957)||
|:red_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971956)||
|:red_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971955)||
|:red_circle: | [test_ActsCheckObjectCountsWorkflowHgtd](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971954)||
|:red_circle: | [test_ActsCheckObjectCountsCachedWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971953)||
|:red_circle: | [test_ActsCheckObjectCountsLegacyWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971952)||
|:red_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971951)||
|:red_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971950)||
|:red_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971949)||
|:red_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971948)||
|:red_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971947)||
|:red_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971946)||
|:red_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971945)||
|:red_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971944)||
|:red_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971943)||
|:red_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971942)||
|:red_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971941)||
|:red_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971940)||
|:red_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971939)||
|:red_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971938)||
|:red_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971937)||
|:red_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971936)||
|:red_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971935)||
|:yellow_circle: | [test_ActsGx2fRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971934)||
|:red_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971933)||
|:red_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971932)||
|:red_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971931)||
|:red_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/61971930)||

PR #4613, opened 2025-09-03
perf: Reimplement `CylinderSurface::intersection` in local frame

andiwand:
Reimplement `CylinderSurface::intersection` based on the local frame. The local frame allows to decompose the intersection into subspaces which should result in less computation. In case the cylinder is not rotated or translated the global to local is skipped.

--- END COMMIT MESSAGE ---

Measurement incoming ...

### DIFF
--- a/Core/src/Surfaces/CylinderSurface.cpp
+++ b/Core/src/Surfaces/CylinderSurface.cpp
@@ -258,8 +258,8 @@ SurfaceMultiIntersection CylinderSurface::intersect(
   if (lro_perp2 > r2) {
     // check the distance of closest approach to the origin in XY
     const Vector2 a = lro.head<2>();
-    const Vector2 n = lrd.head<2>() / std::sqrt(lrd_perp2);
-    const double d2 = (a - a.dot(n) * n).squaredNorm();
+    const Vector2 n = lrd.head<2>();
+    const double d2 = (a - a.dot(n) * n / lrd_perp2).squaredNorm();
     // if the distance of closest approach is greater than the radius, there is
     // no intersection
     if (d2 > r2) {


andiwand:
image

image

Core/src/Surfaces/CylinderSurface.cpp
```diff
@@ -196,88 +197,98 @@ Vector3 CylinderSurface::rotSymmetryAxis(const GeometryContext& gctx) const {
...
-            *this,
-            boundaryTolerance};
+  Vector3 lro = position;   //< local frame ray origin
+  Vector3 lrd = direction;  //< local frame ray direction
+  if (!gctxTransform.linear().isIdentity()) {
+    lro = gctxTransform.inverse() * position;
+    lrd = gctxTransform.linear().inverse() * direction;
+  } else if (!gctxTransform.translation().isZero()) {
+    lro -= gctxTransform.translation();
   }
```
paulgessinger, line 213: These checks will not be *free*, although I suppose they're cheaper than the extra work to do otherwise. We also unfortunately can't cache this, because technically it can vary context-by-context.

... **but**: in Gen3 we could add an overload without the geometry context that e.g. uses a cached boolean. We could then call that from a navigation policy which knows that the cylinder surface is a boundary, which we don't expect to ever move based on the geometry context.

Core/src/Surfaces/CylinderSurface.cpp
```diff
@@ -196,88 +197,98 @@ Vector3 CylinderSurface::rotSymmetryAxis(const GeometryContext& gctx) const {
...
+  const auto buildSolution = [&](double path) -> Intersection3D {
+    Vector3 solution = position + path * direction;
+    IntersectionStatus status = std::abs(path) < std::abs(tolerance)
+                                    ? IntersectionStatus::onSurface
+                                    : IntersectionStatus::reachable;
+    status = boundaryCheck(solution, path, status);
+    return {solution, path, status};
+  };
+
+  const double lrd_perp2 = lrd[0] * lrd[0] + lrd[1] * lrd[1];
```
paulgessinger, line 243: `VectorHelpers::perp` does this I think?

Core/src/Surfaces/CylinderSurface.cpp
```diff
@@ -196,88 +197,98 @@ Vector3 CylinderSurface::rotSymmetryAxis(const GeometryContext& gctx) const {
...
+    const Vector2 a = lro.head<2>();
+    const Vector2 n = lrd.head<2>() / std::sqrt(lrd_perp2);
+    const double d2 = (a - a.dot(n) * n).squaredNorm();
+    // if the distance of closest approach is greater than the radius, there is
+    // no intersection
+    if (d2 > r2) {
+      return {{Intersection3D::invalid(), Intersection3D::invalid()},
+              *this,
+              boundaryTolerance};
+    }
```
paulgessinger: I think this is very similar to what I do in the `CylinderNavigationPolicy`.

Core/src/Surfaces/CylinderSurface.cpp
andiwand: `perp` will do a `sqrt`

paulgessinger: You're right, sorry!

andiwand:
After having a closer look I cannot reproduce the speedup in a microbenchmark and also the assembly seems to be larger due to the transform check (300 original VS 400 changed). Removing the transform check reduces the assembly drastically (400 changed before VS 250 changed after).

In principle we could determine the transform state once, store it in the surface and then use this as a condition. But the speedup is only about 10% for the cylinder intersection.

Closing for now ...

PR #4616, opened 2025-09-04
fix: Performance improvements for AdaptiveHoughTransformSeeder

kciesla93:
This PR fixes some performance issues found in `AdaptiveHoughTransformSeeder`:

- All arguments passed to function bound with `std::bind_front` are taken as `std::cref()` to avoid copies.
- All templates taking `std::bind_front` are now using `F&&` instead of `F`.
- Added `std::vector::reserve` calls where possible.
- Replaced `std::deque` with `std::vector`.


tboldagh:
Thanks @kciesla93 it looks very good to me.

andiwand (dismissed):
LGTM

there are still formatting issue the CI is flagging - can you resolve them @kciesla93 ?

kciesla93:
@andiwand Done

andiwand (approved):

andiwand:
@paulgessinger can you provide ci permission or merge this by hand?

PR #4615, opened 2025-09-04
feat: geometric digitisation for ODD and fix of error parameterisation scripts

asalzburger:
This introduces a odd-digitisation config for the full detector.
* geometric digitisation for all detector components
* fix: geometric digitisation json ( 'rms' entry was holding `rms * rms`) 
* fix `error_parametrisation.py` script which was desynchronised
* add pull distribution plotting option

--- END COMMIT MESSAGE ---

Change of root file hashes expected

To reproduce, run in your favourite environment 
``
python <path_to_src>/Examples/Scripts/Python/digitization.py -n 100000 -d odd -t geometric
``
followed by:
``
python <path_to_src>/Examples/Scripts/Digitization/error_parameterisation.py --root measurements.root --json-in <path_to_src>i/Examples/Configs/odd-digi-geometric-config.json --json-out new_with_variances.json
``

Screenshot 2025-09-04 at 14 57 28

Any further description goes here, @-mentions are ok here!

- Use a *conventional commits* prefix: [quick summary](https://www.conventionalcommits.org/en/v1.0.0/#summary)
  - We mostly use `feat`, `fix`, `refactor`, `docs`, `chore` and `build` types.
- A milestone will be assigned by one of the maintainers

PR #4619, opened 2025-09-05
feat: make ODD LongStrip measurements true 1D in geometric digitisation

asalzburger:
This PR sits on top of #4615 and turns the LongStrip measurements into pure 1D measurements.
It also adapts the the `error_parameterisation.py` script to deal with that correctly and adds some convenient DEBUG output to the DigitizationAlgorithm.

The `error_parameterisation.py` script is also augmented with an overview plot to give first feedback.

It will change performance, so the root file hashes will be updated after the first CI run.

--- END COMMIT MESSAGE ---

Screenshot 2025-09-05 at 15 29 48


andiwand (dismissed):

asalzburger:
This will likely need a root hash update.

andiwand (approved):

paulgessinger:
Any chance you could add the `error_parameterisation.py` to the CI at this point?

asalzburger:
> Any chance you could add the `error_parameterisation.py` to the CI at this point?

Hi, I will do this in a follow-up PR with switching some more tests on, if that's ok.

asalzburger:
> > Any chance you could add the `error_parameterisation.py` to the CI at this point?
> 
> Hi, I will do this in a follow-up PR with switching some more tests on, if that's ok.

Also, I discussed with @noemina - that it would be great to have this usable for ATLAS/ITk as well, so I will do the same way as I did with the `MaterialTrackIO`, i.e. moving the Tree (future RNtuple) definition and filling into `Plugins/Root`.

PR #4637, opened 2025-09-22
feat: Introduce TrackToTruthJetAlgorithm

delitez:
In this PR, an algorithm that matches Kalman tracks to jets, which are clustered from truth particles using FastJet (See [PR:4267](https://github.com/acts-project/acts/pull/4267)), is introduced. Additional WIP: Overlap removal, jet labeling, use of HepMC3 event.


delitez:
Tagging @paulgessinger @pbutti @murnanedaniel @benjaminhuth

Examples/Framework/include/ActsExamples/EventData/TrackJet.hpp
```diff
@@ -0,0 +1,99 @@
...
+#include "Acts/Definitions/Algebra.hpp"
+
+#include <iostream>
+#include <vector>
+
+namespace ActsExamples {
+// TODO: Capitalize enum names for consistency
+enum jetlabel { unknown = -99, ljet = 0, cjet = 4, bjet = 5 };
+// TODO: Remove hadronlabel
+enum hadronlabel {
```
paulgessinger: Where these the final enums that we were using? I thought I remember they were moved to a central place and made consistent. Can you check?

If the jet class lives in Core, these enums also have to live in Core to be usable.

Examples/Algorithms/Jets/src/TruthJetAlgorithm.cpp
paulgessinger, line 1: I would suggest the fist PR to be the labeling tooling, including the truth particle helpers that I grabbed from ATLAS.

Examples/Framework/include/ActsExamples/EventData/TrackJet.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+  std::vector<int> getConstituents() const { return m_constituents; }
+
+  Acts::Vector4 getFourMomentum() const { return m_fourMomentum; }
+
+  void addTrack(const int trk_idx) { m_trk_idxs.push_back(trk_idx); }
+
+  std::vector<int> getTracks() const { return m_trk_idxs; }
+
+  void print() const {
```
paulgessinger: Refactor in favor of an output stream operator please

paulgessinger:
```cpp
std::ostream& operator<< (std::ostream& os, const S& s) {
    os << "my nice S";
    return os;
}
```

Examples/Framework/include/ActsExamples/EventData/TrackJet.hpp
```diff
@@ -0,0 +1,69 @@
...
+ public:
+  explicit TrackJet(const Acts::Vector4& fm) { m_fourMomentum = fm; }
+
+  // TODO::Pass references instead of copies.
+
+  void setConstituents(const std::vector<int>& constituents) {
+    m_constituents = constituents;
+  }
+
+  std::vector<int> getConstituents() const { return m_constituents; }
```
paulgessinger:
```suggestion
  const std::vector<int>& getConstituents() const { return m_constituents; }
```

Examples/Framework/include/ActsExamples/EventData/TrackJet.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+  Acts::Vector4 getFourMomentum() const { return m_fourMomentum; }
+
+  void addTrack(const int trk_idx) { m_trk_idxs.push_back(trk_idx); }
+
+  std::vector<int> getTracks() const { return m_trk_idxs; }
```
paulgessinger:
```suggestion
  const std::vector<int>& getTracks() const { return m_trk_idxs; }
```

Examples/Python/src/TruthJet.cpp
```diff
@@ -29,4 +30,13 @@ void addTruthJet(Context& ctx) {
   ACTS_PYTHON_DECLARE_ALGORITHM(TruthJetAlgorithm, mex, "TruthJetAlgorithm",
                                 inputTruthParticles, outputJets, jetPtMin);
 }  // addTruthJet
+
+void addTrackToTruthJet(Context& ctx) {
+  auto mex = ctx.get("examples");
+
+  ACTS_PYTHON_DECLARE_ALGORITHM(TrackToTruthJetAlgorithm, mex,
```
paulgessinger, line 37: Can you do this in the existing `addTruhJet` function?

Examples/Python/src/TruthJetStub.cpp
```diff
@@ -11,4 +11,5 @@
 
 namespace ActsPython {
 void addTruthJet(Context& /*ctx*/) {}
+void addTrackToTruthJet(Context& /*ctx*/) {}
```
paulgessinger, line 14: I think this can be done in the same function if we're not going to build either of them separately and not the other one.

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -182,6 +192,10 @@
     outputDir=str(outputDir),
 )
 
+if args.jet_clustering:
+    args.reco = False  # disable track reco if jet clustering is requested
+    args.ttbar = True  # force ttbar for jet clustering
```
paulgessinger, line 197: What's the idea behind this? In principle, the algorithm should work either way right?

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -330,6 +344,79 @@
     ),
 )
 
+if args.jet_clustering:
+    addSeeding(
```
paulgessinger, line 348: Why do we need a separate track reco in the jet case? Can't this just use the track reco that is configured otherwise?

Plugins/FastJet/include/Acts/Plugins/FastJet/Jets.hpp
```diff
@@ -83,6 +84,76 @@ class TrackJetBuilder {
   fastjet::ClusterSequence m_clusterSeq{};
 };
 
+class TruthJetBuilder {
```
paulgessinger, line 87: The naming is a bit confusing to me: In the end we have a class that represents a *jet*, and that jet has constituents. At the most basic level, it shouldn't matter a lot what the constituents are. FastJet's own data model uses the four momentum of whatever object that was passed in.

The main thing I see the need for is that we want to (optionally) support associating additional objects like tracks to a jet, and we want to ultimately store the jet label on the object.

- `TruthJetBuilder` does not seem to build anything, it just holds the four momentum
- `JetProperties` is a separate class, which I don't fully understand

Maybe we could use inheritance for this (but I'm not sure jet).

- We could have a `Jet` base class that stores the four momentum associated with the jet.
- Then we could have a `TruthJet` class that carries a vector of particles as constituents. Since this is a separate plugin, we'd have to stick to using indices, or we make this plugin depend on Fatras so we can pull in either the barcode or the particle type itself. In addition this class would carry a vector of indices for the associated tracks, to avoid having to template on the track container type. (This could be solved in the future if we add a type-erased track proxy type)
- A separate `TrackJet` class could instead hold indices to tracks as the constituents. 

I guess in the first iteration, since the constituents would always just be indices, the member and methods could also live in the `Jet` base class, until the types are sorted out down the line.

Plugins/FastJet/include/Acts/Plugins/FastJet/Jets.hpp
```diff
@@ -83,6 +84,76 @@ class TrackJetBuilder {
...
+ public:
+  explicit TruthJetBuilder(const Acts::Vector4& fm) { m_fourMomentum = fm; }
+  const Acts::Vector4 getTruthJetFourMomentum() const { return m_fourMomentum; }
```
paulgessinger, line 90: Generally, a method like this could just be called `fourMomentum` in my opinion.

Plugins/FastJet/include/Acts/Plugins/FastJet/Jets.hpp
murnanedaniel: My two cents: Having a base `Jet` class makes a lot of sense. Both truth-based and track-based forms quack like a jet and walk like a jet - they were just built in different ways. They both (I assume) would still be able to associate with tracks in their cone, and true matching particles. Subclassing I guess is just a convenience to avoid confusion. But keeping as much of the properties and methods in a base will make life a lot easier later when comparing e.g. ceiling performance of jet reconstruction vs. real performance. Then it will be trivial (change the name of the collection).

Plugins/FastJet/include/Acts/Plugins/FastJet/Jets.hpp
```diff
@@ -83,6 +84,76 @@ class TrackJetBuilder {
...
+
+ private:
+  Acts::Vector4 m_fourMomentum{0., 0., 0., 0.};
+};
+
+class JetProperties {
```
pbutti, line 96: Why do we need this class?
The following methods are all valid for a jet object. 

Why do we need a jet and a jet property class? 
I think the following members could (and maybe should) be associated to a jet object, in my opinion.

Plugins/FastJet/include/Acts/Plugins/FastJet/Jets.hpp
```diff
@@ -83,6 +84,76 @@ class TrackJetBuilder {
...
+      : m_truthJet{truthJet} {}
+
+  /// @brief Set the jet constituents
+  /// @param constituents the indices of the constituent tracks
+  void setConstituents(const std::vector<int>& constituents) {
+    m_constituents = constituents;
+  }
+  /// @brief Get the jet constituents
+  /// @return the indices of the constituent tracks
+  const std::vector<int>& getConstituents() const { return m_constituents; }
```
pbutti, line 110: If this class uses the truth jet builder, why the constitues are commented as "constituent tracks" ? 
They are particles, I would suppose.

Plugins/FastJet/include/Acts/Plugins/FastJet/Jets.hpp
pbutti: I still not fully understand. 
A jet is a well-defined object that comes from the clustering of other objects, whatever they are (truth particles, tracks, calo clusters, ... ) under certain rules. 
In my opinion already having TruthJets and TrackJets is even too much as they really are the same thing just constructed with different objects. Also, a reco jet can still have a truth label associated to it, which is the regular way jets are identified for tagging purposes. 

In my view we only would need a Jet class, which holds vectors with associated objects: tracks, particles or calo clusters (if available). To my view this is a similar idea of what ATLAS uses (https://acode-browser.usatlas.bnl.gov/lxr/source/athena/Event/xAOD/xAODJet/xAODJet/versions/Jet_v1.h) 

If we want to make more derived classes, keep in mind that truthJets can have both particles and tracks associated and so on, so a truth jet should have a method to associate tracks to it.

paulgessinger: @pbutti 

> which holds vectors with associated objects: tracks, particles or calo clusters (if available)

that's precisely what I think we could use the subclasses for. Otherwise you'd have these objects on all jets, or the constituents are not strongly typed.

```cpp
class Jet {
  Vector4 m_fourMomentum;
};

class TruthJet : public Jet {
  std::vector<SimParticle> m_constituents;
  std::vector<TrackProxy> m_associatedTracks; // difficult due to lack of type-erasure
};

class TrackJet : public Jet {
  std::vector<TrackProxy> m_constituents; // difficult due to lack of type-erasure
};
```

\+ various getters setters etc.

whereas with a single class it would have to be

```cpp
class Jet {
  Vector4 m_fourMomentum;
  std::vector<unsigned int> m_constituents; // undefined what these are and which collection to index into
  std::vector<unsigned int> m_tracks; // redundant if `m_constituents` are already tracks
  std::vector<unsigned int> m_caloClusters;
};
```

andiwand: personally I would not couple reconstruction and truth EDMs. the redundancy is quite small

paulgessinger: I think my main point is that I'd prefer having a strongly typed `TruthJet` where it's clear that the constituents are truth particles, and ideally are linked as strong types as well.

PR #4639, opened 2025-09-23
ci: switch eic-shell build to use container that doesn't provide acts

veprbl:
This fixes the latest regression with libFatras being picked up from /opt/local/lib of the container. Instead of trying to fix the existing hacks, just getting rid of the pre-installed package is preferred.

--- END COMMIT MESSAGE ---

There is another ongoing issue in which no notifications get delivered when github mentions @acts-project/epic-contacts . We need to resolve that somehow.


paulgessinger:
@veprbl this seems like a great idea. I'll have a closer look tomorrow!

acts-project-service:
## :red_circle: Athena integration test results [e7ce3f036bdc45c23a87c456f21c20a79954eea1]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/62065119) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/12963562)!

PR #4676, opened 2025-09-30
build: Make codegen respect system library flag

stephenswat:
This commit makes the codegen respect the `ACTS_USE_SYSTEM_LIBS` flag, allowing users to build ACTS in a fully offline environment if they have the dependencies already installed. To make this possible, I added a code path which avoids uv and uses the (virtual) Python environment provided by the user.

This is the second attempt at #4141.


stephenswat:
To aid review, please note that:

1. The block of code for setting up uv is unchanged, but I had to put it into a conditional block so the indentation changed.
2. The way this works is by creating a temporary virtual environment that mirrors the outside virtual environment, and then I install the local `codegen` package into that. The outside environment is unchanged.
3. I removed the `build-system` block from the pyproject.toml so it can be build with setuptools rather than requiring uv-build. I think setuptools is much more widely available.

cmake/ActsCodegen.cmake
```diff
@@ -116,8 +167,25 @@ function(acts_code_generation)
...
+                # If we are not using uv, then we use pip to install the
+                # package into the virtual environment in the build directory.
+                # The --no-build-isolation flag ensures that we don't
+                # automatically download setuptools. The --no-index flag
+                # ensures that nothing can be downloaded ever, and the
+                # --no-deps is necessary to convince pip that the dependencies
+                # are already in the environment.
+                execute_process(
+                    COMMAND
+                        ${CMAKE_BINARY_DIR}/codegen_venv/bin/pip install
```
paulgessinger: Maybe it's safer to use pip through python as well like `python3 -m pip`? But I'm not sure here.

cmake/ActsCodegen.cmake
```diff
@@ -1,74 +1,125 @@
...
+    )
+    message(STATUS "uv version: ${uv_version}")
+else()
+    message(
+        STATUS
+        "Configuring codegen in offline mode: preparing virtual environment"
+    )
 
-    file(REMOVE ${UV_DIR}/uv.tar.gz)
+    find_package(Python 3.13 REQUIRED COMPONENTS Interpreter)
```
veprbl: Is there a reason to require 3.13? All should work perfectly fine with 3.12 and, perhaps, even older.

veprbl:
Also, could you explain, what was the idea behind installing packages into venv? If system packages are available, they should be already usable for the purposes of codegen?

paulgessinger:
Part of the codegen code is installed as a package, and we don't want to install it into the site environment (and probably can't)

cmake/ActsCodegen.cmake
paulgessinger: For the `uv` mode, I locked the python version to ensure we get identical output. If you're in a custom environment, I suppose we don't need to do that.

stephenswat: This was requested by @paulgessinger to ensure that the results of this codegen are always consistent.

veprbl: Unless there is a concrete evidence for non-reproducibility, I would remove/relax the version check. Also worth to point out that this does not require a specific version, just sets a lower limit.

stephenswat:
Updated to remove the Python 3.13 requirement and to use `python -m pip` rather than `pip`.

acts-project-service:
## :red_circle: Athena integration test results [470f8c86829d79187629ad90e00a974e24c87725]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/62776072) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13063465)!

PR #4680, opened 2025-09-30
feat: Add jet labeling, jet overlap removal and HepMC3 event

delitez:
This PR follows the [PR #4637](https://github.com/acts-project/acts/pull/4637) and adds jet labeling, overlap removal and access to pythia-8 truth particles via inclusion of HepMC3Event for jet clustering and track-to-jet matching algorithms.


Core/include/Acts/Definitions/ParticleData.hpp
```diff
@@ -79,4 +80,21 @@ std::ostream& operator<<(std::ostream& os, PdgParticle pdg);
 /// @return Optional string view of particle name
 std::optional<std::string_view> pdgToShortAbsString(PdgParticle pdg);
 
+namespace ParticleId {
```
paulgessinger, line 83: Add documentation comments please

Core/include/Acts/Definitions/PdgParticle.hpp
```diff
@@ -83,4 +84,21 @@ static constexpr std::pair<std::int32_t, std::int32_t> extractNucleusZandA(
   return std::make_pair(Z, A);
 }
 
+enum class HadronType {
```
paulgessinger, line 87: Documentation comments

Core/src/Definitions/ParticleId.cpp
paulgessinger, line 1: This file contains code I pulled from ATLAS. I don't think we have precedence for this. From a license point of view, I think this is fine (and @EdwardMoyse had informally confirmed this).

@asalzburger @andiwand @CarloVarni @noemina

Core/include/Acts/Utilities/VectorHelpers.hpp
paulgessinger, line 1: Consider breaking this out into a separate PR.

Examples/Algorithms/Generators/ActsExamples/Generators/EventGenerator.cpp
paulgessinger, line 1: Consider breaking this out into a separate PR. The idea is that we can later use this by convention to identify the hard-scatter event.

paulgessinger (commented):
Here's some comments on the preliminary version of the PR.

I think it might make sense to try to break this up further. I think the changes to the HepMC3 conversion flow, and also the additional index attribute are independent from the other changes (might have to go first, though).

Let me know what you think @delitez!

PR #4681, opened 2025-09-30
feat: Port hashing seeder prototype to `Seeding2`

andiwand:
Ports the hashing seeding plugin and prototype Examples algorithm to the new space point and seeding API.

While doing that, I removed the `Plugin/Hashing` because this is more of a proof of concept. The relevant code was simply moved into the Examples algorithm implementation.

This implementation does not depend on the space point grid anymore and will only work on the predicted space point buckets. For now, the space points remain in an unsorted order without any compute performance tuning.

--- END COMMIT MESSAGE ---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,7 +220,7 @@ if(ACTS_BUILD_PYTHON_WHEEL)
     # CMAKE_PREFIX_PATH should be site-packages (steered by scikit-build-core)
     set(CMAKE_INSTALL_BINDIR ${SKBUILD_NULL_DIR}) # Don't install binaries/scripts
     set(CMAKE_INSTALL_LIBDIR "acts")
-    set(CMAKE_INSTALL_INCLUDEDIR ${SKBUILD_NULL_DIR}}) # Don't install headers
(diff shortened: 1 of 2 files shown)


Plugins/Hashing/src/HashingAlgorithm.cpp
```diff
@@ -0,0 +1,132 @@
...
+    resultSets[binIndex].insert(spacePoint.index());
+    for (const auto& neighborSpacePointIndex : neighborSpacePointIndices) {
+      resultSets[binIndex].insert(neighborSpacePointIndex);
+    }
+  }
+
+  std::vector<std::vector<SpacePointIndex2>> result;
+  result.reserve(resultSets.size());
+  for (const auto& spSet : resultSets) {
+    result.emplace_back(spSet.begin(), spSet.end());
```
CouthuresJeremy: A filter on empty spSet might be missing here, leading to as much returned bucket as the number of bins (while empty buckets are not needed) which might be more than the number of space points.

CouthuresJeremy (approved):
Looks good to me!

paulgessinger (approved):

PR #4638, opened 2025-09-22
unit test for failing tryall policy

dimitra97:
PLEASE DESCRIBE YOUR CHANGES.
THIS MESSAGE ENDS UP AS THE COMMIT MESSAGE.
DO NOT USE @-MENTIONS HERE!

--- END COMMIT MESSAGE ---

Any further description goes here, @-mentions are ok here!

- Use a *conventional commits* prefix: [quick summary](https://www.conventionalcommits.org/en/v1.0.0/#summary)
  - We mostly use `feat`, `fix`, `refactor`, `docs`, `chore` and `build` types.
- A milestone will be assigned by one of the maintainers

PR #4693, opened 2025-10-03
feat: Add test for same position surfaces in gen3

dimitra97:
This checks how the propagation work for surfaces placed in the same position for gen3 geometry (case for trigger muon chambers in atlas).
From navigator side it works. 
From propagator side the change for the `prestepSurface` is required in order not to reject one of the two surfaces.

Also I am reverting the change in the TryAll Policy. I think after the Navigator refactoring for gen3 https://github.com/acts-project/acts/pull/4638 it is not needed.
Unless you have objections or you think makes more sense- it can be kept
Branching out of: https://github.com/acts-project/acts/pull/4638

@andiwand @junggjo9 
--- END COMMIT MESSAGE ---

### DIFF
--- a/Core/include/Acts/Propagator/Propagator.ipp
+++ b/Core/include/Acts/Propagator/Propagator.ipp
@@ -60,8 +60,16 @@ Acts::Result<void> Acts::Propagator<S, N>::propagate(
         // navigator, so we skip this target.
         // This can also happen in a well-behaved geometry with external
         // surfaces.
+        if (state.options.onSurfaceAccept) {
+          ACTS_VERBOSE(
+              "Pre-step surface status is onSurface and flag is enabled, keep "
+              "the target "
+              << nextTarget.surface().geometryId());
+          return nextTarget;
+        }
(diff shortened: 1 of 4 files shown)


dimitra97:
I am closing this!!! Because we said maybe it is not that good to touch the propagator with this change but see how to handle in geometry (e.g shifting of the surfaces)

junggjo9:
In the ATLAS geoemetry the eta & phi surfaces are overlayed on top... I'd still vote for having this supported

andiwand:
> In the ATLAS geoemetry the eta & phi surfaces are overlayed on top... I'd still vote for having this supported

Overlapping surfaces is something that our navigation cannot handle by definition right now and I do not see an easy way to make it reliably work. Anything I have seen related to that until now is a major hack and I would not do that given how much work it was to make the navigation work in the first place.

I did not quite understand why you depend on this and why these surfaces cannot be simply separated in space for example. But lets find a solution and discuss this offline.

PR #4705, opened 2025-10-10
chore: move Plugin python bindings into `Python/Plugins`

asalzburger:
### **This will be split into smaller PRs to trace eventual problems**

This PR moves the Plugins related code into `Python/Plugins` - each plugin will appear in a dedicated library `ActsPluginsPythonBinding@PluginName@` and for each plugin a dedicated `plugin_name.py` is autogenerated that:
- guards the import and throws a warning if the import doesn't work
- applies the `_patch_config` from the adapter.

The dependencies now look pretty clear, with this PR:

*  only one occasion in `Examples/Python` refers back to the main module, this is due to the `FpeMonitor` being in python, but only making sense in the context of the Examples.
* only one module defines actually plugin bindings, i.e. the `GNN` because the backends are only defined there.

This PR will be followed by one doing the same for `Examples/Python` -> `Python/Examples`.

It should not change test results.

--- END COMMIT MESSAGE ---

"main" module needed only in `Framework`

Screenshot 2025-10-16 at 11 06 29

Very strong reduction of `ActsPlugins` appearance in `Python/Examples`:

Screenshot 2025-10-16 at 11 05 50

### DIFF
--- a/Examples/Python/CMakeLists.txt
+++ b/Examples/Python/CMakeLists.txt
@@ -297,12 +297,12 @@ foreach(f ${py_files})
     get_filename_component(_dir ${_target} DIRECTORY)
     file(MAKE_DIRECTORY ${_dir})
 
-    file(
-        CREATE_LINK ${CMAKE_CURRENT_SOURCE_DIR}/python/acts/${f} ${_target}
-        SYMBOLIC
-    )
-
     get_filename_component(_rel ${f} DIRECTORY)
+
+    install(
+        FILES ${CMAKE_CURRENT_SOURCE_DIR}/python/acts/${f}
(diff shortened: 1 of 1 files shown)


Examples/Io/Root/CMakeLists.txt
```diff
@@ -2,7 +2,6 @@ acts_add_library(
     ExamplesIoRoot
     SHARED
     src/RootMeasurementWriter.cpp
-    src/RootMaterialDecorator.cpp
```
asalzburger, line 5: This was malplaced, it should (and now does) live in `ActsPlugins/Root` this is corrected here such that the Plugins python bindings are already correct.

asalzburger:
Superseded with #4750

PR #4710, opened 2025-10-14
fix: Avoid venv race condition in offline build

stephenswat:
In #4676, I implemented the possibility to use the code generation tooling for the ACTS code generation in an offline environment by creating a single virtual environment and reusing it across multiple custom targets. While this works naively, it introduces a race condition in which multiple threads modify the same virtual environment at the same time. In order to resolve this problem, I changed the build system to use a separate virtual environment for every code generation target.


stephenswat:
Unfortunately this will need a bit of follow-up work still, as I had misunderstood the core problem.

PR #4706, opened 2025-10-10
refactor: Common Jet Class

delitez:
In this PR, TrackJets and TruthJets are introduced separately and are inherited from a common "Jet" class.

PR #4725, opened 2025-10-22
feat: Extend HepMC3 utilities for hs vertex for jet labeling

delitez:
In this PR, methods to find hard-scatter vertices are implemented in HepMC3 utilities, to be later used by the jet labeling.

### DIFF
--- a/Examples/Algorithms/Jets/include/ActsExamples/Jets/TruthJetAlgorithm.hpp
+++ b/Examples/Algorithms/Jets/include/ActsExamples/Jets/TruthJetAlgorithm.hpp
@@ -36,7 +36,23 @@ class TruthJetAlgorithm final : public IAlgorithm {
     /// Output jets collection.
     std::string outputJets;
     /// Minimum jet pT.
-    double jetPtMin;
+    double jetPtMin = 20 * Acts::UnitConstants::GeV;
+    /// Jet eta range a pair of doubles defaulted to -inf/+inf
+    std::pair<double, double> jetEtaRange = {
+        -std::numeric_limits<double>::infinity(),
+        std::numeric_limits<double>::infinity()};
+    /// Jet clustering radius
+    double jetClusteringRadius = 0.4;
+    /// Only cluster HS particles
+    bool clusterHSParticlesOnly = true;
+    /// Do jet labeling
+    bool doJetLabeling = true;
+    /// Delta R for labeling
+    double jetLabelingDeltaR = 0.4;
+    /// Minimum hadron pT for labeling
+    double jetLabelingHadronPtMin = 5 * Acts::UnitConstants::GeV;
+    /// Only label HS hadrons
+    bool jetLabelingHSHadronsOnly = true;
   };
 
   TruthJetAlgorithm(const Config& cfg, Acts::Logging::Level lvl);
--- a/Examples/Algorithms/Jets/src/TruthJetAlgorithm.cpp
+++ b/Examples/Algorithms/Jets/src/TruthJetAlgorithm.cpp
@@ -8,14 +8,19 @@
 
 #include "ActsExamples/Jets/TruthJetAlgorithm.hpp"
 
+#include "Acts/Definitions/ParticleData.hpp"
+#include "Acts/Definitions/PdgParticle.hpp"
 #include "Acts/Definitions/Units.hpp"
 #include "Acts/Utilities/Logger.hpp"
+#include "Acts/Utilities/ScopedTimer.hpp"
 #include "ActsExamples/Framework/AlgorithmContext.hpp"
 #include "ActsFatras/EventData/ProcessType.hpp"
 
-#include <ostream>
+#include <algorithm>
+#include <ranges>
 #include <stdexcept>
 
+#include <boost/container/flat_map.hpp>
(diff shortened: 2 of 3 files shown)


Core/include/Acts/Definitions/ParticleData.hpp
```diff
@@ -79,4 +80,20 @@ std::ostream& operator<<(std::ostream& os, PdgParticle pdg);
...
+namespace ParticleId {
+
+bool isHadron(int pdg);
+bool isLepton(int pdg);
+bool isMuon(int pdg);
+bool isElectron(int pdg);
+bool isPhoton(int pdg);
+bool isTau(int pdg);
+bool isQuark(int pdg);
+bool isInteracting(int pdg);
```
paulgessinger: These probably need doxygen-style comments

Examples/Algorithms/Jets/include/ActsExamples/Jets/TruthJetAlgorithm.hpp
```diff
@@ -45,8 +47,8 @@ class TruthJetAlgorithm final : public IAlgorithm {
   Config m_cfg;
   ReadDataHandle<SimParticleContainer> m_inputTruthParticles{
       this, "inputTruthParticles"};
-  WriteDataHandle<std::vector<fastjet::PseudoJet>> m_outputJets{this,
-                                                                "outputJets"};
+  WriteDataHandle<std::vector<ActsPlugins::FastJet::TruthJet<TrackContainer>>>
```
paulgessinger: Consider using a typedef in a file like `ActsExamples/EventData/Jet.hpp`

```cpp
using TruthJet = ActsPlugins::FastJet::TruthJet<TrackContainer>;
```

Examples/Framework/include/ActsExamples/EventData/SimParticle.hpp
```diff
@@ -171,9 +175,17 @@ class SimParticle final {
...
+  void setGenParticle(const HepMC3::GenParticle* genParticle) {
+    m_genParticle = genParticle;
+  }
+  /// Get the HepMC3::GenParticle pointer.
+  const HepMC3::GenParticle* genParticle() const { return m_genParticle; }
+
  private:
   SimParticleState m_initial;
   SimParticleState m_final;
+  const HepMC3::GenParticle* m_genParticle{nullptr};
```
paulgessinger, line 188: Any thoughts @andiwand? (I came up with this, but I was a bit hesitant to add it here`. HepMC3 is essentially required now, and this makes mapping back much simpler.

Examples/Io/HepMC3/src/HepMC3Util.cpp
```diff
@@ -596,4 +597,58 @@ HepMC3Util::NormalizeResult HepMC3Util::normalizeFiles(
...
 }
 
+namespace {
+int eventGeneratorIndexImpl(const auto& obj) {
+  return obj
+      .template attribute<HepMC3::IntAttribute>(
+          std::string{HepMC3Util::kEventGeneratorIndexAttribute})
+      ->value();
+}
+}  // namespace
```
paulgessinger, line 632: @andiwand @murnanedaniel this was my best idea to identify the HS vertex IF we run the simulation inside ACTS: this just retains an index from the input HepMC3 events and assumes that in that set of events, index 0 corresponds to the HS event.

What do you think?

paulgessinger:
Putting this on draft until #4706 is merged.

Examples/Framework/include/ActsExamples/EventData/SimParticle.hpp
andiwand: Looks fine by me - at some point we may need to revisit this and see if improvements could be made by for example containerizing the particles collection and deduplicate some of the fields

Examples/Io/HepMC3/src/HepMC3Util.cpp
andiwand: looks ok to me

andiwand (dismissed):

andiwand:
something seems odd here. `Examples/Python/tests/test_examples.py::test_material_mapping` keeps consistently hanging in the ci-bridge but I don't see why this would be the case @delitez @paulgessinger

delitez:
> something seems odd here. `Examples/Python/tests/test_examples.py::test_material_mapping` keeps consistently hanging in the ci-bridge but I don't see why this would be the case @delitez @paulgessinger

I see that material recording test fails in prior. "FAILED Examples/Python/tests/test_examples.py::test_material_recording "
I'll try to reproduce is locally, thanks!

paulgessinger:
What's weird is that the jobs are *hanging*. So they seem to either be waiting on something or there's some form of infinite loop somewhere. Looking at the diff, it doesn't seem obvious where that comes from.

andiwand:
is this locally reproducible @delitez ? for the CI it seems to happen 100% of the time

if you think the changes can be split into smaller pieces we could start bisecting this by merging individual parts

delitez:
Hi @andiwand and @paulgessinger, I was able to reproduce it locally. The issue seems to be coming from Geant4MaterialRecording, where input HepMC3 particles are used. I'll convert this PR into draft while investigating it.

PR #4735, opened 2025-10-24
feat: added  new algorithm to clusterize cells in-place

goetzgaycken:
The in-place clusterization could be used instead of  the default clusterization but has a slightly different interface. It clusterizes by sorting the cell container in such a way that adjacent cells with the same label belong to the same cluster. This is advantageous since an additional cluster collection is avoid if the  cluster collection resulting from the clusterization is only temporary.


goetzgaycken:
-  In synthetic benchmarks it seems to be 20+-4% faster than the current clusterization. 
-  When running a digitization only example job using a test sample (ttbar, PU200, Geant4, ODD), the improvements seem more substantial. A reason might be that the code path with the current clusterization is not fully optimized.

goetzgaycken:
I haven't heard any strong voices concerning the demonstration of this new clusterization algorithm  in the examples framework. Therefore, the plan would be to leave for the time being the option in the DigitizationAlgorithm to toggle between the two clusterization algorithm which slightly increases its complexity. The longer term plan would be to only have one clusterization algorithm.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+/// @brief base concept for the container that can be used for a cell collection
+// @TODO require that container can be sorted.
+template <typename T_Container>
+concept SequenceContainer = requires(T_Container &cont, unsigned int index) {
+  { cont[index] } -> std::convertible_to<typename T_Container::value_type>;
+} && requires(T_Container &cont) {
+  { cont.empty() } -> std::convertible_to<bool>;
+} && requires(T_Container &cont) {
+  { cont.size() } -> std::convertible_to<std::size_t>;
+};
```
CarloVarni:
```suggestion
concept SequenceContainer = requires(T_Container &cont, unsigned int index) {
  { cont[index] } -> std::convertible_to<typename T_Container::value_type>;
  { cont.empty() } -> std::convertible_to<bool>;
  { cont.size() } -> std::convertible_to<std::size_t>;
} ;
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+concept CellWithLabel = requires(const T_Cell &a) {
+  traits::getLabel(a);
+} && requires(const T_Cell &a, decltype(traits::getLabel(a)) idx) {
+  traits::setLabel(a, idx);
+} && requires(const T_Cell &a, unsigned int axis_i) {
+  traits::getCellCoordinate(a, axis_i);
+} && requires { traits::getCellDimension<T_Cell>(); };
+
+/// @brief base concept for the container that can be used for a cell collection
+// @TODO require that container can be sorted.
```
CarloVarni: why don't we already require this with a `std::sortable` ?

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+#include <algorithm>
+#include <array>
+#include <cassert>
+#include <cmath>
+#include <limits>
+#include <ranges>
+#include <span>
+#include <type_traits>
+
+namespace InPlaceClusterization {
```
CarloVarni: Namespace should be `Acts`. Then you can add something in addition to it but it should still be something like `Acts::SOMETHING`

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+// For cells that should be clusterized the following functions
+// need to be implemented:
+namespace traits {
+// @brief Get the number of dimensions of the grid the cell lives on
+template <typename T_Cell>
```
CarloVarni: please adapt all the code to the ACTS convention of using `cell_t` instead

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+inline auto getLabel(const T_Cell &a) {
+  return a.m_label;
+}
+}  // namespace traits
+
+/// @brief A possible Cell type, that can be clustered
+/// T_Coordinates should be a signed or unsigned integer
+/// T_Index should be an unsigned integer
+template <typename T_Coordinates, std::size_t N, typename T_Index>
```
CarloVarni: why don't we impose it?

```suggestion
/// T_Index should be an unsigned integer
template <typename T_Coordinates, std::size_t N, std::uint32_t T_Index>
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+/// @brief Get the label associated to the given cell
+template <typename T_Cell>
+inline auto getLabel(const T_Cell &a) {
+  return a.m_label;
+}
+}  // namespace traits
+
+/// @brief A possible Cell type, that can be clustered
+/// T_Coordinates should be a signed or unsigned integer
```
CarloVarni: should we add a concept to impose that?

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+      : m_coordinates(coordinates), m_label(T_Index{}), m_srcIndex(src_index) {}
+
+  std::array<T_Coordinates, N>
+      m_coordinates;   ///< the coordinates of the cell on the regular grid
+  T_Index m_label;     ///< a label which will be assigned by the clusterization
+  T_Index m_srcIndex;  ///< the index to find the source cell
+};
+
+/// @brief concept of a call object that can be clustered.
+// @TODO also require decltype(getLabel(cell)) is unsigned integer
```
CarloVarni: can we impose it now?

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+template <typename T_Cell>
+concept CellWithLabel = requires(const T_Cell &a) {
+  traits::getLabel(a);
+} && requires(const T_Cell &a, decltype(traits::getLabel(a)) idx) {
+  traits::setLabel(a, idx);
+} && requires(const T_Cell &a, unsigned int axis_i) {
+  traits::getCellCoordinate(a, axis_i);
+} && requires { traits::getCellDimension<T_Cell>(); };
```
CarloVarni:
```suggestion
concept CellWithLabel = requires(const T_Cell &a,  decltype(traits::getLabel(a)) idx, unsigned int axis_i) {
  traits::getLabel(a);
  traits::setLabel(a, idx);
  traits::getCellCoordinate(a, axis_i);
  traits::getCellDimension<T_Cell>(); 
} ;
```

Also, there is no need of the `&` in `const T_Cell &a`  in this case

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+    connected &= a_coordinate_diff <= 1;
+  }
+  return connected;
+}
+
+/// @brief test whether cells are connected assuming 4 fold connectivity
+/// @param coordinates_diff absolute values of the differences of the cell coordinates
+/// Test whether two cells with the given coordinate differences are connected
+// either horizontally, vertically or are equal.
+template <typename T_Coordinates>
```
CarloVarni: in this context,  `T_Coordinates ` must be an iterable object containing coordinates?

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+inline bool isConnected4(const T_Coordinates &coordinates_diff) {
+  int connections = 0;
+  for (const auto &a_coordinate_diff : coordinates_diff) {
+    connections += a_coordinate_diff;
+  }
+  return connections <= 1;
+}
+
+/// @brief test whether cells are connected assuming 4 or 8 fold connectivity
+enum EFold { k8FoldConnection, k4FoldConnection };
```
CarloVarni: what about 
```suggestion
enum class EFold : std::uint32_t { k8FoldConnection, k4FoldConnection };
```
or whatever type this is supposed to be used as

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+      std::remove_cvref_t<decltype(traits::getCellCoordinate(cells[0], 0))>;
+  static_assert(std::numeric_limits<T_Index>::max() >=
+                std::numeric_limits<T_Label>::max());
+  // cells are sorted in the first coordinate
+  // thus can stop searching for adjacent cells
+  // if distance in the sorted coordinate is too large
+  for (T_Index idx_a = 0; idx_a < cells.size(); ++idx_a) {
+    traits::setLabel(cells[idx_a], idx_a);
+    for (T_Index idx_b = idx_a; idx_b-- > 0;) {
+      std::array<T_Coordinate, NDim> diff;
```
CarloVarni:
```suggestion
      std::array<T_Coordinate, NDim> diff {};
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+        idx_b = 0;
+        break;
+      }
+    }
+  }
+}
+
+/// @brief Sort the cells in ascending order of the coordinate of the specified axis
+template <unsigned int AXIS, CellCollection T_CellCollection,
+          std::size_t NDim = 2, typename T_Index = unsigned int>
```
CarloVarni: why `typename T_Index = unsigned int` and not `unsigned int index_t`? shouldn't we impose it is an unsigned integer?

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+/// accelerate connection tests. Then associate the same label to connected
+/// cells where the label corresponds to the smallest cell index (in the sorted
+/// cell collection) of a cluster, where the cells are in ascending order of one
+/// of the coordinates Finally will sort the cells by the label in ascending
+/// order. As a consequence cells which belong to one cluster are adjacent and
+/// have the same label. T_Index must be large enough to fit the number of cells
+/// in the collection.
+template <unsigned int SORT_AXIS, CellCollection T_CellCollection,
+          typename T_Index = unsigned int, EFold FOLD = k8FoldConnection>
+inline void clusterize(T_CellCollection &cells) {
```
CarloVarni: I believe that templated functions are always `inline`?

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+}
+
+/// @brief call the given function for each cluster of a label sorted cell collection.
+/// It is assumed that @ref clusterize was called previously for the given
+/// cell collection i.e. the cells are labeled and sorted by label.
+/// Will call the given function for each cluster, where the function
+/// gets a reference to the full cell collection and the begin and end
+/// cell index which defines the cell range of a cluster.
+template <CellCollection T_CellCollection, typename T_Func>
+inline void for_each_cluster(T_CellCollection &cells, T_Func func) {
```
CarloVarni: can we upgrade `T_Func ` to a global reference `T_Func &&` ?

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
+// This file is part of the ACTS project.
```
CarloVarni, line 1: This implementation does not support time-based clusterization as well. Can it be extended to cover that case?

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,331 @@
...
+//
```
CarloVarni, line 2: why is this a different implementation of the clustering routine instead of a replacement? If you can demonstrate this provides the same results and it is faster I'd suggest to simply replace the previous version and mark the PR as a breaking one.

CarloVarni (changes requested):

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: I did not mean unsigned integer literally but it should be an unsigned integral type.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: see above.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: you are correct, inline is not necessary.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
CarloVarni: then is it possible to use`std::is_unsigned`  ?

CarloVarni: as well as `std::is_integral`?

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: okay. a bit misleading convention since "a_name_t" typically indicates a concrete type e.g. size_t not a template parameter. But sure if that is the convention.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: Not sure.

```
struct A { 
   A(A &&A) =  default; 
   ....
   operator()(...) const {...}
   std::vector< Payload>
 };
 void example(... cells_a, .... cells_b) {
    A functor;
    ....
    for_each_cluster(cells_a, functor);
    for_each_cluster(cells_b, functor);
 }
 ```
pro: with  `T_Func&& func`  no copy will be made, but with `T_Func func` a copy will be created.
contra: with `T_Func &&func` the payload could be moved away and the second call would not get it.

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 15 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp:1]: I haven't tried it, and I haven't even tried to clusterize more than 2 dimensions. But in theory one could add time as an extra coordinate and transform the time such that a time difference <=1 means connected. The types of all coordinates have to be the same and the type must be an integral type (though the "integral" requirement is likely not really  necessary, but the abort criterion would have to be relaxed to <=2, I suppose ) If the time cannot be transformed in such a way (I would not know why not)  one would have to specialize isConnected or isConnected4 and isConnected8, time could not be used as a sorting direction since a difference of >1 in the sorting axis would abort the search for connections.

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 16 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp:2]: If the current clusterization would produce the final clusters, than that might be advantageous. The final label sorting might add an unnecessary overhead in that case. So far I only encountered cases in which the cluster collection was temporary.

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 15 above)
Comment (CarloVarni) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp:1]: right now we are doing it with way: https://github.com/acts-project/acts/blob/main/Core/include/Acts/Clusterization/TimedClusterization.hpp  

do you think this is not adaptable to this implementation of yours?

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 1 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: ok

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 2 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: The cell is not necessarily comparable without a special comparison operator, I can add one of the requirements of sortable

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 3 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: ok

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 5 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: yes, std::unsigned_integral works.

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 15 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp:1]: to allow for different coordinate types (i.e. int for spacial coordinates and float for time)  would make the code more complicated. Since in the cases I have seen a temporary cell collection is created anyways, then one can also transform the time into something more suitable for the clusterization, I would say.

**Comment by goetzgaycken** [comment]
changes from [9aca2e0](https://github.com/acts-project/acts/commit/9aca2e0ddbe42b995a86ee686eafb5a28295b93a) to [d1fd488](https://github.com/acts-project/acts/commit/d1fd488f57fbe8a5aa52f292fe2055736c9f25fe):
- removed unnecessary inline keyword
- changed template parameter names from  T_Name  to name_t
- put clusterization into namespace Acts
- merged requirements  i.e. `requires(a){..} && requires(b) {...} -> requires(a,b) {...}`

**Comment by goetzgaycken** [comment]
changes [d1fd488](https://github.com/acts-project/acts/commit/d1fd488f57fbe8a5aa52f292fe2055736c9f25fe) to [b54ce39](https://github.com/acts-project/acts/commit/b54ce3917c985e252df51f27e76d8e472e2cce77):
- specify concept for index_t
- removed unnecessary template parameter axis_t

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 12 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: there is now a concept for  T_Index (now index_t)

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 2 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: added std::permutable

**Comment by goetzgaycken** [comment]
btw. the names isConnected4 and isConnected8 only make sense in the 2D case. In the examples framework "commonCorrner" is used instead. To distinguish the two one could use commonEdge commonEdgeAndCorner. But that would get rather verbose.

**Comment by goetzgaycken** [comment]
At the moment the criterion for aborting the search for connections is `diff[SORT_AXIS] > 1`, which assumes that the coordinates are integral. In theory it might be possible to use floats, but then this criterion would be too restrictive. So one possibility would be to replace this by a function like isConnected, then alternative implementations could be provided.

It is not clear to me whether there would be a use case for that (or does  a useful clusterization of location+time would really require none integral times? )

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 6 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: not clear whether that is a useful restriction. The requirements would be that coordinates can be subtracted and that the subtraction yields the same type, and that adjacent cells have differences of <=1 in one or all coordinates (if isConnected is not overloaded). Moreover the coordinates have to be comparable (<), and in the sorting axis a difference of >1  will abort the search. But in principle this should also work with floats (if the abort criterion would be adjusted to e.g. > 1.5 ).

**Comment by goetzgaycken** [comment]
changes [b54ce39](https://github.com/acts-project/acts/commit/b54ce3917c985e252df51f27e76d8e472e2cce77) to [11f46ad](https://github.com/acts-project/acts/commit/11f46ad242ec43218f64ce84a668e79dd22847c1):
- extended requirements for CellWithLabel:
   - coordinates can be compared,
   - label and dimension are of unsigned integral type.

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 7 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: ok

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 8 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: This is kind of how the functions are intended to be used. Is there a reason why it would be useful to drop the & ? 
I haven't tried but in theory the code should work without an implementation of cell_t, as long as the functions are declared.

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 9 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: coordinates_t is a rather concrete type, but to spell it out would get quite complicated:
std::array< decltype(getCoordinatel... ), getCellDimension() >

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 10 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: This would just make the code more verbose without adding extra information: (EFold::k8FoldConnection) instead of just k8FoldConnection which already says everything.

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 11 above)
Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: All elements of the array are set just below. There is no need to first set them to zero to set them to something else directly after.

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp (hunk 8 above)
Comment (CarloVarni) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: I believe the `&` is doing nothing here. The requirement is not really a function that is called in the standard sense. So `cell_t a` and `cell_t& a` are ~the same thing. 

What you are testing is that you can the different functions on a `cell_t` object. Or better, that the functions exist.

Comment (goetzgaycken) [reply Core/include/Acts/Clusterization/InPlaceClusterization.hpp]: indeed does not make a difference. But feels wrong to write `cell_t a` if a only a forward definition of the type cell_t is representing is available. Anyway I would not have a strong opinion either way.

**Comment by goetzgaycken** [comment]
changes [11f46ad](https://github.com/acts-project/acts/commit/11f46ad242ec43218f64ce84a668e79dd22847c1) to [efc6765](https://github.com/acts-project/acts/commit/efc676504eb6f790cb352a3ca4796c355ff1bcd8):
- fixed clang-tidy warnings, and silenced clang-tidy warnings about uninitialized values in 2 cases in which the variables are set just below.

**Comment by goetzgaycken** [comment]
changes [efc6765](https://github.com/acts-project/acts/commit/efc676504eb6f790cb352a3ca4796c355ff1bcd8) to [eb253b4](https://github.com/acts-project/acts/commit/eb253b407ed474743e7f88f807bb0c6f3acd1a78):
- second attempt to fix the remaining clang-tidy warning.

**Comment by goetzgaycken** [comment]
changes [eb253b4](https://github.com/acts-project/acts/commit/eb253b407ed474743e7f88f807bb0c6f3acd1a78) to [7b9aa72](https://github.com/acts-project/acts/commit/7b9aa721a6d272db5719c764611901059f225a18):
- rebase to main 2025-10-30_T1413

**Comment by goetzgaycken** [comment]
changes [7b9aa72](https://github.com/acts-project/acts/commit/7b9aa721a6d272db5719c764611901059f225a18) to [a2898e2](https://github.com/acts-project/acts/commit/a2898e2d410cbc48c311c2e2f3214209eef603aa):
- renamed connection types from fold -> ConnectionType, 8Fold -> CommonEdgeOrCorner, 4Fold -> ConnonEdge
- support unsigned coordinates i.e. when computing differences for unsigned types subtract smaller from larger value, the difference is now a template which can be specialized to support alternative coordinate types.
- moved abort criterion (diff[sort_axis]>1) to template which could be specialized. 
- improved comment concerning types of coordinates.

**Comment by goetzgaycken** [comment]
changes [a2898e2](https://github.com/acts-project/acts/commit/a2898e2d410cbc48c311c2e2f3214209eef603aa) to [94c8288](https://github.com/acts-project/acts/commit/94c82886d3403cc291d09270ed9781c9bccaa73a):
- forgot a couple of places to rename fold -> connect type, common corner ...

File: Core/include/Acts/Clusterization/InPlaceClusterization.hpp [hunk 17]
```diff
@@ -0,0 +1,371 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Utilities to clusterize a collection of cells on a regular grid
+// by sorting them in such a way that adjacent cells with the same
+// label belong to the same cluster, where a regular grid is a grid
+// on which cells have exactly 8 neighbors (except for edge cells),
+// and the indices of adjacent cells differ by one in the corresponding
+// direction.
+
+#pragma once
```
andiwand, line 16:
```suggestion
#pragma once

```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+auto getLabel(const cell_t &a) {
+  return a.m_label;
+}
+}  // namespace traits
+
+/// @brief A possible Cell type, that can be clustered
+/// coordinates_t are expected to be signed integers
+/// index_t should be an unsigned integer
+/// Unsigned should work, float coordinate types may cause problems. See
+/// getCellCoordinate.
```
andiwand:
```suggestion
/// @tparam coordinates_t are expected to be signed integers
/// @tparam index_t should be an unsigned integer
/// @note unsigned should work, float coordinate types may cause problems. See
/// getCellCoordinate.
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+template <typename coordinates_t, std::size_t N, std::unsigned_integral index_t>
+struct Cell {
+  Cell(const std::array<coordinates_t, N> &coordinates, index_t src_index)
+      : m_coordinates(coordinates), m_label(index_t{}), m_srcIndex(src_index) {}
+
+  std::array<coordinates_t, N>
+      m_coordinates;   ///< the coordinates of the cell on the regular grid
+  index_t m_label;     ///< a label which will be assigned by the clusterization
+  index_t m_srcIndex;  ///< the index to find the source cell
```
andiwand: for public struct data members the `m_` prefix should be dropped I think

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+template <typename coordinates_t>
+bool canAbortSearch(const coordinates_t &coordinates_diff,
+                    unsigned int sort_axis_i) {
+  assert(sort_axis_i < coordinates_diff.size());
+  return coordinates_diff[sort_axis_i] > 1;
+}
+
+/// @brief compute the absolute difference of two coordinates
+template <typename coordinate_t>
+auto absDifference(coordinate_t a, coordinate_t b) {
```
andiwand, line 198: everywhere else I saw `const coordinate_t &`, here it is passed by value. is there a reason for that?

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+  int connections = 0;
+  for (const auto &a_coordinate_diff : coordinates_diff) {
+    connections += a_coordinate_diff;
+  }
+  return connections <= 1;
+}
+
+/// @brief test whether cells are connected considering common edges or common corners
+enum class EConnectionType { CommonEdgeOrCorner, CommonEdge };
+template <EConnectionType CONNECTION_TYPE, typename coordinates_t>
```
andiwand: `CONNECTION_TYPE` looks kind of aggressive to me 😅

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+  }
+}
+
+/// @brief Label cells which are in ascending order of the coordinate of the given axis
+/// @param cells the sorted cell collection that should be labeled.
+template <unsigned int SORT_AXIS, CellCollection cell_collection_t,
+          std::unsigned_integral index_t = unsigned int,
+          EConnectionType CONNECTION_TYPE>
+void labelSortedCells(cell_collection_t &cells) {
+  using cell_t = std::remove_cvref_t<decltype(cells[0])>;
```
andiwand: since we have a concept for `CellCollection` I would argue that we can also require it providing a `Cell` type that can be directly accessed and used here

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+  static constexpr std::size_t NDim = traits::getCellDimension<cell_t>();
```
andiwand, line 215: same for the cell dimension - I think that can be required by the `CellCollection`

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+  static_assert(NDim > 0);
+  using label_t = std::remove_cvref_t<decltype(traits::getLabel(cells[0]))>;
+  using coordinate_t =
+      std::remove_cvref_t<decltype(traits::getCellCoordinate(cells[0], 0u))>;
+  static_assert(std::numeric_limits<index_t>::max() >=
+                std::numeric_limits<label_t>::max());
```
andiwand, line 221: also these look like they can be ensured / provided by the `CellCollection` concept

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+  // cells are sorted in the first coordinate
+  // thus can stop searching for adjacent cells
+  // if distance in the sorted coordinate is too large
+  for (index_t idx_a = 0; idx_a < cells.size(); ++idx_a) {
+    traits::setLabel(cells[idx_a], idx_a);
+    for (index_t idx_b = idx_a; idx_b-- > 0;) {
+      // NOLINTBEGIN(cppcoreguidelines-pro-type-member-init)
```
andiwand: is this really necessary? I would expect the optimizer to realize that the initialization is not necessary at the end

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+      // all elements of diff will be set to a value no need to initialize
+      // with a dummy value.
+      std::array<coordinate_t, NDim> diff;
```
andiwand:
```suggestion
      std::array<coordinate_t, NDim> diff{};
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+      if (isConnected<CONNECTION_TYPE>(diff)) {
+        if (cells[idx_a].m_label < idx_a) {
+          // if the label is not its index, the cell was merged to a cluster
+          // thus the two clusters need to be merged rather than just merging
+          // in this one cell
+          // NOLINTBEGIN(cppcoreguidelines-init-variables)
+          // min_label and max_label will be set before use, no reason to set
+          // to unused dummy value before.
+          unsigned int min_label;
+          unsigned int max_label;
```
andiwand:
```suggestion
          unsigned int min_label{};
          unsigned int max_label{};
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+    }
+  }
+}
+
+/// @brief Sort the cells in ascending order of the coordinate of the specified axis
+template <unsigned int AXIS, CellCollection cell_collection_t,
+          std::size_t NDim = 2, std::unsigned_integral index_t = unsigned int>
+  requires(NDim > 0 && AXIS < NDim)
+void sortCellsByCoordinate(cell_collection_t &cells) {
+  using cell_t = std::remove_cvref_t<decltype(cells[0])>;
```
andiwand: same here

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+  std::sort(cells.begin(), cells.end(), [](const cell_t &a, const cell_t &b) {
+    return traits::getCellCoordinate(a, AXIS) <
+           traits::getCellCoordinate(b, AXIS);
+  });
+}
+
+/// @brief Sort the cells in ascending order of the asscoiated label.
+template <CellCollection cell_collection_t>
+void groupCellsByLabel(cell_collection_t &cells) {
+  using cell_t = std::remove_cvref_t<decltype(cells[0])>;
```
andiwand: same here

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+/// @brief determine the number of clusters.
+/// It is assumed that @ref clusterize was called previously for the given
+/// cell collection i.e. the cells are labeled and sorted by label.
+/// It will count the number of different labels do determine the number of
+/// clusters in the collection.
+template <CellCollection cell_collection_t,
+          std::unsigned_integral index_t = unsigned int>
+std::size_t countLabels(const cell_collection_t &cells) {
+  assert(cells.size() <= std::numeric_limits<index_t>::max());
+  std::size_t nlabels = !cells.empty();
```
andiwand:
```suggestion
  if (cells.empty()) {
    return 0;
  }
  std::size_t nlabels = 1;
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,371 @@
...
+
+/// @brief call the given function for each cluster of a label sorted cell collection.
+/// It is assumed that @ref clusterize was called previously for the given
+/// cell collection i.e. the cells are labeled and sorted by label.
+/// Will call the given function for each cluster, where the function
+/// gets a reference to the full cell collection and the begin and end
+/// cell index which defines the cell range of a cluster.
+template <CellCollection cell_collection_t, typename func_t>
+void for_each_cluster(cell_collection_t &cells, func_t func) {
+  if (!cells.empty()) {
```
andiwand:
```suggestion
  if (cells.empty()) {
    return;
  }
  // ...
```
personal preference - reduces indentation, early exit for the uncommon case

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: elsewhere it is possibly multiple coordinates (i.e. `array< coordinate_t, N>` here it is just one coordinate.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: Not that I think this name is perfect, but I cannot think of a better name.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: these are point-less initializations. The values are overwritten in the lines below. It is likely optimized away, but shouldn't the static code analyzer understand that there is no problem here rather than the variables to be set to some useless random values (i.e. the initialization with zero only satisfies  clang-tidy but would not be a meaningful default value).

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: I do not see the advantage.

andiwand: I do.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: ok (but makes searching more complicated).

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: thanks!

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: ok

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: Not a strong argument in this case, but for some reason one instruction less with g++.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: would speed up compilation, containers would not provide this by default, though (i.e. could not  just use `vector < Cell >` ).

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: The idea is that the kind of always true assumption is verified close to where it is assumed. Also if this assert is false, it is not a problem of the cell collection, but the problem is rather that the template parameters here are not matching. One could argue that the index should just always be size_t and then all this could be removed. But there might be reasons to chose something else (e.g. sorting should be faster the smaller the cell collection is, but the label type (or size_t as return value of  getLabel and argument to setLabel) is not necessarily the ideal type for what concerns the CPU).

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: but what is the point of an initialization which only satisfies the static code analyser but otherwise is as good as a random value ? This removes the possibility for a better analyser to spot errors.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: see above

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: Sure one could. Would speed up compilation, std container would provide that, custom container not necessarily.

goetzgaycken: The original CellCpntainer indirectly required that value_type exists, so it can be used instead.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: ok

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: ok

goetzgaycken:
changes [d5bf037](https://github.com/acts-project/acts/commit/d5bf03732b687b38b705f6f71ba72b3b178b30e6) to [ea99ee1](https://github.com/acts-project/acts/commit/ea99ee1c2b1836b257055f271d763df1076b6a89) and [ea99ee1](https://github.com/acts-project/acts/commit/ea99ee1c2b1836b257055f271d763df1076b6a89) to [72b2838](https://github.com/acts-project/acts/commit/72b28386e3bd82f9d482e16c2516034bb03f17ed):
- removed member prefix from `m_` from public members of the reference Cell implementation. 
- exit  early and reduced indention in for_each_cluster
- improved comments for Cell,
- empty line after pragma once
- make use of CellCollecion::value_type  which was required indirectly anyway, now more explicitly.

I did not:
- extend to CellCollection concept  to require the container to provide the dimension of the grid,
- make a static_assert a requirement of the CellCollection since it is only relevant for the function and is more a sanity check.
- satisfy clang-tidy (i.e. I did not initialize variables to some random value just before they are actually set to a meaningful one just to avoid clang-tidy warnings)

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
andiwand: I think it should be either `CamelCase`, like `ConnectionType`, or `snake_case_t`, like `connection_type`

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
andiwand: If we want to support `std::vector`, sure. I suppose it could still be a `CellCollection` trait to access this container constant information.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
andiwand: I don't see any benefit in disabling the linter here tbh. IMO a random value is always worse than `0`, even if we do not rely on it right now. The static analysis is right to flag this. Everywhere is ACTS we use initialized values because it is a point to debug.

goetzgaycken:
change [72b2838](https://github.com/acts-project/acts/commit/72b28386e3bd82f9d482e16c2516034bb03f17ed) to [0c631b5](https://github.com/acts-project/acts/commit/0c631b51005233d8f95fdcc9ed6bfa0098f0635d):
- Like original clusterization there is now an object for the connection testing, which by default has no data  members and only static methods, but could be replaced by something which can provide parameters to the connection test. In the clusterization function the object type replaces the connection type template parameter
- Extended the 2D clusterization unit test by an "example" which clusterizes cells with 2D spatial coordinates and time, where also time is given by a 16 bit integer.

I thought a little bit more about the customization of the connection testing. I suppose some customization will be necessary if the clusterization should take into account spatial and time cell components. I originally wanted to get rid of the connection object  template parameter (current clusterization),  but since I already introduced  a template parameter to steer the connection test (enum), it could also just be turned into an arbitrary type which  is more or less the current design. The overhead is minimal (or zero) if the  "connection" object has no data members and only static methods  which is the case for the default: "ConnectionHelper" (Maybe ConnectionTester would be a better name? ). This helper provides  the original functions `isConnected` and `canAbortSearch`.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: Not that it really matters but it is more lines and more code, more branches.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: Okay fine, I can make the constant more indistinguishable from everything else, if that is what is preferred.

goetzgaycken:
changes [0c631b5](https://github.com/acts-project/acts/commit/0c631b51005233d8f95fdcc9ed6bfa0098f0635d) to [8033eb6](https://github.com/acts-project/acts/commit/8033eb6630bda0af2e7032c9b409d611b97324d5):
- obfuscated an aggressive constant.

goetzgaycken:
changes [033eb6](https://github.com/acts-project/acts/commit/8033eb6630bda0af2e7032c9b409d611b97324d5) to [077f30a](https://github.com/acts-project/acts/commit/077f30a3ba0b1a7949deda452bacaf65578b8346):
- rebase to 2025-11-06_T1142
- solved merge conflicts in ClusterizationTests2D, and 
- adjusted additions to ClusterizationTests2D for duplicate cell exception.

goetzgaycken:
changes [077f30a](https://github.com/acts-project/acts/commit/077f30a3ba0b1a7949deda452bacaf65578b8346) to [cbd9c04](https://github.com/acts-project/acts/commit/cbd9c043a3cb7ab26724a63d79c632a703d6c517):
- clang-tidy and doc fixes

goetzgaycken:
[cbd9c04](https://github.com/acts-project/acts/commit/cbd9c043a3cb7ab26724a63d79c632a703d6c517) to [095167c](https://github.com/acts-project/acts/commit/095167c4e95aa7993594c1e6f84347fc0eb1cec5):
- more line and doc fixes.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: but then `traits::getCellDimension<cell_t>();` just becomes something like `traits::getCellDimension<cell_collection_t>();`, isn't this pretty much the same ?

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: A good static analyzer would understand whether variables which are only used in a local scope are possibly used before being initialized. If all variables are set to a meaningless default value to satisfy not very good analysers than better analysers cannot find the real errors e.g. that there is a hole in the logic and  variables could not be initialized  in all cases.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
andiwand: I don't think this is readable. It took me 100x longer to read your version. Apart from that the code below will need to be safe against `size() == 0`. If you are worried about the number of branching the performance should be studied. To me this argument does not hold if you don't provide any measurement. In the spirit of maintainability I suggest to change the code to a more readable version.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
andiwand: To you it is pointless - ok, but to others, including me, it is not. A good optimizer will make this go away, a good static analyzer would understand that there is no problem. We only have one of both and we should work with what we have and don't turn off the linter for such unimportant cases. If you have better default values you can use those of course.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: I would say the "complicated" part is why does the loop not start at zero, which would not be explained  by exiting early.

Concerning the safety against zero. There are three cases to consider size=0, 1, >1
for 0 and 1 the loop will not be executed because 1 is not < size. and the return value will be equal to size for these cases. Not sure what is so complicated about this but I can add a comment.

In terms of benchmarks, it is a bit difficult to measure. On x86_64, it seems that Your proposal has in most cases a small edge but for some size it is the opposite, on arm (lxplus-arm) my version is significantly faster but in particular for size >=4 where a difference is only really expected for 0 and 1, for >1 there should be just a constant offset. But modern cpus do all kind of weird things .... 
[bench.pdf](https://github.com/user-attachments/files/23476617/bench.pdf)

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: In the  following example clang++ --analyze will complain
```
unsigned int absDiff(int a, int b) {
   int x0;
   int x1;
   // NOLINTBEGIN(cppcoreguidelines-init-variables)
  if ( a<b) {
     x0=a;
     x1=b;
  }
  else {
     x0=b;
  }
  // NOLINTEND(cppcoreguidelines-init-variables)
  return x1-x0;
}
```
If I change it as You say to satisfy clang-tidy:
```
unsigned int absDiff(int a, int b) {
   int x0{};
   int x1{};
  if ( a<b) {
     x0=a;
     x1=b;
  }
  else {
     x0=b;
  }
  return x1-x0;
}
```
Neither of the tools (clang-tidy, clang++ --analyze) will complain. So, is better to satisfy bad tools, or to not add pointless/incorrect initializations. For me it is clearly the latter because in the latter case, a good tool actually helps me to find the bug, where in the first case the bad tool just helps me to hide the mistake under the rug.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
paulgessinger: I would argue until it's proven to be a performance issue, we should go with the least surprising code. Prefer

```cpp
std::size_t nlabels = cells.empty() ? 0 : 1;
```

over 

```cpp
std::size_t nlabels = !cells.empty();
```

for clarity, please.

andiwand:
Sorry but I don't think this discussion is worth the time and effort given what this is about. I asked for very simple adjustments to the code to apply with out guidelines of readability and maintainability. I cannot approve this as. You can of course ask someone else for a review / approval.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
paulgessinger: I strongly think we should avoid toggling the linter even if we're convinced it's safe **until** we have reason to believe there is some performance to be gained.

I would ask you therefore to go with the (potentially unnecessary) initialization for now, and circle back later if we find this can improve things.

goetzgaycken: One possibility, I see, to write the code without bogus initialization and without triggering bogus clang-tidy warnings, would be to use a helper lambda e.g.
```
   auto [min_label,max_label] = [&]() {
      if (a < b) {
         ....
         return std::make_pair(a,b);
      } else {
        ....
         return std::make_pair(b,a);
      }
   }();
```
But that makes things much more complicated than necessary.

goetzgaycken:
changes [ddb6bc2](https://github.com/acts-project/acts/pull/4735/commits/ddb6bc24fb83398d3236975eed0c8fef68394c1c):
- as the comments says, works around the false clang-tidy positive without switching off the warning as requested. Not sure it is an improvement though.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: The point is not so much that it is unnecessary, but that the default value is not necessarily correct.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: ok.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: Another simple example. This is wrong
```
#include <array>
int main() {
   std::array<int,2> a;
   for (unsigned int i=0; i<1; ++i) {
      a[i]=i+1;
   }
   int sum=0;
   for (unsigned int i=0; i<a.size(); ++i) {
      sum += a[i];
   }
   return sum==3 ? 0 : 1;
}
```
but it is not more correct when changing     `std::array<int,2> a;` to `std::array<int,2> a{};`. 

I thought that I encountered a static analyzer which would detect the problem but the ones I tried won't. However valgrind and even clang++ with `-fsanitize=memory` will find the problem, but obviously not if `a` gets initialized. Anyway, initializing variables just for the sake of making clang-tidy happy, makes it harder for better tools to find the actual error. Therefore, I don't think it is an improvement, it just hides a potential problem.

goetzgaycken:
changes [74723d2](https://github.com/acts-project/acts/commit/74723d23a74c3b67909368efb47587cfe63113a5) to [8469555](https://github.com/acts-project/acts/commit/8469555bcdca42e47fec82ddb1fde72026902e9d):
- bowing to the machine. Reverted the latest change because the solution which solved the clang-tidy complaint about uninitialized variables, by actually initializing the variables properly, added too much complexity.  Moreover, instead of switching off the  clang-tidy warning caused by its limited unstanding,  clang-tidy is now satisfied by bogus initializations. Likely compilers are smarter than clang-tidy and understand that these intializations do not serve any purpose and optimize them away, but at the same time smarter tools are taken the possibility to actually find real initialization problems, since there are now the bogus initializations. Anyway clang-tidy is happy, and it seems to be the preferred solution even so the solution has no other purpose than satisfying clang-tidy and it makes harder to find problems with better tools.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: Clusterization with 3 dimensions works, but as mentioned above the types must be the same. Either the raw time could be used which is likely an integer, or  floats could be used also for the spatial coordinates, which may require special implementations of the connection test.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: added bogus initializations.

goetzgaycken:
changes: [8469555](https://github.com/acts-project/acts/commit/8469555bcdca42e47fec82ddb1fde72026902e9d) to [6caffb9](https://github.com/acts-project/acts/commit/6caffb9fb63198def10a99e30d15530959d43690):
- rebased to main of 2026-05-12_T1216, resolved merge conflicts 
- improved some in code comments
- updatedClusterizationTests2D  to hashMixAndCombine

goetzgaycken:
changes [6caffb9](https://github.com/acts-project/acts/commit/6caffb9fb63198def10a99e30d15530959d43690) to [8128d15](https://github.com/acts-project/acts/commit/8128d15cef3e3024ef0b5b904860ff1801b4f041):
- added documentation for items marked as missing
- removed ref tag for referenced functions which did not get resolved.

goetzgaycken:
changes [8128d15](https://github.com/acts-project/acts/commit/8128d15cef3e3024ef0b5b904860ff1801b4f041) to [12ba825](https://github.com/acts-project/acts/commit/12ba82504d4f1b8d900037b33142a83d416a3918):
- forgot to document return value.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
goetzgaycken: Anyway the clusterization is not backward compatible, so that would be something for later.

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,412 @@
...
+}
+
+/// @brief Label cells which are in ascending order of the coordinate of the given axis
+/// @param cells the sorted cell collection that should be labeled.
+/// @param connection_helper a helper object to test for cell connections.
+template <unsigned int SORT_AXIS, CellCollection cell_collection_t,
+          std::unsigned_integral index_t = unsigned int,
+          typename connection_helper_t>
+void labelSortedCells(cell_collection_t &cells,
+                      connection_helper_t &&connection_helper) {
```
andiwand, line 213:
```suggestion
                      const connection_helper_t &connectionHelper) {
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,412 @@
...
+  /// @brief construct cell with its coordinates and index of the origin
+  /// @param the_coordinates coordinates of the cell
+  /// @param src_index index which identifies the source of the cell data
+  Cell(const std::array<coordinates_t, N> &the_coordinates, index_t src_index)
+      : coordinates(the_coordinates), label(index_t{}), srcIndex(src_index) {}
+
+  std::array<coordinates_t, N>
+      coordinates;   ///< the coordinates of the cell on the regular grid
+  index_t label;     ///< a label which will be assigned by the clusterization
+  index_t srcIndex;  ///< the index to find the source cell
```
andiwand, line 83:
```suggestion
  /// the coordinates of the cell on the regular grid
  std::array<coordinates_t, N> coordinates{};
  /// a label which will be assigned by the clusterization
  index_t label{};
  /// the index to find the source cell
  index_t srcIndex{};
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,412 @@
...
+/// @tparam index_t should be an unsigned integer
+/// @note Unsigned coordinates should work, float coordinate types may cause problems. See
+/// getCellCoordinate.
+template <typename coordinates_t, std::size_t N, std::unsigned_integral index_t>
+struct Cell {
+  /// @brief construct cell with its coordinates and index of the origin
+  /// @param the_coordinates coordinates of the cell
+  /// @param src_index index which identifies the source of the cell data
+  Cell(const std::array<coordinates_t, N> &the_coordinates, index_t src_index)
+      : coordinates(the_coordinates), label(index_t{}), srcIndex(src_index) {}
```
andiwand, line 78:
```suggestion
  Cell(const std::array<coordinates_t, N> &coordinates_, index_t srcIndex_)
      : coordinates(coordinates_), srcIndex(srcIndex_) {}
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,412 @@
...
+/// @brief Sort the cells in ascending order of the coordinate of the specified axis
+template <unsigned int AXIS, CellCollection cell_collection_t,
+          std::size_t NDim = 2, std::unsigned_integral index_t = unsigned int>
+  requires(NDim > 0 && AXIS < NDim)
+void sortCellsByCoordinate(cell_collection_t &cells) {
+  using cell_t = typename cell_collection_t::value_type;
+  std::sort(cells.begin(), cells.end(), [](const cell_t &a, const cell_t &b) {
+    return traits::getCellCoordinate(a, AXIS) <
+           traits::getCellCoordinate(b, AXIS);
+  });
```
andiwand, line 283:
```suggestion
  std::ranges::sort(cells, [](const cell_t &c) {
    return traits::getCellCoordinate(c, AXIS);
  });
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,412 @@
...
+
+/// @brief Sort the cells in ascending order of the asscoiated label.
+template <CellCollection cell_collection_t>
+void groupCellsByLabel(cell_collection_t &cells) {
+  using cell_t = typename cell_collection_t::value_type;
+  // stable sort to retain coordinate ordering
+  std::stable_sort(cells.begin(), cells.end(),
+                   [](const cell_t &a, const cell_t &b) {
+                     return traits::getLabel(a) < traits::getLabel(b);
+                   });
```
andiwand, line 294:
```suggestion
  std::ranges::stable_sort(cells,
                   [](const cell_t &c) {
                     return traits::getLabel(c);
                   });
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,412 @@
...
+/// order. As a consequence cells which belong to one cluster are adjacent and
+/// have the same label. index_t must be large enough to fit the number of cells
+/// in the collection.
+template <
+    unsigned int SORT_AXIS, std::unsigned_integral index_t = unsigned int,
+    CellCollection cell_collection_t = std::span<Cell<int, 2, unsigned int> >,
+    typename connection_helper_t =
+        ConnectionHelper<Cell<int, 2, unsigned int> > >
+void clusterize(cell_collection_t &cells,
+                connection_helper_t &&connection_helper =
```
andiwand, line 314:
```suggestion
                const connection_helper_t &connectionHelper =
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,412 @@
...
+}
+
+/// @brief call the given function for each cluster of a label sorted cell collection.
+/// It is assumed that clusterize was called previously for the given
+/// cell collection i.e. the cells are labeled and sorted by label.
+/// Will call the given function for each cluster, where the function
+/// gets a reference to the full cell collection and the begin and end
+/// cell index which defines the cell range of a cluster.
+template <CellCollection cell_collection_t, typename func_t>
+void for_each_cluster(cell_collection_t &cells, func_t func) {
```
andiwand, line 348:
```suggestion
void forEachCluster(cell_collection_t &cells, func_t func) {
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,412 @@
...
+/// Will first sort the cells of each cluster by the coordinate of the specified
+/// axis and add one cell range per cluster to the given range container. The
+/// value type of the range container has to be constructible by two indices the
+/// index of the first cell in a cluster and the index of the cell after the
+/// last cell of the cluster.
+template <unsigned int AXIS, CellCollection cell_collection_t,
+          typename range_collection_t>
+void addCellRangesAndSort(cell_collection_t &cells,
+                          range_collection_t &ranges) {
+  for_each_cluster(cells, [&ranges](cell_collection_t &all_cells,
```
andiwand, line 398:
```suggestion
  for_each_cluster(cells, [&ranges](cell_collection_t &allCells,
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,412 @@
...
+                                    unsigned int idx_begin,
+                                    unsigned int idx_end) {
+    auto cluster_range =
```
andiwand, line 401:
```suggestion
    auto clusterRange =
```

Core/include/Acts/Clusterization/InPlaceClusterization.hpp
```diff
@@ -0,0 +1,412 @@
...
+/// value type of the range container has to be constructible by two indices the
+/// index of the first cell in a cluster and the index of the cell after the
+/// last cell of the cluster.
+template <unsigned int AXIS, CellCollection cell_collection_t,
+          typename range_collection_t>
+void addCellRangesAndSort(cell_collection_t &cells,
+                          range_collection_t &ranges) {
+  for_each_cluster(cells, [&ranges](cell_collection_t &all_cells,
+                                    unsigned int idx_begin,
+                                    unsigned int idx_end) {
```
andiwand, line 400:
```suggestion
                                    unsigned int idxBegin,
                                    unsigned int idxEnd) {
```

Examples/Algorithms/Digitization/src/ModuleClusters.cpp
```diff
@@ -109,6 +117,149 @@ std::vector<ModuleValue> ModuleClusters::createCellCollection() {
...
+  // @param idx the position in the range where 0 refers to the first element in the range
+  const ModuleValue& operator[](unsigned int idx) const {
+    assert(idx < m_cells.size() && m_moduleValues &&
+           m_cells[idx].srcIndex < m_moduleValues->size());
+    return (*m_moduleValues)[m_cells[idx]];
+  }
+
+  std::vector<ModuleValue>* m_moduleValues;
+  std::span<cell_t> m_cells;  // A range of cells which refer to a ModuleValue
+                              // in the ModuleValue collection
```
andiwand, line 191:
```suggestion
  std::vector<ModuleValue>* m_moduleValues{};
  /// A range of cells which refer to a ModuleValue in the ModuleValue collection
  std::span<cell_t> m_cells;
```

Examples/Algorithms/Digitization/src/ModuleClusters.cpp
```diff
@@ -109,6 +117,149 @@ std::vector<ModuleValue> ModuleClusters::createCellCollection() {
...
+      assert(m_moduleValues == other.m_moduleValues);
+      return other.m_iter == m_iter;
+    }
+    iterator& operator++() {
+      ++m_iter;
+      return *this;
+    }
+
+    std::vector<ModuleValue>* m_moduleValues;
+    std::span<cell_t>::iterator m_iter;
```
andiwand, line 153:
```suggestion
    std::vector<ModuleValue>* m_moduleValues{};
    std::span<cell_t>::iterator m_iter;
```

Examples/Algorithms/Digitization/src/ModuleClusters.cpp
```diff
@@ -109,6 +117,149 @@ std::vector<ModuleValue> ModuleClusters::createCellCollection() {
...
+  }
+
+  std::vector<ModuleValue> newVals;
+  newVals.reserve(m_moduleValues.size());
+  // Helper to create a cluster module value for the given cell range
+  // the ModuleValues the cell range refers to will be squashed and then
+  // added to a new Cluster.
+  auto addClusters = [this, &newVals](std::vector<Cell>& all_cells,
+                                      unsigned int idx_begin,
+                                      unsigned int idx_end) {
```
andiwand, line 239:
```suggestion
  auto addClusters = [this, &newVals](std::vector<Cell>& allCells,
                                      unsigned int idxBegin,
                                      unsigned int idxEnd) {
```

Tests/UnitTests/Core/Clusterization/ClusterizationTestData.hpp
```diff
@@ -0,0 +1,217 @@
...
+// Data to test the clusterization algorithms.
+// Contains the coordinates of cells of a cluster where cells are connected if
+// they share an edge. Clusters are separated by 32767,32767, and the end of the
+// cluster collection is given by an empty cell range i.e.
+// 32767,32767,32767,32767. Data was obtained using Pythia ttbar PU200, Geant4,
+// and the ODD detector.
+
+#pragma once
+
+#include <cstdint>
```
andiwand, line 18:
```suggestion
#include <cstdint>

```

Tests/UnitTests/Core/Clusterization/ClusterizationTestDataCommonCorner.hpp
```diff
@@ -0,0 +1,276 @@
...
+// Contains the coordinates of cells of a cluster where cells are connected if
+// they share an edge or corner. Clusters are separated by 32767,32767, and the
+// end of the cluster collection is given by an empty cell range i.e.
+// 32767,32767,32767,32767. Data was obtained using Pythia ttbar PU200, Geant4,
+// and the ODD detector. The first data set corresponds to cluster_test_data but
+// assumes a different connection criterion i.e. common edge or corner.
+
+#pragma once
+
+#include <cstdint>
```
andiwand, line 19:
```suggestion
#include <cstdint>

```

Tests/UnitTests/Core/Clusterization/ClusterizationTests2D.cpp
```diff
@@ -5,10 +5,10 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
-
 #include <boost/test/unit_test.hpp>
```
andiwand, line 8:
```suggestion

#include <boost/test/unit_test.hpp>
```

andiwand (commented):
Do we want to add this directly to `Acts` or rather to `Acts::Experimental` @goetzgaycken ?

paulgessinger:
I this still relevant?

PR #4737, opened 2025-10-25
fix: Seed parameter estimation divides by zero

paulgessinger:
--- END COMMIT MESSAGE ---

### DIFF
--- a/Examples/Python/python/acts/examples/reconstruction.py
+++ b/Examples/Python/python/acts/examples/reconstruction.py
@@ -527,7 +527,7 @@ def addSeeding(
         s.addAlgorithm(
             acts.examples.SeedsToPrototracks(
                 level=logLevel,
-                inputSeeds=seeds,
+                inputSeeds="estimatedseeds",
                 outputProtoTracks=prototracks,
             )
         )


Tests/UnitTests/Core/Seeding/EstimateTrackParamsFromSeedTest.cpp
```diff
@@ -198,6 +198,17 @@ BOOST_AUTO_TEST_CASE(trackparameters_estimation_test) {
   }
 }
 
+BOOST_AUTO_TEST_CASE(trackparm_estimate_aligined) {
+  Vector3 sp0{-72.775, -0.325, -615.6};
+  Vector3 sp1{-84.325, -0.325, -715.6};
+  Vector3 sp2{-98.175, -0.325, -835.6};
```
paulgessinger, line 204: These values were empirically encountered. You can argue that the faulty input is the problem here, but I think the parameter estimation should fail gracefully, allowinig to filter this seed.

Core/src/Seeding/EstimateTrackParamsFromSeed.cpp
```diff
@@ -39,6 +39,11 @@ Acts::FreeVector Acts::estimateTrackParamsFromSeed(const Vector3& sp0,
   Vector3 local1 = transform.inverse() * sp1;
   Vector3 local2 = transform.inverse() * sp2;
 
+  if (local2(1) == 0) {
+    return Result<FreeVector>::failure(
+        SeedingError::InvalidSpacePointsForEstimate);
+  }
```
andiwand, line 45: I think the main problem here is that we use `y = x/a + b` which is not defined for `a=0`. I wonder if we could rather change the computation bellow to use a vector representation of the line instead of slope+intercept

Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp
```diff
@@ -91,10 +92,12 @@ FreeVector estimateTrackParamsFromSeed(spacepoint_range_t spRange,
     spTime = sp->t();
   }
 
-  FreeVector params = estimateTrackParamsFromSeed(
-      spPositions[0], spPositions[1], spPositions[2], bField);
-  params[eFreeTime] = spTimes[0].value_or(0);
-  return params;
+  return estimateTrackParamsFromSeed(spPositions[0], spPositions[1],
+                                     spPositions[2], bField)
+      .transform([&](FreeVector&& params) {
```
andiwand, line 97:
```suggestion
      .transform([&](const FreeVector& params) {
```
I guess this will all be inlined anyways but `FreeVector &&` could result in a move which will actually result in a copy

Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp
```diff
@@ -44,9 +44,10 @@ namespace Acts {
 /// @param bField is the magnetic field vector
 ///
 /// @return the free parameters
-FreeVector estimateTrackParamsFromSeed(const Vector3& sp0, const Vector3& sp1,
-                                       const Vector3& sp2,
-                                       const Vector3& bField);
+Result<FreeVector> estimateTrackParamsFromSeed(const Vector3& sp0,
```
andiwand, line 47: I guess in the general case this can fail because it is slightly overdetermined. But I don't see that we reached this point yet

Core/src/Seeding/EstimateTrackParamsFromSeed.cpp
```diff
@@ -39,6 +39,11 @@ Acts::FreeVector Acts::estimateTrackParamsFromSeed(const Vector3& sp0,
   Vector3 local1 = transform.inverse() * sp1;
   Vector3 local2 = transform.inverse() * sp2;
 
+  if (local2(1) == 0) {
```
andiwand, line 42: checking for `== 0` will not catch cases that are close to `0` but still fail

paulgessinger: The only failure cases I've seen were when this was exactly zero.

Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp
paulgessinger: That's correct. I was assuming this would be inlined into the return value.

Core/src/Seeding/EstimateTrackParamsFromSeed.cpp
paulgessinger: The way I interpreted it is that we can't form a circle of the points are on a line in the transverse plane.

paulgessinger:
Closed in favor of https://github.com/acts-project/acts/pull/4740

PR #4745, opened 2025-10-30
fix: Handle cell duplication with an error

### LINKED ISSUE #4736: Clusterization ignores some connections
Under certain circumstances the clusterization ignores connects, for 4 and 8 fold connectivity. Thus, produces slightly more clusters than necessary. For example In a ttbar, PU200, ODD, Geant4 event,  the clusterization splits 16 clusters into a total of 33. The other >130k  clusters seem fine.

DJDuque:
This makes clusterization (both 1D and 2D) throw an exception when duplicate cells are given as input.

Additionally, the `Connect2d::operator()` is a bit smarter in returning `ConnectResult::eNoConnStop` as quickly as possible (it previously waited unnecessarily until going past a full column). This could provide some performance improvement in cases where your cells are heavily "column-like".

Fixes #4736

--- END COMMIT MESSAGE ---

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -408,7 +408,7 @@ jobs:
         continue-on-error: true
         with:
           image: "ghcr.io/eic/eic_ci_without_acts:nightly"
-          options: -v ${{ github.workspace }}:/work
+          options: -v ${{ github.workspace }}:/work -v ${{ env.CCACHE_DIR }}:/root/.cache/ccache
           shell: bash
           run: |
             for script in /.singularity.d/env/*.sh; do
@@ -420,6 +420,7 @@ jobs:
             set -ue
 
             pushd acts
+            ccache -p | grep cache_dir
             ccache -z && \
             export CMAKE_INSTALL_PREFIX=$HOME/prefix
             export CMAKE_CXX_COMPILER_LAUNCHER=ccache


andiwand (approved):
Thank you for the fix @DJDuque !

An alternative to throwing an exception would be to use `Acts::Result<void>` over the existing `void` return type of `labelClusters` and `createClusters`.

I would argue that this API compatible since it would still compile and the expected standard case should still work. If there are duplicate cells the clustering would silently fail - that is not nice. In this case the exception would be better I guess.

Maybe we can keep it as it is and still treat it as non-breaking.

@goetzgaycken maybe wants to take a look after discovering the bug.
also tagging @CarloVarni and @paulgessinger

DJDuque:
I agree this most likely doesn't need a new major version, but I still wanted to tag the commit as a breaking change because it technically is, and it makes it easier for someone scanning the history to find things that could break their code. For example, adding new variants to an enum would break the Rust bindings I am playing around with [here](https://github.com/DJDuque/acts-rs).

I don't have a strong opinion about `Result` vs throwing here. In this particular case, just throwing feels more natural to me, but I can change it to a `Result<void>` if people think that is better.

Note that [this test](https://github.com/acts-project/acts/blob/main/Tests/UnitTests/Examples/Algorithms/Digitization/ModuleClustersTests.cpp#L71-L89) is now failing. But I am not familiar at all with the `Examples`/`Digitization`/`Utilities` stuff that this involves, so it would be good if someone else could have a look at it and why it depended on clusterization working with duplicate cells.

andiwand:
Yeah personally I am happy with the exception.

Breaking / non breaking is always a bit of a grey area for these kinds of changes. I would go with non breaking given that we currently look at C++ clients and that the standard behavior stays unchanged.

> Note that [this test](https://github.com/acts-project/acts/blob/main/Tests/UnitTests/Examples/Algorithms/Digitization/ModuleClustersTests.cpp#L71-L89) is now failing. But I am not familiar at all with the `Examples`/`Digitization`/`Utilities` stuff that this involves, so it would be good if someone else could have a look at it and why it depended on clusterization working with duplicate cells.

I think what needs to be changed there is that an exception is now expected when merging is disabled because this is now checked by your code changes.

goetzgaycken:
The examples framework uses the clusterization to merge SimHits. In this case the inputs to the clusterization may contain duplicate cells. This change would break running e.g. the DigitizationAlgorithm on a ODD geant4 ttbar input with merging enabled.

DJDuque:
So, [this](https://github.com/acts-project/acts/blob/main/Tests/UnitTests/Examples/Algorithms/Digitization/ModuleClustersTests.cpp#L74-L75) is the one that now throws.

If I understood correctly what @goetzgaycken said, this is because when `merge=true`, the digitization algorithm is internally using the clusterization implementation to "merge duplicates" (?). In any case, then this probably means that whenever `merge=true`, the current results are silently wrong (?).

I think that the options then are:
1. Change the test that is currently failing to:
    ```
    {
        BOOST_CHECK_THROW(
            testDigitizedParametersWithTwoClusters(true, {0, 0}, {0, 0}),
            std::invalid_argument);

        auto result = testDigitizedParametersWithTwoClusters(false, {0, 0}, {0, 0});
        BOOST_CHECK_EQUAL(result.size(), 2);
    }
    ```
    But I guess this will make the `merge=true` useless because whatever input this is expecting usually has duplicates for some reason (?).
2. Someone who uses (and is familiar) with the examples framework and this digitization algorithm makes a separate PR that makes `merge=true` work without using the Core clusterization implementation. Then this PR (in its current state) will be blocked until that happens.
3. We reject this PR, and we modify the Core clusterization to handle arbitrary number of duplicates.

DJDuque:
>     3. We reject this PR, and we modify the Core clusterization to handle arbitrary number of duplicates.

If we decide to move forward with this PR, I just realized that we probably want to make a small change in the `TimedClusterization` [here](https://github.com/acts-project/acts/blob/main/Core/include/Acts/Clusterization/TimedClusterization.ipp#L28-L34) to test that even if a cell is at the same row and column, we don't consider it a duplicate unless the time of both is also within the `timeTolerance`.

But I'll hold making that change (and unit test) until there is a decision on if we want to move forward with this PR or not.

CarloVarni:
how is this PR breaking?

DJDuque:
> how is this PR breaking?

Ok, I was under the (wrong) impression that adding enum variants to a public enum without an explicit underlying type could break ABI. Looking further it looks like there's no possible downstream C++ code that could stop compiling from this change, so I've just removed the `BREAKING CHANGE` tag from the PR title and description (that is enough right? The way PRs are merged ignores whatever commit messages my actual commits have? or should I force update those?).

andiwand:
I think we should go ahead with this PR and fix how the merging is done in the Examples

paulgessinger:
@DJDuque yes the release automation only looks at the squashed commit on the `main` branch. We generally don't version the ABI, so ~ if it compiles without downstream changes, it's not breaking for us.

acts-project-service:
## :red_circle: Athena integration test results [7b1edddb4dad8db5a5ef5beb39a9dc68fbc1d704]
Please investigate the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13244689)!

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930813)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930812)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930811)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930810)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930809)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930808)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930807)||
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930806)||
|:red_circle: | [test_ActsCheckObjectCountsWorkflowHgtd](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930805)||
|:red_circle: | [test_ActsCheckObjectCountsCachedWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930804)||
|:red_circle: | [test_ActsCheckObjectCountsLegacyWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930803)||
|:red_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930802)||
|:red_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930801)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930800)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930799)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930798)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930797)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930796)||
|:green_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930795)||
|:green_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930794)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930793)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930792)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930791)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930790)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930789)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930788)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930787)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930786)||
|:green_circle: | [test_ActsGx2fRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930785)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930784)||
|:green_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930783)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930782)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/63930781)||

goetzgaycken:
> If I understood correctly what @goetzgaycken said, this is because when `merge=true`, the digitization algorithm is internally using the clusterization implementation to "merge duplicates" (?). 

Duplicates are caused either by pileup or, I suppose, if different particles dispose charge directly or indirectly in the sample pixel. It feels kind of wrong/surprising that the clusterization is used/abused to merge multiple SimHits associated to the same pixels, but the algorithm to merge these SimHits and the actual clusterization would be very similar. So, by removing the possibility from the clusterization to be used for this purpose, code needs to be more or less duplicated.

PR #4783, opened 2025-11-07
feat: Allow to dump Tensors in numpy NPY format

benjaminhuth:
--- END COMMIT MESSAGE ---

### DIFF
--- a/Plugins/Gnn/include/ActsPlugins/Gnn/Tensor.hpp
+++ b/Plugins/Gnn/include/ActsPlugins/Gnn/Tensor.hpp
@@ -20,6 +20,8 @@
 #include <memory>
 #include <optional>
 #include <ostream>
+#include <fstream>
+#include <format>
 
 /// Forward declare cuda stream, to be able to use the header without cuda
 struct CUstream_st;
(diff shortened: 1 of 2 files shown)


paulgessinger:
What is this useful for?

benjaminhuth:
Debugging. You can super quickly load a tensor to python and be sure that it is the exact same numbers (other then exporting to csv somehow and reading...)

Plugins/Gnn/include/ActsPlugins/Gnn/Tensor.hpp
```diff
@@ -180,4 +190,38 @@ std::pair<Tensor<std::int64_t>, std::optional<Tensor<float>>> applyEdgeLimit(
...
+    maybeCpuTensor = this->clone({Device::Cpu(), stream});
+  }
+  const Tensor<T> &toDump =
+      maybeCpuTensor.has_value() ? maybeCpuTensor.value() : *this;
+
+  // Simple NPY header for 2D float32 array
+  std::ofstream ofs(filename, std::ios::binary);
+  if (!ofs.is_open()) {
+    throw std::runtime_error("Could not open file for writing: " + filename);
+  }
```
paulgessinger: You're not using this stream in this function, as you pass the path to the other function. Consider removing this here or passing the stream below.

PR #4785, opened 2025-11-08
fix: Fix typo in CMakeLists.txt for stacktrace option

njacazio:
Fix appended compiler flags

--- END COMMIT MESSAGE ---


asalzburger:
Looks like this now fails the `FpeMonitoring` UnitTests

paulgessinger:
Ok I dug around here a bit. I discovered that boost stacktrace has a fallback mode that requires the `dl` library to do the demangling. With the faulty option, we were accidentally using the fallback, and that seems to have worked. Fixing the option now forces the NOOP case, and the test doesn't expect this.

Since I can't push to this branch, I reopened it over in https://github.com/acts-project/acts/pull/4790.

PR #4804, opened 2025-11-13
ci: Simulation with Geant4 for GSF physmon

benjaminhuth:
The GSF monitoring profits from realistic simulated electrons. The simulation is extracted, so that the same simulation can be used for fitting and refitting tests. Also fixed the issue that the refitting workflow used muons instead of electrons previously.


CI/physmon/workflows/physmon_trackfitting_gsf.py
```diff
@@ -11,10 +11,19 @@
...
 setup = makeSetup()
 
+# Paths to pre-simulated data (from physmon_simulation_gsf.py)
+simDir = (setup.outdir / "../simulation_gsf/simulation").resolve()
+particlesPath = simDir / "particles_simulation.root"
+simhitsPath = simDir / "hits.root"
+
+# Verify simulation files exist
+assert particlesPath.exists(), f"Simulation not found: {particlesPath}"
+assert simhitsPath.exists(), f"SimHits not found: {simhitsPath}"
```
andiwand, line 21: after this the workflow can only be executed if the simulation script is run before right?

benjaminhuth: Hmm yeah thats true. But I'm not sure I would make this optional out of the box. Changing it is trivial in principle though (removing the asserts, pass `None`s to the function).
Would you propose to make it optional / adaptive?

acts-project-service:
## :red_circle: Athena integration test results [ea063679e5e1dff4fcc8d65a88e03bf4622ed69e]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/66373189) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13564345)!

PR #4813, opened 2025-11-20
fix: Cone surface fix path correction

junggjo9:
- Fix calculation of the path correction

--- END COMMIT MESSAGE ---

the shift of the direction vector by the transform translation makes the correction dependent on the location of the surface & other surface types are only exploiting the direction vectors themselves. Needs probably also to be fixed in the Trk::Surface inside athena?

Tagging: @paulgessinger. @andiwand, @asalzburger, @noemina


paulgessinger:
This *looks* correct to me, @asalzburger do you remember if there was anything particular about the cones?

Core/src/Surfaces/ConeSurface.cpp
```diff
@@ -15,6 +15,7 @@
 #include "Acts/Surfaces/detail/FacesHelper.hpp"
 #include "Acts/Surfaces/detail/VerticesHelper.hpp"
 #include "Acts/Utilities/Intersection.hpp"
+#include "Acts/Utilities/MathHelpers.hpp"
```
junggjo9:
```suggestion
```

Core/src/Surfaces/ConeSurface.cpp
```diff
@@ -141,11 +141,12 @@ double ConeSurface::pathCorrection(const GeometryContext& gctx,
   // (cos phi cos alpha, sin phi cos alpha, sgn z sin alpha)
   Vector3 posLocal = transform(gctx).inverse() * position;
   double phi = VectorHelpers::phi(posLocal);
-  double sgn = posLocal.z() > 0. ? -1. : +1.;
+  double sgn = -std::copysign(1., posLocal.z());
```
andiwand, line 144: I guess it is just a matter of taste but I think the original version is a bit easier to read (same below)

andiwand (commented):
where did this come up @junggjo9 

lgtm - to me the trinary way of determining the sign seems a bit easier to read

Core/src/Surfaces/ConeSurface.cpp
junggjo9: I found it by accident when I was replacing the tenary operations with a sign function which then turned into another more generic copySign() thingy. Anyhow, tests started to fail depending on whether the swap of the sign >= or >.

andiwand (approved):

acts-project-service:
## :red_circle: Athena integration test results [dc9f0968be0438b998da0d19aa3a69d825380984]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/65386447) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13441143)!

PR #4837, opened 2025-11-28
feat: Add Hough vertex filter config and seed tolerance

Anders200:
This PR adds a configuration switch to toggle the Hough vertex
filtering mechanism during the preprocessing stage. It introduces
a tolerance parameter to tweak the range at which seeds are
accepted by the filter, and exposes these configuration options
to the Python bindings.


andiwand:
@Anders200 can you rebase these changes?

pbalek:
tagging @tboldagh

PR #4862, opened 2025-12-02
feat: Add Hough vertex filter config and seed tolerance

Anders200:
This PR adds a configuration switch to toggle the Hough vertex
filtering mechanism during the preprocessing stage. It introduces
a tolerance parameter to tweak the range at which seeds are
accepted by the filter, and exposes these configuration options
to the Python bindings.


Core/include/Acts/Seeding/SeedFinder.ipp
```diff
@@ -369,6 +371,7 @@ SeedFinder<external_spacepoint_t, grid_t, platform_t>::getCompatibleDoublets(
         mutableData.setDeltaR(otherSP->index(),
                               std::sqrt(deltaR2 + (deltaZ * deltaZ)));
         outVec.push_back(otherSP);
+        // ACTS_INFO("zOrigin4=" << zOriginTimesDeltaR/deltaR);
```
tboldagh: This one and other will need to be removed. Or changed to VERBOSE

Examples/Algorithms/TrackFinding/src/SeedingAlgorithm.cpp
```diff
@@ -276,11 +303,47 @@ ProcessCode SeedingAlgorithm::execute(const AlgorithmContext& ctx) const {
...
+    Eigen::SelfAdjointEigenSolver<Acts::SquareMatrix3> saes(cov);
+    // eigenvalues are sorted in increasing order
+    Acts::Vector3 eivec = saes.eigenvectors().col(2);
+
+    // mean, eivec = start, direction
+    Acts::Vector3 norm{-1. * eivec[1], 1. * eivec[0], 0};
+    float norm_size = norm.norm();
+    float zDist = eivec.cross(norm).dot(mean) / (norm_size * norm_size);
+
+    if (std::abs(zDist - 20.2309) > 30.0)
```
tboldagh: Ideally the 20.2309 should become a constant with meaningful name.

Examples/Scripts/Python/full_chain_itk_hi.py
```diff
@@ -0,0 +1,178 @@
...
+        nMeasurementsMin=6,
+    ),
+    outputDirRoot=outputDir,
+    logLevel=acts.logging.INFO,
+)
+
+addVertexFitting(
+    s,
+    field,
+    # logLevel=acts.logging.VERBOSE,
```
tboldagh: Should be removed.

Examples/Scripts/Python/full_chain_itk_hi.py
```diff
@@ -0,0 +1,178 @@
...
+    initialSigmaPtRel=0.1,
+    initialVarInflation=[1.0] * 6,
+    geoSelectionConfigFile=geo_dir / "itk-hgtd/geoSelection-ITk.json",
+    outputDirRoot=outputDir,
+    logLevel=acts.logging.INFO,
+)
+
+addHoughVertexFinding(
+    s,
+    # logLevel=acts.logging.VERBOSE,
```
tboldagh: Commented lines should be removed.

Examples/Scripts/Python/full_chain_itk_hi.py
```diff
@@ -0,0 +1,178 @@
...
+s = acts.examples.Sequencer(events=1, numThreads=1, outputDir=str(outputDir), skip=1)
+
+addPythia8(
+    s,
+    hardProcess=["Top:qqbar2ttbar=on"],
+    npileup=0,
+    # beam=acts.PdgParticle.eLead,
+    cmsEnergy=5.0 * u.TeV,
+    vtxGen=acts.examples.GaussianVertexGenerator(
+        # stddev=acts.Vector4(0.0125 * u.mm, 0.0125 * u.mm, 55.5 * u.mm, 5.0 * u.ns),
```
tboldagh: Commended out code should be removed. Also in other places

.gitignore
```diff
@@ -39,3 +39,5 @@ codegen/src/codegen.egg-info/
 !.gitlab-ci.yml
 !.clang-tidy
 !.oclint
+
+itk/
```
pbalek: I think I added this to my fork of ACTS, but I would rather not propagate it into the official repository

.gitattributes
```diff
@@ -1 +1,3 @@
 *.ipp gitlab-language=c++
+itk/acts-itk/itk-hgtd/*.cbor filter=lfs diff=lfs merge=lfs -text
+itk/acts-itk/itk-hgtd/*.gdml filter=lfs diff=lfs merge=lfs -text
```
pbalek: this shouldn't be committed, or should it?

Core/include/Acts/Seeding/HoughVtxSeedFilter.hpp
pbalek, line 1: I think we don't need this file, since the filtering is implemented directly in the seeding.

Core/include/Acts/Seeding/HoughVtxSeedFilter.ipp
pbalek, line 1: and we also don't need this file neither

Examples/Algorithms/TrackFinding/src/SeedingAlgorithm.cpp
pbalek: >  Ideally the 20.2309 should become a constant with meaningful name.

No, 20.2309 should be the position of the Hough vertex; 30.0 should become some parameter. And all this should run only if `applySeedFiltering` is true.

Examples/Algorithms/TrackFinding/src/SeedingAlgorithm.cpp
```diff
@@ -276,11 +303,47 @@ ProcessCode SeedingAlgorithm::execute(const AlgorithmContext& ctx) const {
   ACTS_DEBUG("Created " << seeds.size() << " track seeds from "
                         << spacePointPtrs.size() << " space points");
 
+  static thread_local std::vector<seed_type> filtered_seeds;
+  filtered_seeds.clear();
+
+  auto t1 = std::chrono::high_resolution_clock::now();
```
pbalek: I don't think we really need this clock here. You can also remove `#include <chrono>`

Examples/Scripts/Python/full_chain_itk_hi.py
pbalek, line 1: I have a general question for this file. Do we want to add the entire "heavy ion" chain? Or maybe it would be better to modify `full_chain_itk.py`? Or leave our any heavy ions for now? Do people have some opinion regarding this?

Examples/Algorithms/TrackFinding/src/SeedingAlgorithm.cpp
pbalek: Wait, I was totally confused. This part would also filter the seeds, but the filtering is actually done via `setCustomCollisionRegion`. Thus, this whole block is not needed, is it?

pbalek: Hi @Anders200 , do you see this comment?

Anders200: @pbalek done, works

Examples/Python/python/acts/examples/reconstruction.py
```diff
@@ -383,24 +379,20 @@ def addSeeding(
...
-            trackingGeometry,
-            geoSelectionConfigFile,
-            stripGeoSelectionConfigFile,
-            logLevel,
+            s, trackingGeometry, geoSelectionConfigFile, logLevel
         )
-        seeds = None
-        perSeedParticleHypothesis = None
+        
+        addHoughVertexFinding(s, outputDirRoot=outputDirRoot)
```
pbalek: Hi @tboldagh , I guess here should be some "if", so it won't run for pp events. Can it be
```
if seedingAlgorithmConfigArg.fittedHoughVertices is not None:
```
or should we have a new argument for `addSeeding`?

AJPfleger:
discussed in dev-meeting: @pbalek : An improved approach is planed.

PR #4874, opened 2025-12-05
feat: Add naive component merger for GSF and benchmark

benjaminhuth:
It is not clear if the more complicated implementation of the component merging is a) faster and b) equivalent compared to a naive implementation. In order to evaluate this further, this PR adds the naive implementation and a benchmark.

--- END COMMIT MESSAGE ---


Core/src/TrackFitting/detail/GsfComponentMerging.cpp
```diff
@@ -46,8 +46,8 @@ GsfComponent detail::Gsf::mergeTwoComponents(const GsfComponent &a,
 
 namespace detail::Gsf {
 
-double SymmetricKLDistanceMatrix::computeSymmetricKlDivergence(
-    const GsfComponent &a, const GsfComponent &b) {
+double computeSymmetricKlDivergence(const GsfComponent &a,
```
andiwand:
```suggestion
double detail::Gsf::computeSymmetricKlDivergence(const GsfComponent &a,
```
nitpick: if you bind this explicitly to the namespace (and I think you have to pull it out of the namespace block), the compiler will match the declaration and definition. otherwise it is up to the linker to see that the definition really exists

Core/src/TrackFitting/GsfMixtureReduction.cpp
```diff
@@ -74,3 +78,39 @@ void Acts::reduceMixtureWithKLDistance(std::vector<GsfComponent> &cmpCache,
...
+  while (cmpCache.size() > maxCmpsAfterMerge) {
+    // Recompute ALL distances every iteration (naive approach)
+    double minDistance = std::numeric_limits<double>::max();
+    std::size_t minI = 0;
+    std::size_t minJ = 0;
+
+    for (std::size_t i = 0; i < cmpCache.size(); ++i) {
+      for (std::size_t j = i + 1; j < cmpCache.size(); ++j) {
+        double distance =
+            detail::Gsf::computeSymmetricKlDivergence(cmpCache[i], cmpCache[j]);
```
andiwand:
```suggestion
        const double distance =
            detail::Gsf::computeSymmetricKlDivergence(cmpCache[i], cmpCache[j]);
```
nitpick

Examples/Algorithms/TrackFitting/include/ActsExamples/TrackFitting/TrackFitterFunction.hpp
```diff
@@ -85,7 +85,7 @@ std::shared_ptr<TrackFitterFunction> makeKalmanFitterFunction(
                                                          Acts::Logging::INFO));
 
 /// Available algorithms for the mixture reduction
-enum class MixtureReductionAlgorithm { weightCut, KLDistance };
+enum class MixtureReductionAlgorithm { weightCut, KLDistance, naive };
```
andiwand: naive still relates to `KL` right? should this be made explicit here?

benjaminhuth: better `KLDistanceNaive`?

PR #4891, opened 2025-12-09
fix: Re-targeting for `DirectNavigator`

andiwand:
In some rare cases it is necessary to skip unreachable surfaces in the `DirectNavigator` is not implemented currently.

--- END COMMIT MESSAGE ---


Core/include/Acts/Propagator/DirectNavigator.hpp
```diff
@@ -293,29 +293,39 @@ class DirectNavigator {
...
-    // Move the sequence to the next surface
-    state.nextSurface();
+    // Establish & update the surface status
+    while (true) {
+      // Move the sequence to the next surface
+      state.nextSurface();
+
+      if (state.endOfSurfaces()) {
+        break;
+      }
```
benjaminhuth:
```suggestion
    // Establish & update the surface status
    for (state.nextSurface(); !state.endOfSurfaces(); state.nextSurface()) {
```

benjaminhuth (commented):
Maybe that would be easier to understand? Alternatively one could also pull the first `nextSurface` out to make the while loop have a condition.

Core/include/Acts/Propagator/DirectNavigator.hpp
benjaminhuth: or
```c++
state.nextSurface();
for(; !state.endOfSurfaces(); state.nextSurface()) {
```

Core/include/Acts/Propagator/DirectNavigator.hpp
```diff
@@ -376,7 +383,8 @@ class DirectNavigator {
 
     for (auto [intersectionIndex, intersection] :
          Acts::enumerate(intersections)) {
-      if (detail::checkPathLength(intersection.pathLength(), nearLimit,
+      if (intersection.isValid() &&
```
benjaminhuth, line 386: Just out of curilosity: Was this a bug before?

andiwand: discussed offline: this was a bug indeed. surfaces like the cylinder can be missed even with infinite bounds which would not be caught here

acts-project-service:
## :red_circle: Athena integration test results [b22a0d189477ab4587df250159d6963543369c23]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/66442211) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13573837)!

PR #4906, opened 2025-12-12
feat: sliding window alg for hough transform

tboldagh:
This PR adds sliding window as an option to filter peaks in Hough Transform space. 
The algorithm starts with finding bins in HT space that are above threshold and then applies SL window logic to discard nearby peaks. 
In addition, a function to export section of HT space can be exported in a form of a simple array. 
This is intended to feed an ML algorithm in future.

--- END COMMIT MESSAGE ---


Tests/UnitTests/Core/Seeding/HoughTransformTest.cpp
```diff
@@ -122,6 +121,93 @@ BOOST_AUTO_TEST_CASE(hough_transform_seeder) {
...
+          val = std::max(
+              val, peak - static_cast<HoughTransformUtils::YieldType>(dist));
+        }
+        plane.fillBin(x, y, x + y, x + y, val);
+      }
+    }
+  };
+
+  // Add two triangular peaks with maxima 4 at (4,4) and (4,5)
+  std::vector<std::array<std::size_t, 2>> testPeaks({{4, 4}, {4, 5}, {2, 6}});
```
pbalek, line 153: This would add three peaks, not two, wouldn't it?

Core/include/Acts/Seeding/HoughTransformUtils.hpp
```diff
@@ -446,6 +460,63 @@ class IslandsAroundMax {
...
 };
+
+/// @brief Peak finder using sliding window algorithm.
+/// First it finds peaks by scanning all space for cells with number of hits
+/// above threshold/ Then applies sliding window logic to eliminate peaks when
+/// maxima are adjacent leaving only one of them in a window. The logic is that
+/// it applies is to check cells around the peak and require that none on the
+/// upper right corner are above threshold and none in bottom left corner is
+/// below or equal to the peak. It can be illustrated as follows for window size
+/// of 1:
```
pbalek: "The logic is that it applies is to check cells" - this doesn't sound right.
Also, I didn't get the logic. Upper right corner can't be above threshold (i.e. no peak-candidates), and lower left corner can't be higher than. What if the threshold is 2, and window is 3x3:
```
1 1 1
3 5 1
7 4 1
```
Then "5" can't be the peak (because there's "7"), nor "7" (because there's "5" which is above threshold). Thus, where is the peak here?

Core/include/Acts/Seeding/HoughTransformUtils.hpp
```diff
@@ -446,6 +460,63 @@ class IslandsAroundMax {
...
+  /// size of the window in x direction for sliding window
+  std::size_t xWindowSize = 2;
+  /// size of the window in y direction for sliding window
+  std::size_t yWindowSize = 2;
+  /// perform recentering
+  bool recenter = true;
+  /// size of the window in x direction for recentering
+  std::size_t xRecenterSize = 3;
+  /// size of the window in y direction for recentering
+  std::size_t yRecenterSize = 3;
```
pbalek, line 494: Shouldn't this be the same size as `xWindowSize` and `yWindowSize`, or smaller, but definitely not larger? If you find a maximum in 2x2 window, there might be a higher value just outside the window, which will be included here.

Core/include/Acts/Seeding/HoughTransformUtils.hpp
```diff
@@ -446,6 +460,63 @@ class IslandsAroundMax {
...
+std::vector<typename HoughPlane<identifier_t>::Index> slidingWindowPeaks(
+    const HoughPlane<identifier_t>& plane, const SlidingWindowConfig& config);
+
+/// @brief Obtain an image around the peak
+/// @tparam identifier_t Hough plane content
+/// @param plane Hough plane to work on
+/// @param index peak center
+/// @param xSize number of cells around the peak in x direction
+/// @param ySize number of cells around the peak in y direction
+/// @return the vector with count of hits starting from lower left to upper right corner of rectangular window
```
pbalek, line 514: Would vector of vectors be more intuitive?

Core/include/Acts/Seeding/HoughTransformUtils.ipp
```diff
@@ -413,3 +427,118 @@ void Acts::HoughTransformUtils::PeakFinders::IslandsAroundMax<identifier_t>::
...
+bool passWindow(
+    const typename Acts::HoughTransformUtils::HoughPlane<identifier_t>::Index
+        index,
+    const Acts::HoughTransformUtils::HoughPlane<identifier_t>& plane,
+    const Acts::HoughTransformUtils::PeakFinders::SlidingWindowConfig& config) {
+  using YieldType = Acts::HoughTransformUtils::YieldType;
+  auto [xmax, ymax] = index;
+  YieldType max = plane.nHits(xmax, ymax);
+  // window loop
+  // this loop needs to be smarter to take care of wrapping
```
pbalek, line 443: can't this be solved with a version of `checkIndices` what doesn't throw exceptions but simply returns true/false?

Core/include/Acts/Seeding/HoughTransformUtils.ipp
```diff
@@ -413,3 +427,118 @@ void Acts::HoughTransformUtils::PeakFinders::IslandsAroundMax<identifier_t>::
...
+  YieldType tot = 0;
+  auto [xmax, ymax] = index;
+  YieldType maxValue = plane.nHits(xmax, ymax);
+  for (std::size_t x = xmax - config.xRecenterSize;
+       x <= xmax + config.xRecenterSize; ++x) {
+    for (std::size_t y = ymax - config.yRecenterSize;
+         y <= ymax + config.yRecenterSize; ++y) {
+      const YieldType noOfHits = plane.nHits(x, y);
+      if (noOfHits >= maxValue) {
+        // this needs to be smarter to take care of wrapping
```
pbalek: same comment as above - `checkIndices` almost solves this

Core/include/Acts/Seeding/HoughTransformUtils.ipp
```diff
@@ -413,3 +427,118 @@ void Acts::HoughTransformUtils::PeakFinders::IslandsAroundMax<identifier_t>::
...
+  YieldType xw = 0;
+  YieldType yw = 0;
+  YieldType tot = 0;
+  auto [xmax, ymax] = index;
+  YieldType maxValue = plane.nHits(xmax, ymax);
+  for (std::size_t x = xmax - config.xRecenterSize;
+       x <= xmax + config.xRecenterSize; ++x) {
+    for (std::size_t y = ymax - config.yRecenterSize;
+         y <= ymax + config.yRecenterSize; ++y) {
+      const YieldType noOfHits = plane.nHits(x, y);
```
pbalek: I propose to change the variable name to `numOfHits`; `noOfHits` almost looks like "no hits".

Core/include/Acts/Seeding/HoughTransformUtils.ipp
```diff
@@ -413,3 +427,118 @@ void Acts::HoughTransformUtils::PeakFinders::IslandsAroundMax<identifier_t>::
...
+slidingWindowRecenter(
+    const typename Acts::HoughTransformUtils::HoughPlane<identifier_t>::Index
+        index,
+    const Acts::HoughTransformUtils::HoughPlane<identifier_t>& plane,
+    const Acts::HoughTransformUtils::PeakFinders::SlidingWindowConfig& config) {
+  using YieldType = Acts::HoughTransformUtils::YieldType;
+  YieldType xw = 0;
+  YieldType yw = 0;
+  YieldType tot = 0;
+  auto [xmax, ymax] = index;
```
pbalek: since there's later `if(noOfHits >= maxValue) ...`, then `xmax` and `ymax` are not necessarily maximum, are they?

Core/include/Acts/Seeding/HoughTransformUtils.ipp
```diff
@@ -413,3 +427,118 @@ void Acts::HoughTransformUtils::PeakFinders::IslandsAroundMax<identifier_t>::
...
+  };
+
+  for (std::size_t x = 0; x < xSize; ++x) {
+    for (std::size_t y = 0; y < ySize; ++y) {
+      int xPlaneCoord = x + xmax - xSize / 2;
+      int yPlaneCoord = y + ymax - ySize / 2;
+      output.push_back(isInside(xPlaneCoord, yPlaneCoord)
+                           ? static_cast<pixel_value_t>(
+                                 plane.nHits(xPlaneCoord, yPlaneCoord))
+                           : pixel_value_t{});
```
pbalek, line 540: As it is, you would fill the output with zeros, wouldn't you?
so wouldn't be better to do:
```suggestion
      if(isInside(xPlaneCoord, yPlaneCoord))
        output.push_back(static_cast<pixel_value_t>(
                             plane.nHits(xPlaneCoord, yPlaneCoord)));
```

pbalek (commented):
Hi Tomasz, here are few comments.

Core/include/Acts/Seeding/HoughTransformUtils.hpp
pbalek: Maybe I could have checked the code before :-) 
It seems the text should be "... upper right corner are above **the peak** and ..."
And in the example above, I think we can only conclude that "5" is not a peak and that's all. Am I right?

Core/include/Acts/Seeding/HoughTransformUtils.ipp
tboldagh: Going beyond the HT image is memory error. If this happens the consequences should be most severe possible. I.e. exception. I am afraid that mere true/false maybe be easily missed.

Core/include/Acts/Seeding/HoughTransformUtils.hpp
tboldagh: The intention is that it would be passed to some ML algorithm which will linearise it in input, so why not to do it right away. And it is much cheaper to produce 1D vector.

Core/include/Acts/Seeding/HoughTransformUtils.hpp
tboldagh: Thanks for spotting this. Corrected.

Tests/UnitTests/Core/Seeding/HoughTransformTest.cpp
tboldagh: Indeed.

Core/include/Acts/Seeding/HoughTransformUtils.hpp
tboldagh: I would imagine that the reasonable setting would indeed be identical or smaller. But it would not hurt to experiment with larger size too.

Core/include/Acts/Seeding/HoughTransformUtils.ipp
tboldagh: That is precisely the intention. I.e. to pad with zeros.

Core/include/Acts/Seeding/HoughTransformUtils.ipp
tboldagh: Indeed, it is center in fact. Fixed accordingly.

Core/include/Acts/Seeding/HoughTransformUtils.ipp
tboldagh: Followed.

Core/include/Acts/Seeding/HoughTransformUtils.ipp
tboldagh: As we discussed offline, it is trickier. For not this can stay as is.

acts-project-service:
## :red_circle: Athena integration test results [32ba2b7391c7abf436960392e4e6f35bcd9b58b5]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67380514) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13727002)!

paulgessinger:
Hey @tboldagh, sonarcloud [flags](https://sonarcloud.io/project/issues?severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=BUG&id=acts-project_acts&open=AZukZWn2yhSn-4qiMRvY) this line

https://github.com/acts-project/acts/blob/21d3f91bcabf85ee4e3d692a6e6c7e20dc6b3da9/Core/include/Acts/Seeding/HoughTransformUtils.ipp#L450

as a potential issue due to `-xdist` where `xdist` is an `std::size_t`. I'm not sure why this didn't show up in the sonarcloud report on this PR. I think this essentially wraps around the unsigned int (UB?), but the unit test seems to rely on that behavior, so the fix wasn't straightforward to me.

Could you look into this?

tboldagh:
> Hey @tboldagh, sonarcloud [flags](https://sonarcloud.io/project/issues?severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=BUG&id=acts-project_acts&open=AZukZWn2yhSn-4qiMRvY) this line
> 
> https://github.com/acts-project/acts/blob/21d3f91bcabf85ee4e3d692a6e6c7e20dc6b3da9/Core/include/Acts/Seeding/HoughTransformUtils.ipp#L450
> 
> as a potential issue due to `-xdist` where `xdist` is an `std::size_t`. I'm not sure why this didn't show up in the sonarcloud report on this PR. I think this essentially wraps around the unsigned int (UB?), but the unit test seems to rely on that behavior, so the fix wasn't straightforward to me.
> 
> Could you look into this?

Sure, technically it would be a new PR. Right?

paulgessinger:
@tboldagh yes, just thought this is the best place to report this.

tboldagh:
> @tboldagh yes, just thought this is the best place to report this.
Will try fixing it asap. I think it actually points to a genuine bug.

PR #4910, opened 2025-12-15
feat: add alignment workflow for the OpenDataDetector

Zequn-SUN:
This PR introduces a complete detector alignment workflow for the OpenDataDetector (ODD), implementing a full pipeline from simulation with the nominal geometry, alignment with the misaligned geometry, to reconstructions with the nominal, misaligned, and aligned geometry, respectively.  The script implementing the complete workflow is `Examples/Scripts/Python/alignment_odd.py`

This script needs a json file containing the transform information of each ODD sensor. This json file can be found here
https://cernbox.cern.ch/files/spaces/eos/user/z/zesun/odd_transforms.json

For detailed documentation and usage instructions, please refer to:
https://docs.google.com/document/d/1FHyWZoeHhygSMH9F7IPQbE7_Dx9tlPPXQhbNMYYZLJ8/edit?usp=sharing

--- END COMMIT MESSAGE ---


asalzburger:
Hi @Zequn-SUN - there are small conflicts due to the change of the python directories, `Examples/Python` does not exist in main anymore, it moved to `Python/Examples`.

Zequn-SUN:
Hi @asalzburger ,
Thanks for pointing that out. I've updated the code (https://github.com/Zequn-SUN/acts/tree/alignTest) to match the new directory structure (Python/Examples instead of Examples/Python) and merged the latest changes from main. I've also resolved the compilation errors related to the AlignedTransformUpdater API changes (switching from DetectorElementBase* to SurfacePlacementBase*) and updated the Python bindings accordingly.
Sorry for the delay in updating the PR. The changes are now pushed and should be ready for review. Please let me know if there are any other issues.

Best regards
Zequn

AJPfleger:
Hi @Zequn-SUN @asalzburger what's the status on this PR?

Core/src/Surfaces/Surface.cpp
```diff
@@ -83,26 +94,29 @@ AlignmentToBoundMatrix Surface::alignmentToBoundDerivativeWithoutCorrection(
...
   // Calculate the derivative of local frame axes w.r.t its rotation
   const auto [rotToLocalXAxis, rotToLocalYAxis, rotToLocalZAxis] =
       detail::rotationToLocalAxesDerivative(rotation);
   // Calculate the derivative of local 3D Cartesian coordinates w.r.t.
   // alignment parameters (without path correction)
   AlignmentToPositionMatrix alignToLoc3D = AlignmentToPositionMatrix::Zero();
-  alignToLoc3D.block<1, 3>(eX, eAlignmentCenter0) = -localXAxis.transpose();
-  alignToLoc3D.block<1, 3>(eY, eAlignmentCenter0) = -localYAxis.transpose();
-  alignToLoc3D.block<1, 3>(eZ, eAlignmentCenter0) = -localZAxis.transpose();
+  alignToLoc3D.block<1, 3>(eX, eAlignmentCenter0) =
```
goblirsc: Hi, testing this change with an injected global Z shift of a telescope detector seems to make the alignment assign the resulting correction to the global y coordinate. Are you sure this needs to be changed as done here? Maybe you can elaborate a bit on the derivation?

Zequn-SUN: Hi, thanks for your testing! 
I think the confusion comes from a change in the **definition of the translation parameters** between the two versions.

### Before the change (global translation)

In the previous implementation, the translation was defined in the **global coordinate system**, i.e.
$$ \mathbf{x}_g = \mathbf{R}\mathbf{x}_l + \mathbf{t} $$
where:
* (\mathbf{R}): local -> global rotation matrix
* (\mathbf{t}): translation in **global coordinates**
* (\mathbf{x}_g, \mathbf{x}_l): position vectors in global/local frames

In this case, the derivative of the local coordinates w.r.t. the global translation gives:
```cpp
const auto& localXAxis = rotation.col(0);
alignToLoc3D.block<1, 3>(eX, eAlignmentCenter0) = -localXAxis.transpose();
```
which is correct, since the global shift needs to be projected onto the local axis via (\mathbf{R}).

---

### After the change (local translation)

In the updated version, the translation is defined in the **local coordinate system**, i.e.
$$ \mathbf{x}_g = \mathbf{R}(\mathbf{x}_l + \mathbf{t}) $$
Now (\mathbf{t}) lives directly in the **local frame**, so the derivative becomes:
```cpp
alignToLoc3D.block<1, 3>(eX, eAlignmentCenter0) = -Vector3::UnitX().transpose();
```
In other words, since the translation is already expressed in the local frame, the Jacobian no longer involves (\mathbf{R}), and reduces to the identity matrix.

---

### About your observation 
Given this, the effect you're seeing is likely due to a **frame misunderstanding**:

If you're injecting a **local Z shift**, and in your detector geometry the **local Z axis happens to align with global Y**, then the alignment will naturally interpret this as a shift along global Y.

Let me know if that matches what you're seeing :)

XiaocongAi: > Hi, testing this change with an injected global Z shift of a telescope detector seems to make the alignment assign the resulting correction to the global y coordinate. Are you sure this needs to be changed as done here? Maybe you can elaborate a bit on the derivation?

Hi, may I ask how is the telescope detector placed, i.e. are normals of the telescope planes along the global Z axis?

goblirsc: Hi! In principle, this can be picked freely (`binValue` indicates the orientation of the system), see [here](https://github.com/acts-project/acts/blob/main/Examples/Scripts/Python/millepede_alignment.py#L77). The nominal choice in my test was with normals along global y, B-field in global z, and sensor planes in x-z.

XiaocongAi: Thanks! Yes, I implemented the `binValue` for the orientation options a few years. Has your problem being solved already? We didn't have such problem when we played with the telescope detector earlier.

Zequn-SUN: Hi @goblirsc and @XiaocongAi , sorry for my late reply. I was busy with the graduation, so I only just completed the rebase.

The attachment is some slides, including the explanation of the code changes for the derivative definition, and some tests we performed using `telescope ` and `odd`. The results of misalignment and alignment in these tests are consistent with our expectations.
[AlignmentPR4910.pdf](https://github.com/user-attachments/files/27986893/AlignmentPR4910.pdf)

XiaocongAi: Hi @Zequn-SUN , thanks! This looks good to me. Hi @asalzburger , do you have time to have a look?

Examples/Algorithms/Alignment/include/ActsExamples/Alignment/AlignmentAlgorithm.hpp
```diff
@@ -138,6 +148,14 @@ class AlignmentAlgorithm final : public IAlgorithm {
...
   WriteDataHandle<AlignmentParameters> m_outputAlignmentParameters{
       this, "OutputAlignmentParameters"};
+
+  // Collected track data from all events
+  mutable std::vector<std::vector<IndexSourceLink>> m_collectedSourceLinks;
+  mutable TrackParametersContainer m_collectedInitialParameters;
+  mutable std::shared_ptr<MeasurementContainer> m_collectedMeasurements;
+  mutable Acts::GeometryContext m_savedGeoContext;
+  mutable Acts::MagneticFieldContext m_savedMagFieldContext;
+  mutable Acts::CalibrationContext m_savedCalibContext;
```
goblirsc, line 183: Is it safe here to use the context of the first event for tracks from all events? 

Do you store all the individual items rather than the `TrackAlignmentState`s because you want to do a full (non-linear) recalculation of the alignment matrix from fresh track fits at each alignment iteration?

Examples/Algorithms/Alignment/src/AlignmentAlgorithm.cpp
```diff
@@ -106,32 +169,61 @@ ProcessCode AlignmentAlgorithm::execute(const AlgorithmContext& ctx) const {
...
   }
 
-  // add alignment parameters to event store
-  m_outputAlignmentParameters(ctx, std::move(alignedParameters));
+  // Finalize() has no AlgorithmContext / per-event WhiteBoard (each event's
+  // store is destroyed after that event). Single-event main writes
+  // m_outputAlignmentParameters from execute(); here transforms are already
+  // applied via alignedTransformUpdater. Extract from detector / e.g.
+  // MutableGeoIdAlignmentStore for downstream, as in alignment examples.
+
```
goblirsc, line 372: suggest to write results to a text file. Ideally check the implementation in #5432 for consistency with the MP-II related path.

Python/Examples/src/plugins/Alignment.cpp
```diff
@@ -98,4 +134,85 @@ PYBIND11_MODULE(ActsExamplesPythonBindingsAlignment, m) {
...
+            cfg.iterationState.clear();
+            for (auto item : d) {
+              unsigned int iter = item.first.cast<unsigned int>();
+
+              std::uint8_t mask_value;
+              try {
+                auto mask_enum =
+                    item.second.cast<ActsAlignment::AlignmentMask>();
+                mask_value = static_cast<std::uint8_t>(mask_enum);
+              } catch (...) {
```
goblirsc: Will conversion failures here result in a quiet assignment of an empty alignment mask? Better to catch a concrete type and to print an error message?

version_number
```diff
@@ -1 +1 @@
-999.999.999
+44.0.1
```
goblirsc: is this change intentional?

Examples/Algorithms/Alignment/src/AlignmentAlgorithm.cpp
```diff
@@ -106,32 +169,61 @@ ProcessCode AlignmentAlgorithm::execute(const AlgorithmContext& ctx) const {
...
-      ctx.geoContext, ctx.magFieldContext, ctx.calibContext, extensions,
-      Acts::PropagatorPlainOptions(ctx.geoContext, ctx.magFieldContext),
+      m_savedGeoContext, m_savedMagFieldContext, m_savedCalibContext,
+      extensions,
+      Acts::PropagatorPlainOptions(m_savedGeoContext, m_savedMagFieldContext),
       &(*pSurface));
 
-  // Set the alignment options
+  // 转换 iterationState: std::map<uint, bitset<6>> -> std::map<uint,
+  // AlignmentMask>
```
goblirsc: "conversion of" ?

Examples/Algorithms/Alignment/src/AlignmentAlgorithm.cpp
```diff
@@ -106,32 +169,61 @@ ProcessCode AlignmentAlgorithm::execute(const AlgorithmContext& ctx) const {
...
+  std::map<unsigned int, ActsAlignment::AlignmentMask> alignIterationState;
+  for (const auto& [iter, mask] : m_cfg.iterationState) {
+    alignIterationState[iter] =
+        static_cast<ActsAlignment::AlignmentMask>(mask.to_ulong());
+  }
+
+  // Set the alignment options (包括 iterationState)
```
goblirsc: including iterationState?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+import json
+import random
+import numpy as np
+import ctypes
+import os
+import argparse
+import sys
+
+
+def _preload_dd4hep_for_macos() -> None:
```
goblirsc: is this specific to the ODD alignment? Else it should probably go into a more central module?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+    for var in ("DYLD_LIBRARY_PATH", "DD4HEP_LIBRARY_PATH"):
+        cur = os.environ.get(var, "")
+        merged = []
+        for p in path_bits + [x for x in cur.split(":") if x]:
+            if p not in merged:
+                merged.append(p)
+        os.environ[var] = ":".join(merged)
+    mode = getattr(ctypes, "RTLD_GLOBAL", 8)
+    for names in (
+        ("libDD4hepGaudiPluginMgr.1.32.dylib", "libDD4hepGaudiPluginMgr.dylib"),
```
goblirsc: is it safe to hardcode library versions here?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+    plugin manager and libDDCorePlugins in the same process often fixes that.
+    """
+    if sys.platform != "darwin":
+        return
+    root = os.environ.get("DD4HEP_ROOT") or os.environ.get("DD4HEP_INSTALL")
+    libdir: Optional[Path] = Path(root) / "lib" if root else None
+    if libdir is None or not libdir.is_dir():
+        try:
+            candidate = (
+                Path(__file__).resolve().parents[5] / "DD4hep" / "install" / "lib"
```
goblirsc: is this too specific regarding the install location?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+
+import acts
+import acts.examples
+import acts.examples.alignment
+from acts.examples.alignment import AlignmentDecorator
+
+# Sequencer may otherwise raise after Kalman+RootTrackStatesWriter (FLTINV in writeT).
+# ACTS CI uses ACTS_SEQUENCER_FAIL_ON_UNMASKED_FPE=0; set to 1 to fail on any unmasked FPE.
+if os.environ.get("ACTS_SEQUENCER_FAIL_ON_UNMASKED_FPE") is None:
+    os.environ["ACTS_SEQUENCER_FAIL_ON_UNMASKED_FPE"] = "0"
```
goblirsc: is this automatically hiding FPEs if the user does not manually ask for them? 
That seems like an inversion of the usual behaviour. Better to fix the FPEs themselves?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+
+u = acts.UnitConstants
+
+
+def parse_geoid(id_str):
+    """Parse ID string, e.g. 'vol=16|lay=4|sen=1|ext=2'"""
+    parts = {}
+    for part in id_str.split("|"):
+        key, value = part.split("=")
+        parts[key] = int(value)
```
goblirsc: add protection for unexpected / invalid (sub) strings? Will whitespaces break the parsing?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+
+# Sequencer may otherwise raise after Kalman+RootTrackStatesWriter (FLTINV in writeT).
+# ACTS CI uses ACTS_SEQUENCER_FAIL_ON_UNMASKED_FPE=0; set to 1 to fail on any unmasked FPE.
+if os.environ.get("ACTS_SEQUENCER_FAIL_ON_UNMASKED_FPE") is None:
+    os.environ["ACTS_SEQUENCER_FAIL_ON_UNMASKED_FPE"] = "0"
+
+u = acts.UnitConstants
+
+
+def parse_geoid(id_str):
```
goblirsc: this method and the next one both might be better fits for a more common module, as they are relevant for all geo identifiers?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+
+def geoid_to_id_string(gid: acts.GeometryIdentifier) -> str:
+    """Convert GeometryIdentifier to ID string format: 'vol=X|lay=Y|sen=Z|ext=W'"""
+    parts = [f"vol={gid.volume}", f"lay={gid.layer}", f"sen={gid.sensitive}"]
+    if gid.extra != 0:
+        parts.append(f"ext={gid.extra}")
+    return "|".join(parts)
+
+
+def match_filter(geoid_parts, target_vol, target_lay, target_sen, target_ext):
```
goblirsc: could specify required type for `geoid_parts` ?

goblirsc: again a candidate for a separate module? Also very long nested `if`-logic. Is the "backward compatibility" case still used somewhere?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+                if sen_id != target_sen:
+                    return False
+
+    if target_ext != -1:
+        if "ext" not in geoid_parts or geoid_parts["ext"] != target_ext:
+            return False
+    return True
+
+
+def match_surface_to_filter(gid, target_vol, target_lay, target_sen, target_ext):
```
goblirsc: how does this one differ from the previous function? Does it have to be a full reimplementation or could it be a call to the function above? 
Is it specific to the ODD alignment example, or could it go into a more general module?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+                if sen_id not in target_sen:
+                    return False
+            else:
+                if sen_id != target_sen:
+                    return False
+
+    if target_ext != -1:
+        if gid.extra != target_ext:
+            return False
+    return True
```
goblirsc: can all this be replaced by a call to `match_filter`?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+# ============================================================================
+
+
+def runSimulation(
+    trackingGeometry: acts.TrackingGeometry,
+    field: acts.MagneticFieldProvider,
+    digiConfigFile: Path,
+    outputDir: Path,
+    detector=None,
+    numEvents=100,
```
goblirsc: maybe make the default a few more, as 100 tracks are not enough for a serious alignment?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+    t = trf.translation
+    r = trf.rotation
+    return acts.Transform3(
+        acts.Vector3(float(t[0]), float(t[1]), float(t[2])),
+        acts.RotationMatrix3(
+            acts.Vector3(r[0, 0], r[1, 0], r[2, 0]),
+            acts.Vector3(r[0, 1], r[1, 1], r[2, 1]),
+            acts.Vector3(r[0, 2], r[1, 2], r[2, 2]),
+        ),
+    )
```
goblirsc: is this specific to this module? Also not sure about what it is actually supposed to do, is this just explicitly spelling out a deep-copy operation?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+
+
+def collect_matching_sensitive_surfaces(
+    trackingGeometry: acts.TrackingGeometry,
+    target_volume=-1,
+    target_layer=-1,
+    target_sensitive=-1,
+    target_extra: int = -1,
+) -> list:
```
goblirsc: this could go into a common module with all of the other geometry / surface related filtering functions above? None seem specific to ODD alignment

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+    return matches
+
+
+def _asymmetric_random(mag: float, enabled: bool) -> float:
+    if not enabled:
+        return 0.0
+    if random.random() < 0.5:
+        return random.uniform(-mag * 1.5, -mag * 0.5)
+    return random.uniform(mag * 0.5, mag * 1.5)
+
```
goblirsc: better fit for the C++ random generators module?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+
+def apply_random_misalignment_to_transform(
+    trf: acts.Transform3,
+    tx: int = 0,
+    ty: int = 0,
+    tz: int = 0,
+    rx: int = 0,
+    ry: int = 0,
+    rz: int = 0,
```
goblirsc: what do these do? Are they intended as bools?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+    shift_mag_mm: float = 2.0,
```
goblirsc: this seems fairly huge as a default?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+        l_shift_y.shift = _asymmetric_random(shift_mag_mm, True)
+        l_shift_y(trf)
+
+    if tz != 0:
+        l_shift_z = acts.examples.alignment.AlignmentGeneratorLocalShift()
+        l_shift_z.axisDirection = acts.AxisDirection.AxisZ
+        l_shift_z.shift = _asymmetric_random(shift_mag_mm, True)
+        l_shift_z(trf)
+
+    current_rotation = trf.rotation
```
goblirsc: is `current_rotation` needed? It only seems to be read and replaced by itself

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+) -> dict:
+    """Build a JSON-serializable transform record for misalignment/alignment I/O."""
+    rot = trf.rotation
+    trans = trf.translation
+    rotation_matrix = [[float(rot[i, j]) for j in range(3)] for i in range(3)]
+    return {
+        "ID": geoid_to_id_string(gid),
+        "Translation": [float(trans[0]), float(trans[1]), float(trans[2])],
+        "RotationMatrix": rotation_matrix,
+    }
```
goblirsc: possibly interesting beyond this example?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+        f"Misalignment filter: volume={target_volume!r}, "
+        f"layer={target_layer}, sensitive={target_sensitive}, extra={target_extra} "
+        f"-> {num_selected} surfaces"
+    )
+
+
+def setupMisalignment(
+    trackingGeometry: acts.TrackingGeometry,
+    outputDir: Path,
+    target_volume=-1,  # Can be int or list of ints
```
goblirsc: does this also support the nested dict approach?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+    target_layer=-1,  # Can be int or list of ints
+    target_sensitive=-1,  # Can be int or list of ints
+    target_extra: int = -1,
+    tx: int = 0,
+    ty: int = 0,
+    tz: int = 1,
+    rx: int = 0,
+    ry: int = 0,
+    rz: int = 0,
```
goblirsc: are these bools?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+    # Create AlignedTransformUpdater, it will update mutableStore after each iteration
+    aal_cfg.alignedTransformUpdater = (
+        acts.examples.alignment.makeAlignedTransformUpdater(mutableStore)
+    )
+
+    aal_cfg.alignedDetElements = det_elements
+    aal_cfg.align = acts.examples.alignment.makeAlignmentFunction(
+        trackingGeometry, field, acts.logging.INFO
+    )
+    aal_cfg.maxNumIterations = 100  # Reasonable number of iterations
```
goblirsc: isn't 100 full iterations of fitting all the tracks and building the full matrix a bit excessive...?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+    aal_cfg.maxNumTracks = 100000
```
goblirsc: could increase this?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+            current_rotation[0, 2], current_rotation[1, 2], current_rotation[2, 2]
+        )
+        l_rot_z.axis = local_z_axis
+        l_rot_z.angle = _asymmetric_random(rotation_mag_rad, True)
+        l_rot_z(trf)
+
+    return trf
+
+
+def transform_to_alignment_record(
```
goblirsc: If this is intended as a data format for storing alignment records, it might be better to implement this somewhere more central than the ODD alignment example steering script. See my other comment, I was actually wondering why the alignment algorithm does not produce any output file. Tacking this on locally in the example steering hides this feature from users who may not be after this specific example setup.

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+    # These variables must be defined even when using hierarchical structure (for backward compatibility)
+    target_layer = -1  # -1 = all layers (used only if target_volume is not a dict)
+    target_sensitive = -1  # Used only if target_volume is not a dict
+    target_extra = -1  # -1 = all extra
+
+    initialvarinflation = [100, 100, 100, 100, 100, 1]
+
+    tx, ty, tz = 1, 1, 0  # Translation DOF
+    rx, ry, rz = 1, 1, 1  # Rotation DOF
+    shift_mag_mm = 0.5  # Shift magnitude (mm)
```
goblirsc: still a bit on the large side?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+        alignment_dof |= Center1
+    if tz:
+        alignment_dof |= Center2
+    if rx:
+        alignment_dof |= Rotation0
+    if ry:
+        alignment_dof |= Rotation1
+    if rz:
+        alignment_dof |= Rotation2
+
```
goblirsc: this is a lot of code to generate a single-use integer. Put into a function, or just insert directly as a constant?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+
+def runReconstruction(
+    trackingGeometry: acts.TrackingGeometry,
+    field: acts.MagneticFieldProvider,
+    digiConfigFile: Path,
+    inputDir: Path,
+    outputDir: Path,
+    detector=None,
+    numEvents=10,
```
goblirsc: just 10 events?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+            sys.exit(1)
+
+        with open(json_path, "r") as f:
+            transform_records = json.load(f)
+
+        geoIdMap = geo_id_map_from_transform_records(transform_records)
+        print(
+            f"Reconstruction ({mode}): {len(geoIdMap)} aligned elements from {json_path.name}"
+        )
+        mutableStore = acts.examples.alignment.MutableGeoIdAlignmentStore(geoIdMap)
```
goblirsc: again, not very thrilled that the I/O for loading and storing alignment transforms is defined on the fly in a python steering file

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+        help="Output directory (default: reconstruction_output)",
+    )
+
+    args = parser.parse_args()
+
+    if not args.command:
+        parser.print_help()
+        sys.exit(1)
+
+    srcdir = Path(__file__).resolve().parent.parent.parent.parent
```
goblirsc: can this be done more cleanly? Seems heavily dependent on a particular user folder structure

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+
+        runAlignment(
+            trackingGeometry=trackingGeometry_mis,
+            field=field,
+            digiConfigFile=digiConfigFile,
+            inputDir=simulationDir,
+            outputDir=alignmentDir,
+            detector=detector_mis,
+            numEvents=args.num_events,
+        )
```
goblirsc: small nitpick, doesn't Alignment imply (initial) reconstruction?

Examples/Scripts/Python/alignment_odd.py
```diff
@@ -0,0 +1,1387 @@
...
+        print(f"  - Reconstruction (aligned): {reconDir_ali}")
+
+    elif args.command == "simulation":
+        outputDir = (
+            Path(args.output_dir)
+            if args.output_dir
+            else Path.cwd() / "simulation_output"
+        )
+        detector, trackingGeometry = build_odd(misaligned=True)
+
```
goblirsc: no mkdir for the output dir here?

goblirsc (commented):
Hi, adding a few comments on the current version.

From my side, I am fine with moving to local coordinates for the alignment DoF definition. 

One thing I would like to suggest giving one more iteration is the new example steering python macro - it is 1300 lines of code and implements several important pieces of the workflow (geometry selection, random number generators, persistification and loading of alignment corrections via json). All those features will not be available to people using the alignment infrastructure outside this particular ODD alignment example, and not visible to people who do not look for them in the final steering script. This could benefit from some refactoring, with important functionality going into common, re-usable modules. Some components might even be worth porting to the C++ side, to avoid limiting users who do not work with the python examples bindings. 

Cheers, Max

goblirsc:
> Hi, adding a few comments on the current version.
> 
> From my side, I am fine with moving to local coordinates for the alignment DoF definition.
> 
> One thing I would like to suggest giving one more iteration is the new example steering python macro - it is 1300 lines of code and implements several important pieces of the workflow (geometry selection, random number generators, persistification and loading of alignment corrections via json). All those features will not be available to people using the alignment infrastructure outside this particular ODD alignment example, and not visible to people who do not look for them in the final steering script. This could benefit from some refactoring, with important functionality going into common, re-usable modules. Some components might even be worth porting to the C++ side, to avoid limiting users who do not work with the python examples bindings.
> 
> Cheers, Max

sorry, one more :) 
Can the required json which is currently on your personal CernBox be added to the repository? 
At least I was not able to access it via the link.

goblirsc:
Hi @Zequn-SUN, just a ping on this PR, it would be great to see this nice work available in main-ACTS soon!  
Please feel free to ping us in case anything is unclear in the review or if there are difficulties in getting things into place.

Zequn-SUN:
Hi @goblirsc, thank you again for the detailed review and apologize for my so late update. I will first respond to the remaining points one by one below. I will update my Git branch shortly, but due to network connectivity issues in China, I am temporarily unable to push the latest changes. And the attachment is the updated results for the latest version.
[PR4910_update.pdf](https://github.com/user-attachments/files/31171491/PR4910_update.pdf)

version_number
Zequn-SUN: This change was not intentional. The alignment work was originally developed on top of ACTS v44.0.1, and the `version_number` file was accidentally carried over during rebases / merges. It has been reverted to the upstream placeholder `999.999.999`. Sorry for the confusion.

Examples/Algorithms/Alignment/src/AlignmentAlgorithm.cpp
Zequn-SUN: Yes. This Chinese comment has been rewritten in English.

Examples/Algorithms/Alignment/src/AlignmentAlgorithm.cpp
Zequn-SUN: Yes. This Chinese comment has been rewritten in English.

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: `mkdir(parents=True, exist_ok=True)` is now called at the start of every stage (simulation, misalignment, alignment, reconstruction, and the full chain).

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: Yes, it does. To make that clearer, the workflow is now split explicitly into:
 step2.1 misalignment: apply random local misalignment and write txt files only
 step2.2 run seeding + KF track fitting + `AlignmentAlgorithm` on the misaligned geometry

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: Path resolution has been cleaned up. Instead of relying on a fixed `parent.parent.parent...` chain, it is now centralized in `kfalignment_paths.find_acts_source_dir()`, which walks upward from the script location until it finds `Examples/Configs/odd-digi-smearing-config.json`. This removes the dependency on a particular user folder structure and makes the example more robust to different checkout locations.

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: I agree. This part has been refactored so that the transform I/O is no longer defined ad hoc inside the steering script. The loading and storing of alignment transforms, including the 6-column delta text format, index maps, and GeoID-based lookup helpers, has been moved into `kfalignment_transform_io.py`. The steering script now only orchestrates the workflow and calls these reusable helpers.

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: This has been updated. The default is now uniformly `DEFAULT_NUM_EVENTS = 10000`, and it can still be overridden via `--num-events`.

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: I agree. 

Now, first, the transform I/O logic is no longer defined ad hoc inside the ODD steering script; it has been moved into reusable helper modules under `Examples/Scripts/Python/odd_alignment/`, in particular `kfalignment_transform_io.py`.

Second, I also implemented text output directly in `AlignmentAlgorithm::finalize()`, so the alignment algorithm itself now writes a Millepede-style result text and an optional index map via `outputAlignmentFile` / `outputAlignmentIndexFile`, instead of relying only on local steering-script logic.

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: I kept this at `100` and `100000` for now as a practical upper bound for the current example workflow, mainly to avoid collecting an unnecessarily large track sample by default. I agree this could be made configurable or increased if needed for larger studies, but for this PR I kept the default conservative.

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: This is just a conservative safety cap. In the current tests, the alignment usually converges well before that and does not exceed about 25 iterations in practice. So `100` is not meant as a typical runtime expectation, only as a safe upper bound to avoid stopping too early in less well-behaved cases.

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: Yes. The target selection now supports the nested-dict form as well. In the latest workflow, this is handled centrally in the misalignment configuration parsing, so inputs such as `17:8,24:4` are converted into the corresponding nested dictionary structure before being passed into the misalignment setup.

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: Yes, I agree. This is potentially useful beyond this particular example, which is why the transform I/O logic has been pulled out of the monolithic steering script and moved into reusable helper code (`kfalignment_transform_io.py`) instead of keeping it embedded locally in `alignment_odd.py`.

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: This temporary variable is no longer present in the refactored version of the workflow; the redundant read/store pattern was removed during the cleanup of the transform-handling code.

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: I agree. As a first step, these geometry / surface selection helpers have been moved out of the monolithic steering script and grouped into a dedicated module (`kfalignment_geoid.py`) rather than keeping them inline in `alignment_odd.py`.

For this PR I stopped at that level to keep the refactoring scope manageable, but I agree that these helpers are not inherently ODD-specific and could be promoted further into a more common utility module in a follow-up.

If you have a preferred location or design in mind for turning this into a more common utility module, I would be glad to follow that recommendation in a follow-up.

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: Yes, this helper is essentially just an explicit deep-copy operation for `acts.Transform3`.

In the refactored version it is no longer embedded in the monolithic steering script; it now lives in the reusable transform-I/O helper module (`kfalignment_transform_io.py`).

Examples/Scripts/Python/alignment_odd.py
Zequn-SUN: Yes, this has been cleaned up and unified across the workflow interfaces. The default event count is now handled centrally via `DEFAULT_NUM_EVENTS = 10000`, and the individual stages use that shared default unless overridden explicitly via `--num-events`.

Examples/Algorithms/Alignment/include/ActsExamples/Alignment/AlignmentAlgorithm.hpp
Zequn-SUN: Yes, that is the intention.

I went back and forth on whether to add a runtime check here, but in the end I decided that a comment was the better choice for now. The assumption is tied to the current design: tracks are collected in `execute()` and the actual alignment is run once in `finalize()`, where no per-event `AlgorithmContext` is available.

For the current ODD workflow this is safe, because it assumes a single IOV with one shared `MutableGeoIdAlignmentStore`, together with constant B-field and calibration across the processed events. The saved `GeometryContext` points to that shared alignment store, which is updated in place via `alignedTransformUpdater`, rather than to a frozen geometry snapshot from event 0.

And yes, the individual `IndexSourceLink`s, initial track parameters, and measurements are stored intentionally instead of precomputed `TrackAlignmentState`s, precisely so that `ActsAlignment::Alignment` can refit the tracks at every iteration and rebuild the alignment matrix in a fully iterative (non-linear) way.

I added a comment in `AlignmentAlgorithm.hpp` to make these assumptions explicit.

Examples/Algorithms/Alignment/src/AlignmentAlgorithm.cpp
Zequn-SUN: Implemented. `AlignmentAlgorithm::finalize()` now writes a Millepede-style result text file and an optional index map via `outputAlignmentFile` and `outputAlignmentIndexFile`.

The text output follows the same 6-column convention (`label value 0 difference error 99`) so that it is consistent with the MP-II related path and with the format used in #5432.

Python/Examples/src/plugins/Alignment.cpp
Zequn-SUN: Yes, that was a valid concern. I changed this so that the binding no longer fails silently.

`iterationState` now accepts only `AlignmentMask` or `int`; any other type raises a `TypeError` explicitly instead of quietly resulting in an empty alignment mask.

Examples/Algorithms/Alignment/src/AlignmentAlgorithm.cpp
```diff
@@ -8,21 +8,90 @@
...
+
+double covarianceError(const Acts::DynamicMatrix& cov, std::size_t row) {
+  if (cov.rows() <= static_cast<Eigen::Index>(row) ||
+      cov.cols() <= static_cast<Eigen::Index>(row)) {
+    return 0.;
+  }
+  const double var =
+      cov(static_cast<Eigen::Index>(row), static_cast<Eigen::Index>(row));
+  if (!(var > 0.) || !std::isfinite(var)) {
+    return 0.;
```
goblirsc: this looks like a potential silent failure mode if the user passes an invalid row or has a broken covariance matrix

Examples/Algorithms/Alignment/include/ActsExamples/Alignment/AlignmentAlgorithm.hpp
goblirsc: thanks! Did you test if this runs with many events e.g. mu=200 ttbar? I worry a bit about the resulting RAM usage.

Examples/Algorithms/Alignment/src/AlignmentAlgorithm.cpp
```diff
@@ -8,21 +8,90 @@
...
+
+std::string geoIdToString(Acts::GeometryIdentifier id) {
+  std::ostringstream os;
+  os << "vol=" << id.volume() << "|lay=" << id.layer()
+     << "|sen=" << id.sensitive();
+  if (id.extra() != 0u) {
+    os << "|ext=" << id.extra();
+  }
+  return os.str();
+}
```
goblirsc: is there a better location for this, possibly in `Core/Utilities`?

Examples/Detectors/Common/include/ActsExamples/DetectorCommons/AlignedTransformUpdater.hpp
goblirsc: a bit confusing that `AlignedTransformUpdater` does not declare or define `AlignedTransformUpdater` (which actually lives in `ActsAlignment/Kernel/Alignment.hpp`

Examples/Detectors/Common/include/ActsExamples/DetectorCommons/AlignedTransformUpdater.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+ protected:
+  // Allow access to base class's private members
+  using GeoIdAlignmentStore::m_transformMap;
+};
+
+/// @brief Default aligned transform updater
+/// This is a no-op updater that just returns true
+/// Use makeAlignedTransformUpdater() to create a functional updater
+inline ActsAlignment::AlignedTransformUpdater alignedTransformUpdater =
```
goblirsc: do we ever use the no-op updater? The name is what someone would naively give a configured instance, so having this identifier in the ActsExamples namespace (which many will implicitly fetch via `using namespaceActsExamples` might lead to occasional collisions. If it's not urgently needed as a static object, could instead make this a generator function (`makeNoopAlignedTransformUpdater()`)?

Examples/Scripts/Python/odd_alignment/kfalignment_paths.py
```diff
@@ -0,0 +1,79 @@
...
+            p = libdir / n
+            if p.is_file():
+                try:
+                    ctypes.CDLL(str(p), mode=mode)
+                except OSError:
+                    break
+                break
+
+
+def setup_runtime_env() -> None:
```
goblirsc: I still don't think we should override this item - can we fix the FPEs instead?

Examples/Scripts/Python/odd_alignment/kfalignment_paths.py
```diff
@@ -0,0 +1,79 @@
...
+import ctypes
+import os
+import sys
+from pathlib import Path
+from typing import Optional
+
+_DIGI_CONFIG_REL = Path("Examples/Configs/odd-digi-smearing-config.json")
+
+
+def preload_dd4hep_for_macos() -> None:
```
goblirsc: this still seems on the one side very specific (hardcoded library versions e.g. for DDCorePlugins) and at the same time, if it is needed for all mac users working with DD4Hep, too general to go into `odd_alignment`?

andiwand:
@Zequn-SUN to move forward with this I think it would be great to split this PR into smaller junks. my suggestion would be the following
- this PR: Alignment + Core + Python + Test changes
- second PR: Examples changes excluding `Examples/Scripts/Python/odd_alignment`
- third PR: `Examples/Scripts/Python/odd_alignment/full_chain_odd_KFalignment.py` scripts where we should consolidate which of them are generic enough to bring to ACTS

what do you think?

Zequn-SUN:
Hi @andiwand , I think this is a very good idea ! It should make the changes easier to review and keep each PR more focused . BTW, should I close this current PR and then create new PRs separately ?

andiwand:
> Hi @andiwand , I think this is a very good idea ! It should make the changes easier to review and keep each PR more focused . BTW, should I close this current PR and then create new PRs separately ?

Thanks @Zequn-SUN ! yes might even be preferable to close this one and open new ones for the split. you can reference this pr here so the discussion and effort is easier to find

PR #4919, opened 2025-12-16
fix: import acts.examples before importing json modules

pbutti:
While testing the material mapping instructions from 
https://acts.readthedocs.io/en/latest/examples/howto/material_mapping.html
I noticed that importing the json material map module fails due to 
```
ImportError: generic_type: type "JsonDetectorElement"
referenced unknown base type "Acts::DetectorElementBase"
```

I think that's due to the fact that the DetectorElementBase has not been bound in Python when this module is imported. 
It seems that `DetectorElementBase` is defined when importing `acts.examples` module, so I suggest to modify the scripts to load that before the `acts.json` module. I think the solution is bit ugly and maybe we could force the acts.json plugin to always import acts.examples with something like

```
PYBIND11_MODULE(ActsPluginsPythonBindingsJson, json) {
  py::module_::import("acts.examples");   
  ...
}
```
But I let you suggest what do you think it's best.


acts-project-service:
## :red_circle: Athena integration test results [8e0940ef55643b43a6b87f942ce18f85c1ec960b]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/66692676) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13611556)!

paulgessinger:
Shouldn't the detector element base be in core?

PR #4353, opened 2025-06-18
feat: move Root based MaterialTrack reading/writing from Examples to Plugins

asalzburger:
This PR moves `RootMaterialTrack`{`Reader`/`Writer`} bulk code from the Examples to the Root plugin and thus makes it accessible for experiments w/o building the Examples.

It its on top of #4352.

--- END COMMIT MESSAGE ---

Any further description goes here, @-mentions are ok here!

- Use a *conventional commits* prefix: [quick summary](https://www.conventionalcommits.org/en/v1.0.0/#summary)
  - We mostly use `feat`, `fix`, `refactor`, `docs`, `chore` and `build` types.
- A milestone will be assigned by one of the maintainers

PR #4929, opened 2025-12-21
chore: remove broken Mat_map_surface_plot_*.C macros

veprbl:
The sur_type and sur_range_min/max were removed in https://github.com/acts-project/acts/pull/4353, so these macros don't work anymore, as they require this information recorded. It looks like there is a move towards Python macros such as material_mapping_check.py, so these can be removed.

--- END COMMIT MESSAGE ---


andiwand:
I am not familiar with these scripts and there purpose - do you think you could have a look @asalzburger ?

veprbl:
Not needed after #4931

PR #4930, opened 2025-12-21
fix: handle undefined geometry IDs in KalmanFitter verbose output

### LINKED ISSUE #3097: bug: KF verbosity statement
This is not a critical bug, but there is a verbose statement in the Kalman Fitter that causes a seg fault when running with verbose output. This statement [here](https://github.com/acts-project/acts/blob/22730e1e5faa4abb218289269448d89e412b1430/Core/include/Acts/TrackFitting/KalmanFitter.hpp#L892) can cause a crash when the KF is propagating in the final stage back to the perigee surface when the perigee is created like it is in the track fitting algorithm [here](https://github.com/acts-project/acts/blob/22730e1e5faa4abb218289269448d89e412b1430/Examples/Algorithms/TrackFitting/src/TrackFittingAlgorithm.cpp#L88-L89) since it does not have an assigned geometry ID.

I create an issue as there are two solutions, and I don't know which the development team suggests (I'm happy to implement either)

1. Change the verbose statement to not assume a surface has a geometry ID. This will make the verbose statement potentially less useful without the ID.
2. Assign a geometry ID to the perigee surface at construction time (is this even possible?).

heathdutton:
Surfaces without assigned geometry IDs (like perigee surfaces) now get a clearer log message in materialInteractor instead of just printing "undefined" for the geometry ID.

Closes #3097

### DIFF
--- a/Core/include/Acts/Seeding/CompositeSpacePointLineSeeder.hpp
+++ b/Core/include/Acts/Seeding/CompositeSpacePointLineSeeder.hpp
@@ -43,8 +43,12 @@ concept CompositeSpacePointSeedFiller =
       /// @param t0: Offse in the time of arrival of the particle
       /// @param testSp: Reference to the straw space point to test
       {
-        filler.candidatePull(cctx, pos, dir, t0, testSp)
+        filler.candidateChi2(cctx, pos, dir, t0, testSp)
       } -> std::same_as<double>;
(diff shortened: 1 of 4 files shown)


paulgessinger:
I'm wondering if an alternative solution would be to make the geometry id outstream operator say something other than "undefined"

andiwand:
no activity from the author - closing for now

PR #4933, opened 2026-01-06
refactor: Remove raw pointer from portal shell class

dimitra97:
Remove the raw pointer method from the portal shells and get it from the smart pointers

--- END COMMIT MESSAGE ---


Core/include/Acts/Geometry/CuboidPortalShell.hpp
```diff
@@ -70,9 +65,6 @@ class SingleCuboidPortalShell : public CuboidPortalShell {
   /// @copydoc PortalShellBase::size
   std::size_t size() const final;
 
-  /// @copydoc CuboidPortalShell::portal
-  Portal* portal(Face face) final;
-
   /// @copydoc CuboidPortalShell::portalPtr
   std::shared_ptr<Portal> portalPtr(Face face) final;
```
paulgessinger: Do you think this method should be renamed?


```suggestion
  std::shared_ptr<Portal> portal(Face face) final;
```

paulgessinger (dismissed):

Core/include/Acts/Geometry/CuboidPortalShell.hpp
dimitra97: yes why not - although was thinking if the current name is more descriptive? don' t know, I am fine by renaming it also

paulgessinger: Yeah, if you don't mind!

acts-project-service:
## :red_circle: Athena integration test results [6d608d855f63150155d3171aae85ce9e3394c6fd]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67645623) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13765631)!

abarton5:
It this a good idea? You are triggering increments and decrements on every call?

paulgessinger:
@abarton5 this should only affect the construction, and not the behavior while e.g. navigating. I would expect the performance impact to be minimal. We could consider changing to the return value to a const ref.

PR #4934, opened 2026-01-06
refactor: Refactoring of string parsing code for Gbts LUT and connection table

jpreston-cern:
This is the first in a series of PR's that aim to bring GBTS up to date with ACTS coding standards. 

This PR focuses on refactoring the parsing code for the LUT so that its in its own file mimicking that of the connector file as well as moving it to the constructor of GbtsSeedingAlgorithm. It also includes some code clean up, as well as moving the seedFinder constructor to initialize as well. This gives some speed increase due to not parsing the LUT every event. 

--- END COMMIT MESSAGE ---

This will affect the Athena API but a branch is already ready to submit to the canary 

@timadye @andiwand


Core/include/Acts/Seeding/GbtsDataStorage.hpp
```diff
@@ -102,7 +103,7 @@ class GbtsDataStorage {
  protected:
   const GbtsGeometry& m_geo;
   const SeedFinderGbtsConfig& m_config;
-  std::vector<std::array<float, 5>> m_mlLUT;
+  const std::vector<std::array<float, 5>>& m_mlLUT;
   std::vector<GbtsEtaBin> m_etaBins;
```
andiwand: the reference members look concerning to me from a lifetime perspective. these properties are captured by `const &` in the constructor which would leave the impression on me that they will be copied and I can dispose them afterwards.

To improve this I would
- use a `std::shared_ptr` for `GbtsGeometry`
- copy the `SeedFinderGbtsConfig` by value. this is a very common pattern in ACTS
- not absolutely sure about the size and purpose of `m_mlLUT`. if this is the only client I would use value semantics which allows users to move it in. if it is shared between other classes I would make a struct / typedef and pass it by `std::shared_ptr`

Core/include/Acts/Seeding/GbtsDataStorage.hpp
```diff
@@ -79,8 +79,9 @@ class GbtsEtaBin {
 
 class GbtsDataStorage {
  public:
-  explicit GbtsDataStorage(const GbtsGeometry& geometry,
-                           const SeedFinderGbtsConfig& config);
+  explicit GbtsDataStorage(
+      const GbtsGeometry& geometry, const SeedFinderGbtsConfig& config,
+      const std::vector<std::array<float, 5>>& parsedLutFile);
   ~GbtsDataStorage();
```
andiwand, line 87: you mentioned that the destructors will be handled in a different PR right?

Core/include/Acts/Seeding/GbtsLutParser.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include <fstream>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+namespace Acts::Experimental {
+
+class GbtsLutParser {
+ public:
+  explicit GbtsLutParser(std::string& lutInputFile, bool useML) {
```
andiwand: I think this rather involved implementation should be moved to a source file which would also break the include chain towards io libraries

Core/include/Acts/Seeding/GbtsLutParser.hpp
```diff
@@ -0,0 +1,58 @@
...
+        ifs.close();
+      }
+    }
+  }
+  const std::vector<std::array<float, 5>>& getParsedLut() const {
+    return m_mlLUT;
+  }
+
+ private:
+  std::vector<std::array<float, 5>> m_mlLUT{};
```
andiwand: looking at the rest of the implementation I would rather use a free function to model the behavior. generally I would avoid doing heavy lifting in constructors as this will be unexpected. the behavior can be easily modeled in a functional way `params -> std::vector<std::array<float, 5>>`

Core/include/Acts/Seeding/GbtsDataStorage.hpp
jpreston-cern: yes there is a dedicated PR for the DataStorage files that gets rid of the destructor among other things

Core/include/Acts/Seeding/GbtsDataStorage.hpp
jpreston-cern: In the PR that looks at GbtsDataStorage this is changed to `const GbtsGeometry* m_geo;` 

in terms of `GbtsGeometry` that is created in the API constructor as a `std::unique_ptr`  (either in examples in acts or the `GbtsSeedingTool` ) which will mean it stays alive throughout the run. And then a raw pointer to it is passed to `SeedFinderGbts` then `GbtsDataStorage` . is this a good way to do things? or should i always use std::shared_ptr` for those kind of situations?

andiwand: I think you can either move the `std::unique_ptr` in case this is the only client. by doing that you would explicitly move the ownership to `GbtsDataStore`. same can be done with `mlLUT`

if this has multiple clients I would rather use `std::shared_ptr`

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/GbtsSeedingAlgorithm.hpp
```diff
@@ -111,6 +116,9 @@ class GbtsSeedingAlgorithm final : public IAlgorithm {
   // collection of geometry objects used by GBTS
   std::vector<Acts::Experimental::TrigInDetSiLayer> m_layerGeometry{};
 
+  // actual seed finder algorithm
+  std::unique_ptr<Acts::Experimental::SeedFinderGbts> m_finder = nullptr;
+
   // used to assign LayerIds to the GbtsActsMap
   mutable std::map<int, int> m_LayeridMap{};
```
benjaminhuth, line 119: This is outside the diff, but it concerns me that this is `mutable`. As far as I see this is not modified in the `execute` function. If so though, it definitively needs to be explained in the comments.

Examples/Algorithms/TrackFinding/src/GbtsSeedingAlgorithm.cpp
```diff
@@ -69,6 +62,11 @@ ActsExamples::GbtsSeedingAlgorithm::GbtsSeedingAlgorithm(
   // manually convert min Pt as no conversion available in ACTS Examples
   // (currently inputs as 0.9 GeV but need 900 MeV)
   m_cfg.seedFinderConfig.minPt = m_cfg.seedFinderConfig.minPt * 1000;
+
+  m_finder = std::make_unique<Acts::Experimental::SeedFinderGbts>(
+      m_cfg.seedFinderConfig, m_gbtsGeo.get(), &m_layerGeometry,
+      m_lutParser.get(), logger().cloneWithSuffix("GbtsFinder"));
```
benjaminhuth: Besides Andreas' comment that this could be a free function (see above), I would prefer not to release ownership from a `unique_ptr` like that if it is not necessary (same above for `m_gbtsGeo`).

Core/include/Acts/Seeding/GbtsLutParser.hpp
timadye: I think it is reasonable to have something like `GbtsLutParser` as a holder for the lookup table, rather than using `std::vector<std::array<float, 5>>` directly. Perhaps it could better be called something like `GbtsLookupTable`. This could go all the way into `GbtsDataStorage`, so all usage could be via `GbtsLookupTable::get()` (more compact than `getParsedLut()`) or other methods.

As @andiwand suggests, it is probably good to move the file parsing to separate code, perhaps in a `GbtsLookupTable.cpp` file. It could be a `private` method called from the constructor - I personally don't see a problem with constructors that do a lot of work.

andiwand: I think conceptually the class `GbtsLutParser` is not a parser right now. a parser should have a method to parse. currently this class is more like a `GbtsLookupTable` like you suggested. but even then I would not recommend to construct it from a path and letting the constructor read the file.

so ultimately I would recommend either having a class `GbtsLookupTable` or simply a typedef with that name to `std::vector<std::array<float, 5>>` in case this is the only property to hold. to get a `GbtsLookupTable` from file I would either have a static method on `GbtsLookupTable` or a free function that returns the object

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/GbtsSeedingAlgorithm.hpp
jpreston-cern: no worries thank you for looking through ive changed it to a const now. thank you for pointing it out

Examples/Algorithms/TrackFinding/src/GbtsSeedingAlgorithm.cpp
jpreston-cern: yes i agree and have now re worked it so that the `m_lutParser` and `m_gbtsGeo` are now both moved into the seed finder

Core/include/Acts/Seeding/GbtsLutParser.hpp
jpreston-cern: ive pushed some changes to the branch which move the creation of that LUT structure used by datastorage `m_mlLut` to its own function and also `cpp` file, although i pushed this before reading these comments so the naming hasnt been changed (although easily renamed), 

I quite like the idea of some sort of parsing class being passed in to avoid copying in potentially large vectors in the future, currently it is only 60 lines of an `array` size 5 long but as i imagine this part of the code is still under active development this might change and could cause slow down if it has to be copied in every event. 

I now have this parsing class being passed in as a shared pointer where the `storage` which can then pull a ref to the LUT information where it is used

jpreston-cern:
i have added all the changes suggested so far, the only thing i was thinking was that as the shared pointer of both `geometry` and `Lutparser` (name to be changed) is first defined as a member function of `SeedFinderGbts` and is then passed in per instance of the `createSeed`  function, if there is a lot of multi threading of events, would that cause slow down because of the atomic reference count being high/ changing a lot?

Core/include/Acts/Seeding/GbtsDataStorage.hpp
```diff
@@ -100,9 +102,9 @@ class GbtsDataStorage {
   }
 
  protected:
-  const GbtsGeometry& m_geo;
-  const SeedFinderGbtsConfig& m_config;
-  std::vector<std::array<float, 5>> m_mlLUT;
+  std::shared_ptr<const GbtsGeometry> m_geo;
+  const SeedFinderGbtsConfig m_config;
```
andiwand:
```suggestion
  SeedFinderGbtsConfig m_config;
```
no need to use `const` here as the type `GbtsDataStorage` itself can be used as `const`

Core/include/Acts/Seeding/GbtsDataStorage.hpp
```diff
@@ -79,8 +80,9 @@ class GbtsEtaBin {
 
 class GbtsDataStorage {
  public:
-  explicit GbtsDataStorage(const GbtsGeometry& geometry,
-                           const SeedFinderGbtsConfig& config);
+  explicit GbtsDataStorage(std::shared_ptr<const GbtsGeometry> geometry,
+                           const SeedFinderGbtsConfig config,
```
andiwand:
```suggestion
                           const SeedFinderGbtsConfig& config,
```
this should either be passed in as a `const &` or by value and `std::move`d

Core/include/Acts/Seeding/GbtsLutParser.hpp
andiwand: I think the interface ergonomics are still a bit odd. IMO a parser should return something and not change its state if the parsing is done in a single step. so I would either use a parser + data struct where the parser returns the struct or simply omit the parser as it does not have any configuration other than the bool

copying the data is no concern IMO because the `std::vector` can simply be `std::move`d and/or put into a `std::shared_ptr`

Core/include/Acts/Seeding/SeedFinderGbts.hpp
```diff
@@ -57,30 +56,31 @@ class SeedFinderGbts {
 
   SeedContainer2 CreateSeeds(
```
andiwand:
```suggestion
  SeedContainer2 createSeeds(
```

Core/include/Acts/Seeding/SeedFinderGbts.hpp
```diff
@@ -57,30 +56,31 @@ class SeedFinderGbts {
...
       const RoiDescriptor& roi,
-      const SPContainerComponentsType& SpContainerComponents, int max_layers);
+      const SPContainerComponentsType& SpContainerComponents,
+      int max_layers) const;
 
-  std::vector<std::vector<SeedFinderGbts::GNN_Node>> CreateNodes(
-      const auto& container, int MaxLayers);
+  std::vector<std::vector<GbtsNode>> CreateNodes(const auto& container,
```
andiwand:
```suggestion
  std::vector<std::vector<GbtsNode>> createNodes(const auto& container,
```

andiwand: also I would avoid `auto` here

andiwand:
> i have added all the changes suggested so far, the only thing i was thinking was that as the shared pointer of both `geometry` and `Lutparser` (name to be changed) is first defined as a member function of `SeedFinderGbts` and is then passed in per instance of the `createSeed` function, if there is a lot of multi threading of events, would that cause slow down because of the atomic reference count being high/ changing a lot?

`GbtsDataStorage` is only created once per event, no? this also contains the `GbtsLutParser` right now which is held by the `std::shared_ptr`. in this case I would not worry about performance in any way. there might be a bit more indirection but without detailed profiling this should be not acted on

in case this happens for every seed you might be able to see some performance effects but still I would expect it to be extremely small

Core/include/Acts/Seeding/GbtsDataStorage.hpp
jpreston-cern: `GbtsDataStorage` has a member veriable `std::vector<GbtsEtaBin> m_etaBins` which is modified as a part of the `loadnodes` function, if `GbtsDataStorage` is made const is it ok to make that veriable a `mutable` ?

andiwand: no I think in this case `GbtsDataStorage` should just stay non-const downstream. but since `m_config` is hidden from the user it does not need to be `const internally

Core/include/Acts/Seeding/GbtsLutParser.hpp
jpreston-cern: nice ok that makes sense, for `GbtsDataStorage ` it is important that even if the `std::vector<std::array<float, 5>>m_mlLut` veriable is empty, that it is still passed in as the size of the vector is used for the logic in terms of if clusterwidth is used. would it be reasonable to either:

change to the parser style class that returns the data struct like you said which can then be used to pass in if it is available, maybe having another constructor which accepts the creation of `GbtsdataStorage` without passing in the data struct.

or have that `gbtsLookupTable` class and pass that in as a whole which can then be used by `GbtsDataStorage` to get the vector (empty or not)

andiwand: to me the easiest solution seems to be something like this
```c++
// lookup table alias
using GbtsMLLookupTable = std::vector<std::array<float, 5>>;

// unconditionally parse the file
GbtsMLLookupTable parseGbtsMLLookupTable(const std::string &path);
```

then the `GbtsDataStorage` will take `GbtsMLLookupTable` by value in the constructor and move it to the member.
```c++
GbtsDataStorage(
      std::shared_ptr<const GbtsGeometry> geometry, const SeedFinderGbtsConfig& config,
      GbtsMLLookupTable mlLut) : m_geometry(std::move(geometry)), m_config(config), m_mlLut(std::move(mlLut)) {}
```

where `parseGbtsMLLookupTable` is called, it is done conditionally. if it is not parsed an empty `GbtsMLLookupTable` is created. then the `GbtsMLLookupTable` instance is `std::move`d into `GbtsDataStorage`

jpreston-cern: ok that makes sense thank you! in terms of placing this parsing function, should it be in examples the same as the `connector` or within core in `SeedFinderGbts`?

andiwand: I think the parsing function can be part of the `Core` functionality and be called from the Examples

jpreston-cern:
I have added the parsing function to the `SeedFinderGbts` class which stores the parsed information as a member veriable, this is then passed to each instance of `GbtsDataStorage` . i thought this made the most sense due to the seed finder creating the instances of the `GbtsDataStorage` and already having a copy of the config, but i can move it if needed

acts-project-service:
## :red_circle: Athena integration test results [8f3a7cfeb6f3e13d1b176e100ed59058f318c2be]
Please investigate the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13791370)!

|status|job|report|
|------|---|------|
|:green_circle: | [report_pull_request](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837265)||
|:green_circle: | [run_workflow_tests_run4_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837264)||
|:green_circle: | [run_workflow_tests_run2_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837263)||
|:green_circle: | [run_workflow_tests_run2_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837262)||
|:green_circle: | [run_workflow_tests_run3_mc](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837261)||
|:green_circle: | [run_workflow_tests_run3_data](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837260)||
|:green_circle: | [test_ActsPersistifyEDM](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837259)||
|:green_circle: | [test_ActsDumpGeometryIdentifiers](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837258)||
|:red_circle: | [test_ActsCheckObjectCountsWorkflowHgtd](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837257)||
|:red_circle: | [test_ActsCheckObjectCountsCachedWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837256)||
|:red_circle: | [test_ActsCheckObjectCountsLegacyWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837255)||
|:red_circle: | [test_ActsCheckObjectCountsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837254)||
|:green_circle: | [test_ActsEFTrackFit](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837253)||
|:green_circle: | [test_ActsPersistifySeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837252)||
|:green_circle: | [test_ActsBenchmarkWithSpot](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837251)||
|:green_circle: | [test_ActsAnalogueClustering](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837250)||
|:green_circle: | [test_ActsWorkflowHeavyIons](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837249)||
|:green_circle: | [test_ActsWorkflowFastTracking](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837248)||
|:red_circle: | [test_ActsWorkflowCached](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837247)||
|:red_circle: | [test_ActsWorkflow](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837246)||
|:green_circle: | [test_ActsValidateAmbiguityResolution](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837245)||
|:green_circle: | [test_ActsValidateResolvedTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837244)||
|:green_circle: | [test_ActsValidateTracks](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837243)||
|:green_circle: | [test_ActsValidateActsCoreSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837242)||
|:green_circle: | [test_ActsValidateActsSpacePoints](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837241)||
|:green_circle: | [test_ActsValidateSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837240)||
|:green_circle: | [test_ActsValidateOrthogonalSeeds](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837239)||
|:green_circle: | [test_ActsValidateClusters](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837238)||
|:green_circle: | [test_ActsGx2fRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837237)||
|:green_circle: | [test_ActsGSFRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837236)||
|:red_circle: | [test_ActsKfRefitting](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837235)||
|:green_circle: | [test_ActsExtrapolationAlgTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837234)||
|:green_circle: | [test_ActsITkTest](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67837233)||

Core/src/Seeding/GbtsDataStorage.cpp
```diff
@@ -101,38 +101,13 @@ void GbtsEtaBin::generatePhiIndexing(float dphi) {
         std::pair<float, unsigned int>(phi + 2 * std::numbers::pi, nIdx));
   }
 }
-GbtsDataStorage::GbtsDataStorage(const GbtsGeometry& geometry,
-                                 const SeedFinderGbtsConfig& config)
-    : m_geo(geometry), m_config(config) {
+GbtsDataStorage::GbtsDataStorage(std::shared_ptr<const GbtsGeometry> geometry,
+                                 const SeedFinderGbtsConfig& config,
+                                 const GbtsMLLookupTable& parseLutFile)
```
timadye, line 106: `parseLutFile` &rarr; `mlLUT`, since this is the table itself, no longer the parser. Similarly in the `.hpp` file.

Sorry I missed this before (but anyway wouldn't want to hold up the merge). Maybe this can be included in a future PR.

andiwand: to other thing would be to pass this by value, i.e. `GbtsMLLookupTable mlLUT` so the user can use `std::move` and avoid the copy

PR #4935, opened 2026-01-06
chore: move pytests to be co-located with the bindings

asalzburger:
Since the python bindings are now split into Core, Plugins, Examples - also the pytest files can be better co-located.

This is a first step towards this.

--- END COMMIT MESSAGE ---


AJPfleger:
@asalzburger in dev-meeting: Let's close this PR and restart with an updated version.

PR #4939, opened 2026-01-07
feat: Add boost::histogram wrappers and ROOT converter

benjaminhuth:
Adds classes for 1D and 2D Histograms, 1D and 2D Efficiencies, and Profiles to `Core`. Add converters to ROOT histograms to the ROOT plugin.

--- END COMMIT MESSAGE ---

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -106,3 +106,11 @@ repos:
       require_serial: true
       args: [--indent=4]
       exclude: ^.zenodo.json$
+
+    - id: sync_citation_metadata
+      name: sync citation metadata
+      entry: CI/sync_citation_metadata.py
+      language: python
+      files: ^(CITATION\.cff|\.zenodo\.json|AUTHORS)$
+      additional_dependencies: ["typer", "rich", "pyyaml"]
+      pass_filenames: false
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,127 +1,174 @@
 {
   "description": "Experiment-independent toolkit for (charged) particle track reconstruction in (high energy) physics experiments implemented in modern C++ ",
   "license": "MPL-2.0",
-  "title": "acts-project/acts: v11.0.0",
-  "version": "v11.0.0",
+  "title": "acts-project/acts: vv44.2.0",
+  "version": "v44.2.0",
   "upload_type": "software",
   "creators": [
     {
-      "affiliation": "CERN",
-      "name": "Andreas Salzburger"
-    },
-    {
-      "affiliation": "CERN",
-      "name": "Paul Gessinger"
-    },
-    {
-      "affiliation": "CERN / TUM Muenchen",
-      "name": "Fabian Klimpel"
-    },
-    {
-      "affiliation": "CERN",
-      "name": "Moritz Kiehn"
-    },
-    {
-      "affiliation": "CERN / JGU Mainz",
-      "name": "Bastian Schlag"
-    },
-    {
-      "affiliation": "Universite Paris-Saclay / CNRS/IN2P3",
-      "name": "Hadrien G."
-    },
-    {
-      "affiliation": "University of Massachussets",
-      "name": "Robert Langenberg"
-    },
-    {
-      "affiliation": "CERN",
-      "name": "Corentin Allaire"
-    },
-    {
-      "affiliation": "DESY Hamburg",
+      "affiliation": "Zhengzhou University",
       "name": "Xiaocong Ai"
     },
     {
-      "affiliation": "Universitaet Regensburg",
-      "name": "Benjamin Huth"
+      "affiliation": "Rutherford Appleton Laboratory",
+      "name": "Tim Adye"
     },
     {
-      "affiliation": "UC Berkeley",
-      "name": "Louis-Guillaume Gagnon"
+      "affiliation": "",
+      "name": "Guilherme Almeida",
+      "orcid": "0000-0002-4257-4278"
+    },
+    {
+      "affiliation": "IJCLab",
+      "name": "Corentin Allaire"
+    },
+    {
+      "affiliation": "CERN, Aristotle University of Thessaloniki",
+      "name": "Dimitra Amperiadou"
+    },
+    {
+      "affiliation": "AGH University of Krakow",
+      "name": "Tomasz Bold"
+    },
+    {
+      "affiliation": "CERN",
+      "name": "Pierfrancesco Butti"
     },
     {
       "affiliation": "CERN",
       "name": "Noemi Calace"
     },
     {
-      "affiliation": "CERN",
-      "name": "Attila Krasznahorkay"
+      "affiliation": "Annecy LAPP",
+      "name": "Jeremy Couthures"
     },
     {
-      "affiliation": "UC Berkeley",
-      "name": "Beomki Yeo"
+      "affiliation": "CERN, University of Florence",
+      "name": "Lorenzo Damenti"
     },
     {
-      "affiliation": "LBNL",
-      "name": "Charles Leggett"
-    },
-    {
-      "affiliation": "Eotvos Lorand University",
-      "name": "Czirkos Angela"
-    },
-    {
-      "affiliation": "UC Berkeley",
-      "name": "Tomohiro Yamazaki"
+      "affiliation": "University of Manitoba",
+      "name": "Wouter Deconinck"
     },
     {
       "affiliation": "CERN",
-      "name": "Joana Niermann"
+      "name": "Davide Di Croce"
+    },
+    {
+      "affiliation": "CERN, JGU Mainz",
+      "name": "Doga Elitez"
+    },
+    {
+      "affiliation": "Brandeis University",
+      "name": "Luis Falda Coelho",
+      "orcid": "0000-0002-2298-3605"
     },
     {
       "affiliation": "UC Berkeley",
-      "name": "Irina Ene"
-    },
-    {
-      "affiliation": "Oak Ridge National Laboratory",
-      "name": "Joe Osborn"
+      "name": "Louis-Guillaume Gagnon"
     },
     {
       "affiliation": "Stanford University",
       "name": "Rocky Bala Garg"
     },
     {
-      "affiliation": "CERN / TU Wien",
-      "name": "Andreas Stefl"
+      "affiliation": "CERN",
+      "name": "Paul Gessinger"
     },
     {
-      "affiliation": "CERN / University of Coimbra",
-      "name": "Luis Falda Coelho",
-      "orcid": "0000-0002-2298-3605"
+      "affiliation": "DESY",
+      "name": "Maximilian Goblirsch-Kolb"
     },
     {
-      "affiliation": "CERN / University of Graz",
+      "affiliation": "IJCLab",
+      "name": "Hadrien Grasland"
+    },
+    {
+      "affiliation": "LBNL",
+      "name": "Heather Gray"
+    },
+    {
+      "affiliation": "",
+      "name": "Christian Gumpert"
+    },
+    {
+      "affiliation": "University of Massachusetts, Amherst",
+      "name": "Matthew Harris"
+    },
+    {
+      "affiliation": "",
+      "name": "Julia Hrdinka"
+    },
+    {
+      "affiliation": "CERN",
+      "name": "Benjamin Huth"
+    },
+    {
+      "affiliation": "CERN",
+      "name": "Johannes Junggeburth"
+    },
+    {
+      "affiliation": "",
+      "name": "Moritz Kiehn"
+    },
+    {
+      "affiliation": "",
+      "name": "Fabian Klimpel"
+    },
+    {
+      "affiliation": "",
+      "name": "Robert Langenberg"
+    },
+    {
+      "affiliation": "CERN",
+      "name": "Leonardo Monaco"
+    },
+    {
+      "affiliation": "CERN",
+      "name": "Joana Niermann"
+    },
+    {
+      "affiliation": "BNL",
+      "name": "Joe Osborn"
+    },
+    {
+      "affiliation": "CERN, University of Graz",
       "name": "Alexander J. Pfleger",
       "orcid": "0000-0001-5524-7738"
     },
     {
-      "affiliation": "University of Lisbon",
-      "name": "Guilherme Almeida",
-      "orcid": "0000-0002-4257-4278"
-    },
-    {
-      "affiliation": "CERN / University of Amsterdam",
-      "name": "Stephen Nicholas Swatman",
-      "orcid": "0000-0002-3747-3229"
-    },
-    {
-      "affiliation": "CERN / TU Wien",
+      "affiliation": "",
       "name": "Felix Russo",
       "orcid": "0009-0005-8975-2245"
     },
     {
-      "affiliation": "UC Berkeley",
-      "name": "Carlo Varni",
-      "orcid": "0000-0001-6733-4310"
+      "affiliation": "CERN",
+      "name": "Andreas Salzburger"
+    },
+    {
+      "affiliation": "",
+      "name": "Bastian Schlag"
+    },
+    {
+      "affiliation": "Carleton University",
+      "name": "Patrick Scholer"
+    },
+    {
+      "affiliation": "CERN",
+      "name": "Andreas Stefl"
+    },
+    {
+      "affiliation": "CERN",
+      "name": "Stephen Nicholas Swatman",
+      "orcid": "0000-0002-3747-3229"
+    },
+    {
+      "affiliation": "Weizmann Institute of Science",
+      "name": "Roman Urmanov"
+    },
+    {
+      "affiliation": "AGH University of Krakow",
+      "name": "Carlo Varni"
     }
   ],
   "access_right": "open",
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,20 +1,20 @@
 The following people have contributed to the project (in alphabetical order):
 
-- Xiaocong Ai, Zhengzhou University
 - Tim Adye, Rutherford Appleton Laboratory
-- Dimitra Amperiadou, CERN, Aristotle University of Thessaloniki
+- Xiaocong Ai, Zhengzhou University
 - Corentin Allaire, IJCLab
+- Dimitra Amperiadou, CERN, Aristotle University of Thessaloniki
 - Tomasz Bold, AGH University of Krakow
 - Pierfrancesco Butti, CERN
 - Noemi Calace, CERN
+- Luis Falda Coelho, Brandeis University
 - Jeremy Couthures, Annecy LAPP
 - Lorenzo Damenti, CERN, University of Florence
 - Wouter Deconinck, University of Manitoba
 - Davide Di Croce, CERN
 - Doga Elitez, CERN, JGU Mainz
-- Luis Falda Coelho, Brandeis University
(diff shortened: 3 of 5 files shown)


Core/include/Acts/Utilities/Histogram.hpp
```diff
@@ -0,0 +1,295 @@
...
+using BoostHist2D = decltype(boost::histogram::make_histogram(
+    std::declval<BoostVariableAxis>(), std::declval<BoostVariableAxis>()));
+
+using BoostProfileHist =
+    decltype(boost::histogram::make_profile(std::declval<BoostVariableAxis>()));
+
+}  // namespace detail
+
+/// @brief Nested binning struct for booking plots
+class HistBinning {
```
asalzburger: Can we use the `Acts::ProtoAxis` for this ? 
It does not have a logarithmic constructor, but that should be easy to add - and we would already have python bindings for this.

Core/include/Acts/Utilities/Histogram.hpp
```diff
@@ -0,0 +1,295 @@
...
+  /// @param name Histogram name
+  /// @param title Histogram title
+  /// @param binning Binning specification
+  Efficiency1D(std::string name, std::string title, const HistBinning& binning);
+
+  /// Fill efficiency histogram
+  ///
+  /// @param value Value to fill
+  /// @param passed Whether the event passed selection
+  void fill(double value, bool passed);
```
asalzburger: Probably renaming to "accepted" instead of passed ?

Core/include/Acts/Utilities/Histogram.hpp
```diff
@@ -0,0 +1,295 @@
...
+  const std::string& name() const { return m_name; }
+
+  /// Get histogram title
+  const std::string& title() const { return m_title; }
+
+  /// Get axis title
+  const std::string& axisTitle() const { return m_axisTitle; }
+
+  /// Access to passed histogram (for converters and tests)
+  const auto& passedHistogram() const { return m_passed; }
```
asalzburger: Same here.

Core/src/Utilities/Histogram.cpp
```diff
@@ -0,0 +1,145 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "Acts/Utilities/Histogram.hpp"
+
+#include <cmath>
+#include <vector>
+
+namespace Acts {
+
+HistBinning HistBinning::Uniform(std::string title, std::size_t bins,
```
asalzburger: Again, I wonder if we can use the existing `ProtoAxis` mechanism here, and hence we would get a single source Axis description for Grids and Histograms.

Plugins/Root/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ acts_add_library(
     src/TGeoParser.cpp
     src/TGeoPrimitivesHelper.cpp
     src/TGeoSurfaceConverter.cpp
+    src/HistogramToRootConverter.cpp
```
asalzburger: Since this is in the Root directory, it could only be called `HistogramConverter` - would allow to add the other direction converters if needed later.

Plugins/Root/src/HistogramToRootConverter.cpp
```diff
@@ -0,0 +1,277 @@
...
+  const auto& bh = boostHist.histogram();
+  const auto& axis = bh.axis(0);
+
+  // Extract bin edges from boost histogram axis
+  std::vector<double> edges;
+  edges.reserve(axis.size() + 1);
+  for (int i = 0; i < axis.size(); ++i) {
+    edges.push_back(axis.bin(i).lower());
+  }
+  edges.push_back(axis.bin(axis.size() - 1).upper());
```
asalzburger: You could factorize the edge conversion into a helper function at it appears in several places.

Plugins/Root/src/HistogramConverter.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+  return rootHist;
+}
+
+TH2F* toRoot(const Histogram2D& boostHist) {
+  const auto& bh = boostHist.histogram();
+  const auto& xAxis = bh.axis(0);
+  const auto& yAxis = bh.axis(1);
+
+  // Extract bin edges from X axis
```
asalzburger, line 71: The following lines can use the same function, if you encapsulate it in a helper function.

Plugins/Root/src/HistogramConverter.cpp
```diff
@@ -0,0 +1,277 @@
...
+    // - fBinEntries[bin] = sum of w        = count
+    // - fSumw2[bin]      = sum of (w*x)^2  = sum of x^2
+    // - fBinSumw2[bin]   = sum of w^2      = count
+
+    double sum = mean * count;
+
+    // To compute sum of y^2 from sample variance s^2:
+    // s^2 = Sum((x - μ)^2) / (n-1) = (Sum(x^2) - n*µ^2) / (n-1)
+    // Sum(x^2) = (n-1) * s^2 + n*µ^2
+    double sum_of_squares = (count - 1.0) * variance + count * mean * mean;
```
asalzburger: CamelCase coding standard would by `SumOfSquares`

Plugins/Root/include/ActsPlugins/Root/HistogramConverter.hpp
```diff
@@ -0,0 +1,67 @@
...
+namespace ActsPlugins {
+
+/// Convert Histogram1D to ROOT TH1F
+///
+/// Creates a new ROOT TH1F histogram with the same binning, content, and
+/// errors as the input boost histogram. The histogram uses Sumw2() for proper
+/// error tracking with weighted fills.
+///
+/// @param boostHist The boost histogram to convert
+/// @return Raw pointer to new TH1F (caller owns and must delete)
```
asalzburger, line 26: I know it's a bit tricky with ROOT, but shall we do `std::unique_ptr` here ?

asalzburger (changes requested):
Really like this!

Two main remarks:
- probably we could use the existing `ProtoAxis` (or an extended version for this) in order not to have yet another Binning description
- the converters could return `unique_ptr` so it's clear what the ownership does

And then, there are some comments within the code.

Core/include/Acts/Utilities/Histogram.hpp
```diff
@@ -0,0 +1,295 @@
...
+  const std::string& name() const { return m_name; }
+
+  /// Get histogram title
+  const std::string& title() const { return m_title; }
+
+  /// Get axis title
+  const std::string& axisTitle() const { return m_axisTitle; }
+
+  /// Direct access to boost::histogram (for converters and tests)
+  const auto& histogram() const { return m_hist; }
```
andiwand:
```suggestion
  const detail::BoostHist1D& histogram() const { return m_hist; }
```
to me `auto` seems harder to interpret as a human in this case

Core/include/Acts/Utilities/Histogram.hpp
```diff
@@ -0,0 +1,295 @@
...
+  /// Access to total histogram (for converters and tests)
+  const auto& totalHistogram() const { return m_total; }
+
+ private:
+  std::string m_name;
+  std::string m_title;
+  std::string m_xAxisTitle;
+  std::string m_yAxisTitle;
+
+  detail::BoostHist2D m_passed, m_total;
```
andiwand:
```suggestion
  detail::BoostHist2D m_passed;
  detail::BoostHist2D m_total;
```
all variables are equal!

Core/include/Acts/Utilities/Histogram.hpp
```diff
@@ -0,0 +1,295 @@
...
+
+  /// Access to total histogram (for converters and tests)
+  const auto& totalHistogram() const { return m_total; }
+
+ private:
+  std::string m_name;
+  std::string m_title;
+  std::string m_axisTitle;
+
+  detail::BoostHist1D m_passed, m_total;
```
andiwand:
```suggestion
  detail::BoostHist1D m_passed;
  detail::BoostHist1D m_total;
```

Core/include/Acts/Utilities/Histogram.hpp
andiwand: same below a couple of times

Core/include/Acts/Utilities/Histogram.hpp
benjaminhuth: Ah nice, I'll look at this. Just FYI, the `HistBinning` is a copy of the current `ActsExamples::PlotHelper::Binning` class, I did not write it from scratch.

andiwand: I believe this comes from here https://github.com/acts-project/acts/blob/0181d5da563c5c40d84d1bda9e7876f9d3434550/Examples/Framework/include/ActsExamples/Utilities/Helpers.hpp#L26-L67

Core/include/Acts/Utilities/Histogram.hpp
```diff
@@ -0,0 +1,295 @@
...
+using BoostHist1D = decltype(boost::histogram::make_histogram(
+    std::declval<BoostVariableAxis>()));
+
+using BoostHist2D = decltype(boost::histogram::make_histogram(
+    std::declval<BoostVariableAxis>(), std::declval<BoostVariableAxis>()));
+
+using BoostProfileHist =
+    decltype(boost::histogram::make_profile(std::declval<BoostVariableAxis>()));
+
+}  // namespace detail
```
andiwand: are these really a detail if we return them below?

benjaminhuth: Hmm good point. The idea is that they are only used by the converters, but should be fine to make them member of `Acts::`

Plugins/Root/include/ActsPlugins/Root/HistogramConverter.hpp
benjaminhuth: Hmm yeah in principle I would be very happy, I just never fully understood the ownership model of ROOT if this can cause problems somewhere...

Core/include/Acts/Utilities/Histogram.hpp
paulgessinger, line 1: I'm not sure I understand why we need the majority of these wrappers. 

- I think boost-histogram axes can already store names internally
- the bh- axes themselves already define the binning, so why do we need to duplicate that?
- there's built-in support for polymorphic axes (`axis::variant<axis::regular<>, axis::integer<>, ...>`) can we not use these?
- there's built-in support for dynamic-dimension histograms, can't we use those?
- Following our naming conventions, the names would be `BoostHist1`, `BoostHist2` etc.
- If at all possible, I'd like to avoid having this in core, but can do if we have to

Core/src/Utilities/Histogram.cpp
paulgessinger: We don't even need that. Boost-histogram does all of that for us, doesn't it?

Core/include/Acts/Utilities/Histogram.hpp
benjaminhuth: About having this in `Core`:
- If we want to have converters in the ROOT plugin, they either must live in `Core` or in the ROOT plugin
- If we want to use this in the examples, this must either live in `Core` or in `Examples`

The alternative would be to have everything in `Examples` under `Framework/Utilities` and `Io/Root` (which was my first approach)
@asalzburger argued that it would nice to have histograms in `Core` and converters in the plugin. I don't have a strong opinion on this, but see no issue of having this in `Core`.

The main point of the wrapper classes is:
- Consistency (Bundle together histograms for for `Efficiency1D`/`Efficiency2D`
- Restrict the API to make converters to ROOT easier (mainly for Profile)

I agree that for basic histograms, the wrappers don't add much. But I thought it makes sense to follow a consistent style here.

For the binning: This was copied from the previous examples code. See below for discussion if this can use `ProtoAxis`, but its also an option to just use `bh::axis`.

Plugins/Root/include/ActsPlugins/Root/HistogramConverter.hpp
benjaminhuth: Let's not do this for now.

Core/src/Utilities/Histogram.cpp
benjaminhuth: I agree we don't need a third binning/axis class. 

So the points to concern are those:
- If we use the boost regular axis, we can use uniform and log axes, but **no** variable binning.
- If we use the boost variable axis, we cannot use transforms/log and should provide convenience factories for log.
- We could also use `ProtoAxis` and convert it to boost variable axis.

Any opinions @paulgessinger @andiwand @asalzburger ?

Core/include/Acts/Utilities/Histogram.hpp
paulgessinger: I would strongly prefer to use the native types as much as possible, tbh. I'm not against having typedefs, if those make sense, but I don't think we should introduce wrappers that don't do anything, and especially not if they need to duplicate functionality that is in boost-histogram.

For sure, we shouldn't: 
- add API to build axis types, that then get unconditionally converted to boost variable axis types, and then implement logic to turn that back into root axis types
-  add axis metadata that can already be held in boost histogram.


My recommendation would be to add a central typedef for a `bh::axis::variant` that is our polymorphic axis type, and then try to base as much of the API, also for `Efficiency` around that. You can then add `Histogram1/2` wrappers that simply add a title + histo name, and maybe add additional functionality, but e.g. I don't see what we gain from aliasing `project_axis`.

benjaminhuth: Can you be a bit more concrete? Do I understand it correctly that you would prefer to have only have classes for Efficiencies? And would you prefer the code to live in `Examples/Framework` and `Examples/Io/Root`?

paulgessinger: - I won't insist on the location of the code, if you prefer having it in core, we'll do that
- I would prefer only adding "wrapper" code that adds functionality, and not abstracting or reimplementing functionality that already exists in boost-histogram.

asalzburger: The Binning definition from the axes is actually not new code - it lived in the Examples and allowed for python bindings and configuration.

Hence, when moving it to Core we should use the ProtoAxis to define it & we can dump yet another binning definition.

I think the wrappers mainly add name and title which we otherwise have to bookkeep separately as boost doesn't have this functionality.

asalzburger: The efficiency histogram is a wrapper of course because that doesn't exist in boost either, so the storage with accepted / total counting sounds reasonable to me.

paulgessinger: I agree, if we're deciding between `ProtoAxis` and another dedicated type, we should reuse `ProtoAxis`.


**But**: boost-histogram has axis definiton API, types + type erasure for them. What I'm saying is we should use those directly for the histogram API.

You're right about histogram level title, and I think the same is true for axis title (boost-histogram has axis labels)

asalzburger: So here's my suggestion:

- in Core & Plugins we use boost native API

- in Examples, where we really want to use the python bindings, we get rid of the ActsExamples::Binning and translate from ProtoAxis 

That would save us to introduce python bindings for boost types

paulgessinger: Generally seems sensible to me, but I'm not sure I get where the Python connection enters. For the monitoring histograms that we're writing, the binning is hard coded in c++. Is the plan to change this?

benjaminhuth: Actually, we can already modify the binning of the histograms via python as of today: https://github.com/acts-project/acts/blob/main/Python%2FExamples%2Fsrc%2Fplugins%2FRoot.cpp#L94

This is e.g. exposed in the config of the `TrackFinderPerformanceWriter`.

I think I used that for my thesis back then...

paulgessinger: Oh ok, I see. There is actually a first-class python interface for boost-histogram (I think that's actually the main reason why this project exists). Maybe there's a way to reuse that?

benjaminhuth: I would propose to defer the discussion on the python binding to the next step, since this PR only provides basic tools and not yet any connection to Examples.

Right now I have:
- Removed the `HistBinning` class. Core Wrappers use native boost variable axis
- Kept the wrappers as is otherwise

Can we proceed with this @asalzburger @paulgessinger ?

Core/src/Utilities/Histogram.cpp
asalzburger: For me, this would be fine.

Core/include/Acts/Utilities/Histogram.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+#include <boost/histogram.hpp>
+
+namespace Acts {
+
+/// @brief Boost axis type to use for histograms with metadata support
+using BoostVariableAxis = boost::histogram::axis::variable<double, std::string>;
+
+/// @brief Underlying Boost type for Histogram1D
+using BoostHist1D = decltype(boost::histogram::make_histogram(
```
paulgessinger:
```suggestion
using BoostHist1 = decltype(boost::histogram::make_histogram(
```

Core/include/Acts/Utilities/Histogram.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+#include <string>
+#include <vector>
+
+#include <boost/histogram.hpp>
+
+namespace Acts {
+
+/// @brief Boost axis type to use for histograms with metadata support
+using BoostVariableAxis = boost::histogram::axis::variable<double, std::string>;
```
paulgessinger, line 19: Are you still going to switch this to `axis::variant`?

Core/include/Acts/Utilities/Histogram.hpp
```diff
@@ -0,0 +1,269 @@
...
+  void fill(double value);
+
+  /// Get histogram name
+  const std::string& name() const { return m_name; }
+
+  /// Get histogram title
+  const std::string& title() const { return m_title; }
+
+  /// Get axis title from axis metadata
+  const std::string& axisTitle() const { return m_hist.axis(0).metadata(); }
```
paulgessinger: I'd suggest to just expose the axis as well. For the `axisTitle()` I suppose it's useful to have an additional alias like this.

```suggestion
  const std::string& xAxisTitle() const { return m_hist.axis(0).metadata(); }
```

Core/include/Acts/Utilities/Histogram.hpp
```diff
@@ -0,0 +1,269 @@
...
+  const std::string& title() const { return m_title; }
+
+  /// Get X-axis title from axis metadata
+  const std::string& xAxisTitle() const { return m_hist.axis(0).metadata(); }
+
+  /// Get Y-axis title from axis metadata
+  const std::string& yAxisTitle() const { return m_hist.axis(1).metadata(); }
+
+  /// Project the histogram onto x
+  Histogram1D projectionX() const;
```
paulgessinger:
```suggestion
  Histogram1D xProjection() const;
```
?

acts-project-service:
## :red_circle: Athena integration test results [9be78a98433e1dce7e9357af4cac86cd9edea51f]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67701979) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13773692)!

PR #4940, opened 2026-01-07
ci: Bump dependencies version to 19.0.0

paulgessinger:


paulgessinger:
@benjaminhuth can you have a look at the modulemap build failures?

benjaminhuth:
Hmm looks like there is a `#include <cmath>` missing in `CPU/src/TTree_hits.cpp` of ModuleMapGraph.
Fastest solution would be a patch, but I can also upstream the fix (but might take a couple of days for the turn-a-around)

paulgessinger:
@benjaminhuth I'm ok with a patch, I suppose we can apply it inside our cmake?

CI/bump_versions.py
```diff
@@ -48,13 +48,21 @@ class VersionBumper:
         "spack_container": re.compile(
             r"(ghcr\.io/acts-project/spack-container):(\d+\.\d+\.\d+)(_[a-z0-9._-]+)"
         ),
+        # Matches DEPENDENCY_TAG in two formats:
+        # 1. GitLab CI: DEPENDENCY_TAG: v18.0.0
+        # 2. GitHub Actions: DEPENDENCY_TAG: ... default: 'v18.0.0'
```
AJPfleger, line 53:
```suggestion
        # 1. GitLab CI: DEPENDENCY_TAG: v19.0.0
        # 2. GitHub Actions: DEPENDENCY_TAG: ... default: 'v19.0.0'
```
Not crucial, but might look more consistent.

CI/bump_versions.py
```diff
@@ -87,6 +95,9 @@ def find_versions(self, content: str, pattern_name: str) -> set[str]:
         elif pattern_name == "spack_container":
             # Return full version strings
             return {match[1] for match in matches}
+        elif pattern_name == "dependency_tag":
+            # Return full version strings (e.g., v18.0.0)
```
AJPfleger, line 99:
```suggestion
            # Return full version strings (e.g., v19.0.0)
```

CI/bump_versions.py
paulgessinger: But won't that essentially ask the script to update itself? That seems a bit weird, no?

AJPfleger: I am not sure if I understand it then. I thought these were just examples what the regex does. Therefore, I felt just taking the current version would look nice.

I also saw that we have `18` as an example a few lines above, so having everything on arbitrary `18` might also look good :)

paulgessinger: Since it's just an example, I'm not sure this needs to be updated.

CI/bump_versions.py
paulgessinger: Since it's just an example, I'm not sure this needs to be updated.

paulgessinger:
@andiwand output changes seem consistent with updated pythia8 version to me.

andiwand (approved):

acts-project-service:
## :red_circle: Athena integration test results [442a9c6c6205cf77423dd950562c227f5d2d91e1]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67581543) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13757532)!

PR #4951, opened 2026-01-12
fix: Attempt to fix possible div by zero

paulgessinger:


Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -107,8 +107,11 @@ constexpr T sumUpToN(const T N) {
 template <std::integral T>
 constexpr T factorial(const T upperN, const T lowerN = 1) {
   constexpr T one = 1;
-  const T& limit = std::max(one, lowerN);
-  return upperN >= limit ? upperN * factorial(upperN - 1, limit) : one;
+  if (upperN < one) {
+    return one;
+  }
+  const T limit = lowerN < one ? one : lowerN;
```
junggjo9:
```suggestion
  const T limit = std::max(lowerN, one);
```
Is easier comprehensible. Otherwise looks good

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -107,8 +107,11 @@ constexpr T sumUpToN(const T N) {
 template <std::integral T>
 constexpr T factorial(const T upperN, const T lowerN = 1) {
   constexpr T one = 1;
-  const T& limit = std::max(one, lowerN);
-  return upperN >= limit ? upperN * factorial(upperN - 1, limit) : one;
+  if (upperN < one) {
+    return one;
+  }
```
AJPfleger, line 112:
```suggestion
  if (upperN <= one) {
    return one;
  }
```
0! = 1! = 1

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -107,8 +107,11 @@ constexpr T sumUpToN(const T N) {
...
+  const T& limit = std::max(lowerN, one);
+  return upperN >= limit ? upperN * factorial(upperN - one, limit) : one;
```
AJPfleger, line 114:
```suggestion
  return upperN > limit ? upperN * factorial(upperN - one, limit) : limit;
```
The last element will always be the `limit`. e.g.:
with old method:
f(5,3) = 5 * f(4,3) = 5 * 4 * f(3,3) = 5 * 4 * 3 * f(2,3) = 5 * 4 * 3 * 1
vs suggestion:
f(5,3) = 5 * f(4,3) = 5 * 4 * f(3,3) = 5 * 4 * 3

This always saves 1 recursive call.

Core/include/Acts/Utilities/MathHelpers.hpp
AJPfleger, line 1: The modified factorial function is also lacking 2 safety mechanisms:
- `upperN` must not be negative. maybe make a check in case of `signed integral`?
- `upperN` < `lowerN` is not obviously defined/documented. Currently, it would return `1`. But it's not really what I would expect from this function. Omitting the lower terms equals to upperN! / (lowerN - 1)! which I would expect to `0` in that case.

AJPfleger: also `lowerN = 2` could be a better standard value to avoid further unnecessary ultiplications.

paulgessinger:
@AJPfleger @junggjo9 can one of you put together a more comprehensive solution? I'm mostly concerned with avoiding the division by zero.

junggjo9:
I'll open one

PR #4956, opened 2026-01-13
chore: Update `GbtsDataStorage` to ACTS coding standard

jpreston-cern:
This is the second in a series of PR's that aim to bring GBTS up to date with ACTS coding standards. 

This PR focuses on the comments made about the `GbtsDataStorage` files and aims to update them accordingly, this includes things like changing to appropriate containers, re structuring classes and changing to fixed size variables. Other related changes are also made to the files affected by the `GbtsDataStorage` changes.

--- END COMMIT MESSAGE ---

@timadye @andiwand

### DIFF
--- a/Core/include/Acts/EventData/AnyTrackProxy.hpp
+++ b/Core/include/Acts/EventData/AnyTrackProxy.hpp
@@ -91,12 +91,25 @@ class TrackHandlerMutableBase : public TrackHandlerConstBase {
                              HashedString key) const = 0;
 };
 
+template <typename container_t>
+struct TrackHandlerTraits {
+  using Container = std::remove_const_t<container_t>;
+  static constexpr bool ReadOnly =
+      std::is_const_v<container_t> || Container::ReadOnly;
+};
+
 /// Concrete handler for a specific track container type
 /// This templated class provides static instances that implement the virtual
 /// interface for a specific track container type. The static instance approach
 /// avoids heap allocation per handle.
+template <typename container_t,
+          bool read_only = TrackHandlerTraits<container_t>::ReadOnly>
+class TrackHandler;
+
 template <typename container_t>
-class TrackHandler final : public TrackHandlerMutableBase {
+class TrackHandler<container_t, true> final : public TrackHandlerConstBase {
+  using ContainerType = typename TrackHandlerTraits<container_t>::Container;
+
  public:
   /// Get the singleton instance for this track container type
   static const TrackHandler& instance() {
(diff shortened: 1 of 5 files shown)


jpreston-cern:
just to note, i did all of the changes to fixed size variables for int, unsigned int, char, short ect at the end of all my code inspections so they will be fixed at the end (unless it would make more sense do that now)

Core/include/Acts/Seeding/GbtsDataStorage.hpp
```diff
@@ -14,34 +14,54 @@
...
+  void setY(float v) noexcept { m_y = v; }
+  void setZ(float v) noexcept { m_z = v; }
+  void setR(float v) noexcept { m_r = v; }
+  void setPhi(float v) noexcept { m_phi = v; }
+
+  void setLayer(std::uint16_t v) noexcept { m_layer = v; }
+  void setSpIdx(std::uint32_t v) noexcept { m_idx = v; }
+
+  void setPixelClusterWidth(float v) noexcept { m_pcw = v; }
+  void setLocalPositionY(float v) noexcept { m_locPosY = v; }
```
andiwand: alternatively you can also return a mutable reference and assign the value on the caller site:
```c++
float& x() noexcept { return m_x; }
float& y() noexcept { return m_y; }
float& z() noexcept { return m_z; }
```
```c++
.x() = 0;
.y() = 0;
.z() = 0;
```

Core/include/Acts/Seeding/GbtsDataStorage.hpp
```diff
@@ -121,24 +139,26 @@ class GbtsEdge {
     }
   };
 
+  GbtsEdge() = default;
+
   GbtsEdge(const GbtsNode* n1, const GbtsNode* n2, float p1, float p2, float p3)
```
andiwand, line 141: can the nodes ever be `nullptr`? if not you could change this to `const GbtsNode&`

Core/include/Acts/Seeding/GbtsDataStorage.hpp
```diff
@@ -121,24 +139,26 @@ class GbtsEdge {
...
   const GbtsNode* m_n1{nullptr};
   const GbtsNode* m_n2{nullptr};
 
-  signed char m_level{-1}, m_next{-1};
+  std::int8_t m_level{-1};
+  std::int8_t m_next{-1};
 
-  unsigned char m_nNei{0};
-  float m_p[3]{};
+  std::int8_t m_nNei{0};
```
andiwand:
```suggestion
  std::uint8_t m_nNei{0};
```
in case this should stay unsigned `std::uint8_t` would be the right type

Core/include/Acts/Seeding/GbtsDataStorage.hpp
```diff
@@ -83,12 +101,12 @@ class GbtsDataStorage {
...
-  ~GbtsDataStorage();
+                           GbtsMLLookupTable mlLUT);
 
-  int loadPixelGraphNodes(short layerIndex, const std::vector<GbtsNode>& coll,
-                          bool useML);
-  int loadStripGraphNodes(short layerIndex, const std::vector<GbtsNode>& coll);
+  int loadPixelGraphNodes(short layerIndex,
+                          const std::span<const GbtsNode> coll, bool useML);
+  int loadStripGraphNodes(short layerIndex,
+                          const std::span<const GbtsNode> coll);
```
andiwand:
```suggestion
  int loadPixelGraphNodes(short layerIndex,
                          std::span<const GbtsNode> coll, bool useML);
  int loadStripGraphNodes(short layerIndex,
                          std::span<const GbtsNode> coll);
```

I would either go with a mutable value type or a const reference: `std::span<const GbtsNode>` or `const std::span<const GbtsNode>&`

`std::span` is a small object so by value should be fine

Core/src/Seeding/GbtsDataStorage.cpp
```diff
@@ -103,17 +94,15 @@ void GbtsEtaBin::generatePhiIndexing(float dphi) {
...
   // parse the look up table if useML is true
 
   m_etaBins.resize(m_geo->num_bins());
 }
 
-GbtsDataStorage::~GbtsDataStorage() = default;
-
 int GbtsDataStorage::loadPixelGraphNodes(short layerIndex,
-                                         const std::vector<GbtsNode>& coll,
+                                         const std::span<const GbtsNode> coll,
```
andiwand:
```suggestion
                                         std::span<const GbtsNode> coll,
```

Core/src/Seeding/GbtsDataStorage.cpp
```diff
@@ -151,7 +140,7 @@ int GbtsDataStorage::loadPixelGraphNodes(short layerIndex,
 }
 
 int GbtsDataStorage::loadStripGraphNodes(short layerIndex,
-                                         const std::vector<GbtsNode>& coll) {
+                                         const std::span<const GbtsNode> coll) {
```
andiwand:
```suggestion
                                         std::span<const GbtsNode> coll) {
```

Core/include/Acts/Seeding/GbtsDataStorage.hpp
jpreston-cern: in this case GbtsEdge is used in a vector within `buildTheGraph` in `SeedFinderGbts` which is first reserved meaning we the pointers used arnt assigned to anything to start with (that happens later) and as we cant have an epty ref i think thats why they are used. the  `GbtsNode` objects itself in `GbtsDataStorage` is a pointer as we have the big node storage object outside of the instance of `GbtsDataStorage` which we then use pointers to sort these nodes into eta bins (in vectors which again i think is why we don't use) which further down the line are then pushed into the `edgeStorage` vector of `GbtsEdge`.

andiwand (dismissed):

acts-project-service:
## :red_circle: Athena integration test results [c2931c96cfa639bf58fd8ac652caeb5af4d4b864]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67852699) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13793288)!

PR #4962, opened 2026-01-13
refactor: Avoid that factorial can return negative values

junggjo9:
- Rename the capped factorial to product
- Introduce a factorial function with one argument only
- Constrain the domain of the product/factorial/binomial to be unsigned integral values
- Fix bug in the run-time evaluation of the Legendre polynomial derivatives
- Add test comparing the 0-th, first and second derivative of runTime & compile time Legendre polynomials

--- END COMMIT MESSAGE ---

### DIFF
--- a/Core/include/Acts/Propagator/ActorConcepts.hpp
+++ b/Core/include/Acts/Propagator/ActorConcepts.hpp
@@ -30,7 +30,7 @@ concept ActorHasActWithoutResult = requires(
     const actor_t& a, propagator_state_t& state, const stepper_t& stepper,
     const navigator_t& navigator, Args&&... args) {
   {
-    a.act(state, stepper, navigator, std::move<Args>(args)...)
+    a.act(state, stepper, navigator, std::forward<Args>(args)...)
   } -> std::same_as<Result<void>>;
 };
 
@@ -42,7 +42,7 @@ concept ActorHasActWithResult =
              const stepper_t& stepper, const navigator_t& navigator,
              typename actor_t::result_type& result, Args&&... args) {
       {
-        a.act(state, stepper, navigator, result, std::move<Args>(args)...)
+        a.act(state, stepper, navigator, result, std::forward<Args>(args)...)
       } -> std::same_as<Result<void>>;
     };
 
@@ -54,7 +54,7 @@ concept ActorHasOldVoidInterface =
              const stepper_t& stepper, const navigator_t& navigator,
              Args&&... args) {
       {
-        a.act(state, stepper, navigator, std::move<Args>(args)...)
+        a.act(state, stepper, navigator, std::forward<Args>(args)...)
       } -> std::same_as<void>;
     } ||
     // Check with result parameter (for actors with result_type)
@@ -63,7 +63,7 @@ concept ActorHasOldVoidInterface =
               const stepper_t& stepper, const navigator_t& navigator,
               typename actor_t::result_type& result, Args&&... args) {
        {
-         a.act(state, stepper, navigator, result, std::move<Args>(args)...)
+         a.act(state, stepper, navigator, result, std::forward<Args>(args)...)
        } -> std::same_as<void>;
      });
 
@@ -81,7 +81,7 @@ concept ActorHasAbortWithoutResult = requires(
     const actor_t& a, propagator_state_t& state, const stepper_t& stepper,
     const navigator_t& navigator, Args&&... args) {
   {
-    a.checkAbort(state, stepper, navigator, std::move<Args>(args)...)
+    a.checkAbort(state, stepper, navigator, std::forward<Args>(args)...)
   } -> std::same_as<bool>;
 };
 
@@ -94,7 +94,7 @@ concept ActorHasAbortWithResult =
              typename actor_t::result_type& result, Args&&... args) {
       {
         a.checkAbort(state, stepper, navigator, result,
-                     std::move<Args>(args)...)
+                     std::forward<Args>(args)...)
       } -> std::same_as<bool>;
     };
 


Tests/UnitTests/Core/Utilities/MathHelpersTests.cpp
```diff
@@ -43,6 +43,24 @@ BOOST_DATA_TEST_CASE(fastHypot, expDist ^ expDist ^ bdata::xrange(100), xExp,
...
+
+  static_assert(Acts::factorial(2u) == 2u);
+  static_assert(Acts::factorial(2) == 2);
+
+  for (unsigned k = 3; k <= 20; ++k) {
+    BOOST_CHECK_EQUAL(Acts::factorial(k), k * Acts::factorial(k - 1u));
+  }
+  for (int k = -20; k <= 0; ++k) {
+    BOOST_CHECK_EQUAL(Acts::factorial(k), 1);
+  }
```
AJPfleger: Factorial for negative numbers is not defined and should fail.

Tests/UnitTests/Core/Utilities/MathHelpersTests.cpp
AJPfleger, line 1: We also should test for some cases `lowerN != 1`.
I think:
- `lowerN < 1` should fail
- `lowerN > upperN` should fail
- `factorial(a, b) == factorial(a) / factorial(b-1)`

AJPfleger (changes requested):
Let's discuss the function implementation later :)

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -107,7 +107,8 @@ constexpr T sumUpToN(const T N) {
 template <std::integral T>
```
andiwand:
```suggestion
template <std::unsigned_integral T>
```

AJPfleger: I was wondering about that too. But could a call with a reasonable e.g. `int` still find the function?

andiwand: hm good point I guess `T = int` will not be generated

Tests/UnitTests/Core/Utilities/MathHelpersTests.cpp
junggjo9: We can just non-analytically extend the gamma function

gamma(z) = gamma (z) if z != N- otherwise 1

AJPfleger: I don't think, I follow this logic. The Γ-function is a non-integer extension of the factorial with poles at negative integers. So far, no new information for our integer problem. Close to the non-integer numbers the Γ-function diverges. At the poles, no pointwise limit exists.

Core/include/Acts/Utilities/MathHelpers.hpp
junggjo9: changed to unsigned

Tests/UnitTests/Core/Utilities/MathHelpersTests.cpp
junggjo9: lowerN > upperN is defined to return one. lowerN == 0 should be allowed because 0 factorial is one. Third test case has been implemented.

AJPfleger: I am not sure if I am fully on board yet with the `lowerN == 0` case. I think it is not fully clear from description and implementation what it means:
- if `lowerN` is number of elements to remove, but then the name is a bit confusing and the standard value might need to be 0
- if `lowerN` is the starting value that factorial(upperN, lowerN) = lowerN * (lowerN + 1) * ... * (upperN - 1) * upperN, then setting it to 0 would result in 0. (This is, what the first test kinda assumes)

Same for `lowerN > upperN`:
- we cannot really remove more elements than we have
- actually not sure what in this case happens. maybe swap limits? or set to 0 because of the integer division upperN! // (lowerN - 1)!?

third test looks good 👍

junggjo9: If lowerN > upperN then the function returns a one

AJPfleger: I was wondering about the use-case. What do you expect the function actually to do? then we could try to formulate a more precise one around this case, since we are kinda away from a real factorial 😊
Then we could keep the factorial in without the unclear `lowerN` and redefine a more efficient one based one the pure factorial.

Tests/UnitTests/Core/Utilities/MathHelpersTests.cpp
junggjo9: Even if there's no pointwise limit, then the function can be arbitrarily extended to random values. Of course all properties of continuity, differentiablity etc are lost but theoretically possible. Anyhow, the input value needs to be an unsigned value now.

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -99,16 +99,25 @@ template <std::integral T>
...
+/// @param upperN: Upper range of the product calculation
+/// @param lowerN: Lower range of the product calculation
 /// @return Factorial result
-template <std::integral T>
-constexpr T factorial(const T upperN, const T lowerN = 1) {
-  constexpr T one = 1;
-  const T& limit = std::max(one, lowerN);
-  return upperN >= limit ? upperN * factorial(upperN - 1, limit) : one;
+template <std::unsigned_integral T>
+constexpr T product(const T upperN, const T lowerN) {
```
AJPfleger:
```suggestion
constexpr T product(const T lowerN, const T upperN) {
```
I would at least expect it in this order as "from-to"

Core/include/Acts/Utilities/MathHelpers.hpp
```diff
@@ -99,16 +99,25 @@ template <std::integral T>
...
+  T value{1};
+  for (T iter = std::max(static_cast<T>(2), lowerN); iter <= upperN; ++iter) {
```
AJPfleger, line 116:
```suggestion
  for (T iter = lowerN; iter <= upperN; ++iter) {
```
I would always start from `lowerN`. Not much difference if we do the `max` here or the `*1` later on.
But we get the benefit, that we result into `0` in case `lowerN == 0`, which is what I would expect from `0*1*...*N`

AJPfleger (changes requested):
Split into `product` and `factorial` looks good, just the 2 small comments.

Core/include/Acts/Utilities/MathHelpers.hpp
junggjo9: Nope let's leave it as it is for now

Core/include/Acts/Utilities/MathHelpers.hpp
junggjo9: Nope let's leave it as it is for now

AJPfleger: Then please disallow calling this function with `upperN = 0` or `lowerN = 0` or directly return `0` in such a case.

Core/include/Acts/Utilities/MathHelpers.hpp
AJPfleger: Please.

Core/include/Acts/Utilities/detail/Polynomials.hpp
```diff
@@ -103,9 +103,11 @@ constexpr std::array<double, L + 1> coefficients() {
 /// @param d: Order of the derivative
 constexpr double evaluate(const double x, const unsigned l, unsigned d = 0u) {
   double sum{0.};
-  for (unsigned k = l % 2; k + d <= l; k += 2u) {
-    sum += pow(x, k - d) * coeff(l, k) *
-           (d > 0u ? factorial(k - d, k - d + 1u) : 1u);
+  for (unsigned k = l % 2; k <= l; k += 2u) {
+    if (k >= d) {
+      sum +=
+          pow(x, k - d) * coeff(l, k) * (d > 0u ? product(1u + k - d, k) : 1u);
```
AJPfleger, line 109: That one, i didn't verify but it looks reasonable. The previous one was obviously wrong. Let's see :)

AJPfleger (approved):
Thanks for adding the documentation and handling the edge-cases.

acts-project-service:
## :red_circle: Athena integration test results [e875a2e779cb60a38298b2b5af837fd510d1d3f5]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/68128502) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13832250)!

PR #4963, opened 2026-01-14
feat: Introduce infrastructure to make Volumes alignable

junggjo9:
- Introduce the VolumePlacementBase interface class to provide an API hook to the experiment to align the volumes synchronously with the volumes.
- Augment the volumeBounds with the boundarySurfaces() method which is connecting the oriented surfaces to the alignment

--- END COMMIT MESSAGE ---

Tagging: @andiwand, @asalzburger, @paulgessinger


Core/include/Acts/Geometry/Volume.hpp
```diff
@@ -39,21 +41,33 @@ class Volume : public GeometryObject {
...
+  /// @param positioner: Reference to the object aligning the volume
+  /// @param volbounds is the volume boundary definitions
+  explicit Volume(VolumePlacementBase& positioner,
+                  std::shared_ptr<VolumeBounds> volbounds) noexcept;
 
-  /// Copy Constructor - with optional shift
-  ///
+  /// @brief Copy Constructor
+  /// @param vol is the source volume for the copy
+  explicit Volume(const Volume& vol) noexcept;
```
junggjo9:
```suggestion
  Volume(const Volume& vol) noexcept;
```

Core/include/Acts/Geometry/detail/PortalPlacement.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include "Acts/Surfaces/RegularSurface.hpp"
+
+namespace Acts {
+class VolumePlacementBase;
+
+namespace detail {
+/// @brief Implementation of the `SurfacePlacementBase` to synchronize the alignment
+///        of an alignable volume with the alignment of the boundary surfaces
+///        associated with the Volume.
+class PortalPlacement : public DetectorElementBase {
```
paulgessinger: I'm wondering if we shouldn't even make this header hidden entirely, as we don't want it to be used by clients.

Core/include/Acts/Geometry/detail/PortalPlacement.hpp
```diff
@@ -0,0 +1,70 @@
...
+  ///        the portal placement
+  friend class Acts::VolumePlacementBase;
+
+  /// @brief Returns the transform to switch from the portal reference
+  ///       frame into the experiment's global frame taking the alignment
+  ///       correction into account
+  /// @param gctx The current geometry context object, e.g. alignment
+  const Transform3& localToGlobalTransform(
+      const GeometryContext& gctx) const final;
+  /// @brief Returns the const reference to portal surface connected with
```
paulgessinger: NIT

```suggestion

  /// @brief Returns the const reference to portal surface connected with
```

Core/include/Acts/Geometry/Volume.hpp
```diff
@@ -180,19 +194,23 @@ class Volume : public GeometryObject {
...
                  const ViewConfig& viewConfig) const;
 
- protected:
+  VolumePlacementBase* volumePositioner();
+  const VolumePlacementBase* volumePositioner() const;
+
+ private:
   /// @brief Transform matrix that positions the volume in 3D space
-  Transform3 m_transform;
+  std::unique_ptr<const Transform3> m_transform{};
```
paulgessinger: Is this required? We were originally doing this, but got rid of it. If it's meant to communicate the fact that it's optional, I would either use `std::optional` or use the presence of the placement like in the surfaces rather than the presence of a transform.

Core/include/Acts/Geometry/VolumeBounds.hpp
```diff
@@ -113,6 +114,13 @@ class VolumeBounds {
   virtual std::vector<OrientedSurface> orientedSurfaces(
       const Transform3& transform = Transform3::Identity()) const = 0;
 
+  /// @brief Returns the vector of orientedSurfaces but connects the
```
paulgessinger:
```suggestion
  /// @brief Returns the vector of orientedSurfaces and connects the
```

Core/src/Geometry/VolumeBounds.cpp
```diff
@@ -43,3 +44,25 @@ std::ostream& Acts::operator<<(std::ostream& sl,
...
+  if (parentVolume.volumePositioner() == nullptr) {
+    return orientedSurfaces(*parentVolume.m_transform);
+  }
+  std::vector<OrientedSurface> portalSurfaces =
+      orientedSurfaces(Acts::Transform3::Identity());
+  // It is safe to construct a default geometry context here as the oriented
+  // surfaces have their own transform
+  const GeometryContext gctx = GeometryContext::dangerouslyDefaultConstruct();
+  for (std::size_t faceIdx = 0lu; faceIdx < portalSurfaces.size(); ++faceIdx) {
+    std::shared_ptr<RegularSurface>& alignMe = portalSurfaces[faceIdx].surface;
```
paulgessinger:
```suggestion
    std::shared_ptr<RegularSurface>& surface = portalSurfaces[faceIdx].surface;
```

Core/include/Acts/Geometry/Volume.hpp
```diff
@@ -180,19 +194,23 @@ class Volume : public GeometryObject {
   void visualize(IVisualization3D& helper, const GeometryContext& gctx,
                  const ViewConfig& viewConfig) const;
 
- protected:
+  VolumePlacementBase* volumePositioner();
+  const VolumePlacementBase* volumePositioner() const;
```
paulgessinger:
```suggestion
  VolumePlacementBase* volumePlacement();
  const VolumePlacementBase* volumePlacement() const;
```

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,93 @@
...
+  /// @brief Default constructor
+  VolumePlacementBase();
+  /// @brief Default destructor
+  virtual ~VolumePlacementBase() = default;
+  /// @brief Attaches a SurfacePlacementBase to the volume's portal surface
+  ///        to synchronize the portal alignment with the volume alignment
+  ///        Internally, a `PortalPlacement` object is instantiated and
+  ///        connected with the surface. As the number of calls to create
+  ///        the portal surfaces is a priori unrestricted, the ownership
+  ///        of the portal surface is transferred
```
paulgessinger: Ownership will remain shared, even if you pass in the shared pointer as an rvalue ref. I assume this is about the fact that a portal cannot be added to more than one volume like this? Can this be an explicit check + exception maybe?

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,93 @@
...
+  /// @param portalToVolTrf: Transform to switch from the portal's frame into the volume's frame
+  /// @param surface: Mutable reference to the portal surface that is to be aligned
+  std::shared_ptr<RegularSurface> makePortalAlignable(
+      const std::size_t portalIdx, const Transform3& portalToVolTrf,
+      std::shared_ptr<RegularSurface>&& surface);
+  /// @brief Returns the const pointer to the `SurfacePlacementBase` object
+  ///        aligning the i-th portal (May be nullptr if index exceeds the
+  ///        number of portals)
+  /// @param portalIdx: Internal index of the portal surface [0 - number of portals)
+  const detail::PortalPlacement* portalPlacement(
```
paulgessinger: Since this returns a `detail` type, can we make this private?

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,93 @@
...
+  ///        a const-mutable lazy population, this method has to be called
+  ///        at every change of the alignment constants
+  /// @param gctx: Mutable reference of the GeometryContext to populate
+  void populateContextWithPortals(GeometryContext& gctx) const;
+
+ protected:
+  /// @brief Invoke the backend cache to store the localToGlobalTransform
+  ///        for the given portal in the geometry Context
+  /// @param gctx: Reference to the mutable GeometryContext where the
+  ///              tarnsform for the portal shall be stored
```
paulgessinger:
```suggestion
  ///              transform for the portal shall be stored
```

Core/src/Geometry/Volume.cpp
```diff
@@ -19,19 +19,26 @@ namespace Acts {
...
-      m_volumeBounds(vol.m_volumeBounds) {}
+Volume::Volume(VolumePlacementBase& positioner,
+               std::shared_ptr<VolumeBounds> volbounds) noexcept
+    : GeometryObject{},
+      m_volumeBounds{std::move(volbounds)},
+      m_placement{&positioner} {}
+
+Volume::Volume(const Volume& vol) noexcept : GeometryObject{} {
+  (*this) = vol;
+}
```
paulgessinger: Is this not `= default`?

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,49 @@
...
+    const std::size_t portalIdx, const Transform3& portalToVolTrf,
+    std::shared_ptr<RegularSurface>&& portalSurface) {
+  if (portalIdx >= m_portalPlacements.size()) {
+    m_portalPlacements.resize(portalIdx + 1);
+  }
+  std::unique_ptr<detail::PortalPlacement>& placement =
+      m_portalPlacements[portalIdx];
+  if (!placement) {
+    placement.reset(new detail::PortalPlacement(portalIdx, portalToVolTrf, this,
+                                                std::move(portalSurface)));
```
paulgessinger:
```suggestion
    placement = std::make_unique<detail::PortalPlacement>(portalIdx, portalToVolTrf, this,
                                                std::move(portalSurface));
```

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,49 @@
...
+  }
+  assert(portalToVolTrf.isApprox(placement->portalToVolumeCenter()));
```
paulgessinger: Since this is not hot code, maybe this could always be a check and an exception?

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,93 @@
...
+  ///        local volume coordinate system
+  /// @param gctx The current geometry context object, e.g. alignment
+  virtual const Transform3& globalToLocalTransform(
+      const GeometryContext& gctx) const = 0;
+  /// @brief Returns the transform from the portal's frame to the experiment's
+  ///        global frame for the portal surface associated with the volume
+  /// @param gctx The current geometry context object, e.g. alignment
+  /// @param portalIdx: Internal index of the portal surface [0 - number of portals)
+  virtual const Transform3& portalLocalToGlobal(
+      const GeometryContext& gctx, const std::size_t portalIdx) const = 0;
```
paulgessinger, line 134: Doesn't this overlap with `cachePortalTransform`?

Tests/UnitTests/Core/Geometry/VolumeTests.cpp
paulgessinger, line 1: It might be worth adding a test that actually excercises the movement of a volume, and asserts that the portals are indeed moving in unison.

paulgessinger (commented):
Could you add empty lines between methods?

Core/include/Acts/Geometry/Volume.hpp
junggjo9: The intention was to spare the allocation of the transforms in case it's not needed.

junggjo9:
> Could you add empty lines between methods?

sure will do

Core/include/Acts/Geometry/detail/PortalPlacement.hpp
junggjo9: I've made the constructor protected to avoid that client code can use them in its own realm. Though, I'd keep it in detail in case one wants to arrange the filling of the alignment cache differently

Core/src/Geometry/VolumePlacementBase.cpp
junggjo9: Doesn't work with the protected constructor

Core/include/Acts/Geometry/VolumePlacementBase.hpp
junggjo9: The cachePortalTransform mechanism turned out to be useless while I was implementing the test

Core/src/Geometry/Volume.cpp
junggjo9: It's not default given the volume has the unique_ptrs for the transforms.

Tests/UnitTests/Core/Geometry/VolumeTests.cpp
junggjo9: Done in the alignment tests

Core/src/Geometry/VolumePlacementBase.cpp
junggjo9: Actually it's covered by the unit test of the placement

Core/include/Acts/Geometry/detail/PortalPlacement.hpp
```diff
@@ -0,0 +1,99 @@
...
+  /// @brief Returns the transform to switch from the portal reference
+  ///       frame into the experiment's global frame taking the alignment
+  ///       correction into account
+  /// @param gctx The current geometry context object, e.g. alignment
+  const Transform3& localToGlobalTransform(
+      const GeometryContext& gctx) const final;
+
+  /// @brief Returns the const reference to portal surface connected with
+  ///        this placement instance
+  const Surface& surface() const final;
```
paulgessinger: Any reason why this is not also `RegularSurface`?

Core/include/Acts/Geometry/detail/PortalPlacement.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+  /// @brief Returns the mutable reference to portal surface connected with
+  ///        this placement instance
+  Surface& surface() final;
+
+  /// @brief Returns the pointer to the hold surface
+  std::shared_ptr<RegularSurface> surfacePtr();
```
paulgessinger:
```suggestion
  const std::shared_ptr<RegularSurface>& surfacePtr();
```
maybe

Core/include/Acts/Geometry/TrackingVolume.hpp
```diff
@@ -138,14 +149,34 @@ class TrackingVolume : public Volume {
   /// container
   /// @param denseVolumeVector  The contained dense volumes
   /// @param volumeName is a string identifier
-  TrackingVolume(
+  explicit TrackingVolume(
```
paulgessinger: I think this is redundant, as a multi-argument constructor is always explicit.

Core/include/Acts/Geometry/TrackingVolume.hpp
```diff
@@ -138,14 +149,34 @@ class TrackingVolume : public Volume {
...
+  /// @param placement is the VolumePlacement positioning the volume in the experiment's frame
+  ///                  and also dynamically aligning it
+  /// @param volumeBounds is the description of the volume boundaries
+  /// @param volumeMaterial is are materials of the tracking volume
+  /// @param staticLayerArray is the confined layer array (optional)
+  /// @param containedVolumeArray are the sub volumes if the volume is a
+  /// container
+  /// @param denseVolumeVector  The contained dense volumes
+  /// @param volumeName is a string identifier
+  explicit TrackingVolume(
```
paulgessinger: Same here I believe. Outside of that, does a separate constructor add a benefit here? The placement is already pre-constructed, so maybe a setter is sufficient?

Core/include/Acts/Geometry/TrackingVolume.hpp
```diff
@@ -162,6 +193,67 @@ class TrackingVolume : public Volume {
...
+  ///
+  /// @param visitor The callable. Will be called for each reachable surface
+  /// that is found, a selection of the surfaces can be done in the visitor
+  /// @param type: The category of surfaces that shall be processed by the
+  ///              visitor.
+  ///
+  /// @note If a context is needed for the visit, the vistitor has to provide
+  /// this, e.g. as a private member
+  template <SurfaceVisitor visitor_t>
+  void visitSurfaces(visitor_t&& visitor, const SurfaceVisitType type) const {
```
paulgessinger: I would prefer to avoid making this method more powerful. Could you implement this instead as a subclass to [`ITrackingGeometryVisitor`](https://github.com/acts-project/acts/blob/main/Core/include/Acts/Geometry/TrackingGeometryVisitor.hpp)? There you could just add extra virtual methods to be overridden by a client that will dispatch based on the surface case.

Core/include/Acts/Geometry/Volume.hpp
```diff
@@ -39,21 +41,33 @@ class Volume : public GeometryObject {
   ///
   /// @param transform is the transform to position the volume in 3D space
   /// @param volbounds is the volume boundary definitions
-  Volume(const Transform3& transform, std::shared_ptr<VolumeBounds> volbounds);
+  explicit Volume(const Transform3& transform,
```
paulgessinger: Same here, is the setter not sufficient?

Core/include/Acts/Geometry/Volume.hpp
paulgessinger: Right but is that a concern that we have at the moment? If not, I would just keep it as is.

Core/include/Acts/Geometry/Volume.hpp
```diff
@@ -31,6 +32,7 @@ class VolumeBounds;
 /// information.
 class Volume : public GeometryObject {
  public:
+  friend class VolumeBounds;
```
paulgessinger: Can you clarify what this is needed for?

Core/src/Geometry/VolumeBounds.cpp
```diff
@@ -8,12 +8,13 @@
 
 #include "Acts/Geometry/VolumeBounds.hpp"
 
-std::ostream& Acts::operator<<(std::ostream& sl, const VolumeBounds& vb) {
+#include <cassert>
```
paulgessinger: Used?


```suggestion
#include <cassert>

```

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,166 @@
...
+#include "Acts/Surfaces/RegularSurface.hpp"
+
+#include <memory>
+#include <vector>
+
+namespace Acts {
+class GeometryContext;
+struct OrientedSurface;
+
+/// @brief Interface class to define the transform cache backend for alignable volumes.
```
paulgessinger: I don't think `@brief` makes sense here. You can just remove that and deindent the whole comment block. As a general rule, you essentially never need `@brief`.

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,166 @@
...
+///        Alignable volumes can be dynamically moved by the client code using
+///        the information wrapped into the GeometryContext similar to the
+///        alignable surfaces. The challenge is that the boundary surfaces of
+///        the volume need to move accordingly and that ACTS does not know
+///        anything about the caching of the alignable geometry objects a
+///        priori.
+///
+///        A client-based implementation of the VolumePlacementBase can be
+///        parsed to the constructor of the Volume instead of the fixed
```
paulgessinger:
```suggestion
///        passed to the constructor of the Volume instead of the fixed
```

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,166 @@
...
+///        transform. The Volume is then querying its global position from the
+///        VolumePlacements. The associated bonudary surfaces are also
+///        requesting their position in global space. The Volume's transform and
+///        the bounds can the no longer be overwritten at a later stage.
+///
+///        An implementation of the VolumePlacementBase needs to satisfy the
```
paulgessinger:
```suggestion
///        An implementation of the @ref VolumePlacementBase needs to satisfy the
```

maybe

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,166 @@
...
+///        following interface.
+///
+///         1) Transforms switching from the volume's frame into the global
+///            experiment's frame and vice versa:
+///
+///           const Transform3& localToGlobalTransform(const GeometryContext&
+///           gctx) const;
+///
+///           const Transform3& localToGlobalTransform(const GeometryContext&
+///           gctx) const;
```
paulgessinger: Consider using a code snippet that is compile tested for code listings in comments.
At the very least, use

<pre>
```cpp
```
</pre>


to delimit code.

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,166 @@
...
+///
+///
+///        2) At some point during the (tracking) volume construction, the
+///           boundary surfaces, also called the portals, are created from
+///           the associated bounds. ACTS ensures that these surfaces are
+///           connected with the VolumePlacement instance. The client code
+///           needs to ensure that  there is enough memory allocated by the
```
paulgessinger:
```suggestion
///           needs to ensure that there is enough memory allocated by the
```

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,166 @@
...
+///
+///           const Transform3& localToGlobalTransform(const GeometryContext&
+///           gctx) const;
+///
+///           const Transform3& localToGlobalTransform(const GeometryContext&
+///           gctx) const;
+///
+///
+///        2) At some point during the (tracking) volume construction, the
+///           boundary surfaces, also called the portals, are created from
```
paulgessinger: At the code level, they're not quite the same thing, the portals are created from the boundary surfaces. Conceptually they're pretty much the same however.

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,166 @@
...
+///           placement. The argument tells the client how many transforms need
+///           to be cached.
+///
+///             const Transform3& portalLocalToGlobal(const GeometryContext&
+///             gctx,
+///                                                   const std::size_t
+///                                                   portalIdx) const;
+///
+///           Is called every time when the portal is asking for its position in
+///           space. The portalIdx is the unique index of the portal and assists
```
paulgessinger:
```suggestion
///           space. The portalIdx is the unique index of the portal in the scope of this volume and assists
```

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,166 @@
...
+  VolumePlacementBase& operator=(VolumePlacementBase&& other) noexcept;
+
+  /// @brief Delete copy constructor
+  VolumePlacementBase(const VolumePlacementBase& other) = delete;
+
+  /// @brief Delete copy assignment operator
+  VolumePlacementBase& operator=(const VolumePlacementBase& other) = delete;
+
+  /// @brief Abrivation of the portal surface vector
+  using PortalVec_t = std::vector<OrientedSurface>;
```
paulgessinger: Consider whether this is needed (I don't think so, it's used twice only I believe). We've tried to remove trivial typedefs like this in a few places. Also specifically here: `OrientedSurface` != `Portal`, which should be reflected in the naming.

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,126 @@
...
+  }
+  return (*this);
+}
+
+VolumePlacementBase::PortalVec_t VolumePlacementBase::makePortalsAlignable(
+    PortalVec_t portalsToAlign) {
+  // It's safe to use the default geometry context as all incoming portal
+  // surfaces are explicitly requested to have no surface placement and hence
+  // the cached transform is returned
+  const GeometryContext gctx = GeometryContext::dangerouslyDefaultConstruct();
```
paulgessinger: Might be worth checking this explicitly and throwing an exception if it's not the case, just to be safe.

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,126 @@
...
+VolumePlacementBase::VolumePlacementBase() noexcept = default;
+
+VolumePlacementBase::~VolumePlacementBase() = default;
+
+VolumePlacementBase::VolumePlacementBase(VolumePlacementBase&& other) noexcept
+    : VolumePlacementBase{} {
+  (*this) = std::move(other);
+}
+
+VolumePlacementBase& VolumePlacementBase::operator=(
```
paulgessinger: This seems pretty dangerous, e.g. if the `VolumePlacementBase` is allocated on the stack. Maybe we should instead disallow move construction entirely (which would essentially force construction on the heap from the beginning)?

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,126 @@
...
+namespace Acts {
+
+VolumePlacementBase::VolumePlacementBase() noexcept = default;
+
+VolumePlacementBase::~VolumePlacementBase() = default;
+
+VolumePlacementBase::VolumePlacementBase(VolumePlacementBase&& other) noexcept
+    : VolumePlacementBase{} {
+  (*this) = std::move(other);
+}
```
paulgessinger: This is just `= default`, right?

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+VolumePlacementBase::PortalVec_t VolumePlacementBase::makePortalsAlignable(
+    PortalVec_t portalsToAlign) {
+  // It's safe to use the default geometry context as all incoming portal
+  // surfaces are explicitly requested to have no surface placement and hence
+  // the cached transform is returned
+  const GeometryContext gctx = GeometryContext::dangerouslyDefaultConstruct();
+
+  const bool callExpand = portalsToAlign.size() != m_portalPlacements.size();
```
paulgessinger: When would these be different?

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,126 @@
...
+      std::max(portalsToAlign.size(), m_portalPlacements.size()));
+  if (m_portalPlacements.size() != portalsToAlign.size()) {
+    throw std::logic_error(
+        std::format("VolumePlacementBase::makePortalsAlignable() - Has been "
+                    "called previously for a different type of volume bounds. "
+                    "Already registered portals {:} vs incoming portals {:}",
+                    m_portalPlacements.size(), portalsToAlign.size()));
+  }
+  for (std::size_t portalIdx = 0lu; portalIdx < nPortalPlacements();
+       ++portalIdx) {
```
paulgessinger: Maybe 
```suggestion
  for (const auto& [portalIdx, portal] : enumerate(portalsToAlign)) {
```

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,126 @@
...
+    std::shared_ptr<RegularSurface>& portalSurface =
+        portalsToAlign[portalIdx].surface;
```
paulgessinger: If you keep the index based loop:

```suggestion
        portalsToAlign.at(portalIdx).surface;
```

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,126 @@
...
+          "VolumePlacementBase::makePortalsAlignable() - The {:}-th surface is "
+          "already connected to the alignment system",
+          portalIdx));
+    }
+    // It is expected that the volume bounds are callign the orientedSurfaces
+    // method with an identity transform
+    const Transform3 portalToVolTrf =
+        portalSurface->localToGlobalTransform(gctx);
+    std::unique_ptr<detail::PortalPlacement>& placement =
+        m_portalPlacements[portalIdx];
```
paulgessinger:
```suggestion
        m_portalPlacements.at(portalIdx);
```

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,126 @@
...
+    // Just create a new placement which makes the surface alignable
+    if (!placement) {
+      placement.reset(new detail::PortalPlacement(portalIdx, portalToVolTrf,
```
paulgessinger: Use `make_unique` or `make_shared`. I think this is now prevented due to the private constructor, but I would not do a raw `new`.

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,126 @@
...
+                        toString(portalToVolTrf)));
+      }
+      if (!(placement->surface().bounds() == portalSurface->bounds())) {
+        std::stringstream sstr{};
+        sstr << "VolumePlacementBase::makePortalsAlignable() -  The bounds for "
+                "the ";
+        sstr << portalIdx
+             << "-th surface differ with the ones from a previous call: \n";
+        sstr << " -- exist: " << placement->surface().bounds() << "\n";
+        sstr << " -- this call: " << portalSurface->bounds();
```
paulgessinger: Why not also `std::format`?

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,126 @@
...
+    const Transform3 portalToVolTrf =
+        portalSurface->localToGlobalTransform(gctx);
+    std::unique_ptr<detail::PortalPlacement>& placement =
+        m_portalPlacements[portalIdx];
+    // Just create a new placement which makes the surface alignable
+    if (!placement) {
+      placement.reset(new detail::PortalPlacement(portalIdx, portalToVolTrf,
+                                                  this, portalSurface));
+    } else {
+      // The method has been called previously and it needs to be checked that
```
paulgessinger: I don't fully understand under which circumstances this method would be called more than once. If this is a pure error case, I think we should bail early and not allow the method to be run multiple times.

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,126 @@
...
+    m_portalPlacements = std::move(other.m_portalPlacements);
+    for (const auto& portal : m_portalPlacements) {
+      portal->m_parent = this;
+    }
+  }
+  return (*this);
+}
+
+VolumePlacementBase::PortalVec_t VolumePlacementBase::makePortalsAlignable(
+    PortalVec_t portalsToAlign) {
```
paulgessinger: If this object knows which volume it's connected to, can't it figure the portals out from the volume itself?

Also, this currently works on the oriented surface from before the portal construction / synchronization, no? Even though we won't be able to align portals from stacks etc. in any case, I think this is dangrous. Can't we do this on the finalized portals by just installing the placement?

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,126 @@
...
+  // It's safe to use the default geometry context as all incoming portal
+  // surfaces are explicitly requested to have no surface placement and hence
+  // the cached transform is returned
+  const GeometryContext gctx = GeometryContext::dangerouslyDefaultConstruct();
+
+  const bool callExpand = portalsToAlign.size() != m_portalPlacements.size();
+
+  m_portalPlacements.resize(
+      std::max(portalsToAlign.size(), m_portalPlacements.size()));
+  if (m_portalPlacements.size() != portalsToAlign.size()) {
```
paulgessinger: As commented below, I think I would *latch* this with an `initialized` boolean member: if that's already `true` here, that's an exception.

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,126 @@
...
+       ++portalIdx) {
+    std::shared_ptr<RegularSurface>& portalSurface =
+        portalsToAlign[portalIdx].surface;
+    if (portalSurface->surfacePlacement() != nullptr) {
+      throw std::invalid_argument(std::format(
+          "VolumePlacementBase::makePortalsAlignable() - The {:}-th surface is "
+          "already connected to the alignment system",
+          portalIdx));
+    }
+    // It is expected that the volume bounds are callign the orientedSurfaces
```
paulgessinger:
```suggestion
    // It is expected that the volume bounds are calling the orientedSurfaces
```

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+  /// @brief This method is called for the first time when the portal surfaces
+  ///        are registered with the VolumePlacmentBase class. The client
+  ///        receives the information about how many portals exist and it's
+  ///        requested to adapt the size of the transform cache backend
+  ///        accordingly. The method is called after the portal placements have
+  ///        been created.
+  /// @param nPortals: The number of portals that are registered with this volume
+  ///                  placement instance
+  virtual void expandTransformCache(const std::size_t nPortals) = 0;
```
paulgessinger: I don't fully understand this. Don't we expect this to be doable during the *population* phase?

Core/include/Acts/Geometry/detail/PortalPlacement.hpp
```diff
@@ -0,0 +1,99 @@
...
+  ///                experiment's frame
+  /// @param surface: Pointer to the portal surface itself which is becoming alignable
+  ///                 with the construction of this PortalPlacement
+  PortalPlacement(const std::size_t portalIdx, const Transform3& portalTrf,
+                  VolumePlacementBase* parent,
+                  std::shared_ptr<RegularSurface> surface);
+
+ private:
+  /// @brief Orientation of the portal surface w.r.t the volume
+  Transform3 m_interalTrf{Transform3::Identity()};
```
paulgessinger: In the spirit of naming our transforms more explicitly:


```suggestion
  Transform3 m_portalToVolumeCenter{Transform3::Identity()};
```

Core/src/Geometry/PortalPlacement.cpp
paulgessinger, line 1: `detail` folders are not needed in `src`, you can just move this one up.

Tests/UnitTests/Core/Geometry/AlignmentContextTests.cpp
```diff
@@ -121,9 +128,91 @@ inline Surface& AlignableDetectorElement::surface() {
...
+    if (alignContext != nullptr &&
+        alignContext->volumeLocToGlobAlign != nullptr &&
+        alignContext->alignmentIndex < 2) {
+      return alignContext->volumeLocToGlobAlign->at(
+          alignContext->alignmentIndex);
+    }
+    return m_locToGlob;
+  }
+
+  /// @brief Returns the transformation from the experiment's global frame to the
```
paulgessinger: No need for doc comments in the test.

Core/include/Acts/Geometry/detail/PortalPlacement.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+namespace detail {
+/// @brief Implementation of the `SurfacePlacementBase` to synchronize the alignment
+///        of an alignable volume with the alignment of the boundary surfaces
+///        associated with the Volume.
+class PortalPlacement : public SurfacePlacementBase {
+ public:
+  /// @brief Allow the VolumePlacementBase as only class to construct
+  ///        the portal placement
+  friend class Acts::VolumePlacementBase;
```
andiwand, line 25:
```suggestion
  friend class VolumePlacementBase;
```

Core/include/Acts/Geometry/detail/PortalPlacement.hpp
```diff
@@ -0,0 +1,99 @@
...
+#include "Acts/Surfaces/SurfacePlacementBase.hpp"
+
+namespace Acts {
+class VolumePlacementBase;
+
+namespace detail {
+/// @brief Implementation of the `SurfacePlacementBase` to synchronize the alignment
+///        of an alignable volume with the alignment of the boundary surfaces
+///        associated with the Volume.
+class PortalPlacement : public SurfacePlacementBase {
```
andiwand: should the full class be `final`?

Core/include/Acts/Geometry/TrackingVolume.hpp
```diff
@@ -162,6 +193,67 @@ class TrackingVolume : public Volume {
...
+          return surface.isAlignable();
+        case sensitive:
+          return surface.isSensitive();
+        case sensitiveAlignable:
+          return surface.isSensitive() && surface.isAlignable();
+        case passiveAlignable:
+          return !surface.isSensitive() && surface.isAlignable();
+      }
+      return true;
+    };
```
andiwand: wouldn't it be easy enough to reject these surfaces at the callee site?

Core/src/Geometry/Volume.cpp
andiwand: might be another reason to drop the `std::unique_ptr` for now

Core/src/Geometry/Volume.cpp
```diff
@@ -97,23 +126,35 @@ void Volume::update(const GeometryContext& /*gctx*/,
...
 const Transform3& Volume::transform() const {
-  return m_transform;
+  assert(m_transform != nullptr);
+  return (*m_transform);
 }
 
 const Transform3& Volume::itransform() const {
-  return m_itransform;
+  assert(m_itransform != nullptr);
+  return (*m_itransform);
```
andiwand, line 133: I *think* sonarcloud will complain about the `(*...)`

Core/include/Acts/Geometry/Volume.hpp
```diff
@@ -180,19 +194,29 @@ class Volume : public GeometryObject {
...
-  Transform3 m_transform;
+  std::unique_ptr<const Transform3> m_transform{};
 
   /// @brief Inverse of the transform matrix for efficient calculations
-  Transform3 m_itransform;
+  std::unique_ptr<const Transform3> m_itransform{};
 
   /// @brief Center position of the volume in global coordinates
-  Vector3 m_center;
+  Vector3 m_center{Vector3::Zero()};
```
andiwand, line 235: is `center` still useful?

Core/src/Geometry/Volume.cpp
```diff
@@ -97,23 +126,35 @@ void Volume::update(const GeometryContext& /*gctx*/,
...
 }
 
-const Vector3& Volume::center(const GeometryContext& /*gctx*/) const {
-  return m_center;
+Vector3 Volume::center(const GeometryContext& gctx) const {
+  // fast access via transform matrix (and not translation())
+  const auto tMatrix = localToGlobalTransform(gctx).matrix();
+  return Vector3{tMatrix(0, 3), tMatrix(1, 3), tMatrix(2, 3)};
```
andiwand: why is this faster than `transation()`?

could also be done with `.col(3).head<3>()` I believe

junggjo9: That's a comment in the surface center() method which I blindly copied xD

Core/include/Acts/Geometry/Volume.hpp
junggjo9: Was thinking about removing it once we get rid of the volume methods without context. Indeed It's not useful

Core/src/Geometry/Volume.cpp
junggjo9: Sad dog noises

Core/include/Acts/Geometry/TrackingVolume.hpp
junggjo9: yeah can also be done there. My aim was to generalize the surface visitor, but I'll go with a new visitor class as requested by Paul

Core/include/Acts/Geometry/detail/PortalPlacement.hpp
junggjo9: Probably, yes

junggjo9:
```suggestion
class PortalPlacement : public SurfacePlacementBase final {
```

Core/include/Acts/Geometry/detail/PortalPlacement.hpp
junggjo9: to avoid the reference count increase? Don't mind

Core/include/Acts/Geometry/detail/PortalPlacement.hpp
junggjo9: Right that's posible. Will change
```suggestion
  const RegularSurface& surface() const final;
```

Core/src/Geometry/Volume.cpp
andiwand: ok - I think this might be due to allowing sheer and scale in the transforms right now. I think we should restrict this to rotation and translation. everything else will break the geometry and intersections anyways

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,91 @@
...
+  if (&other != this) {
+    m_portalPlacements = std::move(other.m_portalPlacements);
+    for (const auto& portal : m_portalPlacements) {
+      portal->m_parent = this;
+    }
+  }
+  return (*this);
+}
+
+void VolumePlacementBase::makePortalsAlignable(
```
paulgessinger, line 22: In discussion, we decided to make this function virtual with a base implementation.

Core/include/Acts/Geometry/VolumePlacementBase.hpp
```diff
@@ -0,0 +1,174 @@
...
+  /// @param other: Object to be moved into this instance
+  VolumePlacementBase(VolumePlacementBase&& other) noexcept;
+
+  /// Move assignment operator
+  /// @param other: Object to be moved into this instance
+  /// @returns Reference to this instance
+  VolumePlacementBase& operator=(VolumePlacementBase&& other) noexcept;
+
+  /// Abrivation of the portal surface vector
+  using PortalVec_t = std::vector<std::shared_ptr<RegularSurface>>;
```
paulgessinger: Reconsider typedef

Core/src/Geometry/VolumePlacementBase.cpp
paulgessinger: We decided to delete move and copy operator

Core/src/Geometry/VolumePlacementBase.cpp
```diff
@@ -0,0 +1,91 @@
...
+        "empty");
+  }
+
+  if (!m_portalPlacements.empty()) {
+    throw std::runtime_error(
+        "VolumePlacementBase::makePortalsAlignable() - Portals were already "
+        "registered before");
+  }
+
+  for (const auto& [portalIdx, portal] : enumerate(portalsToAlign)) {
```
paulgessinger: Check this again. Maybe use `zip`

junggjo9: portalIdx is used in case of an exception

Core/include/Acts/Geometry/VolumePlacementBase.hpp
junggjo9: Removed

Core/src/Geometry/VolumePlacementBase.cpp
junggjo9: Done

Core/src/Geometry/Volume.cpp
junggjo9: Imlemented as simple translation

Core/include/Acts/Geometry/Volume.hpp
junggjo9: Postponed for folow up PR

Core/include/Acts/Geometry/TrackingVolume.hpp
junggjo9: Removed

Core/src/Geometry/VolumePlacementBase.cpp
junggjo9: Changed to public constructor and make_unique

Core/include/Acts/Geometry/Volume.hpp
junggjo9:
```suggestion
  Volume(const Transform3& transform,
```

Core/include/Acts/Geometry/detail/AlignablePortalVisitor.hpp
```diff
@@ -0,0 +1,50 @@
...
+  ///        un-aligneable subvolumes
+  void enableInspection();
+  /// @brief Visit and potentially modify a tracking volume
+  /// @param volume The tracking volume being visited
+  /// @note Called for each volume in the geometry hierarchy during traversal
+  void visitVolume(TrackingVolume& volume) override;
+
+ private:
+  /// @brief The construction geometry context connected to the experiment's alignment
+  const GeometryContext m_gctx{GeometryContext::dangerouslyDefaultConstruct()};
```
paulgessinger: Do we need to default construct this, if the constructor initializes it with the provided one?

Core/src/Geometry/Volume.cpp
paulgessinger: Ah I see. I would probably define the copy constructor and use copy-and-swap or copy-and-move for the assignment operator, but ok

Core/src/Geometry/AlignablePortalVisitor.cpp
```diff
@@ -0,0 +1,115 @@
...
+  }
+  ACTS_DEBUG("AlignablePortalVisitor() - Found alignable volume "
+             << volume.geometryId() << ", " << volume.volumeBounds());
+  std::vector<std::shared_ptr<RegularSurface>> alignable{};
+  for (Portal& portal : volume.portals()) {
+    ACTS_DEBUG("AlignablePortalVisitor() - Check wheher portal "
+               << portal.surface().geometryId() << " is a boundary surface");
+    if (portal.surface().geometryId().withBoundary(0) != volume.geometryId()) {
+      continue;
+    }
```
paulgessinger, line 60: Should we make this a loud error? I don't think it should realistically happen.


```suggestion
    ACTS_DEBUG("AlignablePortalVisitor() - Check whether portal "
               << portal.surface().geometryId() << " is a boundary surface");
    if (portal.surface().geometryId().withBoundary(0) != volume.geometryId()) {
      continue;
    }
```

Core/src/Geometry/AlignablePortalVisitor.cpp
```diff
@@ -0,0 +1,115 @@
...
+  // unaligned portals
+  if (m_doInspect) {
+    std::ranges::transform(alignable, std::back_inserter(portalTrfBefore),
+                           [&](const std::shared_ptr<RegularSurface>& surface) {
+                             return surface->localToGlobalTransform(m_gctx);
+                           });
+  }
+  volume.volumePlacement()->makePortalsAlignable(m_gctx, alignable);
+
+  if (!m_doInspect) {
```
paulgessinger, line 92: Why only sometimes?

Core/include/Acts/Geometry/detail/AlignablePortalVisitor.hpp
junggjo9: Not necessarily. I'm just default constructing variables where possible when I'm declaring them

Core/src/Geometry/AlignablePortalVisitor.cpp
junggjo9: I need to filter the boundary portals and I thought that a volume with sub-volumes has also the portals of the subvolumes?

Core/src/Geometry/AlignablePortalVisitor.cpp
junggjo9: Good question

Core/include/Acts/Utilities/CloneablePtr.hpp
```diff
@@ -77,45 +77,61 @@ class CloneablePtr {
...
+  {
+    m_ptr = std::move(ptr);
+    if constexpr (!std::is_same_v<const T, const T1>) {
+      m_cloner = [](const T& obj) {
+        return std::make_unique<T1>(static_cast<const T1&>(obj));
+      };
+    }
+    return *this;
+  }
+  // requires std::is_copy_constructible_v<T>
```
paulgessinger:
```suggestion
```

Core/include/Acts/Utilities/CloneablePtr.hpp
```diff
@@ -77,45 +77,61 @@ class CloneablePtr {
...
+  CloneablePtr& operator=(std::unique_ptr<T1>&& ptr)
+    requires(std::is_copy_constructible_v<T1> && std::is_base_of_v<T, T1>)
+  {
+    m_ptr = std::move(ptr);
+    if constexpr (!std::is_same_v<const T, const T1>) {
+      m_cloner = [](const T& obj) {
+        return std::make_unique<T1>(static_cast<const T1&>(obj));
+      };
+    }
+    return *this;
```
paulgessinger, line 95: What's this needed for? The constructor's default argument should be able to create a new cloner, shouldn't it?

Core/include/Acts/Utilities/CloneablePtr.hpp
```diff
@@ -77,45 +77,61 @@ class CloneablePtr {
...
-  /// @brief Get the raw pointer
+  /// Get the raw pointer
   /// @return Pointer to the managed object, or nullptr
   T* get() const { return m_ptr.get(); }
 
-  /// @brief Release ownership of the managed object
+  /// Release ownership of the managed object
   /// @return Pointer to the formerly managed object
-  T* release() {
+  std::unique_ptr<T> release() {
```
paulgessinger, line 130: I think `release` implies that it's no longer owned, so this should return the bare pointer. Do you need the unique for something specific?

junggjo9: I don't need it personally, but I think that's making the pointer more safe against accidential memory leaks. If the uer wants to have ownership, he is forced to transfer it to a smart pointer ;)

Core/include/Acts/Utilities/CloneablePtr.hpp
junggjo9: basically if you want to overwrite the pointer with another unique-ptr

```cpp
CLoneablePtr<blub> kuchen{};
kuchen = std::make_unique<blub>();
```
Given that this might open the domain also to cache child classes,  I thought an update of the cloner will be useful here as well

Core/include/Acts/Geometry/detail/AlignablePortalVisitor.hpp
paulgessinger: But it's redundant, isn't it?

Core/src/Geometry/AlignablePortalVisitor.cpp
paulgessinger: I see what you mean I think. Unfortunately, I'm not sure this will work, because I don't think the numbering is completely guaranteed. It's possible that @dimitra97 made this deterministic now, however. Would it make sense to try to do this inside-out, so basically find the leave volumes first, and then begin assigning them, while skipping if they're already assigned?

Core/include/Acts/Utilities/CloneablePtr.hpp
paulgessinger: I still think just going through the `CloneablePtr` constructor is good enough. We could consider adding automatic upcasting to `CloneablePtr` if we need it.

Core/include/Acts/Utilities/CloneablePtr.hpp
paulgessinger: I'm just saying the precedent in `unique_ptr::release` is that it returns a raw pointer and stops owning the object.

paulgessinger (approved):

PR #4968, opened 2026-01-15
fix: Use (cached) runtime type hash in Acts::Any

paulgessinger:
Hashing `__PRETTY_FUNCTION__` is not consistent across compilers, so a library build with one compiler will not produce the same hashes as a library with another compiler. This PR switches the hash calculation to be based on `typeid(T).name()`. It's calculated at runtime and then cached in a static variable, but should be stable across compilers.

--- END COMMIT MESSAGE ---

@wdconinc

### DIFF
--- a/Alignment/include/ActsAlignment/Kernel/Alignment.ipp
+++ b/Alignment/include/ActsAlignment/Kernel/Alignment.ipp
@@ -173,7 +173,8 @@ ActsAlignment::Alignment<fitter_t>::updateAlignmentParameters(
   for (const auto& [surface, index] : alignResult.idxedAlignSurfaces) {
     // 1. The original transform
     const Acts::Vector3& oldCenter = surface->center(gctx);
-    const Acts::Transform3& oldTransform = surface->transform(gctx);
+    const Acts::Transform3& oldTransform =
+        surface->localToGlobalTransform(gctx);
 
     // 2. The delta transform
     deltaAlignmentParam = alignResult.deltaAlignmentParameters.segment(
@@ -315,7 +316,7 @@ ActsAlignment::Alignment<fitter_t>::align(
     for (const auto& det : alignOptions.alignedDetElements) {
       const auto& surface = &det->surface();
       const auto& transform =
-          det->transform(alignOptions.fitOptions.geoContext);
+          det->localToGlobalTransform(alignOptions.fitOptions.geoContext);
       // write it to the result
       alignResult.alignedParameters.emplace(det, transform);
       const auto& translation = transform.translation();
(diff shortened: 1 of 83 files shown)


tmadlener:
I am most likely missing some constraints here, but could `std::source_location::function_name` do the trick here? As far as I can see, that should all be `constexpr` and available from c++20 and I think something along the [following](https://godbolt.org/z/5WbYeeGGn) could work without giving up the `consteval` part.

edit: It might have the same issue as `__PRETTY_FUNCTION__` and seems to be unstable across compilers.

paulgessinger:
@tmadlener It does, unfortunately: https://godbolt.org/z/jPedPr83E

wdconinc:
Looks good to me, even if I can't currently test it.

Even [`std::type_info::name`](https://en.cppreference.com/w/cpp/types/type_info/name.html) is actually implementation defined, but it seems that at least gcc and clang agree on using the same mangled name... 

And [`std::type_info::hash_code`](https://en.cppreference.com/w/cpp/types/type_info/hash_code.html) sounded reasonable until you read the fine-print that says "objects referring to different types may have the same hash code" and "hash code for the same type can change between invocations of the same program".

andiwand:
> Even [`std::type_info::name`](https://en.cppreference.com/w/cpp/types/type_info/name.html) is actually implementation defined, but it seems that at least gcc and clang agree on using the same mangled name...

Hmm that is not great. But yeah maybe a bit better than what we currently have

> And [`std::type_info::hash_code`](https://en.cppreference.com/w/cpp/types/type_info/hash_code.html) sounded reasonable until you read the fine-print that says "objects referring to different types may have the same hash code" and "hash code for the same type can change between invocations of the same program".

Generally `std::type_info` itself should do the trick. I think `std::type_info::hash_code` would also be fine as we do not need to persitify that info. Hash collisions would already be possible but I would think that this should be very unlikely.
The main problem right now with `std::type_info` is that we experienced same types produce different `std::type_info`s across library boundaries. We think that this is another issue that should be solvable, but we couldn't for now.

paulgessinger:
It's somehow related to the linker not removing duplicate type info symbols for the nominally same type. Documentation around this is sparse and debugging difficult. 

I did some back of the envelope calculation and the risk of hash collisions with the 64Bit version of the hashing algorithm in use here is fairly low, to first order I don't think we need to worry about that.

acts-project-service:
## :red_circle: Athena integration test results [f78f1ca968f311ef738da1f4a7cd61e7868404cb]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/67808961) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/13787364)!

PR #4973, opened 2026-01-15
refactor: Refactor examples plotting tools to not use ROOT

benjaminhuth:
Refactors the plot tools to use the Boost. Histogram-based types that have been introduced in #4939.

--- END COMMIT MESSAGE ---


Examples/Framework/include/ActsExamples/Validation/DuplicationPlotTool.hpp
```diff
@@ -63,45 +44,50 @@ class DuplicationPlotTool {
...
   std::unique_ptr<const Acts::Logger> m_logger;
 
-  /// The logger
-  const Acts::Logger& logger() const { return *m_logger; }
+  Acts::Experimental::ProfileHistogram1 m_nDuplicatedVsPt;
+  Acts::Experimental::ProfileHistogram1 m_nDuplicatedVsEta;
+  Acts::Experimental::ProfileHistogram1 m_nDuplicatedVsPhi;
+  Acts::Experimental::Efficiency1 m_duplicationRatioVsPt;
+  Acts::Experimental::Efficiency1 m_duplicationRatioVsEta;
+  Acts::Experimental::Efficiency1 m_duplicationRatioVsPhi;
```
andiwand: the main difference here is that we do not decouple the state anymore from the tool

what is the reasoning behind this?

I don't think it is a big deal for Examples but if we ever move that to Core / Plugin we may want to change this back

Examples/Io/Root/src/RootTrackFitterPerformanceWriter.cpp
```diff
@@ -58,36 +65,129 @@ ActsExamples::RootTrackFitterPerformanceWriter::
...
+    std::string suffix = baseName.substr(baseName.find('_'));
+
+    // Create mean and width histograms with same X binning as the 2D histogram
+    int nBinsX = hist2d->GetNbinsX();
+    TH1F* meanHist =
+        new TH1F((meanPrefix + suffix).c_str(),
+                 (std::string(hist2d->GetTitle()) + " mean").c_str(), nBinsX,
+                 hist2d->GetXaxis()->GetXmin(), hist2d->GetXaxis()->GetXmax());
+    TH1F* widthHist =
+        new TH1F((widthPrefix + suffix).c_str(),
```
andiwand:
```suggestion
    auto meanHist =
        std::make_unique<TH1F>((meanPrefix + suffix).c_str(),
                 (std::string(hist2d->GetTitle()) + " mean").c_str(), nBinsX,
                 hist2d->GetXaxis()->GetXmin(), hist2d->GetXaxis()->GetXmax());
    auto widthHist =
        std::make_unique<TH1F>((widthPrefix + suffix).c_str(),
```

Plugins/Root/include/ActsPlugins/Root/HistogramConverter.hpp
```diff
@@ -10,13 +10,25 @@
...
 #include "Acts/Utilities/Histogram.hpp"
 
+#include <vector>
+
 class TEfficiency;
 class TH1F;
 class TH2F;
 class TProfile;
 
 namespace ActsPlugins {
```
andiwand, line 20: not sure how we concluded on the name spaces, but either we should add `::Root` or rename `extractBinEdges` to make clear that it is part of the root plugin

alternatively you could elevate this function to the boost histogram code since it looks generic enough

Plugins/Root/src/HistogramConverter.cpp
```diff
@@ -19,7 +19,7 @@
 
 using namespace Acts::Experimental;
 
-namespace {
+namespace ActsPlugins {
 
 std::vector<double> extractBinEdges(const AxisVariant& axis) {
```
andiwand:
```suggestion
std::vector<double> ActsPlugins::extractBinEdges(const AxisVariant& axis) {
```
I usually try to bind the definition to a namespace to avoid linker errors in case the signature changes here or in the declaration

Examples/Framework/include/ActsExamples/Validation/DuplicationPlotTool.hpp
benjaminhuth: Hmm yeah this was mostly a pragmatic decision, since the writer functions anyway are not `const`, and I did not see a benefit in this pattern here.
I didn't thought about potential `Core` integration, but all of this couples relatively strongly to the Examples EDM...

andiwand: yeah I think a Core migration is rather a long term project. another reason for having this as a member would be if we have a base class for these tools. but I don't think there is a good reason to have one

but as said I think this is OK for Examples - resolving

Plugins/Root/include/ActsPlugins/Root/HistogramConverter.hpp
benjaminhuth: Hmm. So in principle this does not couple to ROOT or the plugin code, So not sure if prefixing makes sense here. in principle it could live in core, would that be better?

Examples/Framework/include/ActsExamples/Validation/EffPlotTool.hpp
```diff
@@ -30,18 +29,20 @@ class EffPlotTool {
...
-        {"Pt", PlotHelpers::Binning::Uniform("pT [GeV/c]", 40, 0, 100)},
-        {"LogPt",
-         PlotHelpers::Binning::Logarithmic("pT [GeV/c]", 11, 0.1, 100)},
-        {"LowPt", PlotHelpers::Binning::Uniform("pT [GeV/c]", 40, 0, 2)},
-        {"D0", PlotHelpers::Binning::Uniform("d_0 [mm]", 200, -200, 200)},
-        {"Z0", PlotHelpers::Binning::Uniform("z_0 [mm]", 50, -200, 200)},
-        {"DeltaR", PlotHelpers::Binning::Uniform("#Delta R", 100, 0, 0.3)},
-        {"prodR", PlotHelpers::Binning::Uniform("prod_R [mm]", 100, 0, 200)}};
+    std::map<std::string, Acts::Experimental::AxisVariant> varBinning = {
+        {"Eta", Acts::Experimental::BoostRegularAxis(40, -3.0, 3.0, "#eta")},
```
paulgessinger: I'd probably consider a `using` hre to make it less verbose.

Examples/Framework/include/ActsExamples/Validation/FakePlotTool.hpp
```diff
@@ -66,47 +41,64 @@ class FakePlotTool {
...
-
-  /// @brief delete the fake ratio plots
-  ///
-  /// @param cache cache object for fake ratio plots
-  void clear(Cache& cache) const;
+  void fill(const SimParticleState& truthParticle,
+            std::size_t nTruthMatchedTracks, std::size_t nFakeTracks);
+
+  /// @brief Accessors for histograms (const reference)
+  const Acts::Experimental::Histogram2& nRecoVsPt() const {
```
paulgessinger: I think I get why you need this, but it seems a bit verbose. Would it make sense to hold these histograms keyed by strings in a map and then just expose that map?  Or maybe one map per type?

Examples/Framework/src/Validation/DuplicationPlotTool.cpp
```diff
@@ -9,100 +9,73 @@
...
   const double fit_phi = phi(momentum);
   const double fit_eta = eta(momentum);
   const double fit_pT = perp(momentum);
 
-  PlotHelpers::fillEff(cache.duplicationRatio_vs_pT, fit_pT, status);
-  PlotHelpers::fillEff(cache.duplicationRatio_vs_eta, fit_eta, status);
-  PlotHelpers::fillEff(cache.duplicationRatio_vs_phi, fit_phi, status);
+  m_duplicationRatioVsPt.fill({fit_pT}, status);
+  m_duplicationRatioVsEta.fill({fit_eta}, status);
+  m_duplicationRatioVsPhi.fill({fit_phi}, status);
```
paulgessinger: What is the `{}` doing here?

Examples/Io/Root/src/RootTrackFitterPerformanceWriter.cpp
```diff
@@ -58,36 +65,130 @@ ActsExamples::RootTrackFitterPerformanceWriter::
...
+  ActsPlugins::toRoot(m_trackSummaryPlotTool.nStatesVsEta())->Write();
+  ActsPlugins::toRoot(m_trackSummaryPlotTool.nMeasurementsVsEta())->Write();
+  ActsPlugins::toRoot(m_trackSummaryPlotTool.nHolesVsEta())->Write();
+  ActsPlugins::toRoot(m_trackSummaryPlotTool.nOutliersVsEta())->Write();
+  ActsPlugins::toRoot(m_trackSummaryPlotTool.nSharedHitsVsEta())->Write();
+  ActsPlugins::toRoot(m_trackSummaryPlotTool.nStatesVsPt())->Write();
+  ActsPlugins::toRoot(m_trackSummaryPlotTool.nMeasurementsVsPt())->Write();
+  ActsPlugins::toRoot(m_trackSummaryPlotTool.nHolesVsPt())->Write();
+  ActsPlugins::toRoot(m_trackSummaryPlotTool.nOutliersVsPt())->Write();
+  ActsPlugins::toRoot(m_trackSummaryPlotTool.nSharedHitsVsPt())->Write();
```
paulgessinger: Nitpick, consider `using ActsPlugins::toRoot`.

Python/Examples/src/plugins/Root.cpp
```diff
@@ -93,42 +93,47 @@ PYBIND11_MODULE(ActsExamplesPythonBindingsRoot, root) {
 
   // Output
   {
-    // Bindings for the binning in e.g., TrackFinderPerformanceWriter
+    // Bindings for axis variant used in PlotTool configurations
     {
-      py::class_<PlotHelpers::Binning>(root, "Binning")
+      py::class_<Acts::Experimental::AxisVariant>(root, "AxisVariant")
```
paulgessinger, line 98: Does this need python-level updates?

Examples/Framework/include/ActsExamples/Validation/FakePlotTool.hpp
benjaminhuth: I also thought about maybe using a map of `std::variant`, but I wasn't sure if it would make the thing more complicated in the end. But maybe a map for each type would be a good compromise...

Examples/Framework/src/Validation/DuplicationPlotTool.cpp
benjaminhuth: Well that's the API :D I think we discussed this two weeks ago, I preferred `std::array` over variadic templates here.

Python/Examples/src/plugins/Root.cpp
benjaminhuth: Hmm it was possible to customize the binning before, I don't plan to drop this possibility. As we discussed, at some point we could think of interfacing with Boost.Histogram python package properly, but this is for a later stage I think.

PR #4974, opened 2026-01-15
feat!: template<holder_t> for PodioTrack(State)Container

wdconinc:
Implement template<holder_t> for the PodioTrack(State)Container, in order to allow internally and externally owned podio collections for tracks and track states.

--- END COMMIT MESSAGE ---

This modifies
```cpp
class MutablePodioTrackContainer;
```
to
```cpp
template <template <typename> class holder_t>
class MutablePodioTrackContainer;
```
where the `holder_t` is the (newly public) `RefHolder`, `ConstRefHolder`, `ValueHolder`, or `std::unique_ptr` or `std::shared_ptr`. The same is applied to `ConstPodioTrackContainer`, and the corresponding `PodioTrackStateContainer` classes as well.

The defaults are chosen to remain the same as before: `std::unique_ptr` for owning `MutablePodioTrackContainer`, and `ConstRefHolder` for a viewing `ConstPodioTrackContainer`. Nevertheless, this is a breaking change from a class to a template.

This can be used downstream as, e.g.:
```cpp
  using PodioTrackContainer = Acts::TrackContainer<
      ActsPlugins::MutablePodioTrackContainer<Acts::RefHolder>,
      ActsPlugins::MutablePodioTrackStateContainer<Acts::RefHolder>,
      std::shared_ptr>;

  auto trackStateContainer = 
      std::make_shared<ActsPlugins::MutablePodioTrackStateContainer<Acts::RefHolder>>(
          helper, 
          Acts::RefHolder{*output_track_states}, 
          Acts::RefHolder{*output_track_parameters}, 
          Acts::RefHolder{*output_track_jacobians});
  auto trackContainer =
      std::make_shared<ActsPlugins::MutablePodioTrackContainer<Acts::RefHolder>>(
          helper,
          Acts::RefHolder{*output_tracks});

  PodioTrackContainer acts_tracks(trackContainer, trackStateContainer);
```


paulgessinger:
@wdconinc I've updated this a bit and fixed the remaining CI issues I think.

wdconinc:
> @wdconinc I've updated this a bit and fixed the remaining CI issues I think.

Yeah, sorry about that. The semester started and I got consumed by other stuff.

andiwand (approved):
LGTM 👍

wdconinc:
Over the weekend I was able to come back to this branch and I was able to make the full ActsPodioEdm output of Acts internal types work in our framework with these reference holders.

The only place where we still have to make a data copy now (podio clone) is due to our use of a temporary container for CKF track finding/fitting so we can avoid filling the output container with low quality tracks and track states.

And I have to figure out the dynamic columns that we use for the seed number (even if that isn't persistent).

PR #4981, opened 2026-01-16
refactor: Add policy state to navigation system

paulgessinger:
This PR adds volume-local state to the navigation policies. The way this works is that on volume entry, each policy is asked to create a (possibly empty) state object. The Navigator will then ask the current volume's policy at each step whether the navigation solution is still "valid", where the policy defines what that means. If the policy returns a negative result, the Navigator will renavigate. The policy is allowed to modify the state in `isValid` and in the navigation resolution method as well.

--- END COMMIT MESSAGE ---

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -118,6 +118,18 @@ jobs:
           ls -al build
           ls -al build/coverage
 
+      - name: Verify coverage file integrity
+        run: |
+          echo "--- File info ---"
+          ls -lh build/coverage/cov.xml
+          md5sum build/coverage/cov.xml
+          echo "--- Validate XML ---"
+          xmllint --noout build/coverage/cov.xml && echo "XML is valid" || echo "XML is INVALID"
+          echo "--- Check for truncation ---"
+          tail -5 build/coverage/cov.xml
+          echo "--- Verify closing tag ---"
+          grep '</coverage>' build/coverage/cov.xml && echo "Has closing tag" || echo "MISSING CLOSING TAG"
+
       - name: Run sonar-scanner (PR mode)
         uses: SonarSource/sonarqube-scan-action@v7.0.0
         if: github.event.workflow_run.event == 'pull_request'
@@ -126,6 +138,7 @@ jobs:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }} # Put the name of your token here
         with:
           args: >
+            --define sonar.verbose=true
             --define sonar.cfamily.compile-commands=build/compile_commands.json
             --define sonar.coverageReportPaths=build/coverage/cov.xml
             --define sonar.projectVersion=${{ steps.next_version.outputs.content }}
@@ -142,6 +155,7 @@ jobs:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }} # Put the name of your token here
         with:
           args: >
+            --define sonar.verbose=true
             --define sonar.cfamily.compile-commands=build/compile_commands.json
             --define sonar.coverageReportPaths=build/coverage/cov.xml
             --define sonar.projectVersion=${{ steps.next_version.outputs.content }}


Core/include/Acts/Navigation/INavigationPolicy.hpp
```diff
@@ -68,6 +181,29 @@ class INavigationPolicy {
...
+      const GeometryContext& /*gctx*/, const NavigationArguments /*args*/,
+      NavigationPolicyStateManager& /*stateManager*/,
+      const Logger& logger) const {
+    ACTS_VERBOSE("Default navigation policy state initialization. (noop)");
+    return {};
+  }
+
+  virtual void popState(NavigationPolicyStateManager& /*stateManager*/,
+                        const Logger& logger) const {
+    // By default, we didn't push anything, so we don't need to poop anything
```
paulgessinger, line 242:
```suggestion
    // By default, we didn't push anything, so we don't need to pop anything
```

🤣

Core/include/Acts/Navigation/INavigationPolicy.hpp
```diff
@@ -19,20 +20,131 @@ namespace Acts {
...
+  }
+
+  template <typename T>
+  const T& as() const {
+    return std::any_cast<T&>(payload());
+  }
+
+  NavigationPolicyState() = default;
+
+  bool empty() const { return m_manager == nullptr; }
```
andiwand, line 53: interface wise it would seem easier to me if this was not *empty* constructible. couldn't every policy return a `NavigationPolicyState` object that points to the last stack entry unconditionally? in case there is no state it could populate an empty type which is not read anywhere and should not be costly due to SBO

Core/include/Acts/Navigation/INavigationPolicy.hpp
```diff
@@ -19,20 +20,131 @@ namespace Acts {
...
+                        std::size_t index)
+      : m_manager(&manager), m_index(index) {}
+
+  std::any& payload();
+  const std::any& payload() const;
+
+  NavigationPolicyStateManager* m_manager = nullptr;
+  std::size_t m_index = 0;
+
+  friend class NavigationPolicyStateManager;
```
andiwand, line 66: why do we need `friend` here?

Core/include/Acts/Navigation/INavigationPolicy.hpp
```diff
@@ -19,20 +20,131 @@ namespace Acts {
...
+          "NavigationPolicyStateManager: Attempt to pop from empty stack");
+    }
+    m_stateStack.pop_back();
+  }
+
+ private:
+  // We might want to extend this to a stack
+  std::vector<std::any> m_stateStack;
+
+  friend class NavigationPolicyState;
```
andiwand: `friend` could be avoided if the strack itself would be passed on or if the manager provided an interface for accessing the stack (I guess this is less favorable though)

also the implementation would decouple and there would be no need for the separate definitions below

Core/include/Acts/Navigation/INavigationPolicy.hpp
```diff
@@ -19,20 +20,131 @@ namespace Acts {
...
+  std::any& payload();
+  const std::any& payload() const;
+
+  NavigationPolicyStateManager* m_manager = nullptr;
+  std::size_t m_index = 0;
+
+  friend class NavigationPolicyStateManager;
+};
+
+class NavigationPolicyStateManager {
```
andiwand, line 73: nitpick: *manager* makes me think this class orchestrates something but it is more like a container, right? maybe `NavigationPolicyStateContainer` / `NavigationPolicyStateStack`

Core/include/Acts/Navigation/MultiNavigationPolicy.hpp
```diff
@@ -54,6 +57,98 @@ class MultiNavigationPolicy final : public INavigationPolicy {
...
+    if (states.size() != m_policyPtrs.size()) {
+      ACTS_ERROR("MultiNavigationPolicy popState: number of states ("
+                 << states.size() << ") does not match number of policies ("
+                 << m_policyPtrs.size() << ").");
+      throw std::runtime_error(
+          "MultiNavigationPolicy popState: inconsistent state size.");
+    }
+
+    // @TODO: Possibly use updated zip | reverse
+    for (std::size_t i = states.size(); i-- > 0;) {
```
andiwand: I guess this approach saves us from doing `0 - 1`?

Core/src/Navigation/Navigator.cpp
```diff
@@ -452,8 +476,27 @@ void Navigator::resolveCandidates(State& state, const Vector3& position,
...
   args.direction = direction;
+
+  const INavigationPolicy* policy = state.currentVolume->navigationPolicy();
+  if (policy == nullptr) {
+    ACTS_ERROR(volInfo(state) << "No navigation policy found for volume "
+                              << state.currentVolume->geometryId()
+                              << ". Cannot resolve navigation candidates.");
+    // @TODO: Should this be an exception?
+    return;
+  }
```
andiwand, line 545: might be a case for an `assert`?

Core/src/Navigation/Navigator.cpp
```diff
@@ -452,8 +476,27 @@ void Navigator::resolveCandidates(State& state, const Vector3& position,
...
+
+  // This might be the first volume, check if we have a current state
+  if (!state.policyStateManager.currentState().empty()) {
+    policy->popState(state.policyStateManager, logger());
+  }
+
+  policy->createState(state.options.geoContext, args, state.policyStateManager,
+                      logger());
+
+  auto policyState = state.policyStateManager.currentState();
```
andiwand:
```suggestion
  auto policyState = policy->createState(state.options.geoContext, args, state.policyStateManager,
                      logger());

```

Core/include/Acts/Navigation/INavigationPolicy.hpp
```diff
@@ -19,20 +20,131 @@ namespace Acts {
...
+    return std::pair<T&, NavigationPolicyState>{
+        content, {*this, m_stateStack.size() - 1}};
+  }
+
+  friend class Navigator;
+
+  NavigationPolicyState currentState() {
+    if (m_stateStack.empty()) {
+      return {};  // Emtpy state as sentinel
+    }
```
andiwand, line 93: in case we opt for every policy pushing to the stack this should trigger an exception

Core/include/Acts/Navigation/INavigationPolicy.hpp
```diff
@@ -11,6 +11,7 @@
 #include "Acts/Geometry/GeometryContext.hpp"
 #include "Acts/Navigation/NavigationDelegate.hpp"
 #include "Acts/Navigation/NavigationStream.hpp"
+#include "Acts/Utilities/Any.hpp"
```
andiwand, line 14: ends up not being used?

Core/src/Navigation/Navigator.cpp
```diff
@@ -401,7 +410,22 @@ NavigationTarget Navigator::getNextTargetGen1(State& state,
...
                                               const Vector3& position,
                                               const Vector3& direction) const {
-  if (!state.navCandidateIndex.has_value()) {
+  if (state.currentVolume == nullptr) {
+    ACTS_VERBOSE(volInfo(state) << "No volume to get next target.");
+    return NavigationTarget::None();
+  }
+
+  auto policyState = state.policyStateManager.currentState();
+  bool isValid = state.policyStateManager.currentState().empty() ||
```
andiwand:
```suggestion
  bool isValid = policyState.empty() ||
```

Core/src/Navigation/Navigator.cpp
```diff
@@ -401,7 +410,22 @@ NavigationTarget Navigator::getNextTargetGen1(State& state,
...
+                 state.currentVolume->navigationPolicy()->isValid(
+                     state.options.geoContext,
+                     {.position = position, .direction = direction},
+                     policyState, logger());
```
andiwand: do we only want to check this when we switch targets? I thought it might also be necessary to check this after individual propagation steps. but it might be a fair assumption that the state should be valid as long as the current target is valid?

Core/include/Acts/Navigation/INavigationPolicy.hpp
```diff
@@ -19,20 +20,131 @@ namespace Acts {
 class TrackingVolume;
 class INavigationPolicy;
 class Surface;
+class Navigator;
+
+class NavigationPolicyStateManager;
+
+class NavigationPolicyState {
```
andiwand, line 31: nitpick: this is not really the *state* but a pointer or a proxy to the state. maybe `NavigationPolicyStateProxy`?

Core/include/Acts/Navigation/INavigationPolicy.hpp
```diff
@@ -19,20 +20,131 @@ namespace Acts {
...
+
+template <detail::HasOldInitializeCandidates T>
+void oldToNewSignatureAdapter(const void* instance, const GeometryContext& gctx,
+                              const NavigationArguments& args,
+                              NavigationPolicyState& /*state*/,
+                              AppendOnlyNavigationStream& stream,
+                              const Logger& logger) {
+  const auto* policy = static_cast<const T*>(instance);
+  policy->initializeCandidates(gctx, args, stream, logger);
+}
```
andiwand, line 158:
```suggestion
template <detail::HasOldInitializeCandidates T>
void oldToNewSignatureAdapter(const T* policy, const GeometryContext& gctx,
                              const NavigationArguments& args,
                              NavigationPolicyState& /*state*/,
                              AppendOnlyNavigationStream& stream,
                              const Logger& logger) {
  policy->initializeCandidates(gctx, args, stream, logger);
}
```
isn't this equivalent? or couldn't the caller site do the `static_cast`?

Core/include/Acts/Navigation/INavigationPolicy.hpp
andiwand: thinking about it further, right now a policy could receive a state from another policy if the parent volume populates a state but the child volume does not

Core/include/Acts/Navigation/MultiNavigationPolicy.hpp
andiwand, line 1: all of this seems quite involved. is it possible / would it be easier if the `MultiNavigationPolicy::State` would hold a `NavigationPolicyStateManager` that is presented to the containing policies? this way there would only be a single state per policy on the stack

Core/include/Acts/Navigation/INavigationPolicy.hpp
paulgessinger: I don't think this can happen, since the policy is consulted whether to push or pop a state from the stack. The assumption is that if a policy does not push a state, it will not try to access it.

This is not enforced though, and you're right that it's maybe suboptimal.

Core/include/Acts/Navigation/INavigationPolicy.hpp
paulgessinger: Yes, I wanted to use it but switched to `std::any` for debugging. Ultimately I'll switch back, but fair point.

Core/include/Acts/Navigation/INavigationPolicy.hpp
paulgessinger: Fundamentally, you're right, but I was trying to hide the fact that it's a proxy to the policy. Maybe irrelevant though.

Core/include/Acts/Navigation/INavigationPolicy.hpp
paulgessinger: Might not be needed anymore.

Core/include/Acts/Navigation/INavigationPolicy.hpp
paulgessinger: It should / will / might ultimately manage allocations as well using (maybe) a bump allocator. *manager* sounded appropriate to me, but happy to call it container or stack.

Core/include/Acts/Navigation/INavigationPolicy.hpp
paulgessinger: The idea was that the stack can only be accessed through the state (proxy), to avoid interference from other code. Can revisit though.

Core/include/Acts/Navigation/INavigationPolicy.hpp
paulgessinger: I need a function that I can register into the delegate, and that needs to have this signature with the `void*`. I can register a lambda at the call site, but I thought this was cleaner.

Core/include/Acts/Navigation/MultiNavigationPolicy.hpp
paulgessinger: I didn't consider the approach of unconditionally pushing a state for all policies. That might be a good approach, and probably also work with a future bump allocator and empty 0-sized structs.

I'm not sure I understand the point of passing the manager inside the state.

Core/src/Navigation/Navigator.cpp
paulgessinger: Yes, good point! That's something we need to check I guess. From an interface point of view (for the policy interface) this would be transparent, i.e. we can call `isValid` *whenever* without the policies being aware of it.

Core/src/Navigation/Navigator.cpp
paulgessinger: Yeah makes sense, the return value changed in a few iterations, with the current version this would work.

paulgessinger:
I've added a unit test that also tries out a sort of cone based navigation. This just adds a cone based invalidator that sets a cone width.

I use the following test geometry:

Screenshot 2026-02-03 at 16 19 53

Highlighted are the trajectory and the portals that are hit. There's also some more basic tests in there (one that flip-flops between valid and invalid, and another one with the cone invalidator but a much stronger b field that doesn't curve as much. in that case, the test checks that there are not renavigations.)


@andiwand @dimitra97

Core/src/Navigation/Navigator.cpp
```diff
@@ -236,12 +247,28 @@ NavigationTarget Navigator::nextTarget(State& state, const Vector3& position,
   return NavigationTarget::None();
 }
 
-bool Navigator::checkTargetValid(const State& state, const Vector3& position,
+bool Navigator::checkTargetValid(State& state, const Vector3& position,
                                  const Vector3& direction) const {
+  ACTS_VERBOSE(volInfo(state) << "Entering Navigator::checkTargetValid.");
   static_cast<void>(position);
   static_cast<void>(direction);
```
andiwand:
```suggestion
```

Core/src/Navigation/Navigator.cpp
```diff
@@ -473,8 +537,19 @@ void Navigator::resolveCandidates(State& state, const Vector3& position,
...
   args.position = position;
   args.direction = direction;
+
+  const INavigationPolicy* policy = state.currentVolume->navigationPolicy();
+  if (policy == nullptr) {
+    ACTS_ERROR(volInfo(state) << "No navigation policy found for volume "
+                              << state.currentVolume->geometryId()
+                              << ". Cannot resolve navigation candidates.");
+    // @TODO: Should this be an exception?
+    return;
```
andiwand: I think we can throw in this case or convert it to an `assert`

andiwand:
@paulgessinger can you add a PR description?

Core/src/Navigation/Navigator.cpp
```diff
@@ -473,8 +537,19 @@ void Navigator::resolveCandidates(State& state, const Vector3& position,
   NavigationArguments args;
   args.position = position;
   args.direction = direction;
+
+  const INavigationPolicy* policy = state.currentVolume->navigationPolicy();
+  if (policy == nullptr) {
+    ACTS_ERROR(volInfo(state) << "No navigation policy found for volume "
+                              << state.currentVolume->geometryId()
```
dimitra97: should we log the volumeName() instead?

Tests/UnitTests/Core/Propagator/NavigationPolicyStateTests.cpp
```diff
@@ -0,0 +1,486 @@
...
+    propOpts.stepping.maxStepSize = 100_mm;
+
+    auto result = propagator.propagate(start, propOpts);
+    BOOST_REQUIRE(result.ok());
+
+    const auto& info =
+        result.value().template get<StepCollector::this_result>();
+
+    // Optional trajectory visualization
+    const auto& positions = info.position;
```
dimitra97, line 447: is the visualization really optional here? as I understood since we have the positions we go to track visualization, no? unless we want to visualize eitherway:)

Core/src/Navigation/Navigator.cpp
paulgessinger: `volInfo(state)` will already log the name. I'll remove the geometry id output, actually.

Tests/UnitTests/Core/Propagator/NavigationPolicyStateTests.cpp
paulgessinger: It was optional at one point. I'll remove that. I think visualizing always is fine for a test.

Core/src/Navigation/Navigator.cpp
paulgessinger: Throwing now.

paulgessinger:
@dimitra97 @andiwand should we proceed with this?

PR #4970, opened 2026-01-15
refactor: DetectorElementBase - Introduce isSensitive method && define localToGlobalTransform in the child classes

junggjo9:
- localToGlobalTransform is the primary interface method since https://github.com/acts-project/acts/pull/4958
- Introduce a isSensitive() method to the DetectorElementBase & the Surface

--- END COMMIT MESSAGE ---
Tagging @andiwand

PR #4985, opened 2026-01-19
refactor: Remove `virtual` from `Surface::isSensitive`

andiwand:
We do not expect or want surfaces to overwrite this standard behavior so `virtual` should be dropped. After https://github.com/acts-project/acts/pull/4970

--- END COMMIT MESSAGE ---

@junggjo9

### DIFF
--- a/Core/include/Acts/Surfaces/Surface.hpp
+++ b/Core/include/Acts/Surfaces/Surface.hpp
@@ -445,7 +445,7 @@ class Surface : public virtual GeometryObject,
   virtual std::string name() const = 0;
 
   /// @brief Returns whether the Surface is sensitive
-  virtual bool isSensitive() const;
+  bool isSensitive() const;
 
   /// Return a Polyhedron for surface objects
   ///


junggjo9:
Hmm, I was thinking about exploting the virtualness of the  `isSensitive` method in https://github.com/acts-project/acts/pull/4494. I've made introduced a sensitive surface class which just overridesthis method. Dunno whether this idea is smart. Depends how we  want to define the API of isSensitive in the fitters

andiwand:
> Hmm, I was thinking about exploting the virtualness of the `isSensitive` method in #4494. I've made introduced a sensitive surface class which just overridesthis method. Dunno whether this idea is smart. Depends how we want to define the API of isSensitive in the fitters

I thought the `isSensitive` state should strictly come from the detector element. The problem with a `SensitiveSurface` class is that we would need diamond inheritance to make this happen, no? Like (`Surface` -> `SensitiveSurface`) (`Surface` -> `PlaneSurface`) -> `SensitivePlaneSurface`

For free surfaces this will force us to create temporary `DetectorElement`s, but I think this is what we would expect with the current design? @paulgessinger @asalzburger

PR #4986, opened 2026-01-19
fix: Correct links in getting_started.md

benjaminhuth:
PLEASE DESCRIBE YOUR CHANGES.
DO NOT USE @-MENTIONS HERE!

--- END COMMIT MESSAGE ---

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -56,7 +56,7 @@ For convenience, it is possible to build the required boost and eigen3 dependenc
 Other options are also
 available and are discussed in the [Building ACTS](#building-acts) section.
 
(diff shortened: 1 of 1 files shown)


paulgessinger:
I think the links are correct, **but** the getting started page is potentially not linked in the doc at the moment.

benjaminhuth:
Hmm I think this cannot work when coming to the getting_started.md from the github README.md view, because it tries to link to a local anchor...

paulgessinger:
~~Where do we get there from GitHub?~~

Ah, from the README. I suppose that should link to the rendered documentation then? Otherwise we could also keep that part of the documentation outside of doxygen.

benjaminhuth:
Yeah, that would probably be good

PR #4987, opened 2026-01-19
fix: Duplicate layers in HoughCell

kciesla93:
This PR fixes duplicate layers in `HoughCell` object:

- Added check for duplicate layers using the similar logic as for checking the hit duplicates.
- Added test.

--- END COMMIT MESSAGE ---

Tagging @goblirsc and @junggjo9 since this code is used by muons (also in Athena), but since this was unnoticed, I guess it was not used for anything critical.


Core/include/Acts/Seeding/HoughTransformUtils.hpp
```diff
@@ -126,19 +125,14 @@ class HoughCell {
...
 
-  /// index for the hits -- keeps track of vector's size
-  std::size_t m_iHit{0};
-  /// index for the layers -- keeps track of vector's size
-  std::size_t m_iLayer{0};
-
-  /// a batch to resize the vector of the hits or the layers
-  std::size_t m_assignBatch{20};
+  /// a batch to resize the vector of the hits
+  static constexpr std::size_t m_assignBatch{20};
```
junggjo9: I don't see advantage of enforcing the assignment at compile time.

Core/include/Acts/Seeding/HoughTransformUtils.ipp
```diff
@@ -43,19 +43,25 @@ void Acts::HoughTransformUtils::HoughPlane<identifier_t>::fill(
...
+Acts::HoughTransformUtils::HoughCell<identifier_t>::HoughCell() {
+  m_layers.reserve(m_assignBatch);
+  m_hits.reserve(m_assignBatch);
+}
+
 template <class identifier_t>
 std::span<const identifier_t, std::dynamic_extent>
 Acts::HoughTransformUtils::HoughCell<identifier_t>::getHits() const {
   std::span<const identifier_t, std::dynamic_extent> hits(m_hits.begin(),
-                                                          m_iHit);
```
junggjo9, line 50: The idea was to reuse the same hough cell without reallocating the memory

Core/include/Acts/Seeding/HoughTransformUtils.ipp
```diff
@@ -65,35 +71,32 @@ void Acts::HoughTransformUtils::HoughCell<identifier_t>::fill(
...
 
 template <class identifier_t>
 void Acts::HoughTransformUtils::HoughCell<identifier_t>::reset() {
   m_nHits = 0;
   m_nLayers = 0;
 
-  m_iLayer = 0;
-  m_iHit = 0;
+  m_hits.clear();
+  m_layers.clear();
```
junggjo9: The index manipulation was developed because we saw it performed better. @dimitra97

Core/include/Acts/Seeding/HoughTransformUtils.ipp
```diff
@@ -65,35 +71,32 @@ void Acts::HoughTransformUtils::HoughCell<identifier_t>::fill(
...
-
+  m_hits.push_back(identifier);
   m_nHits += weight;
-  m_nLayers += weight;
+
+  // Check for duplicates
+  if (std::find(m_layers.begin(), m_layers.end(), layer) == m_layers.end()) {
+    m_layers.push_back(layer);
+    m_nLayers += weight;
+  }
```
junggjo9: I foree here already a big CPU bottleneck. Why not checking that the last layer numer corresponds to the current one. Same logic as for the hit filling?

Tests/UnitTests/Core/Seeding/HoughTransformTest.cpp
```diff
@@ -239,6 +239,28 @@ BOOST_AUTO_TEST_CASE(hough_transform_sliding_window) {
...
+  const std::size_t nY = 1;
+  HoughTransformUtils::HoughPlaneConfig config{nX, nY};
+  HoughTransformUtils::HoughPlane<uint8_t> plane(config);
+
+  uint8_t nHits = 0;
+  static constexpr uint8_t nLayers = 10;
+  for (uint8_t layer = 1; layer <= nLayers; ++layer) {
+    // Add hits equal to the layer number
+    for (uint8_t hit = 0; hit < layer; ++hit) {
+      plane.fillBin(0, 0, nHits++, layer);
```
junggjo9, line 256: Here you loop layer by layer. There's'no need to loop through the entire vector of [0-8] if you're filling layer 9. Just check whether the last layer number is different. If people are jumping between the layers, they must do the duplicate removal themselves in the peak finder!

Core/include/Acts/Seeding/HoughTransformUtils.ipp
dimitra97: indeed

Core/include/Acts/Seeding/HoughTransformUtils.ipp
kciesla93: @junggjo9 Thanks for the comments. I've restored the previous behaviour and included the check for the layers with the same logic as for the hit filling.

junggjo9:
Thanks comments from the CI still need to be addressed

junggjo9 (approved):

PR #4998, opened 2026-01-21
refactor: Separate compilation of muon digitization

benjaminhuth:
The muon digitization algorithm depends on ROOT, and complilation needs to be separated to enable compilation of examples without ROOT.

--- END COMMIT MESSAGE ---

@junggjo9


paulgessinger:
Similar to my comment in #5001, do we need these in separate targets necessarily?

benjaminhuth:
Let's focus discussion on #5001

PR #5001, opened 2026-01-21
refactor: Move `ScalingCalibrator to Framework/ROOT subpackage

benjaminhuth:
Step towards compiling examples without ROOT.

--- END COMMIT MESSAGE ---

### DIFF
--- a/Examples/Framework/ROOT/CMakeLists.txt
+++ b/Examples/Framework/ROOT/CMakeLists.txt
@@ -1,7 +1,6 @@
 acts_add_library(
     ExamplesFrameworkRoot
     src/ScalingCalibrator.cpp
-    src/Helpers.cpp
     ACTS_INCLUDE_FOLDER include/ActsExamples
 )
 


benjaminhuth:
@andiwand @paulgessinger Any thoughts about this structure? Now we would have two sub-packages of Framework for two different calibrators.

Maybe it would be more economic to implement the actual calibration functions in `Plugins/{Onnx|Root}`, and just use those functions to compose the calibrators downstream instead of branching in the framework this way...

paulgessinger:
@benjaminhuth Maybe we should just have a general `ROOT` target in the Examples, rather than now having one IO and one "Framework" one. In the end it's not beneficial to have them separate, is it? The only thing that matters is whether we have ROOT available or not.

benjaminhuth:
@paulgessinger But where would you place the code? would you create a general `Examples/Root` or `Examples/Tools/ROOT` or similar package?

paulgessinger:
@benjaminhuth Not sure, but I don't think it's super clean to have single-source-file shared libraries just so we can get the dependencies separated. I suppose the actual solution is to break out the ROOT dependencies through refactoring, and moving the ROOT-dependent code into common targets / libraries.

Examples/Framework/ROOT/include/ActsExamples/EventData/ScalingCalibrator.hpp
paulgessinger, line 1: Also nit: ROOT -> Root

benjaminhuth:
I think the directory structure of Examples is not really allowing for this in a nice way...

@andiwand  proposed something like `Examples/Root`, and one could then do
- `Examples/Root/Io`
- `Examples/Root/Tools`
- `Examples/Root/Algorithms`

PR #4996, opened 2026-01-21
refactor: Relax `KalmanFitter` error logs

andiwand:
Since all the error cases in the `KalmanFitter` are propagated to the user, we can let them decide how severe the error is. One limitation on the severity comes from the ATLAS reconstruction software which will flag a reconstruction job as failed if a single error occurs, depending on the configuration even warnings. Since the reconstruction can continue and we would "only" lose some efficiency, which has to be monitored separately anyways, I would argue that these error cases should not escalate by default.

--- END COMMIT MESSAGE ---

PR #5004, opened 2026-01-22
refactor: Relax Core error logs for Propagation, Navigation, Fitter, Finder

andiwand:
Similar to https://github.com/acts-project/acts/pull/4996

Since most error cases in these components are propagated to the user, we can let them decide how severe the error is. One limitation on the severity comes from the ATLAS reconstruction software which will flag a reconstruction job as failed if a single error occurs, depending on the configuration even warnings. Since the reconstruction can continue and we would "only" lose some efficiency, which has to be monitored separately anyways, I would argue that these error cases should not escalate by default.

--- END COMMIT MESSAGE ---

### DIFF
--- a/Core/include/Acts/Geometry/ContainerBlueprintNode.hpp
+++ b/Core/include/Acts/Geometry/ContainerBlueprintNode.hpp
@@ -142,10 +142,12 @@ class ContainerBlueprintNode : public BlueprintNode {
  protected:
   /// Make the volume stack for the container. This is called by the build
   /// method and is implemented by the derived classes.
+  /// @param gctx The current geometry context object, e.g. alignment
   /// @param volumes The volumes to stack
   /// @param logger The logger to use
   /// @return The volume stack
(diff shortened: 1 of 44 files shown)


Core/include/Acts/Propagator/TryAllNavigator.hpp
```diff
@@ -221,7 +222,9 @@ class TryAllNavigatorBase {
       if (state.currentVolume != nullptr) {
         ACTS_VERBOSE(volInfo(state) << "Start volume resolved.");
       } else {
-        ACTS_ERROR("Start volume not resolved.");
+        ACTS_DEBUG("Start volume not resolved.");
+        state.navigationBreak = true;
```
benjaminhuth, line 226: Do we need this break? I would hope this would  propagate the error back directely

andiwand: I don't think we need it really, I just saw that we also do that in the `Navigator` so I synced it up here

PR #5005, opened 2026-01-22
feat: Introduce the SurfacePlacementBase interface class

junggjo9:
- Introduce the SurfacePlacementBase class
- Replace the useage of DetectorElementBase by the SurfacePlacementBase
- Add the possibility to store the thickness of the surface along the normal direction which is called `depth()` to avoid ambiguity clashes with the Layer thickness definiton, caused by the multiple inheritance of DiscLayer from DiscSurface and Layer.
--- END COMMIT MESSAGE ---

This PR fundamentally changes how `Surfaces`  can be declared as alignable by the experiment. The `DetectorElementBase` is replaced by the `SurfacePlacementBase` interface class. The main reason is the advent of the alignable volumes where it will become possible to let `TrackingVolumes` float with the alignment. The surfaces representing the portals hence need to be declared also alignable and therefore the term `DetectorElementBase` does no longer match the reality what this interface class is actually representing.

How to migrate from `DetElementBase` to  `SurfacePlacement`
--------------------------------------------------------------------

1.) Replace all occurances of 
```cpp
#include "Acts/Geometry/DetectorElementBase.hpp"
```
by 
```cpp
#include "Acts/Surfaces/SurfacePlacementBase.hpp"
```
2.) Let your experiment specific implementation inherit from `SurfacePlacementBase`
3.) Rename
```cpp
const Transform3& transform(const GeometryContext& gctx) const
```
to
```cpp
const Transform3& localToGlobalTransform(const GeometryContext& gctx) const 
```

4.) Override the `isSensitive()`  method

```
 virtual bool isSensitive() const override { return true; }
```

5.) The `thickness()` method is dropped from the `SurfacePlacementBase`  interface. Everytime when you want to parse the `thickness()` towards the trackingGeometry construction, do it via calling `assignSurfaceDepth` directly on the surface.

Tagging: @andiwand, @asalzburger, @paulgessinger, @benjaminhuth

### DIFF
--- a/Core/include/Acts/TrackFitting/GlobalChiSquareFitter.hpp
+++ b/Core/include/Acts/TrackFitting/GlobalChiSquareFitter.hpp
@@ -852,8 +852,8 @@ class Gx2Fitter {
         if (scatteringMapId == scatteringMap->end()) {
           ACTS_DEBUG("    ... create entry in scattering map.");
 
-          detail::PointwiseMaterialInteraction interaction(state, stepper,
-                                                           navigator);
+          Acts::detail::PointwiseMaterialInteraction interaction(state, stepper,
+                                                                 navigator);
           // We need to evaluate the material to create the correct slab
           const bool slabIsValid =
               interaction.evaluateMaterialSlab(MaterialUpdateMode::FullUpdate);


Core/include/Acts/Geometry/DetectorElementBase.hpp
```diff
@@ -8,37 +8,31 @@
...
-
   /// Return the transform for the Element proxy mechanism
   ///
   /// @param gctx The current geometry context object, e.g. alignment
   /// @return reference to the transform of this detector element
   [[deprecated(
       "Please use localToGlobalTransform(const GeometryContext& gctx) "
       "instead")]]
-  virtual const Transform3& transform(const GeometryContext& gctx) const {
+  const Transform3& transform(const GeometryContext& gctx) const {
```
paulgessinger: Does this have to be `virtual` still so that it's not hard breaking? I think keeping it virtual and keeping the default implementation would work? I don't know how the deprecation behaves in that case now.

Core/include/Acts/Surfaces/Surface.hpp
```diff
@@ -212,13 +217,21 @@ class Surface : public virtual GeometryObject,
...
 
+  /// @brief Return the associated surface placement if there is any
+  const SurfacePlacementBase* surfacePlacement() const;
+
   /// Return method for the associated Layer in which the surface is embedded
   /// @return Layer by plain pointer, can be nullptr
   const Layer* associatedLayer() const;
 
+  /// @brief Return the thickness of the surface in the normal direction
+  double depth() const;
```
paulgessinger: Should we stick to the `thickness` name?

Core/src/Surfaces/PlaneSurface.cpp
```diff
@@ -199,8 +197,7 @@ std::pair<std::shared_ptr<PlaneSurface>, bool> PlaneSurface::mergedWith(
   ACTS_VERBOSE("Merging plane surfaces in " << axisDirectionName(direction)
                                             << " direction");
 
-  if (m_associatedDetElement != nullptr ||
-      other.m_associatedDetElement != nullptr) {
+  if (surfacePlacement() != nullptr || other.surfacePlacement() != nullptr) {
```
paulgessinger: Does it make sense to already put that behind a `isAlignable()` method now similar to `isSensitive()`?

Examples/Detectors/GenericDetector/include/ActsExamples/GenericDetector/GenericDetectorElement.hpp
```diff
@@ -29,7 +29,7 @@ namespace ActsExamples {
 /// This is a lightweight type of detector element,
 /// it simply implements the base class.
 ///
-class GenericDetectorElement : public Acts::DetectorElementBase {
+class GenericDetectorElement : public Acts::SurfacePlacementBase {
```
paulgessinger, line 32: Should we also change these names then?

Examples/Detectors/TGeoDetector/src/TGeoDetector.cpp
```diff
@@ -161,7 +161,7 @@ std::vector<ActsPlugins::TGeoLayerBuilder::Config> makeLayerBuilderConfigs(
 /// @param vm is the variable map from the options
 std::shared_ptr<const TrackingGeometry> buildTGeoDetector(
     const TGeoDetector::Config& config, const GeometryContext& context,
-    std::vector<std::shared_ptr<const DetectorElementBase>>& detElementStore,
+    std::vector<std::shared_ptr<const SurfacePlacementBase>>& detElementStore,
```
paulgessinger, line 164: Consider the name here.

Tests/CommonHelpers/include/ActsTests/CommonHelpers/DetectorElementStub.hpp
```diff
@@ -28,7 +28,7 @@ namespace ActsTests {
 ///
 /// This is a lightweight type of detector element,
 /// it simply implements the base class.
-class DetectorElementStub : public Acts::DetectorElementBase {
+class DetectorElementStub : public Acts::SurfacePlacementBase {
```
paulgessinger, line 31: Consider renaming

Core/include/Acts/Surfaces/Surface.hpp
junggjo9: DiscLayer inherits from Layer && DiscSurface and both provide thickness() via the interface.

Core/src/Surfaces/PlaneSurface.cpp
junggjo9: Yeah, in this context this might be worth it

Examples/Detectors/GenericDetector/include/ActsExamples/GenericDetector/GenericDetectorElement.hpp
junggjo9: In these cases it's still a DetectorElement, no?

Core/include/Acts/Geometry/DetectorElementBase.hpp
junggjo9: I've no strong preference for one or the other

Core/include/Acts/Surfaces/Surface.hpp
junggjo9: I've renamed the thickness of the layer (cf. https://github.com/acts-project/acts/pull/5013)

Core/src/Surfaces/PlaneSurface.cpp
junggjo9: Added

Core/include/Acts/Surfaces/Surface.hpp
junggjo9: Changed to thickness

Core/include/Acts/Geometry/DetectorElementBase.hpp
paulgessinger: What I'm saying is that if you remove the `virtual`, client code might not compile without modifications. Can you see if you can keep this `virtual`?

junggjo9:
```suggestion
  virtual const Transform3& transform(const GeometryContext& gctx) const {
```
Sorry my bad I had another snippet of code in my mind. Of course transform only is a method of this one and not declared in surface placement base

PR #5014, opened 2026-01-27
refactor: Extract GSF functions part 1

andiwand:
Extracts a couple of functions from `GsfActor.hpp` to `GsfUtil.hpp`, to later make them available to the CKF and add support for the handling of bremstrahlung during track finding. This first part covers
- `removeLowWeightComponents `
- `updateStepper`
- `convoluteComponents`
- `applyBetheHeitler`

--- END COMMIT MESSAGE ---


Core/include/Acts/TrackFitting/detail/GsfUtils.hpp
```diff
@@ -256,5 +252,132 @@ class Updatable {
   const T &val() const { return m_val; }
 };
 
-}  // namespace detail
-}  // namespace Acts
+namespace gsf {
```
paulgessinger:
```suggestion
namespace Gsf {
```
?

Core/include/Acts/TrackFitting/detail/GsfUtils.hpp
```diff
@@ -256,5 +252,132 @@ class Updatable {
...
+    cmp.derivative() = FreeVector::Zero();
+    cmp.jacTransport() = FreeMatrix::Identity();
+  }
+}
+
+double applyBetheHeitler(
+    const GeometryContext &geoContext, const Surface &surface,
+    Direction direction, const BoundTrackParameters &initialParameters,
+    const double initialWeight, const BetheHeitlerApprox &betheHeitlerApprox,
+    std::vector<BetheHeitlerApprox::Component> &betheHeitlerCache,
```
benjaminhuth, line 340: I wonder if it would make things more complicated or less complicated if we would bundle maybe the caches and the updatables together in some struct...

andiwand: yeah the interface looks pretty depressing. I don't see a great alternative right now. potentially there could be an overload which eats the propagator state and so on to extract some of the parameters.

for now I would leave it as is and not create extra structs since it is a detail and we are in full control of the interface. so we can always revise this when we have a better idea

Core/include/Acts/TrackFitting/detail/GsfUtils.hpp
```diff
@@ -8,36 +8,33 @@
...
-namespace Acts {
-
-/// The tolerated difference to 1 to accept weights as normalized
-constexpr static double s_normalizationTolerance = 1.e-4;
-
-namespace detail {
+namespace Acts::detail {
 
 template <typename component_range_t, typename projector_t>
 bool weightsAreNormalized(const component_range_t &cmps,
```
benjaminhuth, line 36: I wonder if we should move those functions also into `Acts::detail::Gsf`? Then we would have a clear structure that things from `GsfUtils.hpp` belong to `Acts::detail::Gsf`.

andiwand: agreed - I planned to do this in a followup but I can just do it here already

PR #5015, opened 2026-01-27
refactor: Remove `KalmanUpdateHelpers` and refactor KF and GSF

andiwand:
I believe `KalmanUpdateHelpers` is one of the cases where the generalization went too far. I wanted to reuse this component for the CKF+GSF algorithm but could not because it couples tightly to the extensions for example.

This PR ports the functionality to the KF and to the GSF.

Apart from that a couple of other changes were made which could be broken out
- `const Surface &` over `const Surface *` where possible
- clean some includes

--- END COMMIT MESSAGE ---

blocked by
- https://github.com/acts-project/acts/pull/5016

### DIFF
--- a/Core/include/Acts/TrackFitting/KalmanFitter.hpp
+++ b/Core/include/Acts/TrackFitting/KalmanFitter.hpp
@@ -186,6 +186,7 @@ struct KalmanFitterOptions {
   FreeToBoundCorrection freeToBoundCorrection;
 };
 
+/// Result payload returned by the Kalman fitter.
 template <typename traj_t>
 struct KalmanFitterResult {
   /// Fitted states that the actor has handled.
--- a/Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
+++ b/Core/include/Acts/AmbiguityResolution/AmbiguityResolutionML.hpp
@@ -24,6 +24,7 @@ namespace Acts {
 template <AmbiguityNetworkConcept AmbiguityNetwork>
 class AmbiguityResolutionML {
  public:
+  /// @brief Configuration for the ambiguity resolution algorithm.
   struct Config {
     /// Path to the model file for the duplicate neural network
     std::string inputDuplicateNN = "";
--- a/Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.hpp
+++ b/Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.hpp
@@ -33,6 +33,7 @@ namespace Acts {
 ///  4) Back to square 1.
 class GreedyAmbiguityResolution {
  public:
+  /// Configuration parameters for greedy ambiguity resolution.
   struct Config {
     /// Maximum amount of shared hits per track.
     std::uint32_t maximumSharedHits = 1;
@@ -43,20 +44,28 @@ class GreedyAmbiguityResolution {
     std::size_t nMeasurementsMin = 7;
   };
 
+  /// Mutable state used by the greedy ambiguity resolution.
   struct State {
+    /// Total number of tracks
     std::size_t numberOfTracks{};
 
+    /// Track tips for each track
     std::vector<int> trackTips;
+    /// Chi-squared value for each track
     std::vector<float> trackChi2;
+    /// Measurement indices for each track
     std::vector<std::vector<std::size_t>> measurementsPerTrack;
 
     // TODO consider boost 1.81 unordered_flat_map
+    /// Map from measurement index to set of track indices using it
     boost::container::flat_map<std::size_t,
                                boost::container::flat_set<std::size_t>>
         tracksPerMeasurement;
+    /// Number of shared measurements for each track
     std::vector<std::size_t> sharedMeasurementsPerTrack;
 
     // TODO consider boost 1.81 unordered_flat_map
+    /// Set of track indices selected as good tracks
     boost::container::flat_set<std::size_t> selectedTracks;
   };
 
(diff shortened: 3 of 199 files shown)


Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -555,52 +554,101 @@ struct GsfActor {
                             const SourceLink& sourceLink) const {
     const auto& surface = *navigator.currentSurface(state.navigation);
 
-    // Boolean flag, to distinguish measurement and outlier states. This flag
-    // is only modified by the valid-measurement-branch, so only if there
-    // isn't any valid measurement state, the flag stays false and the state
-    // is thus counted as an outlier
-    bool is_valid_measurement = false;
+    // Keep track of all created components for outlier handling
+    std::vector<TrackIndexType> allTips;
```
benjaminhuth: I think we could reserve space for all tips here, right?

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -695,50 +757,70 @@ struct GsfActor {
     }
   }
 
+  enum class TrackStateType { eMeasurement, eOutlier, eHole, eNone };
```
benjaminhuth: Hmm we don't have such a thing somewhere else I guess? Feels a bit weird to define it here

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -695,50 +757,70 @@ struct GsfActor {
...
+                             const Surface& surface,
+                             TrackStateType type) const {
     using PrtProjector =
         MultiTrajectoryProjector<StatesType::ePredicted, traj_t>;
     using FltProjector =
         MultiTrajectoryProjector<StatesType::eFiltered, traj_t>;
 
+    const bool isMeasurement = type == TrackStateType::eMeasurement;
+    const bool isOutlier = type == TrackStateType::eOutlier;
+    const bool isHole = type == TrackStateType::eHole;
```
benjaminhuth: Hmm as far as I can see we could define these constants in the if-branch for forward pass. That would make clear that this property is set only in the forward pass.

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
andiwand: as far as I am aware we do not have something like that anywhere else. maybe it makes sense to elevate this to the new `TrackStateType`. but I have to think a bit about naming

andiwand: ok I think I found a better solution

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -768,11 +767,11 @@ struct GsfActor {
     using FltProjector =
         MultiTrajectoryProjector<StatesType::eFiltered, traj_t>;
 
-    const bool isMeasurement = type == TrackStateType::eMeasurement;
-    const bool isOutlier = type == TrackStateType::eOutlier;
-    const bool isHole = type == TrackStateType::eHole;
-
     if (!m_cfg.inReversePass) {
+      const bool isMeasurement = type.isMeasurement();
```
benjaminhuth: Hmm just a thought: Can one just assign `type` to the trackstate somehow instead of calling `setIsMeasurement` etc.?

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -476,39 +546,45 @@ struct GsfActor {
...
       const auto firstCmpProxy =
           tmpStates.traj.getTrackState(tmpStates.tips.front());
-      const auto isMeasurement = firstCmpProxy.typeFlags().isMeasurement();
 
-      const auto mask =
-          isMeasurement
-              ? TrackStatePropMask::Calibrated | TrackStatePropMask::Predicted |
-                    TrackStatePropMask::Filtered | TrackStatePropMask::Smoothed
-              : TrackStatePropMask::Calibrated | TrackStatePropMask::Predicted;
+      auto mask = TrackStatePropMask::Predicted | TrackStatePropMask::Smoothed;
```
benjaminhuth: hmm shouldn't we allocate `smoothed` only for measurements?

Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -467,18 +467,80 @@ class KalmanFitter {
...
+            TrackStatePropMask::Jacobian | TrackStatePropMask::Calibrated;
+        typename traj_t::TrackStateProxy trackStateProxy =
+            result.fittedStates->makeTrackState(mask, result.lastTrackIndex);
+
+        typename traj_t::ConstTrackStateProxy trackStateProxyConst{
+            trackStateProxy};
+
+        // Set the trackStateProxy components with the state from the ongoing
+        // propagation
+        {
```
paulgessinger: Any reason for the scope here?

Core/include/Acts/TrackFitting/KalmanFitter.hpp
```diff
@@ -517,21 +579,66 @@ class KalmanFitter {
...
-        const auto trackStateProxyRes = detail::kalmanHandleNoMeasurement(
-            state.stepping, stepper, *surface, *result.fittedStates,
-            result.lastTrackIndex, true, logger(), precedingMeasurementExists,
-            freeToBoundCorrection);
 
-        if (!trackStateProxyRes.ok()) {
-          return trackStateProxyRes.error();
+        // Add a <mask> TrackState entry multi trajectory. This allocates
+        // storage for
+        // all components, which we will set later.
```
paulgessinger: I only allocated predicted and jacobian, maybe change the comment accordingly?

PR #5020, opened 2026-01-28
fix: Increment particle number in EDM4hep conv gen particles

paulgessinger:
This pull request makes a minor change to the particle creation logic in the `EDM4hepSimInputConverter::convert` method. The update clarifies the construction of the `particleId` by separating its creation from the construction of the `SimParticle` object, which improves readability and maintainability.

- Refactored the particle creation loop in `EDM4hepSimInputConverter.cpp` to first create the `particleId` before constructing the `SimParticle`, making the code clearer and easier to follow.

### DIFF
--- a/Plugins/Gnn/include/ActsPlugins/Gnn/ModuleMapCuda.hpp
+++ b/Plugins/Gnn/include/ActsPlugins/Gnn/ModuleMapCuda.hpp
@@ -14,51 +14,7 @@
 #include <memory>
 #include <string>
 
-/// @cond
-template <typename T>
-class CUDA_module_map_doublet;
-
-struct CUstream_st;
-using cudaStream_t = CUstream_st *;
-
-template <typename T>
-class CUDA_module_map_triplet;
-/// @endcond
-
 namespace ActsPlugins {
-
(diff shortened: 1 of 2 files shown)


Examples/Io/EDM4hep/src/EDM4hepSimInputConverter.cpp
```diff
@@ -377,11 +377,12 @@ ProcessCode EDM4hepSimInputConverter::convert(const AlgorithmContext& ctx,
                                            generatorStableParticles.size());
 
       for (const auto& genParticle : generatorStableParticles) {
-        SimParticle particle =
-            EDM4hepUtil::readParticle(genParticle)
-                .withParticleId(SimBarcode()
+        nParticles += 1;
```
AJPfleger, line 380: Why do we increment?

paulgessinger: The particle id has a running particle number per primary vertex. So far for the generator particles, it set that number to 0 unconditionally, which is wrong. I incrementing is what we want.

PR #4765, opened 2025-11-03
feat: Definitely not ATLAS toroidal field

davidedicroce:
Core Implementation:
- Add ToroidalField.hpp: toroid magnetic field using Biot-Savart law
- Add ToroidalFieldBindings.cpp: Python bindings with factory functions
- Support both default configuration and custom parameters
- Header-only implementation with dual constructors

Build System Integration:
- Update Examples/Detectors/MagneticField/CMakeLists.txt for conditional compilation
- Update Examples/Python/CMakeLists.txt for Python bindings

Tests and Examples:
- Add toroidal_field_basic.py: Simple usage example
- Add toroidal_field_simulation.py: Complete GeoModel simulation

Features:
- Configurable inner/outer radius, current, and number of coils
- Default configuration (R_in=4.9m, R_out=10m, I=20.5kA, 8 coils)
- Toroidal field symmetry with phi dependence
- Integration with ACTS Examples framework and Geant4 simulation"

References
- https://cds.cern.ch/record/1071117/files/cer-002726678.pdf 
toroidField_y0p1
toroidField_z0

PR #5025, opened 2026-01-29
refactor: Cleanup toroid field, relocate to core

paulgessinger:
Follow up to https://github.com/acts-project/acts/pull/4765


paulgessinger:
Maybe the class should be called `ToroidField` similar to `SolenoidField`.

asalzburger:
This revealed quite a lot of new SonarCloud issues - most of them easy to fix, can we address them ?

Core/src/MagneticField/ToroidField.cpp
```diff
@@ -312,13 +314,14 @@ void ToroidalField::buildGeometry() {
...
+  std::vector<std::array<float, 2>> d_rzE;
+  std::vector<std::array<float, 2>> m_rzE;
   buildSegsMidsRZ(rz_ect, d_rzE, m_rzE);
 
   // ---- Angles ----
   const int nC = m_cfg.layout.nCoils;
-  const float th0 = static_cast<float>(m_cfg.layout.theta0);
-  const float dth = static_cast<float>(m_cfg.layout.thetaStep);
+  const double th0 = static_cast<float>(m_cfg.layout.theta0);
+  const double dth = static_cast<float>(m_cfg.layout.thetaStep);
```
andiwand, line 312: there are a couple of warnings related to this from sonarcloud

paulgessinger: Right, I'll do another round.

paulgessinger: moved everything to double let's see what sonar says.

PR #5028, opened 2026-01-29
build: Add flags, improve VDT discovery

paulgessinger:
--- END COMMIT MESSAGE ---


paulgessinger:
So it was a clang thing. Perfect!

PR #5032, opened 2026-01-30
refactor: Finalize unROOTing Examples

benjaminhuth:
* Restructure `Examples/Framework` to allow complilation without ROOT (`ScalingCalibrator`, ROOT thread safety in `Sequencer`)
* Extract ROOT-based visualization function from the muon digitization and hough seeder.
* Move python bindings of ROOT-based examples functionality to the ROOT plugin binding.
* Enable import of `simulation.py`/`reconstruction.py` without ROOT.
* Some CMake cleanup


Examples/Framework/CMakeLists.txt
```diff
@@ -41,13 +42,14 @@ target_link_libraries(
...
-        ROOT::Core
-        ROOT::Hist
+        $<$<BOOL:${ACTS_BUILD_EXAMPLES_ROOT}>:ROOT::Core>
+        $<$<BOOL:${ACTS_BUILD_EXAMPLES_ROOT}>:ROOT::Hist>
         HepMC3::HepMC3
 )
 target_compile_definitions(
     ActsExamplesFramework
     PRIVATE BOOST_FILESYSTEM_NO_DEPRECATED
+    PUBLIC $<$<BOOL:${ACTS_BUILD_EXAMPLES_ROOT}>:ACTS_BUILD_EXAMPLES_ROOT>
```
paulgessinger: Can you make this private, and possibly only on the `Sequencer.cpp` where this is needed like

```cmake
set_source_files_properties(
  src/special.cpp
  PROPERTIES
    COMPILE_DEFINITIONS MY_SPECIAL_DEFINE=1
)
```

Examples/Framework/CMakeLists.txt
```diff
@@ -8,7 +8,8 @@ acts_add_library(
     src/EventData/MuonSpacePointCalibrator.cpp
     src/EventData/Measurement.cpp
     src/EventData/MeasurementCalibration.cpp
-    src/EventData/ScalingCalibrator.cpp
+    $<$<BOOL:${ACTS_BUILD_EXAMPLES_ROOT}>:src/Root/ScalingCalibrator.cpp>
+    $<$<BOOL:${ACTS_BUILD_EXAMPLES_ROOT}>:src/Root/MuonVisualization.cpp>
```
paulgessinger: This is fancy, but I'm wondering if you could collect this and the other changes to this file into a single

```cmake
if(ACTS_BUILD_EXAMPLES_ROOT)
```

which then adds the source files and libraries? I think that will be simpler to follow.

Python/Examples/src/plugins/Root.cpp
```diff
@@ -274,4 +278,28 @@ PYBIND11_MODULE(ActsExamplesPythonBindingsRoot, root) {
...
+  {
+    py::class_<MeasurementCalibrator, std::shared_ptr<MeasurementCalibrator>>(
+        root, "MeasurementCalibrator");
+
+    root.def(
+        "makeScalingCalibrator",
+        [](const char* path) -> std::shared_ptr<MeasurementCalibrator> {
+          return std::make_shared<ScalingCalibrator>(path);
+        },
+        py::arg("path"));
```
paulgessinger, line 305: Why not just the type `ScalingCalibrator` with the inheritance relationship declared? I think pybind11 should then be able to figure this out.

Python/Examples/src/plugins/Root.cpp
```diff
@@ -274,4 +278,28 @@ PYBIND11_MODULE(ActsExamplesPythonBindingsRoot, root) {
...
+  }
+
+  // Muon visualization
+  {
+    root.def("makeMuonVisualizationFunction", []() {
```
paulgessinger, line 310: Did you try if simply binding the visualizer free function to python and then passing that into the config doesn't work? If that works, I think that would be simpler.

benjaminhuth: Hmm yeah, the reason why I didn't do this is because I have no idea how to run muon digitization, and how to test this... That way I was relatively certain that it should work.

Python/Examples/src/plugins/Root.cpp
benjaminhuth: Hmm this is just copied from the previous binding code... I agree that should be possible, but this is a bit out-of scope of this PR to change the pattern for all calibrators

Python/Examples/src/plugins/Root.cpp
benjaminhuth: If there is an easy recipy I could try to run this, but I would also be happy to leave it like this :D

Examples/Framework/CMakeLists.txt
```diff
@@ -29,6 +24,11 @@ acts_add_library(
     src/Validation/TrackClassification.cpp
     src/Validation/TrackQualityPlotTool.cpp
     src/Validation/TrackSummaryPlotTool.cpp
+)
+
+acts_add_library(
+    ExamplesFramework
+    ${ActsExamplesFramework_SOURCES}
```
paulgessinger: I'd probably just inline this here at this point.

Python/Examples/python/reconstruction.py
```diff
@@ -6,6 +6,22 @@
...
+        RootTrackFinderNTupleWriter,
+        RootTrackFinderPerformanceWriter,
+        RootTrackFitterPerformanceWriter,
+        RootTrackParameterWriter,
+        RootTrackStatesWriter,
+        RootTrackSummaryWriter,
+        RootVertexNTupleWriter,
+    )
+
+    ACTS_ROOT_AVAILABLE = True
```
paulgessinger: nitpick `ACTS_EXAMPLES_ROOT_AVAILABLE`?

CMakeLists.txt
```diff
@@ -597,7 +579,23 @@ if(ACTS_BUILD_EXAMPLES)
...
+        )
+    else()
+        find_package(
+            HepMC3
+            ${_acts_hepmc3_version}
+            REQUIRED
+            CONFIG
+            COMPONENTS search
+        )
+    endif()
```
paulgessinger, line 585: I this case I'd probably store the components in a variable above this and modify it conditionally, so that there's a single call to `find_package`.

PR #5035, opened 2026-02-02
feat: Add jet kinematics into TrackSummaryWriter

delitez:
In this PR, an option to write jet kinematics to TrackSummaryWriter has been added, as has the track-to-jet association based on the deltaR distance.

### DIFF
--- a/CI/clang_tidy/requirements.txt
+++ b/CI/clang_tidy/requirements.txt
@@ -28,7 +28,7 @@ python-slugify==6.1.2
     # via codereport
 pyyaml==6.0.3
     # via -r CI/clang_tidy/requirements.in
-rich==14.3.2
+rich==14.3.3
     # via -r CI/clang_tidy/requirements.in
 setuptools==82.0.0
     # via fs
--- a/CI/fpe_masks/requirements.txt
+++ b/CI/fpe_masks/requirements.txt
@@ -42,13 +42,13 @@ pygments==2.19.2
     # via rich
 pyjwt==2.11.0
     # via gidgethub
-rich==14.3.2
+rich==14.3.3
     # via
     #   -r CI/fpe_masks/requirements.in
     #   typer
 shellingham==1.5.4
     # via typer
-typer==0.24.0
+typer==0.24.1
     # via -r CI/fpe_masks/requirements.in
 uritemplate==4.2.0
     # via gidgethub
--- a/Examples/Scripts/requirements.txt
+++ b/Examples/Scripts/requirements.txt
@@ -85,7 +85,7 @@ python-dateutil==2.9.0.post0
     #   pandas
 pyyaml==6.0.3
     # via -r Examples/Scripts/requirements.in
-rich==14.3.2
+rich==14.3.3
     # via
     #   -r Examples/Scripts/requirements.in
     #   typer
@@ -95,7 +95,7 @@ shellingham==1.5.4
     # via typer
 six==1.17.0
     # via python-dateutil
-typer==0.24.0
+typer==0.24.1
     # via -r Examples/Scripts/requirements.in
 typing-extensions==4.15.0
     # via
--- a/docs/old/requirements.txt
+++ b/docs/old/requirements.txt
@@ -94,7 +94,7 @@ pyyaml==6.0.3
     # via myst-parser
 requests==2.32.5
     # via sphinx
-rich==14.3.2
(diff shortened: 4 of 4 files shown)


benjaminhuth:
Just as a thought: Would a `RootJetWriter` be useful maybe?

Examples/Io/Root/include/ActsExamples/Io/Root/RootTrackSummaryWriter.hpp
```diff
@@ -16,6 +16,7 @@
 #include "ActsExamples/Framework/DataHandle.hpp"
 #include "ActsExamples/Framework/ProcessCode.hpp"
 #include "ActsExamples/Framework/WriterT.hpp"
+#include "ActsExamples/Jets/TruthJetAlgorithm.hpp"
```
andiwand: can we `#include` `TruthJetContainer` without `TruthJetAlgorithm`?

Examples/Io/Root/include/ActsExamples/Io/Root/RootTrackSummaryWriter.hpp
```diff
@@ -281,6 +287,14 @@ class RootTrackSummaryWriter final : public WriterT<ConstTrackContainer> {
...
   /// The number of updates (gx2f)
   std::vector<int> m_nUpdatesGx2f;
+
+  /// The jet information
+  std::vector<int> m_nJets;
+  std::vector<float> m_jet_pt;
+  std::vector<float> m_jet_eta;
+  std::vector<float> m_jet_phi;
+  std::vector<int> m_jet_label;
+  std::vector<std::size_t> m_ntracks_per_jets;
```
andiwand, line 299: these are per jet properties right?

does it make sense to mix them with tracks?

delitez:
> Just as a thought: Would a `RootJetWriter` be useful maybe?

The original idea was to have a separate writer for jets, but for the track-to-jet association, tracks had to be included in that writer as well. Since there are only a few kinematic variables for jets to be added, @paulgessinger and I thought they could be included in the TrackSummaryWriter as optional.

benjaminhuth:
Hmm I see, I'm not feeling strongly about this, but it seems that the variables written are per-jet as @andiwand pointed out, and not per track, so there is no real association happening here, right?

Examples/Io/Root/include/ActsExamples/Io/Root/RootTrackSummaryWriter.hpp
delitez: I would still need the `ActsPlugins/FastJet/Jet.hpp`, or is there any other way?

Examples/Io/Root/include/ActsExamples/Io/Root/RootTrackSummaryWriter.hpp
delitez: discussed offline, resolving the comment

Examples/Io/Root/include/ActsExamples/Io/Root/RootTrackSummaryWriter.hpp
andiwand: ah I only later saw that `TruthJetContainer` lives inside `TruthJetAlgorithm.hpp` - lets keep it as it is

Examples/Io/Root/CMakeLists.txt
```diff
@@ -38,6 +38,8 @@ target_link_libraries(
     ActsExamplesIoRoot
     PUBLIC
         Acts::PluginRoot
+        Acts::PluginFastJet
+        Acts::ExamplesJets
```
andiwand: hm that might be one more reason to split this into a different writer @paulgessinger @murnanedaniel @benjaminhuth

delitez: FastJet plugin is enabled by default for common-ci presets, but not for gnn-ci preset, for which the ci-job fails. I just pushed a commit enabling it, @benjaminhuth would it break anything?

veprbl: This is no just ci issue. Right now you are forcing users of root IO to also enable `ACTS_BUILD_PLUGIN_FASTJET`.

delitez: Drafting this PR for now to investigate another solution

PR #5036, opened 2026-02-02
feat: Add k0 and lambda0 to the list of pdg ids 

pbutti:
Very small PR to add possibility to generate K0s and Lambda0s similarly to the other particles with a particle gun for simple vertexing studies.


Core/include/Acts/Definitions/PdgParticle.hpp
```diff
@@ -38,6 +38,8 @@ enum PdgParticle : std::int32_t {
   eProton = 2212,
   eAntiProton = -eProton,
   eLead = 1000822080,
+  eK0S = 310,
```
andiwand:
```suggestion
  eKaon0Short = 310,
```
nitpick: we already have `eKaonPlus` and `eKaonMinus` so we could stick to this pattern

PR #5038, opened 2026-02-02
ci: Update coverage scripts

paulgessinger:
--- END COMMIT MESSAGE ---


CI/test_coverage.py
```diff
@@ -1,88 +1,282 @@
...
+    gcov_object_directory: str,
+) -> list[str]:
+    script_dir = Path(__file__).resolve().parent
+    source_dir = script_dir.parent.resolve()
+
+    version = gcovr_version(gcovr_exe)
+    extra_flags: list[str] = []
+    if version is not None:
+        if version >= (6, 0):
+            extra_flags.append("--exclude-noncode-lines")
```
AJPfleger:
```suggestion
    if version is not None and version >= (6, 0):
        extra_flags.append("--exclude-noncode-lines")
```

CI/test_coverage.py
```diff
@@ -1,88 +1,282 @@
...
+    "dependencies/",
+    "spack/",
+    "thirdparty/",
+]
+
+
+def _resolve_excludes(source_dir: Path) -> list[str]:
+    """Return exclude patterns: EXCLUDE_PATTERNS as-is plus EXCLUDE_PATHS prefixed with source_dir."""
+    source_prefix = re.escape(source_dir.as_posix()) + r"/"
+    return list(EXCLUDE_PATTERNS) + [source_prefix + p for p in EXCLUDE_PATHS]
```
AJPfleger:
```suggestion
    return EXCLUDE_PATTERNS + [source_prefix + p for p in EXCLUDE_PATHS]
```
`EXCLUDE_PATTERNS` is already a list

.github/workflows/analysis.yml
```diff
@@ -87,15 +87,30 @@ jobs:
...
+        with:
+          name: coverage-html
+          path: build/coverage/html
+
+      - name: Coverage display
+        run: |
+          base_link='https://acts-herald.app.cern.ch/view/${{ github.repository }}/${{ steps.artifact-upload-step.outputs.artifact-id }}'
+          link="$base_link/index.html"
+          echo "🛡️ Code coverage available at: $link"
+          echo "**🛡️ Code coverage available [here]($link)**" >> $GITHUB_STEP_SUMMARY
```
AJPfleger, line 113: Why the `🛡️`? (｡•́‿•̀｡)

CI/test_coverage.py
```diff
@@ -1,88 +1,282 @@
...
+    if not (build_dir / "CMakeCache.txt").exists():
+        console.print(
+            f"Build directory missing CMakeCache.txt: {build_dir}",
+            style="red",
+        )
+        raise typer.Exit(1)
+
+    gcov_exe = gcov or locate_executable(
+        "gcov",
+        "gcov not installed. Install GCC coverage tooling or pass --gcov.",
```
AJPfleger:
```suggestion
        "gcov not installed. Install GCC coverage tooling or pass the path to an executable with --gcov.",
```

CI/test_coverage.py
```diff
@@ -1,88 +1,282 @@
...
+        ),
+    ] = False,
+    html: Annotated[
+        bool,
+        typer.Option("--html/--no-html", help="Also generate HTML coverage report"),
+    ] = True,
+    html_theme: Annotated[
+        str,
+        typer.Option("--html-theme", "-t", help="HTML report theme (e.g. github.blue)"),
+    ] = "github.blue",
```
AJPfleger: What is the long-term goal of the flags? So far, we only pass the path to build and use `--filter`, I think.

CI/test_coverage.py
```diff
@@ -1,88 +1,282 @@
...
+        console.print(f"Input file not found: {input}", style="red")
+        raise typer.Exit(1)
+
+    source_dir = Path(__file__).resolve().parent.parent
+    excludes = list(exclude) if exclude is not None else _resolve_excludes(source_dir)
+    if build_dir_name is not None:
+        excludes.append("^" + re.escape(build_dir_name))
+
+    filter_coverage_xml(input, output, excludes)
+
```
AJPfleger: I think this function is unused.

CI/test_coverage.py
```diff
@@ -1,88 +1,282 @@
...
+
+def filter_coverage_xml(
+    input_path: Path, output_path: Path, excludes: list[str]
+) -> None:
+    from lxml import etree
+
+    patterns = [re.compile(p) for p in excludes]
+
+    tree = etree.parse(input_path)  # noqa: S320
```
AJPfleger: Is the linter-ignore relevant for sonar?

.github/workflows/analysis.yml
paulgessinger: Because I think it looks cool :D

CI/test_coverage.py
paulgessinger: I'll remove this.

CI/test_coverage.py
paulgessinger: I'll drop the html ones. `--jobs`, `--verbose` and `--gcov` I used for local testing and would be helpful to have.

paulgessinger:
@AJPfleger updated now, can you have another look?

.github/workflows/analysis.yml
AJPfleger: We’ll see how history judges it ( •̀ᴗ•́ )و ̑̑

AJPfleger (approved):

PR #5040, opened 2026-02-02
chore: Another round of sonar fixes

paulgessinger:
--- END COMMIT MESSAGE ---


.github/workflows/builds.yml
```diff
@@ -589,7 +589,7 @@ jobs:
         run: ccache -s
 
       - name: Install wheel
-        run: pip install $(ls *.whl)
+        run: pip install $(./ls *.whl)
```
andiwand: intentional change?

paulgessinger: Yes, but incorrect. I'll fix it.

Issue #5033, opened 2026-01-30
Local python installation is not importable after `ACTS_BUILD_PYTHON_WHEEL` changes


asalzburger:
The change in CMake to to this:

```
# Set flags dependent whether we build ACTS in the default configuration or for deployment in pip with scikit-build-core
if(ACTS_BUILD_PYTHON_WHEEL)
    # Check if skbuild is running
    if("${SKBUILD}" STREQUAL "2")
        message(
            STATUS
            "Build with skbuild as ${SKBUILD_PROJECT_NAME},${SKBUILD_PROJECT_VERSION_FULL} (${SKBUILD_STATE})"
        )
    else()
        message(WARNING "scikit-build-core not detected for building wheel")
    endif()

    # For deployment with pip, we build static libs
    set(BUILD_SHARED_LIBS FALSE)
    set(CMAKE_POSITION_INDEPENDENT_CODE ON)

    # Set installation paths for proper python package creation
    # CMAKE_PREFIX_PATH should be site-packages (steered by scikit-build-core)
    set(CMAKE_INSTALL_BINDIR ${SKBUILD_NULL_DIR}) # Don't install binaries/scripts
    set(CMAKE_INSTALL_LIBDIR "acts")
    set(CMAKE_INSTALL_INCLUDEDIR ${SKBUILD_NULL_DIR}}) # Don't install headers
    set(CMAKE_INSTALL_DATAROOTDIR "acts/share")
    set(CMAKE_INSTALL_DATADIR "acts/share")
    set(ACTS_PYTHON_INSTALL_DIR "acts")
else()
    # GNU-like installation paths, e.g. lib/, include/, ...
    include(GNUInstallDirs)
    set(ACTS_PYTHON_INSTALL_DIR "python")
endif()
```

creates now, if you don't want to create a wheel a python installation under `python` and the acts python bindings such as the the `__init__.py` are now in an install folder to be found as `python/__init__.py`.

This renders:
```python
import acts
```

Impossible, as the module acts does not exist anymore.

paulgessinger:
I think it's been working for me still, strange. Ah but you always install from the build directory, I only use the build directory.

asalzburger:
Exactly, the problem is indeed from the install directory - it should be installed into "python/acts" - I will make a fix.

benjaminhuth:
I have a fix prepared for this: https://github.com/acts-project/acts/pull/5041

PR #4948, opened 2026-01-09
feat: pip install acts, first steps

benjaminhuth:
Enable `pip install .` inside the source tree and builds it in the CI with a dedicated cmake prefix. First step to deploy acts on PyPI.

--- END COMMIT MESSAGE ---

Any further description goes here, @-mentions are ok here!

- Use a *conventional commits* prefix: [quick summary](https://www.conventionalcommits.org/en/v1.0.0/#summary)
  - We mostly use `feat`, `fix`, `refactor`, `docs`, `chore` and `build` types.
- A milestone will be assigned by one of the maintainers

PR #5041, opened 2026-02-02
fix: Python setup.sh from installation directory

benjaminhuth:
Recent changes in #4948 broke the python setup from an installation directory as was pointed out in #5033. This fixes this, and also adds a sanity check to the CI.


.github/workflows/builds.yml
```diff
@@ -83,6 +83,11 @@ jobs:
       - name: Install
         run: cmake --build build --target install
 
+      - name: Python sanity check after install
+        run: >
```
asalzburger, line 87:
```suggestion
        shell: bash
        run: >
```

PR #5042, opened 2026-02-02
fix: local python installation

asalzburger:
This PR addresses #5033 for me, by re-introducing the `acts` subdirectory in the installation directory, this allows again to do 

```python 
import acts
```

As the `__init__,py` file now is back at the `acts/__init__,py` location and (as prior) in with `python/__init__py`.

Indeed, we can discuss if the libraries should be in `python`, but I don't really see someone actually using the bare python binding libraries.

The 'setup.sh` file location is will still be within `installed/python/setup.sh`, i.e. parallel to the `acts` module folder where it used to be in order not to break backwards compatibility.

--- END COMMIT MESSAGE ---


asalzburger:
@paulgessinger 👀

asalzburger:
That of course breaks the wheel build ... fixing now.

benjaminhuth:
Hi, just FYI, I opened #5041 yesterday to fix this as well, but this also seem to have still issues

asalzburger:
Ah, we arrived at the same fix.

asalzburger:
Closing this as it is a duplicate of #5041 - let's make the CI of first one work!

asalzburger:
Or let's just try this one with CI.

benjaminhuth:
Hmm this can be closed, right? @asalzburger

PR #4931, opened 2025-12-28
fix: adapt Mat_map_surface_plot.C to read sur_type from geometry-map.json

veprbl:
The sur_type and sur_range_min/max were removed in https://github.com/acts-project/acts/pull/4353, so these macros don't work anymore. This adapts the macros to access the same information from the JSON files.

--- END COMMIT MESSAGE ---

I think, at ePIC we are still interested in surface material maps on up to date Acts.
This is alternative to #4929.
I'm not yet sure about type==4 case yet, it's defunct now, and I don't have a handy test case.

PR #5046, opened 2026-02-03
fix: speedup surface lookup Mat_map_surface_plot.C using precomputed map

veprbl:
This is a small follow up #4931. It was discovered that geant track files will contain lots of invalid surfaces id=0, that lead to negative lookups being redone on each event. That uncovered a problem with performance of the prior way to lookup that iterates over json. This implements a helper structure that allows to speed up the lookup.


veprbl:
This is now deployed at ePIC.

PR #5047, opened 2026-02-03
refactor: Rework GSF outlier handling

andiwand:
Currently, individual outlier states were treated like filtered states if there was at least one measurement state. This logic was reworked. With this change, outlier states are removed from the pool. If all states are outliers the update is performed as if it was a hole while still attaching the measurement to the track state.

--- END COMMIT MESSAGE ---

blocked by
- https://github.com/acts-project/acts/pull/5015


benjaminhuth (approved):
Looks good!

PR #5050, opened 2026-02-04
feat: Track to jet matching

delitez:
In this PR track to jet association is introduced as an optional method within TruthJetAlgorithm. One can either have only clustered jets or if enabled, jets with their associated reconstructed tracks.


Examples/Algorithms/Jets/src/TruthJetAlgorithm.cpp
```diff
@@ -306,4 +315,54 @@ ProcessCode ActsExamples::TruthJetAlgorithm::finalize() {
   return ProcessCode::SUCCESS;
 }
 
+TruthJetContainer TruthJetAlgorithm::trackJetMatching(
```
paulgessinger: This modified the `jets` argument in place, but then also returns a copy. I think you can skip the return and just have the method act on the mutable reference argument.

Examples/Algorithms/Jets/src/TruthJetAlgorithm.cpp
```diff
@@ -296,7 +300,12 @@ ProcessCode ActsExamples::TruthJetAlgorithm::execute(
     ACTS_DEBUG("  - " << label << ": " << count);
   }
 
-  m_outputJets(ctx, std::move(outputJetContainer));
+  if (m_cfg.doTrackJetMatching) {
+    auto trackMatchedJets = trackJetMatching(tracks, outputJetContainer);
+    m_outputJets(ctx, std::move(trackMatchedJets));
+  } else {
+    m_outputJets(ctx, std::move(outputJetContainer));
+  }
```
paulgessinger, line 305: In combination with my other comment (removing the copy return) this would become:


```suggestion
  if (m_cfg.doTrackJetMatching) {
    trackJetMatching(tracks, outputJetContainer);
  }
  m_outputJets(ctx, std::move(outputJetContainer));
```

delitez: Yes, I see, thanks!

Examples/Algorithms/Jets/src/TruthJetAlgorithm.cpp
```diff
@@ -34,6 +34,7 @@ TruthJetAlgorithm::TruthJetAlgorithm(const Config& cfg,
     throw std::invalid_argument("Input particles collection is not configured");
   }
   m_inputTruthParticles.initialize(m_cfg.inputTruthParticles);
+  m_inputTracks.initialize(m_cfg.inputTracks);
```
andiwand: the tracks are only needed if `doTrackJetMatching` is `true` right?

in this case I would either always perform the matching and remove the flag or only depend on the track input if `doTrackJetMatching` is set

Examples/Algorithms/Jets/src/TruthJetAlgorithm.cpp
```diff
@@ -296,7 +299,12 @@ ProcessCode ActsExamples::TruthJetAlgorithm::execute(
     ACTS_DEBUG("  - " << label << ": " << count);
   }
 
-  m_outputJets(ctx, std::move(outputJetContainer));
+  if (m_cfg.doTrackJetMatching) {
+    const ConstTrackContainer& tracks = m_inputTracks(ctx);
+    m_outputJets(ctx, trackJetMatching(tracks, outputJetContainer));
+  } else {
+    m_outputJets(ctx, std::move(outputJetContainer));
+  }
```
andiwand, line 305:
```suggestion
  if (m_cfg.doTrackJetMatching) {
    const ConstTrackContainer& tracks = m_inputTracks(ctx);
    trackJetMatching(tracks, outputJetContainer);
  }

  m_outputJets(ctx, std::move(outputJetContainer));
```
this might be a bit easier to understand

Examples/Algorithms/Jets/src/TruthJetAlgorithm.cpp
```diff
@@ -306,4 +314,54 @@ ProcessCode ActsExamples::TruthJetAlgorithm::finalize() {
   return ProcessCode::SUCCESS;
 }
 
+TruthJetContainer TruthJetAlgorithm::trackJetMatching(
```
andiwand:
```suggestion
void TruthJetAlgorithm::trackJetMatching(
```
in case you go with my suggestion above the return type can be dropped

delitez: yes, looks good to me, thanks!

andiwand (approved):

PR #5051, opened 2026-02-04
chore: Remove old seeders from Examples

andiwand:
Removes `Default` and `Orthogonal` seeders from Examples. They have been replaced by `GridTriplet` and `OrthogonalTriplet` a while ago.

--- END COMMIT MESSAGE ---


paulgessinger (commented):
Should we make `GridTriplet` the default then?

andiwand:
> Should we make `GridTriplet` the default then?

I believe it is already replaced in all our internal cases

Python/Examples/python/reconstruction.py
```diff
@@ -10,7 +10,7 @@
 
 SeedingAlgorithm = Enum(
     "SeedingAlgorithm",
-    "Default TruthSmeared TruthEstimated Orthogonal HoughTransform AdaptiveHoughTransform Gbts Hashing GridTriplet OrthogonalTriplet",
```
paulgessinger, line 13: @andiwand right, but here `Default` goes away. I think having `Default` here is a good idea, and should either lead to `GridTriplet` or replace that enum case. What do you think?

andiwand: I don't think `Default` adds anything other than confusion. Right now it is a distinct enumeration entry. It should rather be an alias that points to `GridTriplet`, but then why not just use `GridTriplet` explicitly on the user site?

paulgessinger: ok

PR #5054, opened 2026-02-05
refactor: `VertexTruthMatcher` Examples algorithm

andiwand:
Extracts the vertex truth matching from `RootVertexNTupleWriter` to its own algorithm `VertexTruthMatcher` - similar to what was done with the `TrackTruthMatcher`.

I created a new Examples utility `VertexTruthUtility` to capture functions that were hidden in the `RootVertexNTupleWriter` implementation.

There are a couple of non-functional other changes related to vertexing in the Examples.

--- END COMMIT MESSAGE ---


Examples/Io/Root/include/ActsExamples/Io/Root/RootVertexNTupleWriter.hpp
```diff
@@ -67,12 +62,6 @@ class RootVertexNTupleWriter final : public WriterT<std::vector<Acts::Vertex>> {
     /// File access mode.
     std::string fileMode = "RECREATE";
 
-    /// Minimum fraction of track weight matched between truth
-    /// and reco vertices to consider as truth matched.
-    double vertexMatchThreshold = 0.7;
-    /// Minimum fraction of hits associated to particle to consider track
-    /// as truth matched.
-    double trackMatchThreshold = 0.5;
```
AJPfleger, line 75: This one is dropped completely?

AJPfleger (approved):

Examples/Io/Root/include/ActsExamples/Io/Root/RootVertexNTupleWriter.hpp
AJPfleger: I mean just the `trackMatchThreshold`

andiwand: right this is was not used at all. I think this comes from back in the days where this performance writer used to do its own truth matching

PR #5067, opened 2026-02-06
fix: guard debug print in CudaTrackBuilding against small bounds vectors

murnanedaniel:
When numberLabels < 3, bounds.size() < 4, so accessing bounds.at(2) and bounds.at(numberLabels - 2) in the ACTS_DEBUG print caused an std::out_of_range crash. 

Guard the detailed print behind a size check and fall back to printing all elements for small vectors.

PLEASE DESCRIBE YOUR CHANGES.
DO NOT USE @-MENTIONS HERE!

--- END COMMIT MESSAGE ---


murnanedaniel:
A little extra context: This is killing GNN testing when running on single particles, since a correctly reconstructed event will give just one label, which crashes this debug line.

Plugins/Gnn/src/CudaTrackBuilding.cu
```diff
@@ -124,11 +126,22 @@ std::vector<std::vector<int>> CudaTrackBuilding::operator()(
...
+                          << bounds.at(numberLabels));
+  } else {
+    std::ostringstream oss;
+    for (std::size_t i = 0; i < bounds.size(); ++i) {
+      if (i > 0) {
+        oss << ", ";
+      }
+      oss << bounds[i];
+    }
+    ACTS_DEBUG("Bounds: " << oss.str());
```
paulgessinger: Why not just do this in all cases? You could use a range based for loop and an immediately executing lambda to put this inside the log macro even.

murnanedaniel: Good point. Will make the fix

Plugins/Gnn/src/CudaTrackBuilding.cu
```diff
@@ -124,11 +126,22 @@ std::vector<std::vector<int>> CudaTrackBuilding::operator()(
   ACTS_CUDA_CHECK(cudaGetLastError());
 
   ACTS_DEBUG("Bounds size: " << bounds.size());
-  ACTS_DEBUG("Bounds: " << bounds.at(0) << ", " << bounds.at(1) << ", "
-                        << bounds.at(2) << ", ..., "
-                        << bounds.at(numberLabels - 2) << ", "
-                        << bounds.at(numberLabels - 1) << ", "
-                        << bounds.at(numberLabels));
+  if (numberLabels >= 3) {
```
benjaminhuth: Hmm this should be more like 7, right?

Plugins/Gnn/src/CudaTrackBuilding.cu
benjaminhuth: Hmm not sure what you mean with all cases, but we should not print out 150k-300k labels in DEBUG mode...

benjaminhuth (dismissed):
Thanks for catching this!

Plugins/Gnn/src/CudaTrackBuilding.cu
murnanedaniel: Take a look at the attempted fix:
- Iterates over the first min(3, bounds.size)
- Then iterates over the last min(3, bounds.size-3)
more or less

AJPfleger:
Hi @murnanedaniel @paulgessinger @benjaminhuth what's the status on this PR?

AJPfleger:
This looks fixed with https://github.com/acts-project/acts/pull/5340

PR #5161, opened 2026-02-24
feat: Detray python bindings proof-of-concept

paulgessinger:


cmake/ActsCreateSetup.cmake
```diff
@@ -5,6 +5,7 @@ configure_file(
     ${PROJECT_BINARY_DIR}/this_acts.sh
     @ONLY
 )
+message(STATUS "making setup script with dependencies: ${ROOT_DIR}")
```
paulgessinger, line 8:
```suggestion
```

PR #5168, opened 2026-02-25
build: export example targets

gondiaz:
Export targets for reading and preparation of simulation data for further processing. The exported targets (and dependencies) are:
- ExamplesFramework
- ExamplesIoPodio
- ExamplesIoEDM4hep (deps: ExamplesDetectorsCommon, ExamplesDetectorDD4hep, ExamplesDigitization)
- ExamplesIoJson (deps: ExamplesMaterialMapping, ExamplesTrackFinding, ExamplesMagneticField)

--- END COMMIT MESSAGE ---

Context: it is framed in a demonstrator project for a new processing framework called [Allen Core](https://gitlab.cern.ch/allen-core/allen-core) which is based on LHCb's online trigger software, [Allen](https://gitlab.cern.ch/lhcb/Allen/-/tree/master). The framework allows heterogeneous architectures, thus the plan is to use `traccc` reconstruction algorithms and Open Data Detector as a benchmark demonstrator.
This PR intends to export the tools we need from `ACTS` to prepare simulation data for further processing.


Examples/Io/Podio/include/ActsExamples/Io/Podio/PodioInputConverter.hpp
```diff
@@ -10,10 +10,9 @@
 
 #include "Acts/Utilities/Logger.hpp"
 #include "ActsExamples/Framework/IAlgorithm.hpp"
+#include "ActsExamples/Framework/DataHandle.hpp"
 
-namespace podio {
-class Frame;
-}
+#include <podio/Frame.h>
```
paulgessinger: Is the forward declaration not sufficient? The type isn't used in the header.

gondiaz: Right, the last commit is not needed, it is a remanent of some other developments. Dropped.

AJPfleger:
@paulgessinger please have a look as discussed in the dev meeting. thanks

paulgessinger:
I'm just noticing thie PR targets the release branch. This should probably target `main`, I imagine?

gondiaz:
> I'm just noticing thie PR targets the release branch. This should probably target `main`, I imagine?

I tryed to rebase into `main`, but found an issue with the `boost` version set for using `boost/hash2` [CMakeLists.txt#L56](https://github.com/acts-project/acts/blob/83b5bd2a95feefc0b5818bfe32df33a4974d2eda/Examples/Io/Root/CMakeLists.txt#L56). It is set to `1.86.0` but should be `1.88.0`.  Also using a lower version is problematic as the library is used anyway [RootFileHasher.cpp#L80](https://github.com/acts-project/acts/blob/83b5bd2a95feefc0b5818bfe32df33a4974d2eda/Examples/Io/Root/src/RootFileHasher.cpp#L80).

PR #5229, opened 2026-03-10
feat: muon buckets sector frame + bucket splitting

davidedicroce:
This MR introduces sector-based muon bucketization for the GeoModel mockup workflow and aligns the mockup geometry IDs so digitization can consistently resolve a common sector frame.

**1. MuonSpacePointDigitizer:**
  - Replaced chamber-frame helper with toSectorFrame() using a representative tracking volume.
  - Grouping changed from chamber-like IDs to sector IDs (volume-level key).
  - Added configurable bucket controls:
      - bucketMaxWindow (max longitudinal extent)
      - bucketNeighborWindow (max inter-point gap before split)
      - bucketOverlapWindow (tail overlap copied to next bucket)
      - minBucketSize (small buckets removed)
  - Implemented bucket split logic in sorted-z order per sector.
  - Added representative-volume resolution and caching per sector.

**2. GeoModelMuonMockupBuilder:**
  - Introduced sector parsing utilities from chamber names:
      - integer extraction by underscore tokenization
      - sector number extraction
      - eta-side sign extraction
  - Added a SectorKey = (station, sector, etaSide) grouping.
  - Assigned tracking volume IDs per sector key with explicit counter and capacity checks.
  - Assigned layer blocks per chamber/child volume to avoid collisions and preserve sector-consistent IDs.
  - Updated processStation() API to pass station index and shared nextVolumeId.

**3. Examples/Scripts/Python/geomodel_G4.py:**
  - Added useful CLI knobs: dumpGeoModelDb, nSectors, nEtaStations, disableEndcaps

**4. Examples/Detectors/MuonSpectrometerMockupDetector/src/GeoMuonMockupExperiment.cpp:**
  - Switched from std::any_cast to std::get and updated exception handling to std::bad_variant_access to match current key type usage.

**5. Python/Examples/src/Digitization.cpp:**
  - Extended Python digitization binding to pass new bucketization window parameters.

Screenshot 2026-03-10 at 17 06 26

@dimitra97 @junggjo9


Examples/Algorithms/Digitization/src/MuonSpacePointDigitizer.cpp
```diff
@@ -42,10 +50,78 @@ constexpr double quantize(const double x, const double pitch) {
...
+
+  std::vector<GeometryIdentifier> sortedIds = sectorLayerIds;
+  std::ranges::sort(sortedIds, [](const GeometryIdentifier& a,
+                                  const GeometryIdentifier& b) {
+    if (a.layer() != b.layer()) {
+      return a.layer() < b.layer();
+    }
+    return a.value() < b.value();
+  });
+  sortedIds.erase(std::unique(sortedIds.begin(), sortedIds.end()), sortedIds.end());
```
dimitra97: do we need to do this? In general the geoIds of the tracking geometry elements (volumes, surfaces etc) are supposed to be unique - otherwise the build of the geometry would fail

Examples/Detectors/GeoModelDetector/src/GeoModelMuonMockupBuilder.cpp
```diff
@@ -202,9 +212,27 @@ GeoModelMuonMockupBuilder::NodePtr_t GeoModelMuonMockupBuilder::processStation(
               ActsPlugins::GeoModel::volumePosInSpace(parent),
               parent->getLogVol()->getShape(), boundFactory);
 
+      const SectorKey key = sectorKey(stationIdx, box);
+      auto volIt = sectorVolumeIds.find(key);
+      if (volIt == sectorVolumeIds.end()) {
+        if (nextVolumeId > 255u) {
```
dimitra97: a bit hardcoded no? I think you can use something like this from here maybe
https://github.com/acts-project/acts/blob/main/Core/include/Acts/Geometry/GeometryIdentifier.hpp#L161
(in case the range also changes in the future)

Examples/Detectors/GeoModelDetector/src/GeoModelMuonMockupBuilder.cpp
```diff
@@ -236,23 +264,31 @@ GeoModelMuonMockupBuilder::NodePtr_t GeoModelMuonMockupBuilder::processStation(
...
     }
     auto trVol = buildChildChamber(gctx, box, boundFactory);
-    trVol->assignGeometryId(geoId.withVolume(it->second.second)
-                                .withExtra(chamberNode->children().size() + 1));
+    // Keep the chamber volume id and assign a unique layer per child volume
+    Acts::GeometryIdentifier chamberGeoId = std::get<2>(it->second);
+    std::size_t& childCounter = std::get<1>(it->second);
+    const auto baseLayer = chamberGeoId.layer();
+    const auto childLayer = baseLayer + (++childCounter);
+    if (childLayer > 0xFFFFu) {
```
dimitra97: same here :)

Examples/Detectors/GeoModelDetector/include/ActsExamples/GeoModelDetector/GeoModelMuonMockupBuilder.hpp
```diff
@@ -91,8 +93,9 @@ class GeoModelMuonMockupBuilder : public Acts::ITrackingGeometryBuilder {
   NodePtr_t processStation(const Acts::GeometryContext& gctx,
                            const std::span<Box_t> boundingBoxes,
                            const std::string& station, const bool isBarrel,
+                           const StationIdx stationIdx,
                            Acts::VolumeBoundFactory& boundFactory,
-                           const Acts::GeometryIdentifier& geoId) const;
+                           std::uint32_t& nextVolumeId) const;
```
dimitra97: What is the reason for this change? I am thinking if this actually limits us to only pass just a number (in this case representing volId). What if we want to have an identifier schema passed which would hold several fields. I dont know if this is a case..

Examples/Algorithms/Digitization/include/ActsExamples/Digitization/MuonSpacePointDigitizer.hpp
```diff
@@ -65,7 +66,17 @@ class MuonSpacePointDigitizer final : public IAlgorithm {
     double strawDeadTime{1. * Acts::UnitConstants::ms};
     /// @brief Applied dead time between two consecutive rpc hits
     double rpcDeadTime{50. * Acts::UnitConstants::ns};
+
+    /// @brief Maximum longitudinal extent of a bucket in the common sector frame
+    double bucketMaxWindow{800. * Acts::UnitConstants::mm};
```
junggjo9:
```suggestion
    double bucketMaxWindow{80. * Acts::UnitConstants::cm};
```

Examples/Algorithms/Digitization/include/ActsExamples/Digitization/MuonSpacePointDigitizer.hpp
```diff
@@ -65,7 +66,17 @@ class MuonSpacePointDigitizer final : public IAlgorithm {
...
+    /// @brief Maximum gap between neighboring space points before splitting
+    double bucketNeighborWindow{200. * Acts::UnitConstants::mm};
+    /// @brief Tail overlap copied from the previous bucket into the next one
+    double bucketOverlapWindow{100. * Acts::UnitConstants::mm};
```
junggjo9:
```suggestion
    double bucketNeighborWindow{20. * Acts::UnitConstants::cm};
    /// @brief Tail overlap copied from the previous bucket into the next one
    double bucketOverlapWindow{10. * Acts::UnitConstants::cm};
```

Examples/Detectors/GeoModelDetector/include/ActsExamples/GeoModelDetector/GeoModelMuonMockupBuilder.hpp
davidedicroce: The reason for switching from passing a GeometryIdentifier to passing nextVolumeId is that in this builder we were effectively using the incoming geoId only as a carrier for an incrementing volume counter, not as a real “schema” object.
With the new sector-based assignment, we need a simple monotonic allocator for unique volume ids that can be shared across stations/sectors, while the rest of the identifier fields (layer/extra) are determined locally.

Examples/Detectors/GeoModelDetector/src/GeoModelMuonMockupBuilder.cpp
davidedicroce: I’ll replace this hardcoded limit with the corresponding helper/constant from GeometryIdentifier so it automatically follows future schema changes.

Examples/Detectors/GeoModelDetector/src/GeoModelMuonMockupBuilder.cpp
davidedicroce: same indeed :)

Examples/Algorithms/Digitization/src/MuonSpacePointDigitizer.cpp
davidedicroce: This change only introduces grouping identifiers used internally by the digitizer logic to bucket space points (e.g. “sector id” = volume-only, and “volume+layer id” = coarse key).
The actual tracking geometry elements keep their full unique GeometryIdentifier as assigned by the builder; these helpers are only used as keys for aggregation and do not get assigned back onto geometry objects.

I can rename these helpers and add a short comment to make it explicit that these are group keys, to avoid confusion with uniqueness requirements.

dimitra97: Ok I see what you mean. Maybe a comment should be enough! Thanks;)

Examples/Algorithms/Digitization/src/MuonSpacePointDigitizer.cpp
```diff
@@ -21,9 +21,15 @@
 #include "ActsExamples/EventData/MuonSpacePoint.hpp"
 
 #include <algorithm>
+#include <cmath>
 #include <format>
+#include <iterator>
```
junggjo9:
```suggestion
```

Examples/Algorithms/Digitization/src/MuonSpacePointDigitizer.cpp
```diff
@@ -42,10 +51,111 @@ constexpr double quantize(const double x, const double pitch) {
...
+
+/// @brief Build a *candidate key* used to resolve a representative tracking volume
+///        within a sector (volume+layer).
+constexpr GeometryIdentifier toVolumeLayerId(const GeometryIdentifier& id) {
   return GeometryIdentifier{}.withVolume(id.volume()).withLayer(id.layer());
 }
 
+inline double bucketSigmaZ(const MuonSpacePoint& sp,
+                           const double sigmaScale = 1.) {
+  const auto cov = sp.covariance();
```
junggjo9:
```suggestion
  const& auto cov = sp.covariance();
```

Examples/Algorithms/Digitization/src/MuonSpacePointDigitizer.cpp
```diff
@@ -42,10 +51,111 @@ constexpr double quantize(const double x, const double pitch) {
...
+  const double var0 = cov[0];
+  const double var1 = cov[1];
+  const double maxVar = std::max({0., var0, var1});
+  return sigmaScale * std::sqrt(maxVar);
```
junggjo9, line 73: Wait that's a bit edgy. The variances describe the measurement's uncertainty in x and y. How does this relate?

Examples/Algorithms/Digitization/src/MuonSpacePointDigitizer.cpp
```diff
@@ -42,10 +51,111 @@ constexpr double quantize(const double x, const double pitch) {
...
+          return a.layer() < b.layer();
+        }
+        return a.value() < b.value();
+      });
+  sortedIds.erase(std::unique(sortedIds.begin(), sortedIds.end()),
+                  sortedIds.end());
+
+  for (const GeometryIdentifier& volId : sortedIds) {
+    if (trackingGeometry.findVolume(volId) != nullptr) {
+      return volId;
```
junggjo9, line 129: How is it possible that an invalid volume identifier is in the list?

Examples/Algorithms/Digitization/src/MuonSpacePointDigitizer.cpp
```diff
@@ -113,7 +223,7 @@ ProcessCode MuonSpacePointDigitizer::initialize() {
   }
   MuonSpacePointCalibrator::Config calibCfg{};
   m_cfg.calibrator =
-      std::make_unique<MuonSpacePointCalibrator>(calibCfg, logger().clone());
+      std::make_shared<MuonSpacePointCalibrator>(calibCfg, logger().clone());
```
junggjo9:
```suggestion
      std::make_unique<MuonSpacePointCalibrator>(calibCfg, logger().clone());
```

Examples/Algorithms/Digitization/src/MuonSpacePointDigitizer.cpp
```diff
@@ -42,10 +51,111 @@ constexpr double quantize(const double x, const double pitch) {
...
+
+inline void sortBucketPerLayer(MuonSpacePointBucket& bucket) {
+  std::ranges::sort(bucket,
+                    [](const MuonSpacePoint& a, const MuonSpacePoint& b) {
+                      if (a.geometryId().layer() != b.geometryId().layer()) {
+                        return a.geometryId().layer() < b.geometryId().layer();
+                      }
+                      if (a.localPosition().z() != b.localPosition().z()) {
+                        return a.localPosition().z() < b.localPosition().z();
+                      }
```
junggjo9:
```suggestion
                      if (a.localPosition().z() != b.localPosition().z()) {
                        return a.localPosition().z() < b.localPosition().z();
                      }
```

Examples/Algorithms/Digitization/src/MuonSpacePointDigitizer.cpp
```diff
@@ -42,10 +51,111 @@ constexpr double quantize(const double x, const double pitch) {
...
+
+  std::vector<GeometryIdentifier> sortedIds = sectorLayerIds;
+  std::ranges::sort(
+      sortedIds, [](const GeometryIdentifier& a, const GeometryIdentifier& b) {
+        if (a.layer() != b.layer()) {
+          return a.layer() < b.layer();
+        }
+        return a.value() < b.value();
+      });
+  sortedIds.erase(std::unique(sortedIds.begin(), sortedIds.end()),
```
junggjo9:
```suggestion
  sortedIds.erase(std::ranges::unique(sortedIds),
```

paulgessinger:
Is this still relevant?

PR #5278, opened 2026-03-23
feat: switch to ODD v6.0.2 with staged build options

asalzburger:
This PR takes profit of the new staged building of the ODD v6.0.2.

* It introduces staged building with boolean flags and tests those.
* It adds the muon hits to the key4hep copy tests (as they appear now)

It will change a some output variables because the detector will not be identical if only the Tracker is built.

--- END COMMIT MESSAGE ---


paulgessinger (commented):
I think this should be configurable when calling the Python function, and I think we should have a Python test that builds the odd with different configurations. 

Could you look into that?

asalzburger:
> I think this should be configurable when calling the Python function, and I think we should have a Python test that builds the odd with different configurations.
> 
> Could you look into that?

Sure, that was the plan - I wanted to see how bad it looks by just shooting at it ... in terms of CI.

Now I see that some tests already require the Calo to be there ... so this should obviously be part of this PR.

andiwand:
@asalzburger this picked up some changes from a different branch

andiwand:
this has picked up quite some conflicts @asalzburger

PR #5295, opened 2026-03-27
feat: Addition of lepton type for jet labelling

delitez:
In this PR, lepton-type classification is introduced for charged leptons to be used in jet labelling in addition to hadron types. Also, the name "unknown" within hadron types is changed to "others", since it is used to label every remaining hadron after labeling b-, c-, and light jets. "unknown" is now used only for unknown origins.


Core/include/Acts/Definitions/PdgParticle.hpp
```diff
@@ -121,7 +121,14 @@ enum class HadronType {
   StrangeBaryon = 9,
   LightMeson = 10,
   LightBaryon = 11,
-  Unknown = 12
+  Other = 12,
```
paulgessinger, line 124: We could consider keeping the `Unkown` case as an alias and using

```cpp
Unknown __attribute__((deprecated("Don't use"))) = Other,
```

to mark it deprecated? That's not in the standard, but both clang and gcc support it, I believe.

Examples/Algorithms/Jets/src/TruthJetAlgorithm.cpp
```diff
@@ -56,6 +56,20 @@ ActsExamples::JetLabel jetLabelFromHadronType(Acts::HadronType hadronType) {
     case LightMeson:
     case LightBaryon:
       return ActsExamples::JetLabel::LightJet;
+    default:
+      return ActsExamples::JetLabel::OtherJet;
+  }
+}
+
+ActsExamples::JetLabel jetLabelFromLeptonType(Acts::LeptonType leptonType) {
```
paulgessinger, line 64: I'm wondering if we couldn't drop `LeptonType` and instead just have this function accept the `PDG` directly. There's one step less than for hadrons, where we map PGDs to hadron category first, which we don't need for leptons.

Going one step further, if there's a single `jetLabelFromPdgId()` you could just pass in the PDG ids and have this function check if it's a hadron or lepton and return the correct label based on that. This would also simplify the `if/else`s that you have to do further below at this time.

Examples/Algorithms/Jets/src/TruthJetAlgorithm.cpp
```diff
@@ -169,74 +195,95 @@ ProcessCode TruthJetAlgorithm::execute(const AlgorithmContext& ctx) const {
...
-          return std::make_pair(label, particle);
+          ActsExamples::JetLabel label = ActsExamples::JetLabel::Unknown;
+          auto htype = Acts::ParticleIdHelper::hadronType(pdgId);
+          auto ltype = Acts::ParticleIdHelper::leptonType(pdgId);
+
+          if (Acts::ParticleIdHelper::isHadron(pdgId)) {
+            label = jetLabelFromHadronType(htype);
+          } else if (Acts::ParticleIdHelper::isLepton(pdgId)) {
+            label = jetLabelFromLeptonType(ltype);
+          }
```
paulgessinger, line 206: If you ditch `LeptonType`, this also simplifies a bit.

Examples/Framework/include/ActsExamples/EventData/Jets.hpp
```diff
@@ -19,10 +19,14 @@ namespace ActsExamples {
...
 enum class JetLabel {
-  Unknown = 0,
+  OtherJet = 0,
   BJet = 5,
   CJet = 4,
   LightJet = 1,
+  MuonJet = 7,
+  ElectronJet = 8,
+  TauJet = 9,
+  Unknown = 10
```
paulgessinger, line 29: Is `Unknown` still needed, and if yes, can it be an alias of `OtherJet` now?

We also need to be careful: like this, if there's an electron in the jet cone at all, we'd classify it as an `ElectronJet`, which is probably not what we want. I think we should put taus between B and C here.

Core/include/Acts/Definitions/PdgParticle.hpp
paulgessinger: Actually this seems to work after all:

```cpp
  Unknown [[deprecated("nope")]] = Other,
```

This might possibly not work on all compilers that we cover but is worth a try.

Examples/Framework/include/ActsExamples/EventData/Jets.hpp
paulgessinger: Actually, it probably needs to sit below B/C and above the light jets, not in between C and B

PR #5309, opened 2026-04-02
feat: CKF brem recovery

andiwand:
feat: Bremsstrahlung recovery in the CKF

Electrons lose a large, non-Gaussian fraction of their energy to
bremsstrahlung, which a single-component Kalman filter models poorly. This
allows the Combinatorial Kalman Filter to run with a multi-component
(GSF-like) stepper to recover these tracks: on each material surface the track
state is convoluted with the Bethe-Heitler energy-loss distribution, the
resulting Gaussian mixture is propagated, and merged back down for the
measurement update.

The single-component behaviour is unchanged and stays the default. The
multi-component path is selected via the stepper type (`if constexpr`), so
there is no runtime cost for the plain filter and the brem-only state adds no
storage in that case (`[[no_unique_address]]`).

--- END COMMIT MESSAGE ---

## Core changes

- **`CombinatorialKalmanFilter`**
  - New `BremCombinatorialKalmanFilterOptions` extends the plain options with the multi-component parameters (`maxComponents`, `weightCutoff`, `mergeMethod`, `betheHeitlerApprox`). The filter exposes an `Options` alias that resolves to the brem options for a multi-stepper and to the plain options otherwise, so the single-component configuration cannot even name the multi-component knobs.
  - The actor gains a stepper-gated `BremState` (empty `NoBremState` for the single-component case) holding the Bethe-Heitler config and scratch buffers, wired in from `findTracks`.
  - Material interaction, the Kalman update and the stepper update are factored into helpers (`performMaterialInteraction`, `kalmanUpdate`, `updateStepper`) that branch on `IsMultiStepper`. In the multi-component branch these apply the Bethe-Heitler mixture, reduce/normalise components, and merge the mixture for the stored track state.
- **`CombinatorialKalmanFilterExtensions`**: new `mixtureReducer` delegate (only used by the multi-stepper path) to reduce the component count.
- **GSF refactor**: `convoluteComponents` is removed from `GsfUtils` and inlined into `GsfActor`, so the per-component Bethe-Heitler application is shared by the GSF and the brem CKF via `applyBetheHeitler`. Minor numeric-literal / `emplace_back` cleanups in `GsfUtils.cpp`.

## Examples / Python

- `TrackFindingAlgorithm` gains a second, optional `findTracksBrem` function. Seeds whose particle hypothesis is `electron` are routed to the brem finder when it is configured; everything else uses the plain finder.
- `makeBremTrackFinderFunction` builds the CKF over a `MultiStepperLoop<SympyStepper>` and is exposed to Python; `addCKFTracks` wires it up alongside the existing finder, using the KL-distance mixture reducer and the default ATLAS Bethe-Heitler approximation.

## Notes / TODOs

A few implementation shortcuts are marked with `TODO` in the code and worth a look during review:
- The multi-component bound state triggers a second covariance transport because stepper components cannot yet be matched to bound-parameter components.
- `updateStepper` currently relies on the temporary states coinciding with the current state; storing the multi-component bound parameters on the track state would be more robust, especially with branching.


Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithmFunction.cpp
benjaminhuth, line 1: Would it make sense to split this in two compilation units? Probably premature optimization though...

benjaminhuth (dismissed):
Looks good to me!

benjaminhuth (approved):

PR #5313, opened 2026-04-03
fix:  narrowing and implicit conversion fixes done by AI

asalzburger:
# Address Sonar implicit conversion and narrowing findings

*** This has been done using our SonarCloud output by cursor *** 

## Summary

This change set makes implicit numeric conversions explicit across Core, Alignment, and related headers/sources, targeting SonarCloud rules around **implicit conversion** and **narrowing** (integer precision loss, `double`→`float`, and `std::size_t` mixed with narrower index or arithmetic types). The intent is mechanical clarity for static analysis and alignment with strict floating-point warnings (for example `-Wfloat-conversion`), without altering algorithms.

## Motivation

Sonar reported a large cluster of findings where:

- `std::size_t` (or other wide integers) were used where API types expect `int`, trajectory index types, or `float`-backed binning.
- Physical quantities expressed with `UnitConstants` (typically `double`) were assigned to `float` configuration fields without an explicit cast.
- Arithmetic combined signed and unsigned index offsets in ways that are easy for analyzers to flag (for example `std::size_t` plus a negative `int`).

Making these conversions explicit documents intent and removes or reduces the corresponding static-analysis noise.

## Approach

- Prefer **`static_cast<T>(…)`** at boundaries (sizes to index types, integrals to `double` for division, `double` to `float` for stored parameters).
- Use **`…f` literals** where a `float` is clearly intended.
- For **axis / neighborhood index math**, use `static_cast<int>(idx)` before adding signed offsets, and `static_cast<int>(getNBins())` where a narrow `int` range is required.
- **`fastCathetus`**: promote `hypotenuse` to `double` before the difference/sum under `sqrt` so mixed-type call sites stay well defined.
- **`hypotSquare`** was **not** changed to always return `double`; doing so altered deduced return types for `fastHypot` and broke a lambda that mixed `fastHypot` and scalar returns. Instead, **`BroadTripletSeedFilter`’s `getTopR`** lambda is explicitly `-> float` with `static_cast<float>(fastHypot(…))`.

## Areas touched (high level)

| Area | Examples |
|------|-----------|
| **Alignment** | χ² / ndf averaging: explicit `double` casts for counts. |
| **Track fitting** | Kalman / GSF / smoother / updater: index types and `float` χ² assignment; fitter result indices aligned with `TrackIndexType` where applicable. |
| **Event data** | Track/space-point containers, proxies: `size()` / `distance` → explicit index or `uint32_t` casts. |
| **Propagator / material** | `float` paths: explicit casts from `double` or Eigen scalars. |
| **Seeding / Seeding2** | Config defaults with `UnitConstants`, GbTS filter literals, geometry `numOfBins` / `numLayers`, graph-seeder floats, triplet/doublet `f` suffixes. |
| **Surfaces** | `FacesHelper`: `anker` and half-count geometry use explicit casts. |
| **Track finding** | `MeasurementSelector`, `TrackParamsLookupAccumulator`: `float` χ₂ and `double` promotion for count-weighted parameters. |
| **Utilities** | `Axis`, `BinAdjustment` (+ volume variant), `BinningData`, `Polynomials` (`product` → `double`), `MathHelpers` (`fastCathetus`). |
| **Vertexing** | `GaussianGridTrackDensity` config: `binSize` and literals. |
| **Visualization** | `EventDataView3D`, `Interpolation3D`: loop indices promoted for `double` math. |
| **Core sources** | Greedy ambiguity resolution ratios, `ParticleData` SEMF terms, seed/space-point `reserve`, `VectorMultiTrajectory` / `VectorTrackContainer`, geometry builders (`Cuboid`/`Cylinder`/`LayerArray`/`CylinderVolumeHelper`). |

## Verification

- Local CMake (minimal preset-style) build with **`-Wfloat-conversion`** enabled in project flags: **`ActsCore`** and **`ActsAlignment`** link successfully.

## Follow-up (optional)

- Re-run Sonar and, if desired, mark matching issues **Resolved** in `sc_issues.json` (same housekeeping as for other rule batches).
- `narrowing_issues.tsv` (~179 rows) was a one-off extract from an export; refresh or remove from the repo if it should not be tracked.

## Related

- Sonar rule families: implicit conversion, narrowing, floating-point precision in assignments.

--- END COMMIT MESSAGE ---


paulgessinger:
There's a number of new issues.

asalzburger:
> There's a number of new issues.

Yeah, many of them are:

```Replace the redundant type with "auto"```

where SonarCloud is giving the opposite recommendation than Athena coding rules, as far as I remember.

File: Core/include/Acts/Seeding2/CylindricalSpacePointGrid2.hpp [hunk 1]
```diff
@@ -53,23 +53,23 @@ class CylindricalSpacePointGrid2 {
     /// distance from x=y=0 (i.e. in r)
     /// WARNING: if rMin is smaller than impactMax, the bin size will be 2*pi,
     /// which will make seeding very slow!
-    float rMin = 0 * UnitConstants::mm;
+    float rMin = static_cast<float>(0 * UnitConstants::mm);
```
asalzburger, line 56: We can debate this if that's really necessary.

Core/src/Definitions/ParticleData.cpp
```diff
@@ -164,24 +164,30 @@ float Acts::calculateNucleusMass(Acts::PdgParticle pdg) {
...
-  float a_Vol = 15.260f * Acts::UnitConstants::MeV;
-  float a_Surf = 16.267f * Acts::UnitConstants::MeV;
-  float a_Col = 0.689f * Acts::UnitConstants::MeV;
-  float a_Sym = 22.209f * Acts::UnitConstants::MeV;
-  float a_Pair =
-      (1 - 2 * (Z % 2)) * (1 - A % 2) * 10.076f * Acts::UnitConstants::MeV;
-
-  float massP = 0.938272f * Acts::UnitConstants::GeV;
-  float massN = 0.939565f * Acts::UnitConstants::GeV;
+  auto a_Vol = static_cast<float>(15.260 * Acts::UnitConstants::MeV);
```
asalzburger, line 167: This doesn't seem to apply our naming conventions!

asalzburger:
This is now reviewed by me, there's a couple of cases where we would not necessary need it, e.g.

```c++
float var = 1. * Units::mm;
```

which will now have a 

```c++
float var = static_cast<float>(1.*Unis::mm);
```

But ok, in order to catch real implicit conversions we may have to accept the latter.

benjaminhuth:
image

andiwand:
this has picked up some conflicts @asalzburger

PR #5330, opened 2026-04-10
feat: Python bindings for SimHit/SimHitContainer + UprootReader

benjaminhuth:
Implement `uproot`-based particle and simhit readers for the PyPI distribution of acts.

--- END COMMIT MESSAGE ---


Python/Examples/python/uproot.py
```diff
@@ -0,0 +1,197 @@
...
+
+        with uproot.open(str(filePath)) as f:
+            tree = f["particles"]
+            self._data = tree.arrays(library="np")
+            event_ids = self._data["event_id"]
+            self._entry_map = {int(eid): i for i, eid in enumerate(event_ids)}
+
+        all_ids = set(self._entry_map.keys())
+        self._min_event = int(min(all_ids))
+        self._max_event = int(max(all_ids))
```
andiwand:
```suggestion
        self._min_event = min(self._entry_map.keys())
        self._max_event = max(self._entry_map.keys())
```
or is there a good reason for the set and additional `int` conversion?

Python/Examples/python/uproot.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def initialize(self):
+        self._particleHandle.initialize(self._outputParticles)
+        return acts.examples.ProcessCode.SUCCESS
+
+    def read(self, context):
+        event_number = context.eventNumber
+        particles = acts.examples.SimParticleContainer()
+
+        if event_number in self._entry_map:
```
andiwand, line 68:
```suggestion
        if event_number not in self._entry_map:
          self._particleHandle(context, particles)
          return acts.examples.ProcessCode.SUCCESS

        # ...
```

might be a good case for early exit? not sure what we expect to happen in this case so success might be fair

similar below

Python/Examples/python/uproot.py
```diff
@@ -0,0 +1,197 @@
...
+
+        all_ids = set(self._entry_map.keys())
+        self._min_event = int(min(all_ids))
+        self._max_event = int(max(all_ids))
+
+    def availableEvents(self):
+        return (self._min_event, self._max_event + 1)
+
+    def initialize(self):
+        self._particleHandle.initialize(self._outputParticles)
```
andiwand, line 61: this we usually do in the constructor no?

Python/Examples/src/Framework.cpp
```diff
@@ -101,6 +101,47 @@ class PyIAlgorithm : public IAlgorithm {
   const Acts::Logger& pyLogger() const { return logger(); }
 };
 
+class PyIReader : public IReader {
+  // Typedef avoids comma in macro argument for PYBIND11_OVERRIDE_PURE
+  using EventRange = std::pair<std::size_t, std::size_t>;
```
andiwand, line 106: macro love 💕

andiwand (dismissed):
LGTM - just left a few minor comments

andiwand (approved):

PR #5336, opened 2026-04-12
feat: Tracking geometry json converter

ssdetlab:
Adding descriptor-based tracking geometry json converter. The converter takes a tracking geometry instance and serializes individual elements. The elements, such as portals and surfaces, are converted independently of their owning objects, like volumes. The relationship between the element and the owner is preserved via index-based linking. 

Adding utility methods to some portal links and navigation classes for (de)serialization. 

Adding axis json deserialization function.    

The tests cover the round-trip conversion of the tracking geometry, including the geometry hierarchy checks and navigation/propagation checks.


Plugins/Json/include/ActsPlugins/Json/TrackingGeometryJsonConverter.hpp
```diff
@@ -0,0 +1,269 @@
...
+///     volumes via portal IDs
+///   - return a reconstructed world `TrackingVolume` (or `TrackingGeometry`)
+class TrackingGeometryJsonConverter {
+ public:
+  /// JSON serialization options for tracking geometry conversion.
+  struct Options {};
+
+  /// Generic lookup from object pointer identity to serialized object ID.
+  template <typename object_t, const char* kContext>
+  struct PointerToIdLookup {
```
paulgessinger: Could this be private? In fact, I think since this is only used as a name in the header, we could possibly even move this to the `.cpp` file.

Plugins/Json/include/ActsPlugins/Json/TrackingGeometryJsonConverter.hpp
```diff
@@ -0,0 +1,269 @@
...
+    // private:
+    std::unordered_map<const object_t*, std::size_t> m_objectIds;
+  };
+
+  /// Generic lookup from serialized ID to pointer-like object holder.
+  ///
+  /// `pointer_t` can be a raw pointer (`object_t*`) or an owning pointer-like
+  /// type such as `std::shared_ptr<object_t>`.
+  template <typename object_t, typename pointer_t, const char* kContext>
+  struct IdToPointerLikeLookup {
```
paulgessinger: Same for this I think.

Plugins/Json/include/ActsPlugins/Json/TrackingGeometryJsonConverter.hpp
```diff
@@ -0,0 +1,269 @@
...
+      return it->second;
+    }
+
+   private:
+    std::unordered_map<std::size_t, pointer_t> m_objects;
+  };
+
+  static inline constexpr char kSurfaceLookupContext[] = "surface";
+  static inline constexpr char kPortalLookupContext[] = "portal";
+  static inline constexpr char kVolumeLookupContext[] = "volume";
```
paulgessinger, line 72: `const char*` ?

Plugins/Json/include/ActsPlugins/Json/TrackingGeometryJsonConverter.hpp
```diff
@@ -0,0 +1,269 @@
...
+                        const TrackingVolume& world,
+                        const Options& options = Options{}) const;
+
+  /// Reconstruct a tracking geometry from JSON.
+  std::shared_ptr<TrackingGeometry> trackingGeometryFromJson(
+      const GeometryContext& gctx, const nlohmann::json& encoded,
+      const Options& options = Options{});
+
+  /// Serialize one portal link using the configured dispatcher.
+  nlohmann::json portalLinkToJson(const GeometryContext& gctx,
```
paulgessinger, line 205: Should we keep these public or try to make them private?

Tests/UnitTests/Plugins/Json/TrackingGeometryJsonConverterTests.cpp
```diff
@@ -0,0 +1,491 @@
...
+  };
+  std::function<void(const Acts::PortalLinkBase* linkA,
+                     const Acts::PortalLinkBase* linkB)>
+      checkLinks = [&](const auto* linkA, const auto* linkB) {
+        const auto* trivialA =
+            dynamic_cast<const Acts::TrivialPortalLink*>(linkA);
+        const auto* trivialB =
+            dynamic_cast<const Acts::TrivialPortalLink*>(linkB);
+        if (trivialA != nullptr) {
+          BOOST_CHECK_NE(trivialB, nullptr);
```
paulgessinger: These need to be `BOOST_REQUIRE_NE` I think, otherwise it will dereference the `nullptr` below and crash. This goes for a number of comparable locations below.

Plugins/Json/src/TrackingGeometryJsonConverter.cpp
```diff
@@ -0,0 +1,1028 @@
...
+                       Acts::GeometryIdentifier::Value& nextVolumeId) {
+  Acts::GeometryIdentifier volumeId = volume.geometryId();
+  if (volumeId == Acts::GeometryIdentifier{}) {
+    volumeId = Acts::GeometryIdentifier{}.withVolume(nextVolumeId++);
+    volume.assignGeometryId(volumeId);
+  }
+
+  for (const auto [ib, boundary] : Acts::enumerate(volume.boundarySurfaces())) {
+    auto& mutableBoundarySurface =
+        const_cast<Acts::RegularSurface&>(boundary->surfaceRepresentation());
```
paulgessinger, line 601: This is acceptable even if sonar complains because it's the Gen1 API.

Plugins/Json/src/TrackingGeometryJsonConverter.cpp
```diff
@@ -0,0 +1,1028 @@
...
+      orderedPortals.push_back(&port);
+    }
+  };
+  auto insertVolume = [&](const auto& vol) {
+    if (volumeIds.emplace(vol, orderedVolumes.size())) {
+      orderedVolumes.push_back(&vol);
+    }
+  };
+
+  std::function<void(const Acts::PortalLinkBase* link)> insertPortalLink =
```
benjaminhuth: Could this be a lambda as well, or is there a reason why this is a `std::function`?

Plugins/Json/src/TrackingGeometryJsonConverter.cpp
benjaminhuth: I would think this deserves a code comment stating what @paulgessinger wrote.

Plugins/Json/src/TrackingGeometryJsonConverter.cpp
```diff
@@ -0,0 +1,1028 @@
...
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+#include <nlohmann/json_fwd.hpp>
+#include <unistd.h>
+
+namespace {
```
benjaminhuth, line 124: I wonder if it would improve structure and readability, if one would split up navigation encoders/decoders in a separate file?

Plugins/Json/src/TrackingGeometryJsonConverter.cpp
```diff
@@ -0,0 +1,1028 @@
...
+#include <ranges>
+#include <stdexcept>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+#include <nlohmann/json_fwd.hpp>
+#include <unistd.h>
```
benjaminhuth: Just out of curiosity: Why do we need the POSIX header here?

Plugins/Json/include/ActsPlugins/Json/TrackingGeometryJsonConverter.hpp
ssdetlab: These are used as template arguments later in the code. `const char*` is not allowed to be a template argument by the standard.

Edit: to clarify the language -- string literals are not allowed to be passed as a template arguments.

Plugins/Json/src/TrackingGeometryJsonConverter.cpp
ssdetlab: My clangd adds them automatically sometimes, when it thinks that the existing headers are not enough. Removed

Plugins/Json/src/TrackingGeometryJsonConverter.cpp
ssdetlab: This function is recursive, and generally lambdas are not allowed to be recursive. 

There's a workaround pattern:
`auto insertPortalLink = [&](const auto& link, auto&& self) -> void`

And you pass the lambda in the call as an argument:
`insertPortalLink(along, insertPortalLink);`

Compared to `std::function` this is a bit less of overhead, but the call syntax is a bit more annoying. 

I'm not married to either of those, so I don't mind switching to the lambda-pattern.

Plugins/Json/src/TrackingGeometryJsonConverter.cpp
ssdetlab: From what we've discussed with @paulgessinger, the general idea is to slowly move the json converters to the `Dispatcher` pattern. eventually, these encoders/decoders will be split away into their dedicated classes (see, e.g., how it looks now in `SurfaceJsonConverter`. 

For the time being I'd keep all of these in the same place, so that it's easier to remember what has to be done. But I don't mind splitting it away right now.

Plugins/Json/include/ActsPlugins/Json/TrackingGeometryJsonConverter.hpp
ssdetlab: Until the encoder/decoder functions that are declared in the `.cpp` file, such as `encodeCompositePortalLink`, are parts of their respective classes they rely on the public access to these to/from json methods to get to the converters in the `TrackingGeometryJsonConverter` config. 

I'm sure that this can be refactored-away, but with its own tradeoffs. Also, we can just iterate the `.cpp` functions out into their own converter classes (which is the goal, from my understanding). So I'd keep them public for now.

Plugins/Json/src/TrackingGeometryJsonConverter.cpp
ssdetlab: Done

Tests/UnitTests/Plugins/Json/TrackingGeometryJsonConverterTests.cpp
ssdetlab: Changed

Plugins/Json/include/ActsPlugins/Json/TrackingGeometryJsonConverter.hpp
ssdetlab: The declaration still has to be public, since the public config depends on the types that depend on these two structs. Moved the definition in the `.cpp` file

Plugins/Json/include/ActsPlugins/Json/TrackingGeometryJsonConverter.hpp
ssdetlab: Same as above

Plugins/Json/src/TrackingGeometryJsonConverter.cpp
benjaminhuth: Yeah makes sense, I don't think this makes any difference performance wise in this I/O code, so you can keep it as is I would say

Plugins/Json/include/ActsPlugins/Json/TrackingGeometryJsonConverter.hpp
```diff
@@ -0,0 +1,253 @@
...
+      Config config = Config::defaultConfig());
+
+  /// @brief Convert a tracking geometry to JSON.
+  ///
+  /// @param gctx geometry context
+  /// @param geometry tracking geometry to convert
+  /// @param options options for the conversion
+  ///
+  /// @return serialized tracking geometry
+  nlohmann::json toJson(const GeometryContext& gctx,
```
benjaminhuth, line 159: A question: Should this take the geometry context?

benjaminhuth: --> properly comment what is needed here

benjaminhuth (approved):
LGTM

PR #5339, opened 2026-04-13
feat: add geometric filter options for strip seeding 

mvessell96:
Adds optional filters on strip doublet cot(theta) + optional weighting on strip seed delta(eta) for displaced seeds
--- END COMMIT MESSAGE ---

Pixel triplet seeding order is sorted by doublet cot(theta) (`sortedByCotTheta`), which saves on execution time since the iteration can be aborted once a threshold is reached. Something similar doesn't exist for strip seeding, likely because the main bottleneck is the very expensive `stripCoordinateCheck` done before the global coord. transformation. It's still possible to implement a very loose criteria based on the approximate cot(theta) _before_ this check is done that can save on execution time while still being conservative. Implementing a threshold of 0.5 here saves ~10-15% off of the total ATLAS LRT seeding time with near-negligible (LLP) efficiency loss.

Additionally, the ATLAS Run 3 LRT pass (https://arxiv.org/pdf/2304.12867, pg 11) had an additional weight component in the seed ordering based on deltaEta = |eta_seed − eta_prime|, where eta_prime is the pseudorapidity of the vector connecting the beam spot position with the point of closest approach of the seed. Seeds from LLP decays tend toward small values of deltaEta due to the correlated direction of the LLP and its decay products, in contrast to seeds corresponding to fake tracks. The ACTS seed filter does not quite function the same as the legacy Athena seed ordering, but a similar function can be emulated in the seed filter weight as an optional factor. This should give a higher weight to genuine displaced seeds, allowing us to reduce the number of seeds per middle SP from 3->2 or 1 downstream, which has another ~15-20% speedup potential in the finding stage.

Both are disabled by default and can just be configured downstream

FYI @jburzy


Core/src/Seeding2/BroadTripletSeedFilter.cpp
```diff
@@ -239,6 +240,17 @@ void BroadTripletSeedFilter::filterTripletTopCandidates(
...
 
+    // Delta eta weight: penalize seeds whose eta is
+    // inconsistent with the vector from the beamspot to the point of
+    // closest approach. Real displaced decay seeds have small delta eta.
+    if (config().deltaEtaWeightFactor > 0 &&
+        impact > config().deltaEtaMinImpact) {
+      float etaSeed = std::asinh(bottomLink.cotTheta());
+      float etaPCA = std::asinh(zOrigin / impact);
+      float deltaEta = std::abs(etaSeed - etaPCA);
+      weight -= config().deltaEtaWeightFactor * deltaEta;
```
andiwand:
```suggestion
      const float etaSeed = std::asinh(bottomLink.cotTheta());
      const float etaPCA = std::asinh(zOrigin / impact);
      const float absDeltaEta = std::abs(etaSeed - etaPCA);
      weight -= config().absDeltaEtaWeightFactor * absDeltaEta;
```

Core/include/Acts/Seeding2/BroadTripletSeedFilter.hpp
```diff
@@ -100,6 +100,16 @@ class BroadTripletSeedFilter final : public ITripletSeedFilter {
     /// Number of seeds required before `seedWeightIncrement` is applied
     float numSeedIncrement = std::numeric_limits<float>::infinity();
 
+    /// Weight factor for the delta-eta penalty between the seed direction
+    /// and the vector from the beamspot to the seed's pca. 
+    /// Seeds from LLP decays tend toward small delta-eta
+    /// values due to the correlated direction of the LLP and its decay
+    /// products. Set to 0 (default) to disable.
+    float deltaEtaWeightFactor = 0;
```
andiwand:
```suggestion
    float absDeltaEtaWeightFactor = 0;
```
though "abs delta eta" seems a bit loaded; in `Core/src/Seeding2/BroadTripletSeedFilter.cpp` it seems more important to me. if we go for that we might also rename `deltaEtaMinImpact`

Core/src/Seeding2/TripletSeedFinder.cpp
```diff
@@ -325,6 +325,17 @@ class Impl final : public TripletSeedFinder {
     const StripData stripB = calculateStripData(spB);
 
     for (auto topDoublet : topDoublets) {
+      // Pre-filter on the doublet stage cot(theta) difference before the
+      // expensive strip coordinate transformation. The doublet cot(theta)
+      // values are computed from SP centers and are approximate
+      // so the cut is very loose
+      {
+        const float deltaCotTheta = cotThetaB - topDoublet.cotTheta();
+        if (deltaCotTheta * deltaCotTheta > m_cfg.cotThetaDiffMax2) {
```
andiwand:
```suggestion
        const float deltaCotTheta = cotThetaB - topDoublet.cotTheta();
        if (std::abs(deltaCotTheta) > m_cfg.cotThetaDiffMax) {
```
do you plan to use other error terms in the future? otherwise we could also go with the non-squared difference

Core/include/Acts/Seeding2/TripletSeedFinder.hpp
```diff
@@ -166,6 +166,12 @@ class TripletSeedFinder {
     /// coordinates in xyz. This is only used in a detector specific check for
     /// strip modules
     float toleranceParam = 1.1 * UnitConstants::mm;
+
+    /// Maximum allowed difference in cot(theta) between the bottom and top
+    /// doublets, applied as a pre-filter before the expensive strip
+    /// coordinate transformation. Only active when useStripInfo is true.
+    /// Set to infinity (default) to disable.
+    float cotThetaDiffMax = std::numeric_limits<float>::infinity();
```
andiwand, line 174: I wonder if this should only be done for strips or if we can further generalize it between pixel and strips; but no concrete idea from my side right now

andiwand (commented):
generally looks good to me 👍  left a few minor suggestions and comments

> Pixel triplet seeding order is sorted by doublet cot(theta) (sortedByCotTheta), which saves on execution time since the iteration can be aborted once a threshold is reached. Something similar doesn't exist for strip seeding, likely because the main bottleneck is the very expensive stripCoordinateCheck done before the global coord. transformation. It's still possible to implement a very loose criteria based on the approximate cot(theta) before this check is done that can save on execution time while still being conservative. Implementing a threshold of 0.5 here saves ~10-15% off of the total ATLAS LRT seeding time with near-negligible (LLP) efficiency loss.

I remember wondering about this - if you think it makes sense to harmonize this we can definitely do it. I faintly remember that `cotTheta` then depends on calibrated / non-calibrated quantities which could be individually above or below threshold potentially hurting efficiency.

Core/include/Acts/Seeding2/TripletSeedFinder.hpp
mvessell96: edit: oop I originally mixed up which line this was on

The main reason this saves so much time in the strip seeding is because it's purposefully being done before the very expensive strip coord check, I think there would be diminishing returns to also doing the "pre filter" in the pixels case given that it's already doing the cottheta sort. My gut feeling is that it wouldn't move the needle that much, but might still be interesting to check

mvessell96:
> generally looks good to me 👍 left a few minor suggestions and comments
> 
> > Pixel triplet seeding order is sorted by doublet cot(theta) (sortedByCotTheta), which saves on execution time since the iteration can be aborted once a threshold is reached. Something similar doesn't exist for strip seeding, likely because the main bottleneck is the very expensive stripCoordinateCheck done before the global coord. transformation. It's still possible to implement a very loose criteria based on the approximate cot(theta) before this check is done that can save on execution time while still being conservative. Implementing a threshold of 0.5 here saves ~10-15% off of the total ATLAS LRT seeding time with near-negligible (LLP) efficiency loss.
> 
> I remember wondering about this - if you think it makes sense to harmonize this we can definitely do it. I faintly remember that `cotTheta` then depends on calibrated / non-calibrated quantities which could be individually above or below threshold potentially hurting efficiency.

Yeah indeed there is this issue with calibrated vs non-calibrated, I think the difference is not too meaningful when the threshold is very coarse, since seeds whose doublets are this incompatible likely shouldn't be pushed into being compatible after any updates. For the LRT/displaced strips case for certain it's well-motivated, since much worse tradeoffs have to be made for CPU vs efficiency in the end... but for this reason I thought it best to just keep the default behavior by default rather than just implementing it directly

Core/src/Seeding2/TripletSeedFinder.cpp
mvessell96: I hadn't thought that far ahead but good point

Core/include/Acts/Seeding2/BroadTripletSeedFilter.hpp
mvessell96: All updated now

Core/src/Seeding2/BroadTripletSeedFilter.cpp
mvessell96: Done

andiwand (approved):

PR #5340, opened 2026-04-13
fix: Propagate edge features in edge classification modules

benjaminhuth:
Previously, the edge features were not selected and propagated, making it impossible to chain classifiers that need the edge features.

--- END COMMIT MESSAGE ---


Plugins/Gnn/src/Tensor.cpp
```diff
@@ -222,4 +182,101 @@ std::pair<Tensor<std::int64_t>, std::optional<Tensor<float>>> applyEdgeLimit(
...
+  }
+  return mask;
+}
+
+template <Acts::Concepts::arithmetic T>
+Tensor<T> selectRows(const Tensor<T> &tensor, const Tensor<bool> &mask,
+                     const ExecutionContext &execContext) {
+  assert(tensor.shape()[0] == mask.shape()[0]);
+  assert(mask.shape()[1] == 1);
+  assert(tensor.device() == mask.device());
```
andiwand: nitpick: could even be an exception in this case?

same below

Plugins/Gnn/src/Tensor.cu
```diff
@@ -34,6 +35,40 @@ __global__ void applyCut(std::size_t size, float cutoff, const float *array,
...
+template <typename T>
+__global__ void gatherRowsKernel(const std::size_t *indices, std::size_t nOut,
+                                 std::size_t nCols, const T *src, T *dst) {
+  const std::size_t k = blockIdx.x * blockDim.x + threadIdx.x;
+  if (k >= nOut) {
+    return;
+  }
+  const std::size_t srcRow = indices[k];
+  for (std::size_t j = 0; j < nCols; ++j) {
+    dst[k * nCols + j] = src[srcRow * nCols + j];
```
andiwand, line 51: is column column-major expected by all as input/output? is this worth documenting?

andiwand (dismissed):
looks fine by me

Plugins/Gnn/src/Tensor.cu
benjaminhuth: Hmm I think row-major is expected here... Hmm yeah I think these are the only tensor-related functions that actually make this assumption, I'll document this in the header

Plugins/Gnn/src/Tensor.cpp
benjaminhuth: Hmm yeah true

andiwand (approved):

PR #5341, opened 2026-04-13
chore!: Remove templated version of HoughVertexFinder

pbalek:
remove templated HoughVertexFinder

--- END COMMIT MESSAGE ---

This PR removes the original version of `HoughVertexFinder` that used a template for spacepoints type. Also removes respective unit tests.

The new version `HoughVertexFinder2` using `SpacePointContainer2` is available since #5108 and remains unchanged. The is also `HoughVertexFinderAlgorithm` that already uses this new version.


andiwand:
thanks for handling this @pbalek ! I would delay merging this until we are ready to cut the next major release

andiwand (approved):

PR #5342, opened 2026-04-13
feat: Adding beam spot as an additional measurement to refitting algorithm 

delitez:
In this PR, an option to include beam-spot measurements in the KF refitting algorithm is introduced. If enabled, a measurement at (0,0,0) with beam-spot covariance will be added to source links, using the surface of the current track state.


delitez:
@andiwand could you take a look at this draft PR?

Examples/Scripts/Python/truth_tracking_kalman.py
andiwand, line 1: I believe this should be reverted

Examples/Io/Root/src/RootTrackStatesWriter.cpp
```diff
@@ -409,6 +409,9 @@ ProcessCode RootTrackStatesWriter::writeT(const AlgorithmContext& ctx,
       } else {
         // get the truth hits corresponding to this trackState
         // Use average truth in the case of multiple contributing sim hits
+        if (state.referenceSurface().geometryId().value() == 0u) {
+          continue;
+        }
```
andiwand, line 414: I think the main limitation is that the source link will not contain an `IndexSourceLink`. we could generalize this by extending the
```c++
if (!state.hasUncalibratedSourceLink()) {
```
with something like
```c++
if (!state.hasUncalibratedSourceLink() || state.getUncalibratedSourceLink().template getPtr<IndexSourceLink>() == nullptr) {
```

we could also turn the `if` upside down to first handle the valid source link case and `else` adding the `nan`s but that is not so important

Examples/Framework/src/Validation/TrackClassification.cpp
```diff
@@ -108,6 +108,9 @@ void ActsExamples::identifyContributingParticles(
     if (state.typeFlags().isOutlier()) {
       continue;
     }
+    if (state.referenceSurface().geometryId().value() == 0u) {
+      continue;
+    }
```
andiwand, line 113:
```suggestion
    if (state.getUncalibratedSourceLink().template getPtr<IndexSourceLink>() == nullptr) {
      continue;
    }
```
similar like below I think this is more generic

Examples/Algorithms/TrackFitting/src/RefittingAlgorithm.cpp
```diff
@@ -109,11 +155,16 @@ ProcessCode RefittingAlgorithm::execute(const AlgorithmContext& ctx) const {
...
+      trackSourceLinks.push_back(Acts::SourceLink{beamSpotSL});
+      surfSequence.push_back(&beamSpotConstTrackState.referenceSurface());
+    }
+
     std::ranges::reverse(surfSequence);
 
-    ACTS_VERBOSE("Initial parameters: "
-                 << initialParams.fourPosition(ctx.geoContext).transpose()
-                 << " -> " << initialParams.direction().transpose());
+    ACTS_VERBOSE("Initial parameters: " << track.parameters().transpose());
```
andiwand, line 167: the original version seems to have more information

Examples/Algorithms/TrackFitting/src/RefittingAlgorithm.cpp
andiwand, line 125: I think it would be good to create a new central perigee surface in the refitting algorithm somewhere at the top of `execute` and supply it also as a reference here

similar to here https://github.com/acts-project/acts/blob/46df76a2d954ee711cd4c4212401ff10342d7cdc/Examples/Algorithms/TrackFitting/src/TrackFittingAlgorithm.cpp#L88-L90

Examples/Algorithms/TrackFitting/src/RefittingAlgorithm.cpp
```diff
@@ -53,6 +54,51 @@ ProcessCode RefittingAlgorithm::execute(const AlgorithmContext& ctx) const {
...
+
+    const Acts::Vector2 beamSpotMeasValue{0., 0.};
+    Acts::SquareMatrix2 inflatedCov =
+        Acts::SquareMatrix2::Zero();  //* 12.5 * Acts::UnitConstants::um;
+    inflatedCov(0, 0) =
+        12.5 *
+        Acts::UnitConstants::um; 
+    inflatedCov(1, 1) =
+        55.5 *
+        Acts::UnitConstants::mm;
```
andiwand, line 72: these should be configurable

Examples/Algorithms/TrackFitting/include/ActsExamples/TrackFitting/RefittingAlgorithm.hpp
```diff
@@ -34,6 +34,8 @@ class RefittingAlgorithm final : public IAlgorithm {
     int pickTrack = -1;
     /// Inflate initial covariance.
     std::array<double, 6> initialVarInflation = {1., 1., 1., 1., 1., 1.};
+    /// Add a beam spot measurement
+    bool addBeamSpotMeasurement = false;
```
andiwand, line 38: since we also need the beam spot position and error we could go with an optional like

```suggestion
    /// Add a beam spot measurement
    std::optional<Acts::SquareMatrix2> beamSpotConstraint;
```

Examples/Io/Root/src/RootTrackStatesWriter.cpp
```diff
@@ -409,6 +409,11 @@ ProcessCode RootTrackStatesWriter::writeT(const AlgorithmContext& ctx,
       } else {
         // get the truth hits corresponding to this trackState
         // Use average truth in the case of multiple contributing sim hits
+        if (!state.hasUncalibratedSourceLink() ||
+            state.getUncalibratedSourceLink()
+                    .template getPtr<IndexSourceLink>() == nullptr) {
+          continue;
+        }
```
andiwand: is this code ever reached? since we first check if there is no source link else again if there is no source link
```c++
if (!state.hasUncalibratedSourceLink()) {
  // ...
} else {
  if (!state.hasUncalibratedSourceLink() ||
      state.getUncalibratedSourceLink()
               .template getPtr<IndexSourceLink>() == nullptr) {
      continue;
  }
  // ...
}
```

Examples/Algorithms/TrackFitting/src/RefittingAlgorithm.cpp
```diff
@@ -109,6 +155,13 @@ ProcessCode RefittingAlgorithm::execute(const AlgorithmContext& ctx) const {
       continue;
     }
 
+    if (m_cfg.beamSpotConstraint.has_value()) {
+      beamSpotSL =
+          RefittingCalibrator::RefittingSourceLink{beamSpotConstTrackState};
+      trackSourceLinks.push_back(Acts::SourceLink{beamSpotSL});
```
andiwand: sonarcloud suggests
```suggestion
      trackSourceLinks.emplace_back(beamSpotSL);
```
not sure if that works with all the templating but we can try

Examples/Algorithms/TrackFitting/src/RefittingAlgorithm.cpp
```diff
@@ -53,6 +55,51 @@ ProcessCode RefittingAlgorithm::execute(const AlgorithmContext& ctx) const {
...
+  auto beamSpotTrackState = beamSpotVectorTrackStateContainer->makeTrackState();
+
+  const Acts::Vector2 beamSpotMeasValue{0., 0.};
+
+  if (inputTracks.size() == 0) {
+    ACTS_INFO("Input tracks collection is empty");
+    return ProcessCode::SKIP;
+  }
+  auto trackRefSurfacePtr = inputTracks.at(0).referenceSurface().getSharedPtr();
+  beamSpotTrackState.setReferenceSurface(trackRefSurfacePtr);
```
andiwand:
```suggestion
  beamSpotTrackState.setReferenceSurface(perigeeSurface);
```
with that the reference surface should be consistent everywhere

one tricky part we don't need to handle here and now is that the start parameters we feed into the KF are taken at the previous reference surface which could be in a different location than the one we create here. also the navigation might get confused by the bound parameters that are exactly in the same spot as the new reference surface. if this works for now somehow we can push this to a follow up. if this causes problems we can revert to your previous working version

ultimately we can extrapolate from the previous starting bound params to the new reference surface and go from there

Examples/Algorithms/TrackFitting/include/ActsExamples/TrackFitting/RefittingAlgorithm.hpp
```diff
@@ -34,6 +34,8 @@ class RefittingAlgorithm final : public IAlgorithm {
     int pickTrack = -1;
     /// Inflate initial covariance.
     std::array<double, 6> initialVarInflation = {1., 1., 1., 1., 1., 1.};
+    /// Add a beam spot measurement
+    std::optional<Acts::SquareMatrix2> beamSpotConstraint;
```
andiwand, line 38: I think it is pretty neat to have this properly bound to python now 👍 

in the future we could think about also binding bound vectors and so on to replace the current `std::array<double, 6>`

Examples/Algorithms/TrackFitting/src/RefittingAlgorithm.cpp
```diff
@@ -73,10 +115,9 @@ ProcessCode RefittingAlgorithm::execute(const AlgorithmContext& ctx) const {
     }
 
     TrackFitterFunction::GeneralFitterOptions options{
-        ctx.geoContext,
-        ctx.magFieldContext,
-        ctx.calibContext,
-        &track.referenceSurface(),
+        ctx.geoContext, ctx.magFieldContext, ctx.calibContext,
+        perigeeSurface.get(),  //&track.referenceSurface(), // use the general
+                               // perigee instead track reference surface
```
andiwand:
```suggestion
        perigeeSurface.get(),
```
I believe in this case we always want to use the new perigee surface

Examples/Algorithms/TrackFitting/src/RefittingAlgorithm.cpp
```diff
@@ -109,6 +150,13 @@ ProcessCode RefittingAlgorithm::execute(const AlgorithmContext& ctx) const {
       continue;
     }
 
+    if (m_cfg.beamSpotConstraint.has_value()) {
+      beamSpotSL =
+          RefittingCalibrator::RefittingSourceLink{beamSpotConstTrackState};
```
andiwand:
```suggestion
      RefittingCalibrator::RefittingSourceLink beamSpotSL{beamSpotConstTrackState};
```
I think we can just move the declaration here and drop it above?

Examples/Algorithms/TrackFitting/src/RefittingAlgorithm.cpp
```diff
@@ -53,6 +55,46 @@ ProcessCode RefittingAlgorithm::execute(const AlgorithmContext& ctx) const {
   auto trackStateContainer = std::make_shared<Acts::VectorMultiTrajectory>();
   TrackContainer tracks(trackContainer, trackStateContainer);
 
+  auto perigeeSurface = Acts::Surface::makeShared<Acts::PerigeeSurface>(
+      Acts::Vector3{0., 0., 0.});
+
+  const Acts::Vector3 beamSpotCenter{0., 0., 0.};
```
andiwand:
```suggestion
  auto perigeeSurface = Acts::Surface::makeShared<Acts::PerigeeSurface>(
      Acts::Vector3{0., 0., 0.});
```
we can either drop `beamSpotCenter` or use it in the `makeShared` call

Examples/Algorithms/TrackFitting/src/RefittingAlgorithm.cpp
```diff
@@ -53,6 +55,46 @@ ProcessCode RefittingAlgorithm::execute(const AlgorithmContext& ctx) const {
...
+          std::move(*beamSpotVectorTrackStateContainer));
+
+  auto beamSpotConstTrackState =
+      beamSpotConstVectorTrackStateContainer->getTrackState(
+          beamSpotTrackState.index());
+  Acts::SourceLink uncalibSLconst =
+      beamSpotConstTrackState.getUncalibratedSourceLink();
+
+  RefittingCalibrator::RefittingSourceLink beamSpotSL{beamSpotConstTrackState};
+
```
andiwand:
```suggestion
```
I believe this can be moved

Examples/Algorithms/TrackFitting/src/RefittingAlgorithm.cpp
```diff
@@ -53,6 +55,46 @@ ProcessCode RefittingAlgorithm::execute(const AlgorithmContext& ctx) const {
...
+  auto beamSpotConstVectorTrackStateContainer =
+      std::make_shared<Acts::ConstVectorMultiTrajectory>(
+          std::move(*beamSpotVectorTrackStateContainer));
+
+  auto beamSpotConstTrackState =
+      beamSpotConstVectorTrackStateContainer->getTrackState(
+          beamSpotTrackState.index());
+  Acts::SourceLink uncalibSLconst =
+      beamSpotConstTrackState.getUncalibratedSourceLink();
+
```
andiwand:
```suggestion
```
this seems unused?

Examples/Algorithms/TrackFitting/src/RefittingAlgorithm.cpp
```diff
@@ -53,6 +55,46 @@ ProcessCode RefittingAlgorithm::execute(const AlgorithmContext& ctx) const {
...
+    ACTS_DEBUG("No beam spot constraint provided, using zero matrix");
+    beamSpotTrackState.allocateCalibrated(beamSpotMeasValue,
+                                          Acts::SquareMatrix2::Zero());
+  }
+
+  Acts::SourceLink testSL{42};
+  beamSpotTrackState.setUncalibratedSourceLink(std::move(testSL));
+
+  Acts::SourceLink uncalibSL = beamSpotTrackState.getUncalibratedSourceLink();
+
```
andiwand:
```suggestion
```
seems unused?

Examples/Algorithms/TrackFitting/src/RefittingAlgorithm.cpp
```diff
@@ -53,6 +55,46 @@ ProcessCode RefittingAlgorithm::execute(const AlgorithmContext& ctx) const {
   auto trackStateContainer = std::make_shared<Acts::VectorMultiTrajectory>();
   TrackContainer tracks(trackContainer, trackStateContainer);
 
+  auto perigeeSurface = Acts::Surface::makeShared<Acts::PerigeeSurface>(
+      Acts::Vector3{0., 0., 0.});
+
+  const Acts::Vector3 beamSpotCenter{0., 0., 0.};
+
+  auto beamSpotVectorTrackStateContainer =
```
andiwand, line 70:
```suggestion
  // The following code is only necessary if a beamspot constraint is in use but unguarded for lifetime and simplicity reasons.
  // The Core KF does not support a beamspot constraint by itself but it can be achieved by injecting a fake measurement on the perigee surface and processing it first in the fitter. This can lead to problems if the beamspot constraint moves the initial parameters too far and we start to miss following surfaces and measurements. But in a refitting scenario with increased bounds and direct navigation it *should* work OK.

  auto beamSpotVectorTrackStateContainer =
```
maybe we should add a comment here to explain what is done and why

Examples/Algorithms/TrackFitting/src/RefittingAlgorithm.cpp
delitez: i shall drop beamSpotCenter

andiwand (approved):
Great - thank you @delitez !

PR #5345, opened 2026-04-14
fix: pass restrictToSensitives to functor

osbornjd:
This PR adds the boolean in addition to the visitor to the functor arguments so that one can actually check if sensitives should be fixed. The current behavior has a bug in that if `restrictToSensitives=false`, this functor is passed in the `else` statement but all non-sensitive surfaces are skipped in the `sensitive` functor because there is not an explicit check on the boolean. This PR fixes this behavior.


Core/include/Acts/Geometry/TrackingVolume.hpp
andiwand:
```suggestion
          [&visitor](const Surface& surface) { visitor(&surface); },
```

andiwand:
```suggestion
      apply([&visitor](const Surface& surface) {
            if (!surface.isSensitive()) {
              return;
            }
            visitor(&surface);
          }});
```

andiwand:
```suggestion
```

andiwand (commented):
good catch! thanks for fixing that @osbornjd 

I added some suggestions that might make the code easier. I don't think there is a purpose for the `sensitive` lambda if it does not generalize well so I would break this out

osbornjd:
Thanks @andiwand , I believe I implemented your suggestion

Core/include/Acts/Geometry/TrackingVolume.hpp
```diff
@@ -185,18 +185,17 @@ class TrackingVolume : public Volume {
...
-      if (surface.geometryId().sensitive() == 0) {
-        return;
-      }
-      visitor(&surface);
-    };
-
     if (restrictToSensitives) {
-      apply(sensitive);
+      apply([&visitor](const Surface& surface) {
+        if (surface.geometryId().sensitive() == 0) {
```
andiwand, line 190: just to check - `surface.isSensitive()` did not work @osbornjd ?

andiwand (dismissed):
LGTM

Core/include/Acts/Geometry/TrackingVolume.hpp
osbornjd: It apparently failed some of the tests in the CI. [Here](https://github.com/acts-project/acts/actions/runs/24412008509/job/71311072206) is an example - but it suggests that looking up the sensitive ID from the geometry ID and `isSensitive()` do not return exactly the same thing

andiwand: I guess this is not expected @paulgessinger @junggjo9

andiwand:
this is changing some of the hashes here https://github.com/acts-project/acts/actions/runs/24414870520/job/71325208679?pr=5345 which should go here https://github.com/acts-project/acts/blob/main/Python/Examples/tests/root_file_hashes.txt
```
test_material_mapping__material_mapping_map.root: bcc7bcfbfed6e576926c24cc6b5808d5f7c890ef1d700e732d5f75cf7119e93f
test_material_mapping__material_validation.root: 63208746c50177723c5efc5d989686dac48c304fe8810680bf34c6a9639605ad
```
can you update them @osbornjd ? afterwards we can get this in

osbornjd:
> this is changing some of the hashes here https://github.com/acts-project/acts/actions/runs/24414870520/job/71325208679?pr=5345 which should go here https://github.com/acts-project/acts/blob/main/Python/Examples/tests/root_file_hashes.txt
> 
> ```
> test_material_mapping__material_mapping_map.root: bcc7bcfbfed6e576926c24cc6b5808d5f7c890ef1d700e732d5f75cf7119e93f
> test_material_mapping__material_validation.root: 63208746c50177723c5efc5d989686dac48c304fe8810680bf34c6a9639605ad
> ```
> 
> can you update them @osbornjd ? afterwards we can get this in

Sorry I'm not sure I understand... what exactly is changing these hashes? I have only changed the TrackingVolume file

andiwand:
> Sorry I'm not sure I understand... what exactly is changing these hashes? I have only changed the TrackingVolume file

I did not dive in but I suppose the material mapping tests depend on the code you changed and produce different outputs now

osbornjd:
>> Sorry I'm not sure I understand... what exactly is changing these hashes? I have only changed the TrackingVolume file

> I did not dive in but I suppose the material mapping tests depend on the code you changed and produce different outputs now

Does this mean I need to go dig into some test and understand what exactly changed, or just that the hashes themselves need to be updated?

andiwand:
> Does this mean I need to go dig into some test and understand what exactly changed, or just that the hashes themselves need to be updated?

I would propose to update the hashes as is like mentioned above. To me it seems plausible that the fix affects the material mapping as it likely discovers different surfaces now.

osbornjd:
Okay sorry, I think I just misunderstood what I needed to do. I updated the root hash

andiwand (approved):

Core/include/Acts/Geometry/TrackingVolume.hpp
junggjo9: Í remember to have encountered that problem as well. The surfaces don't have placements associated and it happened before the setIsSensitive method was introduced

junggjo9: Should be fixed with https://github.com/acts-project/acts/pull/5353

PR #5348, opened 2026-04-14
feat: Traccc sequence example python bindings

paradajzblond:
Updating Traccc, Detray, VecMem versions.

Exposing the full traccc CUDA track reconstruction chain as a Python
module, mirroring the standalone traccc_seq_example executable.


CMakeLists.txt
```diff
@@ -208,7 +208,10 @@ set_option_if(ACTS_BUILD_PLUGIN_GNN ACTS_BUILD_EXAMPLES_GNN)
 set_option_if(ACTS_BUILD_PLUGIN_FPEMON ACTS_BUILD_EXAMPLES)
 set_option_if(ACTS_BUILD_PLUGIN_JSON ACTS_BUILD_PLUGIN_TRACCC)
 set_option_if(ACTS_BUILD_PLUGIN_ACTSVG ACTS_BUILD_PLUGIN_TRACCC)
-set_option_if(ACTS_GNN_ENABLE_CUDA ACTS_GNN_ENABLE_MODULEMAP)
+if(ACTS_ENABLE_CUDA AND ACTS_BUILD_PLUGIN_GNN)
+    set(ACTS_GNN_ENABLE_MODULEMAP ON)
+endif()
+# set_option_if(ACTS_GNN_ENABLE_CUDA ACTS_GNN_ENABLE_MODULEMAP)
```
paulgessinger: I think this should work with the `set_option_if`, you should be able to give a Boolean combination like this I think.

Examples/Algorithms/Traccc/CMakeLists.txt
```diff
@@ -11,6 +11,11 @@ target_link_libraries(
         Acts::ExamplesFramework
         Acts::ExamplesPropagation
         Acts::PluginDetray
+        Acts::PluginCovfie
+        traccc::cuda
+        traccc::core
+        traccc::io
+        vecmem::cuda
```
paulgessinger: You'd have to gate the two cuda targets to the option if we want to be able to compile the CPU code without cuda.

Python/Examples/CMakeLists.txt
```diff
@@ -168,9 +168,13 @@ add_examples_binding_if(Pythia8 Acts::ExamplesGeneratorsPythia8 ACTS_BUILD_EXAMP
 add_examples_binding_if(Root Acts::ExamplesIoRoot ACTS_BUILD_EXAMPLES_ROOT TRUE)
 add_examples_binding_if(Svg Acts::ExamplesIoSvg ACTS_BUILD_PLUGIN_ACTSVG TRUE)
 add_examples_binding_if(TGeo Acts::ExamplesDetectorTGeo ACTS_BUILD_EXAMPLES_ROOT TRUE)
-add_examples_binding_if(Traccc "Acts::PluginCovfie;Acts::ExamplesTraccc" ACTS_BUILD_PLUGIN_TRACCC FALSE)
 add_examples_binding_if(TruthJet Acts::ExamplesJets ACTS_BUILD_EXAMPLES_FASTJET FALSE)
 
+add_examples_binding_if(Traccc
+  "Acts::PluginCovfie;Acts::ExamplesTraccc;traccc::device_common;traccc::cuda"
+  ACTS_BUILD_PLUGIN_TRACCC FALSE
+)
```
paulgessinger, line 176: Also this still hard requires cuda now, doesn't it?

Python/Examples/src/plugins/Traccc.cpp
```diff
@@ -31,102 +69,340 @@ using namespace ActsExamples;
...
+
+  /// The full GPU sequence example (mirrors traccc_seq_example setup)
+  {
+    py::class_<TracccSeqAlg, std::shared_ptr<TracccSeqAlg>>(traccc,
+                                                            "TracccSeqAlg");
 
-          using DetrayPropagator =
-              DetrayPropagator<DetrayRknStepper, DetrayHostStore,
-                               Covfie::ConstantField>;
+    traccc.def(
```
paulgessinger, line 286: Medium term we'll want to move this into the general ACTS examples fw system, and keep the Python specific files light. For testability this is ok at this point.

Tests/UnitTests/Plugins/Detray/DetrayPayloadConverterTests.cpp
```diff
@@ -742,18 +742,28 @@ BOOST_AUTO_TEST_CASE(DetrayTrackingGeometryConversionTests) {
...
+  // Helper to convert std::map<unsigned int, std::string> to detray::name_map
+  auto toDetrayNameMap = [](const std::map<unsigned int, std::string>& src) {
+    detray::name_map result;
+    for (const auto& [idx, name] : src) {
+      result.emplace(static_cast<detray::dindex>(idx), name);
+    }
+    return result;
+  };
+
+  auto detrayNames = toDetrayNameMap(payloads.names);
```
paulgessinger, line 754: Can we just change the type in `payloads`? The type was picked to be consistent with the detray one so we should just unify now.

thirdparty/vecmem/CMakeLists.txt
```diff
@@ -31,5 +31,7 @@ set(VECMEM_BUILD_BENCHMARKING
     "Turn off the build of VecMem benchmarking"
 )
 
+set(VECMEM_BUILD_CUDA ON CACHE BOOL "Build VecMem CUDA library" FORCE)
```
paulgessinger: This would also have to be conditional.

paulgessinger (commented):
Agreed with overall strategy. There are still some spots which hard requires cuda when we build the traccc plugin. 

Also, what's your plan on the orchestration / algorithm code that now live in the Python binding layer? Can we eventually move this out into the plugin proper?

Python/Examples/src/plugins/Traccc.cpp
paradajzblond: Ok, will add to my to-do

Examples/Algorithms/Traccc/include/ActsExamples/Traccc/TracccChain.hpp
```diff
@@ -0,0 +1,114 @@
...
+  traccc::cuda::seed_parameter_estimation_algorithm tp_cuda;
+  traccc::cuda::combinatorial_kalman_filter_algorithm finding_cuda;
+  traccc::cuda::greedy_ambiguity_resolution_algorithm resolution_cuda;
+  traccc::cuda::kalman_fitting_algorithm fitting_cuda;
+
+  TracccChain(const std::string& detector_file,
+              const std::string& digitization_file,
+              const std::string& conditions_file,
+              const std::string& material_file, const std::string& grid_file,
+              const std::string& bfield_file);
```
paulgessinger, line 96: Since these are the only parameters at the API boundary, we could even use PIMPL to hide most of the traccc includes and symbols.

Tests/UnitTests/Plugins/Detray/DetrayPayloadConverterTests.cpp
paradajzblond: yes, can be done after detray merge since that is now frozen

cmake/ActsExternSources.cmake
```diff
@@ -26,8 +26,9 @@ set(ACTS_DETRAY_SOURCE
 )
 mark_as_advanced(ACTS_DETRAY_SOURCE)
 
+# This is temporary until we tag a nice working release of the Traccc code.
 set(ACTS_TRACCC_SOURCE
-    "URL;https://github.com/acts-project/traccc/archive/refs/tags/v${_acts_traccc_version}.tar.gz;URL_HASH;SHA256=de4c88028f51deb63d75c5bdcde40803832e372219cd59d195a748560dacdeb1"
+    "GIT_REPOSITORY;https://github.com/acts-project/traccc.git;GIT_TAG;origin/main"
```
paulgessinger, line 23: This is dangerous because `main` can update any time. Can you just pin the current commit on `main`?

Examples/Algorithms/Traccc/include/ActsExamples/Traccc/TracccChain.hpp
paulgessinger: Point still stands, why not PIMPL to move the includes to the implementation? This has benefits imo even without other backends.

Examples/Algorithms/Traccc/include/ActsExamples/Traccc/TracccSeqAlg.hpp
```diff
@@ -0,0 +1,59 @@
...
+// Forward-declare to keep traccc cuda headers out of this header
+// You can imagine having a separate header with includes for a different
+// backend (e.g. CPU-only)
+namespace ActsExamples {
+struct TracccChain;
+}
+
+namespace ActsExamples {
+
+class TracccSeqAlgorithm final : public IAlgorithm {
```
paulgessinger:
```suggestion
class TracccCudaSeqAlgorithm final : public IAlgorithm {
```
since it's CUDA only?

Examples/Algorithms/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ add_subdirectory_if(GeneratorsPythia8 ACTS_BUILD_EXAMPLES_PYTHIA8)
 add_subdirectory(MaterialMapping)
 add_subdirectory(Printers)
 add_subdirectory(Propagation)
+add_subdirectory_if(Traccc ACTS_BUILD_PLUGIN_TRACCC AND ACTS_ENABLE_CUDA)
```
paulgessinger: The entire subdirectory is now gated with CUDA. I assume this is intentional?

Examples/Scripts/Python/traccc_seq_example.py
```diff
@@ -0,0 +1,188 @@
...
+import sys
+
+import acts
+import acts.examples
+from acts.examples.traccc import *
+
+u = acts.UnitConstants
+
+
+def make_sequencer(
```
paulgessinger: If this function is not used otherwise, you could just inline the content below where you have the arguments.

thirdparty/traccc/CMakeLists.txt
```diff
@@ -43,29 +39,53 @@ set(TRACCC_SETUP_DFELIBS
 )
 set(TRACCC_SETUP_DETRAY OFF CACHE BOOL "Set up Detray as part of Traccc")
 set(TRACCC_SETUP_ACTS OFF CACHE BOOL "Do not set up ACTS as part of Traccc")
-set(TRACCC_SETUP_TBB OFF CACHE BOOL "Do not set up TBB as part of Traccc")
-set(TRACCC_SETUP_BENCHMARKS
+
+if(ACTS_ENABLE_CUDA)
+    message(STATUS "Enabling CUDA support for traccc")
+    set(TRACCC_BUILD_CUDA ON CACHE BOOL "Build Traccc with CUDA support")
+endif()
```
paulgessinger: We never do anything with traccc if CUDA is off, so could potentially pull the if up to the parent scope?

thirdparty/traccc/CMakeLists.txt
```diff
@@ -43,29 +39,53 @@ set(TRACCC_SETUP_DFELIBS
...
 )
+
+if(TRACCC_BUILD_EXAMPLES)
+    set(TRACC_USE_SYSTEM_TBB
+        ON
+        CACHE BOOL
+        "Use the system TBB for the Traccc examples"
+    )
+    find_package(TBB REQUIRED)
+endif()
```
paulgessinger: This is because otherwise TBB isnt's available yet? In general, we only do `find_package` in the top level `CMakeLists.txt` (there might be a handful exceptions), so could you replicate the conditional there and move the `find_package` call?

CMakeLists.txt
```diff
@@ -572,6 +577,14 @@ if(ACTS_BUILD_PLUGIN_TRACCC)
         endif()
     endif()
 
+    if(ACTS_ENABLE_CUDA)
+        find_package(CUDAToolkit REQUIRED)
+        enable_cuda()
+        message(STATUS "Build Traccc plugin with CUDA")
+    else()
+        message(STATUS "Build Traccc plugin for CPU only")
+    endif()
```
paulgessinger, line 599: Can you refactor this so that we only need one 

```cmake
find_package(CUDAToolkit REQUIRED)
enable_cuda()
```

Python/Examples/CMakeLists.txt
paulgessinger: I also **think** this would work to avoid the helper variable, but you'd have to try:

```suggestion
add_examples_binding_if(Traccc
  "Acts::ExamplesTraccc"
 "ACTS_BUILD_PLUGIN_TRACCC;AND;ACTS_ENABLE_CUDA" FALSE
)
```

thirdparty/traccc/CMakeLists.txt
krasznaa: I would leave this here. With Alpaka/HIP/SYCL in mind. Rather, I would write it as:

```cmake
set(TRACCC_BUILD_CUDA ${ACTS_ENABLE_CUDA} CACHE BOOL "Build Traccc with CUDA support")
```

The traccc CMake code will anyway advertise that it's building CUDA. No need to shout about it here.

thirdparty/traccc/CMakeLists.txt
krasznaa: You know why? 😛  `TRACC_USE_SYSTEM_TBB` -> `TRACCC_USE_SYSTEM_TBB`. As long as you set that flag correctly to `ON`, regardless of `TRACCC_BUILD_EXAMPLES`, the build **should** start working.

thirdparty/vecmem/CMakeLists.txt
```diff
@@ -31,5 +31,7 @@ set(VECMEM_BUILD_BENCHMARKING
     "Turn off the build of VecMem benchmarking"
 )
 
+set_option_if(VECMEM_BUILD_CUDA ACTS_ENABLE_CUDA)
```
krasznaa:
```suggestion
set_option_if(VECMEM_BUILD_CUDA_LIBRARY ACTS_ENABLE_CUDA)
```

CMakeLists.txt
krasznaa: 🤔 The GNN configuration was changed recently. And for traccc I think none of this is needed.

As long as traccc's flags are set up correctly, it will take care of finding CUDAToolkit and enabling the language. We don't need to do anything for that here.

krasznaa (commented):
I see the following order right now:
  - We push in #5450, hopefully without much disagreement;
  - We add either `ACTS_SETUP_TRACCC` or `ACTS_BUILD_PLUGIN_DETRAY` in a PR (to be discussed live hopefully today);
  - We update https://github.com/acts-project/traccc/pull/1327 to make use of one of those new cache variables, and merge it in. Making sure that the HEAD of traccc works well with the almost-latest version of Acts;
  - We push forward with this PR. 🥳 

If done right, I'm hopeful that next week we could even get a new Acts tag with all these included. 😄

CMakeLists.txt
```diff
@@ -81,11 +81,11 @@ option(ACTS_BUILD_PLUGIN_EDM4HEP "Build EDM4hep plugin" OFF)
 option(ACTS_BUILD_PLUGIN_FPEMON "Build FPE monitoring plugin" OFF)
 option(ACTS_BUILD_PLUGIN_GEOMODEL "Build GeoModel plugin" OFF)
 option(ACTS_BUILD_PLUGIN_TRACCC "Build Traccc plugin" OFF)
+option(ACTS_ENABLE_CUDA "Enable CUDA for the gnn/Traccc plugins" OFF)
 option(ACTS_BUILD_PLUGIN_GEANT4 "Build Geant4 plugin" OFF)
 option(ACTS_BUILD_PLUGIN_GNN "Build the GNN plugin" OFF)
 option(ACTS_GNN_ENABLE_ONNX "Build the Onnx backend for the gnn plugin" OFF)
 option(ACTS_GNN_ENABLE_TORCH "Build the torchscript backend for the gnn plugin" ON)
-option(ACTS_GNN_ENABLE_CUDA "Enable CUDA for the gnn plugin" OFF)
```
benjaminhuth, line 88: Hmm I guess we do not formally garantuee stability of the CMake, but maybe we could deprecate the option rather then remove it directly. (Like I did for the `ACTS_BUILD_EXAMPLES_PYTHON_BINDINGS`, see a few lines below)

CMakeLists.txt
```diff
@@ -572,6 +577,14 @@ if(ACTS_BUILD_PLUGIN_TRACCC)
         endif()
     endif()
 
+    if(ACTS_ENABLE_CUDA)
+        find_package(CUDAToolkit REQUIRED)
+        enable_cuda()
+        message(STATUS "Build Traccc plugin with CUDA")
```
benjaminhuth: The message should be more general I think

Examples/Algorithms/Traccc/include/ActsExamples/Traccc/TracccSeqAlg.hpp
```diff
@@ -0,0 +1,59 @@
...
+      const Config& cfg, std::unique_ptr<const Acts::Logger> logger = nullptr);
+
+  ProcessCode execute(const AlgorithmContext& ctx) const override;
+  ProcessCode finalize() override;
+
+  const Config& config() const { return m_cfg; }
+
+ private:
+  Config m_cfg;
+  std::shared_ptr<TracccChain> m_chain;
```
benjaminhuth: Hmm this could be a `unique_ptr` as well, right?

Examples/Algorithms/Detray/include/ActsExamples/Detray/DetrayPropagator.hpp
benjaminhuth, line 1: maybe remove this from the diff?

docs/old/getting_started.md
```diff
@@ -284,11 +284,11 @@ components.
 | ACTS_BUILD_PLUGIN_FPEMON            | Build FPE monitoring plugin<br> type: `bool`, default: `OFF`                                                                                                                                                                       |
 | ACTS_BUILD_PLUGIN_GEOMODEL          | Build GeoModel plugin<br> type: `bool`, default: `OFF`                                                                                                                                                                             |
 | ACTS_BUILD_PLUGIN_TRACCC            | Build Traccc plugin<br> type: `bool`, default: `OFF`                                                                                                                                                                               |
+| ACTS_ENABLE_CUDA                    | Enable CUDA for the gnn/Traccc plugins<br> type: `bool`, default: `OFF`                                                                                                                                                            |
```
krasznaa, line 287: I wonder if `ENABLE` is the right word here. 🤔 The behaviour here is that when we turn this on, CUDA **has to** be available and working. Otherwise the configuration/build will fail.

In traccc and vecmem we use the `BUILD` "keyword".
  - https://github.com/acts-project/traccc/blob/main/CMakeLists.txt#L64
  - https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-options.cmake#L30

Maybe we should call this `ACTS_BUILD_CUDA` or `ACTS_BUILD_CUDA_LIBRARIES`? 🤔

Examples/Algorithms/Traccc/include/ActsExamples/Traccc/ActsMeasToTracccAlg.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include <traccc/edm/spacepoint_collection.hpp>
+#include <vecmem/memory/host_memory_resource.hpp>
+
+#include "traccc/geometry/detector.hpp"
+#include "traccc/geometry/host_detector.hpp"
+#include "traccc/io/read_detector.hpp"
+
+namespace ActsExamples {
+
+class ActsMeasToTracccAlg final : public IAlgorithm {
```
krasznaa, line 31: I didn't check what other example algorithms do (😦), but some (Doxygen) documentation would be in order here, right?

Examples/Algorithms/Traccc/include/ActsExamples/Traccc/ActsMeasToTracccAlg.hpp
```diff
@@ -0,0 +1,71 @@
...
+    std::string detectorFile = "";
+    std::string inputActsMeasurements = "measurements";
+    std::string outputDetrayToActsMap = "detray-to-acts-map";
+    std::string outputTracccMeasurements = "acts-to-traccc-measurements";
+    std::shared_ptr<const Acts::TrackingGeometry> trackingGeometry;
+  };
+
+  mutable vecmem::host_memory_resource m_mr;
+  vecmem::host_memory_resource mr;
+  traccc::host_detector host_det;
```
krasznaa: Naming!

```suggestion
  vecmem::host_memory_resource m_mr;
  traccc::host_detector m_host_det;
```

Would've also helped with noticing the duplicate `mr` and `m_mr` variables. 😛

Examples/Algorithms/Traccc/src/ActsMeasToTracccAlg.cpp
```diff
@@ -0,0 +1,218 @@
...
+  m_inputActsMeasurements.initialize(m_cfg.inputActsMeasurements);
+  m_outputDetrayToActsMap.initialize(m_cfg.outputDetrayToActsMap);
+  m_outputTracccMeasurements.initialize(m_cfg.outputTracccMeasurements);
+
+  if (m_detrayToActsMap.empty()) {
+    std::cout << "ActsMeasToTracccAlg: detrayToActsMap is empty" << std::endl;
+    buildSurfaceMap(*m_cfg.trackingGeometry, m_cfg.detectorFile,
+                    m_detrayToActsMap, m_actsToDetrayMap);
+    std::cout << "ActsMeasToTracccAlg: built surface map with "
+              << m_detrayToActsMap.size() << " entries" << std::endl;
```
krasznaa: Please use logger messages for all outputs.

krasznaa (commented):
Just a couple of things to start off...

docs/old/getting_started.md
paradajzblond: propose to discuss and resolve this in another MR

PR #5349, opened 2026-04-15
chore: RootMaterialTrackIo - Don't restrict input tree to be a TChain and add eventId getter

junggjo9:
- Add event ID getter
- TTree for reading can be any kind of TTree
--- END COMMIT MESSAGE ---


Plugins/Root/include/ActsPlugins/Root/RootMaterialTrackIo.hpp
```diff
@@ -44,25 +43,25 @@ class RootMaterialTrackIo {
...
+  /// Destructor
   ~RootMaterialTrackIo() = default;
 
-  /// @brief sets the branch connection for reading from a file
+  /// sets the branch connection for reading from a file
   ///
-  /// @param materialChain the TChain to read the material track from
-  void connectForRead(TChain& materialChain);
+  /// @param materialChain the TTree to read the material track from
+  void connectForRead(TTree& materialChain);
```
paulgessinger, line 57: But what if we want to read a chain?

junggjo9: But a TChain is a TTree

image

paulgessinger: Ah, news to me.

paulgessinger (approved):

PR #5352, opened 2026-04-15
feat: Implement `RiddersStepper`

andiwand:
Implements a ridders stepper on top of our existing stepper infrastructure, by using multiple single steppers, similar to multi stepper, but to extrapolate nudged components. First, I tried to reuse the multi stepper for this but it turns out to be more complicated because of component removal and boundary tolerance handling.

This is similar to the existing `RiddersPropagator` but more modular and flexible. The stepper concept encapsulates the covariance transport therefore it is better suited to perform numerical derivatives on this side. The stepper is then provided to the propagator which does not need to know any details about the covariance transport while also supporting the usual actor infrastructure. Ultimately this means that one can for example run the KF or CKF using the `RiddersStepper` as validation against the analytical implementation.

--- END COMMIT MESSAGE ---

blocked by
- https://github.com/acts-project/acts/pull/5385
- https://github.com/acts-project/acts/pull/5386
- https://github.com/acts-project/acts/pull/5387
- https://github.com/acts-project/acts/pull/5388


Core/include/Acts/Propagator/RiddersStepper.hpp
```diff
@@ -0,0 +1,687 @@
...
+}  // namespace Acts
+
+namespace Acts::Experimental {
+
+/// Collection of bound parameter variations, each entry is a pair of the index
+/// of the bound parameter
+using BoundParameterVariation = std::vector<std::pair<std::size_t, double>>;
+
+/// Base class for generating bound parameter variations
+struct BoundParameterVariationGenerator {
```
benjaminhuth, line 40: maybe this clarifies once I understand the full PR, but wouldn't a delegate/function serve the same purpose here?

Core/include/Acts/Propagator/RiddersStepper.hpp
```diff
@@ -0,0 +1,687 @@
...
+      state.secondaryStepperStates.push_back(
+          m_stepperImpl.makeState(state.primaryStepperState.options));
+      m_stepperImpl.initialize(state.secondaryStepperStates.back(),
+                               nudgedParams, std::nullopt, particleHypothesis,
+                               surface);
+    }
+
+    state.covTransport = true;
+    state.cov = *covariance;
+    // TODO otherwise `update` gets screwed
```
benjaminhuth: hmm not sure I understand, is it uninitialized memory otherwise?

Core/include/Acts/Propagator/RiddersStepper.hpp
```diff
@@ -0,0 +1,687 @@
...
+    state.lastStepConstraintType = stype;
+
+    IntersectionStatus overallStatus = m_stepperImpl.updateSurfaceStatus(
+        state.primaryStepperState, surface, index, navDir, boundaryTolerance,
+        surfaceTolerance, stype, logger);
+    if (overallStatus == IntersectionStatus::unreachable) {
+      return overallStatus;
+    }
+
+    const auto combine = [](const IntersectionStatus a,
```
benjaminhuth, line 407: is this the same logic as in the multistepper? not sure if it makes sense to extract this to some multistepper tools?

benjaminhuth (commented):
Looks good, some comments

Core/include/Acts/Propagator/RiddersStepper.hpp
andiwand: could be done with a delegate as well yes. initially this interface had more functions but then it collapsed to one. the default implementations would still require a struct because of the configuration so I guess a delegate won't really buy us anything?

Core/include/Acts/Propagator/RiddersStepper.hpp
andiwand: right that comment is quite cryptic... since I use `initialize` in the `update` functions and we rely on the jacobian not being reset there I had to comment the line below. not great but at the same time it seems like a problem of our stepping interface

Core/include/Acts/Propagator/RiddersStepper.hpp
andiwand: I guess it is quite similar. we could think about having a first level citizen function to do that

andiwand: I checked now and the multi stepper does something little bit different so I don't think it makes sense to generalize this right now

Core/include/Acts/Propagator/RiddersStepper.hpp
benjaminhuth: I still would try to remove the dead code and improve clarity of the comment (similar above)

Core/include/Acts/Propagator/RiddersStepper.hpp
```diff
@@ -0,0 +1,687 @@
...
+      m_stepperImpl.releaseStepSize(secondaryState, stype);
+    }
+  }
+
+  /// Get a string representation of the step size constraints
+  /// @param state the state of the RiddersStepper
+  /// @return a string representation of the step size constraints
+  std::string outputStepSize(const State& state) const {
+    std::stringstream ss;
+    ss << state.primaryStepperState.stepSize.toString();
```
benjaminhuth, line 483: maybe a visual separation between primary and secondaries would be nice

Core/include/Acts/Propagator/RiddersStepper.hpp
benjaminhuth: I wonder if one potentially if you would forsee this to be steered from python, then it would be benefitial. Otherwise, I agree.

benjaminhuth (dismissed):
LGTM, only some small comments

benjaminhuth (approved):
LGTM

PR #5359, opened 2026-04-16
fix: adapt Mat_map_surface_plot_dist and Mat_map_surface_plot_dist to the new surface type and range lookup

veprbl:
This is a minor sequel to
https://github.com/acts-project/acts/pull/4931
and
https://github.com/acts-project/acts/pull/5046

The goal is to fill some missing code in the two remaining macros.

--- END COMMIT MESSAGE ---


veprbl:
This was tested at ePIC https://github.com/eic/epic/pull/1076#issuecomment-4262311862

andiwand (approved):
fine by me

I fear these scripts are bound to break as they are not monitored or maintained. if they are useful to you I am happy to update them. looking a bit further out we should either test them, migrate them to python, or completely drop them and you can pull them into your repo, in case this is even possible due to internal dependencies

veprbl:
@andiwand I agree, the support model is less than perfect, but I'd like them to be corrected best effort at least. These scripts are in an awkward spot because they support surface material maps only. If we build a testing infrastructure on the Acts' side now, it will add some more work to do during the transition to 3rd gen and python?

andiwand:
I think material mapping validation will be the same between gen1 and gen3? @paulgessinger @noemina know more about this

so in case you have some spare cycles we could either have some testing around the current scripts or port them to python if it is possible without making it terribly slow and monitor that. for example by scanning an eta-phi wedge of ODD and having some reference plots we compare to

PR #5370, opened 2026-04-21
feat: add support for multi-pass tracking

jburzy:
# Summary

This MR introduces `MeasurementSubset` — a lightweight view over a `MeasurementContainer` — and wires it through the reconstruction chain to enable multi-pass tracking without measurement remapping.

## `MeasurementSubset`

`MeasurementSubset` (in `ActsExamples/EventData/Measurement.hpp`) derives from `Acts::detail::ContainerSubset`, following the pattern of `SpacePointContainer2::Subset`. 

On top of `ContainerSubset`, `MeasurementSubset` adds `orderedIndices()` — a geometry-sorted `GeometryIdMultiset<IndexSourceLink>` — as an additional member. This is required by `SpacePointMaker` and the CKF and has no analogue in `ContainerSubset`.

All indices are always in original-container space, so e.g. `MeasurementParticlesMap` from digitization can be used for truth matching across all passes without remapping.

## Digitization

`DigitizationAlgorithm` now writes two objects:
- `"measurements"` — the full `MeasurementContainer` (unchanged),
- `"measurement_subset"` — a `MeasurementSubset` containing all measurement indices

## Algorithm updates

`SpacePointMaker`, `TrackFindingAlgorithm`, and `ProtoTracksToTracks` now accept `MeasurementSubset` as input instead of `MeasurementContainer`. Their internal logic is unchanged — `orderedIndices()` and `getMeasurement()` have identical signatures on both types. `TrackFindingAlgorithm` passes `measurements.container()` to `MeasurementCalibratorAdapter` and sizes the shared-hit arrays to `measurements.container().size()` (original N) so hit indices remain valid.

## `MeasurementFilterAlgorithm`

`MeasurementFilterAlgorithm` it reads a `MeasurementSubset` and a `ConstTrackContainer`, collects the indices of measurements used by the tracks, and writes a new `MeasurementSubset` containing only the remaining (unused) indices. No container copying occurs; the output subset points to the same underlying `MeasurementContainer`.

---

# How filtering works

`MeasurementSubset` carries two index structures:

```
subset()          std::vector<Index>
                  e.g. [0, 1, 2, 4, 7, ...]   <- original-container indices
                  (stored in ContainerSubset base)

orderedIndices()  GeometryIdMultiset<IndexSourceLink>
                  geometry-sorted multiset, each entry stores:
                    .geometryId()  <- which surface
                    .index()       <- same original-container index
```

Both SpacePointMaker and the CKF navigate measurements exclusively through `orderedIndices()`.

---

# Example usage (Python)

The canonical use case is two-pass tracking: a standard prompt pass followed by a large-radius tracking (LRT) pass on the measurements not consumed by pass 1.

```python

# --- Filter: remove measurements used by pass 1 ---
s.addAlgorithm(
    acts.examples.MeasurementFilterAlgorithm(
        level=acts.logging.INFO,
        inputTracks="tracks",                        # pass-1 output
        inputMeasurementSubset="measurement_subset", # full initial subset
        outputMeasurementSubset="lrt_measurement_subset",
        includeOutliers=False,                       # set True to also remove outliers
    )
)

# --- Pass 2: LRT on the remaining measurements ---
# addSeeding and addCKFTracks will read "lrt_measurement_subset"
# via the prefix="lrt_" convention.

addSeeding(
    s, trackingGeometry, field,
    prefix="lrt_",
    seedFinderConfigArg=SeedFinderConfigArg(
        impactMax=300.0 * u.mm,
        collisionRegion=(-350.0 * u.mm, 350.0 * u.mm),
        r=(200.0 * u.mm, 1000.0 * u.mm),
    ),
    ...
)
addCKFTracks(
    s, trackingGeometry, field,
    TrackSelectorConfig(loc0=(-300.0 * u.mm, 300.0 * u.mm), ...),
    prefix="lrt_",
    ...
)
addAmbiguityResolution(s, AmbiguityResolutionConfig(...), prefix="lrt_", ...)
```

The `prefix` parameter causes `addSeeding` and `addCKFTracks` to look for `"lrt_measurement_subset"` automatically, so no further plumbing is needed. A complete working example is provided in `Examples/Scripts/Python/full_chain_odd_LRT_second_pass.py`.


Examples/Framework/include/ActsExamples/EventData/MeasurementMap.hpp
paulgessinger, line 1: Consider putting these into the existing measurement header in the examples. Also potentially name this `UsedMeasurementMap` or something to indicate the usage.

Examples/Algorithms/TrackFinding/src/MeasurementMapAlgorithm.cpp
```diff
@@ -0,0 +1,89 @@
...
+  if (m_inputIndexRemapping.isInitialized()) {
+    remapping = &m_inputIndexRemapping(ctx);
+  }
+
+  std::size_t nUsed = 0;
+  std::size_t nAlreadyPresent = 0;
+
+  for (auto track : tracks) {
+    for (auto state : track.trackStatesReversed()) {
+      if (!state.typeFlags().isMeasurement()) {
```
paulgessinger: Question: should this consider outliers (outliers become *used*) or should they still be available? This will currently does not mark outliers as used (which might be what we want).

Python/Examples/src/TrackFinding.cpp
```diff
@@ -34,6 +36,16 @@ using namespace ActsExamples;
...
+  ACTS_PYTHON_DECLARE_ALGORITHM(MeasurementMapAlgorithm, mex,
+                                "MeasurementMapAlgorithm", inputTracks,
+                                inputMeasurementMap, inputIndexRemapping,
+                                outputMeasurementMap);
+
+  ACTS_PYTHON_DECLARE_ALGORITHM(MeasurementFilterAlgorithm, mex,
+                                "MeasurementFilterAlgorithm", inputMeasurements,
+                                inputMeasurementMap, outputMeasurements,
+                                outputIndexRemapping);
+
```
paulgessinger: These are not used anywhere right now. Consider either modifying one of the existing python scripts to use them, or (even better) add a short python level test that schedules this and runs some basic check.

Examples/Algorithms/TrackFinding/src/MeasurementMapAlgorithm.cpp
jburzy: this is consistent with the [Athena implementation](https://gitlab.cern.ch/atlas/athena/-/blob/main/Tracking/Acts/ActsPrdAssociation/src/PrdAssociationAlg.cxx#L82), so I think it is what we want

pbutti: One can add this as configuration option

jburzy: done :)

Python/Examples/src/TrackFinding.cpp
jburzy: was going to follow up with that in a separate MR, but might as well push something more complete. this required some changes in reconstruction.py to support renaming the 2nd pass output containers in the whiteboard. hopefully this is a reasonable strategy.

andiwand (commented):
Thanks!

A few high level comments:

- I wondered if there is another use case for `inputMeasurementMap` you had in mind
  - if there is no immediate use case we could fold these two algorithms together and simply have one that does `measurements + tracks -> filtered measurements`
  - in case the original measurement index is important we could also attach it to the measurement itself, similar to how it is done for space points. that way we don't need another whiteboard entry and they stick consistently together
- one danger is that the user might mix measurement and space point container which leads to inconsistencies... I don't see an immediate solution to that. without having a general purpose thinning mechanism like Athena
- there are truth matching entries we write in the simulation which would have to be written here as well and consistently used to make performance plots work https://github.com/acts-project/acts/blob/e4285fc0f9d05f78f1c5cd147279882f1a6d4bbe/Examples/Algorithms/Digitization/include/ActsExamples/Digitization/DigitizationAlgorithm.hpp#L47-L54

now all this said it seems to start becoming quite problematic to have a separate measurement collection that is not consistent with the first one... I had the same issue when I started to refactor the simulation EDM and algorithms to get a consistent particle subset after applying selection cuts. my solution there was to redefine what a `ParticleContainer` is on the whiteboard by using a [`ContainerSubset`](https://github.com/acts-project/acts/blob/main/Core/include/Acts/Utilities/detail/ContainerSubset.hpp). a bit sill that the initial container would need to be an array of all indices, but then it generalizes well to post selection and the indices and mappings stay consistent

jburzy:
Thanks for the review @andiwand :) I assumed that maybe this was oversimplified. Here are some thoughts on the various points you raised:

> * I wondered if there is another use case for `inputMeasurementMap` you had in mind
>   
>   * if there is no immediate use case we could fold these two algorithms together and simply have one that does `measurements + tracks -> filtered measurements`
>   * in case the original measurement index is important we could also attach it to the measurement itself, similar to how it is done for space points. that way we don't need another whiteboard entry and they stick consistently together

I think in theory these two algorithms could be combined, I was attempting to keep it conceptually the same as in athena where after each pass we update the PRD map, and then use it during data preparation to skip used measurements. I don't feel super strongly about that though

> * one danger is that the user might mix measurement and space point container which leads to inconsistencies... I don't see an immediate solution to that. without having a general purpose thinning mechanism like Athena

I haven't thought too deeply about this to know it this is really an issue. For now I'm just passing the updated measurement container to the spacepoint formation. what exactly is the concern? 

> there are truth matching entries we write in the simulation which would have to be written here as well and consistently used to make performance plots work

I also don't see why this is needed if it's the same underlying measurements being used. Or is the truth matching based on the indices? In any case, my initial tests show reasonable performance plots :) 

efficiency_vs_prodR

andiwand:
nice plot! 🔥 

> I think in theory these two algorithms could be combined, I was attempting to keep it conceptually the same as in athena where after each pass we update the PRD map, and then use it during data preparation to skip used measurements. I don't feel super strongly about that though

if we do not end up using them individually I would argue that it could simply be one single entity

> I haven't thought too deeply about this to know it this is really an issue. For now I'm just passing the updated measurement container to the spacepoint formation. what exactly is the concern?

that sounds good yes - I believe this is how it should be in any case

my concern is that we have multiple indices flying around at this point so one could use the LRT space points and lookup measurements in the wrong pre-LRT measurement container. if you are lucky this crashes, if you are unlucky you get garbage-in garbage-out

> I also don't see why this is needed if it's the same underlying measurements being used. Or is the truth matching based on the indices? In any case, my initial tests show reasonable performance plots :)

the truth matching does `track state -> measurement -> particle(s)`
https://github.com/acts-project/acts/blob/7bc78ecd82ff4469dd6f688392539928735a111a/Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TrackTruthMatcher.hpp#L28-L37
so this relies on an updated `measurement -> particles` map that is consistent with the measurement subset

jburzy:
Ok I think I'm understanding. So you would suggest:

1. Combine the two algorithms into one -- easy 
2. Implement the [ContainerSubset](https://github.com/acts-project/acts/blob/main/Core/include/Acts/Utilities/detail/ContainerSubset.hpp) for the filtered measurements 
3. Re-calculate the ` measurement -> particles` using the `ContainerSubset`

?

andiwand:
1. 👍 
2. 👍 I imagined there could be an "all measurements" entry and a "selected measurements" entry which by default includes all measurements. the subset can hold a `std::vector<MeasurementIndex>` for this purpose
3. if we go with the container subset for selected measurements we could use the original indices of "all measurements" for consistent matching so I believe in this case we do not need to recreate any association maps but can just pass in and point to the "all measurements" container. does that make sense?

in case 3 gets a bit too convoluted with the examples framework changes we can also push this to a followup. happy to assist in any case

Examples/Framework/include/ActsExamples/Framework/DataHandle.hpp
```diff
@@ -191,6 +191,13 @@ class WriteDataHandle final : public WriteDataHandleBase {
     add(wb, std::move(value));
   }
 
+  /// Read back the value just written via this handle within the same
+  /// execute() call. Calls the protected DataHandleBase::get<T>() rather
+  /// than WhiteBoard::get<T>() directly because WhiteBoard::get is private.
+  const T& readBack(const AlgorithmContext& ctx) const {
+    return this->template get<T>(ctx.eventStore);
+  }
+
```
jburzy: @andiwand I had to add this method to get around `WhiteBoard::get<T>()` being private and `WriteDataHandle` not having access, but maybe you have a cleaner solution I missed.

jburzy:
I've taken a stab at implementing your suggestions @andiwand :) it seems to work and give the same results I had before. happy to receive more feedback.

Examples/Framework/include/ActsExamples/Framework/DataHandle.hpp
andiwand: ok makes sense - seems fair to me. maybe @paulgessinger has an opinion / alternative?

jburzy: Following @paulgessinger suggestion, I've updated the signature of `WhiteBoard::add` to return a pointer to the written object, and adjusted the `DataHandle` accordingly. This removes the nead for `readBack`, so I've removed it.

Examples/Algorithms/TrackFinding/src/TrackFindingAlgorithm.cpp
```diff
@@ -722,15 +722,15 @@ ProcessCode TrackFindingAlgorithm::finalize() {
...
   // Compute shared hits from all the reconstructed tracks
   // Compute nSharedhits and Update ckf results
   // hit index -> list of multi traj indexes [traj, meas]
 
   std::vector<std::size_t> firstTrackOnTheHit(
-      measurements.size(), std::numeric_limits<std::size_t>::max());
+      measurements.container().size(), std::numeric_limits<std::size_t>::max());
   std::vector<std::size_t> firstStateOnTheHit(
-      measurements.size(), std::numeric_limits<std::size_t>::max());
+      measurements.container().size(), std::numeric_limits<std::size_t>::max());
```
andiwand, line 734: 👍 in principle these should be fine to use the subset but lets not deal with the index remapping right now

Examples/Framework/include/ActsExamples/EventData/Measurement.hpp
```diff
@@ -531,4 +533,56 @@ static_assert(
...
+      MeasurementContainer,
+      MeasurementContainer::ConstVariableProxy,
+      std::vector<MeasurementContainer::Index>,
+      true>;
+
+  MeasurementSubset(const MeasurementContainer& container,
+                    std::vector<MeasurementContainer::Index> validIndices,
+                    MeasurementContainer::OrderedIndices orderedIndices)
+      : Base(container, std::move(validIndices)),
+        m_orderedIndices(std::move(orderedIndices)) {}
```
andiwand: ah I forgot about the ordered indices

an alternative to pushing it through would be to generate it during construction so the user does not need to handle it and we avoid redundant information that might be inconsistent

andiwand (commented):
thanks @jburzy ! looks pretty solid to me

left one very minor comment and a suggestion for `MeasurementSubset` construction

andiwand:
LGTM 👍 

there seems to be at least one more case where we need to switch the measurement input looking at the CI log
```
-> white board will contain key 'measurements'
at this point in the sequence (source: DigitizationAlgorithm.OutputMeasurements),
but the type will be
'MeasurementContainer'
and not
'MeasurementSubset'
```

jburzy:
Hi @andiwand, looks like the issues with the CI have been fixed. I think this should be ready now :)

andiwand (approved):

PR #5372, opened 2026-04-21
feat: Add thirdparty Millepede-II

goblirsc:
Add thirdparty Millepede-II installer and plumbing for a future Millepede-II plugin.

--- END COMMIT MESSAGE ---

Add [Millepede-II](https://gitlab.desy.de/millepede/millepede-ii), containing the solver component, as a `thirdparty` package.
This allows to directly call `pede` in the provided Millepede alignment example script to run the fit.
The example now compares the corrections fitted by the ACTS-internal and the Millepede-II solvers in a plot. 

With this, the second stage of the 3-step alignment workflow,
1. dumping of alignment inputs (already in place)
2. running of the alignment fit
3. reading back the corrections and associating them to the surfaces 

is available in the examples.


andiwand (dismissed):
LGTM 👍

paulgessinger:
- I think the mille plugin is currently off in CI, can we turn that on? Same for the MP2 plugin
- Should these be separate plugins or a single one? Having too many knobs can be a bit confusing, and I'm wondering how likely it is in practice to build one and not the other
- In CI, can we make sure that it takes both mille and MP2 from Spack and doesn't build the "vendored" dependencies?

goblirsc:
Will add Millepede to the existing Mille plugin as there are basically no use cases where one could be set up but the other not. Some of the changes in this PR have meanwhile become part of #5432. Will therefore close this iteration and prepare a new (#5433), minimal one, rather than manually reconciling histories to avoid overlaps between the two PR.

PR #5374, opened 2026-04-22
feat: Add Detray to ACTS

niermann999:
Moving the detray repository into ACTS. The detray code is modified to satisfy clang-tidy and sonar analyses. The detray CI is moved over from the detray repository such that it is only triggered if files in the new `Detray` folder have been changed. For the detray gitlab CI, this was not possible, so it will be triggered for every PR to ACTS until the CI Bridge can be configured accordingly. For such changes, a new 'Component - Detray' label has been added. Due to issues with the unused files check, all of the flagged detray files have been added to the exclusion list.

Co-authored-by: beomki-yeo <63090140+beomki-yeo@users.noreply.github.com>
Co-authored-by: Andreas Salzburger <Andreas.Salzburger@cern.ch>
Co-authored-by: Stephen Nicholas Swatman <stephen.nicholas.swatman@cern.ch>
Co-authored-by: Attila Krasznahorkay <30694331+krasznaa@users.noreply.github.com>
Co-authored-by: Paul Gessinger <hello@paulgessinger.com>
Co-authored-by: Guilherme <g.almeida@cern.ch>
Co-authored-by: Fred <92879080+fredevb@users.noreply.github.com>
Co-authored-by: Vakho Tsulaia <qiarik@gmail.com>
Co-authored-by: Yasmine Badr Elhouda LARBAOUI <121395132+Larb-YasmineBadrElhouda@users.noreply.github.com>
Co-authored-by: Anshumaan Singh <anshumaan567@gmail.com>
Co-authored-by: Stewart Martin-Haugh <smartinhaugh@gmail.com>
Co-authored-by: paradajzblond <72857406+paradajzblond@users.noreply.github.com>
Co-authored-by: Abhirikshma Nandi <abhirikshma14@gmail.com>
Co-authored-by: Ryan C <ryan@ryancross.xyz>
Co-authored-by: flg <fabrice@legoff.eu>
Co-authored-by: Tirthankar Mazumder <63574588+wermos@users.noreply.github.com>


CI/check_unused_files.py
```diff
@@ -37,6 +37,10 @@ def main():
         ".github",
         ".idea",
         ".devcontainer",
+        # Used by traccc
+        "Detray/detectors",
+        # CLI tools for users
+        "Detray/tests/tools",
```
AJPfleger: Just checked the issue. The exclusion looks for the directory-name not the path-name. I would suggest to remove these lines and to put after:
```
        # Skip "DD4hep-tests" since their cmake looks a bit different
        # TODO Find a more elegant way
        if str(root).find("Tests/UnitTests/Plugins/DD4hep") != -1:
            processed_files += count_files(root, exclude_dirs, exclude_files)
            continue
```
this block:
```
        # Skip "Detray-detectors". They are used by external traccc
        # TODO Find a more elegant way
        if str(root).find("Detray/detectors") != -1:
            processed_files += count_files(root, exclude_dirs, exclude_files)
            continue

        # Skip "Detray-tests-tools". CLI tools for users
        # TODO Find a more elegant way
        if str(root).find("Detray/tests/tools") != -1:
            processed_files += count_files(root, exclude_dirs, exclude_files)
            continue
```
for now. I have a more elegant way in mind, but lets get this mega-PR in first.

Detray/core/include/detray/geometry/surface.hpp
```diff
@@ -74,6 +74,7 @@ class surface {
   /// Conversion to surface interface around constant detector type
   template <typename detector_type = detector_t>
     requires(!std::is_const_v<detector_type>)
+  // NOLINTNEXTLINE
```
paulgessinger, line 80: If at all possible, specific suppression would be prefferable like:

```cpp
// NOLINTNEXTLINE(google-explicit-constructor, google-runtime-int)
```

niermann999: Fixed

paulgessinger:
Couple observations from the sonar run:

- The log macros could be updated to require `;`
- There's a number of `using namespace`s in headers that I think we should avoid
- A number of templated constructors that don't guard against infinite self-recursion
- Remove implicit conversion operators (I think `surface.hpp` has one, maybe there are more)
- A number of `const_cast`s that we'll need to remove
- Spaceship operator for intersections
- A few forwarding references that are never forwarded

I think these can be follow up PRs.

niermann999:
> Couple observations from the sonar run:
> 
>     * The log macros could be updated to require `;`
> 
>     * There's a number of `using namespace`s in headers that I think we should avoid
> 
>     * A number of templated constructors that don't guard against infinite self-recursion
> 
>     * Remove implicit conversion operators (I think `surface.hpp` has one, maybe there are more)
> 
>     * A number of `const_cast`s that we'll need to remove
> 
>     * Spaceship operator for intersections
> 
>     * A few forwarding references that are never forwarded
> 
> 
> I think these can be follow up PRs.

I went through the sonar issues and the current status is:

* Absolutely, I did not know a good solution here
* There are a few namespaces that are pulled into the detray namespace which depend on the compute backend or algebra plugin, e.g. pulling `using namespace std` vs `using namespace sycl` into `detray::math` in order to switch out the math functions such that they can be used in a sycl kernel
* I think the ranges constructors should be capable to wrap ranges of the same type?
* The implicit conversion operator for the surface is currently used in cases where the surface is converted from the non-const to the const detector type  (which should not be functionally different). It is not nice for sure, but should be safe? I am not sure how to solve this better atm
* The const-casts in the grids are a long standing issue that I have not yet been able to resolve
* The spaceship operator cannot be used for the SoA backends
* If I remember correctly, some forwarding references might get moved when properly forwarded everywhere, which at some point results in undefined behaviour, because they are used multiple times in the constructor. Maybe we should move away form universal references here?

paulgessinger:
@niermann999 

- `do {} while(0)` is the common pattern for the macros
- why not alias the whole namespace? `namespace math = sycl`? `using namespace xxx;` in headers is generally an anti-pattern I think.
- why would you wrap an adaptor in itself? maybe I'm not seeing the use-case. not guarding has the downside of allowing recursive instantiation.
- explicit conversion over implicit conversion. the call site becomes more verbose but much clearer, in my experience.

the rest i'm less sure about.

niermann999:
> * do {} while(0)

I agree, I implemented most of your suggestions now! I am not so sure about the namespace one though, because we will end up with `detray::math == std`, at which point it is not benign to add stuff to it anymore?

For the ranges constructors, it seems that the STL uses the `all` view that needs to be wrapped around if an adaptor wants to adapt a range of the same type. We don't have that atm, so we might need to add that in case we need e.g. the subrange of a subrange.

The surface conversion I am afraid might break in traccc, where the `const detector_t` is used in many places, so let's hold that until a later PR (only need to add the explicit keyword to the conversion operator, I think). In the meantime, I made sure that the class always uses the detector as const internally.

paulgessinger:
I'm fighting Spack to get an updated dependencies package with google test included. Hope to have that converge tomorrow or so, and then we can proceed here.

niermann999:
> I'm fighting Spack to get an updated dependencies package with google test included. Hope to have that converge tomorrow or so, and then we can proceed here.

Do we really depend on googletest for this PR? For now the detray runners just keep building it themselves, as before

paulgessinger:
@niermann999 Right but I would like to avoid that if we can

Detray/core/include/detray/algebra/utils/approximately_equal.hpp
```diff
@@ -0,0 +1,193 @@
...
+  static_assert(std::same_as<detray::traits::value_t<vector1_t>,
+                             detray::traits::value_t<vector2_t>>);
+
+  using index_t = detray::traits::index_t<vector1_t>;
+
+  constexpr index_t size{detray::traits::size<vector1_t>};
+
+  bool ret{true};
+  for (index_t i = 0; i < size; ++i) {
+    ret &= ::detray::algebra::approx_equal(v1[i], v2[i], rel_error, max_error);
```
andiwand: looking at the sonarcloud output - I guess this is intentionally a bit operation over a logical boolean operation to avoid branching?

Detray/core/include/detray/builders/detail/bin_association.hpp
```diff
@@ -0,0 +1,237 @@
...
+
+          auto vertices_per_masks =
+              surface_masks
+                  .template visit<detail::vertexer<point2_t, point3_t>>(
+                      sf.mask());
+
+          // Usually one mask per surface, but design allows - a
+          // single association  is sufficient though
+          for (auto &vertices : vertices_per_masks) {
+            if (!vertices.empty()) {
```
andiwand: looking at the sonarcloud output - the nesting could be reduced with an early `continue`
```suggestion
            if (vertices.empty()) {
              continue
            }
            // ...
```

andiwand (dismissed):
👍 just had a quick browse

@paulgessinger @asalzburger should we get this in? I guess we don't want to review this like a usual PR so from my side it is fine if @niermann999 is happy and you are

Detray/CMakeLists.txt
```diff
@@ -0,0 +1,502 @@
...
+    DETRAY_SETUP_NLOHMANN
+    "Set up the nlohmann::json target(s) explicitly"
+    TRUE
+)
+option(
+    DETRAY_USE_SYSTEM_NLOHMANN
+    "Pick up an existing installation of nlohman::json from the build environment"
+    ${DETRAY_USE_SYSTEM_LIBS}
+)
+if(DETRAY_SETUP_NLOHMANN)
```
benjaminhuth, line 265: I guess this was discussed already, and I missed it, but I wondered going through the PR: Do we want to keep the `extern` part and support building `detray` top-level with its own thirdparty fetching mechanism?
Please feel free to resolve if this was already discussed!

paulgessinger: The idea was that for step zero this would remain possible, yes.

benjaminhuth: Okay seems reasonable. But in the long term I think it would make sense if we only would build detray as top-level project with system libs...

niermann999:
I addressed some more of the suggestions, but I agree that we should probably move ahead with a first workable integration soon and smooth (no pun intended) things out later. So, unless there is a blocker in this for somebody, should we move ahead and then start to also accustom traccc to the new way to pull detray?

andiwand (dismissed):

asalzburger:
Let's do it!

asalzburger (approved):
Approved!

asalzburger:
That makes me very happy. Thanks everyone to make this happy!!

PR #5375, opened 2026-04-22
docs: Add `CODEOWNERS` file

stephenswat:
Following the discussion on governance at this year's developers' workshop, I thought we could open a discussion on _optional_, _opt-in_ ownership of code in ACTS. This commit serves as a first version of that, claiming ownership of some of the code I wrote in this repository.


paulgessinger:
Can we add an additional comment to the file what **we** mean by _owners_ here, just to not discourage anyone from changing these files?

stephenswat:
I have added a draft of an explanation of what we mean by code ownership. I encourage us to use this PR as a place to debate and discuss what we want that to mean (and if we want something like this in the first place).

stephenswat:
Tagging @asalzburger for this discussion.

AJPfleger (dismissed):
We just all had a discussion in the dev-meeting.
- In general, the idea of contact-persons seems good. Auto-suggestions and blame-histories can already work but are not helpful if maintainers left some time ago.
- The name `owner` can imply undesired things like copyright claims. With the current added comment, this could be avoided.

We want to try it out for, if we don't like it we can revert it.

AJPfleger (dismissed):

AJPfleger (approved):

PR #5379, opened 2026-04-23
perf(python): Implement buffered access in uproot based readers

diptaparna:
Implemented buffered access in uproot based readers

--- END COMMIT MESSAGE ---

**Description:** This MR introduces a buffered reading mechanism to the UprootParticleReader and UprootSimHitReader classes to prevent loading entire ROOT trees into memory at initialization.

**Motivation:** Previously, both uproot readers loaded all tree arrays into memory upfront using tree.arrays(library="np"). While this guaranteed thread-safe concurrent access, it led to excessive and often prohibitive memory consumption when processing large simulated datasets. By loading the tree in smaller, configurable chunks, we significantly reduce the memory footprint while maintaining efficient I/O performance.

**Changes:**

- Added a `bufferSize` argument (defaulting to 10) to the constructors of both `UprootParticleReader` and `UprootSimHitReader`.
- Removed the upfront `tree.arrays()` execution in `__init__`, opting instead to only initialize the tree references and event maps.
- Updated the `read()` method in UprootParticleReader to dynamically fetch chunks of `bufferSize` entries whenever the requested event index falls outside the currently cached buffer (`[buffer_start, buffer_end)`).
- Updated the `read()` method in `UprootSimHitReader` to dynamically load rows based on the requested event's start and end indices. If the required block exceeds the default `bufferSize`, the buffer dynamically expands to accommodate the full event.
- Adapted array indexing inside the event population loops to correctly offset against the new `buffer_start`.

**Note:** Generated using Gemini Code Assist in VSCode using the prompt, "Instead of loading the whole tree in memory, implement a buffered way of accessing it. The number of entries to load at a time should be an argument in the constructor, with default value 10." The default `bufferSize` is later changed to 1 as per suggestion in the PR.


Python/Examples/python/uproot.py
```diff
@@ -27,34 +27,38 @@ class UprootParticleReader(acts.examples.IReader):
...
-    concurrent access from the sequencer.
+    vector-valued branches). Data is buffered for access.
     """
 
     def __init__(
         self,
         filePath: Union[Path, str],
         outputParticles: str = "particles",
         level: acts.logging.Level = acts.logging.INFO,
+        bufferSize: int = 10,
```
benjaminhuth: Can we make this "optional" in a sense, that by default we load the whole file? One issue I see here is that potentially, if we run in a multithreaded sequencer, the events are not sequentially drawn from the file, but in an unpredictable order. This might be not the standard use case in `pyacts`, but in this case, I think this becomes very slow with the default buffer size.

benjaminhuth: I guess `bufferSize` could be `None` by default, and then the buffer-start and buffer-end to 0 and n_events or so.

diptaparna: Why would you load the entire array by default? You are not taking advantage of any kind of vectorized operation. The processing happens one event at a time.

I can indeed increase the buffer size, but with (default) geant4 simulation of ttbar event with 200 pileup, just 10 events takes ~ 1 GB of memory. So, I think 10 is a reasonable buffer size.

benjaminhuth: Hmm or maybe 1 as default buffer size? Not sure how `uproot` performs in general

So one problem I think is, if you access the events in the order 100, 1, 99, 2, 98, 3 etc., we would always read 10 events and throwing away 9. This could happen in multithreaded sequencer.

benjaminhuth: I just would try to have a default that is behaving reasonable in most situation, but you can always optimize if you know what you are doing

diptaparna: Yes, 1 as default bufferSize is kinda similar to the default ROOt behaviour. Sounds fine.

benjaminhuth (approved):
LGTM

PR #5380, opened 2026-04-23
feat: Serializer for multilayer navigation

benjaminhuth:


Plugins/Json/src/TrackingGeometryJsonConverter.cpp
```diff
@@ -325,6 +330,25 @@ nlohmann::json encodeSurfaceArrayNavigationPolicy(
   return jPolicy;
 }
 
+std::unique_ptr<Acts::INavigationPolicy> decodeMultiLayerNavigationPolicy(
+    const nlohmann::json& encoded, const Acts::GeometryContext& gctx,
+    const Acts::TrackingGeometryJsonConverter& /*converter*/,
+    const Acts::TrackingVolume& volume, const Acts::Logger& logger) {
+  return std::make_unique<Acts::Experimental::MultiLayerNavigationPolicy>(
```
paulgessinger: Could we at this point just inline the `toJson`/`fromJson` functions into these functions here? We probably won't have a strong need for the disconnected signatures going forward. What do you think?

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
```diff
@@ -74,12 +74,25 @@ class MultiLayerNavigationPolicy : public INavigationPolicy {
...
+
  private:
   // The tracking volume
   const TrackingVolume& m_volume;
 
   // The grid that holds the indexed surfaces
   IndexedUpdatorType m_indexedGrid;
+
+  // The bin expansion used when filling the grid
+  std::vector<std::size_t> m_binExpansion;
```
paulgessinger: Why not store the whole config, and provide a `config()` accessor?

Plugins/Json/include/ActsPlugins/Json/MultiLayerNavigationJsonConverter.hpp
```diff
@@ -0,0 +1,57 @@
...
+#include <memory>
+
+#include <nlohmann/json.hpp>
+
+namespace Acts {
+
+class TrackingGeometryJsonConverter;
+class TrackingVolume;
+
+namespace MultiLayerNavigationJsonConverter {
```
paulgessinger: Could this just go into `TrackingGeometryJsonConverter.cpp` or do we need it as public separate API? The tests could go through `TrackingGeometryJsonConverter::navigationPolicyToJson` and `navigationPolicyFromJson`, with the added benefit that this also tests the registration and dispatch.

Plugins/Json/src/MultiLayerNavigationJsonConverter.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "Acts/Utilities/Grid.hpp"
+#include "ActsPlugins/Json/GridJsonConverter.hpp"
+#include "ActsPlugins/Json/UtilitiesJsonConverter.hpp"
+
+nlohmann::json
+Acts::MultiLayerNavigationJsonConverter::encodeMultiLayerNavigationPolicy(
+    const Experimental::MultiLayerNavigationPolicy& policy,
+    const TrackingGeometryJsonConverter& /*converter*/) {
+  nlohmann::json jPolicy;
+  jPolicy["kind"] = "MultiLayerNavigation";
```
paulgessinger: Can use `getNavigationPolicyKind` if you move it to `TrackingGeometryJsonConverter.cpp`

paulgessinger (changes requested):

Core/include/Acts/Navigation/MultiLayerNavigationPolicy.hpp
dimitra97: I agree with this. apart from that I dont really have anything else to request for change

dimitra97 (approved):

paulgessinger (approved):

PR #5383, opened 2026-04-23
feat: stubgeneration for python bindings

benjaminhuth:


CMakePresets.json
```diff
@@ -195,12 +195,13 @@
...
                 "ACTS_BUILD_PYTHON_WHEEL": "ON",
                 "ACTS_BUILD_PYTHON_BINDINGS": "ON",
+                "ACTS_GENERATE_PYTHON_STUBS": "ON",
                 "ACTS_BUILD_EXAMPLES": "ON",
                 "ACTS_BUILD_PLUGIN_ROOT": "OFF",
                 "ACTS_BUILD_PLUGIN_FPEMON": "ON",
                 "ACTS_BUILD_EXAMPLES_ROOT": "OFF",
                 "ACTS_USE_SYSTEM_NLOHMANN_JSON": "ON",
-                "ACTS_USE_SYSTEM_PYBIND11": "ON"
+                "ACTS_USE_SYSTEM_PYBIND11": "OFF"
```
paulgessinger: Is this necessary?

benjaminhuth: artifact from my local build.

CMakeLists.txt
```diff
@@ -211,6 +212,13 @@ set_option_if(ACTS_BUILD_PLUGIN_ACTSVG ACTS_BUILD_PLUGIN_TRACCC)
 set_option_if(ACTS_GNN_ENABLE_CUDA ACTS_GNN_ENABLE_MODULEMAP)
 set_option_if(ACTS_BUILD_PYTHON_BINDINGS ACTS_BUILD_PYTHON_WHEEL)
 set_option_if(ACTS_BUILD_ALIGNMENT ACTS_BUILD_PLUGIN_MILLE)
+
+# Find uv (used by codegen and python stub generation)
+if(NOT ACTS_USE_SYSTEM_LIBS OR ACTS_GENERATE_PYTHON_STUBS)
+    message(STATUS "Find uv")
+    find_program(uv_exe uv)
+endif()
```
paulgessinger: This is a bit awkward, since the codegen discovery path also auto-downloads uv. Should we push this into a helper `.cmake` file with an `acts_ensure_uv()` function that we can call right before we use it?

That could then do something like 

```cmake
if(NOT TARGET uv::uv)
  # auto-download code currently in `ActsCodegen.cmake` goes here.
  add_executable(uv::uv IMPORTED GLOBAL)
  set_target_properties(uv::uv PROPERTIES
      IMPORTED_LOCATION "${uv_exe}"
  )
endif()

benjaminhuth: done

paulgessinger (approved):

PR #5384, opened 2026-04-24
refactor: Logging and binding for geometry serialization

benjaminhuth:


paulgessinger (dismissed):

Plugins/Json/include/ActsPlugins/Json/TrackingGeometryJsonConverter.hpp
```diff
@@ -147,7 +147,9 @@ class TrackingGeometryJsonConverter {
   ///
   /// @param config is the conversion dispatch configuration
   explicit TrackingGeometryJsonConverter(
-      Config config = Config::defaultConfig());
+      Config config = Config::defaultConfig(),
+      std::unique_ptr<const Acts::Logger> logger = Acts::getDefaultLogger(
+          "TrackingGeometryJsonConverter", Acts::Logging::INFO));
```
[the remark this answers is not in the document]
benjaminhuth: @copilot apply changes based on [this feedback](https://github.com/acts-project/acts/pull/5384#discussion_r3143943242)

Plugins/Json/src/TrackingGeometryJsonConverter.cpp
```diff
@@ -953,9 +967,17 @@ Acts::TrackingGeometryJsonConverter::trackingVolumeFromJson(
...
       encoded.at(kRootVolumeIdKey).get<std::size_t>();
+
+  ACTS_DEBUG("Decoding " << volumeRecords.size() << " volumes, "
+                         << portalRecords.size() << " portals, "
+                         << surfaceRecords.size()
+                         << " surfaces from root volume '"
+                         << volumeRecords.at(rootVolumeId).name << "'");
+
   if (!volumeRecords.contains(rootVolumeId)) {
     throw std::invalid_argument("Serialized root volume ID does not exist");
```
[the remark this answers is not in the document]
benjaminhuth: @copilot apply changes based on [this feedback](https://github.com/acts-project/acts/pull/5384#discussion_r3143943256)

Python/Plugins/src/Json.cpp
```diff
@@ -72,4 +75,30 @@ PYBIND11_MODULE(ActsPluginsPythonBindingsJson, json) {
...
+  {
+    py::class_<TrackingGeometryJsonConverter>(json,
+                                              "TrackingGeometryJsonConverter")
+        .def(py::init([](Acts::Logging::Level level) {
+               return std::make_unique<TrackingGeometryJsonConverter>(
+                   TrackingGeometryJsonConverter::Config::defaultConfig(),
+                   Acts::getDefaultLogger("TrackingGeometryJsonConverter",
+                                          level));
+             }),
+             py::arg("level"))
```
paulgessinger: This should probably expose the config?

benjaminhuth: Hmm not sure, if we bind the config, we also need to bind all this dispatch stuff to be useful, which is not straight forward. I would keep it contained for now, we can expand it later

paulgessinger: Seems to me like it's the common pattern and the obvious future extension point. Also, can you either switch this one or add an overload that takes a logger instance rather than the level?

paulgessinger (dismissed):

paulgessinger (approved):

PR #5387, opened 2026-04-24
test: Rework `PropagatorTest` actor

andiwand:
Rewrite the actor in `PropagatorTest` to mimic the `SurfaceReached` aborter behavior

--- END COMMIT MESSAGE ---

extracted from
- https://github.com/acts-project/acts/pull/5352


Tests/UnitTests/Core/Propagator/PropagatorTests.cpp
```diff
@@ -96,34 +102,49 @@ struct SurfaceObserver {
...
-    state.stepping.stepSize.release(ConstrainedStep::Type::Actor);
-    state.stepping.stepSize.update(distance * state.options.direction,
-                                   ConstrainedStep::Type::Actor);
-
-    // return true if you fall below tolerance
-    if (std::abs(distance) <= tolerance) {
-      ++result.surfaces_passed;
-      result.surface_passed_r = perp(stepper.position(state.stepping));
-      state.stepping.stepSize.release(ConstrainedStep::Type::Actor);
+    // calculate the intersections to the surface
```
AJPfleger:
```suggestion
    // calculate the intersections with the surface
```

Tests/UnitTests/Core/Propagator/PropagatorTests.cpp
```diff
@@ -96,34 +102,49 @@ struct SurfaceObserver {
...
+         Acts::enumerate(multiIntersection)) {
+      if (intersection.isValid() &&
+          detail::checkPathLength(intersection.pathLength(), nearLimit,
+                                  farLimit, logger)) {
+        // Adjust the step size so that we cannot cross the target surface
+        state.stepping.stepSize.release(ConstrainedStep::Type::Actor);
+        state.stepping.stepSize.update(intersection.pathLength(),
+                                       ConstrainedStep::Type::Actor);
+
+        // return true if you fall below tolerance
```
AJPfleger:
```suggestion
        // return true if we fall below tolerance
```

Tests/UnitTests/Core/Propagator/PropagatorTests.cpp
```diff
@@ -96,34 +102,49 @@ struct SurfaceObserver {
...
+    // calculate the intersections to the surface
+    const auto multiIntersection = surface->intersect(
+        state.geoContext, stepper.position(state.stepping),
+        state.options.direction * stepper.direction(state.stepping),
+        BoundaryTolerance::None());
+
+    const double farLimit = std::numeric_limits<double>::max();
+
+    for (auto [intersectionIndex, intersection] :
+         Acts::enumerate(multiIntersection)) {
```
AJPfleger:
```suggestion
    for (auto intersection : multiIntersection) {
```

Tests/UnitTests/Core/Propagator/PropagatorTests.cpp
```diff
@@ -96,34 +102,49 @@ struct SurfaceObserver {
...
+      if (intersection.isValid() &&
+          detail::checkPathLength(intersection.pathLength(), nearLimit,
+                                  farLimit, logger)) {
```
AJPfleger:
```suggestion
      if (!intersection.isValid() ||
          !detail::checkPathLength(intersection.pathLength(), nearLimit,
                                  farLimit, logger)) {
        continue;
      }
```

I was thinking about an early continue and move the rest 1 layer left.

andiwand: good point. I think the down side is that the main body then has a `break;` which might look a bit odd. I now changed it to `std::ranges::find_if` to avoid this. what do you think?

andiwand:
thanks for the review @AJPfleger ! I applied your suggestions and put up another alternative for the `for` loop. I also back synced the changes to the surface reached aborter

AJPfleger (approved):

Tests/UnitTests/Core/Propagator/PropagatorTests.cpp
AJPfleger: implicit loops are even nicer :3

PR #5389, opened 2026-04-24
feat: Add a cartesian space point grid

goblirsc:
Add Cartesian space point grid 

--- END COMMIT MESSAGE ---

Add a cartesian space point grid, leaning heavily on the cylindrical version for inspiration.
Intended for use in cosmic seeding (y-sorted space points), potentially also interesting for telescopes.
The orientation and ordering of the internal sorting is configurable.


Core/include/Acts/Seeding2/CartesianSpacePointGrid.hpp
```diff
@@ -0,0 +1,205 @@
...
+  /// The grid is a 3D grid with the axes representing azimuthal angle (phi),
+  /// z-coordinate, and radial distance (r).
+  using GridType = Grid<BinType, XAxisType, YAxisType, ZAxisType>;
+  /// Type alias for binned group over the cartesian grid
+  using BinnedGroupType = BinnedGroup<GridType>;
+
+  /// Configuration parameters for the cartesian space point grid.
+  struct Config {
+    /// minimum pT
+    float minPt = 0 * UnitConstants::MeV;
```
andiwand: I believe this is unused. should we just drop it?

Core/include/Acts/Seeding2/CartesianSpacePointGrid.hpp
```diff
@@ -0,0 +1,205 @@
...
+    float zMin = -2700 * UnitConstants::mm;
+    /// maximum z coordinate ofspace points used in the seeding
+    float zMax = 2700 * UnitConstants::mm;
+
+    /// enable non equidistant binning in x
+    std::vector<float> xBinEdges{};
+    /// enable non equidistant binning in y
+    std::vector<float> yBinEdges{};
+    /// enable non equidistant binning in z
+    std::vector<float> zBinEdges{};
```
andiwand: also this one seems to be unused. should we just remove it?

Core/src/Seeding2/CartesianSpacePointGrid.cpp
```diff
@@ -0,0 +1,113 @@
...
+  ACTS_VERBOSE("- x Axis  : " << xAxis);
+  ACTS_VERBOSE("- y Axis  : " << yAxis);
+  ACTS_VERBOSE("- z Axis  : " << zAxis);
+
+  GridType grid(
+      std::make_tuple(std::move(xAxis), std::move(yAxis), std::move(zAxis)));
+  m_binnedGroup.emplace(std::move(grid), m_cfg.bottomBinFinder.value(),
+                        m_cfg.topBinFinder.value(), m_cfg.navigation);
+  m_grid = &m_binnedGroup->grid();
+  if (m_cfg.sortingCoord != Acts::eTime) {
```
andiwand: should we throw is the user supplies `eTime`?

andiwand (commented):
Looks good overall, thank you @goblirsc ! I left a few minor comments

Core/include/Acts/Seeding2/CartesianSpacePointGrid.hpp
goblirsc: indeed, as for the cartesian version we do not need a curvature estimate to calculate a phi-bin size on the fly.

Core/include/Acts/Seeding2/CartesianSpacePointGrid.hpp
goblirsc: yep! Can add back in the future easily enough should a non-uniform grid become desirable.

Core/src/Seeding2/CartesianSpacePointGrid.cpp
goblirsc: yep! Should now be happening

goblirsc (commented):
The SonarCloud comments do not seem to be trivial to implement - a range-based loop leads to compile errors with an eldritch `const`-violation, which I was not able to understand as all iterators and accessors are non-`const` 😕 . But there is a pile of nested typedefs, maybe somewhere a `const` is sneaking in. Manually invoking iterators, weirdly, does not run into the same issue.

andiwand (approved):
👍 Thank you @goblirsc !

PR #5390, opened 2026-04-24
feat: Final set of changes for GBTSv3

jpreston-cern:
This is the final PR migrating GBTSv3 to ACTS,

it includes an additional flag that allows for triplets to be passed within the eta barrel to recover efficiency, with the consequence being a slow down in speed due to the extra seeds produced 

--- END COMMIT MESSAGE ---

@timadye @andiwand


Core/include/Acts/Seeding2/GraphBasedTrackSeeder.hpp
```diff
@@ -51,16 +51,22 @@ class GraphBasedTrackSeeder {
     bool matchBeforeCreate = false;
     /// Use legacy tuning parameters.
     bool useOldTunings = false;
-    /// description here
+    /// optional validation for abrrel triplets
     bool validateTriplets = true;
-    /// description here
+    /// widens allowed variation in tau ratio
+    // if layer is missed in edge connecting
```
andiwand:
```suggestion
    /// if layer is missed in edge connecting
```

Core/include/Acts/Seeding2/GraphBasedTrackSeeder.hpp
```diff
@@ -51,16 +51,22 @@ class GraphBasedTrackSeeder {
...
     bool useAdaptiveCuts = true;
+    /// optionally add 3 sp seeds within a cirtain eta range
+    bool addTriplets = false;
     /// Tau ratio cut threshold.
     float tauRatioCut = 0.007;
     /// Tau ratio precut threshold.
     float tauRatioPrecut = 0.009f;
-    /// description here
+    /// correction applied to tau acceptance
+    // if a layer is missed during edge connecting
```
andiwand:
```suggestion
    /// if a layer is missed during edge connecting
```

Core/src/Seeding2/GraphBasedTrackSeeder.cpp
```diff
@@ -665,54 +665,68 @@ void GraphBasedTrackSeeder::extractSeedsFromTheGraph(
...
     GbtsEdge* pS = &(edgeStorage.at(edgeIndex));
 
-    if (pS->level < static_cast<std::int32_t>(minLevel)) {
-      continue;
+    if (m_cfg.lrtMode || !m_cfg.addTriplets) {
+      if (pS->level < minLevel) {
+        continue;
+      }
+    } else {  // eta-dependent cut
+      float edgeEta = std::abs(-std::log(pS->p[0]));
```
andiwand:
```suggestion
      const float edgeAbsEta = std::abs(-std::log(pS->p[0]));
```

Core/src/Seeding2/GraphBasedTrackSeeder.cpp
```diff
@@ -665,54 +665,68 @@ void GraphBasedTrackSeeder::extractSeedsFromTheGraph(
...
+
+      if (edgeEta > m_cfg.maxEtaAddTriplets) {
```
andiwand:
```suggestion
      if (edgeAbsEta > m_cfg.maxAbsEtaAddTriplets) {
```

Core/src/Seeding2/GraphBasedTrackSeeder.cpp
```diff
@@ -723,11 +737,25 @@ void GraphBasedTrackSeeder::extractSeedsFromTheGraph(
       continue;
     }
 
-    if (minLevel > static_cast<std::uint32_t>(rs.vs.size())) {
-      continue;
-    }
+    float seedEta = std::abs(-std::log(pS->p[0]));
```
andiwand:
```suggestion
    const float seedAbsEta = std::abs(-std::log(pS->p[0]));
```

Core/src/Seeding2/GraphBasedTrackSeeder.cpp
```diff
@@ -723,11 +737,25 @@ void GraphBasedTrackSeeder::extractSeedsFromTheGraph(
...
 
-    const float seedEta = std::abs(-std::log(pS->p[0]));
+    std::uint32_t chainLength = static_cast<std::uint32_t>(rs.vs.size());
```
andiwand:
```suggestion
    const std::uint32_t chainLength = static_cast<std::uint32_t>(rs.vs.size());
```

Core/src/Seeding2/GraphBasedTrackSeeder.cpp
```diff
@@ -723,11 +737,25 @@ void GraphBasedTrackSeeder::extractSeedsFromTheGraph(
...
+
+    if (m_cfg.lrtMode || !m_cfg.addTriplets) {
+      if (chainLength < minLevel) {
+        continue;
+      }
+    } else {
+      if (seedEta > m_cfg.maxEtaAddTriplets) {
```
andiwand:
```suggestion
      if (seedAbsEta > m_cfg.maxAbsEtaAddTriplets) {
```

Core/include/Acts/Seeding2/GraphBasedTrackSeeder.hpp
```diff
@@ -51,16 +51,22 @@ class GraphBasedTrackSeeder {
...
     float tauRatioCorr = 0.006;
+    /// description here
```
andiwand: can we add a description here?
```suggestion
    /// Maximum absolute eta value to ...
```
not sure what the cut actually does 😅

Core/include/Acts/Seeding2/GraphBasedTrackSeeder.hpp
```diff
@@ -51,16 +51,22 @@ class GraphBasedTrackSeeder {
...
+    float maxEtaAddTriplets = 1.5;
     /// Eta bin width override (0 uses default from connection file).
     // specify non-zero to override eta bin width from connection file (default
     // 0.2 in createLinkingScheme.py)
```
andiwand:
```suggestion
/// specify non-zero to override eta bin width from connection file (default
/// 0.2 in createLinkingScheme.py)
```

andiwand (commented):
thanks @jpreston-cern ! couple of minor comments, otherwise good to go from my side

Core/include/Acts/Seeding2/GraphBasedTrackSeeder.hpp
jpreston-cern: in general, GBTS currently only provides a seed with a minimum of 4 space points (apart from when we do seed splitting where we can potentially get 2 3sp seeds from splitting a 4sp original seed)

When `addTriplets` is on, we allow 3 spacepoint seeds to be passed through as well, the `maxAbsEtaAddTripelts` veriable then defined the range in which this takes place (currently the default is in the barrel as that is where the issues are at the moment)

andiwand (approved):

PR #5395, opened 2026-04-27
perf: Better optimize cotTheta filter for strip seeding

mvessell96:
Replaces the conditional check per-seed-candidate on approximate strip doublet cotTheta with a sorted one with a break condition as is done in the Pixel seeding 

--- END COMMIT MESSAGE ---

A previous recent PR (https://github.com/acts-project/acts/pull/5339) added an optional filter for strip triplet seeding on the approximate cotTheta before the strip coordinate transformation. This saved execution time in a vacuum, but in practice the branchy condition disrupted the optimizer at compile-time and ultimately did not result in too much measurable CPU improvement in a real-world sense. We can instead implement a version similar to that used already in the pixel seeding with `sortedByCotTheta`, just with the approximate top/bottom cotTheta computed from the sp centers, which saves much more CPU when given an appropriate threshold (testing for ATLAS LRT suggests between 0.3-0.4 is optimal) by ending iteration once the threshold is reached. 

FYI @jburzy


andiwand (approved):
great stuff! thanks @mvessell96 🐑

PR #5396, opened 2026-04-29
chore: Bump cuda architectures

benjaminhuth:
--- END COMMIT MESSAGE ---

Would this make sense as a painless default? Or should we rather remove that at all and rely on cmake autodiscovery? @stephenswat @paradajzblond


paradajzblond:
I think this is fine, I can test, but I'll let Stephen confirm. From my own test I saw we require architectures >= 70 anyway.

stephenswat (dismissed):
Makes sense to me; this supports the A100, AX000, and H100 which are common at CERN. You might want to add 75 because the T4 is still pretty common on lxplus-gpu.

stephenswat (approved):

PR #5400, opened 2026-04-30
feat: Reference trajectory builder + Kalman-based fitter

andiwand:
Implementation of a reference trajectory builder tool that propagates through the detector with provided start parameters and records predicted parameters and jacobians in the track EDM.

On top of that one can attach source links (raw, uncalibrated measurements) and calibrate them. Finally, `filter` can be used to run an updater (usually the Kalman filter in form of the `GainMatrixUpdater`) to forward filter the measurements. This is decoupled form the propagation and uses the existing linearization stored on the track.

An example implementation is provided based on the existing Kalman filter. The difference of the two approaches is where the linearization is formed. While the existing Kalman filter linearizes depending on previous filter steps, this alternative implementation runs a single linearization for the full trajectory based on the start parameters.

--- END COMMIT MESSAGE ---

cc @stephenswat


Core/include/Acts/TrackFitting/ReferenceTrajectoryBuilder.hpp
```diff
@@ -0,0 +1,617 @@
...
+
+    /// Whether to consider energy loss.
+    bool energyLoss = true;
+
+    /// Whether to include non-linear correction during global to local
+    /// transformation
+    FreeToBoundCorrection freeToBoundCorrection;
+
+    /// Input MultiTrajectory
+    std::shared_ptr<traj_t> outputStates;
```
benjaminhuth: This seems not to be referenced anywhere?

Examples/Algorithms/TrackFitting/src/KalmanReferenceTrajectoryFitterFunction.cpp
```diff
@@ -0,0 +1,270 @@
...
+    auto track = *buildResult;
+
+    referenceTrajectoryBuilder.attachSourceLinks(track, sourceLinks,
+                                                 sourceLinkAccessorDelegate);
+
+    referenceTrajectoryBuilder.calibrateMeasurements(options.geoContext,
+                                                     options.calibrationContext,
+                                                     track, calibratorDelegate);
+
+    auto fitterResult = referenceTrajectoryBuilder.filter(
```
benjaminhuth, line 133: The chain of build trajectory -> attach -> calibrate -> filter seems to me like one unit one would chain together in a Core-function. 
My naive assumption of an API would have been more like the KF, having one `.fit` interface... Are there reasons to not do that?

benjaminhuth (commented):
Very nice! only a few comments

Examples/Algorithms/TrackFitting/src/KalmanReferenceTrajectoryFitterFunction.cpp
andiwand: in principle we can have both. I think it is good to have the individual functions available since users might want the flexibility. right now the options and the type do not suggest that this is a KF but one can of course use that way as exercised in the examples. so we can just add an additional high level function like for the KF. what do you think?

benjaminhuth (approved):
Looks good to me!

PR #5407, opened 2026-05-04
perf: Speed up vertex conversion in EDM4hep sim input

paulgessinger:
Speed up the EDM4hep input conversion significantly (~3x or so) by only doing the vertex position search after looking up if we have ID collisions. This was implicitly already done before, but in a costly linear search. This should give identical outputs at significantly reduced runtime cost.


andiwand (dismissed):

paulgessinger:
@andiwand fixed clang-format

andiwand (approved):

PR #5409, opened 2026-05-05
feat: prefix for seeds and seeding performance writers for multi-pass tracking

altsybee:
Adding string prefix for seeds and seeding performance writers in `Python/Examples/python/reconstruction.py`, to have separate output files with seeding info for each iteration of the multi-pass tracking.


andiwand (approved):
fine by me

note that this will likely break due to name changes and it does not generalize to downstream algorithms right now

PR #5410, opened 2026-05-05
feat: Initial Arrow conversion and parquet output

paulgessinger:
Adds an Apache Arrow / Parquet output path for the examples framework, so event data (particles, sim hits, tracks) can be written to a columnar, sharded Parquet dataset and read back per-event — from both C++ and Python/pyarrow.

### What

- **New `ActsPluginArrow` plugin** wrapping Apache Arrow. Arrow is linked
    with hidden visibility so none of its symbols leak across the `.so`
    boundary (enforced by exported-symbol lists). Two visibility-exported,
    pybind-friendly handles — `ArrowSchemaHandle` and `ArrowTable` — move
    schemas and tables across library and language boundaries via the Arrow
    **C Data Interface**, giving ABI-safe pyarrow interop without exposing
    arrow's own typeinfo.

- **Parquet I/O (`Examples/Io/Parquet`)**:
    - `ParquetWriter`: sharded dataset, one directory per collection. Events
      are routed to shard files by `event_id`, so each shard owns a disjoint
      event-id range and footer min/max statistics stay tight — letting the
      reader prune to a single fragment per lookup. Row-group buffering bounds
      peak memory; per-collection schemas are validated on write.
    - `ParquetReader`: dataset reader with per-event lookup via filter
      pushdown + footer-statistics pruning, and added-column schema evolution
      against an optional target schema.

- **Output converters (`Examples/Io/Arrow`)** for particles, sim hits and
    tracks. Each builds a per-event **nested** table: one row per event, every
    field a `list<T>` whose single element holds that event's values, with
    `event_id` as the outer routing key.

- **Python bindings** for the converters, reader and writer, with the arrow
    schema bridged into Python so the producing and consuming sides share one
    schema handle.

- Wired into `full_chain_odd.py` (new `--output-parquet` flag); covered by
    `test_arrow.py` and an ABI-isolation test (`test_arrow_isolation.py`).

### Why

ACTS examples lacked a standard columnar output. Parquet gives compact, typed, schema-stable files that downstream ML/analysis tooling consumes directly, and the C-Data-interface design lets the same buffers be used from pyarrow with no copies and no arrow-symbol clashes.

Blocked by:
- #5439 
- #5440


Examples/Framework/include/ActsExamples/EventData/CaloHit.hpp
benjaminhuth, line 1: Do we need to introduce CaloHit EDM in this PR?

Examples/Framework/include/ActsExamples/EventData/CaloHit.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+/// Energy contribution to a calorimeter cell from a single source particle.
+/// Already aggregated across all step-level depositions, so @c energy is the
+/// total energy and @c time is the energy-weighted average.
+struct CaloHitContribution {
+  /// Row index of the source particle in the simulated
+  /// @c SimParticleContainer; @c std::numeric_limits<std::uint64_t>::max() if
+  /// the source particle was filtered out upstream.
+  std::uint64_t particleRow;
+  /// Sum of step-level energies in the same units as the source EDM4hep
```
benjaminhuth: I think we should know that this is GeV, right? The reader is responsible to provide our native unit system.

Examples/Framework/include/ActsExamples/EventData/CaloHit.hpp
```diff
@@ -0,0 +1,53 @@
...
+namespace ActsExamples {
+
+/// Energy contribution to a calorimeter cell from a single source particle.
+/// Already aggregated across all step-level depositions, so @c energy is the
+/// total energy and @c time is the energy-weighted average.
+struct CaloHitContribution {
+  /// Row index of the source particle in the simulated
+  /// @c SimParticleContainer; @c std::numeric_limits<std::uint64_t>::max() if
+  /// the source particle was filtered out upstream.
+  std::uint64_t particleRow;
```
benjaminhuth: Hmm. Can't we have our particle id here? And maybe a std::optional instead of that max convention?

plans/edm4hep_sim_input_converter_perf.md
benjaminhuth, line 1: Hmm this is from your local environment, right?

Plugins/Arrow/src/ArrowUtil.cpp
```diff
@@ -0,0 +1,260 @@
...
+ public:
+  explicit Impl(std::filesystem::path path) : m_path(std::move(path)) {}
+
+  void write(const arrow::Table& table) {
+    if (!m_writer) {
+      auto outfile = unwrap(arrow::io::FileOutputStream::Open(m_path.string()),
+                            "open parquet");
+      auto properties = parquet::WriterProperties::Builder()
+                            .compression(parquet::Compression::ZSTD)
+                            // ->enable_write_page_index()
```
benjaminhuth: dead code?

Plugins/Arrow/include/ActsPlugins/Arrow/ArrowUtil.hpp
```diff
@@ -0,0 +1,103 @@
...
+/// @param eventId The event id value to stamp.
+/// @return A new table with the event id column prepended.
+ACTS_ARROW_EXPORT std::shared_ptr<arrow::Table> withEventId(
+    const std::shared_ptr<arrow::Table>& table, std::uint64_t eventId);
+
+/// Thin RAII wrapper around @c parquet::arrow::FileWriter that opens lazily
+/// on first write so the schema can be taken from the first event's table.
+/// Page index is enabled so the reader can locate the matching row's pages
+/// in each column without decoding the entire row group on every read.
+class ACTS_ARROW_EXPORT ParquetFileWriter {
```
benjaminhuth, line 156: This only writes one file right, in asymmetry to the reader?

Examples/Io/EDM4hep/include/ActsExamples/Io/EDM4hep/EDM4hepCaloHitInputConverter.hpp
benjaminhuth, line 1: Hmm I also would argue we should separate these changes in a separate PR (seem arrow/parquet unrelated?)

Examples/Io/Parquet/src/ParquetWriter.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  std::filesystem::path shardPath(const std::filesystem::path& directory,
+                                  std::uint64_t shardId) const {
+    // Derive a file-name prefix from the configured per-collection
+    // directory: e.g. `<outputDir>/particles.parquet/` → `particles`,
+    // `<outputDir>/particles/` → `particles`. Falls back to `shard` if the
+    // directory has no usable stem (root dir, trailing slash, etc.).
+    std::string prefix = directory.filename().stem().string();
+    if (prefix.empty()) {
+      prefix = "shard";
```
benjaminhuth: Hmm shouldn't we throw on this?

Examples/Io/Parquet/src/ParquetWriter.cpp
```diff
@@ -0,0 +1,245 @@
...
+    // fewer events than the name implies; consumers should read the file
+    // and trust its row count, not the filename.
+    const std::uint64_t startEvent = shardId * m_cfg.eventsPerShard;
+    const std::uint64_t endEvent = startEvent + m_cfg.eventsPerShard;
+    return directory /
+           std::format("{}_{:06}-{:06}.parquet", prefix, startEvent, endEvent);
+  }
+
+  ParquetWriter::Config m_cfg;
+  std::vector<std::unique_ptr<CollectionState>> m_states;
```
benjaminhuth: Can we rename this to `m_collectionStates`, this makes downstream code easier to read I think

benjaminhuth (commented):
I have some comments, but not fully finished the review. One of the main points is the question if we need to bundle the arrow-interface and the Calo EDM together in one PR

Examples/Framework/include/ActsExamples/EventData/CaloHit.hpp
paulgessinger: I can split it off but I tested it together.

Examples/Framework/include/ActsExamples/EventData/CaloHit.hpp
paulgessinger: I was modeling this only after the output format. I don't foresee us wanting to use this, outside of the conversion. I can still change this to the barcode though.

Examples/Framework/include/ActsExamples/EventData/CaloHit.hpp
paulgessinger: I didn't foresee changing the units (I don't think @murnanedaniel did for the dataset). If needed, I can add the conversion.

Examples/Io/Parquet/src/ParquetWriter.cpp
paulgessinger: Agreed

Plugins/Arrow/src/ArrowUtil.cpp
paulgessinger: I wanted to turn this on again, I commented it out while debugging. I'll look at this.

Plugins/Arrow/include/ActsPlugins/Arrow/ArrowUtil.hpp
paulgessinger: Yes, because I delegate the discovery of the right file to read from to the dataset system. The dataset system (as far as I understand it now) can also write multiple files, but we lose the chunking by event number, the consistent event-to-chunk association, and the file naming. 

Using the dataset system for reading also buys us (as far as I understand it) schema evolution.

Plugins/Arrow/src/ArrowUtil.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+    m_numEvents = 0;
+    for (const auto& f : files) {
+      auto infile =
+          unwrap(arrow::io::ReadableFile::Open(f, arrow::default_memory_pool()),
+                 "open parquet footer");
+      auto reader =
+          unwrap(parquet::arrow::OpenFile(infile, arrow::default_memory_pool()),
+                 "open parquet reader");
+      m_numEvents += reader->parquet_reader()->metadata()->num_rows();
```
paulgessinger, line 343: @benjaminhuth this actually reads the footers of the files to discover the correct number of events. (fyi)

Examples/Framework/include/ActsExamples/EventData/CaloHit.hpp
benjaminhuth: Hmm but if we go in our EDM, we should use our native unites, no?

Examples/Framework/include/ActsExamples/EventData/CaloHit.hpp
benjaminhuth: Okay, but if we don't want to use it, why do we want to add it? Maybe I'm missing something here...

paulgessinger: Pulled out to a separate PR.

Examples/Framework/include/ActsExamples/EventData/CaloHit.hpp
paulgessinger: Pulled out to a separate PR.

Examples/Framework/include/ActsExamples/EventData/CaloHit.hpp
paulgessinger: Split out.

Detray/tests/unit_tests/cpu/navigation/intersection/helix_intersector.cpp
```diff
@@ -58,8 +58,8 @@ const free_track_parameters<test_algebra> free_trk({0.f, 0.f, 0.f}, 0.f,
 // Magnetic field
 const vector3 B{static_cast<scalar>(0.f), static_cast<scalar>(0.f),
                 1.f * unit<scalar>::T};
-const vector3 B_0{0.f * unit<scalar>::T, tol * unit<scalar>::T,
-                  tol * unit<scalar>::T};
+const vector3 B_0{0.f * unit<scalar>::T, tol* unit<scalar>::T,
+                  tol* unit<scalar>::T};
```
paulgessinger: Clang-format formatted this for me (@niermann999 any idea why?)

Examples/Io/Arrow/src/ArrowSimHitOutputConverter.cpp
```diff
@@ -0,0 +1,275 @@
...
+    timeV->UnsafeAppend(t);
+
+    // Digitized global position: project the first matched measurement's
+    // bound parameters through the surface. Multiple measurements per hit
+    // would only happen if a hit migrated across modules during clustering;
+    // we take the first deterministically and leave the rest for a future
+    // "merged hits" extension.
+    float gx = kNaN;
+    float gy = kNaN;
+    float gz = kNaN;
```
paulgessinger, line 197: I'll want to change this to be a proper nullable column, but as a follow up, while explicitly checking if the existing ColliderML files stay readable.

Plugins/Arrow/src/ArrowUtil.cpp
paulgessinger: Back on

benjaminhuth:
Hmm maybe we need to revise the flexibility of the dataset reading: a download of the ColliderML dataset with the colliderml library looks like this:
```
.
└── CERN__ColliderML-Release-1
    ├── ttbar_pu200_particles
    │   ├── data
    │   │   └── ttbar_pu200_particles
    │   │       └── train-00000-of-01000.parquet
    │   └── metadata.json
    └── ttbar_pu200_tracker_hits
        ├── data
        │   └── ttbar_pu200_tracker_hits
        │       └── train-00000-of-01000.parquet
        └── metadata.json
```
I think we cannot map this at the moment...

murnanedaniel:
@benjaminhuth @paulgessinger 

> Hmm maybe we need to revise the flexibility of the dataset reading: a download of the ColliderML dataset with the colliderml library looks like this:
> 
> ```
> .
> └── CERN__ColliderML-Release-1
>     ├── ttbar_pu200_particles
>     │   ├── data
>     │   │   └── ttbar_pu200_particles
>     │   │       └── train-00000-of-01000.parquet
>     │   └── metadata.json
>     └── ttbar_pu200_tracker_hits
>         ├── data
>         │   └── ttbar_pu200_tracker_hits
>         │       └── train-00000-of-01000.parquet
>         └── metadata.json
> ```
> 
> I think we cannot map this at the moment...

Is this a blocker though? One can output to a location then re-arrange as needed. Or do you specifically mean allowing different shard sizes?

paulgessinger:
What flexibility is missing @benjaminhuth ?

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ArrowSimHitOutputConverter.hpp
```diff
@@ -0,0 +1,107 @@
...
+    /// to fill the digitized @c x,y,z columns; otherwise those are NaN.
+    std::string inputMeasurements;
+    /// Optional sim-hit → measurement(s) inverse map; keyed by @c SimHitIndex.
+    std::string inputSimHitMeasurementsMap;
+    /// Output whiteboard key for the resulting @c arrow::Table.
+    std::string outputTable = "simhits";
+    /// Tracking geometry used to find surfaces by @c GeometryIdentifier when
+    /// projecting the matched measurement's local parameters back to global
+    /// coordinates. Required for digitized @c x,y,z to be populated.
+    std::shared_ptr<const Acts::TrackingGeometry> trackingGeometry;
```
benjaminhuth: Hmm our clusters have the global position property, so maybe it would be better to use this instead of doing the projection here?

Examples/Io/Arrow/src/ArrowSimHitOutputConverter.cpp
```diff
@@ -0,0 +1,275 @@
...
+    if (simHitMeasMap != nullptr && measurements != nullptr) {
```
benjaminhuth: As said above, I think this whole block exists basically also here:

https://github.com/acts-project/acts/blob/ae7aa614cdc7ba285fe8809056520aba8cfed893/Examples/Algorithms/Digitization/src/MeasurementCreation.cpp#L36

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ArrowSimHitOutputConverter.hpp
```diff
@@ -0,0 +1,107 @@
...
+namespace ActsExamples {
+
+/// Convert a @c SimHitContainer to an @c arrow::Table.
+///
+/// The output table has one row per event with list-valued columns. Hits are
+/// emitted in @c SimHitContainer iteration order, so the row index of a hit
+/// inside the per-event list equals its @c SimHitIndex; downstream tables
+/// (e.g. tracks) can therefore reference hits by that index.
+///
+/// When @c inputMeasurements, @c inputSimHitMeasurementsMap, and
```
benjaminhuth: This is basically a ColliderML writer, right? Because the simhits and measurements are decoupled in our EDM... So maybe the more appropriate name would be `ColliderMLHitWriter` or so?

Examples/Io/Arrow/src/ArrowParticleOutputConverter.cpp
```diff
@@ -0,0 +1,424 @@
...
+    vyV->UnsafeAppend(static_cast<float>(pos.y()));
+    vzV->UnsafeAppend(static_cast<float>(pos.z()));
+    timeV->UnsafeAppend(static_cast<float>(s.time()));
+    pxV->UnsafeAppend(static_cast<float>(mom.x()));
+    pyV->UnsafeAppend(static_cast<float>(mom.y()));
+    pzV->UnsafeAppend(static_cast<float>(mom.z()));
+
+    std::optional<float> d0;
+    std::optional<float> z0;
+    if (m_perigee->propagator.has_value()) {
```
benjaminhuth: Hmm so this functionality is at least partially duplicated in the `RootTrackSummaryWriter`. I believe we should extract and harmonize this somehow.

Examples/Io/Parquet/src/ParquetReader.cpp
```diff
@@ -0,0 +1,159 @@
...
+      state->handle->initialize(name);
+      m_states.push_back(std::move(state));
+    }
+
+    m_eventsRange = {
+        0, referenceEvents < 0 ? 0 : static_cast<std::size_t>(referenceEvents)};
+  }
+
+  ParquetReader::Config m_cfg;
+  std::vector<std::unique_ptr<CollectionState>> m_states;
```
benjaminhuth: maybe also rename this to `m_collectionStates` for consistency

Examples/Io/EDM4hep/include/ActsExamples/Io/EDM4hep/EDM4hepSimInputConverter.hpp
benjaminhuth, line 1: In principle, also this seems separate form the arrow/parquet stuff, right?

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -230,6 +249,71 @@
             removeNeutral=True,
         ),
     )
+elif args.hepmc3:
```
benjaminhuth: This block seems also unrelated to parquet/arrow

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -456,4 +542,94 @@
...
+        )
+    except ImportError as e:
+        raise RuntimeError(
+            "Parquet output requested but acts.examples.arrow is not available; "
+            "rebuild with ACTS_BUILD_EXAMPLES_PARQUET=ON."
+        ) from e
+
+    # ODD volume -> detector enum mapping used in parquet simhit export.
+    # 0..8 are subdetector-specific enums; 255 marks unknown/unmatched.
+    _odd_detector_resolver = makeVolumeIdDetectorResolver(
```
benjaminhuth, line 494: for now this is okay, but maybe we want to have this as a json file?

Python/Examples/src/plugins/EDM4hep.cpp
benjaminhuth, line 1: Unrelated?

benjaminhuth (commented):
Hey, still another round of comments

Examples/Scripts/Python/full_chain_odd.py
paulgessinger: If we settle on a pattern for this, yes. This will also change with a Gen3 version again

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ArrowSimHitOutputConverter.hpp
paulgessinger: It's optional, it doesn't have to write the measurements, so I'm not sure? I don't think we should have more than one Arrow format, so I'd rather keep the name technical and descriptive rather than a label. Does that make sense?

Examples/Io/Arrow/src/ArrowSimHitOutputConverter.cpp
paulgessinger: But it takes `DigitizedParameters` and not a measurement? I don't think I have access to the cluster here. The code duplication is ~ 3 lines (the `localToGlobal`. I'm not sure I follow.

paulgessinger: Ah, but you're right that it's likely missing the center position for 1D clusters (strips). I don't think I can reuse that function though.

benjaminhuth: No, I rather meant we add clusters to the input of the converter and use the data from the EDM

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ArrowSimHitOutputConverter.hpp
paulgessinger: I think we only have the simhit -> measurement map but not one for mapping to the clusters, so not easily accessible.

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ArrowSimHitOutputConverter.hpp
benjaminhuth: Yeah, I just wondered because this is very different from our usual I/O, where we clearly separate simhits and measurements... but is not critical

Examples/Scripts/Python/full_chain_odd.py
benjaminhuth: Yeah I feel, for some period we might maintain a few conversion files etc...

Examples/Io/Arrow/src/ArrowParticleOutputConverter.cpp
paulgessinger: I'll look into that.

Examples/Io/EDM4hep/include/ActsExamples/Io/EDM4hep/EDM4hepSimInputConverter.hpp
paulgessinger: You're right, I was writing indices from EDM4hep before, but I stopped doing that.

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ArrowSimHitOutputConverter.hpp
benjaminhuth: As discussed, the clusters are indexed the same way as the measurements

paulgessinger: Done

Examples/Io/Arrow/src/ArrowParticleOutputConverter.cpp
paulgessinger: Unified now.

Examples/Framework/src/Utilities/PerigeeParameters.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "Acts/Propagator/Propagator.hpp"
+#include "Acts/Propagator/PropagatorOptions.hpp"
+#include "Acts/Surfaces/BoundaryTolerance.hpp"
+#include "Acts/Surfaces/PerigeeSurface.hpp"
+
+#include <utility>
+
+namespace ActsExamples {
+
+std::optional<Acts::BoundTrackParameters> propagateToPerigee(
```
andiwand:
```suggestion
std::optional<Acts::BoundTrackParameters> ActsExamples::propagateToPerigee(
```
nitpick: would avoid linker errors if one modifies the parameters or function name

Examples/Framework/src/Utilities/PerigeeParameters.cpp
```diff
@@ -0,0 +1,69 @@
...
+    if (!lp.ok()) {
+      return std::nullopt;
+    }
+    // Direction and momentum are unchanged by a straight-line extrapolation,
+    // so only the local position differs from the truth parameters.
+    Acts::BoundVector params;
+    params << lp.value(), start.phi(), start.theta(), start.qOverP(),
+        start.time();
+    return Acts::BoundTrackParameters(std::move(surface), params, std::nullopt,
+                                      start.particleHypothesis());
```
andiwand: this would not extrapolate time. should this rather just call the normal propagation?

Examples/Framework/src/Utilities/PerigeeParameters.cpp
```diff
@@ -0,0 +1,69 @@
...
+  // reached; we only steer which side to propagate towards.
+  Acts::PropagatorPlainOptions options(geoContext, magFieldContext);
+  options.direction =
+      Acts::Direction::fromScalarZeroAsPositive(intersection.pathLength());
+
+  auto result = propagator.propagateToSurface(start, *surface, options);
+  if (!result.ok()) {
+    return std::nullopt;
+  }
+  return std::move(*result);
```
andiwand: in case the above is true, this code block might be easy enough to move to the caller side?

andiwand:
This adds quite some complexity, I remember some of it after you explained it to me. Can we summarize this in the PR description? LLM summary would be enough from my side which could also guide the review a bit

Examples/Framework/src/Utilities/PerigeeParameters.cpp
paulgessinger: I *never* remember this 😅

andiwand: I think this is also very much a matter of taste 😅 it jus always pops into my eye

Examples/Framework/src/Utilities/PerigeeParameters.cpp
paulgessinger: I believe this is what `RootParameterWriter` was doing before. I think this could also call the full propagation with neutral params, yes.

Examples/Framework/include/ActsExamples/Utilities/PerigeeParameters.hpp
```diff
@@ -0,0 +1,48 @@
...
+///
+/// @param geoContext Geometry context for the surface operations.
+/// @param magFieldContext Magnetic field context for the propagation.
+/// @param propagator Propagator used for charged particles.
+/// @param surface Perigee surface to express the parameters on.
+/// @param start Truth (curvilinear) parameters at the production vertex.
+///
+/// @return The bound parameters on @p surface, or @c std::nullopt if the
+///         extrapolation (neutral) or propagation (charged) failed.
+std::optional<Acts::BoundTrackParameters> propagateToPerigee(
```
benjaminhuth: Is there a reason not to return Acts::Result?

Examples/Framework/src/Utilities/PerigeeParameters.cpp
benjaminhuth: Okay, after checking again, the `RootTrackSummaryWriter` does not invoke the propagater at all...

https://github.com/acts-project/acts/blob/ff34fc388dc72df5c3f46fdcd9904aff93e72cba/Examples/Io/Root/src/RootTrackSummaryWriter.cpp#L385

But we should unify this in a subsequent PR I think...

benjaminhuth:
> What flexibility is missing @benjaminhuth ?

For the record: I think I was not correct, I don't really remember what exact issue I was encountering...

Examples/Framework/src/Utilities/PerigeeParameters.cpp
paulgessinger: @benjaminhuth I'm not sure I follow. The particle writer was calling the propagation prior to this PR. Is that what you meant?

paulgessinger: Ah, you mean extending this to also cover the summary writer.

Examples/Framework/src/Utilities/PerigeeParameters.cpp
paulgessinger: Removed this for now.

benjaminhuth (dismissed):
LGTM!

paulgessinger:
@andiwand added a PR description now.

Examples/Scripts/Python/full_chain_odd.py
```diff
@@ -127,19 +134,25 @@
 parser.add_argument(
     "--output-root",
     help="Switch root output on/off",
-    default=True,
+    default=False,
```
benjaminhuth, line 137: Hmm could these changed defaults cause the ML ambit test to fail?

```
FAILED Python/Examples/tests/test_examples.py::test_ML_Ambiguity_Solver - AssertionError: assert False
 +  where False = exists()
 +    where exists = PosixPath('/tmp/pytest-of-root/pytest-0/test_ML_Ambiguity_Solver0/odd_output/performance_finding_ambiML.root').exists
```

benjaminhuth (approved):

PR #5411, opened 2026-05-06
feat: Add hierarchical alignment structures

knutzk:
This PR introduces group-level alignment, i.e., the ability to align a rigid set of detector modules as a single unit (layer, stave), rather than only individual surfaces. Two new classes implement this and are wired into the existing code.

--- END COMMIT MESSAGE ---

**What is added:**
- `AlignableStructure`: a group of surfaces sharing alignment DoFs, with an `AlignmentMask` selecting which of the six rigid-body DoFs are free and optional per-DoF variance constraints for regularisation.
- `AlignmentHierarchy`: a registry over a list of structures providing a detector-element-to-structure lookup, plus two validation checks:
  - `validate()`: rejects modules assigned to more than one structure,
  - `detectMaskConflicts()`: rejects configurations where the same DoF floats at both structure and module level simultaneously.

The alignment engine now builds and validates a hierarchy from `AlignmentOptions` before each iteration and tags measurement states with their owning structure. Python bindings are exposed in `acts.examples.alignment`. Handing a hierarchy to `align()` from Python will require binding `AlignmentOptions/Alignment<Fitter>` (to be followed up).

**Tests:** New file `Tests/UnitTests/Alignment/Kernel/HierarchyTests.cpp` with three cases: 
- hierarchy validation (clean vs. overlapping assignment),
- structureFor() lookup correctness,
- and mask-conflict detection (overlapping vs. disjoint masks).

**Not yet included:** the actual math ...

### DIFF
--- a/Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
+++ b/Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
@@ -8,7 +8,7 @@
 
 #pragma once
 
-// #include "Acts/Definitions/Alignment.hpp"
+#include "Acts/Definitions/Alignment.hpp"
 #include "Acts/Geometry/GeometryIdentifier.hpp"
 #include "Acts/Surfaces/Surface.hpp"
 #include "ActsAlignment/Kernel/AlignmentMask.hpp"
@@ -19,67 +19,69 @@
 
 namespace ActsAlignment {
 
-/**
- * @brief A class that groups multiple surfaces to be aligned together.
- *
- * AlignableStructure acts as a logical parent for a set of modules (surfaces).
- * It manages the 6 standard degrees of freedom (Center/Rotation) and
- * constraints.
- */
+/// @brief A logical group of surfaces that share alignment degrees of freedom.
+///
+/// AlignableStructure aggregates a set of modules (surfaces) into a single
+/// alignable unit parameterised by the 6 standard rigid-body DoFs
+/// (Center0..2, Rotation0..2) plus optional per-DoF rigidity constraints.
+/// It owns no transform of its own; alignment deltas live in the
+/// AlignmentContext store, exactly like the per-module case.
 class AlignableStructure {
  public:
-  /**
-   * @brief Constructor
-   * @param id The geometric identifier for this structure
-   */
-  explicit AlignableStructure(Acts::GeometryIdentifier id)
-      : m_id(id), m_surfaces(), m_constraints(), m_mask(AlignmentMask::None) {}
+  /// @brief Constructor
+  /// @param id The geometric identifier for this structure
+  explicit AlignableStructure(Acts::GeometryIdentifier id) : m_id(id) {}
 
   /// Disallow copy construction to ensure identity uniqueness
   AlignableStructure(const AlignableStructure&) = delete;
   AlignableStructure& operator=(const AlignableStructure&) = delete;
 
-  /**
-   * @brief Add a surface to this alignable structure
-   * @param surface The surface to add
-   */
+  /// @brief Add a surface to this alignable structure
+  /// @param surface The surface to add
   void addSurface(std::shared_ptr<Acts::Surface> surface) {
     m_surfaces.push_back(std::move(surface));
   }
 
-  /**
-   * @brief Get the geometric identifier
-   * @return The ID of the structure
-   */
+  /// @brief Add a child structure for nested hierarchies
+  /// @param child Non-owning pointer to a child structure
+  void addChild(AlignableStructure* child) { m_children.push_back(child); }
+
+  /// @brief Get the geometric identifier
+  /// @return The ID of the structure
   Acts::GeometryIdentifier geometryId() const { return m_id; }
 
-  /**
-   * @brief Access the list of associated surfaces
-   * @return A vector of shared pointers to surfaces
-   */
+  /// @brief Access the list of associated surfaces
+  /// @return A vector of shared pointers to surfaces
   const std::vector<std::shared_ptr<Acts::Surface>>& surfaces() const {
     return m_surfaces;
   }
 
-  /**
-   * @brief Access the constraints map (Rigidity)
-   *
-   * The map key is the standard alignment index.
-   * The value is the variance constraint (W matrix entry).
-   * @return Reference to the constraints map
-   */
+  /// @brief Access the list of child structures
+  /// @return A vector of non-owning pointers to child structures
+  const std::vector<AlignableStructure*>& children() const {
+    return m_children;
+  }
+
+  /// @brief Access the constraints map (rigidity)
+  ///
+  /// The map key is the standard alignment index. The value is the variance
+  /// constraint \f$\sigma_i^2\f$ contributing \f$W_{ii} = 1/\sigma_i^2\f$ to
+  /// the Hessian diagonal. Setting \f$\sigma_i \to 0\f$ locks DoF @c i.
+  /// @return Reference to the constraints map
   std::map<Acts::AlignmentIndices, double>& constraints() {
     return m_constraints;
   }
+  /// @brief Access the constraints map (const overload)
+  /// @return Const reference to the constraints map
   const std::map<Acts::AlignmentIndices, double>& constraints() const {
     return m_constraints;
   }
 
-  /**
-   * @brief Access the alignment mask (floating DoFs)
-   * @return Reference to the mask
-   */
+  /// @brief Access the alignment mask (floating DoFs)
+  /// @return Reference to the mask
   AlignmentMask& alignmentMask() { return m_mask; }
+  /// @brief Access the alignment mask (const overload)
+  /// @return The mask value
   AlignmentMask alignmentMask() const { return m_mask; }
 
  private:
@@ -89,7 +91,10 @@ class AlignableStructure {
   /// The collection of sensors governed by this structure
   std::vector<std::shared_ptr<Acts::Surface>> m_surfaces;
 
-  /// Map of parameter index -> constraint value (rigidity)
+  /// The child structures for nested hierarchies (non-owning)
+  std::vector<AlignableStructure*> m_children;
+
+  /// Map of parameter index -> variance constraint (rigidity)
   std::map<Acts::AlignmentIndices, double> m_constraints;
 
   /// Bitmask for determining which of the 6 DoFs are floating
--- a/Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
+++ b/Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
@@ -43,8 +43,10 @@ class AlignableStructure {
   }
 
   /// @brief Add a child structure for nested hierarchies
-  /// @param child Non-owning pointer to a child structure
-  void addChild(AlignableStructure* child) { m_children.push_back(child); }
+  /// @param child Shared pointer to a child structure
+  void addChild(std::shared_ptr<AlignableStructure> child) {
+    m_children.push_back(std::move(child));
+  }
 
   /// @brief Get the geometric identifier
   /// @return The ID of the structure
@@ -57,8 +59,8 @@ class AlignableStructure {
   }
 
   /// @brief Access the list of child structures
-  /// @return A vector of non-owning pointers to child structures
-  const std::vector<AlignableStructure*>& children() const {
+  /// @return A vector of shared pointers to child structures
+  const std::vector<std::shared_ptr<AlignableStructure>>& children() const {
     return m_children;
   }
 
@@ -91,8 +93,8 @@ class AlignableStructure {
   /// The collection of sensors governed by this structure
   std::vector<std::shared_ptr<Acts::Surface>> m_surfaces;
 
-  /// The child structures for nested hierarchies (non-owning)
-  std::vector<AlignableStructure*> m_children;
+  /// The child structures for nested hierarchies
+  std::vector<std::shared_ptr<AlignableStructure>> m_children;
 
   /// Map of parameter index -> variance constraint (rigidity)
   std::map<Acts::AlignmentIndices, double> m_constraints;
--- a/Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
+++ b/Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
@@ -1,0 +1,99 @@
+// This file is part of the ACTS project.
+//
+// Copyright (C) 2026 CERN for the benefit of the ACTS project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+// #include "Acts/Definitions/Alignment.hpp"
+#include "Acts/Geometry/GeometryIdentifier.hpp"
+#include "Acts/Surfaces/Surface.hpp"
+#include "ActsAlignment/Kernel/AlignmentMask.hpp"
+
+#include <map>
+#include <memory>
+#include <vector>
+
+namespace ActsAlignment {
+
+/**
+ * @brief A class that groups multiple surfaces to be aligned together.
+ *
+ * AlignableStructure acts as a logical parent for a set of modules (surfaces).
+ * It manages the 6 standard degrees of freedom (Center/Rotation) and
+ * constraints.
+ */
+class AlignableStructure {
+ public:
+  /**
+   * @brief Constructor
+   * @param id The geometric identifier for this structure
+   */
+  explicit AlignableStructure(Acts::GeometryIdentifier id)
+      : m_id(id), m_surfaces(), m_constraints(), m_mask(AlignmentMask::None) {}
+
+  /// Disallow copy construction to ensure identity uniqueness
+  AlignableStructure(const AlignableStructure&) = delete;
+  AlignableStructure& operator=(const AlignableStructure&) = delete;
+
+  /**
+   * @brief Add a surface to this alignable structure
+   * @param surface The surface to add
+   */
+  void addSurface(std::shared_ptr<Acts::Surface> surface) {
+    m_surfaces.push_back(std::move(surface));
+  }
+
+  /**
+   * @brief Get the geometric identifier
+   * @return The ID of the structure
+   */
+  Acts::GeometryIdentifier geometryId() const { return m_id; }
+
+  /**
+   * @brief Access the list of associated surfaces
+   * @return A vector of shared pointers to surfaces
+   */
+  const std::vector<std::shared_ptr<Acts::Surface>>& surfaces() const {
+    return m_surfaces;
+  }
+
+  /**
+   * @brief Access the constraints map (Rigidity)
+   *
+   * The map key is the standard alignment index.
+   * The value is the variance constraint (W matrix entry).
+   * @return Reference to the constraints map
+   */
+  std::map<Acts::AlignmentIndices, double>& constraints() {
+    return m_constraints;
+  }
+  const std::map<Acts::AlignmentIndices, double>& constraints() const {
+    return m_constraints;
+  }
+
+  /**
+   * @brief Access the alignment mask (floating DoFs)
+   * @return Reference to the mask
+   */
+  AlignmentMask& alignmentMask() { return m_mask; }
+  AlignmentMask alignmentMask() const { return m_mask; }
+
+ private:
+  /// The unique identifier for this structure
+  Acts::GeometryIdentifier m_id;
+
+  /// The collection of sensors governed by this structure
+  std::vector<std::shared_ptr<Acts::Surface>> m_surfaces;
+
+  /// Map of parameter index -> constraint value (rigidity)
+  std::map<Acts::AlignmentIndices, double> m_constraints;
+
+  /// Bitmask for determining which of the 6 DoFs are floating
+  AlignmentMask m_mask{AlignmentMask::None};
+};
+
+}  // namespace ActsAlignment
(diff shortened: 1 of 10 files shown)


Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
```diff
@@ -0,0 +1,106 @@
...
+    m_children.push_back(std::move(child));
+  }
+
+  /// @brief Get the geometric identifier
+  /// @return The ID of the structure
+  Acts::GeometryIdentifier geometryId() const { return m_id; }
+
+  /// @brief Access the list of associated surfaces
+  /// @return A vector of shared pointers to surfaces
+  const std::vector<std::shared_ptr<Acts::Surface>>& surfaces() const {
```
paulgessinger: Consider returning a `std::span<const std::shared_ptr<Acts::Surface>>`, or a dereferencing range like in `TrackingVolume`.

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
```diff
@@ -0,0 +1,106 @@
...
+/// AlignableStructure aggregates a set of modules (surfaces) into a single
+/// alignable unit parameterised by the 6 standard rigid-body DoFs
+/// (Center0..2, Rotation0..2) plus optional per-DoF rigidity constraints.
+/// It owns no transform of its own; alignment deltas live in the
+/// AlignmentContext store, exactly like the per-module case.
+class AlignableStructure {
+ public:
+  /// @brief Constructor
+  /// @param id The geometric identifier for this structure
+  explicit AlignableStructure(Acts::GeometryIdentifier id) : m_id(id) {}
```
paulgessinger, line 42: Do we expect each alignable structure to get a unique identifier, or is this intended to share one with some other geometry object? If separate, how do we pick / assign them?

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
```diff
@@ -0,0 +1,106 @@
...
+
+  /// @brief Access the list of associated surfaces
+  /// @return A vector of shared pointers to surfaces
+  const std::vector<std::shared_ptr<Acts::Surface>>& surfaces() const {
+    return m_surfaces;
+  }
+
+  /// @brief Access the list of child structures
+  /// @return A vector of shared pointers to child structures
+  const std::vector<std::shared_ptr<AlignableStructure>>& children() const {
```
paulgessinger, line 75: Consider `std::span`.

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
```diff
@@ -0,0 +1,153 @@
...
+ public:
+  /// @brief Outcome of validate()
+  struct ValidationResult {
+    /// Detector elements that appear in more than one structure. Each offender
+    /// is listed once.
+    std::vector<const Acts::SurfacePlacementBase*> overlapping;
+
+    /// @brief Whether the hierarchy passes validation
+    bool ok() const { return overlapping.empty(); }
+  };
```
paulgessinger, line 48: Is the list of surfaces useful for debugging and that's why it's returned here? Otherwise it might be enough to log them as errors and / or return a `Result` (but maybe what you're doing now is the best approach)

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
```diff
@@ -0,0 +1,153 @@
...
+/// construction and exposes it for later steps of the alignment pipeline
+/// (validation, structure-level derivative accumulation). It does not own the
+/// structures; the caller retains ownership.
+class AlignmentHierarchy {
+ public:
+  /// @brief Outcome of validate()
+  struct ValidationResult {
+    /// Detector elements that appear in more than one structure. Each offender
+    /// is listed once.
+    std::vector<const Acts::SurfacePlacementBase*> overlapping;
```
paulgessinger: Does it make sense to keep the `Surface`s instead, since it's possible to get from a surface to the placement (but not the other way round in all cases)?

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
```diff
@@ -0,0 +1,106 @@
...
+  /// @brief Access the alignment mask (const overload)
+  /// @return The mask value
+  AlignmentMask alignmentMask() const { return m_mask; }
+
+ private:
+  /// The unique identifier for this structure
+  Acts::GeometryIdentifier m_id;
+
+  /// The collection of sensors governed by this structure
+  std::vector<std::shared_ptr<Acts::Surface>> m_surfaces;
```
paulgessinger, line 109: Just for completeness: do we need to keep the surfaces as pointer-to-mutable or would pointer-to-const work?

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
```diff
@@ -0,0 +1,153 @@
...
+    for (const auto& structure : m_structures) {
+      if (structure == nullptr) {
+        continue;
+      }
+      for (const auto& surface : structure->surfaces()) {
+        const auto* detElement = surface->surfacePlacement();
+        if (detElement == nullptr) {
+          continue;
+        }
+        m_detElementToStructure.emplace(detElement, structure.get());
```
paulgessinger: Not critical, but `std::unordered_map::emplace()` returns a `std::pair<Iterator, bool>` where the bool tells you if the element was indeed inserted or if it already existed. So you could check the returned bool to determine the overlap, I think.

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
```diff
@@ -0,0 +1,153 @@
...
+      const Acts::SurfacePlacementBase* detElement) const {
+    auto it = m_detElementToStructure.find(detElement);
+    return it == m_detElementToStructure.end() ? nullptr : it->second;
+  }
+
+  /// @brief Find the structure that owns a given surface
+  /// @param surface The surface to look up
+  /// @return The owning structure, or nullptr if the surface's element is a
+  ///         standalone floating module (or has no attached placement)
+  AlignableStructure* structureFor(const Acts::Surface* surface) const {
```
paulgessinger:
```suggestion
  AlignableStructure* structureFor(const Acts::Surface& surface) const {
```
?

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
```diff
@@ -0,0 +1,153 @@
...
+      return nullptr;
+    }
+    return structureFor(surface->surfacePlacement());
+  }
+
+  /// @brief Check that no detector element is assigned to more than one
+  ///        structure. Mixed mode is allowed: a detector element may be absent
+  ///        from every structure and float as a standalone module.
+  /// @return The validation result
+  ValidationResult validate() const {
```
paulgessinger: Since this doesn't actually compute the validation, maybe


```suggestion
  ValidationResult isValid() const {
```

?

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
```diff
@@ -0,0 +1,153 @@
...
+  ///
+  /// A structure-level DoF and a module-level DoF on the same physical
+  /// element are correlated: if both are floating, the Hessian picks up a
+  /// zero eigenvalue along that direction. This check returns every such
+  /// overlap so the caller can reject the configuration.
+  /// @param moduleMask The DoFs currently floating at module level
+  ///                   (i.e. the iteration's global alignment mask)
+  /// @param floatingModules Detector elements listed in @c alignedDetElements
+  /// @return One entry per conflicting (detElement, structure) pair
+  std::vector<MaskConflict> detectMaskConflicts(
```
paulgessinger, line 137: So for my understanding, is every DoF *conflict* a problem in this sense? A rotation of the overall structure is not equivalent to a roatation of a shifted nested structure, or is it?

Does this have to be recursive?

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
```diff
@@ -0,0 +1,153 @@
...
+    std::vector<const Acts::SurfacePlacementBase*> overlapping;
+
+    /// @brief Whether the hierarchy passes validation
+    bool ok() const { return overlapping.empty(); }
+  };
+
+  /// @brief Construct from a flat list of structures
+  /// @param structures Shared pointers to the alignable structures
+  explicit AlignmentHierarchy(
+      const std::vector<std::shared_ptr<AlignableStructure>>& structures)
```
paulgessinger, line 53: Not critical, but do you foresee the need to have shared ownership for these structures? If they will always only be registered in a single parent, `std::unique_ptr` might be sufficient.

Alignment/include/ActsAlignment/Kernel/Alignment.ipp
```diff
@@ -243,6 +246,47 @@ ActsAlignment::Alignment<fitter_t>::align(
...
+  const auto validation = hierarchy.validate();
+  if (!validation.ok()) {
+    for (const auto* detElement : validation.overlapping) {
+      ACTS_ERROR("Detector element "
+                 << detElement
+                 << " (Surface ID: " << detElement->surface().geometryId()
+                 << ") is assigned to multiple alignable structures");
+    }
+    return AlignmentError::HierarchyValidationFailure;
+  }
```
paulgessinger, line 302: Ok I see. An alternative (I'm not suggesting it's better) would be to pass in a logger to `validate` and let it internally print a log message, and then return the error directly from `validate`. OTOH, this also works and keeps it higher level, so I'm on board.

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
goblirsc: Will this / should this also return recursively surfaces within children?

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
```diff
@@ -0,0 +1,106 @@
...
+    return m_surfaces;
+  }
+
+  /// @brief Access the list of child structures
+  /// @return A vector of shared pointers to child structures
+  const std::vector<std::shared_ptr<AlignableStructure>>& children() const {
+    return m_children;
+  }
+
+  /// @brief Access the constraints map (rigidity)
```
goblirsc, line 79: could add documentation on how a free DoF is encoded. Are they intended to be represented by missing entries?

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
```diff
@@ -0,0 +1,153 @@
...
+      : m_structures(structures) {
+    std::unordered_map<const Acts::SurfacePlacementBase*, unsigned int> counts;
+    for (const auto& structure : m_structures) {
+      if (structure == nullptr) {
+        continue;
+      }
+      for (const auto& surface : structure->surfaces()) {
```
goblirsc: for this to work with nested substructures, would we need a recursive version of `surfaces()`?

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
```diff
@@ -0,0 +1,153 @@
...
+        m_overlapping.push_back(detElement);
+      }
+    }
+  }
+
+  /// @brief Find the structure that owns a given detector element
+  /// @param detElement The detector element to look up
+  /// @return The owning structure, or nullptr if the element is a standalone
+  ///         floating module
+  AlignableStructure* structureFor(
```
goblirsc: One general conceptual question here would be, what do we do for nested structures? 
E.g. (ITk example) 
Strip system -->  endcap A-side --> disk 1 --> petal 2 --> module X ?  

Will it only be the direct parent allowd to have D.o.F? 
What if all of the structures have alignment D.o.F. enabled? 
This is a conveivable scenario (used for example in CMS) , where the global structure movements and local relative movements of sub-assemblies / sensors relative to their parent can be nicely encoded using the `AlignableStructure` mechanism with its recursion logic - but it would not be supported in a formalism where we assume each surface to have zero or one alignable parent structure.

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
goblirsc, line 137: I have some doubts about this logic - doesn't it assume that the D.o.F of the modules and the D.o.F of the structures line up? This is a very strong assumption which we may not want to impose on the user. 
For example, let's look at an ITk endcap again: 
- The alignment D.o.F of a strip module would usually be oriented in the module local frame
- while the disk is more likely to be parametrised in the global frame. 
- So a module-level D.o.F would actually map to a _linear combination_ of structure-level D.o.F and vice-versa. 
- The blind directions from the hessian would depend on the concrete transforms, and not be simply overlapping entries in a bit mask. 

Even if the entire detector used aligned degrees of freedom, this approach of forbidding overlaps between structures and their children risks imposing some strong limitations: 
- Consider a setup where we want to allow, when writing our alignment inputs, both high-level and low-level alignment. We would deliberately want to allow all D.o.F., and then later add restrictions when we fit. 
- Or consider an alignment where we have both high- and low-level degrees of freedom active at the same time, with (possibly metrology motivated) constraints on the local D.o.F but not on the structure-level D.o.F, and equality constraints in the fit to lock the global structure movement rather than strict vetoes. Such a setup would be rejected by the `MaskConflict` check.

goblirsc (commented):
Hi, nice work, thanks a billion! :) 
I added a few comments. My main concern is that the assumption of at most one alignable structure per surface, and the explicit veto of shared alignment indices between modules and structures, form a very rigid framework that limits how this infrastructure can be used in practice - possibly unnecessarily?

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
```diff
@@ -0,0 +1,106 @@
...
+  ///
+  /// The map key is the standard alignment index. The value is the variance
+  /// constraint \f$\sigma_i^2\f$ contributing \f$W_{ii} = 1/\sigma_i^2\f$ to
+  /// the Hessian diagonal. Setting \f$\sigma_i \to 0\f$ locks DoF @c i.
+  /// @return Reference to the constraints map
+  std::map<Acts::AlignmentIndices, double>& constraints() {
```
andiwand: if the order is not relevant `std::unordered_map` is usually the "better" choice

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
```diff
@@ -0,0 +1,153 @@
...
+      const Acts::SurfacePlacementBase* detElement) const {
```
andiwand:
```suggestion
      const Acts::SurfacePlacementBase& detElement) const {
```

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
andiwand:
```suggestion
  const AlignableStructure* structureFor(
```
do we need mutable access?

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
andiwand:
```suggestion
  const AlignableStructure* structureFor(const Acts::Surface* surface) const {
```
do we need mutable access?

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
knutzk: Now uses a `SurfaceRange` which dereferences in a similar way as done in `TrackingVolume`

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
knutzk: I think unique is preferred? We don't want to be in a position where we 'have to' stick to pre-existing geometric structures. We'll need to find a strategy to assign unique IDs, so I flagged this in the comment.

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
knutzk: I think the explicit container type is preferred here, since we're exposing internal members. The alternative would be dereferencing (as done above), but since we want to be able to regress through the children, I think a direct shared pointer vector is preferable in this case

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
knutzk: Just to remain fully flexible in how clients would want to continue using that information, I think returning the actual SurfacePlacementBase elements seems like the best solution

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
knutzk: Good idea! Now doing that instead.

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
knutzk: Since we now use the `ConstDereference` in the exposed `SurfaceRange`, which enforces read-only access, I think it's easiest (and more practical) to keep the pointer-to-mutable version.

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
knutzk: Interesting, I didn't know that. Now explicitly checking what the bool return is.

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
knutzk: Changed here and in other places

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
knutzk: Yeah, you're right. As discussed in our chat a few weeks ago, this was an overly simplistic approach, neglecting the fact that mask conflicts are only true if the reference frames are identical. We'll need the chain-rule logic implemented first, before we can truly detector conflicts. So I turned this function signature into a stub now, and we'll have to implement it at a later point.

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
knutzk: To me, `isValid` would imply a boolean return type, but since we return a result object here, I think the original name makes more sense

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
knutzk: In principle no, so unique_ptr could be better. But as far as I know, this is not very straightforward to implement on the Python side (we would have to use `py::smart_holder` to transfer ownership or drop the Python bindings). Unless we want to do any of these changes, `shared_ptr` seems easier to me.

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
knutzk: I think it shouldn't. At least not in this method, because here we really want to know which surfaces are directly associated with the structure. We might need a second method to walk through all surfaces transitively.

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
knutzk: Now explicitly mentioned, thanks!

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
knutzk: Absolutely right, this is now implemented

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
knutzk: I do think we generally want to support this multi-level alignment. The key design question is whether the alignment code is set up separately for each of these fits (and each time with its own alignment hierarchy, which can shift from call to call), or whether we want to be able to do multi-level alignment *in the same call*. The current code only supports the former. It's something we should discuss

Alignment/include/ActsAlignment/Geometry/AlignmentHierarchy.hpp
knutzk: Yes, I agree with your doubts. I removed the implementation of this function and replaced it with a stub. We'll need to implement the chain rule first, before we can truly assess overlap in the degrees of freedom. My assumption for this implementation above was a bit too naive.

knutzk:
Rebasing onto `main` and undrafting

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
paulgessinger: It would just hide the fact that it's a vector, but would basically behave the same otherwise.

Alignment/include/ActsAlignment/Geometry/AlignableStructure.hpp
paulgessinger: ok

paulgessinger:
Hey @knutzk, sorry this sat for a while. Should we try to get it in? 

@goblirsc what do you think?

goblirsc:
> Hey @knutzk, sorry this sat for a while. Should we try to get it in?
> 
> @goblirsc what do you think?

Hi, would be great to get this in! 

I think we still need to converge on the final way of supporting multi-level hierarchies - but this can be finalised in a future PR, and what we have here will be a great start. 

The only other concern I remember was the too rigid check for redundancy, but I think this was already removed.

PR #5416, opened 2026-05-07
chore: Add `referenceFrame` overload without position/direction for plane surfaces

andiwand:
--- END COMMIT MESSAGE ---


paulgessinger (commented):
Could we define it on `RegularSurface` similar to the `localToGlobal` overloads without direction?

paulgessinger (approved):

PR #5417, opened 2026-05-07
feat: Iterative gauss fit for performance plots

andiwand:
--- END COMMIT MESSAGE ---


Examples/Io/Root/include/ActsExamples/Io/Root/RootTrackFitterPerformanceWriter.hpp
```diff
@@ -54,10 +54,11 @@ class RootTrackFitterPerformanceWriter final
 
     /// Minimum number of entries in a bin for it to be included in the
     /// mean/width fit.
-    int minEntriesForFit = 10;
-    /// Fit option string to pass to ROOT when performing Gaussian fits for
-    /// mean/width extraction.
-    std::string fitOption = "QS0";
+    int fitMinEntries = 10;
```
benjaminhuth, line 57: I think we should expose these configs to python

Plugins/Root/src/HistogramConverter.cpp
```diff
@@ -23,6 +25,49 @@
...
+  double sigmaError;
+};
+
+std::optional<FitResult> iterativeGaussFit(TH1& hist, double sigmaRange,
+                                           int iterations,
+                                           const Acts::Logger& logger) {
+  TFitResultPtr result = hist.Fit("gaus", "SQ0", nullptr);
+  if ((result.Get() == nullptr) || ((result->Status() % 1000) != 0)) {
+    ACTS_DEBUG("Failed to fit Gaussian: status "
+               << (result.Get() != nullptr ? result->Status() : -1));
```
benjaminhuth: Hmm is -1 a good status in case result is `nullptr`? I would maybe rather emit a string...

benjaminhuth (commented):
Looks good, two comments here.

benjaminhuth (approved):

Plugins/Root/src/HistogramConverter.cpp
```diff
@@ -37,18 +37,23 @@ struct FitResult {
...
-  TFitResultPtr result;
+  TFitResultPtr result = hist.Fit("gaus", "SQ0", nullptr);
+  if ((result.Get() == nullptr) || ((result->Status() % 1000) != 0)) {
+    ACTS_DEBUG("Failed to fit Gaussian: status "
+               << (result.Get() != nullptr ? result->Status() : -1));
+    return std::nullopt;
+  }
 
-  double mean = hist.GetMean();
-  double sigma = hist.GetRMS();
```
paulgessinger: By the way, turns out `GetRMS()` aliases and returns the standard deviation, meaning that non-zero distributions are fine. Maybe the fit is overkill in that case actually.

andiwand: I did expect this to return the standard deviation but it seems less stable than an initial global fit

paulgessinger: Fit is definitely not wrong, so let's got with it.

PR #5418, opened 2026-05-07
feat: Enhance spacepoint strip building and fix pairing 

murnanedaniel:
--- END COMMIT MESSAGE ---


Core/include/Acts/SpacePointFormation2/StripSpacePointBuilder.hpp
```diff
@@ -45,6 +45,12 @@ struct ConstrainedOptions final {
   double stripLengthTolerance = 0.01;
   /// Tolerance scaling factor of the gap between strip detector elements
   double stripLengthGapTolerance = 0.01;
+  /// ATLAS-SCTGapParameter analogue: scales an extra |m|, |n| allowance
```
benjaminhuth: Hmm I would try to avoid ATLAS references here, and just explain what the parameter does (similar below), also , what is m and n here?

Core/src/SpacePointFormation2/StripSpacePointBuilder.cpp
```diff
@@ -161,6 +162,27 @@ Result<void> computeConstrainedFormationState(
...
+  // Geometric extra allowance from the wafer-thickness gap between the two
+  // stereo surfaces (ATLAS SCTGapParameter analogue). Disabled when
+  // stripGapParameter == 0 (legacy behaviour, regression-safe).
+  if (stripGapParameter > 0.0) {
+    const Vector3 gapVec =
+        0.5 * (stripEnds2.top + stripEnds2.bottom - stripEnds1.top -
+               stripEnds1.bottom);
+    const double sdir1n = state.firstBtmToTop.norm();
+    const double sdir2n = state.secondBtmToTop.norm();
+    if (sdir1n > 1e-9 && sdir2n > 1e-9) {
```
benjaminhuth: Can we make the 1e-9 a named constant?

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/SpacePointMaker.hpp
```diff
@@ -61,6 +59,26 @@ class SpacePointMaker final : public IAlgorithm {
     std::vector<Acts::GeometryIdentifier> geometrySelection;
 
     std::vector<Acts::GeometryIdentifier> stripGeometrySelection;
+
+    /// Strip cluster pairing strategy, applied after partner-module
```
benjaminhuth: I think this should be an `enum` and not a string.

Examples/Algorithms/TrackFinding/src/SpacePointMaker.cpp
```diff
@@ -221,18 +231,19 @@ SpacePointMaker::SpacePointMaker(Config cfg,
...
-    m_cfg.geometrySelection.erase(geoSelLastUnique.begin(),
-                                  geoSelLastUnique.end());
+  // dedupe both selection lists: stripGeometrySelection's per-(vol, layer)
+  // partner setup throws on duplicate insert otherwise.
+  for (auto* sel :
+       {&m_cfg.geometrySelection, &m_cfg.stripGeometrySelection}) {
+    std::ranges::sort(*sel, std::less<Acts::GeometryIdentifier>{});
+    const auto last = std::ranges::unique(*sel, isDuplicate);
+    if (last.begin() != last.end()) {
+      ACTS_LOG_WITH_LOGGER(this->logger(), Acts::Logging::WARNING,
```
benjaminhuth: I think we can use our usual logging macro here, because the logger is exposed as `logger()`

Examples/Algorithms/TrackFinding/src/SpacePointMaker.cpp
```diff
@@ -272,17 +283,11 @@ void SpacePointMaker::initializeStripPartners() {
       allSensitivesVector.begin(), allSensitivesVector.end());
 
   for (auto selector : m_cfg.stripGeometrySelection) {
-    // Apply volume/layer range
-    auto rangeLayer =
+    // No `extra` filter: stereo partners live at different `extra` values,
```
benjaminhuth: Hmm is that true? I would have thought the `extra` bits distinguish rings, and not sides of stereo partners, see here:

https://github.com/acts-project/acts/blob/8ecb262043a42f1dd7912834f55be61a074f6bd3/Python/Examples/python/odd.py#L124-L127

Should be only have different extra bits if they are different in r.

benjaminhuth (changes requested):
Looks good! A few things to address before merging that I think...

Core/include/Acts/SpacePointFormation2/StripSpacePointBuilder.hpp
andiwand: I thought this was a 1:1 port from Athena - was part of it missing? do you have a link to the original code?

Examples/Algorithms/TrackFinding/src/SpacePointMaker.cpp
benjaminhuth: Yeah, so our usual pattern is to rename the constructor argument to `logger_` or so and still use the logger. But fine for me

Examples/Algorithms/TrackFinding/src/SpacePointMaker.cpp
```diff
@@ -272,15 +281,19 @@ void SpacePointMaker::initializeStripPartners() {
...
   for (auto selector : m_cfg.stripGeometrySelection) {
-    // Apply volume/layer range
+    // Apply (volume, layer) selection first, then filter by `extra` when
+    // the selector specifies a specific ring. `extra` distinguishes
+    // concentric rings within a layer (see GeometryIdentifier docs and
+    // e.g. the ODD geometry config which assigns `extra` from radial
+    // cuts). Stereo partners live in the *same* ring (same `extra`); we
+    // run the partner search per-ring so the closest-distance heuristic
+    // never has to disambiguate across rings.
     auto rangeLayer =
```
benjaminhuth: where I see that, we could combine the `selectLowestNonZeroGeometryObject` + `std::views::filter` to a small helper function, as this also contains some logic.

benjaminhuth (commented):
Code wise that looks good to me, only to minor comments. Before approving, I would like to understand if the interaction between this and https://github.com/acts-project/acts/pull/5445 is understood, to my understanding they are orthogonal and can both be merged independently

Core/include/Acts/SpacePointFormation2/StripSpacePointBuilder.hpp
murnanedaniel: The original code is here: 

https://gitlab.cern.ch/atlas/athena/-/blob/main/InnerDetector/InDetRecTools/SiSpacePointTool/src/SiSpacePointMakerTool.cxx#L106-122

```
double dm = (m_SCTgapParameter * r) * std::abs(s * x12);
double d  = dm / std::sqrt((1. - x12) * (1. + x12));   // dm / sin(stereo)
if (std::abs(T1(2,2)) > 0.7) d *= (r / std::abs(T1(2,3)));   // endcap
// ...added to stripLengthGapTolerance, m_SCTgapParameter capped at 0.002
```

Athena's dm is in mm (uses radial distance r and longitudinal-gap projection s·x12) and is added to stripLengthGapTolerance. The ACTS version is stripGapParameter · |gapVec| / sin(stereo) / halfLen, dimensionless, added to the |m|, |n| ≤ 1+tol limit. 

They should be doing to the same thing, just the units are different. Indeed, adding this gap significantly boosts the efficiency (~40% to 99%, with enough tolerance)

andiwand:
these changes are still valuable I think @murnanedaniel - should we get this rolling again? I could also put some time in

PR #5420, opened 2026-05-07
fix: Revise zero-charge q/p

andiwand:
Revise `ChargeHypothesis` to handle $q=0$ consistently for the $q/p$ parameter. Previously we populated $1/p$ for a neutral particle hypothesis. But this is then often consumed downstream as $q/p$ effectively assigning $q=1$. This is the reason our RK4 steppers will bend trajectories for neutral particle hypotheses right now.

--- END COMMIT MESSAGE ---


Core/include/Acts/EventData/ParticleHypothesis.hpp
```diff
@@ -134,26 +139,90 @@ class ParticleHypothesis final {
...
+        m_momentum{momentum} {
     assert(absPdg == makeAbsolutePdgParticle(absPdg) &&
            "pdg is expected to be absolute");
   }
 
+  /// Create a new particle hypothesis with the same mass and charge but a
+  /// different absolute PDG.
+  /// @param absPdg The new absolute PDG value
+  /// @return A new ParticleHypothesis with the updated absolute PDG
+  ParticleHypothesis withAlteredPdg(PdgParticle absPdg) const {
```
paulgessinger: You could consider marking these `[[nodiscard]]` to make it harder to accidentally misuse.

Fatras/include/ActsFatras/EventData/Particle.hpp
```diff
@@ -237,7 +237,9 @@ class Particle {
   Acts::ParticleHypothesis hypothesis() const {
     return Acts::ParticleHypothesis(
         absolutePdg(), static_cast<float>(mass()),
-        Acts::ChargeHypothesis{static_cast<float>(absoluteCharge())});
+        Acts::ChargeHypothesis{static_cast<float>(absoluteCharge())},
+        charge() != 0 ? std::nullopt
+                      : std::optional<double>(absoluteMomentum()));
```
paulgessinger, line 242: OK naively, this means we're going from `q=0` being sort of a sentinel by itself to the momentum being explicitly optional? I'm still a bit confused what conceptually a "momentum hypothesis" is when we have an actual momentum.

paulgessinger (commented):
So the point is that we free up `qop==0` for neutrals, while storing the momentum inside the hypothesis?

Fatras/include/ActsFatras/EventData/Particle.hpp
andiwand: this converter is practically only used during fatras to launch our propagation for a particle using "the right" particle hypothesis.

for the propagation to work correctly we need a momentum to make the time transport work (and potentially in the future to make interactions work). our track parametrization cannot do that because we don't have a momentum parameter in there, only a curvature parameter which is 0 for neutrals and if `B=0`.

the best way forward in these cases to me is to add another property to the hypothesis, which already carries input parameters "to make the propagation and interactions work", like particle mass and charge. these are kind of bootstrapping the propagation with inputs we cannot observe on the go with a filter.

we cross a bit the reconstruction and simulation boundaries here. the propagation is written for reconstruction purposes and we use it for simulation here. but we can repurpose the momentum hypothesis  here, which one needs in any case during reconstruction if the momentum is not observable like for cases where `B=0`.

andiwand:
> So the point is that we free up `qop==0` for neutrals, while storing the momentum inside the hypothesis?

that and in case the momentum is not observable, like for `B=0`, one can provide a hypothesis to make interactions work. one would then use a fixed value rather than setting `q/p` and therefore also assuming a charge which is overall cleaner and more controlled I believe.

Fatras/include/ActsFatras/EventData/Particle.hpp
paulgessinger: Ok.,

paulgessinger (dismissed):
Ok. I don't love it, but I don't have any better ideas either, so let's do it

paulgessinger (approved):

PR #5421, opened 2026-05-08
refactor: Switch Examples digitization to `DirectedProtoAxis`

andiwand:
Moves our Examples digitization from `BinUtility` to `DirectedProtoAxis`. This gets us one step further to get rid of `BinUtility`

--- END COMMIT MESSAGE ---

blocked by
- https://github.com/acts-project/acts/pull/5531
- https://github.com/acts-project/acts/pull/5540
- https://github.com/acts-project/acts/pull/5541


Examples/Io/Json/src/JsonDigitizationConfig.cpp
```diff
@@ -131,7 +134,11 @@ void ActsExamples::to_json(nlohmann::json& j, const GeometricConfig& gdc) {
     indices.push_back(static_cast<std::size_t>(idx));
   }
   j["indices"] = indices;
-  j["segmentation"] = nlohmann::json(gdc.segmentation);
+  Acts::BinUtility segmentation;
```
benjaminhuth, line 139: just as a question: How big is the overlap here actually? Can we implement a json serialization from `vector<axis>` that is json compatible with the binutility? so that one could in practice do binutility <-> json <-> vector<axis>?
That would make moving forward easier maybe?

benjaminhuth (commented):
Looks good to me overall! As discussed, a deprecation of `ProtoAxis` and `IAxis::createXXX` seems desirable to me...

Examples/Io/Json/src/JsonDigitizationConfig.cpp
asalzburger: Actually, what I would suggest is:
- when reading old files (for a bit), we should be doing binUnitlity -> json (old) -> vector (new), i.e. support reading old files
- I would not spend time in writing old format, so only have it one way.

asalzburger (changes requested):
For me almost good to go, we should just clarify what we do with backward compatibility,

asalzburger:
I marks this ready for review, I think it's practically there, and I guess we should make a move.

Examples/Io/Json/src/JsonDigitizationConfig.cpp
andiwand: I don't think we version these files and I fear there is no good way to detect old way new way other than trying to parse it and falling back or detecting it by inspecting what properties are present.

in any case I don't think we want to have multiple formats in flight so the easiest way forward IMO is to just stick with the old format as long as we can. out of convenience I did this right now through reading/writing `BinUtilities` and then converting. but we would also switch over to a different type but still reading from the existing json format

benjaminhuth: Actually, we version the files:

```
    "acts-geometry-hierarchy-map": {
        "format-version": 0,
        "value-identifier": "digitization-configuration"
    },
```

benjaminhuth (commented):
Hash checks do fail, do we know why?

Examples/Io/Json/src/JsonDigitizationConfig.cpp
andiwand: the file we do version but not a dumped `BinUtilities`

benjaminhuth: Hmm but doesn't that boil down to the same? a part of the file becomes incompatible, so the whole file becomes incompatible?

asalzburger (approved):
LGTM

PR #5422, opened 2026-05-08
feat: Add appending of track states to tracks in PyACTS bindings

H4m5t3r:
feat: Python bindings for appending track states to tracks + setting track state flags

This pull request adds Python bindings that allow for appending track states to tracks and setting track state flags. A brief example of usage has also been added in `track_finding_python_only.py`.

### DIFF
--- a/Examples/Scripts/Python/track_finding_python_only.py
+++ b/Examples/Scripts/Python/track_finding_python_only.py
@@ -79,6 +79,7 @@ def runTrackFindingPythonOnly(
         )
     )
 
+    # Option 1: Implement a track finding algorithm...
     class PythonTrackFinder(acts.examples.IAlgorithm):
         def __init__(self, name, level):
             acts.examples.IAlgorithm.__init__(self, name, level)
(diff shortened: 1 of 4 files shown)


Python/Examples/src/EventData.cpp
```diff
@@ -78,6 +80,85 @@ void addEventData(py::module& mex) {
...
+          "smoothedCovariance",
+          [](const TrackStateProxy& self) {
+            return Acts::BoundMatrix{self.smoothedCovariance()};
+          })
+      .def("setReferenceSurface",
+           [](TrackStateProxy& self,
+              std::shared_ptr<const Acts::Surface> srf) {
+             self.setReferenceSurface(std::move(srf));
+           })
+      .def("getUncalibratedSourceLink",
```
paulgessinger: Properties for these?

Python/Examples/src/EventData.cpp
```diff
@@ -78,6 +80,85 @@ void addEventData(py::module& mex) {
...
+           &TrackStateProxy::getUncalibratedSourceLink)
+      .def("setUncalibratedSourceLink",
+           [](TrackStateProxy& self, const Acts::SourceLink& sourceLink) {
+             self.setUncalibratedSourceLink(Acts::SourceLink{sourceLink});
+           })
+      .def("setHasMeasurement",
+           [](TrackStateProxy& self, bool value = true) {
+             self.typeFlags().setHasMeasurement(value);
```
paulgessinger: Can you bind them through the type flags instead of calling through here?

Also, maybe properties?

Examples/Scripts/Python/track_finding_python_only.py
```diff
@@ -108,7 +109,26 @@ def execute(self, context):
             self.prototracks(context, prototracks)
             return acts.examples.ProcessCode.SUCCESS
 
-    s.addAlgorithm(PythonTrackFinder("PythonTrackFinder", acts.logging.INFO))
+    # s.addAlgorithm(PythonTrackFinder("PythonTrackFinder", acts.logging.INFO))
```
benjaminhuth: I like the idea of showing both versions. for test reasons, I would prefer having the PythonTrackFinder being executed (this script is run in the pyacts tests...)

benjaminhuth (approved):
LGTM!

PR #5423, opened 2026-05-08
feat: Add ODD json roundtrip conversion

benjaminhuth:
--- END COMMIT MESSAGE ---

@paulgessinger I needed to introduce a tolerance at one point to avoid round trip failures because of numeric issues.


Examples/Scripts/Python/geometry.py
```diff
@@ -97,6 +98,12 @@ def runGeometry(
 
             jmw.write(trackingGeometry)
 
+        if outputTrackingGeometryJson:
+            converter = TrackingGeometryJsonConverter(level=acts.logging.INFO)
+            jsonStr = converter.toJson(context.geoContext, trackingGeometry)
+            outPath = outputDir / "json" / f"event{ievt:>09}-tracking-geometry.json"
```
paulgessinger: Is this event dependent?

Plugins/Json/src/GridJsonConverter.cpp
```diff
@@ -32,15 +32,15 @@ std::unique_ptr<Acts::IAxis> Acts::AxisJsonConverter::fromJson(
...
 
   if (axisType == Acts::AxisType::Equidistant) {
+    std::array<double, 2u> range = jAxis.at("range");
     return Acts::IAxis::createEquidistant(boundaryType, range.at(0),
                                           range.at(1), jAxis.at("bins"));
   }
 
   return Acts::IAxis::createVariable(
-      boundaryType, jAxis.at("edges").get<std::vector<double>>());
+      boundaryType, jAxis.at("boundaries").get<std::vector<double>>());
```
paulgessinger, line 43: Do we use this for anything else? Is changing the key a problem?

Python/Examples/tests/test_geometry.py
```diff
@@ -178,3 +179,92 @@ def test_odd_gen3(constructionMethod):
...
+    assert orig["pathLength"].mean() > 0.0
+
+    np.testing.assert_array_equal(orig["nSensitives"], rebuilt_data["nSensitives"])
+    np.testing.assert_array_equal(orig["nPortals"], rebuilt_data["nPortals"])
+    np.testing.assert_array_equal(
+        orig["nSuccessfulSteps"], rebuilt_data["nSuccessfulSteps"]
+    )
+    np.testing.assert_allclose(
+        orig["pathLength"], rebuilt_data["pathLength"], rtol=1e-5
+    )
```
paulgessinger, line 263: I suppose the root hash check is too sensitive to minimal fluctuations?

Core/src/Geometry/SurfaceArrayCreator.cpp
```diff
@@ -27,6 +27,12 @@
 
 namespace Acts {
 
+// Relative guard factor applied to layer tolerances to absorb floating-point
+// rounding when surface centers sit exactly at the geometric boundary (e.g.
+// after a JSON serialization roundtrip where vertex positions may shift by
+// a few ULPs).
+static constexpr double s_layerToleranceEpsilon = 1e-8;
```
paulgessinger: This one I don't understand. If the tolerances are too right for the JSON roundtrip to break things, that's a user error, no?

Python/Examples/tests/test_geometry.py
benjaminhuth: Hmm so since I needed to introduce some floating point tolerance, we have some floating point variance in the roundtrip somewhere... I did not try a hash check though

Examples/Scripts/Python/geometry.py
benjaminhuth: well, in theory yes (geocontext), in practice not right now of course. I just kept this consistent with the rest of the script.

Core/src/Geometry/SurfaceArrayCreator.cpp
benjaminhuth: Hmm so one would need to change the ODD definition in that case? 
So I'm not an expert on the geo construction, so this is definitively a ad-hoc fix in some sense. The issue was that about 25 surfaces were missing I think because of differences of the order of 1e-13... Not sure where this comes from, maybe because the order of operations changes or so?

Plugins/Json/src/GridJsonConverter.cpp
benjaminhuth: This is just a bug, the key for reading and writing were inconsistent. I guess its safer to adjust the reading, as existing files stay or become valid. The conversion was not unit-tested (added in this PR)

Core/src/Geometry/SurfaceArrayCreator.cpp
paulgessinger: Only reason I can see why this would happen is if we're using zero-tolerances, which would mean that the surfaces clip out of the containing layer.

benjaminhuth: Okay, not sure where to look for that, can you point me to where this would be defined?

paulgessinger: It's here: https://github.com/acts-project/acts/blob/main/Plugins/DD4hep/include/ActsPlugins/DD4hep/OpenDataDetectorBuilder.hpp#L60-L63

Should be 2mm, so I don't understand how JSON precision can break this.
But I think this should be understood and not worked around by wiggling the numbers immediately.

Examples/Scripts/Python/geometry.py
benjaminhuth: renamed the output file to be not event dependent, as with obj

Core/src/Geometry/SurfaceArrayCreator.cpp
benjaminhuth: Okay, claude and I refactored this, can you have a look again?

Python/Examples/tests/test_geometry.py
benjaminhuth: Hmm event with the refactor, the pathlength is not bit-identical (seems to have 1e-11 differences). I would keep it like that for now.

Examples/Scripts/Python/geometry.py
```diff
@@ -66,7 +67,7 @@ def runGeometry(
             )
             vis.write(outputDir / "obj" / "geometry.obj")
 
-        if outputJson:
+        if outputPlainJson:
```
paulgessinger: I'm not sure I understand the difference

Python/Examples/tests/test_geometry.py
```diff
@@ -178,3 +178,87 @@ def test_odd_gen3(constructionMethod):
         assert visitor.num_surfaces == 19261
         assert visitor.num_volumes == 109
         assert visitor.num_portals == 437
+
+
+@pytest.mark.skipif(not dd4hepEnabled, reason="DD4hep not set up")
+@pytest.mark.odd
+@pytest.mark.slow
```
paulgessinger: Should this be marked slow?

benjaminhuth: probably not

Examples/Scripts/Python/geometry.py
benjaminhuth: hmm we also have this surface dump thing, which is non-roundtrip-able. I thought it would make sense to distinguish them... But feel free to suggest something better

benjaminhuth: renamed to `outputSurfacesJson`

paulgessinger (approved):

PR #5424, opened 2026-05-09
refactor!: Packed strip space point calibration column

andiwand:
- pack strip space point calibration column
- rename `CopyFromIndex` column and accessors to `CopiedFromIndex`

--- END COMMIT MESSAGE ---

blocked by
- https://github.com/acts-project/acts/pull/5437
- https://github.com/acts-project/acts/pull/5444
- https://github.com/acts-project/acts/pull/5445
- https://github.com/acts-project/acts/pull/5436


andiwand:
> Elevates the pre-computable strip space point information out of the seeding code. This way the computation only needs to happen once, but the user is now responsible for the computation. Therefore this is a breaking change.
> 
> Also allows to deduplicate the `stripCoordinateCheck` again. In a followup PR I plan to elevate this this function and rename it to something like `stripSpacePointCalibration` to make it useable in the parameter estimation.
>
> To be seen what performance difference this makes for ITk strip seeding

After fighting the performance I get a marginal 2% improvement in Athena with the SPOT script. I am back to two functions again but elevated them now to public functions in `Core/SpacePointerFormation2/StripSpacePointBuilder.hpp`.

With the single function and the 5 calibration vectors attached to the space point I saw about ~10% slowdown. My interpretation is that all of these are in the cache and loading them from there is slower than recalculating them from 4 calibration vectors. I think this will highly depend on the CPU and scenario but lets not go down that road.

I also experimented with using Eigen to various degrees but the more we use it the more it slows down...

What is left from this is mainly a refactor and some breaking renames that I believe overall improve things. Next I would explore a bit how to best estimate the seed params using the calibration.

Core/include/Acts/SpacePointFormation2/StripSpacePointBuilder.hpp
```diff
@@ -98,6 +101,128 @@ Vector2 computeVarianceZR(const GeometryContext& gctx, const Surface& surface1,
...
+      ihv[0] * dOuter[0] + ihv[1] * dOuter[1] + ihv[2] * dOuter[2];
+
+  // sInner = stripSeparation dot dOuter
+  // Check if sInner is inside the inner detector element
+  // TODO should this be `sOuter`?
+  const float sInner =
+      scd[0] * dOuter[0] + scd[1] * dOuter[1] + scd[2] * dOuter[2];
+  if (std::abs(sInner) > std::abs(scale) * tolerance) {
+    return false;
+  }
```
andiwand: should this be `sOuter`?

Core/include/Acts/SpacePointFormation2/StripSpacePointBuilder.hpp
```diff
@@ -98,6 +101,128 @@ Vector2 computeVarianceZR(const GeometryContext& gctx, const Surface& surface1,
...
+      ihv[0] * direction[1] - ihv[1] * direction[0]};
+
+  // sOuter = stripSeparation dot dInner
+  // Check if sOuter is inside the outer detector element
+  // TODO should this be `sInner`?
+  const float sOuter =
+      scd[0] * dInner[0] + scd[1] * dInner[1] + scd[2] * dInner[2];
+  if (std::abs(sOuter) > std::abs(scale) * tolerance) {
+    return false;
+  }
```
andiwand: should this be `sInner`?

Core/include/Acts/SpacePointFormation2/StripSpacePointBuilder.hpp
```diff
@@ -98,6 +101,128 @@ Vector2 computeVarianceZR(const GeometryContext& gctx, const Surface& surface1,
...
+
+  // Corrected position using the outer strip center and direction
+  // TODO use inner?
+  const auto& oc = sp.outerStripCenter;
+  const float sOuterNorm = sOuter / scale;
+  calibrated[0] = oc[0] + ohv[0] * sOuterNorm;
+  calibrated[1] = oc[1] + ohv[1] * sOuterNorm;
+  calibrated[2] = oc[2] + ohv[2] * sOuterNorm;
+  return true;
```
andiwand: outer vs inner

Core/include/Acts/EventData/StripSpacePointCalibrationDetails.hpp
```diff
@@ -0,0 +1,32 @@
...
+  std::array<float, 3> outerStripHalfVector;
+  std::array<float, 3> innerStripHalfVector;
+};
+
+struct StripSpacePointCalibrationDetailsDerived final {
+  std::array<float, 3> stripSeparationCrossInnerHalfVector;
+  std::array<float, 3> stripSeparationCrossOuterHalfVector;
+  std::array<float, 3> innerCrossOuterStripHalfVector;
+  // TODO use inner?
+  std::array<float, 3> outerStripCenter;
```
andiwand: outer vs inner

Core/include/Acts/EventData/StripSpacePointCalibrationDetails.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+#pragma once
+
+#include <array>
+
+namespace Acts {
+
+struct StripSpacePointCalibrationDetails final {
+  // TODO use inner?
+  std::array<float, 3> outerStripCenter;
```
andiwand: outer vs inner

Core/include/Acts/SpacePointFormation2/StripSpacePointBuilder.hpp
```diff
@@ -98,6 +101,128 @@ Vector2 computeVarianceZR(const GeometryContext& gctx, const Surface& surface1,
...
+
+  // sInner = stripSeparation dot (outerStripHalfVector cross direction)
+  // Check if direction is inside the inner detector element
+  // TODO should this rather use `sCrossIhv`?
+  const float sInner = direction[0] * sCrossOhv[0] +
+                       direction[1] * sCrossOhv[1] +
+                       direction[2] * sCrossOhv[2];
+  if (std::abs(sInner) > std::abs(scale) * tolerance) {
+    return false;
+  }
```
andiwand: should this rather use `sCrossIhv`?

Core/include/Acts/SpacePointFormation2/StripSpacePointBuilder.hpp
```diff
@@ -98,6 +101,128 @@ Vector2 computeVarianceZR(const GeometryContext& gctx, const Surface& surface1,
...
+
+  // sOuter = stripSeparation dot (innerStripHalfVector cross direction)
+  // Check if direction is inside the outer detector element
+  // TODO should this rather use `sCrossOhv`?
+  const float sOuter = direction[0] * sCrossIhv[0] +
+                       direction[1] * sCrossIhv[1] +
+                       direction[2] * sCrossIhv[2];
+  if (std::abs(sOuter) > std::abs(scale) * tolerance) {
+    return false;
+  }
```
andiwand: should this rather use `sCrossOhv`?

Core/include/Acts/SpacePointFormation2/StripSpacePointBuilder.hpp
```diff
@@ -98,6 +101,128 @@ Vector2 computeVarianceZR(const GeometryContext& gctx, const Surface& surface1,
...
+
+  // Corrected position using the outer strip center and direction
+  // TODO use inner?
+  const float sOuterNorm = sOuter / scale;
+  calibrated[0] = oc[0] + ohv[0] * sOuterNorm;
+  calibrated[1] = oc[1] + ohv[1] * sOuterNorm;
+  calibrated[2] = oc[2] + ohv[2] * sOuterNorm;
+  return true;
```
andiwand: outer vs inner

andiwand (commented):
A few things I was not sure about while trying to understand the code

mvessell96:
> > Elevates the pre-computable strip space point information out of the seeding code. This way the computation only needs to happen once, but the user is now responsible for the computation. Therefore this is a breaking change.
> > Also allows to deduplicate the `stripCoordinateCheck` again. In a followup PR I plan to elevate this this function and rename it to something like `stripSpacePointCalibration` to make it useable in the parameter estimation.
> > To be seen what performance difference this makes for ITk strip seeding
> 
> After fighting the performance I get a marginal 2% improvement in Athena with the SPOT script. I am back to two functions again but elevated them now to public functions in `Core/SpacePointerFormation2/StripSpacePointBuilder.hpp`.
> 
> With the single function and the 5 calibration vectors attached to the space point I saw about ~10% slowdown. My interpretation is that all of these are in the cache and loading them from there is slower than recalculating them from 4 calibration vectors. I think this will highly depend on the CPU and scenario but lets not go down that road.
> 
> I also experimented with using Eigen to various degrees but the more we use it the more it slows down...
> 
> What is left from this is mainly a refactor and some breaking renames that I believe overall improve things. Next I would explore a bit how to best estimate the seed params using the calibration.

Yes I tried something similar (ok but much less elegant) originally and also came to the conclusion that loading from cache was slower than recalculating them but also was not sure how much this result could be trusted for the same CPU/scenario reason so left it unmentioned ;) This is ultimately why I settled on the messy two function solution which I wasn't exactly happy with but was the fastest implementation out of the stuff I threw at the wall

andiwand:
I also tried using only the function with the pre-calculated struct and it seems to produce the same timing. Maybe we can deduplicate to this variant? Or did it behave differently for you?

There are a few things I was not sure about and left a few comments above, maybe you have an opinion on these @mvessell96.

If we could reduce the information necessary for the calibration it might go faster but I didn't have a great idea about it.

mvessell96:
> I also tried using only the function with the pre-calculated struct and it seems to produce the same timing. Maybe we can deduplicate to this variant? Or did it behave differently for you?
> 
> There are a few things I was not sure about and left a few comments above, maybe you have an opinion on these @mvessell96.
> 
> If we could reduce the information necessary for the calibration it might go faster but I didn't have a great idea about it.

I think it gave like O(5-8%) which was a much bigger number when I put this in, I think likely some of the other recent improvements have eaten most of the gain? Since we are likely getting to the bad top sps less often

andiwand:
Ok good to know! Then lets just keep it - doesn't hurt

andiwand:
I coded up a little toy visualization and the current implementation looks indeed correct https://github.com/andiwand/cern-scripts/blob/main/scripts/2026-05-10_strip-space-point-calibration.py

image

benjaminhuth (approved):
LGTM!

PR #5433, opened 2026-05-11
feat: mp-ii thirdparty installer 

goblirsc:
Add millepede-ii thirdparty component, update CI 

--- END COMMIT MESSAGE ---

Cleaned-up version of #5372. Additionally update CI chain to pull in Mille/-pedeII via Spack. Only use one common Plugin for Mille and MP2 - keep the name for now, will later rename to `Millepede-II` once related PR are merged.


goblirsc:
Update to run scripts and plotting utilities not forgotten - next PR :)

andiwand (dismissed):
LGTM 👍 

there were some issues in the CI, some of them looked temporary. let's see what it does after re-run

goblirsc:
Odd - seems the CI is providing MP2-V05-01-00, but the requested 05-01-01 is available on SPack.

goblirsc:
Enabling MP-II in the CI uncovered a complaint by Eigen about using non-recommended includes in the Helpers of the MP2 plugin - should now hopefully be fixed :crossed_fingers:

Plugins/Mille/src/ActsToMille.cpp
andiwand:
```suggestion
    if (isMeasurementOnSurface) {
      ++iMeas;
    }
```

andiwand:
```suggestion
      if (isMeasurementOnSurface) {
        targetState.projectionMatrix(iMeas, localIndex) =
          measurement.localDerivatives[iLoc];
      }
```

goblirsc:
:fire_engine:  :fire_engine:  :fire_engine:

goblirsc:
The "errors" in the `CI Bridge` tests are fixed in MillepedeII V05-01-02. This is not yet on Spack, so this PR still provides the older version. As they are just style comments and no real issues, is there a way of temporarily silencing them?

goblirsc:
To share this with the poor souls that wonder about the large number of commits for this trivial-looking change: 
The ACTS CI is configured to treat warnings as errors.

And `gfortran` receives some flags via the CI preset that make it emit warnings on every unmatched apostrophe... in **comment lines**. 

Writing `!!  let's write something completely harmless` as a comment would cause a CI error. 
As do several such sentences and "prime" symbols in the MP2 codebase.

A hacky local fix consisting of patching in-place the MP2 build options to mask the `Werror` coming from the CI preset seems to be foiled by some subtle caching in the pipelines.

goblirsc:
Conclusion: 
- The `CI Bridge` compilation failure is not reproducible locally, but fixed with the latest Millepede version 05-01-02 
- This version is supported in the `development` branch of `Spack` already 
- will wait for this to propagate to the ACTS CI (which will un-break the now-failing tests) rather than piling on more hacks to fix spurious CI issues.

CMakePresets.json
```diff
@@ -78,7 +79,9 @@
                 "ACTS_BUILD_EXAMPLES_FASTJET": "ON",
                 "ACTS_USE_SYSTEM_NLOHMANN_JSON": "ON",
                 "ACTS_USE_SYSTEM_PYBIND11": "ON",
-                "ACTS_USE_SYSTEM_GBL": "OFF"
+                "ACTS_USE_SYSTEM_GBL": "OFF",
```
paulgessinger, line 95: We could take GBL from spack as well in principle. What do you think?

goblirsc: Sure. Let's do this in the next PR where Christian will populate the GBL plugin?

andiwand (dismissed):

andiwand:
there seems to be a new CI issue with [`CI Bridge / lcg_109: [alma9, clang19]`](https://github.com/acts-project/acts/pull/5433/checks?check_run_id=79687660094) @goblirsc @paulgessinger

paulgessinger:
@andiwand yes and I don't understand what it is to be honest. It could be related to the updated dependencies package? @goblirsc do you want to try splitting that off into a separate pr?

goblirsc:
> @andiwand yes and I don't understand what it is to be honest. It could be related to the updated dependencies package? @goblirsc do you want to try splitting that off into a separate pr?

trying in https://github.com/acts-project/acts/pull/5550 :crossed_fingers:

goblirsc:
> > @andiwand yes and I don't understand what it is to be honest. It could be related to the updated dependencies package? @goblirsc do you want to try splitting that off into a separate pr?
> 
> trying in #5550 🤞

So it seems: 
- the errors flagged in the pipeline overview are also there when only touching the deps. 
- However, the #5550 pipeline contines after the errors
- Instead, looking into the detailed job on the present PR, we get another error 
```
$ cd src
$ pytest -rsfExw --durations=10 -k "not gnn and not plot" -v -n1
ImportError while loading conftest '/builds/acts/ci-bridge/src/Python/conftest.py'.
Python/conftest.py:2: in <module>
    import acts
E   ModuleNotFoundError: No module named 'acts' 
```
- and this appears only here, and not in #5550  - the true culprit? 
@paulgessinger any idea what this could signify?

andiwand:
ah I only now saw this https://github.com/acts-project/acts/pull/5433#issuecomment-4631757812

that is even more mysterious... that usually means ACTS is not build properly. for example python bindings are missing or built against a different python version. could any of the MPII stuff be messing with the environment? @goblirsc

goblirsc:
> ah I only now saw this [#5433 (comment)](https://github.com/acts-project/acts/pull/5433#issuecomment-4631757812)
> 
> that is even more mysterious... that usually means ACTS is not build properly. for example python bindings are missing or built against a different python version. could any of the MPII stuff be messing with the environment? @goblirsc

I don't think any MPII stuff touches the python env - it is a standalone program in Fortran. There are a handful of utility scripts in python shipped with it, but those should just get copied over without touching the env. Also, the error only happens for one specific OS/compiler combo in the CI - which makes this even weirder :exploding_head:

paulgessinger:
Ok so I think the actual failure here is:

- CMake is configured with `ACTS_BUILD_PYTHON_BINDINGS="ON"`
- We only build the `ActsCore` library
- We try to run the python unit tests, but the python bindings were never build.

@andiwand could this be related to your build simplification?

andiwand:
I also had another look yesterday and saw that we only build core. in the log cmake runs a second time because MP-II changes `CMAKE_CXX_COMPILER` which I think it should not. the second ACTS cmake run misses somehow most of the flags, which I also don't think it should

paulgessinger:
I think what happens is that we run cmake multiple times. Once with the correct setting and then again with the core library only. Then we run the build for the second step and then the tests can't work. 

This might be triggered by the compiler changing, and I'm wondering if this could have something to do with the MP2 build needing Fortran?

Workaround could be locking the compilers in the beginning already?

This could also explains why it's only this job: maybe the other jobs don't trigger a compiler change.

paulgessinger:
@goblirsc do we need to test the MP2 code in lcg? Otherwise we could just rely on the main tests outside of lcg

andiwand:
only in the LCG setup we fetch MP-II from source. I think it is good to cover this

this is the part in the MP-II cmake which causes the compiler flag to change

```
# Discover compilers from PATH when not already specified (e.g. via -DCMAKE_Fortran_COMPILER
# or the FC/CC/CXX environment variables). Defaults to gfortran/gcc/g++ for standalone
# cmake invocations.
if(NOT DEFINED CACHE{CMAKE_Fortran_COMPILER} AND NOT DEFINED ENV{FC})
    find_program(gf_loc gfortran)
    set(CMAKE_Fortran_COMPILER ${gf_loc})
endif()
if(NOT DEFINED CACHE{CMAKE_C_COMPILER} AND NOT DEFINED ENV{CC})
    find_program(gcc_loc gcc)
    set(CMAKE_C_COMPILER ${gcc_loc})
endif()
if(NOT DEFINED CACHE{CMAKE_CXX_COMPILER} AND NOT DEFINED ENV{CXX})
    find_program(gpp_loc g++)
    set(CMAKE_CXX_COMPILER ${gpp_loc})
endif()
```

why the second cmake invocation does not result in a valid setup I am not sure. sounds like a problem on the ACTS side. in the best case we would fix both

paulgessinger:
I think you're never supposed to change  `CMAKE_CXX_COMPILER` from inside CMake. That might be our bug right there. 

What happens is that when the compiler changes the whole cache is invalidated and the build call reconfigures on the fly, but without the preset variables set. This can't work, basically. 

To me that's an anti-pattern/bug in the MP2 configuration. 

We can work around this in the ci config by setting the variables before CMake runs or the MP2 CMake drops this section.

andiwand:
yes agreed. I also think setting `CMAKE_CXX_COMPILER` is disruptive and should not be done. @goblirsc mentioned there was a reason to do this like CMake picking a wrong compiler, but I would be surprised if this cannot be handled any other way. I think we deliberately never set `CMAKE_CXX_COMPILER`, even from outside and let CMake discover the compiler with the `CXX` environment variable or from `PATH`.

We can either wait for a fix in MP-II, or patch it in ACTS, or do as @paulgessinger and set `CMAKE_CXX_COMPILER` for now.

goblirsc:
> yes agreed. I also think setting `CMAKE_CXX_COMPILER` is disruptive and should not be done. @goblirsc mentioned there was a reason to do this like CMake picking a wrong compiler, but I would be surprised if this cannot be handled any other way. I think we deliberately never set `CMAKE_CXX_COMPILER`, even from outside and let CMake discover the compiler with the `CXX` environment variable or from `PATH`.
> 
> We can either wait for a fix in MP-II, or patch it in ACTS, or do as @paulgessinger and set `CMAKE_CXX_COMPILER` for now.

Hi, 
at this stage, if updating the CI or temporarily disabling MP2 for the lcg tests is an option, i would prefer that. Otherwise we have to tag a third release of MP2 just to satisfy the CI. Setting CMAKE_CXX_COMPILER was the recommended way I found at the time of moving the package to CMake to ensure the CMake configuration does not incorrectly pick up the system compiler when an lcg release is set up. I can experiment with alternatives, but would need to test such changes with the other client codes first. So this could be updated in the future beyond this PR if you agree.

goblirsc:
> > yes agreed. I also think setting `CMAKE_CXX_COMPILER` is disruptive and should not be done. @goblirsc mentioned there was a reason to do this like CMake picking a wrong compiler, but I would be surprised if this cannot be handled any other way. I think we deliberately never set `CMAKE_CXX_COMPILER`, even from outside and let CMake discover the compiler with the `CXX` environment variable or from `PATH`.
> > We can either wait for a fix in MP-II, or patch it in ACTS, or do as @paulgessinger and set `CMAKE_CXX_COMPILER` for now.
> 
> Hi, at this stage, if updating the CI or temporarily disabling MP2 for the lcg tests is an option, i would prefer that. Otherwise we have to tag a third release of MP2 just to satisfy the CI. Setting CMAKE_CXX_COMPILER was the recommended way I found at the time of moving the package to CMake to ensure the CMake configuration does not incorrectly pick up the system compiler when an lcg release is set up. I can experiment with alternatives, but would need to test such changes with the other client codes first. So this could be updated in the future beyond this PR if you agree.

... and the working day after writing this, I updated the packages :D 

In testing, it seems the "system compiler issue" is gone, likely fixed in one of the config updates since it was initially seen... crossing fingers it will stay that way! 

Now running a CI with the new package versions. This will currently **fail** with the Spack pipelines - those will need another dependencies tag, after I propagate the versions to Spack.  But one thing at a time :)

paulgessinger (dismissed):

andiwand:
@goblirsc CI is failing because https://gitlab.desy.de/millepede/mille/-/archive/V01-00-05/mille-V01-00-05.tar.gz is not accessible

goblirsc:
> https://gitlab.desy.de/millepede/mille/-/archive/V01-00-05/mille-V01-00-05.tar.gz

Hi, this is worrying - I can access this from within DESY, but no longer from without. 

_Update:_
DESY gitlab is expected to be only partially available until the end of the week, due to "high load from external accesses".
From testing, web browser access is fine, but `wget` currently seems to time out outside the DESY network.

Let's retry the CI in a few days when the situation is back to normal!

paulgessinger:
Hm this seems to be a more fundamental problem then? If external access is disabled like this, DESY GitLab is not the right location to host this, since the software is meant to be used outside of DESY. 

I would strongly suggest moving MP2 to GitHub at this point.

goblirsc:
> Hm this seems to be a more fundamental problem then? If external access is disabled like this, DESY GitLab is not the right location to host this, since the software is meant to be used outside of DESY.
> 
> I would strongly suggest moving MP2 to GitHub at this point.

It seems to be a temporary emergency measure, similar to the issue CERN had recently. 
GitHub probably has equal chanced to be hit by AI scrapers? 
Let's see. If this does not resolve by the weekend, I can mirror to github.

paulgessinger (approved):

PR #5434, opened 2026-05-12
`feat`: detray geometry conversion

paradajzblond:
Enabling the build of Detray examples and plugin independent of Traccc.
Adding geometry conversion algorithm

### DIFF
--- a/Examples/Algorithms/Detray/include/ActsExamples/Detray/ActsToDetrayDetectorAlg.hpp
+++ b/Examples/Algorithms/Detray/include/ActsExamples/Detray/ActsToDetrayDetectorAlg.hpp
@@ -1,0 +1,54 @@
+#pragma once
+#include "Acts/Geometry/GeometryContext.hpp"
+#include "Acts/Geometry/GeometryIdentifier.hpp"
+#include "Acts/Geometry/TrackingGeometry.hpp"
+#include "Acts/Utilities/Logger.hpp"
+#include "ActsExamples/Framework/DataHandle.hpp"
+#include "ActsExamples/Framework/IAlgorithm.hpp"
+#include "ActsExamples/Framework/ProcessCode.hpp"
+#include "ActsPlugins/Detray/DetrayConversionUtils.hpp"
+
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+#include <vecmem/memory/host_memory_resource.hpp>
+#include <detray/detectors/odd_metadata.hpp>
+
+namespace ActsExamples {
+
+/// Converts Acts TrackingGeometry → detray detector in memory,
+/// and writes the detray↔Acts geometry ID map to the whiteboard.
+class ActsToDetrayDetectorAlg final : public IAlgorithm {
+ public:
+  struct Config {
+    /// The Acts tracking geometry to convert
+    std::shared_ptr<const Acts::TrackingGeometry> trackingGeometry;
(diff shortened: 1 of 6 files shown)


Examples/Algorithms/Detray/include/ActsExamples/Detray/ActsToDetrayDetectorAlg.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <unordered_map>
+
+#include <vecmem/memory/host_memory_resource.hpp>
+#include <detray/detectors/odd_metadata.hpp>
+
+namespace ActsExamples {
+
+/// Converts Acts TrackingGeometry → detray detector in memory,
+/// and writes the detray↔Acts geometry ID map to the whiteboard.
+class ActsToDetrayDetectorAlg final : public IAlgorithm {
```
paulgessinger, line 22: This should not be an algorithm. You can orchestrate this directly from Python and use the config to propagate this static info

paulgessinger (commented):
Do we need these as separate plugins? What's the use case of building only detray without traccc?

Also: the geometry should not flow through the whiteboard. The mapping and detray geometry can be wrapped in a python bound struct and passed through the config directly into algorithms that need it.

paradajzblond:
> Do we need these as separate plugins? What's the use case of building only detray without traccc?
> 
> Also: the geometry should not flow through the whiteboard. The mapping and detray geometry can be wrapped in a python bound struct and passed through the config directly into algorithms that need it.

well exactly for this type of geometry conversions and testing, since we will be stuck with files for a little while longer.

Examples/Algorithms/Detray/include/ActsExamples/Detray/ActsToDetrayDetectorAlg.hpp
asalzburger: Hey @paulgessinger - as I wrote to @paradajzblond, I have picked up her branch and will work on it a bit. 

This won't be an algorithm anymore, but a conversion function which we can call from python.

asalzburger:
> > Do we need these as separate plugins? What's the use case of building only detray without traccc?
> > Also: the geometry should not flow through the whiteboard. The mapping and detray geometry can be wrapped in a python bound struct and passed through the config directly into algorithms that need it.
> 
> well exactly for this type of geometry conversions and testing, since we will be stuck with files for a little while longer.

Eventually, the DETRAY plugin will disappear and that code should go into the Detray folder we have already after detray import, I assume.

paulgessinger:
I'd say this is longer term but yes, @asalzburger.

@paradajzblond Maybe I'm missing something, but why does that require two separately toggled pieces of code? You would be able to write the code exactly as is with the only difference being that you build traccc as well. Does that not make sense? My worry is that we already have a huge number of flags and I don't see what the use case for building one and not the other is. Traccc always needs detray anyway, so the only remaining combination is detray but not traccc. But what do you do with detray that would be better served without building traccc alongside it?

paradajzblond:
@paulgessinger you can imagine a wonderful world in the future where we always use detray geometry description regardless if the track reconstruction happens on cpu or device. as andreas pointed, this MR is solely to show how i did the conversions, we can close it now if you wish

paulgessinger:
I think it would be cool to surface this and start testing the conversion. I'm all for that!

asalzburger:
Superseded by #5449

PR #5435, opened 2026-05-12
ci: Rewrite physmon using Snakemake

paulgessinger:


CI/physmon/Snakefile
```diff
@@ -0,0 +1,860 @@
...
+# for every rule's shell, appended so they cannot shadow ACTS/ROOT. The rules
+# run from the repo root, so the relative paths resolve.
+#
+# This replaces snakemake's default prefix, so keep its `set -euo pipefail`.
+# Braces are doubled: snakemake runs str.format() over the prefix too, so a
+# literal shell ${...} must be written as ${{...}}.
+shell.prefix(
+    'set -euo pipefail; '
+    'export PYTHONPATH="${{PYTHONPATH:+$PYTHONPATH:}}CI/physmon:Examples/Scripts/Python"; '
+)
```
andiwand, line 62: alternatively we could add / generate an activation script that stacks `this_acts_withdeps.sh` and the python path changes and source it in `wrap.sh`

CI/physmon/config/physmon.yaml
```diff
@@ -0,0 +1,249 @@
...
+#                      `SeedingAlgorithm` enum member names so the script can
+#                      resolve them with getattr(); outputs land in
+#                      <slug>/<variant>/.
+#   trackfinding:      expand the trackfinding pack for this generator.
+#                      `label` (or `label_fmt` with {variant_label} when the
+#                      generator has variants), optional `extras`, and
+#                      `vertexing_config` if any vertexing extra is used.
+# ---------------------------------------------------------------------------
+
+generators:
```
andiwand: generator might be a bit overloaded in context of particle gun and pythia8. what do you think about `workflows` or `producers` or `jobs` as alternative? (might very well be a me issue)

andiwand (commented):
had a first high altitude look - looks quite good to me already. I think we will benefit from having a rigid framework around the physmon workflows

CI/physmon/config/physmon.yaml
paulgessinger: I like "producer"

CI/physmon/Snakefile
paulgessinger: The only downside is that snakemake then has to assume where the build is located or be told in the CI run.

CI/physmon/config/physmon.yaml
paulgessinger: changed

CI/physmon/Snakefile
paulgessinger: It does assume now.

PR #5436, opened 2026-05-12
refactor: Revise track parameter estimation implementation

andiwand:
- modularize implementation so it can be reused for other estimation functions
  - planning to add an `estimateTrackTangentsFromSeed` function
  - planning to add back an estimation function for N space points
- account for the arc length of the circle segment vs straight line estimate for `dzds`

--- END COMMIT MESSAGE ---

cc @jburzy @mvessell96 might make a difference for the LRT track param estimation


Core/src/Seeding/EstimateTrackParamsFromSeed.cpp
```diff
@@ -8,21 +8,23 @@
...
-Acts::FreeVector Acts::estimateTrackParamsFromSeed(const Vector3& sp0,
-                                                   const Vector3& sp1,
-                                                   const Vector3& sp2,
-                                                   const Vector3& bField) {
-  return estimateTrackParamsFromSeed(sp0, 0, sp1, sp2, bField);
+namespace Acts {
+namespace {
+
+double sinc(double x) {
+  if (std::abs(x) < 1e-10) {
```
AJPfleger: The Taylor expansion has an error of $\mathcal{O}(x^4)$. I think, with $\epsilon \approx 10^{-16}$, we could (should?) go up to `1e-4`.

AJPfleger: Also with `1e-10`, it always goes to `1 - 1e-20`, which is exactly `1` as `double`.

andiwand: yes very good point 👍 let me change that

andiwand: I experimented a bit more with this and it looks like `sin(x) ~ x` works very well for small numbers. combined with your argument I think I will just fix `x~0` at the numerical limit

AJPfleger (dismissed):

AJPfleger (dismissed):

acts-project-service:
## :red_circle: Athena integration test results [946222254848e819f358b399891f74775a012476]
Please investigate the [build job](https://gitlab.cern.ch/acts/acts-athena-ci/-/jobs/74897774) for the [pipeline](https://gitlab.cern.ch/acts/acts-athena-ci/-/pipelines/14768053)!

AJPfleger (approved):

PR #5437, opened 2026-05-12
feat: Provide access to tangents from seed estimation function

andiwand:
Adds outputs for tangents at each space point to `estimateTrackParamsFromSeed`. It now allows to fit three space points on a helix and estimate the trajectory tangents at the space points. This is useful for strip space point calibration where one needs the trajectory direction as an input to derive calibrated space points, which in turn then get used to estimate calibrated seed parameters.

Also adds Python bindings that were used to test the output of this new function.

--- END COMMIT MESSAGE ---

image

blocked by
- https://github.com/acts-project/acts/pull/5436

cc @jburzy @mvessell96 relevant for LRT


Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp
```diff
@@ -181,6 +181,24 @@ BoundMatrix estimateTrackParamCovariance(
...
+/// @param sp1 is the middle space point
+/// @param sp2 is the top space point
+/// @param bField is the magnetic field vector
+/// @param tangent0 is the output tangent at the bottom space point
+/// @param tangent1 is the output tangent at the middle space point
+/// @param tangent2 is the output tangent at the top space point
+void estimateTrackTangentsFromSeed(const Vector3& sp0, const Vector3& sp1,
+                                   const Vector3& sp2, const Vector3& bField,
+                                   Vector3& tangent0, Vector3& tangent1,
+                                   Vector3& tangent2);
```
andiwand: alternatively we could
- return a `std::tuple` instead of using output parameters
- merge this into the estimation function having output pointers defaulted with `nullptr` and only compute and output the tangents if requested

benjaminhuth (approved):
LGTM!

PR #5444, opened 2026-05-13
feat: Elevate strip space point calibration to public API

andiwand:
Brings the currently hidden strip space point calibration from the seeding code to public API. There are two layers to it because the critically depend on its performance during strip seeding. One is implemented on top of `std::array` with all the linalg rolled out into primitive arithmetic C++ expressions. The public ones uses Eigen and wires internally to the accelerated implementation.

This can then be used together with https://github.com/acts-project/acts/pull/5437 to calibrate strip space points for track parameter estimation.

--- END COMMIT MESSAGE ---

cc @jburzy @mvessell96 relevant for LRT


Core/include/Acts/EventData/StripSpacePointCalibrationDetails.hpp
```diff
@@ -0,0 +1,47 @@
...
+#pragma once
+
+#include <array>
+
+namespace Acts {
+
+/// Collection of outer strip space point details, used for the calibration.
+struct OuterStripSpacePointCalibrationDetails final {
+  /// Center of the outer strip.
+  std::array<float, 3> outerCenter;
```
benjaminhuth, line 18: Hmm so is there a good reason to keep the `std::array<float, 3>` instead of a `Vector3`?

Core/include/Acts/SpacePointFormation2/detail/StripSpacePointCalibrationImpl.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+namespace Acts::detail {
+
+inline OuterStripSpacePointCalibrationDetailsDerived
+deriveOuterStripSpacePointCalibrationDetails(std::span<const float, 3> ihv,
+                                             std::span<const float, 3> ohv,
+                                             std::span<const float, 3> iosv,
+                                             std::span<const float, 3> oc) {
+  OuterStripSpacePointCalibrationDetailsDerived result{};
+  result.innerCrossOuterHalfVector = {ihv[1] * ohv[2] - ihv[2] * ohv[1],
```
benjaminhuth: also this could be implemented with our cross helpers for Eigen types I think

benjaminhuth (commented):
Hmm I think we should move to the Eigen-based math for consistency and readability here.

benjaminhuth (dismissed):
LGTM. Just a minor comment: Naming wise, I think `spanMap`, `spanDot`, ... would fit better, because `cmath` might be a bit misleading (no `#include <cmath>` actually).

benjaminhuth (approved):
LGTM!

PR #5445, opened 2026-05-13
fix: Set strip half length vector in `SpacePointMaker`

andiwand:
We actually depend on the strip half length down stream in the strip space point seeding

--- END COMMIT MESSAGE ---


andiwand:
cc @benjaminhuth @murnanedaniel

benjaminhuth (approved):
LGTM!

PR #5446, opened 2026-05-13
feat(detray): Add detray github nightly build

niermann999:
Move some of the detray github workflows to a dedicated nightly build/test. These include the algebra plugins and HIP, SYCL workflows. Also fixed the label maker to extend to the correct github workflow files and fixed a few warnings that blocked the nightly 

--- END COMMIT MESSAGE ---


niermann999:
For testing, the nightly builds trigger on the PR for now

PR #5448, opened 2026-05-13
feat: remove unused nodes before each step in GNN workflow

benjaminhuth:


mimodekjaer (dismissed):
Seems okay.

The only real comment is that the clang formatting should be fixed. :)

Python/Examples/tests/test_examples.py
```diff
@@ -950,6 +950,65 @@ def test_gnn_metric_learning(tmp_path, trk_geo, field, assert_root_hash, hardwar
...
+                outputDir=out,
+                shrinkNodes=shrink,
+                s=s,
+            )
+        del s  # Ensure ROOT TFile is closed (happens in sequencer destructor)
+        output_dirs[shrink] = out
+
+    root_files = ["performance_finding_gnn.root", "ntuple_finding_gnn.root"]
+    for fname in root_files:
+        h_no_shrink = hash_root_file(output_dirs[False] / fname)
```
mimodekjaer, line 1012: I think it would be good to check if they exists before accessing the files.

mimodekjaer (changes requested):
I think all in all, that it is good :), but I think it would be good to check that the root files are generated, before accessing them.

Python/Examples/tests/test_examples.py
benjaminhuth: Hmm I hope this would be clear from the error message in `hash_root_file`...

mimodekjaer: It should give a file not found from uproot. Most of the reason I added this was because I noticed that most of the other tests does this does this before running the `assert_hash_root`, which runs the `hash_root_file` on the file.

But I do not have any strong opinion on this.

mimodekjaer (approved):
I think it is good to go.

I dont have strong opinions on what error that should be thrown if the GNN fails. So I think this is fine.

PR #5449, opened 2026-05-13
feat: detray geometry converter (and ODD tests)

asalzburger:
This PR is largely based on the work by @paradajzblond and supersedes:

https://github.com/acts-project/acts/pull/5434

*Here's what it does*:

**Geometry** 

- adds a new `DetrayGeometryConverter` with a conversion method that is templated on memory resource and `detray::detector_metadata` (hence will be useful for all detray detectors)
- adds a `Vecmem` python binding to define in one place `pmr::memory_resource`, `pmr::host_memory_resource`
- adds a `convertODD()` python method that specifies host memory resource, ODD-metadata and converts

**Propagation**

- re-established the `DetrayPropagator` this time using the detray internal validation methods

--- END COMMIT MESSAGE ---

Python snipped showing the conversion (tested):

@paradajzblond @paulgessinger 

```python
        host_memory = acts.vecmem.HostMemoryResource()
        detrayGeometry = acts.detray.convertODD(
            host_memory,
            gContext,
            trackingGeometry,
            beampipeVolumeName="BeamPipe",
            logLevel=acts.logging.INFO,
        )
```


asalzburger:
How can I make @paradajzblond co-author of this PR ?

Plugins/Detray/include/ActsPlugins/Detray/DetrayGeometryConverter.hpp
```diff
@@ -0,0 +1,158 @@
...
+                            this](const Acts::TrackingVolume& volume) {
+      ACTS_DEBUG("Found volume: " << volume.volumeName());
+      if (volume.volumeName() == beampipeVolumeName) {
+        convCfg.beampipeVolume = &volume;
+        ACTS_INFO("Found beampipe volume: " << volume.volumeName());
+      }
+    });
+
+    // Find beampipe volume
+    trackingGeometry.apply(
```
niermann999: Is this not the same as above (only that the logging is missing)?

Python/Plugins/src/Detray.cpp
```diff
@@ -25,7 +27,30 @@ PYBIND11_MODULE(ActsPluginsPythonBindingsDetray, detray) {
   using namespace ActsPlugins;
 
   {
-    py::class_<DetrayHostDetector, std::shared_ptr<DetrayHostDetector>>(
-        detray, "detray_detector");
+    using DetrayMetaDataODD = detray::odd_metadata<detray::array<double>>;
```
niermann999:
```suggestion
    using DetrayMetaDataODD = detray::odd_metadata<detray::array<float>>;
```

Plugins/Detray/include/ActsPlugins/Detray/DetrayGeometryConverter.hpp
```diff
@@ -0,0 +1,158 @@
...
+  /// 4. It checks the consistency of the built detray detector using
+  /// detray::detail::check_consistency.
+  /// 5. It constructs a mapping from detray surface identifiers to ACTS
+  /// GeometryIdentifiers for all sensitive surfaces in the detray detector.
+  /// 6. It returns a tuple containing the built detray detector and the
+  /// detray→ACTS mapping.
+  ///
+  /// @return A tuple containing the built detray detector and the detray→ACTS mapping.
+  template <typename metadata_t>
+  detray::detector<metadata_t> toDetray(
```
paulgessinger: I suppose this could be a free function since the only class member is the logger?

Plugins/Detray/include/ActsPlugins/Detray/DetrayGeometryConverter.hpp
asalzburger: Oh yes, removed one.

Python/Plugins/src/Detray.cpp
asalzburger: Fixed.

Plugins/Detray/include/ActsPlugins/Detray/DetrayGeometryConverter.hpp
asalzburger: changed to free function.

asalzburger:
Ready to go for review from my side @niermann999

Plugins/Detray/include/ActsPlugins/Detray/DetrayGeometryConverter.hpp
```diff
@@ -0,0 +1,148 @@
...
+      from_payload<detector_t>(detectorBuilder,
+                               std::move(*payloads.materialGrids));
+
+  detray::io::surface_grid_reader<typename detector_t::surface_type,
+                                  std::integral_constant<std::size_t, 0>,
+                                  std::integral_constant<std::size_t, 2>>::
+      template from_payload<detector_t>(detectorBuilder,
+                                        *payloads.surfaceGrids);
+
+  return std::make_shared<detector_t>(detectorBuilder.build(mr));
```
niermann999, line 107: Does it do a consistency check as claimed in the comment? If so, it also should build the name map here, so that error messages contain the volume names instead of indices

Plugins/Detray/include/ActsPlugins/Detray/DetrayGeometryConverter.hpp
```diff
@@ -0,0 +1,148 @@
...
+  // ───────────────────────────────────
+  std::unordered_map<std::uint64_t, Acts::GeometryIdentifier> detrayToActsMap;
+
+  for (const auto& surface : detrayDetector.surfaces()) {
+    // surface.source is the Acts GeometryIdentifier encoded as uint64
+    const Acts::GeometryIdentifier actsId(surface.source);
+    if (actsId.sensitive() == 0) {
+      continue;  // skip portals and passives
+    }
+    detrayToActsMap[surface.identifier().value()] = actsId;
```
niermann999: Why not use the `detray::geometry::identifier` type directly? It *should* be hashable

Examples/Algorithms/Detray/include/ActsExamples/Detray/DetrayPropagator.hpp
```diff
@@ -24,49 +23,28 @@
...
-    /// The field (if any)
-    field_t field = field_t();
-  };
-
-  /// Create a DetrayPropagator
+  /// Create a DetrayConcretePropagator - this is for testing purposes only,
+  /// it runs without actor chain.
   ///
-  /// @param cfg configuration struct
+  /// @param dGeometry configuration struct
```
niermann999: This is the detray geometry?

Examples/Algorithms/Detray/include/ActsExamples/Detray/DetrayPropagator.hpp
```diff
@@ -85,101 +63,106 @@ class DetrayPropagator : public PropagatorInterface {
...
 
     // Now follow that ray with the same track and check, if we find
     // the same volumes and distances along the way
-    detray::free_track_parameters<DetrayAlgebraType> track(
-        {position.x(), position.y(), position.z()}, 0.f,
-        {direction.x(), direction.y(), direction.z()},
-        startParameters.charge());
+    using DAlgebra = typename detector_t::algebra_type;
+
+    detray::free_track_parameters<DAlgebra> track(
```
niermann999, line 75: Should the detray propagation not also start from bound track parameters?

Examples/Algorithms/Detray/include/ActsExamples/Detray/DetrayPropagator.hpp
```diff
@@ -85,101 +63,106 @@ class DetrayPropagator : public PropagatorInterface {
...
 
     // Return material
     RecordedMaterial recordedMaterial;
     PropagationSummary summary(startParameters);
 
-    if (!m_cfg.sterile) {
-      /// Aggregation of multiple inspectors
-      using DetrayInspector = detray::aggregate_inspector<DetrayObjectTracer>;
+    using DPropagationCfg = detray::propagation::config;
+    DPropagationCfg dCfg{};
```
niermann999: This type alias does not seem to be used anywhere else
```suggestion
    detray::propagation::config dCfg{};
```

Examples/Algorithms/Detray/include/ActsExamples/Detray/DetrayPropagator.hpp
```diff
@@ -85,101 +63,106 @@ class DetrayPropagator : public PropagatorInterface {
...
-        propagator.propagate(propagation);
+        typename DPropagator::state propagation(
+            track, m_field, *m_detrayGeometry, dCfg.context);
+        DPropagator propagator(dCfg);
+        propagateAndRecord<DMaterialTracer, DObjectTracer>(
+            propagation, propagator, summary, recordedMaterial);
       }
     }
 
+    // Return empty results
```
niermann999: This is not necessarily empty?

Examples/Scripts/Python/propagation_test_odd.py
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env python3
```
niermann999, line 1: Does this script do anything detray related, yet?

Plugins/Detray/include/ActsPlugins/Detray/DetrayGeometryConverter.hpp
asalzburger: Actually, I want to add later on that you can call the consistency checks on the object itself to decouple building from validation, I will update the comment.

asalzburger: I also updated the docu regarding the map creation (which I also encapsulated from Neža's original code)

Plugins/Detray/include/ActsPlugins/Detray/DetrayGeometryConverter.hpp
asalzburger: Don't know, that is from Neža's code, probably because we wanted to write it out to json?
I will change and if someone then needs it serialized, they can do it.

asalzburger: Fixed.

Examples/Algorithms/Detray/include/ActsExamples/Detray/DetrayPropagator.hpp
asalzburger: Doesn't matter here, this is not a Propagator that can be used anywhere else than in the PropagationAlgorithm, if we want an Acts::Propagator with Detray backend, that's more work, but would be cool - and that one would need this, indeed.

Examples/Algorithms/Detray/include/ActsExamples/Detray/DetrayPropagator.hpp
asalzburger: Yup.

Examples/Algorithms/Detray/include/ActsExamples/Detray/DetrayPropagator.hpp
asalzburger: Fixed.

Examples/Algorithms/Detray/include/ActsExamples/Detray/DetrayPropagator.hpp
asalzburger: Yes, old behavior - comment updated.

Examples/Scripts/Python/propagation_test_odd.py
asalzburger: Will come in a follow-up pr.

asalzburger:
Comments addressed.

niermann999 (approved):

andiwand (approved):

paulgessinger:
Just saw this pop in my feed again and had a handful more thoughts:

- I suggested avoiding the class (my bad), but actually the pattern might work better if the class' config struct includes the config for the payload converter, or even the payload converter itself. This would also avoid the string->volume lookup that the function does internally (I think that should be an explicit action on the caller)
- The class doesn't have to be templated, only the conversion method does. 
- We could use explicit template specialization to push specifically the ODD template instantiation to a single .cpp file already at this point. this allows us to forward declare detray types (I specifically did this for the payload converter header) 
- I think we should avoid referring to things as "ActsSOMETHING" in this repository. The assumption should be that unqualified SOMETHING is ACTS, and everything else is qualified. 
- there's a number of leftover includes, doc strings etc
- the identifier map is a separate call and object. I would suggest that the main detray conversion method returns a class that has the detray geometry as a member as well as an already Build identifier map. The class can the. Expose a lookup function. It can also keep the original tracking geometry shared pointer, which would cleanly communicate their relationship with less external book keeping. 

Happy to discuss these (at the latest in June)

PR #5450, opened 2026-05-14
build: Detray Configuration Update, main branch (2026.05.14.)

krasznaa:
Modified how Detray would be used by the Acts code.

Removed the `thirdparty/detray` subdirectory, as we should now always pick up Detray from within the same source tree that everything else is in.

Switched `DETRAY_SETUP_<FOO>` flags off by default for all externals that Acts (also) provides.

Switched `DETRAY_SVG_DISPLAY` on by default, as the Detray plugin is making use of it.

--- END COMMIT MESSAGE ---

This was all inspired by: https://github.com/acts-project/traccc/pull/1327

Pinging @paradajzblond, @niermann999 and @stephenswat.


stephenswat:
This doesn't really solve any of our issues. All ls this is still locked
behind `ACTS_BUILD_PLUGIN_TRACCC`. It needs to be moved outside of that
conditional block to be at all useful for building traccc...

krasznaa:
P.S. Note that I did my local tests with:

```
cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_STANDARD=20 -DACTS_BUILD_PLUGIN_TRACCC=TRUE -DACTS_BUILD_PLUGIN_JSON=TRUE -DTRACCC_BUILD_CUDA=TRUE ../acts/
```

krasznaa:
> This doesn't really solve any of our issues. All ls this is still locked behind `ACTS_BUILD_PLUGIN_TRACCC`. It needs to be moved outside of that conditional block to be at all useful for building traccc...
> 

Never said that this would resolve everything. But in the past @paulgessinger was quite adamant that we shouldn't split the configuration into too many pieces. So I wanted to start with something that should not be controversial.

Then we can discuss tomorrow how we would operate while we still have 2 separate repositories.

krasznaa:
Made irrelevant by other PRs.

PR #5456, opened 2026-05-15
feat: added correct handling of strip spacepoints to GBTS

jpreston-cern:
added correct handling of strip space point loading, this was previously disabled as focus was on pixel space points

all relevant strip seeding hard coded variables are also now pulled to the `config` from `buildTheGraph`

--- END COMMIT MESSAGE ---

@timadye @andiwand


Core/include/Acts/Seeding2/GraphBasedTrackSeeder.hpp
```diff
@@ -89,6 +89,31 @@ class GraphBasedTrackSeeder {
     float minDeltaRadius = 2.0 * Acts::UnitConstants::mm;
     /// Maximum d0 impact parameter when validating edge connection triplet
     float d0Max = 3.0 * UnitConstants::mm;
+    /// Maximum difference in allowed tangent between candidate edge connection
+    float cutDPhiMax = 0.012f;
+    /// Maximum difference in allowed tangent between candidate edge connection
+    /// in LRT mode
+    float cutDPhiMaxLrt = 0.07f;
```
andiwand: if standard tracking and LRT tracking can be generalized via different values for existing cuts I would not pull this switching mechanism into the core algorithm. I would rather set different values on the client side since these are experiment dependent in any case

Core/include/Acts/Seeding2/GraphBasedTrackSeeder.hpp
```diff
@@ -191,14 +216,16 @@ class GraphBasedTrackSeeder {
...
   /// @param maxLayers Maximum number of layers
   /// @param filter Tracking filter to be applied
   /// @param options Event based options such as magnetic field strength
   /// @param outputSeeds Container with generated seeds
   void createSeeds(const SpacePointContainer2& spacePoints,
-                   const GbtsRoiDescriptor& roi, std::uint32_t maxLayers,
-                   const GbtsTrackingFilter& filter, const Options& options,
-                   SeedContainer2& outputSeeds) const;
+                   const GbtsRoiDescriptor& roi,
+                   const std::vector<bool>& isPixelLayer,
```
andiwand, line 215: note that `std::vector<bool>` is a bit special in STL but for this case it should be ok

Core/include/Acts/Seeding2/GraphBasedTrackSeeder.hpp
jpreston-cern: nice ok that makes sense, in this case would that be adding all the modified cuts to the `ActsStripGbtsSeedingToolCfg` via `kwargs.setdefault` ?

andiwand: yes I think in Athena we could split the algorithm instantiation similar to what is done for the triplet seeder

jpreston-cern: some of the values (specifically min and max z regions are pulled from the region of interest) so some of the settings dont have a pixel equivalent (in the sense that its pulled from the config), in this case, would it be best practice to just have these LRT versions in the config?

andiwand: in these cases I would model the option property with a `std::optional<T>` and document what happens if it is not set

Core/include/Acts/Seeding2/GbtsTrackingFilter.hpp
```diff
@@ -91,7 +91,7 @@ class GbtsTrackingFilter final {
     /// Maximum track curvature.
     float maxCurvature = 1e-3f / Acts::UnitConstants::mm;
     /// Maximum longitudinal impact parameter.
-    float maxZ0 = 170.0 * Acts::UnitConstants::mm;
+    float filterMaxZ0 = 170.0 * Acts::UnitConstants::mm;
```
andiwand: any particular reason to rename this?

Core/include/Acts/Seeding2/GraphBasedTrackSeeder.hpp
```diff
@@ -89,6 +89,20 @@ class GraphBasedTrackSeeder {
...
+    /// Maximum difference in allowed tangent between candidate edge connection
+    float cutDPhiMax = 0.012f;
+    /// Maximum allowed curvature tolerance for candidate edge connections
+    float cutDCurvMax = 0.001f;
+    /// Minimum z0 value, set as optionl as if in pixel mode,
+    /// The value is picked from the ROI
+    float minZ0;
+    /// Minimum z0 value, set as optionl as if in pixel mode,
+    /// The value is picked from the ROI
+    float maxZ0;
```
andiwand: can you add some defaults here?
```suggestion
    /// Minimum z0 value, set as optionl as if in pixel mode,
    /// The value is picked from the ROI
    float minZ0 = -600;
    /// Minimum z0 value, set as optionl as if in pixel mode,
    /// The value is picked from the ROI
    float maxZ0 = 600;
```

Core/src/Seeding2/GraphBasedTrackSeeder.cpp
```diff
@@ -196,23 +203,11 @@ std::vector<std::vector<GbtsNode>> GraphBasedTrackSeeder::createNodes(
...
-  // curv cut for triplets
-  const float cutDCurvMax = m_cfg.lrtMode ? 0.015f : 0.001f;
-  // tau cut for doublets and triplets
-  const float cutTauRatioMax = m_cfg.lrtMode ? 0.015f : m_cfg.tauRatioCut;
-  const float minZ0 = m_cfg.lrtMode ? -600.0f : static_cast<float>(roi.zMin());
-  const float maxZ0 = m_cfg.lrtMode ? 600.0f : static_cast<float>(roi.zMax());
-  const float minDeltaPhi = m_cfg.lrtMode ? 0.01f : 0.001f;
-
   // used to calculate Z cut on doublets
-  const float maxOuterRadius = m_cfg.lrtMode ? 1050.0f : 550.0f;
```
andiwand, line 210: I believe we should remove the `lrtMode` property if this was the only reason to have it in the first place

Core/include/Acts/Seeding2/GbtsTrackingFilter.hpp
jpreston-cern: theres a few different Z0 values depending on what area of the algorithm is using it, (geometry, tau ratio cuts and the filter all use different values), im assuming as there are different tolerances needed, maybe in future it would be good to have a single one then apply tolerences directly

Core/src/Seeding2/GraphBasedTrackSeeder.cpp
jpreston-cern: `lrtMode` in this case is also used to change the minimum length of a seed, select the type of layer connections, and apply some tolerances on the curvature, so has some scope outside of the current removed cuts (although i think in future this could be looked at more, as now we are moving towards just having all needed algorithm changes in `ActsSeedingConfig` i think this can be streamlined), but i think might be outside the scope of this PR (but can add it to the PR where i take out the rest of the magic numbers)

Core/include/Acts/Seeding2/GbtsTrackingFilter.hpp
andiwand: but I think that is generally ok. we have the same in the triplet seeder where doublet and triplet cuts with the same name can have a different value. these values do not strictly need to be synchronized as they can be leveraged for different purposes

I believe the filter context is clear with `GbtsTrackingFilter::Config::maxZ0`, no need to repeat `filter` in `GbtsTrackingFilter::Config:: filterMaxZ0`

Core/src/Seeding2/GraphBasedTrackSeeder.cpp
andiwand: I agree that is outside the scope of this PR but I think it would be good to lift this out and let the user steer the non-LRT vs LRT config

Core/include/Acts/Seeding2/GbtsTrackingFilter.hpp
jpreston-cern: nice ok no worries, I will change it back

Core/src/Seeding2/GraphBasedTrackSeeder.cpp
jpreston-cern: for sure i agree

andiwand (approved):

PR #5461, opened 2026-05-18
build(detray): Turn detray into an ACTS component

stephenswat:
This commit changes the way we build the `Detray/` directory to be in line with how `Fatras/` works. Instead of including detray as a subproject, it is a full-fledged component. Some technical debt remains which can be cleanup in a future version.

--- END COMMIT MESSAGE ---

This is a more complete replacement for #5450.


stephenswat:
Tagging @andiwand, @niermann999, @asalzburger, @benjaminhuth.

stephenswat:
Let me turn this into a draft by Joana's request until we can discuss what road we want to take for this.

PR #5462, opened 2026-05-18
build: Allow ACTS to export detray targets

stephenswat:
As it stands right now, there is no way for ACTS to export targets related to the Detray/ directory. This commit is a near-minimal change set allowing clients pulling in ACTS (including traccc) to get the detray targets.

--- END COMMIT MESSAGE ---

This is essentially the smallest, least controversial change set that allows traccc to pull in detray via ACTS. Tagging @andiwand, @niermann999, @krasznaa, @benjaminhuth, @asalzburger.


stephenswat:
This may eventually evolve into #5450 or #5461, but this keeps the existing structure as much as possible while allowing a build of traccc using ACTS and an ACTS-provided detray.

CMakeLists.txt
```diff
@@ -549,23 +555,50 @@ if(ACTS_BUILD_PLUGIN_GEANT4)
...
+    option(DETRAY_SETUP_EIGEN3 "" FALSE)
+    option(DETRAY_SETUP_ACTSVG "" FALSE)
+    option(DETRAY_SETUP_COVFIE "" FALSE)
+    option(DETRAY_SVG_DISPLAY "" ON)
+    if(ACTS_BUILD_TRACCC)
+        option(DETRAY_BUILD_TEST_UTILS "" ON)
+    endif()
+    if(NOT ("itk_metadata" IN_LIST DETRAY_GENERATE_METADATA))
+        list(APPEND DETRAY_GENERATE_METADATA "itk_metadata")
+    endif()
```
niermann999: This is a traccc requirement, right? Should it not be moved into the `if(ACTS_BUILD_TRACCC)` block above?

stephenswat: Makes sense; done.

krasznaa: It should be there. And in a different way:

```suggestion
    set(DETRAY_GENERATE_METADATA
        "itk_metadata"
        CACHE STRING
        "Semicolon-separated list of metadata generator scripts to run in this build"
    )
```

As I did in #5450. I.e. we set it to `"itk_metadata"` by default. But if the user of Acts wants to set it to something else, they must not be prevented from doing so. 🤔

stephenswat: _Not_ having `itk_metadata` in the list at all makes the traccc compilation fail right now, so I think it makes sense to set it in all cases.

niermann999: Indeed, the problem is that currently traccc hard requires the header to be present, even if the ITk detector specialization is not built. This needs to be fixed in traccc at some point, but is not an ACTS issue (for many ACTS users the ITk is not available)

niermann999 (dismissed):
I preferred the list append for the metadata, but let's move ahead with this for now

andiwand (approved):

PR #5463, opened 2026-05-18
fix: Track container resize dynamic columns if the column is created after tracks have been made

junggjo9:
- Track container resize dynamic columns if the column is created after tracks have been made
- Improve documentation of the DynamicColumnBase

--- END COMMIT MESSAGE ---

Tagging @andiwand, @paulgessinger


andiwand (dismissed):
LGTM 👍 thanks @junggjo9 !

I would still wait for @paulgessinger to have a look before getting this in

Core/include/Acts/EventData/detail/DynamicColumn.hpp
```diff
@@ -15,50 +15,113 @@
...
   virtual std::unique_ptr<DynamicColumnBase> clone(
       bool empty = false) const = 0;
 };
 
+/// Actual implementation of the Column memory using a std::vector as backend.
+/// The template parameter can be anything which is default constrictible and
+/// copy assignable.
+/// @tparam T: Data type of the data to be stored by the column
 template <typename T>
 struct DynamicColumn : public DynamicColumnBase {
```
andiwand:
```suggestion
class DynamicColumn : public DynamicColumnBase {
public:
```
nitpick `class` over `struct` as this has private members now

Core/include/Acts/EventData/detail/DynamicColumn.hpp
```diff
@@ -67,45 +130,66 @@ struct DynamicColumn : public DynamicColumnBase {
...
   }
 
+ private:
   std::vector<T> m_vector;
 };
 
+/// Template specification to circumvent the flaws of the bool implementation
+/// of a std::vector.
 template <>
 struct DynamicColumn<bool> : public DynamicColumnBase {
```
andiwand:
```suggestion
class DynamicColumn<bool> : public DynamicColumnBase {
public:
```

Core/include/Acts/EventData/VectorMultiTrajectory.hpp
```diff
@@ -496,7 +496,11 @@ class VectorMultiTrajectory final
   template <typename T>
   void addColumn_impl(std::string_view key) {
     HashedString hashedKey = hashStringDynamic(key);
-    m_dynamic.insert({hashedKey, std::make_unique<detail::DynamicColumn<T>>()});
+    auto insertItr = m_dynamic.insert(
+        {hashedKey, std::make_unique<detail::DynamicColumn<T>>()});
+    if (insertItr.second && size() > 0ul) {
+      insertItr.first->second->resize(size());
+    }
   }
```
andiwand, line 504: in case the user double calls `addColumn` with the same `key` this will overwrite the existing column with empty values, right?

not directly tied to this PR but since we touch it I wondered if this is the expected behavior

junggjo9: If the key is already assigned, then nothing changes

Core/include/Acts/EventData/detail/DynamicColumn.hpp
```diff
@@ -15,50 +15,113 @@
 
 namespace Acts::detail {
 
+/// Interface of the dynamic vector column used in the track container backend.
+/// The interface provides basic methods to retrieve the data in a type erase
+/// form. Further methods for memory pre-allocation and memory release are
+/// defined
 struct DynamicColumnBase {
```
junggjo9:
```suggestion
class DynamicColumnBase {
```

Core/include/Acts/EventData/VectorMultiTrajectory.hpp
andiwand: ah right this does not overwrite. seems fair to me

Core/include/Acts/EventData/detail/DynamicColumn.hpp
```diff
@@ -114,13 +196,19 @@ struct DynamicColumn<bool> : public DynamicColumnBase {
...
+  /// @copydoc DynamicColumnBase::copyFrom
   void copyFrom(std::size_t dstIdx, const std::any& srcPtr) override {
     const auto* other = std::any_cast<const bool*>(srcPtr);
     assert(other != nullptr &&
            "Source column is not of same type as destination");
     m_vector.at(dstIdx).value = *other;
   }
 
+ private:
+  /// Auxiliary struct to wrap the boolean
```
paulgessinger:
```suggestion
  /// Auxiliary struct to wrap a boolean for use in a vector
```

paulgessinger (dismissed):
Fine by me. The only question I have is if we could locate this behavior in the front end somehow so that we could get it consistently also for back ends like  and xAOD? (even if the latter isn't really used anymore)

junggjo9:
I am not really sure how we could do this. I've encountered this issue when I tried to decorate the track returned from the global chi2 fitter (https://gitlab.cern.ch/atlas/athena/-/merge_requests/88064)

andiwand (approved):

PR #5464, opened 2026-05-18
feat: Bind measurement container to python

benjaminhuth:
Restricted API for now: For now const-only access into container, measurement creation API exposed.

--- END COMMIT MESSAGE ---


andiwand (approved):
LGTM 👍 

we may want to touch the measurement EDM one more time to switch to a container that has a single measurement type and then having N containers for N subsystems as this maps more to what actual experiments do. but not really critical

PR #5468, opened 2026-05-19
fix: GNN4ITk example + AthenaDumpWriter

benjaminhuth:
This adds a runtime test for the otherwise untested `gnn4itk_example.py`, and writer for the "Athena ROOT Dump format". This will also facilitate creating module maps for the ODD in the future.


andiwand (approved):
fine by me

maybe we should reconsider the naming `Athena` in the future if we think this is could be a broader format that can now also be written from ACTS

benjaminhuth:
Yeah the naming is a bit unfortunate, and I would like to avoid this format to be adopted very broadly, but unfortunately GNN4ITk tools are using this relatively widely.

PR #5469, opened 2026-05-20
build: update traccc build

paradajzblond:
Updating build flags to build latest Traccc


thirdparty/traccc/CMakeLists.txt
```diff
@@ -43,29 +37,46 @@ set(TRACCC_SETUP_DFELIBS
...
+)
+
 set(TRACCC_BUILD_TESTING
     OFF
     CACHE BOOL
     "Turn off the build of the Traccc unit tests"
 )
+
+set(TRACCC_SETUP_TBB OFF CACHE BOOL "Do not set up TBB as part of Traccc")
+set(TRACCC_USE_SYSTEM_TBB ON CACHE BOOL "Use the system TBB for Traccc")
```
krasznaa: It would make sense to rather just leave the existing line 46 alone. `TRACCC_SETUP_TBB=FALSE` is correct. But as long as that is being set, `TRACCC_USE_SYSTEM_TBB` plays no role. (I.e. you shouldn't set it either.)

thirdparty/vecmem/CMakeLists.txt
```diff
@@ -31,5 +31,7 @@ set(VECMEM_BUILD_BENCHMARKING
     "Turn off the build of VecMem benchmarking"
 )
 
+set(VECMEM_BUILD_CUDA_LIBRARY ACTS_ENABLE_CUDA)
```
krasznaa: As we discussed just now...

krasznaa (commented):
Other than these, nothing else that I could see wrong. 🤔

andiwand (dismissed):

andiwand:
This if failing at least one CI build right now
- https://github.com/acts-project/acts/pull/5469/checks?check_run_id=76970608276

benjaminhuth (dismissed):

andiwand (approved):

PR #5472, opened 2026-05-20
ci: Run detray GitLab CI only for detray changes

stephenswat:
Currently, the detray GitLab CI runs for any PR. This consumes a lot of resources, so this commit changes the CI configuration to run only when the `Detray/` directory is changed.

--- END COMMIT MESSAGE ---

Tagging @niermann999, @andiwand.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -511,6 +511,10 @@ detray_build_cuda:
   artifacts:
     paths:
       - build
+  only:
+    changes:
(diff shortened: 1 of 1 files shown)


andiwand (approved):

paulgessinger:
I don't think this will work due to the way the GitLab jobs are dispatched. I think the only way is to check the diff inside the job itself and early return with exit code 0.

PR #5486, opened 2026-05-20
feat: Space point performance writer for Examples

andiwand:
--- END COMMIT MESSAGE ---

cc @benjaminhuth @murnanedaniel


andiwand:
@murnanedaniel with ttbar 200 I definitely see degraded strip space point efficiency

image

this is for "many muon" (200 vertices * 4 muons each) events

image

I think the ttbar result might be polluted by secondaries and decay products. I can try to clean out the denominator

murnanedaniel:
@andiwand Is it possible to include the fake rate / ghost rate / however you want to call it?

andiwand:
> @andiwand Is it possible to include the fake rate / ghost rate / however you want to call it?

yes this is already there. I just checked it again and fixed a bug. now the fake output looks more reasonable. I also added a particle selection mechanism

Examples/Io/Root/src/RootSpacePointPerformanceWriter.cpp
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  for (const auto [module1, module2] : m_stripModulePairMap) {
+    const Acts::Surface* surface1 =
+        m_cfg.trackingGeometry->findSurface(module1);
+    const Acts::Surface* surface2 =
+        m_cfg.trackingGeometry->findSurface(module2);
+
+    if (surface1 == nullptr || surface2 == nullptr) {
+      ACTS_DEBUG("Could not find surfaces for modules " << module1 << " and "
```
benjaminhuth: Hmm that should be a WARNING, shouldn't it?

benjaminhuth (commented):
Looks good! I think it would be interesting to also write out the simhit residuals for matched spacepoints in r, phi, z.

andiwand:
> Looks good! I think it would be interesting to also write out the simhit residuals for matched spacepoints in r, phi, z.

this is already possible using the output of the `RootSpacePointWriter` but I didn't plot residuals in the performance writer yet

benjaminhuth (approved):
LGTM

PR #5488, opened 2026-05-21
feat: Implement dynamical loading of additional particle data from CSV

njacazio:
Add functionality to load additional particle data at compile time from a CSV `/tmp/newparticles.py` file if it exists.

--- END COMMIT MESSAGE ---


andiwand:
thanks @njacazio !

in its current form this looks very specific and not something I would add to ACTS.

what is your use case here?

since we generate the particles during the build I would think it should be wired there if we want to have this as a general solution

njacazio:
Hi @andiwand thanks for the comments. Indeed it's specific to an analysis of non SM particles being simulated within ACTS. 
This is indeed transparent for the standard user as particles are only added if the csv file is found.
> since we generate the particles during the build I would think it should be wired there if we want to have this as a general solution

Do you mean to define the input csv as cmake options?
Many thanks!

andiwand:
Right, I don't think we want to rely on a file being read from a hard coded path at `/tmp/newparticles.py`. It would need to be at least a cmake option and gracefully wired in cmake to the code generation. For example by adding an argparse parameter

njacazio:
Hi @andiwand thank you for the suggestions, indeed, now the choice is more explicit.

Core/src/Definitions/CMakeLists.txt
```diff
@@ -5,8 +5,16 @@ target_sources(
 
 include(ActsCodegen)
 
+set(PARTICLE_TABLE_CODEGEN_PYTHON_ARGS "")
```
andiwand: I think this should use `option` and directly fill the path. but I am not exactly sure if it then should rather go into the central https://github.com/acts-project/acts/blob/main/CMakeLists.txt

Fatras/scripts/generate_particle_data_table.py
```diff
@@ -138,8 +139,24 @@ def clang_format(content):
     p.add_argument(
         "--format", action="store_true", help="Run clang-format on the output."
     )
+    p.add_argument("--additional-particles", type=Path, help="CSV file with additional particles.", default=None)
 
     args = p.parse_args()
+    if args.additional_particles is not None:
+        import os
```
andiwand: import can be elevated to the global ones

Fatras/scripts/generate_particle_data_table.py
```diff
@@ -138,8 +139,24 @@ def clang_format(content):
...
+        additional_particle_csv = args.additional_particles
```
andiwand:
```suggestion
```
no need to alias that IMO

Fatras/scripts/generate_particle_data_table.py
```diff
@@ -138,8 +139,24 @@ def clang_format(content):
...
+            print(f"Loading additional particle data from: {additional_particle_csv}")
+            particles_before = Particle.all()
+            Particle.load_table(additional_particle_csv, append=True)
+            particles_after = Particle.all()
+            new_particles = set(particles_after) - set(particles_before)
+            print(f"Loaded {len(new_particles)} additional particles from {additional_particle_csv}")
+            for p in new_particles:
+                print(f"  - {p.name} (PDG ID: {p.pdgid})")
+        else:
+            print(f"Warning: Additional particle CSV file not found: {additional_particle_csv}")
```
andiwand: should be a hard error if this file does not exist which I hope would be triggered by `Particle.load_table`
```suggestion
        print(f"Loading additional particle data from: {additional_particle_csv}")
        particles_before = Particle.all()
        Particle.load_table(additional_particle_csv, append=True)
        particles_after = Particle.all()
        new_particles = set(particles_after) - set(particles_before)
        print(f"Loaded {len(new_particles)} additional particles from {additional_particle_csv}")
        for p in new_particles:
            print(f"  - {p.name} (PDG ID: {p.pdgid})")
```

andiwand (commented):
much better already - thanks @njacazio !

can you update the pr description?

I left a few more comments. other than that I would ask @paulgessinger to have another look as he designed this in the first place

paulgessinger:
Is there any reason not to add them unconditionally? Or are they not available in the particle library?

paulgessinger:
Is there any reason not to add them unconditionally? Or are they not available in the particle library?

njacazio:
> Is there any reason not to add them unconditionally? Or are they not available in the particle library?

Hi @paulgessinger the particles are not available in the common library.

paulgessinger:
@njacazio how large are the csvs that you're adding? I'm wondering if we could just ship those with the library and unconditionally include them.

njacazio:
Hi @paulgessinger the csv per se is quite limited, but I guess it's user dependent, especially if the particle parameters are not measured yet.
Shipping it with the library imho would be against the purpose of the change, i.e. allow the user to add custom PDGs in the simulation.

andiwand:
Just to throw in a few thoughts. I think one can argue in both directions.

It would definitely be easier to set up the build while having the extra file in the repo. Users could maintain a modified CSV in their clone. But depending on how exotic these particles are we may not want to put them in and it will be hard to audit this. And if we start changing it conflicts will be a challenge.

The other alternative would be to make the particle data runtime extensible so we don't have to bother with the build.

njacazio:
Hi @andiwand @paulgessinger as we discussed yesterday we now support (if the csv file is proveded) both additional particles and dynamic masses.
This would make an additional line in the particledata header e.g.:
`{ pdg, readCsvFloatAt("/tmp/data.csv", 3, 3) },`

The only thing that I would keep is the PDG code and the name. These are hardcoded. By doing this one does not impact on the CPU for lookup, nor on additional maps. If one wants to add more particles it makes sense to me to retrigger the compilation as this is a major change.
Tagging also @jovankaas

Fatras/scripts/generate_particle_data_table.py
```diff
@@ -39,6 +40,55 @@
 
 """
 
+# Only needed if the user wants to add additional particles at runtime from a CSV file, otherwise it can be ignored
+CSV_FILE_READER = """
```
andiwand, line 44: I don't think we should put multiple lines of non generated code into the code generator. this should really go into the c++ library

Fatras/scripts/generate_particle_data_table.py
andiwand, line 1: I don't think we need to touch this at all if we keep all the changes in c++

andiwand (commented):
hi @njacazio !

after the discussion last week my takeaway was rather to keep everything outside the generator and add a layer of user defined particles on top. this would be part of `ParticleData.{hpp,cpp}` where we keep a runtime mutable map hidden in the source file which can be modified with setters accessible from the header file. these can then be bound to python and called from your script.

one thing we might want to be careful about is thread safety. do we want to guard the read and writes with a mutex or do we delegate the responsibility to the user?

if there is a common CSV format for particles we can also think about having a high level interface to read them in from file. but not sure if we want to have this in core?

PR #5489, opened 2026-05-21
ci: Exclude `StrawLineFitterTest` from debug build tests

andiwand:
Currently this thing alone runs 25 minutes in our CI.

--- END COMMIT MESSAGE ---

debug build
```
347/347 Test #135: StrawLineFitterTest ..................   Passed  1409.57 sec
```

normal build
```
358/358 Test #135: StrawLineFitterTest ..................   Passed   24.03 sec
```

Ultimately we should reconsider if this is a test or how this is implemented

cc @junggjo9 @benjaminhuth


benjaminhuth:
Maybe we can think where to put this executable, because its not really a unit test I think. Maybe to the benchmarks?

AJPfleger (approved):

PR #5471, opened 2026-05-20
feat: Integrate the traccc repository

stephenswat:
This commit incorporates the entire codebase of traccc, the ACTS R&D
line on parallel track reconstruction, into the ACTS project.

The traccc commit used is 5dc136296a6bdb0568f1444df78cd82cfd46e5a7.

Co-authored-by: Beomki Yeo <beomki.yeo@gmail.com>
Co-authored-by: Andreas Salzburger <andreas.salzburger@cern.ch>
Co-authored-by: Attila Krasznahorkay <attila.krasznahorkay@cern.ch>
Co-authored-by: Paul Gessinger <paul.gessinger@cern.ch>
Co-authored-by: Joana Niermann <joana.niermann@cern.ch>
Co-authored-by: Guilherme Metelo Rita de Almeida <guilherme.almeida@tecnico.ulisboa.pt>
Co-authored-by: Neža Ribaric <neza.ribaric@cern.ch>
Co-authored-by: Konrad Kusiak <konrad.kusiak@kx27.eu>
Co-authored-by: Mateusz Jakub Fila <mateusz.jakub.fila@cern.ch>
Co-authored-by: Stewart Martin-Haugh <stewart.martin-haugh@stfc.ac.uk>
Co-authored-by: Xola Gugulethu Mapekula <xola.mapekula@cern.ch>
Co-authored-by: Sylvain Joube <sylvain.joube@gmail.com>
Co-authored-by: Mark Brettell <mark.brettell@cern.ch>
Co-authored-by: Mikkel Mødekjær <mimodekj@cern.ch>
Co-authored-by: Fabrice le Goff <flegoff@cern.ch>
Co-authored-by: Hadrien Grasland <knights_of_ni@gmx.com>
Co-authored-by: Ryan Cross <ryan.cross@warwick.ac.uk>
Co-authored-by: Georgiana Mania <georgiana.mania@gmail.com>
Co-authored-by: Petr Fiedler <petr.fiedler@cern.ch>
Co-authored-by: Gabriel Carmona <gabriel.carmona@phd.unipi.it>
Co-authored-by: Vakho Tsulaia <qiarik@gmail.com>
Co-authored-by: Shima Shimizu <shima.shimizu@cern.ch>
Co-authored-by: Malik Shahzad Muzaffar <shahzad.malik.muzaffar@cern.ch>
Co-authored-by: Frederik Verdoner Barba <fbarba@ethz.ch>
[...]

PR #5490, opened 2026-05-21
build(detray): Fix vecmem CMake include

stephenswat:
This commit fixes an inherited bug from the detray repository in which the use of `vecmem-check-language` does not correctly use the path exported by vecmem. Also turns off the setup of detray by default, in order not to disturb the user experience while we complete the integration.

--- END COMMIT MESSAGE ---

see discussion https://mattermost.web.cern.ch/acts/pl/jbuxpdkyh3b18y9o48pepmcx3o


andiwand (dismissed):

niermann999 (dismissed):

stephenswat:
Sorry, made a change to this. Instead of unconditionally including that file, I believe it should actually not be included _at all_ when using an installed version of vecmem.

andiwand (dismissed):

jmcarcell:
Slightly related to this PR, and because building ACTS in the LCG stacks failed today, so now vecmem is required by default (and also in the future)? It's a bit weird that it broke today since the default was turned ON a few days ago, that I don't understand yet. Will detray also be required by default?

andiwand:
> Slightly related to this PR, and because building ACTS in the LCG stacks broke yesterday, so now vecmem is required by default (and also in the future)? It's a bit weird that it broke today since the default was turned ON a few days ago, that I don't understand yet.

@jmcarcell a few more details were shared here https://mattermost.web.cern.ch/acts/pl/jbuxpdkyh3b18y9o48pepmcx3o

stephenswat:
> Slightly related to this PR, and because building ACTS in the LCG stacks broke yesterday, so now vecmem is required by default (and also in the future)? It's a bit weird that it broke today since the default was turned ON a few days ago, that I don't understand yet.

Hi Juan, what broke in this case is a line inside of the detray configuration. vecmem has been enabled by default for a while now. Similarly, detray has been enabled for a while. However, it is only as of #5462 that turning on the detray code actually does anything if you don't have the traccc code turned on. In other words, detray was being masked by the traccc flag, even if the setup of detray was enabled. The underlying bug here is in how ACTS and detray set up vecmem, which was leading to colliding vecmem versions.

As I described on the ACTS Mattermost, the simplest ways of mitigating this issue is to set `ACTS_SETUP_DETRAY` to false, which will avoid setting up detray at all, and means that detray won't set up its own vecmem.

This PR also makes that the default behaviour so that future changes to the detray and traccc build systems won't cause downstream failures.

jmcarcell:
In the LCG stacks it wasn't enough to turn detray OFF, vecmem also had to be OFF not to be found : https://gitlab.cern.ch/sft/stacks/lcgcmake/-/merge_requests/3917.

stephenswat:
> In the LCG stacks it wasn't enough to turn detray OFF, vecmem also had to be OFF not to be found : https://gitlab.cern.ch/sft/stacks/lcgcmake/-/merge_requests/3917.

The fact that only turning detray off didn't fix the issue is worrying. Do you have logs showing what went wrong with detray turned off but vecmem turned on?

jmcarcell:
Yes, but they only show that vecmem was not found, for example on Alma9: https://lcgapp-services.cern.ch/cdash/build/172617/file/11834. This is with the first commit of https://gitlab.cern.ch/sft/stacks/lcgcmake/-/merge_requests/3917 (with detray OFF but vecmem not set, so the default value).

stephenswat:
Superseded by #5504 and #5471.

PR #5492, opened 2026-05-21
feat: python bindings for host and device memory resources

paradajzblond:
As discussed with krasznaa this builds on top of asalzburger  #5449 and adds the resources needed in #5348


andiwand:
@paradajzblond the title prefix should not contain `

also using @ in the description will result in it being copied into the commit message and people will be notified by github every time the commit gets pushed somewhere

Python/Plugins/src/TracccHost.cpp
andiwand, line 1: I think this should be split in `TracccHost.cpp` and `TracccDevice.cpp`. then the stub mechanism will deal with the guarded compilation

Python/Plugins/src/Vecmem.cpp
andiwand, line 1: same here

Python/Plugins/CMakeLists.txt
andiwand, line 1: the suggestion above should also simplify the logic here

Python/Plugins/src/Traccc.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+#endif
+
+namespace py = pybind11;
+using namespace pybind11::literals;
+
+void bind_traccc_host(py::module_& m) {
+  // ---- traccc types ----
+  py::class_<traccc::memory_resource, std::shared_ptr<traccc::memory_resource>>(
+      m, "TracccMemoryResource")
```
krasznaa: With this the type name in Python would end up being `acts.traccc.TracccMemoryResource`? 🤔 

It's not a black or white question, but as long as we keep the C\+\+ type names with their current naming convention, I would prefer the Python name to be `acts.traccc.memory_resource`. 🤔

Python/Plugins/src/Traccc.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <string>
+
+#include <traccc/utils/memory_resource.hpp>
+
+#if defined(ACTS_ENABLE_CUDA)
+#include <traccc/cuda/utils/stream.hpp>
+
+void bind_traccc_cuda(py::module_& m) {
+  py::class_<traccc::cuda::stream, std::shared_ptr<traccc::cuda::stream>>(
+      m, "CudaStream")
```
krasznaa: So, what's the conclusion here? Is it not possible to achieve a type name of `acts.traccc.cuda.stream`?

If not, then I would still prefer `acts.traccc.cuda_stream`.

But I'd really be interested in the opinion of people more experienced with pybind11 on whether we could add "additional namespaces" into these names. 🤔

paradajzblond: apparently you can do this with submodules, so will change this

Python/Plugins/src/TracccHost.cpp
paradajzblond: done!

asalzburger:
I have already created a 
`Python/Examples/src/Vecmem.cpp` which does part of the job, right ?

- should we not expand this one ?

paradajzblond:
> I have already created a `Python/Examples/src/Vecmem.cpp` which does part of the job, right ?
> 
> * should we not expand this one ?

We need a safe guard against building cuda components, originally I had everything in a single file because I wanted to construct composites ie you can retrieve the correct bundle of memory resources, stream and copy in a single call. However, Stefl suggested that we rather have separate files and the compilation is safeguarded by the stub mechanism, so this is where those changes come from.

andiwand:
where do we stand here @paradajzblond @asalzburger ?

apart from the discussion there seem to be cmake configuration errors in the CI

PR #5493, opened 2026-05-21
feat: support geometric (segmented) digitization on cylinder surfaces

altsybee:
The geometric digitizer (realistic pixel/strip segmentation, as opposed to simple smearing) previously worked only on plane and disc surfaces. Cylindrical sensitive surfaces silently produced zero channels because the drift, mask, and segmentation steps had no cylinder code path. 

This PR adds digitization for cylindrical surfaces, and combines the functionality into unified surface-type-agnostic classes, `SurfaceDrift` and `SurfaceMask` (instead of `PlanarSurfaceDrift/PlanarSurfaceMask`; the essence of the code for the planar "path" remained untouched).

The cylindrical pixel and strip layers can be digitized with a grid defined in the `digi.json` (section `segmentation/binningdata`), using `binRPhi` + `binZ` (both in mm at the sensor radius).

No JSON schema change is required — `binRPhi/binZ` were already accepted by the binning parser.

Example usage in `digi.json`:
```
{
  "NAME": "OuterPixels::Barrel",
  "volume": 17,
  "layer": 2,
  "value": {
    "geometric": {
      "digital": true,
      "indices": [0, 1],
      "segmentation": {
        "binningdata": [
          { "bins": 2000, "max":  30, "min": -30, "option": "closed",
            "type": "equidistant", "value": "binRPhi" },
          { "bins": 4800, "max":  48, "min": -48, "option": "open",
            "type": "equidistant", "value": "binZ" }
        ]
      },
      "thickness": 0.02,
      "threshold": 0.001,
      "charge-smearing": { "mean": 0.0, "stddev": 0.001, "type": "Gauss" }
    }
  }
}
```

In this snippet for cylinders:

`binRPhi`: min=-30, max=30 means the rPhi grid covers a 60 mm arc length on the cylinder (= R·φ in mm at the sensor radius).
`binZ`: min=-48, max=48 means the z grid covers ±48 mm along the cylinder axis.

(For the grid to actually match the sensor, those numbers should agree with the cylinder bounds: `binZ` extent ≈ `2·halfLengthZ`, and the `binRPhi` extent ≈ `2·R·halfPhiSector` (so ±π·R for a closed cylinder, less for a sector). `bins` is the number of cells across that range; with `type: "equidistant"` the cell pitch is just `(max − min) / bins.`)

Example output for cylindrical layer: cluster size vs eta:
image

(Note: unit tests  `Tests/UnitTests/Fatras/Digitization/ChannelizerTests.cpp,  PlanarSurfaceDriftTests.cpp, PlanarSurfaceMaskTests.cpp` still cover Planar surfaces, tests for Cylinders can be added later).


Fatras/src/Digitization/SurfaceMask.cpp
```diff
@@ -56,29 +57,82 @@ void checkIntersection(std::vector<Acts::Intersection2D>& intersections,
...
+
+ClipResult clipLiangBarsky(double x0, double y0, double x1, double y1,
+                           double xmin, double xmax, double ymin, double ymax) {
+  const double dx = x1 - x0;
+  const double dy = y1 - y0;
+
+  double tEnter = 0.0;
+  double tExit = 1.0;
+
+  const double p[4] = {-dx, dx, -dy, dy};
```
asalzburger: Can we change that to `std::array>double,4>`, please ?

Fatras/src/Digitization/SurfaceMask.cpp
```diff
@@ -56,29 +57,82 @@ void checkIntersection(std::vector<Acts::Intersection2D>& intersections,
...
+  const double q[4] = {x0 - xmin, xmax - x0, y0 - ymin, ymax - y0};
```
asalzburger: Same here, also flagged by SonarCloud.

Fatras/src/Digitization/SurfaceMask.cpp
```diff
@@ -56,29 +57,82 @@ void checkIntersection(std::vector<Acts::Intersection2D>& intersections,
...
+/// On success returns a {tEnter, tExit} pair in [0, 1] describing the
+/// parameter range of the clipped segment along (end - start). If the segment
+/// lies completely outside the rectangle, valid is false.
+struct ClipResult {
+  double tEnter;
+  double tExit;
+  bool valid;
+};
+
+ClipResult clipLiangBarsky(double x0, double y0, double x1, double y1,
```
asalzburger: To follow Acts style,  change this to `Acts::Result<>` pattern,  which carries implicitly the `isValid()` flag.

asalzburger (changes requested):
Small changes to be done, the eventual SonarCloud issues or the original code, I can follow up in a separate PR if needed.

asalzburger:
Looks already very good, a few changes and we can move this forward.

asalzburger (approved):
LGTM now, awaiting CI

PR #5495, opened 2026-05-22
fix: Don't mark Any copy operations noexcept

paulgessinger:
Let's try not making the copy operations in Any noexcept. They mask actual paths that can throw (the contained object's constructor). For vector<Any> optimization, I think the moves being noexcept should be sufficient. 

This also fixes an exception bug that was masked by noexcept before, where an exception on copy would produce an Any pointing at invalid memory.


Core/include/Acts/Utilities/Any.hpp
```diff
@@ -192,9 +192,13 @@ class AnyBase : public AnyBaseAll {
...
   /// @tparam T Type of the value to store
   /// @param value Value to store in the Any
+  /// @note Not noexcept: storing a type larger than the small buffer allocates
+  ///       on the heap (may throw std::bad_alloc) and the stored type's own
+  ///       constructor may throw.
   template <typename T>
-  explicit AnyBase(T&& value) noexcept(detail::kAnyNoexcept)
-    requires(isStorable<std::decay_t<T>>())
+  explicit AnyBase(T&& value)
+    requires(!std::is_base_of_v<AnyBaseAll, std::decay_t<T>> &&
```
paulgessinger, line 200: Adding this just to make sonar happy

andiwand (approved):

PR #5497, opened 2026-05-22
chore!: Remove deprecated `RiddersPropagator`

andiwand:
Removed the already deprecated `RiddersPropagator`, which was replaced by the new `RiddersStepper`

--- END COMMIT MESSAGE ---


asalzburger (approved):
Easy one.

PR #5499, opened 2026-05-22
chore!: Remove deprecated space point formation

andiwand:
Removed the already deprecated space point formation which is replaced by `Core/include/Acts/SpacePointFormation2`. A followup PR will rotate the new package name back to the original `Core/include/Acts/SpacePointFormation`

--- END COMMIT MESSAGE ---


andiwand:
fyi @osbornjd this might be a bigger chunk to absorb on your side

asalzburger (approved):
One CI failure, but looks like a time-out, approved here as it just doing code removal.

PR #5500, opened 2026-05-22
chore!: Remove deprecated seeding components

andiwand:
Removed the already deprecated seeding components in `Core/include/Acts/Seeding` which have been replaced by `Core/include/Acts/Seeding2`. A future PR will rotate the component name back to the original one

--- END COMMIT MESSAGE ---


andiwand:
fyi @osbornjd this might be a bigger chunk to absorb on your side

osbornjd:
Thanks for the notification - what is the motivation for this? Is this deprecating the grid based seed finder?

andiwand:
> Thanks for the notification - what is the motivation for this? Is this deprecating the grid based seed finder?

the seeding has beed rewritten and currently lives in `Seeding2`. overall this is supposed to be functionally equivalent but the interface changed completely. besides this the EDM is not completely defined within ACTS. the Examples exercise this

osbornjd:
If the interface changed completely, is there an obvious mapping from the `Seeding` interface to `Seeding2` interface? Or will we be effectively having to retune our implementation of the seeding algorithm?

andiwand:
> If the interface changed completely, is there an obvious mapping from the `Seeding` interface to `Seeding2` interface? Or will we be effectively having to retune our implementation of the seeding algorithm?

I cannot say if it very obvious. the config params should be mostly the same but got shuffled around. arriving at the exact same performance might take some iterations.

we can discuss this further in the next meeting. I believe we can also offer to delay the removal if that helps. but I should also mention that we are pushing this already for quite some time while mentioning this switchover many times in past meetings

asalzburger:
Looks like we need to make and help with the migration here.
ACTS runs in default already in `Seeding2` and that will turn into the sole `Seeding` schema.

@osbornjd what release are you currently running with & what are your next, planned update?

On a side note: we are almost finished with defining the experiment contact group and structure, which is exactly the place where these longer term planning items will be discussed in the future.

osbornjd:
I apologize for missing the dev meetings where this was discussed, unfortunately our weekly sPHENIX tracking meeting is also on Tuesdays at 17:00 CERN time.

We are currently running on v45.0.0. My naive question would be, is it necessary to completely deprecate the old version? For example, right now there is `Seeding` and `Seeding2` (for which the plan is for Seeding2 is to replace Seeding). Is there any problem with keeping both in existence?

andiwand:
> I apologize for missing the dev meetings where this was discussed, unfortunately our weekly sPHENIX tracking meeting is also on Tuesdays at 17:00 CERN time.
> 
> We are currently running on v45.0.0. My naive question would be, is it necessary to completely deprecate the old version? For example, right now there is `Seeding` and `Seeding2` (for which the plan is for Seeding2 is to replace Seeding). Is there any problem with keeping both in existence?

yes the problem is maintaining it and keeping it working, which might already not be the case. we moved the tests and integration fully towards `Seeding2` already. `Seeding` is only hanging in there right now with no connections. an argument for a transition period can be made, but it should be clear that the code needs to be removed at some point. from my point of few it should be sooner than later because people continue to target the wrong one

AJPfleger (approved):

PR #5502, opened 2026-05-22
build(detray): Increase usability of detray build

stephenswat:
This commit fixes several issues with the detray build system which have been identified when trying to use it through ACTS to build traccc. More specifically, it:

1. Includes the detray directory directory, without relying on FetchContent. Consequently, removes the ability to use a pre-installed detray.
2. Ensures that the detray metadata files are always generated and installed.
3. Removes the dependency on Google Test and Google Benchmark from installed versions of detray.

--- END COMMIT MESSAGE ---

Tagging @paradajzblond @krasznaa; aims to incorporate everything that #5450 does.


krasznaa (approved):
These look good. Hopefully the CI agrees.

andiwand (approved):

PR #5504, opened 2026-05-23
build: Toggle default off for covfie, detray, vecmem

andiwand:
Having `ACTS_SETUP_COVFIE`, `ACTS_SETUP_DETRAY`, and `ACTS_SETUP_VECMEM` on by default causes various problems downstream. I propose to turn them off by default until these issues are resolved

--- END COMMIT MESSAGE ---


stephenswat (approved):
Good move, let's do this. Potentially we should keep this permanently, makes sense to me.

PR #5510, opened 2026-05-26
feat: json based performance writers

asalzburger:
This PR adds a json writing possibility for histograms collected in the Examples framework.

This takes the Boost histograms and dumps them into json.

It also fixes SonarCloud issues not only in the Json writer but the exact same ones in the other writers.

--- END COMMIT MESSAGE ---


asalzburger:
Discussed, and decided to drop this in favor of using the python handles directly.

PR #5513, opened 2026-05-27
build(detray): Add CMake external include guards

stephenswat:
The detray setup currently makes it fairly easy to include dependencies twice through the FetchContent channel, and this can lead to extremely difficult to debug issues where multiple versions of a dependency fight for precedence. This commit therefore introduces some include guards, throwing an error if a configuration leads to the re-fetching of dependencies that are already in place.


stephenswat:
Tagging @niermann999.

andiwand (approved):

PR #5515, opened 2026-05-27
chore: expand TrackFinderPerformanceWriter usage boolean throughout python

JustusRudolph:
Refactor current TruthMatching for particle to track matching & update TrackFinderPerformance writer accordingly
--- END COMMIT MESSAGE ---

- `TruthMatching` for particles to tracks now includes vectors of `TrackIndexWithWeight`, a pair of the track index in the track collection, and the associated weight related to the particle, for both duplicates and fakes. The weight of a track is simply the number of hits it contains of the particle divided by the total number of hits that the particle left. These vectors are sorted from most to least matched tracks.
- Add these vectors as outputs for the `RootTrackFinderPerformanceWriter`.
- Add the `writeMatchingDetails` parameter throughout calls of `reconstruction.py:addTrackWriters()` to make output of tree tunable from highest level execution.


Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TrackTruthMatcher.cpp
```diff
@@ -96,46 +96,58 @@ ProcessCode TrackTruthMatcher::execute(const AlgorithmContext& ctx) const {
...
         static_cast<double>(nMajorityHits) / track.nMeasurements() >=
         m_cfg.matchingRatio;
-    const bool truthMatched =
-        static_cast<double>(nMajorityHits) /
-            particleTruthHitCount.at(majorityParticleId) >=
-        m_cfg.matchingRatio;
+    const double trackTruthRatio = static_cast<double>(nMajorityHits) /
+                                   particleTruthHitCount.at(majorityParticleId);
+    const bool truthMatched = trackTruthRatio >= m_cfg.matchingRatio;
+    const auto trackWithWeight = std::make_pair(track.index(), trackTruthRatio);
```
andiwand:
```suggestion
    const TrackIndexWithWeight trackWithWeight(track.index(), trackTruthRatio);
```

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TrackTruthMatcher.cpp
```diff
@@ -96,46 +96,58 @@ ProcessCode TrackTruthMatcher::execute(const AlgorithmContext& ctx) const {
...
                                               std::nullopt, particleHitCounts};
-
-      auto& particleTrackMatch = particleTrackMatching[majorityParticleId];
-      ++particleTrackMatch.fakes;
+      particleTrackMatch.fakes.push_back(trackWithWeight);
     }
   }
+  // Sort duplicate and fakes by decreasing weight
+  for (auto& [_, entry] : particleTrackMatching) {
+    std::sort(entry.duplicates.begin(), entry.duplicates.end(),
```
andiwand:
```suggestion
    std::ranges::sort(entry.duplicates,
```

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TrackTruthMatcher.cpp
```diff
@@ -96,46 +96,58 @@ ProcessCode TrackTruthMatcher::execute(const AlgorithmContext& ctx) const {
...
+              [](const TrackIndexWithWeight& t1, const TrackIndexWithWeight& t2) {
+                return t1.second > t2.second;  // sort by weight
+              });
+    std::sort(entry.fakes.begin(), entry.fakes.end(),
```
andiwand:
```suggestion
    std::ranges::sort(entry.fakes,
```

andiwand (commented):
Thanks for the contribution @JustusRudolph ! I left a few minor implementation comments.

What is your motivation for changing this? I can see that the truth matching can retain a bit more information which can be utilized downstream. But currently you only wire that to the `RootTrackFinderPerformanceWriter` which writes out tracks. So tracks that match to a particle will also contain about information about other tracks that match to this particle? To me this seems a bit counterintuitive and potentially tied to a very specific use case.

JustusRudolph:
> Thanks for the contribution @JustusRudolph ! I left a few minor implementation comments.

Thanks for having a look @andiwand! Your suggestions all looked good (I wasn't aware of ranges sort, looks like I'm stuck in 2019) so I applied them all, thank you!

> What is your motivation for changing this? I can see that the truth matching can retain a bit more information which can be utilized downstream. But currently you only wire that to the `RootTrackFinderPerformanceWriter` which writes out tracks. So tracks that match to a particle will also contain about information about other tracks that match to this particle? To me this seems a bit counterintuitive and potentially tied to a very specific use case.

The reason I only applied it to the performance finder for now is because that is the only location I've seen where by default the matching details are written out to -- but I have limited experience with this so please let me know if there additional places where nominally a "matching bool" could be added.

Tracks do not contain any additional information, only particles that are written following the matching process now contain information regarding which tracks indices are matched and not matched to them. All tracks that are written will contain the same information as before. The `RootTrackFinderPerformanceWriter` writes both the performance wrt particles and tracks with the `TrackFinderPerformanceCollector`, but the `matchingdetails` tree was just for particle performance.

My motivation is because I want to be able to access duplicate indices (and fakes) of a particle for both seeding and tracking (ambi) in the future when I'm checking the impact of increased localised occupancies due to detector services and thus making more informed decisions for both placement of such services and how to adapt the tracking. As for the `chore`, right now I'm already using the `matchingdetails` extensively and I wanted to be able to set it from a higher level and not have to change it in `reconstruction.py::addWriters()` every time to toggle it.

Please let me know if anything is still unclear or if I have misunderstood something. Thank you!

andiwand:
To me it seems the additional information would rather belong to an NTuple writer like `RootTrackSummaryWriter`. but this is a per track writer so I don't think we should store information about tracks that also matched to the matched particle. `RootTrackFinderPerformanceWriter` is supposed to write out plots so I don't think we should write out a tree. if you want to relate duplicated tracks from particles I think this should be rather down offline after reading the output from `RootTrackSummaryWriter`. does that make sense?

andiwand (approved):

PR #5516, opened 2026-05-27
fix: Bugs in the record_propagation function

niermann999:
Fix two bugs in the record propagation function. These mainly arise, since the state type deduplication in the actor chain is not preserving the order of the states in the tuple. The second bug revolved around extracting types from the state reference tuple not as actual reference types, which fails during compile time. These bugs were not detected earlier, since they only appear if the parameter transport is compiled into the validation. And that is at this point in time only done in an open PR in traccc.

--- END COMMIT MESSAGE ---


niermann999:
Related to issue #5517

stephenswat (approved):
Looks good to me. 👍

PR #5518, opened 2026-05-28
feat: fix in detray for loglevel change

asalzburger:
This fix is necessary in Detray to allow changing the log level to DEBUG,
another fix in traccc is also necessary.

--- END COMMIT MESSAGE ---


Detray/io/include/detray/io/backend/detail/grid_reader.hpp
```diff
@@ -384,8 +384,9 @@ class grid_reader {
 #if DETRAY_LOG_LVL > 2
     using single_axes_t =
         types::list<axis::single_axis<bounds_ts, binning_ts>...>;
-#endif
+    static_assert(sizeof(single_axes_t) > 0u);
```
asalzburger: Not sure we need the `static_assert` here, doesn't harm though.

asalzburger: Let me double-check if this is needed.

asalzburger: It is indeed needed, in order to avoid a unused warning which breaks compilation.

niermann999: What is the warning that appears otherwise? Maybe it would make sense to put it into the `DETRAY_TYPENAME` macro if it is related to that?

niermann999: you could try 
```
    using single_axes_t [[maybe_unused]] =
        types::list<axis::single_axis<bounds_ts, binning_ts>...>;
```
Despite the odd placement of the attribute, it would make the intention easier to read

asalzburger: That does indeed do the job.

Detray/io/include/detray/io/backend/detail/grid_reader.hpp
```diff
@@ -382,10 +382,10 @@ class grid_reader {
     DETRAY_DEBUG_HOST("--> frame:   " << DETRAY_TYPENAME(local_frame_t));
     DETRAY_DEBUG_HOST("--> bins:    " << DETRAY_TYPENAME(bin_t));
 #if DETRAY_LOG_LVL > 2
```
niermann999: If the maybe unused worked, then maybe these preprocessor guards can be removed altogether?

niermann999 (dismissed):

Detray/io/include/detray/io/backend/detail/grid_reader.hpp
asalzburger: And right you are - the guards are not needed anymore, I push a new commit.

niermann999 (approved):

Detray/io/include/detray/io/backend/detail/grid_reader.hpp
niermann999: Thanks for cleaning this up

PR #5524, opened 2026-05-29
feat: add options to use spacepoint time information in TripletSeeder, with a new ROOT seeding analysis tool

wangyanqi191025:
feat: add options to use spacepoint time information in TripletSeeder, with a new ROOT seeding analysis tool

--- END COMMIT MESSAGE ---

1. Core code change: add options to use time information in TripletSeeder, actually apply the cut of time in BroadTripletSeedFilter and DoubletSeedFinder
2. use time cut in Examples GridTripletSeedingAlgorithm 
3. add a ROOT seeding analysis tool

Still a draft for the moment. Some potential concerns:

- Extra CPU time consumption in DoubletSeedFinder in no-time situation, since reading too much useless `nan ` for time and time variance. even causing problems when no time info in spacepoint?

To @XiaocongAi  @pbutti

### DIFF
--- a/Core/src/Seeding2/BroadTripletSeedFilter.cpp
+++ b/Core/src/Seeding2/BroadTripletSeedFilter.cpp
@@ -348,6 +348,49 @@ void BroadTripletSeedFilter::filterTripletsMiddleFixed(
   std::size_t numTotalSeeds = 0;
   for (const auto& [bottom, middle, top, bestSeedQuality, zOrigin,
                     qualitySeed] : sortedCandidates) {
+    // calculate chi2 of the triplet time compatibility
+    if (config().applyTripletTimeFilter) {
+      const auto& spM = spacePoints[middle];
+      const auto& spB = spacePoints[bottom];
+      const auto& spT = spacePoints[top];
+
+      const float tM = spM.time();
+      const float tB = spB.time();
+      const float tT = spT.time();
+
+      if (!std::isnan(tB) && !std::isnan(tM) && !std::isnan(tT)) {
+        const float varianceTM = spM.varianceT();
+        const float t0M = tM - std::sqrt(spM.zr()[0] * spM.zr()[0] +
+                                         spM.zr()[1] * spM.zr()[1]) /
+                                   Acts::PhysicalConstants::c;
+        const float varianceTB = spB.varianceT();
+        const float t0B = tB - std::sqrt(spB.zr()[0] * spB.zr()[0] +
+                                         spB.zr()[1] * spB.zr()[1]) /
+                                   Acts::PhysicalConstants::c;
+        const float varianceTT = spT.varianceT();
+        const float t0T = tT - std::sqrt(spT.zr()[0] * spT.zr()[0] +
+                                         spT.zr()[1] * spT.zr()[1]) /
+                                   Acts::PhysicalConstants::c;
+
+        const float t0Mean = (t0B + t0M + t0T) / 3.f;
+        const float chi2_t0 = ((t0B - t0Mean) * (t0B - t0Mean)) / varianceTB +
+                              ((t0M - t0Mean) * (t0M - t0Mean)) / varianceTM +
+                              ((t0T - t0Mean) * (t0T - t0Mean)) / varianceTT;
+        if (chi2_t0 > config().tripletTimeFilterChi2) {
+          ACTS_VERBOSE("Triplet time compatibility chi2 is "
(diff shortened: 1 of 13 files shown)


Examples/Io/Root/include/ActsExamples/Io/Root/RootSeedEventAnaWriter.hpp
andiwand, line 1: I believe we already have similar writers for this, like `RootSeedWriter.hpp` `RootTrackParameterWriter.hpp`. can they be used for your use case?

Core/src/Seeding2/DoubletSeedFinder.cpp
```diff
@@ -52,6 +52,10 @@ class Impl final : public DoubletSeedFinder {
     const float rM = middleSp.zr()[1];
     const float varianceZM = middleSp.varianceZ();
     const float varianceRM = middleSp.varianceR();
+    const float tM = middleSp.time();
+    const float varianceTM = middleSp.varianceT();
+    const float t0M =
+        tM - std::sqrt(rM * rM + zM * zM) / Acts::PhysicalConstants::c;
```
andiwand, line 58: I fear this might have quite some impact on the compute performance of the seeding without time. did you look at this?

I implemented this template mechanism for these kinds of features. that might be the better choice in this case

Core/src/Seeding2/BroadTripletSeedFilter.cpp
```diff
@@ -348,6 +348,49 @@ void BroadTripletSeedFilter::filterTripletsMiddleFixed(
...
+
+      const float tM = spM.time();
+      const float tB = spB.time();
+      const float tT = spT.time();
+
+      if (!std::isnan(tB) && !std::isnan(tM) && !std::isnan(tT)) {
+        const float varianceTM = spM.varianceT();
+        const float t0M = tM - std::sqrt(spM.zr()[0] * spM.zr()[0] +
+                                         spM.zr()[1] * spM.zr()[1]) /
+                                   Acts::PhysicalConstants::c;
```
andiwand, line 365: note that `c=1` in ACTS and we usually don't write the constant into formulas for readability reasons. not wrong but for consistency reasons I would tend to remove them

Core/src/Seeding2/BroadTripletSeedFilter.cpp
```diff
@@ -348,6 +348,49 @@ void BroadTripletSeedFilter::filterTripletsMiddleFixed(
   std::size_t numTotalSeeds = 0;
   for (const auto& [bottom, middle, top, bestSeedQuality, zOrigin,
                     qualitySeed] : sortedCandidates) {
+    // calculate chi2 of the triplet time compatibility
+    if (config().applyTripletTimeFilter) {
```
andiwand, line 352: I believe this filtering should actually be done in the triplet seed finder

andiwand (commented):
hey @wangyanqi191025 - thanks for bringing this to ACTS! I left a few structural comments

Core/src/Seeding2/DoubletSeedFinder.cpp
wangyanqi191025: Thanks @andiwand for having a look at this PR. CPU consumption is what we also worried about, for which we don't want to influence seeding without time too much. May I ask where can I find the template mechanism you mentioned so I can make some change to this PR? Also I plan to do more test on CPU consumption. Thanks!

andiwand: hey @wangyanqi191025 !

the templating mechanism you can find here
- for the doublet finder: https://github.com/acts-project/acts/blob/b0405619b0faea970326fec66ece6ec9f1f81041/Core/src/Seeding2/DoubletSeedFinder.cpp#L23-L25
- for the triplet finder: https://github.com/acts-project/acts/blob/main/Core/src/Seeding2/TripletSeedFinder.cpp#L27-L28

this allows to generate different version of the seeding code and dispatch to it depending on what is available during runtime. I imagine that we could use this mechanism to use time information and only emit code when needed

paulgessinger:
what's the plan here @jburzy @wangyanqi191025 ?

wangyanqi191025:
Hi Paul @paulgessinger , I am trying to apply the suggestions from Andreas. Timeline may be the end of September or the start of October.

wangyanqi191025:
blocked by #6101

PR #5527, opened 2026-06-01
chore: cleanup geometry building examples

asalzburger:
This PR cleans out the geometry building python files and removes all broken gen2 geometry building tests.

Instead it allows to build:
Gen1 / Gen3 / Detray (via Gen3) / Geant4

--- END COMMIT MESSAGE ---


Examples/Scripts/Python/geometry_validation.py
```diff
@@ -0,0 +1,121 @@
...
+        type=str,
+        default="gen3",
+        choices=["gen1", "gen3", "detray", "geant4"],
+        help="Convert to detray detector and run detray navigation and propagation",
+    )
+
+    # Set log level for the geometry construction and conversion
+    p.add_argument(
+        "--log-level",
+        type=int,
```
andiwand:
```suggestion
        type=acts.logging.Level,
```
would that also work? in that case we can drop `getLogLevel`

asalzburger: I wouldn't know how to change it then on the command line ... as acts.logging.Level does not have a construction from int on the python side, but I think we should have it ... let me check

andiwand (approved):

PR #5528, opened 2026-06-01
ci: Add pypi deployment and nightly to main repo

benjaminhuth:
This adds the deployment CI scripts from https://github.com/acts-project/pypi-build to the main repo. It will trigger a nightly acts build to the test server, release builds (for now) and custom nightlies are a manually triggered.


.github/workflows/pypi.yml
```diff
@@ -0,0 +1,187 @@
...
+    needs: prepare
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      max-parallel: 1
+      fail-fast: false
+      matrix:
+        os:
+         - ubuntu-latest
+        # - ubuntu-24.04-arm
```
paulgessinger, line 63: Is that something you looked into? Would be useful to have I think.

.github/workflows/pypi.yml
```diff
@@ -0,0 +1,187 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          name: build-version
+
+      - name: print build version
+        if: github.event_name != 'workflow_dispatch' || github.event.inputs.acts_ref == ''
+        run: cat build_version.txt
+
+      #- name: Remote terminal
+      #  run: curl -sSf https://sshx.io/get | sh -s run
```
paulgessinger:
```suggestion
```

.github/workflows/pypi.yml
```diff
@@ -0,0 +1,187 @@
...
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.13"
+
+      - name: Patch version to dev version
+        if: github.event_name != 'workflow_dispatch' || github.event.inputs.acts_ref == ''
+        run: |
+          PATCHED_VERSION="$(cat acts/version_number).$(cat build_version.txt)"
+          echo "Patch to version $PATCHED_VERSION"
+          python3 -m pip install change-wheel-version==0.6.0
```
paulgessinger, line 124: Consider uv (not important)

.github/workflows/pypi.yml
benjaminhuth: not yet, let's enable it in a second step?

.github/workflows/pypi.yml
benjaminhuth: Good point, but I would update this later. the problem is, I cannot really test the workflow as long it is not in main, so I would merge it as-is if possible.

paulgessinger (dismissed):

paulgessinger (dismissed):

paulgessinger (approved):

PR #5529, opened 2026-06-01
ci: harden GitHub Actions and enforce hash-pinning via zizmor

paulgessinger:
Add the zizmor GitHub Actions security linter as a pre-commit hook and resolve every finding it reports, then enforce commit-hash pinning going forward.

Tooling / config:
- Add zizmor to .pre-commit-config.yaml.
- Add .github/zizmor.yml: enforce "*": hash-pin for unpinned-uses, and ignore artipacked for release.yml / update-pip-requirements.yml (their checkouts must keep credentials to push). These ignores live in config rather than inline comments so dependabot version bumps can't drop them.
- dependabot: add a 7-day cooldown for the github-actions and pre-commit ecosystems.

unpinned-uses (hash-pin):
- Pin all GitHub Actions to full commit SHAs (with # vX.Y.Z annotations).

dangerous-triggers:
- Document the intentional pull_request_target / workflow_run triggers (backport, pr-lint, milestone, labels, remove-automerge-label, report, sonarcloud) with inline zizmor ignores; none check out or execute untrusted PR code.

template-injection:
- Move ${{ ... }} context expansions out of run blocks into env vars referenced as shell variables (analysis, docs, report, sonarcloud, update-pip-requirements, builds, and the dependencies composite action).

artipacked:
- Set persist-credentials: false on checkouts that don't need the token; the two that push are handled via the config ignores above.

excessive-permissions:
- Add explicit minimal permissions: blocks to every workflow (contents: read by default, permissions: {} for trigger_athena, and targeted write scopes only where a job actually needs them, plus actions: read for cross-run artifact download in report / sonarcloud).

superfluous-actions:
- Replace actions-ecosystem/action-remove-labels with a gh CLI step in remove-automerge-label.yml.


.github/workflows/remove-automerge-label.yml
```diff
@@ -1,15 +1,25 @@
...
+      pull-requests: write
     steps:
-      - uses: actions-ecosystem/action-remove-labels@v1
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          labels: automerge
+      - name: Remove automerge label
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
```
AJPfleger, line 22: do we need these 2?

.github/workflows/trigger_athena.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+permissions: {}
+
```
AJPfleger, line 8: no explicit permissions?

.pre-commit-config.yaml
```diff
@@ -116,3 +116,8 @@ repos:
       files: ^(CITATION\.cff|\.zenodo\.json|AUTHORS)$
       additional_dependencies: ["typer", "rich", "pyyaml"]
       pass_filenames: false
+
+  - repo: https://github.com/woodruffw/zizmor-pre-commit
+    rev: v1.25.2   # check latest
```
AJPfleger, line 121: no hash-check here (and in general for pre-commit-config)? 😈

.github/workflows/trigger_athena.yml
paulgessinger: Doesn't need repo permissions, no.

.pre-commit-config.yaml
paulgessinger: Yeah I need to look into how that would work. I'd consider this out of scope for this PR.

.github/workflows/remove-automerge-label.yml
paulgessinger: I think so since:
- We don't check out the repo so we need to tell the CLI which repo to operate on
- The CLI needs a token to perform actions

AJPfleger (approved):

PR #5530, opened 2026-06-01
feat: expose PlotTool configs in .py, add per-layer particleId to tree

altsybee:
This PR makes two (largely independent) improvements to the ActsExamples I/O and performance QA code:

1. Expose configurations of the plotting tools (`EffPlotToolConfig`, `ResPlotToolConfig` etc.) in the `reconstruction.py`, so binning, pT/eta ranges and other plotting parameters can be configured from .py steering scripts ("full_chain.py") instead of being hard-coded in C++. A small example is added to `full_chain_odd.py`.

Also, `truthPtRangesForEta` and `truthAbsEtaRangesForPt` from `EffPlotToolConfig` are now exposed to the python level and can be changed in .py scripts; a log-pT variant of the efficiency-vs-pT plots is added.

2. Add a per-measurement truth particleId branch to the track summary tree, allowing each measurement on a track to be linked back to the truth particle that produced it (e.g. for per-layer particle-identity studies).

### DIFF
--- a/Examples/Io/Root/include/ActsExamples/Io/Root/RootTrackSummaryWriter.hpp
+++ b/Examples/Io/Root/include/ActsExamples/Io/Root/RootTrackSummaryWriter.hpp
@@ -56,6 +56,12 @@ class RootTrackSummaryWriter final : public WriterT<ConstTrackContainer> {
     std::string inputTrackParticleMatching;
     /// Input jet collection (optional).
     std::string inputJets;
+    /// Input collection of simulated hits (optional, only for truth-linked
+    /// per-measurement particle id branch).
+    std::string inputSimHits;
+    /// Input map from measurement index to sim-hit indices (optional, only
+    /// for truth-linked per-measurement particle id branch).
+    std::string inputMeasurementSimHitsMap;
     /// Output filename.
     std::string filePath = "tracksummary.root";
     /// Name of the output tree.
@@ -102,6 +108,9 @@ class RootTrackSummaryWriter final : public WriterT<ConstTrackContainer> {
   ReadDataHandle<TrackParticleMatching> m_inputTrackParticleMatching{
       this, "InputTrackParticleMatching"};
   ReadDataHandle<TruthJetContainer> m_inputJets{this, "InputJets"};
+  ReadDataHandle<SimHitContainer> m_inputSimHits{this, "InputSimHits"};
+  ReadDataHandle<MeasurementSimHitsMap> m_inputMeasurementSimHitsMap{
+      this, "InputMeasurementSimHitsMap"};
 
   /// Mutex used to protect multi-threaded writes
   std::mutex m_writeMutex;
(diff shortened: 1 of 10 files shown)


Examples/Io/Root/include/ActsExamples/Io/Root/RootTrackSummaryWriter.hpp
```diff
@@ -136,6 +145,12 @@ class RootTrackSummaryWriter final : public WriterT<ConstTrackContainer> {
   std::vector<std::vector<std::uint32_t>> m_measurementVolume;
   /// The layer id of the measurements
   std::vector<std::vector<std::uint32_t>> m_measurementLayer;
+  /// The particle id of the truth particle that left each measurement.
+  /// Outer: one entry per track. Inner: one entry per measurement on
+  /// that track, ordered identically to m_measurementVolume /
+  /// m_measurementLayer (reverse state order). Sentinel value 0 means
+  /// no truth link or no truth inputs configured.
+  std::vector<std::vector<std::uint64_t>> m_measurementParticleId;
```
andiwand, line 153: I fear this gets a bit into the direction of the `RootTrackStatesWriter` which I would rather not duplicate in this writer

Examples/Io/Root/src/RootTrackSummaryWriter.cpp
```diff
@@ -41,6 +43,42 @@ using Acts::VectorHelpers::perp;
...
+    const auto& simHit = *simHits.nth(simHitIdx);
+    const double dep = simHit.depositedEnergy();
+    if (dep > bestDeposit) {
+      bestDeposit = dep;
+      bestId = static_cast<std::uint64_t>(simHit.particleId().hash());
+    }
+  }
+  return bestId;
+}
+}  // namespace
```
andiwand, line 80: this seems like an alternative approach to truth matching which was extracted into `TrackTruthMatcher` some time ago to be consistent across all writers. I would discourage us to split this logic again. So I would rather upstream this there but also check against what https://github.com/acts-project/acts/pull/5515 is doing

Python/Examples/python/reconstruction.py
```diff
@@ -24,6 +24,19 @@
...
+# Duplication Plots
+DuplicationPlotToolConfig = acts.examples.root.DuplicationPlotToolConfig()
+# Efficiency Plots
+EffPlotToolConfig = acts.examples.root.EffPlotToolConfig()
+# Fake rate Plots
+FakePlotToolConfig = acts.examples.root.FakePlotToolConfig()
+# TrackQuality Plots
+TrackQualityPlotToolConfig = acts.examples.root.TrackQualityPlotToolConfig()
+# Track Summary Plots
+TrackSummaryPlotToolConfig = acts.examples.root.TrackSummaryPlotToolConfig()
```
andiwand, line 38: the python scrips are already messy, I fear adding global variables will hurt this even further. I would either use custom `addAlgorithm` functions on the caller side with the correct params or pass them through names arguments on the high-level `add*` functions

PR #5531, opened 2026-06-02
feat: Extract `MultiAxis` from `Grid`

andiwand:
Extracts the handling of multiple axis from `Grid` into its own `IMultiAxis` abstraction with a concrete templated `MultiAxis` implementation. Use this new abstraction in `Grid` as much as possible.

other changes
- cleanup includes
- simplify `MultiAxisHelper.hpp`

--- END COMMIT MESSAGE ---

blocked by
- https://github.com/acts-project/acts/pull/5534
- https://github.com/acts-project/acts/pull/5535
- https://github.com/acts-project/acts/pull/5537


Core/include/Acts/Utilities/Grid.hpp
```diff
@@ -525,21 +552,7 @@ class Grid final : public IGrid {
...
-        current_size *= value + 2;
-      }
-    }
-    // ignore under-and overflow bins for each axis and multiply all bins
-    else {
-      for (const auto& value : nBinsArray) {
-        current_size *= value;
-      }
-    }
-    return current_size;
```
asalzburger, line 543: That's actually nice that this is now in the multi axis.

Core/include/Acts/Utilities/IMultiAxis.hpp
```diff
@@ -0,0 +1,511 @@
...
+/// multi-index (one local bin index per axis) or via a single flattened global
+/// bin index. As for @c IAxis, local bin indices start at @c 1, with index
+/// @c 0 and <tt>nBins + 1</tt> denoting the underflow and overflow bins of an
+/// axis; flattened global indices include these under-/overflow bins.
+class IMultiAxis {
+ public:
+  /// Small vector type used to hold per-axis values without heap allocation
+  /// for the common low-dimensional cases.
+  template <typename T>
+  using SmallVector = boost::container::small_vector<T, 3>;
```
asalzburger, line 53: How does this relate to the definition in the IGrid:

```
  /// @{
  using AnyIndexType = boost::container::small_vector<std::size_t, 3>;
  /// Type alias for dynamic point type (coordinates as vector of doubles)
  using AnyPointType = boost::container::small_vector<double, 3>;
```

Can we harmonize them ?

Core/include/Acts/Utilities/IMultiAxis.hpp
```diff
@@ -0,0 +1,511 @@
...
+      result.push_back(axis.getNBins());
+    }
+    return result;
+  }
+
+  /// Get the total number of bins in the grid
+  /// @param fullCounter if @c true the under-/overflow bins of every axis are
+  ///        included in the count, otherwise only the regular bins are counted
+  /// @return product of the per-axis bin counts
+  virtual std::size_t getNTotalBins(bool fullCounter = true) const {
```
asalzburger: Shall we have default false or try here ?

I think it depends whether one can make a lookup from `[0, getNTotalBins()]` via global bin access, right ?

Core/include/Acts/Utilities/IMultiAxis.hpp
```diff
@@ -0,0 +1,511 @@
...
+    }
+    return result;
+  }
+
+  /// Get the upper boundary of the grid range along each axis
+  /// @return point holding the maximum of each axis
+  virtual AnyPoint getMaxPointAny() const {
+    AnyPoint result;
+    result.reserve(getNAxes());
+    for (const IAxis& axis : *this) {
```
asalzburger, line 145: Does SonarCloud complain about this *this ? Let's see.

Core/include/Acts/Utilities/IMultiAxis.hpp
```diff
@@ -0,0 +1,511 @@
...
+///
+/// On top of the dynamically sized @c IMultiAxis API this adds a statically
+/// sized API (using @c std::array of fixed length @c DIM) and the grid index
+/// conversions between points, multi-indices and flattened global indices.
+/// The actual axis storage is provided by the concrete @c MultiAxis derived
+/// class.
+///
+/// @tparam _DIM number of axes (dimension of the grid)
+template <std::size_t _DIM>
+class IMultiAxisND : public IMultiAxis {
```
asalzburger: With the Range, we called id `XD` instead of `ND` - I don't care, but I usually prefer consistent naming.

asalzburger (changes requested):
Some clarification and probably a discussion about consistent naming XD/ND ?

Core/include/Acts/Utilities/IMultiAxis.hpp
andiwand: I don't think it does. do you have an alternative in mind?

asalzburger (dismissed):
Thanks for implementing the comments.

Core/include/Acts/Utilities/IMultiAxis.hpp
```diff
@@ -0,0 +1,511 @@
...
+///
+/// On top of the dynamically sized @c IMultiAxis API this adds a statically
+/// sized API (using @c std::array of fixed length @c DIM) and the grid index
+/// conversions between points, multi-indices and flattened global indices.
+/// The actual axis storage is provided by the concrete @c MultiAxis derived
+/// class.
+///
+/// @tparam _DIM number of axes (dimension of the grid)
+template <std::size_t _DIM>
+class IMultiAxisXD : public IMultiAxis {
```
paulgessinger, line 367: We have diverging conventions here: in `Vector2` etc we decided not to add `D(imension)` at the end, but `RangeXD` / `Range1D` already does this.

Core/include/Acts/Utilities/IMultiAxis.hpp
```diff
@@ -0,0 +1,511 @@
...
+  /// @return global bin index of the bin containing the point
+  virtual FlatIndex getFlatIndexFromPoint(const Point& point) const {
+    return getFlatIndexFromMultiIndex(getMultiIndexFromPoint(point));
+  }
+
+  /// Determine the flattened global bin index from a multi-index
+  /// @param multiIndex local bin indices along each axis (under-/overflow bins
+  ///        are allowed)
+  /// @return global bin index of the bin
+  virtual FlatIndex getFlatIndexFromMultiIndex(
```
paulgessinger: So far the terminology was `local` vs `global` I think. Should we stick to that?

Core/include/Acts/Utilities/IMultiAxis.hpp
andiwand: yeah not sure - I would prefer `N` over `X`. the `D` I also don't think we need for the concrete dimension. I think it is rather confusing as it could indicate `double`.

since we already have `RangeXD` and `Range1D` I agree that it makes sense to stick to that. if we want to change that I would propose to push this to a future consistent breaking rename

Core/include/Acts/Utilities/IMultiAxis.hpp
andiwand: ah yes that is a good point we should discuss here. I just sneaked this in right now.

my argument is that `local` / `global` does not really mean anything. these terms we use for positions and they do not indicate to me that one is an index vector and one is a scalar index.

happy to discuss this here further or in the meeting today

cc @benjaminhuth @AJPfleger @paulgessinger

AJPfleger: Happy to discuss in the meeting. I have strong opinions on what local and global means but too little understanding of this snippet right now to see how it connects 😂

Core/include/Acts/Utilities/IMultiAxis.hpp
asalzburger: The Range ones are actually not too much used outside, I am happy to drop/rename even if breaking.

Core/include/Acts/Utilities/IMultiAxis.hpp
```diff
@@ -0,0 +1,538 @@
...
+                                                 const IAxis& axis2);
+
+  /// Factory method to create a multi-axis of dimension 3
+  /// @param axis1 the first axis of the grid
+  /// @param axis2 the second axis of the grid
+  /// @param axis3 the third axis of the grid
+  /// @return unique pointer to the created multi-axis
+  static std::unique_ptr<IMultiAxisXD<3>> create(const IAxis& axis1,
+                                                 const IAxis& axis2,
+                                                 const IAxis& axis3);
```
andiwand: added these factory functions now @asalzburger @paulgessinger 

I first wanted to go with a generic approach but ended up exploding build times. I imagine we will only use 1D, 2D, 3D anyways and that allows to push it to cpp.

asalzburger (dismissed):

Core/include/Acts/Utilities/Axis.hpp
```diff
@@ -231,7 +231,7 @@ class Axis<AxisType::Equidistant, bdt> : public IAxis {
 
   /// get bin width
   /// @return constant width for all bins
-  double getBinWidth(std::size_t /*bin*/ = 0) const { return m_width; }
+  double getBinWidth(std::size_t /*bin*/) const final { return m_width; }
```
andiwand, line 234: technically a breaking change? @paulgessinger 

this is a sonarcloud complaint right? I ignored that because in principle the input is not necessary in this case. maybe it would be cleaner to use an overload tough

paulgessinger: Yeah, I'll try that. I agree with sonar that it's surprising if the defaults are not consistent.

Core/include/Acts/Utilities/detail/MultiAxisHelper.hpp
```diff
@@ -31,19 +31,22 @@ namespace Acts::detail {
 template <std::size_t DIM>
 class FlatNeighborHoodIndices {
  public:
+  using size_type = std::size_t;
+
```
junggjo9, line 35: Strange that the linter didn't mock about the indentation or my coffee level is not yet high enough and I'm seeing it wrongly

Core/include/Acts/Utilities/detail/MultiAxisHelper.hpp
```diff
@@ -31,19 +31,22 @@ namespace Acts::detail {
...
       return;
-    }
-    std::size_t flatStride = 1;
-    for (long i = DIM - 2; i >= 0; --i) {
-      flatStride *= (nBinsArray[i + 1] + 2);
-      m_flatStride[i] = flatStride;
+    } else {
+      std::size_t flatStride = 1;
+      for (long i = DIM - 2; i >= 0; --i) {
+        flatStride *= (nBinsArray[i + 1] + 2);
```
junggjo9, line 47: Overflow and underflow shall always be assigned?

Core/include/Acts/Utilities/detail/MultiAxisHelper.hpp
```diff
@@ -548,15 +228,20 @@ struct MultiAxisHelper {
...
-    const std::size_t area = 1;
-
-    MultiAxisHelperImpl<sizeof...(Axes) - 1>::getFlatIndexFromMultiIndex(
-        multiIndex, axes, bin, area);
-
+    std::size_t area = 1;
+    forEachAxisReverse(std::index_sequence_for<Axes...>{}, [&](auto I) {
+      constexpr std::size_t i = I;
+      bin += area * localBins[i];
+      area *= std::get<i>(axes).getNBins() + 2;
```
junggjo9: overflow really included in the area?

Core/include/Acts/Utilities/detail/MultiAxisHelper.hpp
```diff
@@ -575,14 +260,45 @@ struct MultiAxisHelper {
   /// dimensions where d is the number of axis objects in the tuple.
   /// @note This could be a under-/overflow bin along one or more axes.
   template <class Point, class... Axes>
-  static std::array<std::size_t, sizeof...(Axes)> getMultiIndexFromPoint(
+  static std::array<std::size_t, sizeof...(Axes)> getLocalBinsFromPoint(
       const Point& point, const std::tuple<Axes...>& axes) {
-    std::array<std::size_t, sizeof...(Axes)> multiIndex{};
+    std::array<std::size_t, sizeof...(Axes)> localBins{};
+    forEachAxis(std::index_sequence_for<Axes...>{}, [&](auto I) {
+      constexpr std::size_t i = I;
```
junggjo9:
```suggestion
    forEachAxis(std::index_sequence_for<Axes...>{}, [&](sd::size_t i) {
      constexpr std::size_t i = I;
```

would that work?

Core/include/Acts/Utilities/detail/MultiAxisHelper.hpp
```diff
@@ -596,58 +312,72 @@ struct MultiAxisHelper {
...
-        bin, axes, area, multiIndex);
+    // Compute the per-axis stride (same layout as getGlobalBinFromLocalBins),
+    // then peel off the local bin indices from the most significant axis down.
+    std::array<std::size_t, sizeof...(Axes)> strides{};
+    std::size_t area = 1;
+    forEachAxisReverse(Seq{}, [&](auto I) {
+      constexpr std::size_t i = I;
+      strides[i] = area;
+      area *= std::get<i>(axes).getNBins() + 2;
+    });
```
junggjo9, line 324: same as above

Core/include/Acts/Utilities/detail/MultiAxisHelper.hpp
andiwand: for now I fear we have to https://github.com/acts-project/acts/issues/5602

Core/include/Acts/Utilities/detail/MultiAxisHelper.hpp
andiwand: similar as above - I fear for now we have to stick to it https://github.com/acts-project/acts/issues/5602

Core/include/Acts/Utilities/detail/MultiAxisHelper.hpp
andiwand: yes good point!

andiwand: ok it only worked for one case. in all the other we need `constexpr i` so it can be used for the `get<i>`

andiwand: did something similar to simplify this now

Core/include/Acts/Utilities/detail/MultiAxisHelper.hpp
andiwand: to me it looks ok

Core/include/Acts/Utilities/IMultiAxis.hpp
```diff
@@ -0,0 +1,563 @@
...
+/// multi-index (one local bin index per axis) or via a single flattened global
+/// bin index. As for @c IAxis, local bin indices start at @c 1, with index
+/// @c 0 and <tt>nBins + 1</tt> denoting the underflow and overflow bins of an
+/// axis; flattened global indices include these under-/overflow bins.
+class IMultiAxis {
+ public:
+  /// Small vector type used to hold per-axis values without heap allocation
+  /// for the common low-dimensional cases.
+  template <typename T>
+  using SmallVector = boost::container::small_vector<T, 3>;
```
paulgessinger, line 53: Not important but could be private.

paulgessinger (dismissed):

junggjo9 (approved):

PR #5534, opened 2026-06-02
chore: Extract `IGrid` into its own header

andiwand:
--- END COMMIT MESSAGE ---


asalzburger (approved):
Trivial move, but good one.

PR #5539, opened 2026-06-02
feat: Add MMG/EdgeLayerConnector for GNN track building

benjaminhuth:


Examples/Scripts/Python/gnn4itk_example.py
```diff
@@ -143,13 +146,21 @@ def runGNN4ITk(
...
-        "level": logLevel,
-        "useOneBlockImplementation": False,
-        "doJunctionRemoval": True,
-    }
-    trackBuilder = CudaTrackBuilding(**trackBuilderConfig)
+    # Stage 3: Track building
+    if useEdgeLayerConnector:
+        trackBuilderObj = EdgeLayerConnector(
+            level=logLevel,
+            nBlocks=512,
```
mimodekjaer: I would change this to nThreads

Plugins/Gnn/include/ActsPlugins/Gnn/EdgeLayerConnector.hpp
```diff
@@ -0,0 +1,56 @@
...
+/// @{
+
+/// Track building stage based on the EdgeLayerConnector algorithm from
+/// the ModuleMapGraph library (CUDA based)
+class EdgeLayerConnector final : public TrackBuildingBase {
+ public:
+  /// Configuration for the EdgeLayerConnector
+  struct Config {
+    /// Number of thread blocks for parallel edge processing
+    std::size_t nBlocks = 512;
```
mimodekjaer: nThreads

Plugins/Gnn/src/EdgeLayerConnector.cu
```diff
@@ -0,0 +1,127 @@
...
+  CUDA_graph<float> graph(spacepointIDsTensor.data(), numSpacepoints,
+                          edgeSrc.data(), edgeTgt.data(),
+                          tensors.edgeScores->data(), numEdges);
+  ACTS_CUDA_CHECK(cudaDeviceSynchronize());
+  ACTS_CUDA_CHECK(cudaGetLastError());
+
+  ACTS_DEBUG("Setup EdgeLayerConnector...");
+  CUDA_edge_layer_connector<float> connector(
+      &graph, m_cfg.weightsCut, static_cast<int>(m_cfg.minHits),
+      static_cast<int>(m_cfg.nBlocks), static_cast<int>(m_cfg.maxHitsPerTrack));
```
mimodekjaer: m_cfg.nThreads

Plugins/Gnn/src/EdgeLayerConnector.cu
```diff
@@ -0,0 +1,127 @@
...
+  // Convert std::int64_t edge indices to int using Tensor for memory management
+  auto srcInt64Ptr = tensors.edgeIndex.data();
+  auto tgtInt64Ptr = tensors.edgeIndex.data() + numEdges;
+
+  auto edgeSrc =
+      Tensor<int>::Create({1, static_cast<std::size_t>(numEdges)}, execContext);
+  auto edgeTgt =
+      Tensor<int>::Create({1, static_cast<std::size_t>(numEdges)}, execContext);
+
+  thrust::copy(thrust::cuda::par.on(stream), srcInt64Ptr,
```
mimodekjaer, line 39: Is there any reason to use thrust instead of cuda's memcpyasync here?
Mostly just to have all the copies done in the same way, if there is no reason.

mimodekjaer (changes requested):
It generally seems fine, the biggest thing is to rename the nBlocks, to avoid misconception, because it is defining the number of threads.

Plugins/Gnn/src/EdgeLayerConnector.cu
benjaminhuth: because we convert between two int types, its no a memcpy.

Examples/Scripts/Python/gnn4itk_example.py
benjaminhuth: Hmm the parameter in ModuleMapGraph is called blocks, but yeah, it probably should be blockSize...

mimodekjaer: Yes I checked the MMG code, the _blocks is the block size. So yes changing it to blockSize is good.

mimodekjaer (approved):
I think it is good now.

PR #5540, opened 2026-06-02
feat: `IAxis` gains `AxisDirection`

andiwand:
Adds metadata to `IAxis`, which is inherited by `Axis`, to describe the quantity the axis records. This is useful for validation and printing, especially when dealing with multiple axis.

--- END COMMIT MESSAGE ---


Core/include/Acts/Utilities/AxisDefinitions.hpp
```diff
@@ -16,6 +16,8 @@ namespace Acts {
 
 /// @enum AxisDirection to specify a local axis direction
 enum class AxisDirection : int {
+  /// Unknown axis direction - unspecified or none of the below
+  AxisUnknown = -1,
```
paulgessinger: Brings back the discussion about an in-band unknown value vs an optional in `Axis`.

andiwand: I thought we already have a placeholder for that. out of laziness I just added one right now but we can also use `std::optional`. I am sort of in the middle and happy with both

I suppose the main difference is
- `std::optional<AxisDirection>` has a bigger footprint
- `AxisDirection` does not express that it has an optional value

andiwand: https://github.com/acts-project/acts/pull/5540/commits/1cb1a30fa0b05cd481c65082df4ba12469a2e152 switches to `std::optional`

paulgessinger: I think footprint for these types is probably tolerable to increase. I'd also lean towards out-of-band optional.

paulgessinger (approved):

PR #5542, opened 2026-06-03
feat: AnyOf polymorphic value

paulgessinger:
Extends our Any type to be able to handle upcasting


Core/include/Acts/Utilities/Any.hpp
```diff
@@ -740,6 +827,19 @@ using Any = AnyBase<sizeof(void*), true>;
...
+/// @details Stores any value of a type @c T convertible to @c Base* (i.e.
+///          publicly and unambiguously derived from @c Base) and hands it back
+///          as @c Base& / @c Base* through @ref AnyBase::asBase,
+///          @ref AnyBase::operator* and @ref AnyBase::operator-> without naming
+///          @c T. The concrete type can still be recovered with
+///          @ref AnyBase::as / @ref AnyBase::asPtr. Copyable; values up to
+///          @c sb_size bytes are stored inline, larger ones on the heap.
+/// @tparam Base Common base class exposed by the accessors
+/// @tparam sb_size Size of the small-buffer-optimization storage
+template <typename Base, std::size_t sb_size = 48>
```
paulgessinger: Debating whether this should be `sizeof(void*)` like in the general case.

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -438,6 +442,80 @@ class AnyBase : public AnyBaseAll {
...
+  template <typename B = Base>
+    requires(!std::is_void_v<B>)
+  const B* asBase() const {
+    if (m_handler == nullptr) {
+      return nullptr;
+    }
+    // upcast only adjusts the pointer (base-subobject offset); it never touches
+    // the pointee, so casting away const here is safe and the const is restored
+    // in the return type.
+    return m_handler->upcast(const_cast<void*>(dataPtr()));
```
andiwand: an alternative would be to overload `upcast` with a `const void *`

Core/include/Acts/Utilities/Any.hpp
andiwand: I think it can. I imagine that we will rather use the templated version downstream and just define this here for illustration

andiwand (commented):
😍

I imagine this will make value semantic type erasure way easier downstream

Core/include/Acts/Utilities/Any.hpp
paulgessinger: Yes, indeed. Do you prefer this?

andiwand: both is fine from my side

Core/include/Acts/Utilities/Any.hpp
paulgessinger: Changed.

Core/include/Acts/Utilities/Any.hpp
paulgessinger: Added separate upcast pointer.

paulgessinger: Added separate upcast pointer.

andiwand (dismissed):

andiwand (approved):

PR #5543, opened 2026-06-03
chore: Revise interpolation implementation

andiwand:
Mainly makes use of C++20 to get rid of template specialization using `if constexpr` and bit utilities as short hands.

--- END COMMIT MESSAGE ---


Tests/UnitTests/Core/Utilities/InterpolationTests.cpp
```diff
@@ -110,18 +106,15 @@ BOOST_AUTO_TEST_CASE(interpolation_mixed_point_values) {
...
-  CHECK_CLOSE_REL(interpolate((p << 2.3).finished(), low, high, v), 23., 1e-6);
+  const Point2 low{1.};
+  const Point3 high{2.};
+  const Values v{{10., 20.}};
+
+  CHECK_CLOSE_REL(interpolate(Point1{1}, low, high, v), 10., 1e-6);
+  CHECK_CLOSE_REL(interpolate(Point1{1.3}, low, high, v), 13., 1e-6);
+  CHECK_CLOSE_REL(interpolate(Point1{1.5}, low, high, v), 15., 1e-6);
+  CHECK_CLOSE_REL(interpolate(Point1{1.8}, low, high, v), 18., 1e-6);
+  CHECK_CLOSE_REL(interpolate(Point1{2}, low, high, v), 20., 1e-6);
```
AJPfleger, line 117: It looks like there have been 2 cases dropped.

Tests/UnitTests/Core/Utilities/InterpolationTests.cpp
```diff
@@ -26,82 +25,79 @@ BOOST_AUTO_TEST_CASE(interpolation_1d) {
...
-  CHECK_CLOSE_REL(interpolate(Point({{2.3}}), low, high, v), 23., 1e-6);
+  const Point low{1.};
+  const Point high{2.};
+  const Values v{10., 20.};
+
+  CHECK_CLOSE_REL(interpolate(Point{1.}, low, high, v), 10., 1e-6);
+  CHECK_CLOSE_REL(interpolate(Point{1.3}, low, high, v), 13., 1e-6);
+  CHECK_CLOSE_REL(interpolate(Point{1.5}, low, high, v), 15., 1e-6);
+  CHECK_CLOSE_REL(interpolate(Point{1.8}, low, high, v), 18., 1e-6);
+  CHECK_CLOSE_REL(interpolate(Point{2.}, low, high, v), 20., 1e-6);
```
AJPfleger, line 36: Here also 2 cases missing.

Tests/UnitTests/Core/Utilities/InterpolationTests.cpp
andiwand: yes these fire the new `assert` I added because we probe the interpolation outside its range of definition. same below

AJPfleger (dismissed):

AJPfleger (approved):

PR #5545, opened 2026-06-04
draft: try `dynamic_cast` for type checks once again

andiwand:
--- END COMMIT MESSAGE ---


andiwand:
the design diverged quite a bit from what I remembered. we have podio specific data handles and also python ones which rightfully use a common type checking mechanism, more general than the previous one. so I don't think we want to go back from that. for other cases like the dynamic space point columns removing `final` from the class still makes sense to circumvent the problem, but I will do this in a different PR.

Issue #5401, opened 2026-04-30
Rework digitization for Gen3 geometry


benjaminhuth:
The workflow with the Geometry-Id-Hook that assigns the extra bit for different rings in the ODD won't work straight forward with Gen3. In that context, a more general rework of the digitzation could be done, which also would improve several other aspects.

Goals are:
- Decouple the digi config from the geometry hierarchy, make it stable against changes in geometry ID numbering
- Try to avoid redundant information on surface size in the config file
- Future: Allow to use encoded information on digitzation from DD4Hep etc.

Proposed steps for ODD:
- [ ] Encode subsystem information in the ODD detector element / DD4hep detector element (walk DD4hep tree on construction)
- [ ] In digitization: Use `dynamic_cast` (or a design interface) to retrieve subsystem
- [ ] Only encode pitch of subsystems in json file, located in ODD repository (independent of geo id)
- [ ] Keep compatibility with current workflow to ensure correctness

@andiwand @paulgessinger

PR #5546, opened 2026-06-04
feat: ColliderML reader

benjaminhuth:
Blocked by 

- #5410 
- #5557


Examples/Io/Parquet/src/ColliderMLInputConverter.cpp
```diff
@@ -0,0 +1,486 @@
...
+  // 2. Parse hits table
+  // ------------------------------------------------------------------
+  const bool needSimHits = m_outputSimHits.isInitialized();
+  const bool needMeasurements = m_outputMeasurements.isInitialized();
+
+  if (!needSimHits && !needMeasurements) {
+    return ProcessCode::SUCCESS;
+  }
+
+  if (m_cfg.geoIdMap.empty() && ctx.eventNumber == 0) {
```
benjaminhuth: put this in constructor

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
```diff
@@ -0,0 +1,486 @@
...
+                        << " not found in tracking geometry — geoIdMap may be "
+                           "stale, regenerate it");
+      return ProcessCode::ABORT;
+    }
+
+    Acts::Vector3 globalPos{static_cast<double>(hxArr->Value(hOff + i)),
+                            static_cast<double>(hyArr->Value(hOff + i)),
+                            static_cast<double>(hzArr->Value(hOff + i))};
+
+    // ColliderML digitized positions are full 3D positions inside the sensor
```
benjaminhuth: Rather then a two-step conversion, I would prefer a one-step conversion with euclidean boundary tolerance with configurable tolerance with a good default (e.g., 5mm)

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
```diff
@@ -0,0 +1,486 @@
...
+  MeasurementParticlesMap measParticlesMap;
+
+  if (needSimHits || needMeasurements) {
+    hitSeq.reserve(static_cast<std::size_t>(nHits));
+  }
+  if (needMeasurements) {
+    measurements.reserve(static_cast<std::size_t>(nHits));
+  }
+
+  for (std::int64_t i = 0; i < nHits; ++i) {
```
benjaminhuth, line 371: separating the loop for simhits and for measurements would simplify the handling of ordering and creation of the index maps I think.

Examples/Io/Parquet/CMakeLists.txt
```diff
@@ -78,3 +78,58 @@ target_link_libraries(
 )
 
 acts_compile_headers(ExamplesIoParquet GLOB include/ActsExamples/**/*.hpp)
+
+# ColliderML input converter — absorbed into libActsPluginArrow like the
+# other Parquet TUs. It needs ExamplesDigitization (for createMeasurement and
+# smearing types), so we add that as a PRIVATE link to ActsPluginArrow and
+# include its headers via genex (not a link-level dep on the OBJECT target).
+add_library(
```
benjaminhuth: I think all of this should go into `ActsExamplesIoArrow_obj`, and be placed under `Examples/Io/Arrow`.

Examples/Scripts/Python/colliderml_performance_plots.py
benjaminhuth, line 1: this should not be part of the checked-in source

Plugins/Arrow/include/ActsPlugins/Arrow/ArrowUtil.hpp
```diff
@@ -90,6 +90,18 @@ class ACTS_ARROW_EXPORT ArrowTable {
   ArrowSchemaHandle schema() const;
   std::string toString() const;
 
+  /// Extract a flat (non-nested) column as a typed vector.
```
benjaminhuth: If we move this into the arrow object libraries, are these helper functions required?

Examples/Io/Parquet/include/ActsExamples/Io/Parquet/ColliderMLInputConverter.hpp
```diff
@@ -0,0 +1,148 @@
...
+#include <string>
+#include <unordered_map>
+
+namespace Acts {
+class TrackingGeometry;
+}
+
+namespace ActsExamples {
+
+/// Packed ColliderML geometry key: det(8b)|vol(8b)|layer(16b)|surface(32b).
```
benjaminhuth: this is an implementation detail and should not be exposed in the header file I think

Examples/Io/Parquet/include/ActsExamples/Io/Parquet/ColliderMLInputConverter.hpp
```diff
@@ -0,0 +1,148 @@
...
+inline std::uint64_t colliderMLGeoKey(std::uint8_t detector,
+                                      std::uint8_t volume, std::uint16_t layer,
+                                      std::uint32_t surface) {
+  return (static_cast<std::uint64_t>(detector) << 40) |
+         (static_cast<std::uint64_t>(volume) << 32) |
+         (static_cast<std::uint64_t>(layer) << 16) |
+         static_cast<std::uint64_t>(surface);
+}
+
+/// Load a ColliderML geometry ID map from a CSV file.
```
benjaminhuth: In fact, we don't load CSV files, but parquet files now

Examples/Configs/colliderml_geo_map.parquet
paulgessinger, line 1: Is this the volume mapping? Should that be parquet? I think there's benefit in having this be ASCII, no? How large is it?

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
```diff
@@ -0,0 +1,493 @@
...
+namespace {
+
+// Return the row-0 (offset, length) for a list column in the table.
+std::pair<std::int64_t, std::int64_t> rowBounds(const arrow::Table& table,
+                                                const std::string& name) {
+  auto col = table.GetColumnByName(name);
+  if (!col) {
+    throw std::runtime_error("ColliderMLInputConverter: missing column '" +
+                             name + "'");
+  }
```
paulgessinger, line 44: This should be checked using the schema mechanism and an authoritative schema (I think we have this already it just needs to be used). I would then probably drop all other defensive checks and trust the schema validation.

We could also try to decompose the schema a bit further so we can centrally define the column types and names, rather than duplicating them between here and the output converter.

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
```diff
@@ -0,0 +1,493 @@
...
+    return g->sigma;
+  }
+  if (const auto* g = fn.target<const Digitization::GaussClipped>()) {
+    return g->sigma;
+  }
+  if (const auto* g = fn.target<const Digitization::Exact>()) {
+    return g->sigma;
+  }
+  return std::nullopt;
+}
```
paulgessinger: I we need this information here we should either provide it as an explicit input (json) or rethink the digitization to make this part of an interface (go away from an opaque function).

Examples/Io/Arrow/src/ColliderMLInputConverter.cpp
```diff
@@ -0,0 +1,493 @@
...
+                               "'");
+    }
+    return arr;
+  };
+
+  auto detArr = getCol.operator()<arrow::UInt8Array>("detector");
+  auto volArr = getCol.operator()<arrow::UInt8Array>("volume");
+  auto layArr = getCol.operator()<arrow::UInt16Array>("layer");
+  auto surfArr = getCol.operator()<arrow::UInt32Array>("surface");
+  auto geoArr = getCol.operator()<arrow::UInt64Array>("acts_geo_id");
```
paulgessinger: Consider using `TypeTag` or a free template function to avoid this.

Examples/Scripts/Python/generate_colliderml_geo_map.py
```diff
@@ -0,0 +1,257 @@
...
+    parser.add_argument(
+        "--tolerance",
+        type=float,
+        default=1.0,
+        help="Normal-vector perpendicular distance tolerance in mm (default: 1.0)",
+    )
+    args = parser.parse_args()
+
+    hits_dir = (
+        args.input / "ttbar_pu200_tracker_hits" / "data" / "ttbar_pu200_tracker_hits"
```
paulgessinger: Hard coded path should probably not be here.

Examples/Scripts/Python/generate_colliderml_geo_map.py
```diff
@@ -0,0 +1,257 @@
+#!/usr/bin/env python3
```
paulgessinger: This is probably fairly slow in python? Might be worth writing in C++ instead.

Plugins/Arrow/include/ActsPlugins/Arrow/ArrowUtil.hpp
```diff
@@ -149,6 +149,17 @@ ACTS_ARROW_EXPORT std::shared_ptr<arrow::Schema> trackSchema();
...
+/// Read a flat (non-event-indexed) Parquet file into an @c ArrowTable.
+///
+/// Unlike @c ParquetDatasetReader, this function reads a single Parquet file
+/// that has no @c event_id column in one shot — suited for static lookup
+/// tables such as geometry-ID maps. All Parquet/Arrow symbols are confined
+/// to @c ArrowUtil.cpp so callers do not need to include Parquet headers.
+///
+/// Throws on I/O or schema errors.
+ACTS_ARROW_EXPORT ArrowTable
+readFlatParquetFile(const std::filesystem::path& path);
```
paulgessinger: Since there's only a single usage side, I would probably argue this doesn't have to be in the common util header.

Python/Examples/tests/test_colliderml_reader.py
```diff
@@ -0,0 +1,128 @@
...
+def colliderml_pu0_sample(tmp_path_factory):
+    """Locate or download a ttbar PU0 ColliderML sample.
+
+    If COLLIDERML_DATA_DIR is set, use that directory directly (expected
+    to contain CERN__ColliderML-Release-1/). Otherwise download via the
+    colliderml CLI (requires network access and uv in PATH).
+
+    Returns the path to the CERN__ColliderML-Release-1 directory.
+    """
+    data_dir_env = os.environ.get("COLLIDERML_DATA_DIR")
```
paulgessinger: Why don't we just produce a file on the fly with the correct schema? I'm not a huge fan of downloading this behind-the-scenes.

paulgessinger (commented):
Thanks!

To be completely honest, with the amount of massaging needed here, it's getting close to the point where I would just conclude that the current ColliderML data content is just not suitable to be used as an input here.

We can still go ahead with this, but I would make it a priority to try to augment the parquet file content to take the extra work out of this implementation, like encoding the local dimensions and a way to map the geometry ids.

/cc @murnanedaniel

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
benjaminhuth: Yeah so I think we only have not-nice solutions here. But I think the canonical source of truth on subspace and sigmas are the digitization config files. I do not want to create a new file for this, and think this is the best we can do to fill in the missing info as of now.

I thought about a sigma interface, but not all smearers have a sigma canonically...

Examples/Io/Arrow/src/ColliderMLInputConverter.cpp
benjaminhuth: I just told claude to not use a `TypeTag` :D 
But can do a free function.

Examples/Scripts/Python/generate_colliderml_geo_map.py
benjaminhuth: good point, but its a do-once task...

Plugins/Arrow/include/ActsPlugins/Arrow/ArrowUtil.hpp
benjaminhuth: Agree, I thought I would have refactored this...

Python/Examples/tests/test_colliderml_reader.py
benjaminhuth: Hmm but this thing is meant to read in the ColliderML file as they are on the internet. So the second-best thing is to store a small sample of ColliderML on CERN ressources...

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
benjaminhuth: yeah good idea, I did not think of schema validation...

Examples/Configs/colliderml_geo_map.parquet
benjaminhuth: I had it in CSV and thought I put it in parquet because its CSV with ten-thousands of lines of samples... but can do CSV as well

Python/Examples/tests/test_colliderml_reader.py
paulgessinger: Or we just generate it: the Arrow schema pretty much guarantees we're testing the right thing, and avoids both these pitfalls.

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
benjaminhuth: @andiwand whats your take on this?

Examples/Io/Arrow/src/ColliderMLInputConverter.cpp
```diff
@@ -0,0 +1,493 @@
...
+                            static_cast<double>(hyArr->Value(hOff + i)),
+                            static_cast<double>(hzArr->Value(hOff + i))};
+
+    // ColliderML digitized positions are full 3D positions inside the sensor
+    // volume. Project with unlimited perpendicular tolerance, then check
+    // that the 2D local position is within the sensor boundary + configurable
+    // Euclidean tolerance. Out-of-bounds means the geoIdMap assigned the wrong
+    // surface.
+    auto localResult =
+        surface->globalToLocal(ctx.geoContext, globalPos, Acts::Vector3{},
```
andiwand: `direction = Acts::Vector3{}` will only work for "nice" surfaces. but might be fair to assume in this case as ODD sensors are all planar. would still be worth pointing this out as these kinds of lines get copied everywhere. in the future hopefully with a comment

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
```diff
@@ -0,0 +1,493 @@
...
+    const double tz = tzArr->Value(hOff + i);
+    const double tt = htArr->Value(hOff + i);
+
+    if (!needMeasurements) {
+      // Without measurements: add all simhits that pass the geoId check.
+      if (needSimHits) {
+        Acts::Vector4 pos4{tx, ty, tz, tt};
+        Acts::Vector4 zero4 = Acts::Vector4::Zero();
+        hitSeq.emplace_back(geoId, barcode, pos4, zero4, zero4,
+                            static_cast<std::int32_t>(i));
```
andiwand: we don't have the direction of the hit, right? so we cannot re-digitize it

Examples/Io/Arrow/src/ColliderMLInputConverter.cpp
```diff
@@ -0,0 +1,493 @@
...
+          "Hit " << i << " geoId " << geoId
+                 << " projected local position outside sensor bounds (tol="
+                 << m_cfg.hitBoundsTolerance
+                 << " mm) — geoIdMap wrong surface, regenerate it");
+      return ProcessCode::ABORT;
+    }
+    const Acts::Vector2& lp = localResult.value();
+
+    DigitizedParameters dParams;
+    for (const auto& param : smearing.params) {
```
andiwand: this happens per hit right now?

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
andiwand: only had a quick look but could we simply schedule the digitization after reading the hits? decoupling the digitization from the converter and breaking the input out. one could accidentally then schedule the geometric one which produces funny output but I would not worry too much for this workflow

Python/Examples/tests/test_colliderml_reader.py
benjaminhuth: okay, convinced. the PU0 download is roughly 350MB because it downloads the whole shard, which is really a lot...

Examples/Configs/colliderml_geo_map.parquet
benjaminhuth: Okay, the CSV is roughly 500KB, vs 131KB, so CSV is acceptable.

Examples/Io/Arrow/src/ColliderMLInputConverter.cpp
benjaminhuth: yeah, we could assert that we have a regular surface at hand here.

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
benjaminhuth: nope

Python/Examples/tests/test_colliderml_reader.py
benjaminhuth: But one caveat: we cannot check them if the GeoID resolving and the local-to-global mapping really works in colliderml. because the data are just different.

Examples/Io/Arrow/src/ColliderMLInputConverter.cpp
benjaminhuth: Hmm I don't see a real benefit to group it by module, maybe this can be done in a refactor if this turns out to be a bottleneck.

Examples/Scripts/Python/generate_colliderml_geo_map.py
benjaminhuth: removed for now

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
benjaminhuth: I would like to keep it like this for now, as redigitization is not easily possible currently.

Examples/Configs/odd_gen1_gen3_geo_id_map.csv
andiwand, line 1: not sure if we want to commit this?

I think colliderml has to come with some mapping. when we have that we could also just remap on the fly

Examples/Configs/odd-digi-smearing-config-notime-gen3.json
andiwand, line 1: I thought we wanted to have https://github.com/acts-project/acts/issues/5401 for this and effectively get rid of this kind of configuration?

having another file for gen3 goes in the opposite direction. also this one is likely to break since developments are still ongoing. also here I wonder if it would be more effective to read that from a colliderml metadata

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ColliderMLInputConverter.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+/// Load a ColliderML geometry ID map from a CSV file.
+///
+/// Expected columns (with header): detector, volume, layer, surface,
+/// acts_geo_id.
+///
+/// @param path  Path to the CSV file.
+/// @return Map from packed ColliderML key to @c Acts::GeometryIdentifier.
+ACTS_ARROW_EXPORT std::unordered_map<std::uint64_t, Acts::GeometryIdentifier>
+loadColliderMLGeoIdMap(const std::filesystem::path& path);
```
andiwand: I thought we only need `ACTS_ARROW_EXPORT` if there is arrow in the signature?

andiwand (commented):
I think it would be more effective to get the issues we work around here fixed in the dataset by augmenting it or by working on the next release. We already talked about a solution for digitization config and also the sensor positions.

I feel like we would commit this and it would break with the next gen3 change and it will most definitely not work for the next release. I believe there has to be a good argument why this is need to live in main now instead of having it in a branch equally accessible for other people.

benjaminhuth:
Thanks for your review @andiwand. Hmm while I share the argument that the dataset is not optimal and requires some work-arounds to be read in, I would strongly argue for merging this to main, and not postponing it to the next ColliderML release that god knows when will happen:
- V1 is out now, people are doing research  with that.
- I would like to enable this in the pip-install-acts build. This is not possible from a private branch

In my opinion the spirit of the examples framework is not to insist on clean designs etc but to allow for R&D. In this particular case, the reader is self-contained and does not compromise other parts of the code.

Of course, once a V2 version is out, this code should be deprecated at some point.

Examples/Configs/odd_gen1_gen3_geo_id_map.csv
benjaminhuth: I think thats not true actually. Similar to the question wether the ODD config files should live in the ACTS or the ODD repo: The right point would be ACTS, as here we can ensure consistency. Similar here: ColliderML cannot come with a mapping, unless it tags e.g. the ACTS release, which would it make again useless here.

Examples/Configs/odd-digi-smearing-config-notime-gen3.json
benjaminhuth: Well, https://github.com/acts-project/acts/issues/5401 is not yet done. 

I mean we can do a on-the-fly conversion with the help of the gen1-gen3 geometry map it that is better?

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ColliderMLInputConverter.hpp
benjaminhuth: good point, probably not necessary.

Examples/Configs/odd_gen1_gen3_geo_id_map.csv
benjaminhuth: We could also generate this though, but this requires DD4hep... So it would be convenient to have it and ensured its validity with a test.

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ColliderMLInputConverter.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+namespace Acts {
+class TrackingGeometry;
+}
+
+namespace ActsExamples {
+
+/// Pre-extracted sigma for a single bound parameter from the digitization
+/// config. Used to avoid re-extracting per hit at runtime.
+struct DigitizationSigmaConfig {
```
benjaminhuth: Better make this member of `ColliderMLInputConverter` class, not in public namespace.

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ColliderMLInputConverter.hpp
```diff
@@ -0,0 +1,173 @@
...
+  Acts::BoundIndices index;
+  double sigma;
+};
+
+/// A digitization config entry paired with its pre-extracted sigma values.
+using DigitizationConfigWithSigmas =
```
benjaminhuth: Better make this member of ColliderMLInputConverter class, not in public namespace.

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ColliderMLInputConverter.hpp
benjaminhuth: I think this function does not need to be exposed in a header file at all.

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ColliderMLInputConverter.hpp
```diff
@@ -0,0 +1,173 @@
...
+///
+/// Modes (controlled by which output keys are set in the config):
+///   - Particles only:  set @c outputParticles, leave hits keys empty.
+///   - + SimHits:       also set @c outputSimHits.
+///   - + Measurements:  also set @c outputMeasurements; requires
+///                      @c trackingGeometry and @c digiConfig.
+///
+/// @note SimHit momentum fields are zero-filled; ColliderML does not
+///       record per-hit momentum.
+class ACTS_ARROW_EXPORT ColliderMLInputConverter : public IAlgorithm {
```
benjaminhuth: Rename to ColliderMLRelease1InputConverter

Examples/Configs/odd-digi-smearing-config-notime-gen3.json
benjaminhuth: Okay, the after some discussion, the consensus is: Don't provide any config file in the PR, provide the conversion scripts, and execute them on the fly.
In this case, this means: We need a function that acts like `digiConfigGen3 = convert(digiConfigGen1, gen1Gen3Map)`.

Examples/Configs/odd_gen1_gen3_geo_id_map.csv
benjaminhuth: Consensus: We provide the mapping script, the map will be provided outside the ACTS repo somewhere.

Plugins/Arrow/include/ActsPlugins/Arrow/ArrowUtil.hpp
```diff
@@ -149,6 +149,12 @@ ACTS_ARROW_EXPORT std::shared_ptr<arrow::Schema> trackSchema();
 /// @c ArrowSimHitOutputConverter.
 ACTS_ARROW_EXPORT std::shared_ptr<arrow::Schema> simHitSchema();
 
+/// Expected schema for the per-event particle table in the ColliderML
+/// dataset format. Contains only the columns that ColliderML provides; the
+/// ACTS @c particleSchema() is a strict superset (it adds @c energy,
+/// @c perigee_d0/z0, @c parent_id which ColliderML omits).
+ACTS_ARROW_EXPORT std::shared_ptr<arrow::Schema> collidermlParticleSchema();
```
benjaminhuth: Both particles and hits schema should live in the colliderml reader header, to be independent of the potentially changing schemas.

Python/Core/src/Surfaces.cpp
benjaminhuth, line 1: I think this is not needed

Python/Examples/tests/root_file_hashes.txt
```diff
@@ -94,3 +94,6 @@ test_truth_tracking_kalman[odd-0.0]__tracksummary_kf.root: f5f74d202235e77f3032e
 test_truth_tracking_kalman[odd-1000.0]__trackstates_kf.root: 8448baae3a7fbbafd4a8401cb633ad1e4782f6951606eb2374c58c5e455ae093
 test_truth_tracking_kalman[odd-1000.0]__tracksummary_kf.root: e10bf72eadec344fc3c20d0a43a82485a00d96730d81d2059ba880d21ac55c0f
 test_strip_space_points[odd]__strip_space_points.root: 52b4e4b1027f4ce7b178205ee5a460e6ed3871d4703fc4f17b67278877996058
+test_colliderml_truth_tracking_pu0__trackstates_kf.root: 7e77849bf4ed7318b62ab80da8c5bbb53d34290feec7eb3bb5d4c74db1ea4b0d
```
benjaminhuth: This does not refer to any test anymore, right?

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
```diff
@@ -0,0 +1,629 @@
...
+    return g->sigma;
+  }
+  if (const auto* g = fn.target<const Digitization::Exact>()) {
+    return g->sigma;
+  }
+  return std::nullopt;
+}
+
+// Packed ColliderML geometry key: det(8b)|vol(8b)|layer(16b)|surface(32b).
+std::uint64_t colliderMLGeoKey(std::uint8_t detector, std::uint8_t volume,
```
benjaminhuth: Is this used? I think we should use the general Acts::GeometryIdentifier for packing.

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
```diff
@@ -0,0 +1,629 @@
...
+    const double tz = tzArr->Value(hOff + i);
+    const double tt = htArr->Value(hOff + i);
+
+    if (!needMeasurements) {
+      // Without measurements: add all simhits that pass the geoId check.
+      if (needSimHits) {
+        Acts::Vector4 pos4{tx, ty, tz, tt};
+        Acts::Vector4 zero4 = Acts::Vector4::Zero();
+        hitSeq.emplace_back(geoId, barcode, pos4, zero4, zero4,
+                            static_cast<std::int32_t>(i));
```
benjaminhuth: I'm not sure this is correct. the index of simhit is not the index in the container, it is the index in the particle (can be evaluated from true time). So the mapping below might be also wrong.

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
```diff
@@ -0,0 +1,564 @@
...
+                           static_cast<std::int32_t>(i));
+    }
+  }
+
+  // Build measSimHitsMap from the sorted SimHitContainer, using the temporary
+  // index() as cross-reference, then reset index() to -1 (undefined).
+  MeasurementSimHitsMap measSimHitsMap;
+  if (needMeasurements && needSimHits && !hitIndexToMeas.empty()) {
+    SimHitIndex sortedPos = 0;
+    for (auto& hit : simHits) {
```
benjaminhuth, line 480: Use `Acts::enumerate`

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
```diff
@@ -0,0 +1,564 @@
...
+      auto it = hitIndexToMeas.find(hit.index());
+      if (it != hitIndexToMeas.end()) {
```
benjaminhuth, line 482: This should never happen. put an assertion rather then an `if`

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ColliderMLRelease1InputConverter.hpp
```diff
@@ -0,0 +1,189 @@
...
+    /// inside the sensor volume; at incidence angles the projected 2D
+    /// coordinate can land a few mm outside the exact sensor boundary.
+    /// Default 5 mm tolerates physical incidence effects while still catching
+    /// wrong-surface assignments from a stale geoIdMap (tens of mm off).
+    double hitBoundsTolerance = 5.0;
+  };
+
+  /// Expected Arrow schema for the per-event particle table in the
+  /// ColliderML Release 1 dataset format.
+  static std::shared_ptr<arrow::Schema> particleSchema();
```
paulgessinger: I would use the schema handle as a return value here as well so it's consistent between C++ and Python, like the shared schemas in ArrowUtil do.

Python/Examples/src/plugins/Arrow.cpp
```diff
@@ -70,6 +71,31 @@ PYBIND11_MODULE(ActsExamplesPythonBindingsArrow, m) {
...
+  ACTS_PYTHON_DECLARE_ALGORITHM(
+      ColliderMLRelease1InputConverter, m, "ColliderMLRelease1InputConverter",
+      inputParticlesTable, inputHitsTable, outputParticles, outputSimHits,
+      outputMeasurements, outputMeasurementSubset, outputMeasSimHitsMap,
+      outputMeasParticlesMap, outputParticleMeasurementsMap, trackingGeometry,
+      digiConfig, geoIdMapPath, geoIdMapSourcePrefix, geoIdMapTargetPrefix,
+      geoIdMap, hitBoundsTolerance);
+
+  m.def(
+      "collidermlRelease1ParticleSchema",
```
paulgessinger: Bind this onto the class as a static rather than on the module.

.pre-commit-config.yaml
```diff
@@ -16,6 +16,7 @@ repos:
     - id: check-yaml
       exclude: \.gitlab-ci.yml$
     - id: check-added-large-files
+      args: ['--maxkb=1000']
```
paulgessinger: Untelated?

CMakePresets.json
```diff
@@ -29,6 +29,7 @@
             "cacheVariables": {
                 "CMAKE_BUILD_TYPE": "RelWithDebInfo",
                 "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
+                "ACTS_ENABLE_LOG_FAILURE_THRESHOLD": "ON",
```
paulgessinger: Unrelated I think.

Examples/Scripts/Python/convert_digi_config.py
```diff
@@ -0,0 +1,162 @@
...
+"""Convert a digitization config JSON from one geometry to another.
+
+Uses a geometry ID map CSV (produced by ``generate_geoid_map.py``) to remap
+the volume IDs in a digitization config JSON file.  The digi config entries
+are keyed by volume, so the conversion builds a volume→volume mapping from
+the full surface-level geo ID map and applies it.
+
+Usage
+-----
+./run_in_env.sh python3 Examples/Scripts/Python/convert_digi_config.py \\
```
paulgessinger: `run_in_env.sh` is local to your setup right?

paulgessinger (dismissed):

paulgessinger (dismissed):

paulgessinger (approved):

PR #5536, opened 2026-06-02
chore: Make `Axis` implementations final

andiwand:
--- END COMMIT MESSAGE ---

PR #5547, opened 2026-06-04
fix: Circumvent symbol visibility issues with templated final classes

andiwand:
As discovered with https://github.com/acts-project/acts/pull/5536 there seems to be a general risk of using `template class final` which makes `dynamic_cast` fail across shared library boundaries. The easiest path to circumvent the problem right now appears to be dropping `final` from the class and making the functions final instead which results almost in the same behavior.

--- END COMMIT MESSAGE ---

References
- https://stackoverflow.com/questions/79192304/xcode-16-breaks-dynamic-cast-for-final-types-defined-in-shared-library
- https://github.com/llvm/llvm-project/issues/120129


jpreston-cern:
Just tested on running GBTS in examples which was originally getting an `std::bad_cast` error from the linking issue, and is now fixed with these changes

paulgessinger (approved):

paulgessinger:
It's annoying that we have to do this, but oh well.

PR #5548, opened 2026-06-05
feat: augment portal merge failures with context

paulgessinger:
### Problem

In gen3 blueprint construction, designating material on a portal face that later gets **merged** during container stacking fails with `PortalMergingException`. The old message — `"Failure to merge portals"` — carried no context (no face, volume, or node), and the failure surfaced deep inside stack-shell construction, making it hard to trace back to the offending material designation.

### Changes

Three layers of diagnostics, each independently tested:

- **Context-rich exception** (`Portal::merge`): `PortalMergingException` can now carry a message. The material check reports the stacking direction, which side(s) carry material, and the offending surface bounds/center, plus a hint to move the designation to an unmerged face.
- **Catch-and-rethrow in stack shells** (`Cylinder`/`CuboidStackPortalShell`): merge failures are rethrown with the face, stack label/direction, and the volumes involved, appending the original reason as "Underlying error".
- **Early, node-scoped detection** (`ContainerBlueprintNode::connectImpl`): before any merge work, the container checks each *merged* face of its child shells for material and fails fast with the blueprint **node path** and a list of all clashes at once. Fused faces (which legitimately carry material) are not flagged. Adds `mergedFaces(AxisDirection)` helpers to both stack shells.

### Tests

- `BlueprintApiTests::MaterialOnMergedPortalThrows` — reproduces the failure at blueprint level, exercising the early node-level detection (D).

### DIFF
--- a/Core/include/Acts/Geometry/BlueprintOptions.hpp
+++ b/Core/include/Acts/Geometry/BlueprintOptions.hpp
@@ -20,6 +20,14 @@ struct BlueprintOptions {
   std::shared_ptr<NavigationPolicyFactory> defaultNavigationPolicyFactory{
       makeDefaultNavigationPolicyFactory()};
 
+  /// If set, material designated on a portal face that must be merged during
+  /// container stacking does not abort construction. Instead the offending
+  /// material is discarded, the merged surface is tagged with a
+  /// @ref MergedMaterialMarker, and a warning is emitted. This is lossy and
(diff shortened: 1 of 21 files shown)


noemina (approved):
This is very nice and useful! I have also made use of it for ATLAS and provides the required info for debugging the geometry construction!
Thank you Paul!

paulgessinger:
image

PR #5549, opened 2026-06-05
fix: Replacing Gen3 onLayer/onContainer can wrap

paulgessinger:
This pull request refactors and generalizes the callback and customization interfaces in the geometry blueprint builder, making them more flexible and consistent. The main change is to allow layer and container customization callbacks to return any type of `BlueprintNode` (not just their specific subclasses), and to clarify and rename related concepts and types accordingly. This improves extensibility for users who may want to wrap or replace nodes with custom implementations.


paulgessinger:
@madbaron I think I mentioned yesterday that `onLayer`/`onContainer` can return arbitraty `BlueprintNode`s, but that was actually subtly broken. This PR fixes it.

andiwand (dismissed):
👍 

linter job is still failing @paulgessinger

andiwand (approved):

PR #5551, opened 2026-06-05
feat(python): add GUNTAM Python-callable seeding example

EliMe5:
feat(python): add GUNTAM Python-callable seeding example

Add a PythonCallable seeding mode that allows reconstruction seeding to be
provided by a Python function.

Expose the event-data bindings needed by Python seeders to construct seeds
from Python, including mutable space point proxies, mutable seed proxies,
seed-spacepoint container assignment, and tests for the ownership path.

Add a GUNTAM ONNX seeding example using the OpenDataDetector workflow. The
integration is implemented for GUNTAM, while the PythonCallable interface
and ONNX example structure are intended to support other externally provided
ONNX seeding models as well.

--- END COMMIT MESSAGE ---

This PR adds:

* `SeedingAlgorithm.PythonCallable` support through the reconstruction helpers
* mutable `SpacePointContainer2` / `MutableSpacePointProxy2` bindings
* mutable `SeedContainer2` / `MutableSeedProxy2` bindings, including seed creation and space-point container assignment
* `IndexSourceLink` constructor and `makeSourceLink` bindings in the examples module
* tests for mutable event-data bindings and seed/space-point container ownership
* a GUNTAM ONNX-based PythonCallable seeding example using the ODD workflow

Local validation:

* rebased onto latest `origin/main`
* `pytest Python/Core/tests/test_event_data.py` passed
* `py_compile` passed on both example scripts
* ran `onnx_seeding.py` with a GUNTAM model for 2 particle-gun events, single thread; full chain completed successfully through Fatras, digitization, seeding, CKF, ambiguity resolution, and vertex fitting

cc @benjaminhuth  @Corentin-Allaire


Examples/Scripts/Python/guntam_transformer_seeder.py
```diff
@@ -0,0 +1,183 @@
...
+
+import acts
+import acts.examples
+
+# Geometry acceptance window matching the GUNTAM training domain.
+_R_MAX = 500.0  # mm
+_Z_MAX = 1000.0  # mm
+
+# Avoid z-degenerate triplets, which can make track-parameter estimation ill-defined.
+_Z_EPS = 0.01  # mm
```
Corentin-Allaire: In general I am not a huge fan of global variable. Those could be parameter of _apply_model_acceptance ?

Examples/Scripts/Python/guntam_transformer_seeder.py
```diff
@@ -0,0 +1,183 @@
...
+        self._seeds_handle.initialize(seeds_key)
+
+        sess_options = ort.SessionOptions()
+        sess_options.intra_op_num_threads = num_threads
+        sess_options.inter_op_num_threads = num_threads
+
+        self._session = ort.InferenceSession(
+            model_path,
+            sess_options=sess_options,
+            providers=["CPUExecutionProvider"],
```
Corentin-Allaire: You probably could add an option to switch between CPU and GPU ?

Examples/Scripts/Python/guntam_transformer_seeder.py
```diff
@@ -0,0 +1,183 @@
...
+        container = acts.SeedContainer2()
+        container.assignSpacePointContainer(sp)
+        seed = None
+        for triplet, score in zip(seeds_v, scores_v):
+            sp_indices = [int(orig_idx[j]) for j in triplet]
+            seed = container.createSeed()
+            seed.quality = float(score)
+            seed.vertexZ = 0.0
+            seed.assignSpacePointIndices(sp_indices)
+
```
Corentin-Allaire: Not sure if it is an issue but it could probably be vectorised. By first creating a list of seed of the size of triplet then doing a for each ?

You could also add a score filter here. Usually I only keep score> 0.35 and It removes a lot of fakes.

Corentin-Allaire (commented):
Here are a few comments on the PR. One thionk I am not 100% sure is if we want to push `Examples/Scripts/Python/onnx_seeding.py` as is to the repo ?

Python/Core/src/EventData.cpp
andiwand, line 1: shall we split the binding changes out in to a separate PR? I imagine we can merge these in separately without lots of iterations

andiwand:
drafting for now until we have a conclusion on https://github.com/acts-project/acts/pull/5571 which takes load off the gitlab ci

benjaminhuth:
@EliMe5 Can we disentangle https://github.com/acts-project/acts/pull/5571 and this PR? i.e., one only introducing the necessary bindings with the current ownership model here, and then adding the guard features later in a PR?

EliMe5:
> @EliMe5 Can we disentangle #5571 and this PR? i.e., one only introducing the necessary bindings with the current ownership model here, and then adding the guard features later in a PR?

@benjaminhuth I could try, but the code requires some of the guardrails (particularly the missing `py::keep_alive` I added), which would be hard to untangle with the other changes and bloat this pull request a bit.

benjaminhuth (approved):
Looks good! Only minor comment (maybe also a bit matter of taste): I think right now it onnx_seeding.py suggests some generality which it does not have, its really just running guntam. So maybe merging the two py files into one called `guntam_transformer_seeder.py` makes sense? 
But not critical.

PR #5557, opened 2026-06-07
ci: Turn on Arrow plugin in CI

paulgessinger:
Probably a good idea to actually turn this on in CI 🙈


Plugins/CMakeLists.txt
```diff
@@ -18,4 +17,15 @@ add_component_if(Root PluginRoot ACTS_BUILD_PLUGIN_ROOT)
 add_component_if(DD4hep PluginDD4hep ACTS_BUILD_PLUGIN_DD4HEP)
 add_component_if(EDM4hep PluginEDM4hep ACTS_BUILD_PLUGIN_EDM4HEP)
 
+# Arrow must be registered last. When the examples are built, the Arrow plugin
+# absorbs the Examples/Io/{Arrow,Parquet} converter object files and therefore
+# links Acts::ExamplesFramework, which is dragged into the ActsPluginArrowTargets
+# export set. ExamplesFramework publicly links other Acts components
```
AJPfleger:
```suggestion
# export set. ExamplesFramework publicly links other ACTS components
```

CMakeLists.txt
```diff
@@ -349,7 +349,7 @@ set(_acts_covfie_version 0.15.2)
 set(_acts_vecmem_version 1.24.0)
 set(_acts_annoy_version 1.17.3)
 set(_acts_fastjet_version 3.4.1)
-set(_acts_arrow_version 23.0.0)
+set(_acts_arrow_version 21.0.0)
```
AJPfleger, line 352: why downgrade the min version?

paulgessinger: I was trying to see if we can use a low enough version to get away with LCG builds, but we can't. This version should be the lowest actual Arrow version that we can compile against.

AJPfleger (approved):

PR #5558, opened 2026-06-08
fix: spatial grid accelerator intersection

niermann999:
Handle missed intersections in the spatial grid acceleration structure by instead reverting to projecting the ray origin onto the grid multi-axis. Also adds a number of assertions to make finding issues of a similar sort easier in the future.

--- END COMMIT MESSAGE ---


Detray/core/include/detray/navigation/accelerators/spatial_grid.hpp
```diff
@@ -115,13 +117,33 @@ class spatial_grid_impl : public grid_t {
...
+    // Use ray origin as query position
+    if (!is_inside) [[unlikely]] {
+      // In case the intersection results come in local coordinates
+      if constexpr (std::same_as<decltype(intr_point), query_type>) {
+        intr_point =
+            frame_t::global_to_local(trf, tangential.pos(), tangential.dir());
+      } else {
+        intr_point = tangential.pos();
+      }
     }
```
andiwand, line 147: how can this happen?

looks a bit like a case of the input being already malformed and then we patch it up with a backup plan

niermann999: In backward propagation (so going through a cylinder from the outside in) when the tangential passes outside the cylinder radius, it will miss the cylinder altogether. This results in a negative descriminator in the intersector and the intersector will issue an invalid result.

andiwand: not sure if I get it. in ACTS we handle this by having multiple solutions for the intersection and keep track of them by using a stable index.

andiwand (approved):

Detray/core/include/detray/navigation/accelerators/spatial_grid.hpp
niermann999: There is no geometric solution in this case, the ray just misses the cylinder

andiwand: I see but why would we deal with that in the grid then? shouldn't the intersection just report that there is not valid solution? and we do not probe the grid then

niermann999: I know the track hit a sensitive in the forward pass and we are missing the grid because the tangential is a linear approximation to a highly bent track (pT of 0.5 GeV). So I decided to give it a chance, but I only checked that it stops segfaulting, tbh

Also, I have no abstract interface between the grid and the neighbourhood search code, so I would need to have the grid return an empty in-place bin neighbourhood range object for it. I think, it might be better trying to do the same, homogeneous workload across all threads and just intersect surfaces?

andiwand: right this is what I meant with patching it up in the wrong place. I don't think this should happen in the first place. if the surface was discovered in the forward direction but fails due to tangential cylinder approach during the backward pass, I think there is another problem lurking somewhere causing the observed behavior.

niermann999: The track state is valid (so no `nan`, `inf` etc), but I am indeed not sure if the KF returned a correct track state or if e.g. the direction and position are inaccurate. Anyway, this should make the propagation resilient in such an event.

PR #5559, opened 2026-06-08
feat: Add option to set the propagator max step in Geant4 if needed

pbutti:
This PR adds the possibility to specify the maximum length for the Geant4 propagator stepper. 
By default is set to `inf` which means it will just to use Geant4 default (100m), so no change is expected when keeping the configuration as is. 

The reason to tune the length of the stepper is to avoid Geant4 to evaluate the field outside the passed magnetic field map during Step0. To my understanding during step0 G4 might just try to check what possible intersections to a certain volume a particle might have so in the condition of a volume that over-extend the magnetic field map it might be that we see several ERRORs from retrieving the B Field in the wrong point. This happens, for example, for ALICE3 current detector layout, where the beampipe is very long in z and over-extend the bfield map and causes BField retrieval errors for high-eta particles. I'm attaching a log.

```
*********************************************************************************************************
* G4Track Information:   Particle = pi+,   Track ID = 1,   Parent ID = 0
*********************************************************************************************************

Step#    X(mm)    Y(mm)    Z(mm) KinE(MeV)  dE(MeV) StepLeng TrackLeng  NextVolume ProcName
    0 0.000266 3.81e-06 -4.1e-05  6.68e+03        0        0         0 av_140_impr_2_A3IP_VACUUM_pv_95 initStep
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at 0.000265938 3.80627e-06 -4.10313e-05
17:34:51    MagneticFiel   ERROR     PF:: Field evaluated at -1852.72 -93.5001 12511.9
17:34:51    MagneticFiel   ERROR     Field lookup error: MagneticFieldError:1 at -1852.72,-93.5001,12511.9
17:34:51    MagneticFiel   ERROR     PF:: Field evaluated at -2867.47 2152.43 18583.7
17:34:51    MagneticFiel   ERROR     Field lookup error: MagneticFieldError:1 at -2867.47,2152.43,18583.7
17:34:51    MagneticFiel   ERROR     PF:: Field evaluated at -6575.85 -24680.1 52686.9
17:34:51    MagneticFiel   ERROR     Field lookup error: MagneticFieldError:1 at -6575.85,-24680.1,52686.9
17:34:51    MagneticFiel   ERROR     PF:: Field evaluated at -5434.25 -92119.1 67926.6
17:34:51    MagneticFiel   ERROR     Field lookup error: MagneticFieldError:1 at -5434.25,-92119.1,67926.6
17:34:51    MagneticFiel   ERROR     PF:: Field evaluated at -7367.55 -80713.3 77071.4
17:34:51    MagneticFiel   ERROR     Field lookup error: MagneticFieldError:1 at -7367.55,-80713.3,77071.4
17:34:51    MagneticFiel   ERROR     PF:: Field evaluated at -9863.07 15729.9 61090.3
17:34:51    MagneticFiel   ERROR     Field lookup error: MagneticFieldError:1 at -9863.07,15729.9,61090.3
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -11.0633 -0.558332 74.7147
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -16.5992 -0.754711 112.072
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -44.2738 -1.64401 298.859
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -49.1866 -1.75311 332.067
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -55.3407 -1.87018 373.575
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -55.3535 -1.87146 373.573
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at 0.000265938 3.80627e-06 -4.10313e-05
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -0.958611 -0.0483873 6.4755
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -1.43808 -0.0719609 9.71326
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -3.83546 -0.189138 25.9021
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -4.26167 -0.209607 28.7801
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -4.79445 -0.235041 32.3776
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -4.79446 -0.235041 32.3776
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at 0.000265938 3.80627e-06 -4.10313e-05
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -0.958567 -0.048385 6.4752
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -1.43802 -0.0719576 9.71282
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -3.83528 -0.189129 25.9009
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -4.26148 -0.209598 28.7788
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -4.79423 -0.23503 32.3762
17:34:51    MagneticFiel   INFO      PF:: Field evaluated at -4.79424 -0.235031 32.3762
    1    -4.79   -0.235     32.4  6.68e+03 5.63e-11     32.7      32.7 av_140_impr_2_VD_InnerWallCyl_pv_0 Transportation
17:34:51    Geant4Simula   VERBOSE   Check whether volume av_140_impr_2_A3IP_VACUUM_pv_95 is sensitive
```

The error only happens in a specific, probably too big, volume (beampipe) in this case.  The ERRORs do not impact the actual propagation as Geant4 then is able to properly propagate the particle and produce sensible results.
There are few ways to "silence / fix" the error.
- Change the geometry specifications in the .gdml for Geant4 => but might be complicated due to unions, recursive volume references etc etc
- Cap the  propagator extrapolator step to avoid having G4 to evaluate un-physical intersections => easy to transfer to different geometries and can be kept as G4 default.
- Silence the error => my least favourite as I find the ERROR useful to understand problems. 

This PR adds the option to cap the propagator if wanted. After setting it to 500mm:
```
*********************************************************************************************************
* G4Track Information:   Particle = pi+,   Track ID = 1,   Parent ID = 0
*********************************************************************************************************

Step#    X(mm)    Y(mm)    Z(mm) KinE(MeV)  dE(MeV) StepLeng TrackLeng  NextVolume ProcName
    0 0.000266 3.81e-06 -4.1e-05  6.68e+03        0        0         0 av_140_impr_2_A3IP_VACUUM_pv_95 initStep
    1    -4.79   -0.235     32.4  6.68e+03 5.63e-11     32.7      32.7 av_140_impr_2_VD_InnerWallCyl_pv_0 Transportation
11:59:49    Geant4Simula   VERBOSE   Check whether volume av_140_impr_2_A3IP_VACUUM_pv_95 is sensitive
    2    -4.99   -0.244     33.7  6.68e+03    0.457     1.36      34.1 av_140_impr_2_A3IP_VACUUM_pv_95 Transportation
```

--- END COMMIT MESSAGE ---

@altsybee


Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
```diff
@@ -303,6 +305,30 @@ Geant4Simulation::Geant4Simulation(const Config& cfg,
 
 Geant4Simulation::~Geant4Simulation() = default;
 
+ProcessCode Geant4Simulation::initialize() {
+  auto ret = Geant4SimulationBase::initialize();
+  if (ret != ProcessCode::SUCCESS) {
+    return ret;
+  }
```
andiwand: looks like this could be done in `Geant4SimulationBase::initialize`

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
```diff
@@ -303,6 +305,30 @@ Geant4Simulation::Geant4Simulation(const Config& cfg,
...
+  if (std::isfinite(m_cfg.propagatorLargestAcceptableStep)) {
+    ACTS_LOG_WITH_LOGGER(
+        this->logger(), Acts::Logging::INFO,
+        "Setting Largest Acceptable Step for in-field propagator to "
+            << m_cfg.propagatorLargestAcceptableStep << " mm.");
+    G4PropagatorInField* propagator =
+        G4TransportationManager::GetTransportationManager()
+            ->GetPropagatorInField();
+    propagator->SetLargestAcceptableStep(m_cfg.propagatorLargestAcceptableStep *
+                                         Acts::UnitConstants::mm);
```
andiwand: is there only a singleton interface to this? in the past I tried to avoid relying on that and encapsulate these things into `Geant4Manager` and `Geant4Handle`

andiwand (commented):
other things we did in the past is cutting generated particles and filter these ones out. apart from that one could just define the magnetic field outside the detector. the kill volume would then handle it after the successful step. but having the step length adjustable is also generally good

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
pbutti: Happy to move it in Geant4SimulationBase if you prefer. 

- Cutting Generated particles doesn't help in this case: this happens for a pion with 6.7 GeV and |eta| < 3 so in principle in a reasonable acceptance. 
- For the magnetic field I thought about returning 0 outside the map. Do you think I could add that? But I still like the ERROR as it is quite informative
- The kill Volume doesn't handle the case where the BField interpolation is out of bounds in the step0 because the particle is not propagated yet and doesn't cross a boundary. 
- The step length adjustable is the first thing (among few tries) that I found it's working.

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
pbutti: As far as I understand the propagator is available for grab after `runManager.initialize()`, so I think this is a way to get it. 

If you have a better design suggestion, I'm happy to follow it.

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
andiwand: I think it would fit better in the base. but let me check alternative APIs first

- if the particle is $|\eta| < 3$ it still sounds like a geometry / navigation issue?
- one can just decorate the magnetic field provider to have a default value. to me personal the error has no benefit, it usually only pops up if something unrelated broke and labels it as "error" even though nothing bad would happen downstream. the particle would get trashed by G4 or by one of our kill actions.

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
andiwand: looks like what you propose is the best option G4 has to offer. other things rather couple to the geometry or to the particle interaction list.

I think I would lean to rather putting this before the sim is actually executed. it will be set every event then but should ensure nobody else gets a chance to sneak in a different value than we selected

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
pbutti: - Yes I think that it’s due to the current ALICE3 TGeo description that uses volumes that are too big wrt the field map. I think this error only pops in ACTS because we use our magnetic field provider
- I Don’t see this in ODD, which makes me think that things are specific to ALICE3 geometry and not a general problem. I could add to the failed retrieval a return of a 0 field.

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
pbutti: Where do you suggest you put it? 
this still gets set before every event since we only instantiate one manager no? 

Also it’s in the initialise method which is only called once. Maybe I’m not understanding the details of your suggestion :)

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
pbutti: Moved in Geant4SimulationBase::execute

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
pbutti: Moved in Geant4SimulationBase::execute

Examples/Algorithms/Geant4/src/Geant4Simulation.cpp
```diff
@@ -109,6 +111,17 @@ ProcessCode Geant4SimulationBase::execute(const AlgorithmContext& ctx) const {
...
 
+  // If requested cap the max propagator in field to avoid initial stepping
+  // errors in case volumes are larger than field map
+
+  if (std::isfinite(config().propagatorLargestAcceptableStep)) {
+    G4PropagatorInField* propagator =
+        G4TransportationManager::GetTransportationManager()
+            ->GetPropagatorInField();
+    propagator->SetLargestAcceptableStep(
+        config().propagatorLargestAcceptableStep * Acts::UnitConstants::mm);
```
andiwand:
```suggestion
        config().propagatorLargestAcceptableStep / Acts::UnitConstants::mm * CLHEP::mm);
```
Convert from ACTS `mm` to G4 `mm` requires dividing with `Acts::UnitConstants::mm` and then multiplying with `CLHEP::mm`. that said I think in both cases `mm = 1` so it will not make a difference but would be good to be changed for overall correctness

pbutti: Nice catch.

andiwand (dismissed):

andiwand (approved):

PR #5561, opened 2026-06-08
feat: enable reading of detector files for detray in Examples

asalzburger:
Allows to test also written-out files for `geometry_validation` and `propagation_validation`.

--- END COMMIT MESSAGE ---

Screenshot 2026-06-08 at 15 58 24


paulgessinger:
This goes via json right? What's the plan for the detray json format going forward?

niermann999:
> This goes via json right? What's the plan for the detray json format going forward?

In the near future it will stay functional. Many tools that are currently in detray and also traccc unit tests still rely on it. Besides, being able to persist a detray detector still seems quite handy to me.

asalzburger:
> This goes via json right? What's the plan for the detray json format going forward?

Just to be clear - this goes via `detray::writer::json` not the `Acts::toDetrayJson` - which is how we did it for Gen2.

asalzburger:
Here's my current snapshot for `ActsExamples/Propagation`:
- [x] ODD: Gen1/Gen3 ACTS Propagation runs
- [x] ODD: Geant4 reference propagation runs
- [ ] ODD: Detray does not run (neither in-memory nor with file I/O): Gen3 -> Detray drops grid portal links

Also visible in the svg:
Screenshot 2026-06-08 at 16 01 12

Cross-check:
- [x] Detray propagation runs on ToyDetector files runs (generated by detray itself)

::eyes:: @paulgessinger @paradajzblond

Plugins/Detray/include/ActsPlugins/Detray/DetrayGeometryConverter.hpp
```diff
@@ -104,7 +105,16 @@ std::shared_ptr<detray::detector<metadata_t>> toDetray(
       template from_payload<detector_t>(detectorBuilder,
                                         *payloads.surfaceGrids);
 
-  return std::make_shared<detector_t>(detectorBuilder.build(mr));
+  if (!detectorName.empty()) {
+    detectorBuilder.set_name(detectorName);
+  } else if (payloads.names.count(0) > 0) {
```
niermann999:
```suggestion
  } else if (payloads.names.contains(0)) {
```

Python/Plugins/src/Detray.cpp
```diff
@@ -28,27 +33,61 @@ PYBIND11_MODULE(ActsPluginsPythonBindingsDetray, detray) {
...
         });
 
+    detray.def(
+        "readODD",
+        [](vecmem::memory_resource& mr, const std::vector<std::string>& files) {
+          auto cfg = detray::io::detector_reader_config{}.do_check(false);
+          for (const auto& f : files) {
+            cfg.add_file(f);
+          }
+          auto [det, names] =
```
niermann999: If you don't need to use the detector and name map here, can you not just pass on the pair directly?

Examples/Scripts/Python/propagation_validation.py
```diff
@@ -170,18 +173,24 @@ def main():
...
 
-            detrayGeometry = acts.detray.convertODD(
-                __pmr,
-                gContext,
-                trackingGeometry,
-                beampipeVolumeName="BeamPipe",
-                logLevel=logLevel,
-            )
+            if args.input != "":
+                files = sorted(glob.glob(args.input.rstrip("/") + "/*.json"))
```
niermann999: Why do the files need to be sorted? The type of content is encoded in the file header section

asalzburger: Removed.

Python/Plugins/src/Detray.cpp
asalzburger: return directly, was a left-over from a different attempt.

Plugins/Detray/include/ActsPlugins/Detray/DetrayGeometryConverter.hpp
asalzburger: Fixed.

Python/Plugins/src/Detray.cpp
```diff
@@ -28,27 +33,57 @@ PYBIND11_MODULE(ActsPluginsPythonBindingsDetray, detray) {
...
     detray.def(
         "convertODD",
         [](vecmem::memory_resource& mr, const Acts::GeometryContext& gctx,
            const Acts::TrackingGeometry& trackingGeometry,
            const std::string& beampipeVolumeName,
+           const std::string& detectorName = "",
            Acts::Logging::Level logLevel = Acts::Logging::INFO) {
-          return DetrayGeometryConverter::toDetray<DetrayMetaDataODD>(
-              mr, gctx, trackingGeometry, beampipeVolumeName, logLevel);
+          auto [det, names] =
```
niermann999, line 80: Same here?

niermann999 (approved):

asalzburger:
CI errors unrelated to the PR.

PR #5566, opened 2026-06-09
fix: Prevent dangling bfield reference in record propagation function

niermann999:
Remove dangling reference to magnetic field in record_propagation function and use std::optional for type erasure instead.

--- END COMMIT MESSAGE ---


andiwand (approved):

asalzburger (approved):
Also approved.

PR #5567, opened 2026-06-09
feat: Add the option of bin counting histograms for the surfaces material maps

dimitra97:
This PR introduces the possibility of recording the bin counts optionally and create an extra histogram in the material maps for every mapped surface.

It has been enabled by a flag when estimating the total average material to optionally record the bin counts and pass it in the constructor of the `BinnedSurfaceMaterial` (otherwise an empty matrix is passed).

It can be useful for statistical studies (how many bins have been mapped with how many hits etc during the material mapping) as a root macro that I have included can show this use case.

I tried to make it as simple as possible and not to affect much clients' side and experiments that already rely on these functions. 

--- END COMMIT MESSAGE ---

Tagging @asalzburger @noemina


Examples/Scripts/MaterialMapping/Mat_map_stats.C
andiwand, line 1: I wonder if we should add more ROOT scrips or rather start to get rid of them. not that untested Python code is much better but I think we have more integration towards testing for it. `*.C` flags that it is broken to me 😅

dimitra97: if we want to move away from the root script, I can move it to python scrips - no problem :)

Core/include/Acts/Material/BinnedSurfaceMaterial.hpp
```diff
@@ -34,10 +34,12 @@ class BinnedSurfaceMaterial : public ISurfaceMaterial {
   /// @param binUtility defines the binning structure on the surface (copied)
   /// @param materialVector is the vector of material slabs as recorded (moved)
   /// @param splitFactor is the pre/post splitting directive
+  /// @param binCounts is the vector of the bin counts recorded
   /// @param mappingType is the type of surface mapping associated to the surface
   BinnedSurfaceMaterial(const BinUtility& binUtility,
                         MaterialSlabVector materialVector,
                         double splitFactor = 0.,
+                        std::vector<unsigned int> binCounts = {},
```
andiwand, line 42: for this to be non-breaking the parameter would have to be the last one

Core/include/Acts/Material/BinnedSurfaceMaterial.hpp
```diff
@@ -46,10 +48,12 @@ class BinnedSurfaceMaterial : public ISurfaceMaterial {
   /// @param binUtility defines the binning structure on the surface (copied)
   /// @param materialMatrix is the matrix of material slabs as recorded (moved)
   /// @param splitFactor is the pre/post splitting directive
+  /// @param binCounts is the matrix of the bin counts recorded
   /// @param mappingType is the type of surface mapping associated to the surface
   BinnedSurfaceMaterial(const BinUtility& binUtility,
                         MaterialSlabMatrix materialMatrix,
                         double splitFactor = 0.,
+                        std::vector<std::vector<unsigned int>> binCounts = {},
```
andiwand, line 56: same here

Core/include/Acts/Material/AccumulatedSurfaceMaterial.hpp
```diff
@@ -142,8 +142,10 @@ class AccumulatedSurfaceMaterial {
   void trackAverage(const Vector3& gp, bool emptyHit = false);
 
   /// Total average creates SurfaceMaterial
-  /// @return Unique pointer to the averaged surface material
-  std::unique_ptr<const ISurfaceMaterial> totalAverage();
+  /// @param recordBinCounts flag to record the counts of the mapped bins from all the tracks
+  /// @return Unique pointer to the surface material with the bin counts information if enabled
+  std::unique_ptr<const ISurfaceMaterial> totalAverage(
+      bool recordBinCounts = false);
```
andiwand, line 148: I wonder if we really want to pipe everything through the reconstruction abstraction. if these were two different interfaces we could leave the reco untouched

one way out might be to containerize the accumulated material with dynamic columns and trimming that for reconstruction purposes

but all of this is rather long term compared to this change here

cc @asalzburger

asalzburger: Indeed, the GridSurfaceMaterial has already some sort of containerized option. I am sort of waiting until we have the axis work through to clean this one up - probably we wait for the binCounts to added at that stage then?

dimitra97: ok sounds like an option. I draft it for now and go with the  containerized option, once the axis refactoring is also done

PR #5571, opened 2026-06-10
feat(python): add mutable EventData proxy bindings with lifetime safety

EliMe5:
feat(python): add mutable EventData proxy bindings with lifetime safety

Add mutable Python bindings for `SpacePointContainer2`,
`SeedContainer2`, and measurement containers, together with lifetime-safe
proxy and iterator handling.

Python proxies currently hold raw pointers into their backing C++
containers. If the container is consumed by a `std::unique_ptr<T>`
transfer, for example through the whiteboard, the Python wrapper can
remain alive while the C++ object has been moved out. Accessing an
existing proxy then dereferences freed memory.

This PR introduces `ProxyTether`, a binding-side wrapper that keeps the
owning `py::object` and revalidates it before proxy access. If the
backing container has been disowned, access now raises `ValueError`
instead of causing undefined behaviour or a segfault.

Also guard optional `SpacePointColumns` access so missing columns raise
`AttributeError` instead of relying on C++ assertions.

--- END COMMIT MESSAGE ---

## Changes

* Add mutable `SpacePointContainer2` / `MutableSpacePointProxy2` bindings:

  * `createSpacePoint()`
  * `__getitem__`
  * `__iter__`
  * guarded read/write access for scalar and array columns
* Add `SeedContainer2` / `MutableSeedProxy2` bindings:

  * `createSeed()`
  * `assignSpacePointContainer()`
  * `assignSpacePointIndices()`
  * read/write `quality` and `vertexZ`
* Make `MeasurementContainer` / `MeasurementSubset` proxy access use tethered proxies.
* Add checked iterator handling for example flat multimaps.
* Add `py::keep_alive` to `ConstTrackContainer` proxy accessors.

## `ProxyTether`

`ProxyTether` stores the Python owner object, the proxy value, and a type-erased owner-validity check.

Before each proxy access, the owner is re-cast to the backing C++ type. If pybind11's `smart_holder` has disowned the object after a `unique_ptr<T>` transfer, this fails and is translated into a Python `ValueError`.

This handles both relevant lifetime cases:

* the ordinary Python GC path, because the owner `py::object` is retained;
* the whiteboard/disown path, because the owner is revalidated before access.

## Tests

Adds a test-only pybind11 module, `_acts_core_test_bindings`, with helpers that consume `SpacePointContainer2` and `SeedContainer2` via `std::unique_ptr<T>`. This reproduces the same disowning behaviour as the whiteboard without depending on `acts.examples`.

`test_event_data.py` now covers:

* mutable space point proxy read/write,
* all-columns round trip,
* missing-column `AttributeError`,
* seed container/proxy access,
* fail-loud behaviour after disown for space point and seed proxies,
* proxy survival after Python GC.

All added tests pass.

## Known follow-up work

This PR focuses on proxies and iterators whose own backing container has been consumed/disowned. It does not fully solve separate cross-owner or buffer-lifetime cases, including:

* `sourceLinks` sub-iterators whose underlying container is transferred after the iterator is created;
* `SeedContainer2` references to an assigned `SpacePointContainer2` if future Python APIs expose dereferencing that relationship;
* `TrackProxy` objects kept alive after `makeConst()` drains the mutable track container;
* `MeasurementSubset` views after the original `MeasurementContainer` is independently transferred;
* NumPy arrays that outlive move-only backing storage.

Those require separate true-owner or buffer-base lifetime handling and are left for follow-up PRs.

cc @benjaminhuth @andiwand


Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
andiwand, line 1: I am unsure if this complexity is necessary. The idea of the proxies is that the container stays in one place, if that changes it is the users responsibility to create new proxies. This is similar to C++ iterators and how they "invalidate" in certain situations.

since we capture the container with a `std::unique_ptr` it should always stay on the same address and the proxies should be stable after the creation of the container and after the container goes to the whiteboard. I believe during this transition we `std::move` the container which will change the container address?

is it possible to restructure the code so we do not hold on to proxies when this transition happens?

cc @benjaminhuth

andiwand: it might be a similar story like the python specific readers and writers. the question is if we want to buy into an all-the-way python look and feel interface which requires all these wrappers to avoid segfaults and look nice. cc @paulgessinger

Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
```diff
@@ -0,0 +1,168 @@
...
+  return [assign = std::move(assign)](Tether& self, const V& value) {
+    assign(self.checked(), value);
+  };
+}
+
+/// A Python iterator that gates an underlying pybind11 iterator on the owner
+/// still being alive. Used for containers whose iteration yields values/pairs
+/// whose Python conversion is still registered (e.g. the flat multimaps).
+template <typename Container>
+struct CheckedIterator {
```
benjaminhuth: Hmm maybe we can make the naming consistent with the `Tether` naming used above...

andiwand:
drafting for now until we have a conclusion on the discussion which takes load off the gitlab ci

Python/Core/tests/DisownTestBindings.cpp
benjaminhuth, line 1: Not sure if we want to have a C++ file here, I think we can also have other test-related functionality `src` folders (e.g., for histogram creation)

Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
```diff
@@ -0,0 +1,168 @@
...
+/// Return whether `owner`'s wrapped C++ container is still owned (not disowned
+/// by a whiteboard transfer). Implemented by attempting to re-acquire the C++
+/// reference: this throws iff the instance was disowned. Any pending Python
+/// error from the failed cast is cleared so the caller can raise a clean one.
+template <typename Container>
+bool ownerAlive(const pybind11::object& owner) {
+  try {
+    owner.cast<const Container&>();
+    return true;
+  } catch (...) {
```
benjaminhuth, line 42: `(...)` is generally not great, is it possible to name the exception we want to catch here?

Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
```diff
@@ -0,0 +1,168 @@
...
+/// rather than a Container template parameter, so a single bound proxy type can
+/// be produced by several different container types (e.g. a measurement proxy
+/// returned by both MeasurementContainer and MeasurementSubset).
+template <typename Proxy>
+struct ProxyTether {
+  using AliveFn = bool (*)(const pybind11::object&);
+
+  pybind11::object owner;
+  Proxy proxy;
+  AliveFn aliveFn;
```
benjaminhuth, line 59: I wonder if it would be nice to pull the `ownerAlive` function into the `ProxyTether` class, and provide an additional template parameter `Owner`? Are there cases we want another alive function?
Then we also wouldn't handle the case of `aliveFn == nullptr` below.

Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
```diff
@@ -0,0 +1,168 @@
...
+  }
+}
+
+/// Tethers a C++ proxy to its parent container's Python object. The disown
+/// check is type-erased via a function pointer (`&ownerAlive<Container>`)
+/// rather than a Container template parameter, so a single bound proxy type can
+/// be produced by several different container types (e.g. a measurement proxy
+/// returned by both MeasurementContainer and MeasurementSubset).
+template <typename Proxy>
+struct ProxyTether {
```
benjaminhuth, line 54: In general I wonder if a proper class with constructor would make sense here to avoid uninitialized objects...

Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
```diff
@@ -0,0 +1,168 @@
...
+      .def("__next__", &CheckedIterator<Container>::next);
+}
+
+/// A Python iterator over an index-accessible container that yields freshly
+/// tethered proxies. Used for containers whose proxy type is itself bound as a
+/// ProxyTether (so a plain make_iterator over raw proxies would hit an
+/// unregistered type). `maker(owner, container, index)` builds the tethered
+/// proxy for each element.
+template <typename Container>
+struct CheckedIndexIterator {
```
benjaminhuth: Can this not follow the pattern as above of having a `inner` iterator that doe the iteration, and only wraps the creation of the `ProxyTether`? Maybe I'm missing something, but, having an index count here seems odd to me...

benjaminhuth (commented):
Hi, really cool initiative. I think personally it would be cool to have a mechanism like this to avoid segfaults on the python-level that are due to our C++ ownership model.
I think there are still some things to be discussed about the implementation.

Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
benjaminhuth: Hmm not sure we can solve all issues related to this with pointer stability, as we have also `moveToConst`, consuming handles, and potential use of proxies after `s.run()` finishes, when all whiteboards are destroyed. 

So in some cases the memory is literally freed already on the C++ side, and we should not try to keep a pointer to this without beeing attached to the livecycle...

In general I like the approach, maybe we can think if we can reduce complexity in some cases a bit (e.g, iterators)

EliMe5: I agree with you that in C++, iterator invalidation is standard and the user's responsibility. But the reason I added the security is that the existing (before this PR) bindings already invited users to use the bindings in a "python-way" with, for example, the memory arrays to Python users through NumPy. The main danger isn't necessarily a loud segfault, but rather UB/UAF. Because Python doesn't have strict scoping or compile-time lifetime checks like C++, a user can easily save a proxy to a variable and later unknowingly read freed memory or uninitialised columns. This can silently corrupt physics data without crashing the program.

Finally, the reason I created a separate `ProxyTether.hpp` file is to avoid redundant code and to follow the existing utilities pattern. But this is definitely an architectural choice, and I can modify the code accordingly.

Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
andiwand, line 54: `ProxyTether` is not really tied to proxies, right? it could be used for anything that has the same ownership concept

Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
```diff
@@ -0,0 +1,168 @@
...
+  /// Validate the container is still owned, then return the proxy. Throws
+  /// pybind11::value_error (a Python ValueError) if it was consumed.
+  const Proxy& checked() const {
+    validate();
+    return proxy;
+  }
+  Proxy& checked() {
+    validate();
+    return proxy;
+  }
```
andiwand, line 85: for proxies I don't think the non `const` is used? if we widen the scope of this shell it can be

Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
EliMe5: Yes, I tried to keep it general for flexibility, but that is true; the naming then is a bit restrictive. Perhaps I could rename it to `OwnershipTether`?

andiwand: thinking about it further `ProxyTether` might be the best choice for now. the pattern is more generic than for proxies but we will only use it in that case for now I believe

paulgessinger: Doesn't pybind11 give you the ability to mark something as "internal reference"?

benjaminhuth: I don't think so, because the problem is different:
`keep_alive` and `return_value_policy::internal_reference` keep the parent object alive if it goes out of scope and the proxy is still valid.

In our case, we might have the inverse situation:
The python object might still be there, but the underlying C++ memory might be moved away or so.

benjaminhuth: A different approach would be to completely avoid `std::move` on the python -> c++ boundary and to copy always. then livetime should work...

Python/Core/tests/DisownTestBindings.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+namespace py = pybind11;
+
+PYBIND11_MODULE(_acts_core_test_bindings, m) {
+  m.doc() = "Test-only helpers to disown ACTS EventData containers";
+
+  // Taking the argument by std::unique_ptr<T> tells pybind11's smart_holder to
+  // disown the Python wrapper (and frees the object when the unique_ptr goes
+  // out of scope) -- exactly what WhiteBoardRegistry::fromPython does.
+  m.def("consume_spacepoints",
```
benjaminhuth: I think we can do this with the whiteboard actually, it has mechanisms to consume handles. If that for some reason doesn't work, you could put it in the relevant `EventData.cpp`, so we avoid creating an extra cpp file.

Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
EliMe5: I used `catch (...)` because pybind11's `smart_holder` throws an undocumented, implementation-specific exception when casting a disowned object.

Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
EliMe5: I tried it, but it breaks `MeasTether` in `Examples`, which wraps measurements owned by both `MeasurementContainer` and `MeasurementSubset`. Since pybind11 requires a strict 1-to-1 C++ to Python type mapping, keeping `aliveFn` as a runtime function pointer is required here to act as type erasure. 
I added the constructor to ensure `aliveFn` is never uninitialised.

Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
EliMe5: The non-const `checed()` is actively used. `tetheredWrite` and the `assignSpacePointIndices`/`assignSourceLinks` lambdas all call it on mutable proxies.

Python/Core/tests/DisownTestBindings.cpp
EliMe5: I put it in `EventData.cpp` because using the whiteboard would require the tests to depend on the examples, which is hard to justify architecturally. Because the core bindings don't expose direct whiteboard control.

Python/Utilities/include/ActsPython/Utilities/ProxyTether.hpp
EliMe5: We need the index count here because `py::make_iterator` would yield raw C++ proxies, which aren't registered Python types. The index lets us construct the registered `ProxyTether` wrappers on the fly for each element during iteration.

Python/Core/tests/DisownTestBindings.cpp
benjaminhuth: Hmm not sure if I agree, because the test could live in `Python/Examples/tests` and cover all EDM types... But fine for me to do it like that as well.

benjaminhuth (approved):
LGTM

PR #5572, opened 2026-06-10
fix: the perigee stopper

niermann999:
As it is used in an open PR in traccc, but not in detray, this testing actor broke in the meantime

--- END COMMIT MESSAGE ---


stephenswat (approved):
LGTM.

PR #5575, opened 2026-06-11
docs: document Fatras barcode identifiers

### LINKED ISSUE #2796: Write fatras documentation
https://acts.readthedocs.io/en/latest/fatras/fatras.html#id1

Improve documentation of Barcode!

hadronic-arvind:
Closes #2796

### DIFF
--- a/docs/groups/fatras.md
+++ b/docs/groups/fatras.md
@@ -1,0 +1,93 @@
+@defgroup fatras Fatras
+@brief Fast track simulation on ACTS tracking geometries.
+
+# Fatras
+
+Fatras is the ACTS fast track simulation package. It uses the
+@ref Acts::Propagator and a reconstruction-oriented @ref Acts::TrackingGeometry
+to transport particles through the detector and to produce simulated hits on
+sensitive surfaces. It is intended for fast algorithm validation and detector
+studies that do not require a full, detailed detector simulation.
+
+Fatras uses parametrized material interactions, including multiple scattering,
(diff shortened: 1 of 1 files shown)


docs/old/fatras/fatras.md
andiwand, line 1: @paulgessinger this file for example we do not show at all right now? is the correct replacement `docs/groups/fatras.md`, added in this PR? in this case we should migrate more content from old to new and not touch the old one

andiwand (commented):
thanks @hadronic-arvind ! this reads well and represents what is in the code right now

I am not sure about the location and what to do with the "old" file. lets wait for @paulgessinger to have a look

docs/old/fatras/fatras.md
hadronic-arvind: @andiwand do you mean the location of the documentation for the Barcode within the `docs/groups/fatras.md‎` or the location of the document itself? Also, if needed, I can help move content from the old `fatras.md` to the new (after checking if relevant content is up-to-date). 


Also, looks like `docs/groups/fatras.md` feeds the current Doxygen docs, while `docs/old/fatras/fatras.md` still matters because `readthedocs.yml` points through the old docs stack. Since the original issue linked ReadTheDocs, I figured updating only the new docs would have been incomplete.

paulgessinger: This file indeed is only referenced in the old documentation. It's an open question if we want to push this update there. To be honest, I would probably think: no.

andiwand (approved):
LGTM 👍

PR #5581, opened 2026-06-12
fix: Revive mixed digitization in Examples

andiwand:
Mixed digitization (smeared + geometric) is broken right now and results in a `std::vector` out-of-bounds access. I tracked this down to `ModuleClusters::squash` where we first act on the smeared digi and then on the geometric digi. This is the other way around as in other pars of the code and results in the out-of-bounds access as the assumed geometric params are not present yet.

Apart from that this change adds a couple of `const` where possible.

--- END COMMIT MESSAGE ---


Examples/Algorithms/Digitization/src/ModuleClusters.cpp
```diff
@@ -251,29 +253,11 @@ ModuleValue ModuleClusters::squash(std::vector<ModuleValue>& values) {
...
-        mval.paramVariances.at(j) += f2 * other.paramVariances.at(j);
-      }
-    }
-  }
-
-  // Now do the geometric indices
-  Cluster clus;
+  // Do the geometric indices first: they precede the smeared ones in
+  // the parameter vectors filled by the digitization algorithm, so
+  // seeding `mval' with them keeps the slot positions used in the
```
andiwand:
```suggestion
  // seeding `result' with them keeps the slot positions used in the
```

Examples/Algorithms/Digitization/include/ActsExamples/Digitization/ModuleClusters.hpp
```diff
@@ -56,13 +57,17 @@ class ModuleClusters {
   double m_nsigma;
   bool m_commonCorner;
 
-  std::vector<ModuleValue> createCellCollection();
+  std::vector<ModuleValue> createCellCollection() const;
+
   void merge();
-  ModuleValue squash(std::vector<ModuleValue>& values);
+
+  ModuleValue squash(std::vector<ModuleValue>& values) const;
```
benjaminhuth, line 128: Hmm even though this is not a core type, I think  small doc string would be nice what the methods do

Examples/Algorithms/Digitization/src/ModuleClusters.cpp
```diff
@@ -251,29 +253,11 @@ ModuleValue ModuleClusters::squash(std::vector<ModuleValue>& values) {
...
-        mval.paramVariances.at(j) += f2 * other.paramVariances.at(j);
-      }
-    }
-  }
-
-  // Now do the geometric indices
-  Cluster clus;
+  // Do the geometric indices first: they precede the smeared ones in
+  // the parameter vectors filled by the digitization algorithm, so
+  // seeding `result' with them keeps the slot positions used in the
```
benjaminhuth, line 262: So just to ensure I understand this: Slot positions means that the `std::vector`s in `ModuleResult` are filled so we can savely access indices without an out of bounds error?

In terms of order: Couldn't we pre-allocate the memory so the ordering is not so fragile, or do I miss something here?

Examples/Algorithms/Digitization/src/ModuleClusters.cpp
```diff
@@ -313,33 +297,61 @@ ModuleValue ModuleClusters::squash(std::vector<ModuleValue>& values) {
...
 
   if (tot > 0) {
     pos /= tot;
-    var /= (tot * tot);
+    var /= tot * tot;
+  }
+
+  // Without any cells there is no geometric position; skip the
+  // geometric entries instead of emitting a bogus (0, 0) measurement.
+  if (!cluster.channels.empty()) {
```
benjaminhuth: Hmm is that the correct handling here? I havn't checked the full downstream code, but do we skip the measurement at this point?
Also, I would not refer in the comment to a previous faulty behavior

benjaminhuth (commented):
Looks good, thank you for looking into that! A few comments

Examples/Algorithms/Digitization/include/ActsExamples/Digitization/ModuleClusters.hpp
andiwand: I can ask claude to document it because what the functions do is really beyond me and I don't think I want to know 😅

Examples/Algorithms/Digitization/src/ModuleClusters.cpp
andiwand: I thought about pre allocating but ultimately we would have to loop through the geometric and smeared digi to do this. so I think it would would be close to what already happens

Examples/Algorithms/Digitization/src/ModuleClusters.cpp
andiwand: I don't think this is strictly necessary and will not happen with any calling site right now. I think both versions are not great. the one before emits `(0,0)`, this one just drops the unknown quantities. might be better to just throw?

benjaminhuth: throw seems good until we run into something that justifies more complicated treatement...

benjaminhuth (approved):
LGTM

PR #5583, opened 2026-06-12
chore!: Remove deprecated cmake options

benjaminhuth:
Removed deprecated options for CUDA and Python bindings in CMake configuration. Not sure if strictly breaking according to our policy, but now is a good point to remove those.


andiwand (dismissed):
thanks! I would have forgotten about these

andiwand (dismissed):

docs/groups/python_bindings.md
andiwand, line 1: should we break these changes out @benjaminhuth ?

at least right now this is not captured by the PR title / description

benjaminhuth: wtf. I did not add these changes :D I think I mis-clicked on "fix with copilot", because the button is next to "re-run" :D

andiwand (approved):

PR #5586, opened 2026-06-12
feat: measurement and sim-hit tables for the Arrow/Parquet output

murnanedaniel:
Follow-up to #5410: splits tracker-hit output into a reco table and a truth table, linked by row indices.

**`tracker_hits`** - one row per measurement (row index = the id in track `hit_ids`):
- `loc0`, `loc1`, `var_loc0`, `var_loc1` (full bound layout)
- `time`, `var_time`
- `subspace` bitmask (bit0/1/2 = loc0/loc1/time measured)
- `x`, `y`, `z` (reco global)
- `detector`, `volume_id`, `layer_id`, `surface_id`
- `size_loc0`, `size_loc1`, `n_channels`, `sum_activation`, `local_eta`, `local_phi`, `global_eta`, `global_phi`, `eta_angle`, `phi_angle`
- `particle_ids`, `simhit_ids` (nested truth links; merged clusters carry several contributors)

**`tracker_simhits`** - one row per sim-hit, *all* sim-hits, in container order (= the `simhit_ids` indices):
- `true_x`, `true_y`, `true_z`, `true_time`
- `tpx`, `tpy`, `tpz`, `tE`, `dE`
- `particle_id`, `hit_index`
- `detector`, `volume_id`, `layer_id`, `surface_id`

Only `inputMeasurements` + `trackingGeometry` are required: clusters and truth inputs are optional (shape columns zeroed / truth links empty), all collection names configurable, and the `detector` numbering can be supplied as a custom mapping. Python bindings and tests included.

The shape columns give cheap cluster topology without writing out full cell lists (~1M cells/event at ttbar PU200). They currently come out zeroed: clusters reach the converter with empty `channels` even in geometric digitization, though `localParameters` fills them and `ModuleClusters` should preserve them. Am I misreading the flow, or are the cells dropped somewhere? The columns are there either way, so they'll fill in once the cells do.

Validated on the ODD up to ttbar PU200: ~2% merged measurements, <=0.006% unmatched sim-hits, all links resolve.

--- END COMMIT MESSAGE ---

cc @paulgessinger @andiwand @benjaminhuth

### DIFF
--- a/Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ArrowMeasurementOutputConverter.hpp
+++ b/Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ArrowMeasurementOutputConverter.hpp
@@ -44,11 +44,15 @@ class ACTS_ARROW_EXPORT ArrowMeasurementOutputConverter final
   struct Config {
     /// Measurement container (required). One output row per measurement.
     std::string inputMeasurements;
-    /// Cluster container parallel to the measurements (required for the
-    /// shape columns; produced by the geometric digitization).
+    /// Optional cluster container parallel to the measurements (produced by
+    /// the geometric digitization). When unset - e.g. smearing-only
+    /// digitization - the shape columns are emitted as zeros so the schema
+    /// stays stable.
     std::string inputClusters;
-    /// Input @c SimHitContainer (required; used to resolve contributing
-    /// particles for @c particle_ids).
+    /// Optional input @c SimHitContainer; must be set together with
+    /// @c inputSimHitMeasurementsMap. When unset (e.g. data, or a reco-only
+    /// conversion) the @c particle_ids / @c simhit_ids link columns are
+    /// emitted as empty lists.
     std::string inputSimHits;
     /// Optional input particle container used to resolve a contributing
     /// sim-hit's particle barcode to a row index in the corresponding parquet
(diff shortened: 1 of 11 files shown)


Plugins/Arrow/include/ActsPlugins/Arrow/ArrowUtil.hpp
andiwand, line 1: one general question I briefly discussed with @benjaminhuth is should this live in Plugins which is almost on the level of Core, or rather in examples.

this breaks writing colliderml v1 output and we have no handle on having different schemas of the same conceptual object. looking at https://github.com/acts-project/acts/pull/5546 which purpose is to read v1 files we are about to blow this up because there are too many moving parts.

@benjaminhuth suggestion was to move the schema to examples and version it explicitly there

cc @murnanedaniel @paulgessinger

benjaminhuth: So. this should not break ColliderML v1 as of now, because I on purpose not use the shared plugin schemas in the ColliderML reader, see:

https://github.com/acts-project/acts/blob/a0289a6b4d9ca6207781da271e6824deadeeca39/Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ColliderMLRelease1InputConverter.hpp#L116

I think its okay to change the schemas in the plugin, and for ColliderML v2 reader, I would again copy them to the reader.

Whether we should version the schemas in examples, is another question. There are pro and cons here I think.

Examples/Io/Arrow/include/ActsExamples/Io/Arrow/ArrowMeasurementOutputConverter.hpp
```diff
@@ -0,0 +1,109 @@
...
+    std::shared_ptr<const Acts::TrackingGeometry> trackingGeometry;
+    /// Diagnostic only: sim-hits contributing to no measurement are absent
+    /// from this table's links (they remain in the truth table). A warning is
+    /// logged when the fraction exceeds this value.
+    double maxUnmatchedSimHitFraction = 0.001;
+    /// Resolves the @c detector subsystem id for a given geometry id.
+    /// Defaults to reading the geometry id's @c extra byte.
+    std::function<std::uint8_t(Acts::GeometryIdentifier)> detectorResolver =
+        [](Acts::GeometryIdentifier gid) {
+          return static_cast<std::uint8_t>(gid.extra());
```
benjaminhuth, line 81: Hmm is this a good default? In ODD gen1, extra is set to distinguish different endcap rings for digitization currently.

Plugins/Arrow/include/ActsPlugins/Arrow/ArrowUtil.hpp
benjaminhuth: But I agree with @andiwand that I'm not sure the schemas should live in the plugin...

PR #5591, opened 2026-06-15
fix: remove rejected AMVF vertices safely

### LINKED ISSUE #2380: Bug: AMVF `deleteLastVertex` might be wrong
https://github.com/acts-project/acts/blob/main/Core/include/Acts/Vertexing/AdaptiveMultiVertexFinder.ipp#L549

I noticed that `deleteLastVertex` might be bugged.
- we destroy the vertex candidate and access it again
- we refit with the vertex we actually want to delete

I tried fixing it here https://github.com/andiwand/acts/tree/try-fix-amvf-delete-last-vertex

hadronic-arvind:
## Summary
- remove rejected AMVF vertex candidates from fitter state before any refit
- erase the exact rejected vertex from the owning and pointer collections instead of assuming it is the tail element
- add regression coverage for rejecting one of multiple newly seeded AMVF candidates

## Root Cause
`AdaptiveMultiVertexFinder::deleteLastVertex` removed the last owned vertex before using the rejected vertex reference to update fitter state. In multi-seed iterations this could also remove the wrong vertex: rejecting an earlier candidate destroyed the tail candidate, while the subsequent refit still had stale vertex pointers in the fitter state.

The updated cleanup keeps the rejected vertex alive while it is removed from the multimap and vertex collection, removes candidate-specific track/info state, refits only the remaining affected vertices, and then erases the rejected vertex from the finder-owned collections.

Closes #2380

## Tests
- `pre-commit run --files Core/include/Acts/Vertexing/AdaptiveMultiVertexFinder.hpp Core/src/Vertexing/AdaptiveMultiVertexFinder.cpp Tests/UnitTests/Core/Vertexing/AdaptiveMultiVertexFinderTests.cpp`
- `cmake --build /workspaces/acts/build_codex --target ActsUnitTestAdaptiveMultiVertexFinder --parallel 4`
- `ctest --test-dir /workspaces/acts/build_codex -R AdaptiveMultiVertexFinder --output-on-failure`
- `cmake --build /workspaces/acts/build_codex --target ActsUnitTestFullBilloirVertexFitter ActsUnitTestIterativeVertexFinder ActsUnitTestKalmanVertexUpdater ActsUnitTestAdaptiveMultiVertexFitter ActsUnitTestAdaptiveMultiVertexFinder ActsUnitTestZScanVertexFinder ActsUnitTestTrackDensityVertexFinder ActsUnitTestGridDensityVertexFinder ActsUnitTestHoughVertexFinder2 --parallel 4`
- `ctest --test-dir /workspaces/acts/build_codex -R Vertex --output-on-failure`


Tests/UnitTests/Core/Vertexing/AdaptiveMultiVertexFinderTests.cpp
```diff
@@ -671,6 +698,80 @@ BOOST_AUTO_TEST_CASE(
   }
 }
 
+BOOST_AUTO_TEST_CASE(adaptive_multi_vertex_finder_removes_rejected_seed) {
```
andiwand, line 701: it is a bit unfortunate that we have to do so much setup to test this properly. did you try to make it fail as well? differently said, without the change to `Core/src/Vertexing/AdaptiveMultiVertexFinder.cpp` this fails?

andiwand (commented):
thanks you @hadronic-arvind ! this looks like an important fix

to add some more context: currently our seeders only provide a single seed so this should not be a problem there. but the seeder can be a limiting factor for the performance and the only way to overcome this to me was to allow multiple seeds per single iteration.

I left one curiosity comment - otherwise good to go from my side

Tests/UnitTests/Core/Vertexing/AdaptiveMultiVertexFinderTests.cpp
hadronic-arvind: Yeah.

Keeping the new regression test as-is but temporarily restoring the old `deleteLastVertex` behavior in `Core/src/Vertexing/AdaptiveMultiVertexFinder.cpp` makes the focused AMVF test fail. In that case the regression returns zero vertices instead of the expected one:

`check allVertices.size() == 1u has failed [0 != 1]`

and then hits a memory access violation on the stale/bad access path.

With the production change restored, the same focused AMVF test passes again. I also reran the broader vertexing CTest group on the rebased branch, and all 9 vertexing tests pass.

andiwand: great thank you!

andiwand (approved):

andiwand:
Very good to get this out of the way - thank you @hadronic-arvind !

hadronic-arvind:
The PR is up to date with main at a8c3607, but merge-sentinel is blocked because the latest fork CI runs are action_required and CI Bridge refused with "No pipeline was triggered for this user". Could someone with maintainer/CI permissions approve the GitHub Actions runs and trigger CI Bridge for this SHA?

PR #5592, opened 2026-06-15
feat: change to multimask rather than multi links

asalzburger:
This PR changes the standard behavior in the Gen3->detray conversion to use the new multi-mask mechanism instead of multilinks.

In pure ODD navigation, this gives a rough 10 % speed improvement as only one intersection has to be done and checked against several masks. This only works for portal cylinders and disks, and if anything else falls back to the default behavior of one surface per mask.

The output results are identical.

--- END COMMIT MESSAGE ---

detray_multimask_vs_multilink


niermann999 (approved):

niermann999:
Nice! About 10% improvement is also what I seem to remember from the detray standalone benchmarks (but I think without covariance transport)

PR #5498, opened 2026-05-22
chore!: Cleanup `SurfaceArray` from deprecations

andiwand:
Gets rid of most deprecations in `SurfaceArray`. Some functions are still required by the Gen1 geometry so left them mostly untouched.

--- END COMMIT MESSAGE ---

PR #5594, opened 2026-06-15
chore: Elevate axis construction exceptions

andiwand:
We only ever validated the correctness of the axis construction in the `IAxis` helpers. This just fell on my head in https://github.com/acts-project/acts/pull/5498

This PR elevates the validation to the concrete `Axis` constructors.

--- END COMMIT MESSAGE ---


Python/Core/src/Navigation.cpp
```diff
@@ -123,8 +126,8 @@ void addNavigation(py::module_& m) {
 
         auto detElem = std::make_unique<Test::DetectorElementStub>();
 
-        auto surface = Surface::makeShared<CylinderSurface>(
```
benjaminhuth: Is this related?

benjaminhuth (approved):
Looks good to me, I didn't go deep enough to though to see the connection to the changes in the tests...

Python/Core/src/Navigation.cpp
andiwand: yes this causes the new exception to pop. the reason is that we create a surface array on a disc with the only surface being a cylinder so min and max R are the same

PR #5596, opened 2026-06-16
fix: Bit more consistency in usage of particles and particles_selected for…

pbutti:
We noticed that in some tracking algorithm examples configuration "particles_selected", see `addCKFTracks`, while in others "particles" are used instead, see `addAmbiguityResolution`. We propose to be little bit more consistent, even if this code shouldn't be taken as holy grail for setting up a reconstruction pass, but more like an example. 

--- END COMMIT MESSAGE ---


paulgessinger:
Changes outputs. Should we update the references @pbutti @andiwand ?

PR #5598, opened 2026-06-16
feat: VectorHelpers - Add generic eta method

junggjo9:
--- END COMMIT MESSAGE ---


Core/include/Acts/Utilities/VectorHelpers.hpp
```diff
@@ -143,6 +143,20 @@ double eta(const Eigen::MatrixBase<Derived>& v) noexcept
   }
 }
 
+/// Calculate the pseudo rapdity from anything implementing a method
+/// like `theta()` returning anything convertible to `double`.
+/// @tparam T anything that has a theta method
+/// @param v Any type that implements a theta method
+/// @return The pseudo rapidity value
+template <typename T>
+double eta(const T& v) noexcept
```
benjaminhuth, line 152: Hmm why not `double eta(double theta)`, wouldn't that be even more generic?

junggjo9: I was taking the phi implementation as my template here. Yeah, that would be another option

andiwand: in the best case we could have both. we have many implementations for eta -> theta and the other way flying around which could be centralized. but a scalar transformation is out of scope for vector helpers I believe

junggjo9: Do we already have a central eta implementation?

benjaminhuth: Hmm yeah make sense that its not for vector helpers. But I think code like `eta(myFancyType)` which internally calls `myFancyType.theta()` is just less transparent as `eta(myFancyType.theta())`...

So I don't like the `phi` function either I think :D

andiwand: ah we actually have a helper https://github.com/acts-project/acts/blob/0dda19a5707637b756a17c92ec578bc2c8503bad/Core/include/Acts/Utilities/AngleHelpers.hpp#L35-L89

Core/include/Acts/Utilities/VectorHelpers.hpp
junggjo9:
```suggestion
#include "Acts/Utilities/AngleHelpers.hpp"
```

Core/include/Acts/Utilities/VectorHelpers.hpp
```diff
@@ -143,6 +143,20 @@ double eta(const Eigen::MatrixBase<Derived>& v) noexcept
...
+  requires requires {
+    { v.theta() } -> std::floating_point;
+  }
+{
+  return -std::log(std::tan(0.5 * v.theta()));
```
junggjo9:
```suggestion
  return  Acts::AngleHelpers::etaFromTheta(v.theta());
```

andiwand (approved):

PR #5599, opened 2026-06-17
refactor!: Read Bethe-Heitler-Approx from json

benjaminhuth:
So far, ACTS had a reader for the custom ATLAS/Athena native parameterization format. However, this format is non-standard and untested. This replaces this by introducing a json-based reading, and also adds a file to the repository that encodes the default parameterization used in ATLAS.


Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
```diff
@@ -136,24 +134,31 @@ class AtlasBetheHeitlerApprox : public BetheHeitlerApprox {
...
-  /// the parameterization for low x/x0
-  /// @param high_parameters_path Path to the foo.par file that stores
-  /// the parameterization for high x/x0
-  /// @param lowLimit the upper limit for the low x/x0-data
-  /// @param highLimit the upper limit for the high x/x0-data
-  /// @param clampToRange forwarded to constructor
-  /// @param noChangeLimit forwarded to constructor
-  /// @param singleGaussianLimit forwarded to constructor
-  /// @return AtlasBetheHeitlerApprox instance loaded from parameter files
-  static AtlasBetheHeitlerApprox loadFromFiles(
```
andiwand, line 152: technically breaking?

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
```diff
@@ -169,24 +174,25 @@ class AtlasBetheHeitlerApprox : public BetheHeitlerApprox {
...
 
   /// Returns the number of components the returned mixture will have
   /// @return Number of components in the mixture
   std::size_t maxComponents() const override {
-    return std::max(m_lowData.size(), m_highData.size());
+    std::size_t max = 0;
+    for (const auto &range : m_ranges) {
+      max = std::max(max, range.data.size());
+    }
+    return max;
```
andiwand: alternatively
```suggestion
    return std::ranges::max(m_ranges, {}, [](const auto &range) { return range.data.size(); });
```

Core/src/TrackFitting/BetheHeitlerApprox.cpp
```diff
@@ -8,58 +8,43 @@
...
+      m_singleGaussianLimit(singleGaussianLimit) {
+  if (m_ranges.empty()) {
+    throw std::invalid_argument("At least one range is required");
+  }
 
-    Data data;
+  // Sort ranges by minimum value
+  std::sort(m_ranges.begin(), m_ranges.end(), [](const auto &a, const auto &b) {
+    return a.range.min() < b.range.min();
+  });
```
andiwand: we could also push this to the users responsibility and only check if it is sorted already

alternatively
```suggestion
  // Sort ranges by minimum value
  std::ranges::sort(m_ranges, {}, [](const auto &r) { return r.range.min(); });
```

Core/src/TrackFitting/BetheHeitlerApprox.cpp
```diff
@@ -8,58 +8,43 @@
...
+  }
 }
 
-std::span<AtlasBetheHeitlerApprox::Component> AtlasBetheHeitlerApprox::mixture(
+std::span<PolynomialBetheHeitlerApprox::Component>
+PolynomialBetheHeitlerApprox::mixture(
     double xOverX0, const std::span<Component> mixture) const {
   if (m_clampToRange) {
-    xOverX0 = std::clamp(xOverX0, 0.0, m_highLimit);
+    xOverX0 = std::clamp(xOverX0, 0.0, m_ranges.back().range.max());
```
andiwand: since `m_ranges.back().range.max()` comes up quite a bit we could have a short hand either public or private on the class

Plugins/Json/include/ActsPlugins/Json/BetheHeitlerApproxJsonConverter.hpp
```diff
@@ -0,0 +1,50 @@
...
+               PolynomialBetheHeitlerApprox::RangeData& data);
+
+/// Load a Bethe-Heitler approximation from a JSON file.
+///
+/// @param filepath Path to the JSON file
+/// @param clampToRange Whether to clamp x/x0 values to the valid range
+/// @param noChangeLimit Limit below which no change is applied
+/// @param singleGaussianLimit Limit below which a single Gaussian is used
+/// @return Shared pointer to the loaded PolynomialBetheHeitlerApprox
+std::shared_ptr<const PolynomialBetheHeitlerApprox>
```
andiwand: - this could just return a value right?
- I wonder if we could also use `from_json` here. one problem is the additional configuration

Plugins/Json/src/BetheHeitlerApproxJsonConverter.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "ActsPlugins/Json/BetheHeitlerApproxJsonConverter.hpp"
+
+#include "Acts/Utilities/RangeXD.hpp"
+
+#include <fstream>
+#include <stdexcept>
+
+namespace Acts {
+
+void to_json(nlohmann::json& j,
```
andiwand: alternatively you can bind them to the namespace to avoid drifting declaration and definition which would only be caught during link time
```suggestion
void Acts::to_json(nlohmann::json& j,
```

Python/Examples/src/TrackFitting.cpp
```diff
@@ -100,19 +101,18 @@ void addTrackFitting(py::module& mex) {
...
+        mex, "PolynomialBetheHeitlerApprox")
+        .def(py::init<std::vector<PolynomialBetheHeitlerApprox::RangeData>,
+                      bool, double, double>(),
+             "ranges"_a, "clampToRange"_a = false, "noChangeLimit"_a = 0.0001,
+             "singleGaussianLimit"_a = 0.002);
+
+    mex.def("loadBetheHeitlerApproxFromJson",
+            &Acts::loadBetheHeitlerApproxFromJson, "filepath"_a,
+            "clamp_to_range"_a = false, "no_change_limit"_a = 0.0001,
+            "single_gaussian_limit"_a = 0.002);
```
andiwand: these defaults appear in multiple places. should we collect them rather in a single place and use `None` defaults on the python side?

andiwand (commented):
nice! left a few suggestions

since we are in a breaking cycle we could just absorb the API change and mark this as breaking

Core/include/Acts/TrackFitting/BetheHeitlerApprox.hpp
benjaminhuth: true. lets add the !

Plugins/Json/include/ActsPlugins/Json/BetheHeitlerApproxJsonConverter.hpp
benjaminhuth: yeah, I a value would be better. Otherwise, I would leave it a function as entry point due to the config things...

Python/Examples/src/TrackFitting.cpp
benjaminhuth: Actually I think there should be no defaults at all for these values...

Plugins/Json/include/ActsPlugins/Json/BetheHeitlerApproxJsonConverter.hpp
benjaminhuth: Hmm actually, all our APIs take `std::shared_ptr`, so lets stick with that.

paulgessinger:
Compile errors seem real @benjaminhuth

benjaminhuth:
Just to document the impact on physmon:

## Fit
image
image
image



## Refit

image
image

Plugins/Json/include/ActsPlugins/Json/BetheHeitlerApproxJsonConverter.hpp
```diff
@@ -38,12 +49,10 @@ void from_json(const nlohmann::json& j,
 /// @param clampToRange Whether to clamp x/x0 values to the valid range
 /// @param noChangeLimit Limit below which no change is applied
 /// @param singleGaussianLimit Limit below which a single Gaussian is used
-/// @return Shared pointer to the loaded PolynomialBetheHeitlerApprox
-std::shared_ptr<const PolynomialBetheHeitlerApprox>
-loadBetheHeitlerApproxFromJson(const std::string& filepath,
-                               bool clampToRange = false,
-                               double noChangeLimit = 0.0001,
-                               double singleGaussianLimit = 0.002);
+/// @return The loaded PolynomialBetheHeitlerApprox
```
AJPfleger, line 52: Why the type-change?

AJPfleger: After discussion: I commented to the changes after the last review. I think it's fine to restore to the "original" version.

Now we have `std::make_shared` in the python-binding. In theory we could put it here but I guess it's fine as it is.

AJPfleger (approved):

PR #5603, opened 2026-06-18
chore: Enhance `enumerate`

andiwand:

- Fix dangling references with proxy/temporary elements: `operator*` now returns `std::tuple<index_type, decltype(*iter)>` instead of `std::tie(...)`. Iterators whose operator* yields a prvalue (e.g. `vector<bool>`, computed/zip-style ranges) are now stored by value; plain `T&` elements still bind by reference, so in-place mutation through the structured binding is unchanged.
- More robust index type: index_type is deduced from `std::size(container)` and falls back to std::size_t when the range has no `size()` (via the new `detail::enumerate_index_type` trait). This also makes C-style arrays work, where the old ::size_type lookup failed.
- Constrain with a concept instead of SFINAE: added an enumerable concept (checks `std::begin`/`std::end`) replacing the trailing SFINAE template parameter; using `std::begin`/`std::end` also admits C-style arrays.
- Make the iterator an explicit `LegacyInputIterator`: added `iterator_traits` typedefs (`iterator_category`, `value_type`, `reference`, `pointer`, `difference_type`), changed `operator++` to return `iterator&`, added post-increment, and switched to `operator==` (C++20 synthesizes `!=`). It is intentionally not declared forward, since the proxy tuple reference can't satisfy the C++20 iterator concepts in C++20 (no tuple `common_reference` until C++23).
- Moved the iterator/wrapper into detail as templates (previously function-local structs) to avoid `-Wunused-local-typedef` on the new trait aliases under `-Werror`.
- Doc fix: corrected the usage example from `for (auto [...] = enumerate(...))` to `for (auto [...] : enumerate(...))`.

--- END COMMIT MESSAGE ---


Core/include/Acts/Utilities/Enumerate.hpp
```diff
@@ -8,48 +8,107 @@
...
+template <typename T>
+struct enumerate_index_type<
+    T, std::void_t<decltype(std::size(std::declval<T>()))>> {
+  using type = decltype(std::size(std::declval<T>()));
+};
+
+/// A type is enumerable if it exposes `std::begin` and `std::end` (covers
+/// stl-like containers as well as C-style arrays).
+template <typename T>
+concept enumerable = requires(T &&t) {
```
benjaminhuth: Hmm isn't that exactly what the concept of a C++20 `range` is?

Core/include/Acts/Utilities/Enumerate.hpp
benjaminhuth, line 1: Hmm in general, this should follow our `CamelCase` guide, right?

Core/include/Acts/Utilities/Enumerate.hpp
```diff
@@ -8,48 +8,107 @@
...
+    ++*this;
+    return tmp;
+  }
+
+  reference operator*() const { return reference{i, *iter}; }
+};
+
+/// Range wrapper that yields `enumerate_iterator`s over a (possibly owned)
+/// container.
+template <typename container_type, typename index_type, typename iterator_t>
```
benjaminhuth: Hmm this could also take the concept, right?

benjaminhuth (commented):
Looks good, some comments though

Core/include/Acts/Utilities/Enumerate.hpp
andiwand: ah yes - I was unsure about the `.begin()` vs `std::begin()` but of course `ranges` also captures this

benjaminhuth (approved):
LGTM!

PR #5605, opened 2026-06-19
feat: record Geant4 production process of simulated particles

altsybee:
# Geant4: record the production process of simulated particles

## Summary

The standalone Geant4 simulation (`Examples/Algorithms/Geant4`) never sets the `ActsFatras::GenerationProcess` on the particles it creates, so every simulated particle was written out with `process = eUndefined`. (Fatras sets it via its physics modules, but the Geant4 backend did not.) 
This made it impossible to tell, downstream, whether a secondary came from a decay, a conversion, bremsstrahlung, a nuclear interaction, etc.

This PR maps the Geant4 creator process of each track to a `GenerationProcess` and stores it on the particle, so it is propagated to the output (e.g. the `process` branch of `RootParticleWriter`).

## Changes

- **`Examples/Algorithms/Geant4/src/ParticleTrackingAction.cpp`**: in
  `convert()`, set the process from `track.GetCreatorProcess()` via a new helper
  `g4CreatorToGenerationProcess()`. Mapping:
  - no creator process (primary) → `eUndefined`
  - `fDecay` → `eDecay`
  - `fHadronic` / `fPhotolepton_hadron` → `eNuclearInteraction`
  - EM sub-types `fBremsstrahlung` → `eBremsstrahlung`,
    `fGammaConversion` → `ePhotonConversion`, `fIonisation` → `eIonisation`
  - anything else (Compton, photoelectric, annihilation, …) → `eOther`
- **`Fatras/include/ActsFatras/EventData/GenerationProcess.hpp`**: add
  `eIonisation = 5` and `eOther = 6` to carry the finer Geant4 detail (with a
  note that 4/5/6 are Geant4-fed; Fatras itself only sets 1/2/3).
- **`Fatras/src/EventData/GenerationProcess.cpp`**: add the matching
  `operator<<` cases.

## Notes / compatibility

- Set on the initial state in `PreUserTrackingAction`, which is what  `SimParticle::process()` / `RootParticleWriter` write out.
- Adding enum values is backward compatible: the only `switch` over  `GenerationProcess` (`GenerationProcess.cpp`) has a `default`, and all other  consumers cast uint↔enum or stream it. Existing files written with codes 0–4  are unaffected.
- Fatras-based simulation is unchanged.

### Example usage: origin of particles which produce fake measurements on tracks 
(Pythia pp; categories can overlap, so they need not sum to 100%):
 -   primary: 316  (4.21558 %)
 -   decay: 1240  (16.5422 %)
  -  material: 5992  (79.936 %)
      -  photon-conversion: 106  (1.41409 %)
      -  bremsstrahlung: 0  (0 %)
      -  nuclear-interaction: 528  (7.04376 %)
      -  delta-ray: 5337  (71.198 %)
      -  other: 47  (0.627001 %)
 -   unknown: 11  (0.146745 %)


benjaminhuth (approved):
Seems like a good addition to me, since we apparently did not populate the `process` field at all for out Geant4 integration.

PR #5610, opened 2026-06-22
feat: Add optional time-of-flight cuts to Seeding2 doublet finding and seed filtering

jburzy:
Use the per-space-point time to reject combinations that are inconsistent with a particle travelling from the interaction point at the speed of light.

- DoubletSeedFinder: new Config fields deltaTMax and tofInverseSpeed. A doublet is rejected when |(t_other - t_middle) - (L_other - L_middle) * tofInverseSpeed| exceeds deltaTMax, with L = sqrt(r^2 + z^2).
- BroadTripletSeedFilter: the same deltaTMax / tofInverseSpeed cut applied to a triplet's bottom and top space points.

deltaTMax defaults to infinity (no cut) and the cut is only evaluated when the space-point container provides a time column, so seeding without time behaves exactly as before.

--- END COMMIT MESSAGE ---

This is meant for 4D trackers to improve CPU performance by rejecting spurious doublet combinations early in the chain.

### DIFF
--- a/Core/include/Acts/Seeding2/BroadTripletSeedFilter.hpp
+++ b/Core/include/Acts/Seeding2/BroadTripletSeedFilter.hpp
@@ -65,6 +65,16 @@ class BroadTripletSeedFilter final : public ITripletSeedFilter {
  public:
   /// @brief Structure that holds configuration parameters for the seed filter algorithm
   struct Config {
+    /// Maximum allowed deviation of the time difference between a triplet's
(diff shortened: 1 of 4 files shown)


jburzy:
Looks like maybe this is a duplicate of https://github.com/acts-project/acts/pull/5524... will leave open for now but feel free to close if the other implementation is preferred

jburzy:
Closing this in favor of the template implementation downstream in k4ActsTracking

PR #5611, opened 2026-06-22
chore: DevContainer / VSCode improvements

paulgessinger:
This pull request updates the development container setup to improve CMake integration, environment configuration, and pre-commit hook management. The main changes include switching from `pre-commit` to `prek` for git hooks, enhancing CMake support in VS Code, and improving environment consistency and caching for builds.

**Dev environment improvements:**

* Switched from installing `pre-commit` to `prek` in the Dockerfile and updated documentation to use `prek install` for enabling git hooks. (`.devcontainer/Dockerfile`, `.devcontainer/setup.sh`) [[1]](diffhunk://#diff-13bd9d7a30bf46656bc81f1ad5b908a627f9247be3f7d76df862b0578b534fc6L5-R5) [[2]](diffhunk://#diff-aaf9e7764a12876cbe70d492c719e21e8590390a38d0f037193a765df858fdbfL27-R27)
* Removed the automatic installation of pre-commit hooks from the setup script to avoid unintended side effects. (`.devcontainer/setup.sh`)

**CMake and VS Code integration:**

* Added the `ms-vscode.cmake-tools` extension and updated VS Code settings to set the default CMake preset to `devcontainer`, configure test environment variables, and improve test output handling. (`.devcontainer/devcontainer.json`)
* Added a new `devcontainer` CMake preset for consistent builds inside the container. (`CMakePresets.json`)

**Environment and caching enhancements:**

* Configured additional environment variables (`LC_ALL`, `LANG`, `CCACHE_DIR`) and mounted a named volume for ccache to speed up rebuilds and ensure locale consistency. (`.devcontainer/devcontainer.json`)


.devcontainer/devcontainer.json
```diff
@@ -7,19 +7,30 @@
         "vscode": {
             "extensions": [
                 "xaver.clang-format",
-                "twxs.cmake",
-                "llvm-vs-code-extensions.vscode-clangd"
+                "llvm-vs-code-extensions.vscode-clangd",
+                "ms-vscode.cmake-tools"
```
krasznaa, line 11: I know I'm a Microsoft shill at this point, but why don't we just use `ms-vscode.cpptools-extension-pack` and be done with it? 🤔

Those packages by themselves (I believe) even provide clang-format. Without having to rely on a "3rd party" plugin.

.devcontainer/devcontainer.json
```diff
@@ -7,19 +7,30 @@
...
             ],
             "settings": {
                 "clangd.path": "/usr/bin/clangd",
                 "clangd.arguments": [
                     "--compile-commands-dir=${containerWorkspaceFolder}/build_devcontainer"
-                ]
+                ],
```
krasznaa, line 17: One of the upsides of buying into the Microsoft ecosystem is that we don't need to fine-tune clangd like this.

Though we do get Intellisense and its own issues. (Which for me personally have been more or less acceptable recently.)

.devcontainer/devcontainer.json
```diff
@@ -7,19 +7,30 @@
...
     },
     "containerEnv": {
-        "CXXFLAGS": "-Wno-psabi"
+        "CXXFLAGS": "-Wno-psabi",
+        "CCACHE_DIR": "/ccache",
+        "LC_ALL": "C.UTF-8",
+        "LANG": "C.UTF-8"
     },
+    "mounts": [
+        "source=acts-ccache,target=/ccache,type=volume"
```
krasznaa, line 33: I don't actually understand what this does. 😦 What local directory do you mount into the container as `/ccache`? And why is this not just handled through the build directory?

These are genuine questions. I don't use ccache practically ever. But in such a setup I imagined that we'd want to have the container use its own cache area, no?

.devcontainer/devcontainer.json
```diff
@@ -7,19 +7,30 @@
...
+                "cmake.ctest.testSuiteDelimiter": "/",
+                "cmake.testEnvironment": {
+                    "BOOST_TEST_COLOR_OUTPUT": "0"
+                }
             }
         }
     },
     "containerEnv": {
-        "CXXFLAGS": "-Wno-psabi"
+        "CXXFLAGS": "-Wno-psabi",
```
krasznaa, line 27: Where does this come from? 😕 It has a weird smell that we'd need to turn off a specific warning like this.

.devcontainer/Dockerfile
```diff
@@ -2,4 +2,6 @@ FROM ghcr.io/acts-project/spack-container:23.2.0_linux-ubuntu24.04_gcc-13.3.0_cx
 
 RUN apt-get update && apt-get install -y clangd vim
 
-RUN uv tool install --no-cache pre-commit
+RUN uv tool install -U --no-cache --no-build --force uv
+
+RUN uv tool install --no-build --no-cache prek
```
krasznaa: SonarCloud may have a point here. 🤔

krasznaa (commented):
I'm all for perking up the DevContainer config a little. Since making a proper development environment for Acts has been a bit challenging to me as well in the past.

I don't see anything deal-breaking here. I'm mainly wondering if we could simplify the configuration a little further. 🤔

.devcontainer/devcontainer.json
paulgessinger: Possible, I've found `clangd` to be more reliable than MS' C++ intelligence. With the compilation DB hooked up, clangd interops well with cmake-tools.

.devcontainer/devcontainer.json
paulgessinger: It's a named docker volume. Docker creates this on the fly when the container launches, and it persists across container launches. `/ccache` is **inside** the container.

.devcontainer/devcontainer.json
paulgessinger: Yeah this is annoying. The origin is a warning on aarch64 GCC that just tells you that the compiler/stdlib changed their ABI for `std::pair` (IIRC) at some point. From all of my research, this is not associated with any code pattern or code defect in particular, it's just that if you use `std::pair` as a function argument in GCC on aarch64, it will tell you that the ABI changed back in GCC10 or so. As far as I've been able to work out, this is not actionable from our side.

paulgessinger: @krasznaa makes sense?

.devcontainer/Dockerfile
paulgessinger: Added.

paulgessinger:
@krasznaa can you have another lok now?

.devcontainer/devcontainer.json
krasznaa: Yeah, that ARM warning has been the bane of me in some cases as well... 😦 I think you're right, this is as good a "solution" for it as we can get.

krasznaa (approved):
Sorry for the silence. Yeah, I don't think we can do much better than this. 😉

PR #5613, opened 2026-06-23
fix: TrackingGeometry from JSON - Fix deserialization from straws

junggjo9:
- Straw surfaces are not regular surfaces -> current desrialization throw an exception
- Change interface to parse the string to the actual JSON instead of a desrialized JSON -> ensure that the JSON does not remain in memory after construction

--- END COMMIT MESSAGE ---
Tagging @paulgessinger, @benjaminhuth


Examples/Scripts/Python/material_merge_rz_view.py
```diff
@@ -97,7 +97,7 @@ def load_tracking_geometry(
     from acts.json import TrackingGeometryJsonConverter
 
     converter = TrackingGeometryJsonConverter()
-    return converter.fromJson(gctx, Path(json_path).read_text())
+    return converter.fromJson(gctx, str(json_path.absolute()))
```
paulgessinger: Can we keep this as `Path` (without the `read_text()`)?

Plugins/Json/include/ActsPlugins/Json/TrackingGeometryJsonConverter.hpp
```diff
@@ -167,12 +167,12 @@ class TrackingGeometryJsonConverter {
...
   ///
   /// @param gctx geometry context
-  /// @param encoded serialized tracking geometry
+  /// @param jsonPath The path to the JSON file on disc
   /// @param options options for the conversion
   ///
   /// @return pointer to deserialized geometry
   std::shared_ptr<TrackingGeometry> fromJson(
-      const GeometryContext& gctx, const nlohmann::json& encoded,
+      const GeometryContext& gctx, const std::string& jsonPath,
```
paulgessinger:
```suggestion
      const GeometryContext& gctx, const std::filesystem::path& jsonPath,
```

Plugins/Json/src/TrackingGeometryJsonConverter.cpp
```diff
@@ -1204,8 +1206,21 @@ Acts::TrackingGeometryJsonConverter::trackingVolumeFromJson(
 
 std::shared_ptr<Acts::TrackingGeometry>
 Acts::TrackingGeometryJsonConverter::fromJson(const GeometryContext& gctx,
-                                              const nlohmann::json& encoded,
+                                              const std::string& path,
```
paulgessinger:
```suggestion
                                              const std::filesystem::path& path,
```

Python/Plugins/src/Json.cpp
```diff
@@ -113,8 +115,8 @@ PYBIND11_MODULE(ActsPluginsPythonBindingsJson, json) {
         .def(
             "fromJson",
             [](const TrackingGeometryJsonConverter& self,
-               const GeometryContext& gctx, const std::string& encoded) {
-              return self.fromJson(gctx, nlohmann::json::parse(encoded));
+               const GeometryContext& gctx, const std::string& filePath) {
+              return self.fromJson(gctx, filePath);
```
paulgessinger: You can then just bind the `fromJson` directly, without a wrapper lambda.

If this doesn't work, there might be some pybind11 adapter header that needs to be included.

Examples/Scripts/Python/material_merge_rz_view.py
junggjo9:
```suggestion
    return converter.fromJson(gctx, json_path.absolute())
```

Python/Examples/tests/test_geometry.py
```diff
@@ -213,7 +213,7 @@ def test_odd_gen3_json_roundtrip(tmp_path):
     assert json_path.stat().st_size > 0
 
     converter = TrackingGeometryJsonConverter()
-    rebuilt_geometry = converter.fromJson(gctx, json_path.read_text())
+    rebuilt_geometry = converter.fromJson(gctx, str(json_path.absolute()))
```
junggjo9:
```suggestion
    rebuilt_geometry = converter.fromJson(gctx, json_path.absolute())
```

Python/Plugins/src/Json.cpp
junggjo9: Done

Plugins/Json/src/TrackingGeometryJsonConverter.cpp
```diff
@@ -409,25 +411,15 @@ std::unique_ptr<Acts::INavigationPolicy> decodeCylinderNavigationPolicy(
...
-
 std::unique_ptr<Acts::GridPortalLink> makeGridPortalLink(
     const std::shared_ptr<Acts::RegularSurface>& surface,
     Acts::AxisDirection direction, const Acts::IAxis& axis0,
     const Acts::IAxis* axis1) {
   std::unique_ptr<Acts::GridPortalLink> grid;
-
+  if (!surface) {
+    throw std::invalid_argument(
+        "makeGridPortalLink() - Non regular srurface was passed");
```
benjaminhuth:
```suggestion
        "makeGridPortalLink() - Non regular surface was passed");
```

benjaminhuth (commented):
One problem I currently see that we never run a geometry conversion with straws in any test in ACTS, and given the nature of straw surfaces, this could break easily, so adding a unittest for this would be nice maybe

junggjo9:
good point, I'll add one. Nevertheless, the excellent python binding crashes in the CI but not on my machine :(

benjaminhuth (approved):
LGTM.

PR #5614, opened 2026-06-24
refactor: C++ implementation of ROOT file hasher

paulgessinger:
This is an implementation of the ROOT file hashing in C++. The hashes are not identical, but the guarantees should be the same. It should be strictly faster than the Ptyhon implementation.

--- END COMMIT MESSAGE ---


Examples/Io/Root/src/RootFileHasher.cpp
```diff
@@ -0,0 +1,438 @@
...
+#include <boost/hash2/sha2.hpp>
+#else
+#include <TMD5.h>
+#endif
+
+namespace ActsExamples {
+
+namespace {
+
+#if defined(ACTS_ROOT_FILE_HASHER_USE_BOOST_HASH2)
```
benjaminhuth: Nitpick: I think one can make the `#if` `#else` `#endif` smaller by pulling out some stuff. but its a bit matter of taste

Examples/Io/Root/src/RootFileHasher.cpp
```diff
@@ -0,0 +1,438 @@
...
+         ROOT::Internal::TTreeReaderValueBase::kSetupMatch;
+}
+
+/// Reader for a fundamental scalar branch.
+template <typename T>
+struct ScalarReader final : IBranchReader {
+  TTreeReaderValue<T> value;
+  ScalarReader(TTreeReader& reader, const char* name) : value(reader, name) {}
+  bool valid() const override { return setupOk(value); }
+  void append(std::string& out) override {
```
benjaminhuth: Is there a reason why we use `std::string` here? wouldn't `std::vector<std::byte>` fit better, especially we feed `std::span<const std::byte>` to the hash later?

Examples/Io/Root/src/RootFileHasher.cpp
```diff
@@ -0,0 +1,438 @@
...
+  if (type == "UChar_t" || type == "unsigned char") {
+    return makeReader<unsigned char>(kind, reader, name);
+  }
+  if (type == "Bool_t" || type == "bool") {
+    return makeReader<bool>(kind, reader, name);
+  }
+  return nullptr;
+}
+
+std::string trim(const std::string& s) {
```
benjaminhuth: Can we use `boost::trim` here?

Examples/Io/Root/src/RootFileHasher.cpp
```diff
@@ -0,0 +1,438 @@
...
+  if (begin == std::string::npos) {
+    return "";
+  }
+  auto end = s.find_last_not_of(" \t");
+  return s.substr(begin, end - begin + 1);
+}
+
+/// Return the first top-level template argument of a type name, e.g.
+/// "vector<float>" -> "float" and "vector<vector<int> >" -> "vector<int>".
+std::string firstTemplateArg(const std::string& s) {
```
benjaminhuth: I wonder if we should add a unittest for that... maybe overkill

Examples/Io/Root/src/RootFileHasher.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+struct Reader {
+  std::string name;
+  std::unique_ptr<IBranchReader> reader;
+  std::unique_ptr<Hasher> hasher;  // only used in non-order-invariant mode
+};
+
+/// Compute the digest of a single tree.
+Digest hashTree(TTree& tree, bool orderInvariant) {
+  std::vector<BranchInfo> infos = describeBranches(tree);
```
benjaminhuth, line 303: Is there a reason why info collection and reader creation is split in two loops? Wouldn't it be more natural to have a single loop?

benjaminhuth (commented):
Looks good, a few comments. Do we know how much faster it is?

paulgessinger:
@benjaminhuth i've found the C++ version runs in "constant" time (clearly it doesn't, but the file size in my benchmarks it didn't move). On the smallest files the perf is the same, for larger files I get speedups of 2x+

Examples/Io/Root/src/RootFileHasher.cpp
paulgessinger: The branches are sorted after being loaded, which is needed to make it fully reproducible. We could also do a single loop and then sort before accumulating the hash, but that seems fairly similar to this.

paulgessinger:
@benjaminhuth here are the benchmarks:

```
===================================================================
### tracksummary_ckf.root  (372K)
===================================================================
Benchmark 1: python-uproot
  Time (mean ± σ):     354.8 ms ±   8.0 ms    [User: 299.6 ms, System: 51.0 ms]
  Range (min … max):   343.1 ms … 368.1 ms    10 runs

Benchmark 2: cpp-sha256
  Time (mean ± σ):     343.0 ms ±   7.6 ms    [User: 277.8 ms, System: 58.7 ms]
  Range (min … max):   332.1 ms … 356.4 ms    10 runs

Summary
  cpp-sha256 ran
    1.03 ± 0.03 times faster than python-uproot

===================================================================
### hits.root  (504K)
===================================================================
Benchmark 1: python-uproot
  Time (mean ± σ):     849.9 ms ±  15.2 ms    [User: 789.3 ms, System: 55.1 ms]
  Range (min … max):   811.7 ms … 867.7 ms    10 runs

Benchmark 2: cpp-sha256
  Time (mean ± σ):     336.7 ms ±   8.5 ms    [User: 275.1 ms, System: 55.8 ms]
  Range (min … max):   324.3 ms … 351.9 ms    10 runs

Summary
  cpp-sha256 ran
    2.52 ± 0.08 times faster than python-uproot

===================================================================
### measurements.root  (1.1M)
===================================================================
Benchmark 1: python-uproot
  Time (mean ± σ):      2.169 s ±  0.051 s    [User: 2.089 s, System: 0.072 s]
  Range (min … max):    2.066 s …  2.247 s    10 runs

Benchmark 2: cpp-sha256
  Time (mean ± σ):     348.9 ms ±  11.1 ms    [User: 284.9 ms, System: 57.9 ms]
  Range (min … max):   336.2 ms … 365.3 ms    10 runs

Summary
  cpp-sha256 ran
    6.22 ± 0.25 times faster than python-uproot
```

benjaminhuth:
nice. 2s to 300ms is some good speedup.

Examples/Io/Root/include/ActsExamples/Io/Root/detail/RootFileHasherHelpers.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <string>
+
+#include <boost/algorithm/string/trim.hpp>
+
+namespace ActsExamples::detail {
+
+/// Return the first top-level template argument of a type name, e.g.
+/// "vector<float>" -> "float" and "vector<vector<int> >" -> "vector<int>".
+/// Returns an empty string if the name has no template argument.
+inline std::string firstTemplateArg(const std::string& s) {
```
andiwand, line 21: nitpick:
- could be part of `RootFileHasher.hpp`
- implementation could go to source

Examples/Io/Root/src/RootFileHasher.cpp
```diff
@@ -0,0 +1,404 @@
...
+#else
+constexpr std::size_t kDigestSize = 16;
+#endif
+
+using Digest = std::array<unsigned char, kDigestSize>;
+
+/// Incremental hasher backed by Boost.Hash2 SHA-256 (preferred) or ROOT's
+/// TMD5 (fallback for Boost < 1.86). Only the backend member and the two
+/// method bodies differ between the two configurations.
+class Hasher {
```
andiwand: I wonder if it would be cleaner to template on the underlying hasher and then only use the preprocessor for a typedef or something like this

andiwand (commented):
pretty nice!

Examples/Io/Root/include/ActsExamples/Io/Root/detail/RootFileHasherHelpers.hpp
paulgessinger: This is in the header so it can be covered in the unit test, we can drop the unit test however, if we think we don't need the coverage.

Even if we keep the test, we can move this to the `RootFileHasher.hpp` of course.

Examples/Io/Root/src/RootFileHasher.cpp
paulgessinger: The hashers signatures aren't identical:

```cpp
    m_hasher.update(reinterpret_cast<const unsigned char*>(data.data()), data.size());
    m_hasher.Update(reinterpret_cast<const UChar_t*>(data.data()), data.size());
```

I'll try to reduce the preprocessor surface.

andiwand: right I guess we could then either use `if constexpr` or have a minimal wrapper to align the interface into a common one. at the end of the day it does the same but avoid using the preprocessor multiple times. up to you

Examples/Io/Root/include/ActsExamples/Io/Root/detail/RootFileHasherHelpers.hpp
andiwand: right I just thought the additional header is not strictly necessary and the `namespace detail {}` could be moved over to `RootFileHasher.hpp`. up to you

andiwand (approved):

PR #5619, opened 2026-06-29
perf: speed up Parquet reading via LRU shard cache

benjaminhuth:
Replace the per-event Arrow dataset scan with an LRU shard cache. The old path decompressed the full shard row group (~750 MB for ColliderML) on every readEvent() call. The new path decompresses each shard once on first access and serves subsequent events with an in-memory Table::Slice, reducing per-event I/O cost by ~100x for sequential processing.


Core/include/Acts/Utilities/detail/LruCache.hpp
paulgessinger, line 1: Pedantic comment of the day: `LruCache`!

paulgessinger (dismissed):
lgtm. can you check the sonar issues though?

benjaminhuth:
fixed the issues, can you reapprove?

paulgessinger (approved):

PR #5620, opened 2026-06-29
feat: Add a propagation validation function

niermann999:
Adds a new function that compares a full propagation (with covariance transport) against a reference truth trace. The result is a detailed statistic about additionally encountered and missed surfaces for the track population, as well as SVG for tracks with deviations and a log of the navigation steps of failed tracks.

This function is used in the traccc tests to compare a bare propagation (without KF) against several truth scenarios (fast sim with and without material effects) to determine if roadbuilding without KF updates are feasible.

--- END COMMIT MESSAGE ---

### DIFF
--- a/CI/check_unused_files.py
+++ b/CI/check_unused_files.py
@@ -78,6 +78,7 @@ EXCLUDE_FILES = (
     "Detray/codegen/detray-sympy/tests/test_matrices.py",
     # Used in traccc
     "Detray/tests/include/detray/test/utils/perigee_stopper.hpp",
+    "Detray/tests/include/detray/test/validation/propagation_validation.hpp",
     # Build-time metadata generation
     "Detray/python/detray/detectors/impl/definitions.py",
     "Detray/python/detray/detectors/impl/type_helpers.py",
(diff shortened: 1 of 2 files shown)


stephenswat (approved):

CI/check_unused_files.py
```diff
@@ -78,6 +78,7 @@
     "Detray/codegen/detray-sympy/tests/test_matrices.py",
     # Used in traccc
     "Detray/tests/include/detray/test/utils/perigee_stopper.hpp",
+    "Detray/tests/include/detray/test/validation/propagation_validation.hpp",
```
andiwand, line 81: rather than ignoring that the file is not used anywhere it would be good to compile and smoke test them somewhere. right now they could break silently only for someone downstream to find out

niermann999: I agree, I am working on it as part of this [PR](https://github.com/acts-project/acts/pull/5625/changes#diff-903bb4ae647939ff76a2fe4cdb033cdfffed2f1ff5c9893c35dbf197efe98314) , but this needed to go in asap to make the traccc integration

PR #5623, opened 2026-06-30
perf!: Replace Barcode::asVector with asArray

benjaminhuth:
This causes heap allocations in each comparison, and is very painful in maps with barcodes as keys.

--- END COMMIT MESSAGE ---


paulgessinger (commented):
Ok, but wouldn't it be better to key on the hash? That's what e.g. `unordered_map` will already do? Where are you hitting issues with this?

paulgessinger (approved):

PR #5626, opened 2026-06-30
feat: Add TrackingGeometry::findVolumeByName

paulgessinger:


Core/src/Geometry/TrackingGeometry.cpp
```diff
@@ -279,6 +279,17 @@ const TrackingVolume* TrackingGeometry::findVolume(
...
 
+const TrackingVolume* TrackingGeometry::findVolumeByName(
+    std::string_view name) const {
+  const TrackingVolume* found = nullptr;
+  apply([&](const TrackingVolume& volume) {
+    if (found == nullptr && volume.volumeName() == name) {
+      found = &volume;
+    }
+  });
+  return found;
```
junggjo9, line 290: Probably, I am overseeing here something but if you've a volume structure lik
```
- House
   -- Room
     -- Table 
       -- Plate
         -- Cake
```
and you want to ask for Table. The traversing continues to the rock bottom of the volume hierachy and doesn't bail out earlier right?

paulgessinger: I don't think these a mechanism in the visit system yet to early abort. I imagine this function wouldn't be called on a hot path, so i would not prematurely optimize.

junggjo9 (approved):

PR #5628, opened 2026-07-01
feat: Include and set up python bindings in the build options

VladimirDrobny:
This commit adds the DETRAY_BUILD_PYTHON_BINDINGS cmake option.
If enabled, it also sets up pybind11, the same way as it is done in acts.
As with other Detray dependencies, this commit provides the option to use the system pybind11 or to fetch and build it from an external source.

--- END COMMIT MESSAGE ---


Detray/extern/pybind11/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+    set(DETRAY_PYBIND11_SOURCE_FULL "${DETRAY_PYBIND11_SOURCE}")
+endif()
+mark_as_advanced(DETRAY_PYBIND11_SOURCE_FULL)
+
+# Declare where to get pybind from.
+FetchContent_Declare(pybind11 ${DETRAY_PYBIND11_SOURCE_FULL})
+
+# Set up its build.
+set(PYBIND11_TEST OFF)
+set(PYBIND11_INSTALL OFF)
```
niermann999: Should these be cache variables?
```suggestion
set(PYBIND11_TEST FALSE CACHE BOOL)
set(PYBIND11_INSTALL  FALSE CACHE BOOL)
```

niermann999 (approved):
Looks good!

PR #5629, opened 2026-07-01
chore: Update pypi dependencies to 23.2.0

benjaminhuth:
--- END COMMIT MESSAGE ---

@paulgessinger In the long term it would be good to find a mechanism to sync this automatically...


paulgessinger:
@benjaminhuth it's not checked by CI, but I put a script to bump the versions in one go here: https://github.com/acts-project/acts/blob/main/CI/bump_versions.py that can probably be extended to cover this location as well.

paulgessinger (approved):

PR #5630, opened 2026-07-01
feat: add minimal MC truth information for HF tagging

fgrosa:
This PR adds in the examples a minimal MC truth information to trace back tracks produced by HF hadrons/quarks. In particular, it adds in the root outputs:
- A branch containing the information whether the track comes from charm/beauty. If `searchUpToQuark=false` (default) the origin is traced back to HF hadrons, if instead `searchUpToQuark=true` the search in the kine tree goes up to the quarks, including also particles produced in the fragmentation of the heavy quark, but not from the hadron decay.
- A branch containing the information of the HepMC3 particle index to map it to the HepMC3 output that contains the full information

This is needed for ALICE3 performance studies, to assess the reconstruction performance of displaced tracks originating from HF decays, as discussed with @altsybee.


Examples/Io/HepMC3/src/HepMC3InputConverter.cpp
```diff
@@ -105,6 +107,64 @@ std::string printListing(const auto& vertices, const auto& particles) {
...
+
+    if (visited.count(part->id()) != 0u) {
+      continue;
+    }
+    visited.insert(part->id());
+
+    int pdgCode = std::abs(part->pid());
+
+    // --- beauty PDG IDs ---
+    if (static_cast<Acts::HfOrigin>(pdgCode / 100) == Acts::HfOrigin::Bottom ||
```
delitez: For the hadron-level checks from the ancestor PDG (here `pdgCode/100` etc.), would ' hadronType ' be worth reusing instead of a manual check? `Acts::ParticleIdHelper::hadronType` method already decodes the PDG IDs in ATLAS style. You could map, e.g., `BottomMeson/BottomBaryon/BBbarMeson → HfOrigin::Bottom`

Examples/Io/HepMC3/src/HepMC3InputConverter.cpp
```diff
@@ -105,6 +107,64 @@ std::string printListing(const auto& vertices, const auto& particles) {
...
+        static_cast<Acts::HfOrigin>(pdgCode / 1000) == Acts::HfOrigin::Bottom ||
+        (m_cfg.searchUpToHfQuark &&
+         static_cast<Acts::HfOrigin>(pdgCode) == Acts::HfOrigin::Bottom)) {
+      return Acts::HfOrigin::Bottom;
+    }
+
+    // --- charm PDG IDs ---
+    if (static_cast<Acts::HfOrigin>(pdgCode / 100) == Acts::HfOrigin::Charm ||
```
delitez: Same comment from above.

Core/include/Acts/Definitions/PdgParticle.hpp
```diff
@@ -124,6 +124,9 @@ enum class HadronType {
   Unknown = 12
 };
 
+/// Hadron heavy-flavour origin classification
+enum class HfOrigin : std::uint8_t { None = 0, Charm = 4, Bottom = 5 };
```
delitez: Could you elaborate on what `None` means here? No ancestor or prompt-light quark? (also as a doc comment perhaps)

delitez (commented):
Leaving some minor comments, I believe these changes could also be useful for jet labelling. Thanks!

fgrosa:
> Leaving some minor comments, I believe these changes could also be useful for jet labelling. Thanks!

Thanks for your suggestions! I have implemented them in [94562d4](https://github.com/acts-project/acts/pull/5630/commits/94562d417eada8740f3c4f51edf0a7a0e938e24a)
If you think that it would be useful add other information in this PR for the jet labelling, let me know!

Examples/Io/HepMC3/src/HepMC3InputConverter.cpp
fgrosa: I didn't realise about it, done now!

Core/include/Acts/Definitions/PdgParticle.hpp
fgrosa: In this implementation I just tag the heavy-flavour origin, so whatever is not tagged as from charm or bottom, is with "none" tag. I added more information in the comments.

In case you want me to add more tagging in case of `searchUpToQuark=true` (that would be the useful usecase for jet labelling I guess), e.g. from u/d/s/t quarks or gluons, I can add it (but in this case I would propose to change the name of the `enum` and the branch in the output in `Origin` rather than `HfOrigin`).

delitez:
Thank you! The changes look good to me. I think this PR is fine as it is! For the jet labeling, we can discuss opening another PR in the future.

fgrosa:
Hi @delitez, since the PR looks good to you, do you think it could be merged? Thanks!

delitez:
To me it looks good, but I don’t have rights to merge. @paulgessinger could you please take a look as well?

paulgessinger:
@delitez you can still approve and that will be counted.

Core/include/Acts/Definitions/PdgParticle.hpp
```diff
@@ -124,6 +124,17 @@ enum class HadronType {
...
+/// Hadron heavy-flavour origin classification
+/// Bottom: decay product of a bottomed hadron (it can also be B->D->X),
+///         or particle originate in the fragmentation of a bottom quark
+///         depending on the searchUpToHfQuark flag
+/// Charm: decay product of a charmed hadron, or particle originate
+///        in the fragmentation of a charm quark depending
+///        on the searchUpToHfQuark flag
+/// None: no heavy-flavour origin, depending on the criterion regulated
+///       by the searchUpToHfQuark flag
+enum class HfOrigin : std::uint8_t { None = 0, Charm = 4, Bottom = 5 };
```
andiwand: at least to me `HfOrigin` looks a bit odd - how about `HeavyFlavorOrigin`?

Examples/Io/HepMC3/src/HepMC3InputConverter.cpp
```diff
@@ -27,6 +27,8 @@
 #pragma clang diagnostic pop
 #endif
 
+#include <unordered_set>
```
andiwand: nitpick: move the `#include` up next to the other ones

andiwand (commented):
at least for me `HF` does not really ring a bell. `heavy-flavour` does not seem to be too long of a name to just spell it out

Core/include/Acts/Definitions/PdgParticle.hpp
andiwand: two more things
- the values look a bit curious, why 0, 4, 5?
- not sure if this should go to ACTS Core. for now I would rather put it in Fatras

Examples/Io/HepMC3/src/HepMC3InputConverter.cpp
```diff
@@ -105,6 +107,68 @@ std::string printListing(const auto& vertices, const auto& particles) {
 };
 }  // namespace
 
+Acts::HfOrigin HepMC3InputConverter::checkHfOrigin(
```
andiwand:
```suggestion
Acts::HeavyFlavourOrigin HepMC3InputConverter::deriveHeavyFlavourOrigin(
```

Examples/Io/HepMC3/src/HepMC3InputConverter.cpp
```diff
@@ -105,6 +107,68 @@ std::string printListing(const auto& vertices, const auto& particles) {
...
+    const std::shared_ptr<const HepMC3::GenParticle>& particleToCheck) const {
+  std::stack<std::shared_ptr<const HepMC3::GenParticle>> st;
```
andiwand: can this just be a `std::vector`?

Examples/Io/HepMC3/src/HepMC3InputConverter.cpp
```diff
@@ -105,6 +107,68 @@ std::string printListing(const auto& vertices, const auto& particles) {
...
+  std::unordered_set<int> visited;
```
andiwand, line 114: is `int` the correct upstream hepmc type?

Core/include/Acts/Definitions/PdgParticle.hpp
delitez: 0, 4, 5 come from the quark flavour numbering scheme, I suggest to keep these numbers as they are.

andiwand: makes sense 👍 but then this should be documented

Examples/Io/HepMC3/src/HepMC3InputConverter.cpp
fgrosa: Yes it is, see https://gitlab.cern.ch/hepmc/HepMC3/-/blob/master/include/HepMC3/GenParticle.h?ref_type=heads#L151

Examples/Io/HepMC3/include/ActsExamples/Io/HepMC3/HepMC3InputConverter.hpp
```diff
@@ -50,6 +52,9 @@ class HepMC3InputConverter : public IAlgorithm {
 
     /// If true, check the consistency of the generated event.
     bool checkConsistency = false;
+
+    /// Search up to quark in HF tagging
+    bool searchUpToHfQuark = false;
```
andiwand:
```suggestion
    /// Search up to quark in heavy-flavour tagging
    bool searchUpToHeavyFlavourQuark = false;
```

Fatras/include/ActsFatras/EventData/HeavyFlavourOrigin.hpp
andiwand, line 1: @paulgessinger and I discussed this yesterday after I made my comments and we think it would be best for now to push this into Examples and attach the information onto the `SimParticle` instead of the Fatras particle. Sorry for the second rotation... does that work for you @fgrosa ?

Fatras/include/ActsFatras/EventData/HeavyFlavourOrigin.hpp
```diff
@@ -0,0 +1,38 @@
...
+///         depending on the searchUpToHfQuark flag
+/// Charm: decay product of a charmed hadron, or particle originate
+///        in the fragmentation of a charm quark depending
+///        on the searchUpToHfQuark flag
+/// None: no heavy-flavour origin, depending on the criterion regulated
+///       by the searchUpToHfQuark flag
+enum class HeavyFlavourOrigin : std::uint8_t {
+  None = 0,
+  Charm = 4,
+  Bottom = 5
```
andiwand: you could attach the docstrings to the enum literals

```suggestion
enum class HeavyFlavourOrigin : std::uint8_t {
/// None: no heavy-flavour origin, depending on the criterion regulated
///       by the searchUpToHfQuark flag
  None = 0,
  /// Charm: decay product of a charmed hadron, or particle originate
///        in the fragmentation of a charm quark depending
///        on the searchUpToHfQuark flag
  Charm = 4,
  /// Bottom: decay product of a bottomed hadron (it can also be B->D->X),
///         or particle originate in the fragmentation of a bottom quark
///         depending on the searchUpToHfQuark flag
  Bottom = 5
```

andiwand (commented):
left a few more comments @fgrosa - afterwards good to go from my side 👍

Fatras/include/ActsFatras/EventData/HeavyFlavourOrigin.hpp
fgrosa: Hi @andiwand sure, there is no problem, I will ping you when done!

fgrosa: @andiwand it should be done in [ee190d8](https://github.com/acts-project/acts/pull/5630/commits/ee190d8ea4bec6b24ad17501205d056558cb9b04), could you check that the implementation is the one you expected?

Examples/Io/HepMC3/include/ActsExamples/Io/HepMC3/HepMC3InputConverter.hpp
```diff
@@ -64,6 +69,9 @@ class HepMC3InputConverter : public IAlgorithm {
   void convertHepMC3ToInternalEdm(const AlgorithmContext& ctx,
                                   const HepMC3::GenEvent& genEvent) const;
 
+  ActsFatras::HeavyFlavourOrigin deriveHeavyFlavourOrigin(
```
andiwand:
```suggestion
  HeavyFlavourOrigin deriveHeavyFlavourOrigin(
```

andiwand (commented):
I may have found one potential problem. let me also trigger the ci for you

fgrosa:
> I may have found one potential problem. let me also trigger the ci for you

Indeed I see the problem now, I had to remove the `HeavyFlavourOrigin` from the `Particle` because otherwise I had to include `"SimParticle.hpp"` in `Particle.hpp` and viceversa, but this does not work. Do you have a solution or should be revert the latest commits and define it in `ActsFatras` again?

Examples/Framework/include/ActsExamples/EventData/SimParticle.hpp
```diff
@@ -111,13 +132,33 @@ class SimParticle final {
...
+  /// Original particle index (to match HepMC).
+  SimParticle& setOrigParticleIdx(std::uint32_t idx) {
+    initialState().setOrigParticleIdx(idx);
+    finalState().setOrigParticleIdx(idx);
+    return *this;
+  }
+  /// Particle heavy-flavour origin (0->none, 4->charm, 5->beauty)
+  SimParticle& setHeavyFlavourOrigin(HeavyFlavourOrigin origin) {
+    initialState().setHeavyFlavourOrigin(origin);
+    finalState().setHeavyFlavourOrigin(origin);
```
andiwand:
```suggestion
    m_hfOrigin = origin;
```
shouldn't that do it?

Examples/Framework/include/ActsExamples/EventData/SimParticle.hpp
```diff
@@ -111,13 +132,33 @@ class SimParticle final {
...
   SimBarcode parentParticleId() const {
     return initialState().parentParticleId();
   }
+  /// Original particle index (to match HepMC)
+  std::uint32_t origParticleIdx() const {
+    return initialState().origParticleIdx();
+  }
+  /// Particle heavy-flavour origin (0->none, 4->charm, 5->beauty)
+  HeavyFlavourOrigin heavyFlavourOrigin() const {
+    return initialState().heavyFlavourOrigin();
```
andiwand:
```suggestion
    return m_hfOrigin;
```

Examples/Framework/include/ActsExamples/EventData/SimParticle.hpp
fgrosa: Ah right!

Examples/Io/HepMC3/src/HepMC3InputConverter.cpp
```diff
@@ -105,6 +108,70 @@ std::string printListing(const auto& vertices, const auto& particles) {
...
 
+ActsExamples::HeavyFlavourOrigin HepMC3InputConverter::deriveHeavyFlavourOrigin(
+    const std::shared_ptr<const HepMC3::GenParticle>& particleToCheck) const {
+  std::vector<std::shared_ptr<const HepMC3::GenParticle>> st;
+  std::unordered_set<int> visited;
+  st.push_back(particleToCheck);
+
+  bool isFromCharm{false};
+  while (!st.empty()) {
+    auto part = std::move(st.back());
```
andiwand:
```suggestion
    const std::shared_ptr<const HepMC3::GenParticle> part = st.back();
```
in principle the move should not be necessary but it avoids one refcount increase. but I don't think we are bound by that here performance wise

andiwand:
@fgrosa can you add an ignore statement here so physmon goes green without needing to update the reference? https://github.com/acts-project/acts/blob/main/CI/physmon/config/simulation.yml#L67

andiwand:
sorry there is one more file to update @fgrosa I forgot about https://github.com/acts-project/acts/blob/main/CI/physmon/config/pythia8_ttbar.yml - similar to the `simulation` one

andiwand:
@fgrosa linter is complaining - but that should be really the last thing to fix

andiwand (approved):

PR #5639, opened 2026-07-06
feat: part 1 of GBTS layer connection training tool implementation in ACTS 

jpreston-cern:
Adds the layer connection tool needed for creating the layer map to ACTS so it can be trained in house, it works by taking the RZ coordinates on hits in a truth track to then find the corresponding layer associated with that hit with the transition between each pair of hits then recorded.

From this, the transitions that pass a probability threshold are kept and printed out in either a new simplified format or the old Athena format (which will be depreciated in future).

It also adds the feature of being able to symmetrize the table to account for holes in the training based on stats, but this will only work for the an "idealized" symmetric detector and should be turned of when run on a real (aligned) detector.

The naming of the file itself is a little vague, but this is due to wanting it to be a general area in which all the training tools can be found. Such as the ones needed for tuning the cuts on graph building and filtering.

Future PR's will focus on adding direct eta bin training (for GPU but also potential performance increase and tuning of eta binning in general) and adding direct ID loading in `addTrack` with corresponding changes in the GBTS core algorithm to get rid of dependency on specific ID naming

--- END COMMIT MESSAGE ---

@andiwand @timadye @mimodekjaer 

TO DO:

add docs to pass CI test
clang tidy 
variable name changes in `recontruction.py`


jpreston-cern:
using a table trained on 1,000,000 single muon events, running on 2000 ttbar200 events, the  [dcube plots](https://hepunx.rl.ac.uk/webtempfiles/jpreston/dcube_gbtsacts_1000kSingleMu_gbtsftf/)  gives good agreement with the previous table although these are without any probability cut which gives around a 10 percent slowdown in performance (currently looking at what the correct cut should be)

Core/include/Acts/Seeding2/GbtsTrainingTool.hpp
```diff
@@ -0,0 +1,101 @@
...
+//
+// Copyright (C) 2016 CERN for the benefit of the ACTS project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// includes needed
+#pragma once
+#include "Acts/Utilities/Logger.hpp"
```
andiwand:
```suggestion
#pragma once

#include "Acts/Utilities/Logger.hpp"
```

Core/include/Acts/Seeding/GbtsTrainingTool.hpp
```diff
@@ -0,0 +1,101 @@
...
+#include <memory>
+#include <optional>
+#include <set>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace Acts::Experimental {
+
+class GbtsLayerConnectionTool {
```
andiwand, line 25: all of this could use some doc: `class` / `struct` / `variables`

Core/include/Acts/Seeding2/GbtsTrainingTool.hpp
```diff
@@ -0,0 +1,101 @@
...
+    float probThreshold = -1;
+  };
+  struct TrackCoordinates {
+    float r{};
+    float z{};
+  };
+
+  struct LayerDescription {
+    LayerDescription(float minR_, float maxR_, float minZ_, float maxZ_,
+                     std::int32_t gbtsId_);
```
andiwand: can we reuse / repurpose the existing layer description?

Core/include/Acts/Seeding/GbtsTrainingTool.hpp
```diff
@@ -0,0 +1,101 @@
...
+  using LayerIdPair = std::pair<std::int32_t, std::int32_t>;
+
+  struct LayerIdPairHash {
+    std::size_t operator()(const LayerIdPair& pair) const noexcept {
+      const auto h1 = std::hash<std::int32_t>{}(pair.first);
+      const auto h2 = std::hash<std::int32_t>{}(pair.second);
+
+      return h1 ^ (h2 << 1);
+    }
+  };
```
andiwand, line 95: below I see you use `std::set` and `std::unordered_map`. in case `std::map` is enough you could also drop the custom hasher

Core/include/Acts/Seeding2/GbtsTrainingTool.hpp
```diff
@@ -0,0 +1,101 @@
...
+      const auto h1 = std::hash<std::int32_t>{}(pair.first);
+      const auto h2 = std::hash<std::int32_t>{}(pair.second);
+
+      return h1 ^ (h2 << 1);
+    }
+  };
+  explicit GbtsLayerConnectionTool(
+      Config config, const std::string& geometryInformation,
+      std::unique_ptr<const Acts::Logger> logger =
+          Acts::getDefaultLogger("Finder", Acts::Logging::Level::INFO));
```
andiwand:
```suggestion
      std::unique_ptr<const Logger> logger =
          getDefaultLogger("GbtsLayerConnectionTool", Logging::Level::INFO));
```
- logger name
- `Acts::` prefix should not be necessary

Core/include/Acts/Seeding2/GbtsTrainingTool.hpp
```diff
@@ -0,0 +1,101 @@
...
+  struct LayerIdPairHash {
+    std::size_t operator()(const LayerIdPair& pair) const noexcept {
+      const auto h1 = std::hash<std::int32_t>{}(pair.first);
+      const auto h2 = std::hash<std::int32_t>{}(pair.second);
+
+      return h1 ^ (h2 << 1);
+    }
+  };
+  explicit GbtsLayerConnectionTool(
+      Config config, const std::string& geometryInformation,
```
andiwand: I think instead of requiring an existing file as input with some predefined format it would be more generic to either input existing ACTS objects, for example would the tracking geometry do it?, or rather define a "parsed" version using `struct` and likely `std::vector`

Core/include/Acts/Seeding2/GbtsTrainingTool.hpp
```diff
@@ -0,0 +1,101 @@
...
+    float zMaxTol = 0.2340f;
+    float rMinTol = 2.5337f;
+    float rMaxTol = 2.5337f;
+
+    bool doSymmetrization = false;
+    bool useOldFormatting = false;
+
+    float probThreshold = -1;
+  };
+  struct TrackCoordinates {
```
andiwand: maybe rather `Hit`? can / should this also carry the geometry ID?

Core/include/Acts/Seeding2/GbtsTrainingTool.hpp
andiwand: also
- should `geometryInformation` rather be part of the config?
- should `config` be passed by `const &`?

Core/include/Acts/Seeding2/GbtsTrainingTool.hpp
jpreston-cern: the previous description only has an average for either z or r (not bounds) depending on if it is a endcap or barrel, maybe if we wanted to use the old one we could inherit that and add the extra bounds? but it might be simpler to keep them separate as the variables used in the original `GbtsLayerDiscription` is quite vague

Core/include/Acts/Seeding/GbtsTrainingTool.hpp
jpreston-cern: originally i used a `std::map` but was worried about the speed due to not being able to reserve and look up being slower, but now changed it so both sets and maps are now unordered to use the same hash

Core/include/Acts/Seeding2/GbtsTrainingTool.hpp
jpreston-cern: I think in future it should be for sure for when you want to input GBTS ID directly, but i need to work out the best way to pull the "GBTS ID dependencies" out of Core before it can be used, In a future PR ill probably do that all in one go basically (both refactor of core and adding hit level ID's)

andiwand:
@jpreston-cern can you have a look at the clang-tidy failure and the conflict?

Core/include/Acts/Seeding2/GbtsTrainingTool.hpp
```diff
@@ -0,0 +1,154 @@
...
+  void createConnectionTable(const std::string& outputFileLocation) const;
+
+ private:
+  /// returns the Acts logger
+  const Acts::Logger& logger() const { return *m_logger; }
+
+  /// Finds the Gbts Coordinate of a given hits coordinate
+  /// @param r radius
+  /// @param z z coordinate
+  std::optional<std::int32_t> findGbtsIdByCoord(float r, float z) const;
```
andiwand: should this also consume a `HitCoordinates`?

Core/include/Acts/Seeding2/GbtsTrainingTool.hpp
```diff
@@ -0,0 +1,154 @@
...
+  /// @param outputFileLocation the location for the layer connection table
+  /// @param tempTable the temporary container for holding layer transitions before sorting
+  void oldStyleFormatting(const std::string& outputFileLocation,
+                          const LayerIdPairs& tempTable) const;
+
+  /// Config for layer connection tool
+  Config m_cfg;
+  /// Acts logger
+  std::unique_ptr<const Acts::Logger> m_logger =
+      Acts::getDefaultLogger("Finder", Acts::Logging::Level::INFO);
```
andiwand:
```suggestion
  std::unique_ptr<const Acts::Logger> m_logger;
```
default is coming from the constructor

Core/include/Acts/Seeding2/GbtsTrainingTool.hpp
```diff
@@ -0,0 +1,154 @@
...
+  using LayerIdPairs = std::unordered_set<LayerIdPair, LayerIdPairHash>;
+  /// Map of layer pair transitions, quantifying the amount of times they occur
+  using LayerIdPairMap =
+      std::unordered_map<LayerIdPair, std::uint32_t, LayerIdPairHash>;
+
+  /// Constructor for layer connection training tool
+  /// @param config The training tool config
+  /// @param logger The Acts logger
+  explicit GbtsLayerConnectionTool(
+      Config& config,
```
andiwand:
```suggestion
      const Config& config,
```

Core/include/Acts/Seeding2/GbtsTrainingTool.hpp
```diff
@@ -0,0 +1,154 @@
...
+      std::unique_ptr<const Logger> logger = Acts::getDefaultLogger(
+          "GbtsLayerConnectionTool", Acts::Logging::Level::INFO));
```
andiwand:
```suggestion
      std::unique_ptr<const Logger> logger = getDefaultLogger(
          "GbtsLayerConnectionTool", Logging::Level::INFO));
```

Core/include/Acts/Seeding2/GbtsTrainingTool.hpp
```diff
@@ -0,0 +1,154 @@
...
+  /// converts layer hits to layer transitions
+  /// @param track the layer hits of a particle
+  void addTrack(const std::vector<HitCoordinates>& track);
+  /// Creates the connection table
+  /// @param outputFileLocation the location for the layer connection table
+  void createConnectionTable(const std::string& outputFileLocation) const;
+
+ private:
+  /// returns the Acts logger
+  const Acts::Logger& logger() const { return *m_logger; }
```
andiwand:
```suggestion
  const Logger& logger() const { return *m_logger; }
```

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/GbtsTrainingAlgorithm.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+  for (std::uint32_t l = 0; l < lines; l++) {
+    inStream >> minR >> maxR >> minZ >> maxZ >> gbtsId;
+
+    detectorGeometry.emplace_back(minR, maxR, minZ, maxZ, gbtsId);
+  }
+}
+
+GbtsTrainingAlgorithm::GbtsTrainingAlgorithm(
+    const Config& config, std::unique_ptr<const Acts::Logger> inputLogger)
```
andiwand, line 75:
```suggestion
    const Config& config, std::unique_ptr<const Logger> inputLogger)
```

Core/src/Seeding2/GbtsTrainingTool.cpp
```diff
@@ -0,0 +1,259 @@
...
+#include <string>
+
+namespace Acts::Experimental {
+
+GbtsLayerConnectionTool::LayerDescription::LayerDescription(
+    float minR_, float maxR_, float minZ_, float maxZ_, std::int32_t gbtsId_)
+    : minR(minR_), maxR(maxR_), minZ(minZ_), maxZ(maxZ_), gbtsId(gbtsId_) {}
+
+GbtsLayerConnectionTool::GbtsLayerConnectionTool(
+    Config& config, std::unique_ptr<const Logger> logger)
```
andiwand:
```suggestion
    const Config& config, std::unique_ptr<const Logger> logger)
```

Core/src/Seeding/GbtsTrainingTool.cpp
```diff
@@ -0,0 +1,259 @@
...
+  // layers
+  for (std::uint32_t i = 0; i < m_cfg.detectorGeometry.size(); i++) {
+    for (std::uint32_t j = 0; j < m_cfg.detectorGeometry.size(); j++) {
+      if (i == j) {
+        continue;
+      }
+
+      LayerIdPair pair;
+      pair.first = m_cfg.detectorGeometry[i].gbtsId;
+      pair.second = m_cfg.detectorGeometry[j].gbtsId;
```
andiwand, line 44: would it make sense to symmetrize the pair by sorting `first, second` and treat `second, first` as equivalent?

Core/src/Seeding/GbtsTrainingTool.cpp
```diff
@@ -0,0 +1,259 @@
...
+  // check to see if tracks have backward transitions
+  for (std::uint32_t idx = 0; idx + 1 < track.size(); idx++) {
+    const float hit1Magnitude = std::hypot(track[idx].r, track[idx].z);
+    const float hit2Magnitude = std::hypot(track[idx + 1].r, track[idx + 1].z);
+
+    if (hit1Magnitude > hit2Magnitude) {
+      ACTS_WARNING("Track travels inwards, skipping");
+      return;
+    }
+  }
```
andiwand: note that this is will be unstable which a large z0 / displaced tracks. I would rather add `time` as a hit feature` or explicitly a hit index or rely on the ordering the user provides

Core/src/Seeding2/GbtsTrainingTool.cpp
```diff
@@ -0,0 +1,259 @@
...
+  // container for gbts IDs of the track
+  std::vector<std::optional<std::int32_t>> layerGbtsIds{};
+  layerGbtsIds.reserve(track.size());
+
+  // find GBTS ids for all measurements in a track
+  for (const auto& measurement : track) {
+    const float r = measurement.r;
+    const float z = measurement.z;
+
+    const auto gbtsId = findGbtsIdByCoord(r, z);
```
andiwand:
```suggestion
  for (const auto& measurement : track) {
    const auto gbtsId = findGbtsIdByCoord(measurement);
```

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/GbtsTrainingAlgorithm.cpp
```diff
@@ -0,0 +1,184 @@
...
+  geometryParser(m_cfg.geometryFileDir,
+                 m_cfg.gbtsLayerConnectionToolConfig.detectorGeometry);
+
+  m_layerConnectionTool.emplace(
+      m_cfg.gbtsLayerConnectionToolConfig,
+      this->logger().cloneWithSuffix("GbtsLayerConnectionTool"));
+}
+
+GbtsTrainingAlgorithm::~GbtsTrainingAlgorithm() {
+  m_layerConnectionTool->createConnectionTable(m_cfg.outputFileDir);
```
andiwand: I would not do this during destruction but rather in `finalize`

Core/src/Seeding/GbtsTrainingTool.cpp
```diff
@@ -0,0 +1,259 @@
...
+  // finally, add transitions to output file (old or new format)
+  if (m_cfg.useOldFormatting) {
+    oldStyleFormatting(outputFileLocation, tempPairs);
+  } else {
+    outputFile << tempPairs.size() << "\n";
+    for (const auto& layerPair : tempPairs) {
+      // swap order as we want outward -> inward ordering
+      outputFile << layerPair.second << " " << layerPair.first << "\n";
+    }
+  }
```
andiwand: one more reason why I would decouple data structure and IO

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/GbtsTrainingAlgorithm.hpp
```diff
@@ -0,0 +1,95 @@
...
+    /// The layer connection tool config
+    Acts::Experimental::GbtsLayerConnectionTool::Config
+        gbtsLayerConnectionToolConfig;
+    /// geometry file used for creating layers
+    std::string geometryFileDir =
+        "/home/ppd/jasper/QT/acts/gbts_layer_geometry.txt";
+    /// output directory for layer connection table
+    std::string outputFileDir =
+        "/home/ppd/jasper/QT/acts/connection-tables/"
+        "conditional_prob_2kttbar200_new.txt";
```
andiwand: these defaults should be changed

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/GbtsTrainingAlgorithm.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+  /// run time execution per event
+  /// @param ctx event context
+  ProcessCode execute(const AlgorithmContext& ctx) const final;
+
+  /// Get readonly access to the config parameters
+  const Config& config() const { return m_cfg; }
+
+  /// GBTS training algorithm destructor
+  ~GbtsTrainingAlgorithm();
```
andiwand:
```suggestion
  ~GbtsTrainingAlgorithm() override;
```

jpreston-cern:
Conflict and clang tidy came from the refactor back to `Seeding` from `Seeding2` and not using an `override` on the destructor (although got rid of the destructor now and using `finalize()` ) Ive done the easier fixes, so will now look at the hit ordering and last tidy up ones in the next day or so

Core/src/Seeding/GbtsTrainingTool.cpp
jpreston-cern: would a good way to do this to have the Tool itself output a vector of the allowed connections then the Algorithm in `Examples` parse to the file (which is the change i have currently made) ?

Core/src/Seeding/GbtsTrainingTool.cpp
jpreston-cern: I've gone with the third option due to time time ordering needing to take place in the interface anyway

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/GbtsTrainingAlgorithm.cpp
jpreston-cern: I'm not sure that Logger by itself is available in the `ActsExamples` name space (at least i get an error when i remove the `Acts::` prefix

andiwand:
@jpreston-cern docs and clang tidy is still complaining. also can you resolve the open comments that have been addressed?

Core/src/Seeding/GbtsTrainingTool.cpp
jpreston-cern: I think doing it this way removes some connection that where present before, i think this is because in pooling the left and right side probabilities together it can make both transitions fail if one side has a very low probability so think i might have to keep it as it was so it acts as an "or" statement instead of an average of probabilities of each side

Core/src/Seeding/GbtsTrainingTool.cpp
andiwand: yes I would output a vector of connections and the add a static function, for now this can life in core on this class or be a free function to read and write such a vector

Core/src/Seeding/GbtsTrainingTool.cpp
andiwand: let's keep it as is for now

Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/GbtsTrainingAlgorithm.cpp
andiwand: whoops right that can not be done in Examples - never mind

andiwand (approved):

PR #5642, opened 2026-07-06
fix(detray): Surface sequencer in case no surfaces should be collected

niermann999:
This is needed, in case the surface sequence buffer is empty, e.g. when MBF smoothing is enabled instead of Klaman smoothing.

--- END COMMIT MESSAGE ---


andiwand:
@niermann999 @stephenswat @paradajzblond @paulgessinger should we add `(detray)` and co to the title so for example in this case `fix(detray): Surface sequencer in case no surfaces should be collected`, otherwise from the tittle alone I would think this touches ACTS core. the title then goes into the git history and is picked up from there for the release which will likely confuse anyone looking at it since labels are gone from there

stephenswat:
> @niermann999 @stephenswat @paradajzblond @paulgessinger should we add `(detray)` and co to the title so for example in this case `fix(detray): Surface sequencer in case no surfaces should be collected`, otherwise from the tittle alone I would think this touches ACTS core. the title then goes into the git history and is picked up from there for the release which will likely confuse anyone looking at it since labels are gone from there

I fully support this. Only problem is that e.g. "refactor(detray)" takes up 18 characters, but I don't think the convention in ACTS is to strictly adhere to the 50 character commit title limit, right?

andiwand:
right this will start to get a bit long. I don't have a good idea to avoid that. we could abbreviate `refactor`, I don't think it makes sense to abbreviate detray?
but we don't have a char limit on our PR titles, except what GitHub might impose on us so I think it is ok

stephenswat:
I would just go with "refactor(detray)"; it's long but doesn't make the message unreadable.

niermann999:
We could just call it `refactor(GPU)`, no? Which then also covers traccc

stephenswat (approved):
Looks correct to me.

PR #5644, opened 2026-07-06
ci: Disable Dependabot auto-rebase

andiwand:

## Problem

Dependabot's default `rebase-strategy: auto` rebases every open Dependabot PR whenever `main` moves. On this repo with the merge queue, each rebase force-pushes a new head commit, which:

- re-triggers the full CI suite, and
- dismisses stale approvals (branch protection),

knocking the PR out of the merge queue in a continuous loop as the queue advances `main`.

## Change

Set `rebase-strategy: "disabled"` on both update entries (`github-actions` and `pre-commit`). Dependabot no longer auto-rebases open PRs when `main` moves; manual `@dependabot rebase` still works, and PRs are still recreated on the weekly schedule.

Note: this only takes effect once merged to `main`, since Dependabot reads its config from the default branch. Existing open PRs keep their current behavior until recreated.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,9 @@ updates:
       timezone: "Europe/Zurich"
     cooldown:
       default-days: 7
+    # Don't auto-rebase open PRs when main moves: rebases re-trigger CI and
+    # dismiss stale approvals, which knocks PRs out of the merge queue.
+    rebase-strategy: "disabled"
 
   - package-ecosystem: "pre-commit"
     directory: "/"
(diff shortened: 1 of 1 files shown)


paulgessinger:
Dependabot can also make grouped updates per dependency ecosystem. Maybe what would also reduce churn for us?

andiwand:
that might also make sense yes. but with the rebase right now it is very annoying to have them open because they will not go in at all and just consume resources. with this change the main problem should go away

AJPfleger (approved):

Issue #59, opened 2020-03-27
Refine `debugLog` throughout acts-core (was ACTS-457)


paulgessinger:
[GitLab issue: #458](https://gitlab.cern.ch/acts/acts-core/issues/458)

Original author Andreas Salzburger @ asalzbur

paulgessinger:
[Original comment by: msmk at 2019-03-27 16:52:35](https://gitlab.cern.ch/acts/acts-core/issues/458#note_2483892)

 Original author ATS Jenkins @ atsjenkins

[Andreas Salzburger](https://gitlab.cern.ch/asalzbur) mentioned this issue in [a merge request of acts/acts-core](https://gitlab.cern.ch/acts/acts-core/merge_requests/408):
'Extrapolator prototype'

paulgessinger:
[Original comment by: msmk at 2019-03-27 16:52:36](https://gitlab.cern.ch/acts/acts-core/issues/458#note_2483898)

 Original author Hadrien Benjamin Grasland @ hgraslan

Here is a summary of the suggestions that I made on MR 408:
 * We may want to centralize the many debugLog methods in a single place
 * Whether we do this or not, we may want them to converge towards a single interface style.
 * Given that debugLog internally uses an std::stringstream, passing an std::stringstream& to the logging lambda may be both more efficient and cleaner than requiring the lambda to return an std::string.

paulgessinger:
[Original comment by: msmk at 2019-03-27 16:52:36](https://gitlab.cern.ch/acts/acts-core/issues/458#note_2483907)

 Original author Hadrien Benjamin Grasland @ hgraslan

Another suggestion made at the time: we currently have several instances of this set of struct members:
```cpp
    bool        debug         = false;
    std::string debugString   = "";
    size_t      debugPfxWidth = 30;
    size_t      debugMsgWidth = 50;
```
It would probably be better to turn this into a nullable unique_ptr to a debug settings struct, as it will take up less room in memory in release mode and allow less repetition in struct members.

If we like, we'll also be able to factor out common debugLog-like functionality as a member of such a struct.

paulgessinger:
[Original comment by: msmk at 2019-03-27 16:52:37](https://gitlab.cern.ch/acts/acts-core/issues/458#note_2483914)

 Original author Andreas Salzburger @ asalzbur

I agree, I will make such a struct and integrate it first in the Propagator schema.

paulgessinger:
[...]

PR #5646, opened 2026-07-07
feat: Add node feature selection to OnnxEdgeClassifier

Baucki1:
- adding an option to select specific node features to OnnxEdgeClassifier
- adding an option to assign prefactors (scales) to specific node features
- adding template for scaling Tensor by column-wise factors (device-aware)
--- END COMMIT MESSAGE ---

Tagging @madbaron

This is needed for the GNNTrackFinder pipeline in [k4ActsTracking](https://github.com/key4hep/k4ActsTracking) ([#59](https://github.com/key4hep/k4ActsTracking/pull/59)), enabling us to run multiple edge classifiers, each on a different set of node features.

### DIFF
--- a/Plugins/Gnn/include/ActsPlugins/Gnn/Tensor.hpp
+++ b/Plugins/Gnn/include/ActsPlugins/Gnn/Tensor.hpp
@@ -237,5 +237,15 @@ template <Acts::Concepts::arithmetic T>
 Tensor<T> selectCols(const Tensor<T> &tensor, const Tensor<bool> &mask,
                      const ExecutionContext &execContext);
 
+/// Multiply each column of src with the corresponding scale value
+/// dst[r,n] = src[r,n] * scales[n]
+/// @param src Source tensor [R, N] in row-major layout
+/// @param scales Vector of scale factors of length N, same type as src
+/// @param execContext Device and stream for output allocation
+/// @return New tensor [R, N] in row-major layout containing scaled values
+template <Acts::Concepts::arithmetic T>
+Tensor<T> mulPerColumn(const Tensor<T> &src, const std::vector<T> &scales,
+                       const ExecutionContext &execContext);
+
 /// @}
 }  // namespace ActsPlugins
--- a/Plugins/Gnn/src/OnnxEdgeClassifier.cpp
+++ b/Plugins/Gnn/src/OnnxEdgeClassifier.cpp
@@ -124,33 +124,41 @@ PipelineTensors OnnxEdgeClassifier::operator()(
   // Apply feature selection if configured
   std::optional<Tensor<float>> selectedNodeFeatures;
   ActsPlugins::Tensor<float> *nodeFeatures = &tensors.nodeFeatures;
+
   if (!m_cfg.selectedFeatures.empty()) {
-    // Create filtered node features with only selected indices
-    selectedNodeFeatures.emplace(Tensor<float>::Create(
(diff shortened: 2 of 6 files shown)


Plugins/Gnn/src/OnnxEdgeClassifier.cpp
```diff
@@ -121,12 +121,52 @@ PipelineTensors OnnxEdgeClassifier::operator()(
...
+    std::size_t numNodes = tensors.nodeFeatures.shape()[0];
+    std::size_t numAllFeatures = tensors.nodeFeatures.shape()[1];
+
+    for (std::size_t n = 0; n < numNodes; ++n) {
+      for (std::size_t j = 0; j < m_cfg.selectedFeatures.size(); ++j) {
+        int featureIdx = m_cfg.selectedFeatures[j];
+        if (featureIdx < 0 || featureIdx >= static_cast<int>(numAllFeatures)) {
+          throw std::runtime_error("Selected feature index out of range");
+        }
+        dstData[n * m_cfg.selectedFeatures.size() + j] =
```
benjaminhuth: This will not work with CUDA tensors. But you should be able to implement the functionality with an existing function, see:

https://github.com/acts-project/acts/blob/48c2709732162cbeea64883ce817624c9edbad88/Plugins/Gnn/include/ActsPlugins/Gnn/Tensor.hpp#L237

benjaminhuth (changes requested):
Nice addition, but as of now this would not work on CUDA I think.

Plugins/Gnn/src/OnnxEdgeClassifier.cpp
Baucki1: Yes, you are right, thanks. I switched to selectCols with the node feature mask cloned to the respective device now.

Plugins/Gnn/include/ActsPlugins/Gnn/Tensor.hpp
```diff
@@ -237,5 +237,15 @@ template <Acts::Concepts::arithmetic T>
...
                      const ExecutionContext &execContext);
 
+/// Multiply each column of src with the corresponding scale value
+/// dst[r,n] = src[r,n] * scales[n]
+/// @param src Source tensor [R, N] in row-major layout
+/// @param scales Vector of scale factors of length N, same type as src
+/// @param execContext Device and stream for output allocation
+/// @return New tensor [R, N] in row-major layout containing scaled values
+template <Acts::Concepts::arithmetic T>
+Tensor<T> mulPerColumn(const Tensor<T> &src, const std::vector<T> &scales,
```
benjaminhuth, line 247: Looks good from my side, could you maybe implement a small UnitTest for this? The existing tests have some patterns for CPU and GPU tests that you could use.

benjaminhuth (commented):
Looks good, only one small comment!

Plugins/Gnn/include/ActsPlugins/Gnn/Tensor.hpp
Baucki1: Thanks for the pointer. I added something similar for mulPerColumn. This just tests that the multiplication yields the expected results on the same example tensor as for the other tests. Is this what you had in mind?

benjaminhuth: Looks good!

benjaminhuth (dismissed):
LGTM!

benjaminhuth (dismissed):

benjaminhuth (approved):

PR #5649, opened 2026-07-07
perf: Speed up GSF component reduction

benjaminhuth:


Core/include/Acts/TrackFitting/detail/GsfComponentMerging.hpp
```diff
@@ -346,37 +349,50 @@ std::tuple<BoundVector, BoundMatrix> mergeGaussianMixture(
...
+  std::vector<std::size_t> m_originalToActive;
   std::size_t m_numberComponents{};
+  std::size_t m_numberActive{};
 
-  template <typename array_t, typename setter_t>
-  void setAssociated(std::size_t n, array_t &array, const setter_t &setter) {
-    const std::size_t indexConst = (n - 1) * n / 2;
+  static std::size_t triangularIndex(std::size_t i, std::size_t j) {
+    return i * (i - 1) / 2 + j;
+  }
```
junggjo9, line 371: Does that to the same thing as these two methods?

https://github.com/acts-project/acts/blob/main/Core/include/Acts/Utilities/AlgebraHelpers.hpp#L258-L293

Core/src/TrackFitting/detail/GsfComponentMerging.cpp
```diff
@@ -85,38 +93,111 @@ SymmetricKLDistanceMatrix::SymmetricKLDistanceMatrix(
...
+  std::size_t i = 0;
+  for (; i + BLOCK <= nActivePairs; i += BLOCK) {
+    double blockMin = data[i];
+    for (std::size_t k = 1; k < BLOCK; ++k) {
+      blockMin = std::min(blockMin, data[i + k]);
+    }
+    if (blockMin < currentMin) {
+      currentMin = blockMin;
+      currentBlockStart = i;
+      currentBlockLen = BLOCK;
```
junggjo9, line 160:
```suggestion
      currentBlockLen = std::min(BLOCK, BLOCK -k);
```
if you've a block of 50 an the minimum appears at 49, the block length would go out of scope, no?

Tests/UnitTests/Core/TrackFitting/GsfMixtureReductionTests.cpp
```diff
@@ -288,6 +288,87 @@ BOOST_AUTO_TEST_CASE(test_naive_vs_optimized) {
...
+  std::shared_ptr<PlaneSurface> surface =
+      CurvilinearSurface(Vector3{0, 0, 0}, Vector3{1, 0, 0}).planeSurface();
+
+  std::mt19937 rng(1234);
+  std::uniform_real_distribution<double> weightDist(0.5, 1.5);
+  std::uniform_real_distribution<double> qopDist(0.1, 5.0);
+  std::uniform_real_distribution<double> covDist(0.5, 2.0);
+
+  for (int seed = 0; seed < 500; ++seed) {
+    const std::size_t NComps = 72;
```
junggjo9, line 332:
```suggestion
    constexpr std::size_t NComps = 72;
```

Tests/UnitTests/Core/TrackFitting/GsfMixtureReductionTests.cpp
```diff
@@ -288,6 +288,87 @@ BOOST_AUTO_TEST_CASE(test_naive_vs_optimized) {
...
+    std::vector<GsfComponent> cmps;
```
junggjo9, line 333: Maybe use

std:;array<GsfComponent, NComps> cmps = filledArray<GsfComponent, NComps>(makeDefaultComponent(weightDist(rng));


???

junggjo9 (commented):
Minor stuff

Core/include/Acts/TrackFitting/detail/GsfComponentMerging.hpp
benjaminhuth: Maybe, but these functions are for a compile-time fixed matrix, which we don't have here

Core/src/TrackFitting/detail/GsfComponentMerging.cpp
benjaminhuth: No, this should be avoided by the loop condition `i + BLOCK <= nActivePairs`. The idea is that we don't introduce dependencies on loop variable `k` to enable vectorization.

Tests/UnitTests/Core/TrackFitting/GsfMixtureReductionTests.cpp
benjaminhuth: Well the interface of `reduceMixtureWithKLDistance` requires `std::vector`...

Tests/UnitTests/Core/TrackFitting/GsfMixtureReductionTests.cpp
benjaminhuth: could do it, but I think in a unit test this is fine...

Core/include/Acts/TrackFitting/detail/GsfComponentMerging.hpp
```diff
@@ -346,37 +349,50 @@ std::tuple<BoundVector, BoundMatrix> mergeGaussianMixture(
...
+/// that are still present (rather than scanning a mask over the full,
+/// never-shrinking array). All public methods take and return the original
+/// component indices (indices into the range passed to the constructor); the
+/// swap-based compaction is purely an internal bookkeeping detail.
+///
 class SymmetricKLDistanceMatrix {
   using Array = Eigen::Array<double, Eigen::Dynamic, 1>;
-  using Mask = Eigen::Array<bool, Eigen::Dynamic, 1>;
 
   Array m_distances;
```
andiwand, line 362: does this have any benefit over `std::vector` at this point?

Core/src/TrackFitting/detail/GsfComponentMerging.cpp
```diff
@@ -47,8 +50,10 @@ GsfComponent detail::Gsf::mergeTwoComponents(const GsfComponent &a,
   return ret;
 }
 
-double detail::Gsf::computeSymmetricKlDivergence(const GsfComponent &a,
-                                                 const GsfComponent &b) {
+namespace detail::Gsf {
+
+double computeSymmetricKlDivergence(const GsfComponent &a,
```
andiwand: this is now unbound to any namespace. if definition and declaration gets out of sync it can only be caught during linking

Core/src/TrackFitting/detail/GsfComponentMerging.cpp
```diff
@@ -85,38 +93,111 @@ SymmetricKLDistanceMatrix::SymmetricKLDistanceMatrix(
...
+  // one fixed-size block instead of the whole array. This replaces a
+  // two-pass search (vectorized min-value pass + full-array early-exit
+  // index scan) with one pass plus a small bounded rescan, avoiding the
+  // O(nActivePairs/2)-average scalar index search of the two-pass approach.
+  // BLOCK=16 was found to be the sweet spot for the array sizes that
+  // dominate this reduction (N=48-72 active components, i.e. up to ~2556
+  // active pairs).
+  constexpr std::size_t BLOCK = 16;
+
+  const double *data = m_distances.data();
```
andiwand: could be a span to avoid raw pointer

andiwand (dismissed):

Core/include/Acts/TrackFitting/detail/GsfComponentMerging.hpp
benjaminhuth: Eigen does explicit alignment, not sure if this is important here, but I would leave it like that.

Core/src/TrackFitting/detail/GsfComponentMerging.cpp
benjaminhuth: Probably we should use m_distances directly...

andiwand (approved):

PR #5651, opened 2026-07-07
fix(detray): preserve actor chain state order

### LINKED ISSUE #5517: Ordering in the actor chain state tuple is not preserved
When accessing the flattened tuple of actor states (actor state references)  that the actor chain expects to be handed to the propagator the order of the states is not preserved with respect to the actor type pack that the actor chain was instantiated with. This lead to some unexpected bugs in downstream code that made assumptions on the ordering of the actor state types.

hadronic-arvind:
## Summary

- Preserve actor-chain state tuple ordering while removing duplicate state types.
- Reuse `types::push_back_unique` from `type_list.hpp` for tuple de-duplication.
- Add regression coverage for composite observing-actor `state_tuple` and `state_ref_tuple` order.
- Update tuple/type-list helper tests to assert stable left-to-right uniqueness.
- Keep the now-vendored Traccc Kalman fitter state references in the preserved actor-chain order.
- Update Traccc's backward Kalman-filter comparison and Detray's validation helper to consume states by the preserved order and type.

## Motivation

Fixes #5517.

`unique_t` previously removed duplicate tuple types by rotating entries, which could reorder actor state tuples.
Downstream code expects the flattened actor-chain state tuple to follow the actor dependency tree order.

## Upstream integration

The branch is merged with current `main` at `606efeb0d`.
Since Traccc is now integrated directly into ACTS, the old `thirdparty/traccc` FetchContent patch was replaced by the equivalent change in `Traccc/core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp`.

## Validation

- All repository hooks pass on the nine PR-owned files.
- A fresh Release/C++20 build configured successfully with `ACTS_BUILD_TRACCC=ON`, `ACTS_BUILD_PLUGIN_TRACCC=ON`, `ACTS_SETUP_DETRAY=ON`, `DETRAY_BUILD_UNITTESTS=ON`, and `DETRAY_SETUP_GOOGLETEST=ON`.
- `detray_unit_test_cpu`, `detray_unit_test_cpu_array`, `traccc_core`, `traccc_performance`, and the full `traccc_test_cpu` executable built successfully with four parallel jobs.
- `detray_unit_test_cpu`: 31/31 tests passed.
- `detray_unit_test_cpu_array`: all 271 locally runnable tests passed, including the nested observing-actor ordering regression.
- The remaining inhomogeneous-field test requires a `DETRAY_BFIELD_FILE` fixture that is not present in the local devcontainer.
- `git diff upstream/main...HEAD --check` passes.


Detray/tests/unit_tests/cpu/propagator/actor_chain.cpp
```diff
@@ -116,6 +116,14 @@ using composite3 = composite_actor<example_actor_t, composite1>;
 // Implements example_actor through composite2<-composite3 with composite1 obs.
 using composite4 = composite_actor<example_actor_t, composite1>;
 
+template <int I>
+struct ordered_actor : public detray::base_actor {
+  struct state {};
+};
+
+using ordered_actor_chain =
```
niermann999: I think, this test would make more sense, if tested with an observing actor pattern, where the order of the state tuple then depends on the order in which the actor dependency tree is resolved. This is where the problem that was mentioned in the issue lies. We rarely have duplicate actor types in the tuple, but observing actors do appear in conjunction with the parameter updater

Basically, if you have something like this:
```
  using comp_actor = actor::some_actor<observer1, observer2, obsever3<observer4>>;

  using actor_chain_t = actor_chain<comp_actor>;
```
check that the state tuple comes out as
```
dtuple<some_actor::state, observer1::state, observer2::state, observer3::state, observer4::state>
```

Detray/core/include/detray/utils/tuple_helpers.hpp
```diff
@@ -204,51 +204,50 @@ using tuple_cat_t = typename tuple_cat_type<tuple_ts...>::type;
...
 /// @{
-template <std::size_t I, typename... tuple_ts>
-struct unique_types {};
-
-/// No elements in tuple
-template <std::size_t I>
-struct unique_types<I, std::tuple<>> {
-  using type = std::tuple<>;
+template <typename result_tuple_t, typename arg_t>
+struct push_back_unique_type {};
```
niermann999: Could this perhaps be done by adding `push_back_unique` to the existing type list implementation, like e.g. https://github.com/acts-project/acts/blob/cf14abc44fdb28a4d1790d3fcb2544adea2fb6e8/Detray/core/include/detray/utils/type_list.hpp#L117-L125
 and then simply convert the type list into a tuple? This way the push back tool would be easier to reuse

niermann999 (commented):
Thank you for the fix!

hadronic-arvind:
Can a maintainer run the CI Bridge Pipeline please.

hadronic-arvind:
Current-head handoff:

- Merged with current `main` at `606efeb0d`.
- All fork workflows are green on `98fd81b83`, including the full Traccc and Detray matrices, Builds, Docs, Checks, and devcontainer.
- Repository hooks, the full local Traccc CPU test executable build, Detray unit tests, and diff checks pass.
- Both review threads are resolved.

Could a maintainer please trigger CI Bridge for this head and review/approve it?
Once policy-bot is satisfied, this is ready for automerge.

PR #5652, opened 2026-07-07
fix: PyPI build, improve CI

benjaminhuth:
Fixes the pypi build that was broken due to a segfault. Also refactors the CI so that `cibuildwheel` is actually run in the CI for each PR for Linux and python 3.13. Implements the `cibuildwheel` call in `CI/cibuildhweel.sh`


paulgessinger:
So what was the segfault from? Do we know?

CI/cibuildwheel.sh
```diff
@@ -0,0 +1,34 @@
...
+export CI
+export CIBW_MANYLINUX_X86_64_IMAGE="manylinux_2_34" # based on almalinux9
+export CIBW_SKIP="*-musllinux* *-manylinux_i686"
+SETUP_CMD="bash {package}/CI/dependencies/setup.sh -t v23.2.0 -d deps -e env.sh"
+export CIBW_BEFORE_ALL_LINUX="dnf install -y bc ccache && ${SETUP_CMD}"
+export CIBW_BEFORE_ALL_MACOS="brew install ninja ccache && ${SETUP_CMD}"
+export CIBW_ENVIRONMENT="GITHUB_TOKEN=${GITHUB_TOKEN:-}"
+export CIBW_ENVIRONMENT_PASS="CI"
+export CIBW_BEFORE_BUILD="ccache -z"
+export CIBW_ENVIRONMENT_LINUX="CMAKE_PREFIX_PATH=\$PWD/deps/venv:\$PWD/deps/view CCACHE_DIR=/host/\$PWD/ccache LD_LIBRARY_PATH=\$PWD/deps/view/lib64:\$PWD/deps/view/lib:\$PWD/deps/venv/lib64:\$PWD/deps/venv/lib"
```
paulgessinger: I think this means the action-level ccache dir will stay empty.

paulgessinger (approved):

PR #5447, opened 2026-05-13
refactor(detray): Exclude the ACTS unittests from detray workflows

niermann999:
Exclude changes in the Detray directory from triggering ACTS builds and instead adding a dedicated plugin build to the detray workflow

--- END COMMIT MESSAGE ---

PR #5653, opened 2026-07-08
ci: Skip build and analysis flows for Detray PRs

stephenswat:
This commit uses the `paths-ignore:` mechanism to skip the analysis and build steps for changes that touch only detray-related files.

--- END COMMIT MESSAGE ---

Tagging @niermann999 @andiwand. Hopefully this will take some load off the CI.


.github/workflows/analysis.yml
```diff
@@ -11,6 +11,7 @@ on:
       - 'develop/**'
     paths-ignore:
       - "docs/**"
+      - "Detray/**"
```
niermann999: I think we cannot skip all of the analysis jobs or we loose e.g. clang-tidy

.github/workflows/builds.yml
```diff
@@ -9,6 +9,7 @@ on:
       - 'develop/**'
     paths-ignore:
       - "docs/**"
+      - "Detray/**"
```
niermann999: If this is turned off, we need a dedicated small workflow that runs the detray plugin. See #5447

paulgessinger: I would prefer to keep the builds but we can skip the downstream tests if you want

stephenswat:
@paulgessinger whenever you have time check out the `build.yml` in this file. It grabs the changeset and computes which components (e.g. Core, Examples, etc.) are touched. It then uses that info to determine whether to run the physmon or not. Would something like that be good?

Ignore the fact that I nuked the entire CI apart from that, that's just so I can test this on GitHub without taking up a ton of resources.

PR #5655, opened 2026-07-08
refactor: Drop residual std::enable_if

paulgessinger:


Core/include/Acts/EventData/SpacePointContainer.hpp
andiwand, line 1: actually these ones we should kill. I might have overlooked them in the seeding cleanup / rotation prs

paulgessinger: Fine by me

PR #5656, opened 2026-07-08
feat: Estimate track parameters from N space points

andiwand:
`estimateTrackParamsFromSpacePoints` fits N >= 3 space points as a
least-squares helix, where `estimateTrackParamsFromSeed` puts an exact helix
through three: a Taubin circle fit transverse to the field, optionally
refined with Gauss-Newton steps, and a linear fit of the field coordinate
against the transverse arc length. Collinear input falls back to a principal
axis with q/p at zero. An optional per-point weight expresses relative
trust; an empty span is uniform, so existing callers are unchanged.


andiwand:
Notes for review, moved out of the description now that it is the commit message verbatim.

A single scalar weight per point is used rather than a transverse/longitudinal split: the estimation frame is a general rotation fixed by the field, so a diagonal split would not be rotation-invariant and is only meaningful when the field is along global z.

Worth a look: on the straight-line fallback the function succeeds with `q/p` exactly zero, and does not distinguish a vanishing field, where the momentum is genuinely unmeasurable, from a track too stiff to resolve the curvature. The three-point estimator has the same behaviour, but only for exactly collinear points, whereas the N-point fit has an explicit straightness threshold and so reaches it far more often.

Tests: Taubin fit against a known circle, collinear input reported invalid, weighted fits recovering truth where the unweighted fit is biased, an N=3 cross-check against the closed-form estimator, large-arc phi unwrapping and the zero-field fallback.

Core/include/Acts/Seeding/detail/CircleFit.hpp
```diff
@@ -0,0 +1,209 @@
...
+namespace Acts::detail {
+
+/// Result of a circle fit in a plane.
+struct CircleFit {
+  /// Circle center.
+  Vector2 center = Vector2::Zero();
+  /// Circle radius.
+  double radius = 0.;
+  /// False in the straight-line limit, i.e. (near-)collinear input.
+  bool valid = false;
```
benjaminhuth: Hmm is there a reason why this is not modeled with `Acts::Result` or `std::optional`?

Core/include/Acts/Seeding/detail/CircleFit.hpp
```diff
@@ -0,0 +1,209 @@
...
+///
+/// @param points the points whose transverse projection is fitted
+/// @param weights optional per-point weights (empty span = uniform)
+/// @return the fitted circle (with `valid = false` if effectively straight)
+inline CircleFit fitCircleTaubin(std::span<const Vector3> points,
+                                 std::span<const double> weights = {}) {
+  CircleFit result;
+  const std::size_t n = points.size();
+  if (n < 3) {
+    return result;
```
benjaminhuth: e.g., here I would expect either `std::nullopt` or an error code

benjaminhuth (commented):
Looks good, only small comment about `CircleFit` struct and error handling.

benjaminhuth (approved):
LGTM

PR #5659, opened 2026-07-09
feat: enable Arrow/Parquet for PyPI, use pytest in cibuildwheel

benjaminhuth:
Adds a new test marker `pypi` to select tests for the PyPI build inside cibuildwheel. Refactor tests so the test modules can be imported without ROOT enabled (lazy `acts.examples.root` imports).


Python/Core/python/__init__.py
```diff
@@ -68,20 +68,18 @@ def _decoratorFromFile(file: Union[str, Path], **kwargs):
...
+        from .json import MaterialMapJsonConverter, JsonMaterialDecorator
+
         c = MaterialMapJsonConverter.Config()
         for k in kwargs.keys():
             if hasattr(c, k):
                 setattr(c, k, kwargs.pop(k))
 
         return JsonMaterialDecorator(jFileName=str(file), rConfig=c, **kwargs)
     elif file.suffix == ".root":
+        from .root import RootMaterialDecorator
```
paulgessinger, line 81:
```suggestion
        from acts.root import RootMaterialDecorator
```

CI/cibuildwheel.sh
```diff
@@ -27,8 +27,8 @@ export CIBW_ENVIRONMENT_PASS="CI"
 export CIBW_BEFORE_BUILD="ccache -z"
 export CIBW_ENVIRONMENT_LINUX="CMAKE_PREFIX_PATH=\$PWD/deps/venv:\$PWD/deps/view CCACHE_DIR=/host${CCACHE_DIR} LD_LIBRARY_PATH=\$PWD/deps/view/lib64:\$PWD/deps/view/lib:\$PWD/deps/venv/lib64:\$PWD/deps/venv/lib"
 export CIBW_ENVIRONMENT_MACOS="CMAKE_PREFIX_PATH=\$PWD/deps/venv:\$PWD/deps/view CCACHE_DIR=${CCACHE_DIR} MACOSX_DEPLOYMENT_TARGET=26.0"
-export CIBW_BEFORE_TEST="ccache -s"
-export CIBW_TEST_COMMAND="python {package}/Examples/Scripts/Python/track_finding_python_only.py"
+export CIBW_BEFORE_TEST="ccache -s && pip install -r {package}/Python/Examples/tests/requirements.txt"
```
paulgessinger: `uv pip install` would save some time but might not be important.

CI/cibuildwheel.sh
```diff
@@ -27,8 +27,8 @@ export CIBW_ENVIRONMENT_PASS="CI"
...
+export CIBW_TEST_COMMAND="pytest {package}/Python/Examples/tests/test_arrow.py {package}/Python/Examples/tests/test_track_finding_python_only.py -m pypi -v"
```
paulgessinger: Why not just let it run the whole test discovery and select by `pypi`?

benjaminhuth: good point, this was predating the lazy-root imports...

paulgessinger (approved):

PR #5663, opened 2026-07-10
feat(detray): Add detector generate and read Python bindings for Detray

VladimirDrobny:
Adds Python bindings for Detray's toy detector generation and generic detector reading from JSON. Code structure aims to be faithful to ACTS.

In Detray, the top Python module is already occupied by the existing metadata code generator. To avoid refactoring, this PR places the Python bindings in submodules (detray.core and detray.examples).

Binds:
- generate_toy_detector in detray.examples: Builds the toy detector and writes it to a JSON file (equivalent of the generate_toy_detector CLI tool). It is bound only when DETRAY_BUILD_TEST_UTILS is enabled, which guarantees that the functions the binding calls are compiled.
- read_detector in detray.core: Reads a (for now only generic) detector from a JSON file.

Includes option to emit stubs, toggled by the new DETRAY_GENERATE_PYTHON_STUBS build option. Requires uv, for which ACTS supports fetching. This PR just makes it a requirement.

--- END COMMIT MESSAGE ---


Detray/python/CMakeLists.txt
```diff
@@ -34,3 +34,53 @@ install(
     FILES ${CMAKE_CURRENT_SOURCE_DIR}/detray/__init__.py
     DESTINATION ${DETRAY_PYTHON_INSTALL_DIR}
 )
+
+# Generate stubs for a user-facing Python module using pybind11-stubgen.
+# _module: dotted Python module name of the wrapper (e.g. detray.core)
+# _target: pybind11 module target; its .so is importable as detray.<_target>
+# [ARGN]:  additional CMake targets that must be built first
+function(detray_pybind_generate_stubs _module _target)
```
paulgessinger: Any way to reuse the ACTS level CMake code to include with a relative path one folder up from detray? That's what I did for the codegen setup recently.

Detray/python/Core/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+set_target_properties(
+    DetrayPythonBindings
+    PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${_python_dir}/detray
+)
+install(TARGETS DetrayPythonBindings DESTINATION ${DETRAY_PYTHON_INSTALL_DIR})
+
+# Public submodule wrapper(s): `core.py` exposes the bindings as `detray.core`.
+set(py_files core.py)
+
+foreach(f ${py_files})
```
paulgessinger: Mostly for @niermann999: this is some of the full duplication that I was thinking of when we discussed this. 

Now this is not a phenomenal amount of code of course, but @asalzburger put in some chunk of time to unify this across ACTS about 6mo ago.

Detray/python/Core/tests/test_read_detector.py
```diff
@@ -0,0 +1,20 @@
+import os
+import tempfile
+
+import detray.core
+import detray.examples
+
+
+def test_generate_and_read_detector():
```
paulgessinger, line 8: Again for @niermann999 (discussed with @VladimirDrobny earlier today): medium term I think it would make sense to have a combined python test suite across acts and detray.

Detray/python/Examples/python/__init__.py
paulgessinger, line 1: In acts, @asalzburger built it so that these are autogenerated.

paulgessinger (commented):
Thanks @VladimirDrobny!

Happy with the overall direction!

Detray/python/Core/src/CoreModuleEntry.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+namespace {
+
+using algebra_t = detray::array<DETRAY_CUSTOM_SCALARTYPE>;
+using detector_t = detray::detector<detray::default_metadata<algebra_t>>;
+using name_map_t = detector_t::name_map;
+
+/// Owns a detector together with the memory resource its data lives in.
+struct detector_handle {
+  std::unique_ptr<vecmem::host_memory_resource> memory_resource;
```
niermann999, line 46: We might just want to add this to the detector it self at this point. Then, we would not require these wrappers

Detray/python/CMakeLists.txt
VladimirDrobny: Done, although a little hacky, using acts_ prefix in Detray.

Detray/python/Examples/python/__init__.py
VladimirDrobny: Done.

Python/Examples/CMakeLists.txt
```diff
@@ -140,7 +140,11 @@ install(
     DESTINATION ${_examples_python_install_dir}
 )
 
-acts_pybind_generate_stubs(acts.examples ActsExamplesPythonBindings ActsPythonBindings)
+acts_pybind_generate_stubs(
```
niermann999, line 143: Are these changes intentional?

Detray/python/Core/src/CoreModuleEntry.cpp
niermann999: Ok, let's leave it as it is for the moment

Python/Examples/CMakeLists.txt
VladimirDrobny: Yes. For some reason, the gersemi lint CI job started failing on those lines all of a sudden. I just reformatted them as the job suggested, but maybe I'm missing something.

cmake/ActsPybindStubs.cmake
```diff
@@ -0,0 +1,33 @@
+include_guard(GLOBAL)
+
+if(NOT "${CMAKE_CURRENT_LIST_DIR}" IN_LIST CMAKE_MODULE_PATH)
+    list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})
+endif()
+
```
paulgessinger: I'm not sure I understand this. Does this need to be here or can we do that in the detray cmakelists?

Detray/python/Core/src/CoreModuleEntry.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+  py::class_<detector_handle>(m, "Detector")
+      .def(
+          "n_volumes",
+          [](const detector_handle &d) { return d.detector.volumes().size(); },
+          "Number of volumes in the detector")
+      .def(
+          "n_surfaces",
+          [](const detector_handle &d) { return d.detector.surfaces().size(); },
+          "Number of surfaces in the detector");
```
paulgessinger: Generally, I would advise to keep the python API as close to the C++ as reasonable, with some concessions to make it pythonic, e.g. using properties over getters and setters.

Detray/python/Core/src/CoreModuleEntry.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+  return {detector_handle{std::move(mr), std::move(det)}, std::move(names)};
+}
+
+}  // namespace
+
+PYBIND11_MODULE(DetrayPythonBindings, m) {
+  m.doc() = "Detray core bindings";
+
+  py::class_<detector_handle>(m, "Detector")
```
paulgessinger:
```suggestion
  py::class_<detector_handle>(m, "DetectorDefaultMetadata" STRINGIFY_HELPER(DETRAY_CUSTOM_SCALARTYPE))
```

with a suitably defined `STRINGIFY_HELPER` somewhere, so that the python name reads `DetectorDefaultMetadatafloat` or whatever it may be.

Detray/python/Core/src/CoreModuleEntry.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include <string>
+#include <utility>
+
+namespace py = pybind11;
+
+namespace {
+
+using algebra_t = detray::array<DETRAY_CUSTOM_SCALARTYPE>;
+using detector_t = detray::detector<detray::default_metadata<algebra_t>>;
+using name_map_t = detector_t::name_map;
```
paulgessinger: I don't think name mape depends on the detector type, does it? So could just be `detray::name_map`, and then you don't even need the alias.

Detray/python/Core/src/CoreModuleEntry.cpp
```diff
@@ -0,0 +1,78 @@
...
+  py::class_<detector_handle>(m, "Detector")
+      .def(
+          "n_volumes",
+          [](const detector_handle &d) { return d.detector.volumes().size(); },
+          "Number of volumes in the detector")
+      .def(
+          "n_surfaces",
+          [](const detector_handle &d) { return d.detector.surfaces().size(); },
+          "Number of surfaces in the detector");
+  py::class_<name_map_t>(m, "NameMap");
```
paulgessinger: Again, worth debating whether 

```suggestion
  py::class_<name_map_t>(m, "name_map");
```

is the right call here.

Detray/python/Core/src/CoreModuleEntry.cpp
VladimirDrobny: These are helper functions, just so I have something to test. They will be removed and replaced by `len(detector.volumes)` and `len(detector.surfaces)` in my next PR.

Detray/python/Core/src/CoreModuleEntry.cpp
VladimirDrobny: Why? It seems to me there's a convention of naming Python classes in CamelCase, and I think it would be weird to go against it.

Traccc/extras/benchmark/requirements.txt
stephenswat, line 1: I don't think you meant to include this; please delete it from the changeset.

Detray/codegen/detray-sympy/requirements.txt
stephenswat, line 1: Same here.

cmake/ActsPybindStubs.cmake
VladimirDrobny: Yes, we can do it in Detray CMakeLists. Fixed.

paulgessinger (dismissed):

niermann999 (dismissed):

paulgessinger (approved):

niermann999 (approved):

PR #5665, opened 2026-07-10
feat: Add track-state iteration to AnyTrackProxy

paulgessinger:
## What

The type-erased `AnyTrackProxy` previously only exposed `nTrackStates()`, with no way to actually reach a track's track states. This adds the same basic track-state access surface as the concrete `TrackProxy`, returning type-erased `AnyTrackStateProxy` objects:

- `outermostTrackState()` / `innermostTrackState()` (the latter returns `std::optional`, empty when the track is not forward-linked — matching `TrackProxy`)
- `trackStatesReversed()` (tip → stem) / `trackStates()` (stem → tip, requires forward-linking)

## How

- The handler vtable gains a single `trackStateAtIndex(container, stateIndex)` primitive (const → `AnyConstTrackStateProxy`, mutable → `AnyMutableTrackStateProxy`), implemented in both `TrackHandler` specializations via `container->trackStateContainer().getTrackState(...)`.
- A new `AnyTrackStateRange<reverse, read_only>` mirrors `detail_lt::TrackStateRange`, walking `previous()` / the `"next"` component and yielding type-erased proxies. Everything else is built on the already-available tip/stem indices.

## Incidental fix

`AnyTrackStateProxy`'s wrapping constructor takes its argument by non-const reference and the proxy satisfies `TrackStateProxyConcept`, so it greedily hijacked copy-construction of non-const `AnyTrackStateProxy` lvalues. This was latent until the range's `std::optional` storage exercised it. The constructor is now constrained to exclude `AnyTrackStateProxy` itself.

## Tests

Extended `AnyTrackProxyTests.cpp` with cases for reverse iteration order, forward iteration after `linkForward()`, the not-forward-linked and empty-track cases, and mutable write-through (verifying mutable ranges yield mutable proxies). `ActsUnitTestAnyTrackProxy` and `ActsUnitTestAnyTrackStateProxy` both pass.

### DIFF
--- a/Core/include/Acts/EventData/AnyTrackProxy.hpp
+++ b/Core/include/Acts/EventData/AnyTrackProxy.hpp
@@ -8,6 +8,7 @@
 
 #pragma once
 
+#include "Acts/EventData/AnyTrackStateProxy.hpp"
 #include "Acts/EventData/ParticleHypothesis.hpp"
 #include "Acts/EventData/TrackProxyCommon.hpp"
 #include "Acts/EventData/TrackProxyConcept.hpp"
@@ -16,6 +17,7 @@
 
 #include <any>
 #include <cassert>
+#include <optional>
 #include <type_traits>
 
 namespace Acts {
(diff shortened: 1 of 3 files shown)


paulgessinger:
Closing — will re-open from fork branch.

PR #5667, opened 2026-07-10
feat: Add track-state iteration to AnyTrackProxy

paulgessinger:
## What

The type-erased `AnyTrackProxy` previously only exposed `nTrackStates()`, with no way to actually reach a track's track states. This adds the same basic track-state access surface as the concrete `TrackProxy`, returning type-erased `AnyTrackStateProxy` objects:

- `outermostTrackState()` / `innermostTrackState()` (the latter returns `std::optional`, empty when the track is not forward-linked — matching `TrackProxy`)
- `trackStatesReversed()` (tip → stem) / `trackStates()` (stem → tip, requires forward-linking)

## How

- The handler vtable gains a single `trackStateAtIndex(container, stateIndex)` primitive (const → `AnyConstTrackStateProxy`, mutable → `AnyMutableTrackStateProxy`), implemented in both `TrackHandler` specializations via `container->trackStateContainer().getTrackState(...)`.
- A new `AnyTrackStateRange<reverse, read_only>` mirrors `detail_lt::TrackStateRange`, walking `previous()` / the `"next"` component and yielding type-erased proxies. Everything else is built on the already-available tip/stem indices.

## Incidental fix

`AnyTrackStateProxy`'s wrapping constructor takes its argument by non-const reference and the proxy satisfies `TrackStateProxyConcept`, so it greedily hijacked copy-construction of non-const `AnyTrackStateProxy` lvalues. This was latent until the range's `std::optional` storage exercised it. The constructor is now constrained to exclude `AnyTrackStateProxy` itself.

## Tests

Extended `AnyTrackProxyTests.cpp` with cases for reverse iteration order, forward iteration after `linkForward()`, the not-forward-linked and empty-track cases, and mutable write-through (verifying mutable ranges yield mutable proxies). `ActsUnitTestAnyTrackProxy` and `ActsUnitTestAnyTrackStateProxy` both pass.


Core/include/Acts/EventData/AnyTrackProxy.hpp
```diff
@@ -475,6 +510,101 @@ class AnyTrackProxy : public TrackProxyCommon<AnyTrackProxy<read_only>,
...
+    if (stem == kTrackIndexInvalid) {
+      return std::nullopt;
+    }
+    return mutableHandler()->trackStateAtIndex(mutableContainerPtr(), stem);
+  }
+
+  /// Get a range over the track states of this track, from the outside inwards.
+  /// Compatible with range-based for loops. Const version.
+  /// @return Track state range to iterate over
+  detail_anytstate::AnyTrackStateRange<true, true> trackStatesReversed() const {
```
andiwand: could use some typedefs
- `AnyConstReverseTrackStateRange`
- `AnyMutableReverseTrackStateRange`
- `AnyConstTrackStateRange`
- `AnyMutableTrackStateRange`

Core/include/Acts/EventData/AnyTrackStateProxy.hpp
```diff
@@ -809,6 +831,103 @@ class AnyTrackStateProxy
...
+  /// Construct an empty range.
+  AnyTrackStateRange() : m_begin{std::nullopt} {}
+
+  /// @return Iterator to the first track state.
+  Iterator begin() { return Iterator{m_begin}; }
+  /// @return Past-the-end iterator.
+  Iterator end() { return Iterator{std::nullopt}; }
+
+ private:
+  Iterator m_begin;
```
andiwand, line 926: should this also support a custom `end`? can be defaulted to end end

paulgessinger: The concrete (non type erased) track state iterator also doesn't do this, so I would lean towards keeping the end fixed.

Core/include/Acts/EventData/AnyTrackProxy.hpp
paulgessinger: Aliased, but they stay in the detail namespace.

andiwand (approved):

PR #5668, opened 2026-07-11
feat: displaced-track (LRT) support for GBTS seeding

jburzy:
The GBTS graph builder assumed beamline-pointing tracks in several places, making its LRT mode inefficient for large impact parameters. This PR removes or parametrizes those assumptions so that GBTS can seed tracks with |d0| up to a configured d0Max (300 mm for ITk LRT), and fixes several related bugs. All changes are should not affect  prompt
configurations. The widenings scale with d0Max (negligible at prompt values) or are gated on lrtMode, and new configuration fields default to the previous behaviour.

Bug fixes:
  - Fix the ptCoeff units error (0.5*B*UnitConstants::m with pT in GeV), which made the curvature cuts ~1000x too loose and silently disabled them in LRT mode and in the old-tunings path
  - Keep the entry-layer graph vetoes (isolated-node skip, z0-bitmask) out of LRT mode: they require a confirmed continuation two hits beyond the candidate doublet, which is only valid for minLevel >= 3; in LRT mode (minLevel 2) they rejected every three-space-point seed
  - Bounds-guard the z0 histogram fill, which wrote out of bounds when doubletFilterRZ was disabled

Displaced-aware cut widenings, evaluated as upper bounds at d0Max and radius-dependent so they vanish for r >> d0Max:
  - phi sliding window: widened by the position-azimuth swing asin(d0Max/r1) - asin(d0Max/r2); the phi wrap-around indexing margin scales accordingly
  - curvature proxy |dphi/dr|: widened by the same swing per unit radius, evaluated only for candidates in the marginal band
  - segment tau-ratio matching: widened by the ds/dr chord-factor ratio bound (cot(theta) cancels); each edge caches its chord factor
  - segment direction-phi matching: both estimates anchor at the shared node, so they differ by (c12 - c23)*r2 with c the per-segment asin(d0Max/r) chord slope; widened by that bound, band-gated

Configurability and generalization:
  - GbtsGeometry: the eta-bin pair connectivity z0 range (previously hardcoded to +-168 mm) is a constructor argument; a per-layer inward depth derived from the connection map replaces hardcoded layer IDs for the LRT matchBeforeCreate gate
  - Barrel/endcap decisions use the geometry layer type instead of ITk pixel layer-ID arithmetic, enabling triplet d0/pT validation and adaptive tau corrections for strip seeding
  - GbtsConnectionFilter enum (PixelOnly/StripOnly/All) replaces the boolean lrtMode volume filtering (bool overloads kept)
  - GbtsTrackingFilter: initial state covariances are configurable and an optional d0-driven r-z slope process noise accounts for the geometric dz/dr variation of displaced trajectories
  - z0 bitmask resolution and histogram tolerance made configurable
  - Per-cut debug accounting in the graph builder, chain statistics in the seed extraction, and rejection counters in the tracking filter, reported at DEBUG level with per-rejection kinematics at VERBOSE

  --- END COMMIT MESSAGE ---

cc @jpreston-cern

### DIFF
--- a/Core/include/Acts/Seeding/GbtsDataStorage.hpp
+++ b/Core/include/Acts/Seeding/GbtsDataStorage.hpp
@@ -89,6 +89,11 @@ struct GbtsEtaBin final {
 
   /// Layer ID for this bin
   std::uint32_t layerId{0};
+  /// Whether the bin belongs to a barrel layer
+  bool isBarrel{false};
+  /// Inward depth of the bin's layer in the connection graph
+  /// (0 = entry/innermost layer, see GbtsGeometry::layerDepthById)
+  std::uint32_t layerDepth{std::numeric_limits<std::uint32_t>::max()};
 };
 
 /// Storage container for GBTS nodes
@@ -118,6 +123,10 @@ class GbtsNodeStorage final {
   /// Get the total number of nodes
   /// @return Total number of nodes
   std::uint32_t numberOfNodes() const;
+  /// Get the smallest node radius over all non-empty bins.
+  /// Only valid after initializeNodes has been called.
+  /// @return Minimum node radius, or 0 if no nodes are loaded
+  float minNodeRadius() const;
   /// Sort nodes by phi
   void sortByPhi();
   /// Initialize node attributes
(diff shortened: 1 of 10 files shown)


jpreston-cern:
Thanks for doing this @jburzy!  can you validate that the prompt tracking still looks relatively the same (although i imagine it'll be a little different with the corrected `ptCoeff` ), as well as if running in `Examples` stand alone is affected.

jburzy:
Closing this initial attempt in favor of @jpreston-cern 's smarter approach :)

PR #5671, opened 2026-07-12
feat: Add PointSurface (PCA to a 3D point)

andiwand:

## Summary

Introduces a **`PointSurface`**, the point-analogue of the `LineSurface`: its intersection with a track is the **point of closest approach (PCA) to a single fixed 3D point** rather than to a line. The driving use case is **vertexing** — expressing track parameters at the PCA to a vertex / beam spot, just as a Perigee does to a line.

## Design

Geometrically a `PointSurface` is a **curvilinear plane anchored at the fixed center**: a plane through the center whose normal always equals the track momentum direction, so the track's crossing of that plane *is* the PCA. Key properties:

- **Local coordinates are Cartesian `(x, y)`** in the measurement plane perpendicular to the direction — not polar `(r, φ)`, which is degenerate at the point (`r = 0`).
- **Measurement frame** is built from the direction like `CurvilinearSurface`, with a global-X fall-back when the direction is (anti-)parallel to global Z (`s_curvilinearProjTolerance`).
- **No parallel degeneracy** — unlike `LineSurface`, the PCA is a plain projection `u = (c − p)·d`, always solvable.
- **Bounds**: unbounded (like Perigee), or a new `PointBounds` holding a single maximum distance `R` (Cartesian disc `x² + y² ≤ R²`).

## What's included

- `PointSurface` + `PointBounds` (`Core/…/Surfaces/`), `SurfaceType::Point` and `BoundsType::ePoint` (appended to avoid renumbering persisted values).
- Jacobians: `boundToFreeJacobian` **generalizes the LineSurface frame correction to both local axes** (the point's frame rotates fully with the direction, whereas a line's second axis is fixed); `freeToPathDerivative` keeps the direction block (nonzero residual at the PCA); alignment is **center-only** (a point has no orientation).
- Propagator treats a `Point` target like a `Perigee` (`ForcedSurfaceReached`), since the PCA can sit behind the current step.
- Integration: **JSON** (de)serialization, **EDM4hep/Podio**, and **Python** bindings (`Surface.createPoint`, `SurfaceType.Point`, `SurfaceBoundsType.Point`).

## Testing

- `PointSurfaceTests` / `PointBoundsTests`: construction, round-trip global↔local (incl. large-η and direction ∥ Z), PCA intersection (residual ⊥ direction, no degeneracy), bounds rejection.
- **Finite-difference validation** of `boundToFreeJacobian`, `freeToPathDerivative`, and `alignmentToPathDerivative` — this validates the jacobian derivation, the main technical risk.
- JSON round-trip tests (bounded + unbounded); Python test parity with Perigee.

All new and adjacent surface unit tests pass locally.

## Notes / open items

- The analytic `boundToFreeJacobian` correction is exact in the standard (Z-based) frame branch; near direction ∥ Z the X-based fall-back makes it approximate (analogous to the documented forward-η limitation of `LineSurface`). Round-trip behavior is exact in both branches.
- Detector-geometry converters (Geant4, GeoModel, Detray, ActSVG, Root TGeo) are intentionally **not** wired up — a point has no analogue there and their existing default/throw branches are simply never exercised for this type.


paulgessinger:
If we make the jacobians surface specific we could use this to introduce a telescope safe parametrization as well

asalzburger:
> If we make the jacobians surface specific we could use this to introduce a telescope safe parametrization as well

In addition, this could be our adHoc parameterization for TPC, where space charge distortions make any extended surface a bit of a pain.

Core/include/Acts/Surfaces/Surface.hpp
```diff
@@ -70,13 +70,14 @@ class Surface : public virtual GeometryObject,
...
+    Point = 8
   };
 
   /// Helper strings for screen output
-  static constexpr std::array<std::string_view, Surface::SurfaceType::Other + 1>
-      s_surfaceTypeNames = {"Cone",  "Cylinder", "Disc",        "Perigee",
-                            "Plane", "Straw",    "Curvilinear", "Other"};
+  static constexpr std::array<std::string_view, Surface::SurfaceType::Point + 1>
+      s_surfaceTypeNames = {"Cone",  "Cylinder",    "Disc",  "Perigee", "Plane",
+                            "Straw", "Curvilinear", "Other", "Point"};
```
asalzburger: Is there a case that prevents us from leaving "Other" at the end ?
Usually that's the left-over category.

Core/include/Acts/Surfaces/SurfaceBounds.hpp
```diff
@@ -44,7 +44,8 @@ class SurfaceBounds {
     eConvexPolygon = 10,
     eAnnulus = 11,
     eBoundless = 12,
-    eOther = 13
+    eOther = 13,
+    ePoint = 14
```
asalzburger: Same here, the idea with eOther is that we can just push it to the back count, however, if someone had written this out (JSON) we introduce a schema evolution, I guess that's what you want to avoid.

Core/src/Surfaces/PointSurface.cpp
```diff
@@ -0,0 +1,295 @@
...
+Polyhedron PointSurface::polyhedronRepresentation(
+    const GeometryContext& gctx, unsigned int /*quarterSegments*/) const {
+  // Non-physical marker: a small cross centered on the point in the transform
+  // frame (a point has no extent, this is for display/debugging only).
+  std::vector<Vector3> vertices;
+  std::vector<Polyhedron::FaceType> faces;
+  std::vector<Polyhedron::FaceType> triangularMesh;
+
+  const Transform3& ctransform = localToGlobalTransform(gctx);
+  double d = 1.;
```
asalzburger: Indeed, that's a bit challenging to display.

To remove a fixed `d=1` we could enhance the polyhedronRep (quaterSegments) into a PolyHedronOptions that allows to set segment number to mimic an arc, reference length, etc.

asalzburger (changes requested):
A quick check on the enum / string ordering - otherwise I find this very good.

asalzburger (approved):
GTG now, the re-ordering made me particularly happy.

PR #5672, opened 2026-07-12
feat: Use PointSurface for the 3D PCA in ImpactPointEstimator

andiwand:
`ImpactPointEstimator::estimate3DImpactParameters` now propagates directly to a `PointSurface` at the vertex. Its normal follows the local track momentum, so the propagator converges exactly to the 3D PCA. This replaces the analytic PCA estimate plus propagation to a fixed-normal plane, which was only exact when the vertex-to-PCA vector was orthogonal to the momentum.

`getVertexCompatibilityImpl` takes the measurement frame from `referenceFrame(gctx, pos, dir)` instead of the transform rotation, since the `PointSurface` frame is direction-dependent. Behavior is unchanged for direction-independent surfaces.

--- END COMMIT MESSAGE ---

Depends on #5671.

The `PerigeeSurface` uses producing the 2D parameters `(d0, z0)` are untouched — linearizers, Billoir residuals, lifetime signs and density finders all need the Perigee parametrization, which the Cartesian `PointSurface` does not provide.

All vertexing unit tests pass. One assertion is relaxed for transverse tracks (θ = 90°), where 2D and 3D PCA coincide and the local position is legitimately unchanged; the old plane differed only through its arbitrary in-plane frame choice.

### DIFF
--- a/Core/src/Vertexing/ImpactPointEstimator.cpp
+++ b/Core/src/Vertexing/ImpactPointEstimator.cpp
@@ -11,7 +11,7 @@
 #include "Acts/Definitions/Algebra.hpp"
 #include "Acts/Propagator/PropagatorOptions.hpp"
 #include "Acts/Surfaces/PerigeeSurface.hpp"
-#include "Acts/Surfaces/PlaneSurface.hpp"
+#include "Acts/Surfaces/PointSurface.hpp"
 #include "Acts/Utilities/AngleHelpers.hpp"
 #include "Acts/Utilities/Intersection.hpp"
 #include "Acts/Utilities/MathHelpers.hpp"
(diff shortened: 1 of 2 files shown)


benjaminhuth (approved):
LGTM. But it seems we have no hash-check for vertexing algorithms?

andiwand:
_This comment was written by Claude, driven by @andiwand._

Local A/B of this diff against its base. Relevant because Athena's default primary-vertex finder is `Acts::AdaptiveMultiVertexFinder`, so `estimate3DImpactParameters` sits on a production path.

**Physics — no change, but not bit-identical.** Over 5000 track/vertex pairs: median |Δχ²|/χ² 9e-9 (const 2 T) / 7e-8 (solenoid), max 1.5e-4; the 3D PCA moves by pm–nm; zero flips of a χ² < 18.42 accept/reject. End-to-end AMVF (Pythia8 1+50 PU, 20 events): identical vertex count in every event, 155/159 vertices bitwise identical, max position shift 2.3e-7 mm. Physics references should hold; digest/byte-level ones will move.

**Compute — a wash.** The isolated `estimate3DImpactParameters` + `getVertexCompatibility` call is −2.6 % (const) / −2.55 % (solenoid) against a 0.4–1.1 % noise floor, from dropping the Newton solve. At AMVF-algorithm level that dilutes to −1.4 % against a −1.2 % noise floor, i.e. not measurable.

PR #5674, opened 2026-07-13
fix: MuonHoughSeeder fix to min-max logic and unit test

pingponghero12:
Commit fixes logic for min max axis range calculation in MuonHoughSeeder and adds a unit test based on one from HoughTransformUtilsTests, which runs one bucket and compares the result to provided parameters.

--- END COMMIT MESSAGE ---

MuonHoughSeeder is an example class that did not have unit tests before, the class before fix did not return any maximums for the valid unit test, after fix it returns expected maximum.

The test involves one simple case taken from HoughTransformUtilsTests.cpp.

@junggjo9 @lmonaco99


Examples/Algorithms/TrackFinding/src/MuonHoughSeeder.cpp
```diff
@@ -113,8 +113,8 @@ MuonHoughMaxContainer MuonHoughSeeder::constructEtaMaxima(
     const AlgorithmContext& ctx, const MuonSpacePointBucket& bucket,
     HoughPlane_t& plane) const {
   MuonHoughMaxContainer etaMaxima{};
-  AxisRange_t axisRanges{-3., 3., 100. * Acts::UnitConstants::m,
-                         -100. * Acts::UnitConstants::m};
+  AxisRange_t axisRanges{-3., 3., -100. * Acts::UnitConstants::m,
+                         100. * Acts::UnitConstants::m};
```
junggjo9: Isn't the constructor {minX, maxX, minY, maxY};?

pingponghero12: Yes, I made slight mistake in the code, I'm fixing it right now, there was no need to change the signs in the axis initialization, but just change min min to max min.

pingponghero12:
@junggjo9 the logic should be sound right now.

For context, the original was implemented as:
```cpp
    AxisRange_t axisRanges{-3., 3., 100. * Acts::UnitConstants::m,
                           -100. * Acts::UnitConstants::m};
```
Then for all points it would check:
```cpp
      axisRanges.yMax =
          std::min(axisRanges.yMax, x + m_cfg.phiPlaneMarginIcept);
      axisRanges.yMin =
          std::min(axisRanges.yMin, x - m_cfg.phiPlaneMarginIcept);
```

The mistake was that it would for both yMax and yMin take min instead of max and min accordingly. Example of behavior of original by hand:
```
Original:
yMin = 100 000
yMax = -100 000

example points:
DriftCircle{-427.981, -225.541, 14.5202, uncert},
DriftCircle{-412.964, -199.530, 1.66237, uncert},

// I dont account for left-right
for -> 0
    yMax = min(-427, -100 000) = -100 000
    yMin = min(-427, 100 000) = -427

for -> 0
    yMax = min(-412, -100 000) = -100 000
    yMin = min(-412, -427) = -412
```

junggjo9 (approved):

junggjo9 (approved):

benjaminhuth:
`automerge` won't work as CI bridge won't run automatically. need to merge manually

benjaminhuth:
works now

PR #5676, opened 2026-07-13
feat!: extend EDM4hep plugin track conversion for drop-in converter parity

madbaron:
Make ActsPlugins::EDM4hepUtil a drop-in replacement for the ACTS->EDM4hep track converters maintained in k4ActsTracking, and fix the track state covariance layout while doing so.

writeTrack:
- Take an Acts::MagneticFieldProvider and MagneticFieldContext instead of a scalar Bz, and evaluate the local field at each track state, so the perigee/omega conversion is correct in a non-uniform field. The scalar-Bz signature is kept as a convenience overload backed by a ConstantBField.
- Write the number of holes (setNholes) alongside chi2/ndf.
- Add an optional hitLookup callback (TrackerHitLookup, a std::function over AnyConstTrackStateProxy) returning the edm4hep::TrackerHit for a measurement state, or nullopt. Returned hits are attached via addToTrackerHits. 
- Emit track states in the EDM4hep/LCIO convention: the AtIP state first, then the measurement states inside-out (AtFirstHit ... AtLastHit).
- Guard the AtFirstHit/AtLastHit tagging against tracks with no measurement states (previously an unchecked front()/back()).

readTrack:
- Take a magnetic field provider as well, so a non-uniform field can be read back, not just written. The scalar-Bz signature is likewise kept as a convenience overload.
- Sample the field at the track state's point of closest approach, which is where writeTrackState samples it.
- Iterate forward and skip the IP state, matching the new write order.
- Read back the number of holes.

New writeTrackState():
- Convert a single Acts::BoundTrackParameters into an edm4hep::TrackState. writeTrack is reimplemented on top of it.
- Exposed because downstream needs track states that do not come from a fitted track: k4ActsTracking builds AtFirstHit states for seeds and AtCalorimeter states after extrapolation, and was reimplementing this conversion privately.

Fix the packed covariance parameter order:
- edm4hep::TrackParams is {d0, phi, omega, z0, tanLambda, time}, but the covariance was packed in the internal order {d0, z0, phi, tanLambda, omega, time}, so 16 of 21 elements were stored in the wrong slot.
- packCovariance/unpackCovariance now take an edm4hep::CovMatrix6f and delegate element placement to its setValue/getValue, so the layout is EDM4hep's to define rather than ours.

Also:
- Move the readTrack helper lambdas and the field lookup into detail:: helpers  in the .cpp (localBz, referencePoint, unpackTrackState, pointOfClosestApproach).
- Drop the EDM4hep < 0.99 compatibility branch for getParticle/setParticle, which is dead now that the minimum version is 0.99.

Tests:
- CovariancePackingConvention checks all 36 covariance elements against EDM4hep's own accessor, which a pack/unpack round trip cannot do.
- HitLookupInvocationOrder pins the callback invocation order against the emitted state order.
- NonUniformFieldRoundTrip round-trips a track with large d0/z0 through a z-dependent field.
- RoundTripTests additionally checks chi2, ndf and nHoles.

BREAKING CHANGE: the EDM4hep track conversion changes its API, its on-disk format and its minimum EDM4hep version.

Source:
- writeTrack takes an optional hitLookup callback before the logger, so calls passing a logger positionally now need an explicit `{}` for the callback.
- readTrack takes a GeometryContext as its first argument, needed to locate the point of closest approach at which the local field is sampled.

On-disk:
- Track states are written IP-first, then the measurement states inside-out (AtFirstHit ... AtLastHit); previously outside-in with the IP state last.
- The track state covariance is packed in the EDM4hep parameter order (edm4hep::TrackParams) rather than the internal order. Files written by earlier versions are read differently by this version, and vice versa.

Build:
- The minimum supported EDM4hep version is raised from 0.10.5 to 0.99, the first release providing the TrackerHit interface, the CovMatrix types, Track::Nholes and the edm4hep::TrackParams enum.

--- END COMMIT MESSAGE ---

Follow-up targeting the discussion in key4hep/k4ActsTracking#44. The companion change is key4hep/k4ActsTracking#95, which deletes the local converter and calls into this one.

Follow-up points:
- The stale ordering promise in the hitLookup docs was a real mis-association, not just a doc bug: the callback ran outside-in while the buffers were reversed before emission. It now runs in emission order.
- Marking this BREAKING rather than preserving compatibility: the positional logger break alone could be absorbed with an extra writeTrack(gctx, track, to, Bz, const Logger&) overload, but that covers one of five breaks, and the readTrack context argument, both on-disk changes and the EDM4hep floor cannot be made compatible, so the label is needed either way.


Plugins/EDM4hep/include/ActsPlugins/EDM4hep/EDM4hepUtil.hpp
```diff
@@ -106,22 +119,58 @@ edm4hep::MCParticle getParticle(const edm4hep::SimTrackerHit& hit);
...
-/// Write an Acts track to EDM4hep format
+/// Callable returning the local magnetic field z-component (in Acts native
+/// units) at a given global position.
+///
+/// This is used by @ref writeTrack to support spatially varying magnetic
+/// fields: the LCIO/EDM4hep perigee parametrization (in particular the
+/// curvature @c omega) depends on the local field, so for a non-uniform field
+/// the conversion should use the field value at each track state's location
+/// rather than a single global constant.
+using LocalBzProvider = std::function<double(const Acts::Vector3& position)>;
```
paulgessinger: Could we use the existing `MagneticFieldProvider` abstraction for this? That's already created from dd4hp in k4ActsTracking so would be easy to pass in here.

Plugins/EDM4hep/include/ActsPlugins/EDM4hep/EDM4hepUtil.hpp
```diff
@@ -88,6 +91,16 @@ Parameters convertTrackParametersToEdm4hep(
 Acts::BoundTrackParameters convertTrackParametersFromEdm4hep(
     double Bz, const Parameters& params);
 
+/// Default tracker-hit lookup used by @ref writeTrack: never associates a hit,
+/// reproducing the historical behaviour where no tracker hits are written.
+struct NoTrackerHitLookup {
+  template <typename state_proxy_t>
```
paulgessinger: This is templated I think only because the state varies. At this time, we have `AnyTrackStateProxy` which erases the concrete type. With that, you would be able to turn this into a simple function, and the argument to `writeTrack` into an `std::function`

Plugins/EDM4hep/include/ActsPlugins/EDM4hep/EDM4hepUtil.hpp
```diff
@@ -106,22 +119,58 @@ edm4hep::MCParticle getParticle(const edm4hep::SimTrackerHit& hit);
...
-template <Acts::TrackProxyConcept track_proxy_t>
+/// @param hitLookup Optional callback mapping a measurement track state to its
+///                  EDM4hep tracker hit (defaults to writing no tracker hits)
+template <Acts::TrackProxyConcept track_proxy_t,
+          typename hit_lookup_t = detail::NoTrackerHitLookup>
 void writeTrack(const Acts::GeometryContext& gctx, track_proxy_t track,
-                edm4hep::MutableTrack to, double Bz,
-                const Acts::Logger& logger = Acts::getDummyLogger()) {
+                edm4hep::MutableTrack to, const LocalBzProvider& bzAtPosition,
+                const Acts::Logger& logger = Acts::getDummyLogger(),
```
paulgessinger: I would probably keep the logger argument last.

paulgessinger (commented):
Thanks @madbaron. Not that many changes needed for this to be usable, great! I only have a couple of detail comments.

Plugins/EDM4hep/include/ActsPlugins/EDM4hep/EDM4hepUtil.hpp
madbaron: Done!

Plugins/EDM4hep/include/ActsPlugins/EDM4hep/EDM4hepUtil.hpp
madbaron: `LocalBzProvider` is gone entirely: both writeTrack and now readTrack take `const Acts::MagneticFieldProvider&`.

Plugins/EDM4hep/include/ActsPlugins/EDM4hep/EDM4hepUtil.hpp
madbaron: `NoTrackerHitLookup` and the `hit_lookup_t` template parameter are both gone.

madbaron:
Thanks for the review @paulgessinger. Sorry it took me a while to react, I wanted to test this in tandem with the implementation in k4ActsTracking downstream and only got to it now. All three comments are addressed, plus two follow-ups that came out of testing this against k4ActsTracking.

`readTrack` now takes a field provider too, mirroring `writeTrack`. Previously you could write correctly in a non-uniform field but not read back.
New `writeTrackState(...)` converting a single BoundTrackParameters into an `edm4hep::TrackState`. `writeTrack` is reimplemented on top of it, which removed the `setParameters` lambda and the `referencePoint` block that was  duplicated between the measurement loop and the IP state. It's public because downstream needs states that don't come from a fitted track. k4ActsTracking builds `AtFirstHit` states for seeds and `AtCalorimeter` states after extrapolation, and was reimplementing this conversion privately ([key4hep/k4ActsTracking#44](https://github.com/key4hep/k4ActsTracking/issues/44)).

Two behaviour changes:
- Covariance ordering fix. The packed covariance did not follow EDM4hep's parameter order. `edm4hep::TrackParams` is {d0, phi, omega, z0, tanLambda, time}; the plugin packed in its internal order {d0, z0, phi, tanLambda, omega, time}, so 16 of 21 slots were wrong. It round-tripped cleanly because `unpackCovariance` had the same permutation, which is why the existing `CovariancePacking` test never caught it, but any non-ACTS consumer read scrambled covariances. pack/unpackCovariance now take `edm4hep::CovMatrix6f` and delegate placement to `setValue/getValue`, so the layout is EDM4hep's to define, not ours. Added CovariancePackingConvention which checks all 36 elements against EDM4hep's own accessor (verified it fails against the old packing).
- Track state ordering. States are now emitted IP-first, then inside-out (AtIP, AtFirstHit … AtLastHit), matching the LCIO/EDM4hep convention that downstream converters use. readTrack moves in lockstep.

Both change on-disk content. ACTS→ACTS round-trips are consistent either way, but files written by released ACTS read differently under this branch and vice versa.

paulgessinger:
@madbaron sounds reasonable, I'll take another look

Plugins/EDM4hep/include/ActsPlugins/EDM4hep/EDM4hepUtil.hpp
```diff
@@ -106,154 +130,258 @@ edm4hep::MCParticle getParticle(const edm4hep::SimTrackerHit& hit);
...
-  auto unpack = [](const edm4hep::TrackState& trackState) {
+  // Global position the track state is expressed relative to. This is also
+  // where the local field is evaluated.
+  auto referencePoint = [](const edm4hep::TrackState& trackState) {
+    return Vector3{
+        trackState.referencePoint.x,
+        trackState.referencePoint.y,
+        trackState.referencePoint.z,
+    };
+  };
```
paulgessinger: Could probably go to a detail helper as well?

Plugins/EDM4hep/include/ActsPlugins/EDM4hep/EDM4hepUtil.hpp
```diff
@@ -106,154 +130,258 @@ edm4hep::MCParticle getParticle(const edm4hep::SimTrackerHit& hit);
...
+/// @param gctx The geometry context
+/// @param mctx The magnetic field context
 /// @param track The Acts track to convert
 /// @param to The EDM4hep track to write to
-/// @param Bz The magnetic field z-component
+/// @param magneticField The magnetic field provider, evaluated at each state
+/// @param hitLookup Optional callback mapping a measurement track state to its
+///                  EDM4hep tracker hit (defaults to writing no tracker hits)
 /// @param logger The logger instance
 template <Acts::TrackProxyConcept track_proxy_t>
```
paulgessinger, line 232: Since the `TrackerHitLookup` now already uses the type-erased track state access, this function here could likely also use the type-erased track proxy type and become regular function, which could be pushed fully into the `.cpp` file.

I think I'll look at this as a follow up item.

paulgessinger (dismissed):
nice

paulgessinger:
Some more follow up points:

- EDM4hepUtil.cpp:248 samples Bz at params.position(gctx) while readTrack (EDM4hepUtil.hpp:398,415) samples it at the reference point; for parameters already on a perigee (the IP state, normally) those differ by d0/z0, so in a real field map q/p comes back biased and the header's round-trip claim at EDM4hepUtil.hpp:327 is false.
- This RAISES the edm4hep floor supported by ACTS to v0.99 I believe. This should be recorded in the CMake.
- inserting hitLookup before logger in writeTrack (EDM4hepUtil.hpp:211) is source-breaking for any caller passing a logger positionally (the in-repo test had to add {}), and the PR is titled plain feat: despite that plus the on-disk covariance and ordering changes. We either mark this as BREAKING or we figure out a backwards compatible way
- EDM4hepUtil.hpp:205 promises the hitLookup callback is invoked "in the same order the states are written", but it runs outside-in at line 248 and the buffers are reversed afterwards at line 284, so a stateful callback mis-associates.
- writeTrack sets Nholes (EDM4hepUtil.hpp:226) but readTrack never reads it back (EDM4hepUtil.hpp:425-427)

Plugins/EDM4hep/include/ActsPlugins/EDM4hep/EDM4hepUtil.hpp
madbaron: Done!

Plugins/EDM4hep/include/ActsPlugins/EDM4hep/EDM4hepUtil.hpp
madbaron: Just to say it explicitly: I left this as a follow-up

madbaron:
> Some more follow up points:
> 
> * EDM4hepUtil.cpp:248 samples Bz at params.position(gctx) while readTrack (EDM4hepUtil.hpp:398,415) samples it at the reference point; for parameters already on a perigee (the IP state, normally) those differ by d0/z0, so in a real field map q/p comes back biased and the header's round-trip claim at EDM4hepUtil.hpp:327 is false.
> * This RAISES the edm4hep floor supported by ACTS to v0.99 I believe. This should be recorded in the CMake.
> * inserting hitLookup before logger in writeTrack (EDM4hepUtil.hpp:211) is source-breaking for any caller passing a logger positionally (the in-repo test had to add {}), and the PR is titled plain feat: despite that plus the on-disk covariance and ordering changes. We either mark this as BREAKING or we figure out a backwards compatible way
> * EDM4hepUtil.hpp:205 promises the hitLookup callback is invoked "in the same order the states are written", but it runs outside-in at line 248 and the buffers are reversed afterwards at line 284, so a stateful callback mis-associates.
> * writeTrack sets Nholes (EDM4hepUtil.hpp:226) but readTrack never reads it back (EDM4hepUtil.hpp:425-427)

All done

paulgessinger (dismissed):

CMakeLists.txt
```diff
@@ -593,14 +595,7 @@ if(ACTS_BUILD_PLUGIN_ONNX OR ACTS_GNN_ENABLE_ONNX)
     find_package(onnxruntime ${_acts_onnxruntime_version} MODULE REQUIRED)
 endif()
 if(ACTS_BUILD_PLUGIN_EDM4HEP)
-    find_package(EDM4HEP ${_acts_edm4hep_version} CONFIG)
-    if(NOT EDM4HEP_FOUND)
-        message(
```
paulgessinger, line 598: I think we'll still want to support 0.99 and 1.0 at the same time, something like this fallback version mechanism would still be needed.

madbaron: I've put it back!

paulgessinger (approved):

PR #5678, opened 2026-07-13
ci: Drop redundant 2d/3d projections in physmon

benjaminhuth:


benjaminhuth:
Ultra low stats comparison (1 PR) gives roughly 54 -> 42 minutes on github CI, 68 -> 56 min on gitlab CI.

benjaminhuth:
For context: this collapses these 6 plots to one (only pull_z0): 


image

andiwand (approved):

andiwand:
there seem to be a couple more cases @benjaminhuth 
image

benjaminhuth:
@andiwand  Hmm but those we cannot tackle with the filter, so we would decide if we want to remove `pullmean_vs_phi` from the set

andiwand:
ok in this case it would be good to do it upstream in `histcmp` I think

paulgessinger:
@andiwand Right but the job is to compare the histograms, not ignore them by default, right?

andiwand:
> @andiwand Right but the job is to compare the histograms, not ignore them by default, right?

yes that works well for 1D plots but poorly for 2D and 3D. so my argument is why do it in the first place

paulgessinger:
Not actually faster, consensus that the projections are redundant and can be dropped.

PR #5679, opened 2026-07-13
feat: add spherical (phi, eta, r) grid seeding algorithm

Aksth070600:
# feat: add Spherical (φ, η, r) grid seeding algorithm

## Summary

Adds a new triplet seeding algorithm that bins space points on a **spherical
(φ, η, r)** grid instead of the cylindrical (φ, z, r) grid. Added `SphericalSpacePointGrid` alongside the existing `CylindricalSpacePointGrid` and `CartesianSpacePointGrid`. The existing seeders are untouched, and all downstream doublet/triplet/filter code is reused unchanged.

It is provided so the ITk community can evaluate the binning directly in Athena. This grew out of my master's thesis on ACTS performance optimisation
(*Performance Optimization in ACTS*, [doi.org/10.5281/zenodo.20717572](https://doi.org/10.5281/zenodo.20717572)).

## Motivation

Binning on `cot(θ) = z/r = sinh(η)` (monotone in η) groups near-collinear space
points, aligning the grid with the collision geometry rather than the detector's
z-extent:

- Uniform Δz bins are non-uniform in η; particle density is more uniform in (φ, η).
- Detector acceptance `|η| < η_max` gives a fixed, compact binning range,
  instead of the large/unbounded z-range the forward region needs.
- Contiguous η-bin arrays → simpler neighbour queries.

## Design

- **New, cleanly-named classes:** `Acts::SphericalSpacePointGrid` (Core) and
  `ActsExamples::SphericalGridTripletSeedingAlgorithm` (Examples), mirroring the
  pattern ACTS already uses for Cartesian/Cylindrical. The existing grid and
  algorithm classes are not modified.
- **No EDM change:** The grid computes `cot(θ) = z/r` at fill from the existing
  `zr()` accessor and stores `sinh(etaBinEdges)` on the axis, avoiding any new
  `SpacePointContainer` columns.
- **Downstream reused unchanged:** `DoubletSeedFinder`, `TripletSeedFinder`,
  `TripletSeeder`, `BroadTripletSeedFilter` are untouched; only the grid fill and
  neighbour gather differ.
- **Added Example Algorithm:** Fully wired into the Python layer (`SeedingAlgorithm.SphericalGridTriplet`,
  `addSphericalGridTripletSeeding`) with an example ITk operating point that
  leaves all generic selection cuts at the `itkSeedingAlgConfig` defaults and
  overrides only the (η) binning parameters.

## Performance

- **ODD (thesis, ACTS v44):** the spherical operating points showed substantial
  pipeline speedups and preserved/improved efficiency vs the ODD cylindrical
  default. Full study + plots in the thesis (DOI above). *(Note: the cylindrical default was not optimized for the ODD geometry; the thesis reported seeding-only results.)*
- **ITk (current main, ttbar+PU200, 200 ev, matched non-grid parameters):** with
  every non-grid-specific parameter held at the cylindrical ITk-tuned defaults, the
  **binning itself is roughly efficiency-neutral but a few percent slower in timing**
  (≈ −0.002 in ambiguity efficiency; few % slower seeding, few % slower total pipeline). Larger ITk speedups seen in earlier exploration were mostly attributable to operating-point choices that also benefit the cylindrical grid, not to the binning per se. Some early tests also compared against an older, less-optimised cylindrical implementation, which further contributed to the promising early results.

## Further work

- **Optimization and testing in Athena:** Even though the spherical implementation seems to be roughly equal to the cylindrical version, a version tuned for Athena might give better results. Some operating points seem to be faster on the spherical grid than the cylindrical one, though not at the default ITk parameters this PR ships with.

### DIFF
--- a/Core/include/Acts/Seeding/SphericalSpacePointGrid.hpp
+++ b/Core/include/Acts/Seeding/SphericalSpacePointGrid.hpp
@@ -1,0 +1,208 @@
+// This file is part of the ACTS project.
+//
+// Copyright (C) 2016 CERN for the benefit of the ACTS project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+#include "Acts/Definitions/Units.hpp"
+#include "Acts/EventData/SpacePointContainer.hpp"
+#include "Acts/Seeding/BinnedGroup.hpp"
+#include "Acts/Utilities/Grid.hpp"
+#include "Acts/Utilities/Logger.hpp"
+#include "Acts/Utilities/RangeXD.hpp"
+
+#include <numbers>
+#include <vector>
+
+namespace Acts {
+
+/// A spherical space point grid used for seeding in a cylindrical detector
+/// geometry.
+/// The grid is defined in spherical coordinates (phi, eta, r)
+class SphericalSpacePointGrid {
+ public:
+  /// Space point index type used in the grid.
+  using SpacePointIndex = std::uint32_t;
+  /// Type alias for bin container holding space point indices
+  using BinType = std::vector<SpacePointIndex>;
+  /// Type alias for phi axis with equidistant binning and closed boundaries
+  using PhiAxisType = Axis<AxisType::Equidistant, AxisBoundaryType::Closed>;
+  /// Type alias for the eta axis with variable binning and open boundaries. The
+  /// stored edges are cot(theta) = sinh(eta), so space points bin by z / r.
+  using EtaAxisType = Axis<AxisType::Variable, AxisBoundaryType::Open>;
+  /// Type alias for r axis with variable binning and open boundaries
+  using RAxisType = Axis<AxisType::Variable, AxisBoundaryType::Open>;
+  /// Spherical space point grid type (phi, eta, r)
+  using GridType = Grid<BinType, PhiAxisType, EtaAxisType, RAxisType>;
+  /// Type alias for binned group over the spherical grid
+  using BinnedGroupType = BinnedGroup<GridType>;
+
+  /// Configuration parameters for the spherical space point grid.
+  struct Config {
+    /// minimum pT
+    float minPt = 0 * UnitConstants::MeV;
+    /// minimum extension of sensitive detector layer relevant for seeding as
+    /// distance from x=y=0 (i.e. in r)
+    float rMin = 0 * UnitConstants::mm;
+    /// maximum extension of sensitive detector layer relevant for seeding as
+    /// distance from x=y=0 (i.e. in r)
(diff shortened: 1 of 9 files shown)


Core/include/Acts/Seeding/SphericalSpacePointGrid.hpp
```diff
@@ -0,0 +1,208 @@
...
+  ///         `std::nullopt` if the space point is outside the grid bounds.
+  std::optional<std::size_t> binIndex(const Vector3& position) const {
+    if (!grid().isInside(position)) {
+      return std::nullopt;
+    }
+    return grid().globalBinFromPosition(position);
+  }
+  /// Get the bin index for a space point.
+  /// @param phi The azimuthal angle of the space point in radians
+  /// @param cotTheta The polar coordinate cot(theta) = z / r of the space point
```
andiwand, line 124: why not use `eta` as a grid variable here?

Examples/Algorithms/TrackFinding/include/ActsExamples/TrackFinding/SphericalGridTripletSeedingAlgorithm.hpp
andiwand, line 1: we should try to deduplicate this with the existing seeder. there should be a way to branch into one or another grid based on configuration. potentially it will need some template unrolling to deal with different types

andiwand (commented):
thanks @Aksth070600 !

I wonder if we should first concentrate to get the grid in and then see how to integrate this in the Examples.

alternatively we could follow the current approach and then do a followup to recombine them. but this would have to happen soon to avoid divergence

Aksth070600:
Thanks @andiwand! I'd prefer to keep this as one PR and do the deduplication here rather than defer it, so there's no divergence to reconcile. I'd template the existing GridTripletSeedingAlgorithm over the grid type and select cylindrical vs spherical by config, keeping the cylindrical path unchanged. Does that work for you, or would you rather land the grid first and integrate in a follow-up?

Core/include/Acts/Seeding/SphericalSpacePointGrid.hpp
Aksth070600: This was a small perf change: by binning on cotθ = z/r = sinh(eta) and storing sinh(etaBinEdges) on the axis, the fill is a single division per space point instead of computing eta = asinh(z/r). It should be marginally cheaper to fill, but if readability matters more, i can switch to using eta directly.

andiwand:
@Aksth070600 I think landing the spherical grid first would be nice because it should go smoothly. For the seeder implementation we need a few more cycles to iron it out and try to remove the duplication. The benefit is that part of your work lands almost immediately and then we iterate on the rest.

If you agree you can create another PR that just targets the spherical grid. I would put it into the `Acts::Experimental` namespace for now so we do not have to mess with versioning in future updates. If you want you could also add some unit tests so this is exercised even without the seeding code.

PR #5680, opened 2026-07-13
perf: defer error2 behind the deltaCotTheta2 pre-filter in the triplet cut

Aksth070600:
# perf: defer error² behind the cheap deltaCotTheta² pre-filter in the triplet cut

## Summary

In `BroadTripletSeedFinder`'s triplet cot-θ compatibility cuts (both the pixel and
strip loops), the expensive `error²` term is computed for every candidate before it
is used. This defers it behind the cheap `deltaCotTheta²` condition, so it
is computed at most once and only when it can actually matter. The change is
bit-exact, the produced seeds are identical.

## What / why

Both cot-θ cuts have the form

```
deltaCotTheta² > error² + <scatter>      (error² ≥ 0)
```

so a cut can only fire when `deltaCotTheta² > <scatter>`. We gate the expensive
`error²` (the spB–spM + spM–spT error sum plus the spM correlation term) behind that
cheap comparison and compute it lazily, at most once. `error²` is skipped entirely
for the candidates that are cot-θ-compatible but rejected by the helix/curvature cut
before the refined cut is reached.

## Performance

Measured on current ACTS main, ITk, ttbar + PU200, single thread (median of 3
repeats): the seed stage is **~1.5 % faster**

### DIFF
--- a/Core/src/Seeding/TripletSeedFinder.cpp
+++ b/Core/src/Seeding/TripletSeedFinder.cpp
@@ -74,14 +74,14 @@ class Impl final : public TripletSeedFinder {
       // use geometric average
       const float cotThetaAvg2 = cotThetaB * cotThetaT;
 
-      // add errors of spB-spM and spM-spT pairs and add the correlation term
-      // for errors on spM
-      const float error2 = topDoublet.er() + erB +
-                           2 * (cotThetaAvg2 * varianceRM + varianceZM) *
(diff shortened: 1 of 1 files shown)


andiwand:
thanks @Aksth070600 !

I will try to get some numbers out of Athena using this patch. Depending on this we can make a decision on how to proceed.

Aksth070600:
Thanks @andiwand! Sounds good :-)

PR #5681, opened 2026-07-13
fix: propagate PadBlueprintNode padding to parent container

noemina:
I have seen an unexpected behaviour while investigating the use of the `PadBlueprintNode` for the ATLAS use-case.

When a `PadBlueprintNode` is nested as a child of a
`ContainerBlueprintNode` (rather than only used as the implicit root
"World" node created by `Blueprint::construct()`), its padding is
silently lost one level up. The parent container sizes itself from
`PadBlueprintNode::build()`'s return value, which by design is the
unpadded child volume. `Blueprint::construct()` already knows this and
calls `trackingVolume()` explicitly for the outermost node; no other
container does.
This means that if a sub-detector builder wraps its own subtree
in a `PadBlueprintNode` (e.g. to add a local envelope) and that node
ends up as a plain child of some other container, the container's own
bounds are computed as if the padding never happened; and two nested
pads can end up producing volumes of identical size, causing at the very end navigation failures.

`ContainerBlueprintNode::build()` now detects `PadBlueprintNode`
children explicitly and uses `trackingVolume()` (the padded volume)
for its own sizing and shell bookkeeping.

I have added a unit test to explicitly test this behaviour as well.

--- END COMMIT MESSAGE ---

@paulgessinger @andiwand @junggjo9


Core/src/Geometry/ContainerBlueprintNode.cpp
```diff
@@ -52,11 +53,20 @@ Volume& ContainerBlueprintNode::build(
...
-    Volume& volume = child.build(options, gctx, logger);
-    m_childVolumes.push_back(&volume);
+    Volume& built = child.build(options, gctx, logger);
+    Volume* volume = &built;
+    if (auto* pad = dynamic_cast<PadBlueprintNode*>(&child); pad != nullptr) {
+      // PadBlueprintNode::build() intentionally returns the *unpadded* child
+      // volume (padding is only reflected in trackingVolume()). Containers
+      // must size themselves using the padded volume, or the padding is
+      // silently lost for every ancestor except the implicit root node,
+      // which Blueprint::construct() special-cases via trackingVolume().
```
andiwand, line 63: what is the reason for `PadBlueprintNode` not returning the padded volume?

andiwand (commented):
I cannot say much other than the change looking consistent. relying on @paulgessinger expertise here

Core/src/Geometry/ContainerBlueprintNode.cpp
paulgessinger: Right, I don't think it should be necessary to special case this, actually. Let me think about what to do.

paulgessinger: I think changing `PadBlueprintNode` to return the modified padded volume would also fix the problem @noemina was seeing in ITk.

andiwand (approved):

PR #5683, opened 2026-07-14
ci: Run physmon histcmp comparisons in parallel

andiwand:

The physmon CI job currently takes ~40 minutes, and most of it is not physics: dataset generation takes ~9 min while the histcmp comparison stage takes ~27 min. The comparison time is dominated by 12 invocations on track fitting performance files (~110 s each), which render ~400 matplotlib ratio plots per invocation. Since all comparisons are independent, they can run concurrently.

### Changes

- `phys_perf_mon.sh` accepts an optional third argument with the number of concurrent comparison processes (default 1, i.e. unchanged behavior).
- `run_histcmp` enqueues each comparison into a bounded pool of background processes. Each job writes its log and exit code to a scratch directory; after all jobs finish, the logs are replayed inside the usual `::group::` markers and the results CSV is written in submission order, so console output, summary and exit code behave exactly as before.
- Dataset generation stays sequential because its wall time and memory are measured and tracked in the runtime metrics.
- Both the GitHub Actions and GitLab CI jobs pass `$(nproc)`.

### Test

Ran the comparison stage locally (generation inputs seeded from the reference files) with 4 jobs: 31 comparisons totalling ~700 s of histcmp work completed in ~3.5 min wall clock, with identical CSV/summary content and correct exit code propagation for failing jobs.

Expected effect on CI: comparison stage ~27 min → ~7 min on the 4-core runners.

Part of a series to speed up physmon:
- histcmp-side parallel plot rendering and a flag to skip projection plots (PRs in paulgessinger/histcmp)
- optional PDF plot output (follow-up PR)


benjaminhuth:
image
image

This is very nice!

benjaminhuth (approved):
LGTM!

PR #5686, opened 2026-07-14
refactor: Reimplement padding with shared functionality

paulgessinger:
...instead of literal node class reuse;

Co-authored-by: Noemi Calace <noemi.calace@cern.ch>


Core/include/Acts/Geometry/PadBlueprintNode.hpp
```diff
@@ -45,9 +45,25 @@ class PadBlueprintNode final : public StaticBlueprintNode {
...
+  ///       @p inner.
+  /// @param gctx The geometry context
+  /// @param inner The volume to enclose
+  /// @param envelope The envelope to add to the bounds of @p inner
+  /// @param name The name of the padded volume
+  /// @param logger The logger to use
+  /// @return The padded volume enclosing @p inner
+  /// @throws std::logic_error if @p inner has unsupported bounds, or if the
+  ///         envelope is asymmetric where it must not be
+  static std::unique_ptr<TrackingVolume> padded(
```
dimitra97, line 63: sorry now I start following this more and mostly for my own education: I thought in general there was no ownership of the tracking volume transfered (i remember from the static node for example). do we want this to happen in general? I mean creating the volumes before the finalization of the geometry int he blueprint

paulgessinger: This is a bit of a special case: this here is a factory that produces a new tracking volume that's passed around the volume given as an argument. The pad node uses it to generate the volume that it then owns internally (like the static one) while the root node just uses it during construction to produce the world volume. Does that make sense?

dimitra97: yes ok, thanks!

dimitra97 (approved):

PR #5687, opened 2026-07-14
feat: Add traccc independent Cuda implementation of MuonSpacePoint container

pingponghero12:
The commit introduces alternative CUDA friendly structure-of-arrays implementation of MuonSpacePoint container and space point itself with a ptr object. The implementation fulfills the CompositeSpacePoint and related concepts. It is independent of traccc. Commit includes Root reader utility and unit tests.

--- END COMMIT MESSAGE ---

The container it it's current form uses proxy objects as MuonSpacePoint alternative and stores memory in container object, in similar manner as traccc.

It will be used for CUDA implementation of HoughTransform.

Things worth discussing:
1. I use manually managed memory not unified memory, can I keep it that way or should I change it.
2. Should I use some smart way to minimize the code size of memory operations in .cpp, as it is really bloated. If so what would be preferred.
3. Should I include unittest of reader, tho it is difficult to make one without the writer, and adding both unit test and writer as I have it will bloat the code size again.

Also I would need to abstract some part of my code such as `cudaCheck` or `allocalteDeviceCollumn`, is there preferred place to where I do it, as there is no such thing in the code, with sole exception of some GNN functionality.

@junggjo9 @stephenswat


pingponghero12:
@lmonaco99 this is mostly the container, so the part of code we went through, the HoughPlane is not part of this PR.

Examples/CudaMuonSegmentFitting/include/ActsExamples/EventData/CudaMuonSpacePoint.hpp
```diff
@@ -0,0 +1,394 @@
...
+
+  /// Deleted because the container owns CUDA device memory.
+  CudaMuonSpacePointContainer(const CudaMuonSpacePointContainer&) = delete;
+  CudaMuonSpacePointContainer& operator=(const CudaMuonSpacePointContainer&) =
+      delete;
+
+  /// Movable to transfer ownership of host/device storage.
+  CudaMuonSpacePointContainer(CudaMuonSpacePointContainer&& other) noexcept;
+  CudaMuonSpacePointContainer& operator=(
+      CudaMuonSpacePointContainer&& other) noexcept;
```
junggjo9, line 265: CAn this one be defaulted?

Examples/Framework/src/EventData/CudaMuonSpacePoint.cpp
```diff
@@ -0,0 +1,497 @@
...
+
+template <typename T>
+void copyColumnToHost(std::vector<T>& hostColumn, const T* deviceColumn) {
+  if (hostColumn.empty() || deviceColumn == nullptr) {
+    return;
+  }
+
+  cudaCheck(cudaMemcpy(hostColumn.data(), deviceColumn,
+                       hostColumn.size() * sizeof(T), cudaMemcpyDeviceToHost));
+}
```
junggjo9: I suppose these can be replaced with traccc function now?

Examples/Framework/src/EventData/CudaMuonSpacePoint.cpp
```diff
@@ -0,0 +1,497 @@
...
+
+CudaMuonSpacePointContainer::size_type CudaMuonSpacePointContainer::bucketEnd(
+    size_type bucket) const {
+  checkBucket(bucket);
+  return m_host.bucketEnd[bucket];
+}
+
+void CudaMuonSpacePointContainer::addBucket(size_type start, size_type end) {
+  if (start > end || end > m_size) {
+    throw std::out_of_range("CudaMuonSpacePointContainer invalid bucket range");
```
junggjo9:
```suggestion
    throw std::out_of_range(std::format("CudaMuonSpacePointContainer invalid bucket range [{:};{:}]. Allowed [0;{:})", start,end, m_size));
```
To give the use more information

Examples/CudaMuonSegmentFitting/src/CudaMuonSpacePoint.cpp
```diff
@@ -0,0 +1,497 @@
...
+  checkBucket(bucket);
+  return m_host.bucketEnd[bucket];
+}
+
+void CudaMuonSpacePointContainer::addBucket(size_type start, size_type end) {
+  if (start > end || end > m_size) {
+    throw std::out_of_range("CudaMuonSpacePointContainer invalid bucket range");
+  }
+
+  m_host.bucketStart.push_back(static_cast<std::uint32_t>(start));
```
junggjo9, line 374: why not changing the signature to std::uint_32? Then you spare a cast

Examples/Framework/src/EventData/CudaMuonSpacePoint.cpp
```diff
@@ -0,0 +1,497 @@
...
+  m_onDevice = false;
+}
+
+CudaMuonSpacePointContainer::value_type CudaMuonSpacePointContainer::at(
+    size_type index) const {
+  checkIndex(index);
+  return value_type{*this, index};
+}
+
+void CudaMuonSpacePointContainer::checkIndex(size_type index) const {
```
junggjo9:
```suggestion
inline void CudaMuonSpacePointContainer::checkIndex(size_type index) const {
```

Examples/Framework/src/EventData/CudaMuonSpacePoint.cpp
```diff
@@ -0,0 +1,497 @@
...
+  if (index >= m_size) {
+    throw std::out_of_range("CudaMuonSpacePointContainer index out of range");
```
junggjo9:
```suggestion
    throw std::out_of_range(std::format("CudaMuonSpacePointContainer index {:}out of range. Max allowed: {:}", index, m_size-1));
```

Examples/Framework/src/EventData/CudaMuonSpacePoint.cpp
```diff
@@ -0,0 +1,497 @@
...
+  }
+}
+
+void CudaMuonSpacePointContainer::checkBucket(size_type bucket) const {
+  if (bucket >= bucketCount()) {
+    throw std::out_of_range("CudaMuonSpacePointContainer bucket out of range");
```
junggjo9:
```suggestion
    throw std::out_of_range(std::format("CudaMuonSpacePointContainer bucket {:} out of range. Max allowed: {:}" , bucket, bucketCount() - 1));
```

junggjo9:
@stephenswat do you have a minute to review ;)

Examples/CudaMuonSegmentFitting/include/ActsExamples/EventData/CudaMuonSpacePoint.hpp
pingponghero12: To my understanding it cannot, because on structure of raw pointers, same as on just raw pointer the move would copy the data but leave the old data not changed to nullptr, but probably undefined, meaning it will likely in reality hold the pointer still, and then when first object goes of scope it will on destructor free the memory which of the pointer both old and new hold.

That is why I just use exchange here:       m_device{std::exchange(other.m_device, {})}

The move assignment  also cannot be, since we have to clear GPU memory of the first one before coping new pointer into it.

I tested the behavior on small prototype and the default in fact did not change the ptr to nullptr.

Examples/CudaMuonSegmentFitting/src/CudaMuonSpacePoint.cpp
pingponghero12: bucketStart is an index for the array, so I think it makes sense for consistency to keep it of size_type, it just is stored as uint32 for processing to limit memory usage. This will not add much overhead as this operation is only called once per bucket during loading and it is like 2 lines so I think we could maybe keep it.

Examples/Framework/CMakeLists.txt
```diff
@@ -92,4 +93,12 @@ if(NOT TBB_FOUND)
     find_package(TBB ${_acts_tbb_version} MODULE REQUIRED)
 endif()
 
+if(ACTS_ENABLE_CUDA)
+    find_package(CUDAToolkit REQUIRED)
+
+    target_compile_definitions(ActsExamplesFramework PRIVATE ACTS_ENABLE_CUDA)
+
+    target_link_libraries(ActsExamplesFramework PRIVATE CUDA::cudart)
+endif()
```
paulgessinger: This will need proper integration, rather than modifying the existing target. 

We'll need to add a separate cuda aware target that handles this, and I believe there is some traccc precedent at this point.

Examples/Framework/include/ActsExamples/EventData/CudaMuonSpacePoint.hpp
```diff
@@ -0,0 +1,394 @@
...
+//
+// Copyright (C) 2016 CERN for the benefit of the ACTS project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+#ifdef ACTS_ENABLE_CUDA
```
paulgessinger: No ifdefs like this please.

Examples/Framework/src/EventData/CudaMuonSpacePoint.cpp
```diff
@@ -0,0 +1,504 @@
...
+//
+// Copyright (C) 2016 CERN for the benefit of the ACTS project
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "ActsExamples/EventData/CudaMuonSpacePoint.hpp"
+
+#ifdef ACTS_ENABLE_CUDA
```
paulgessinger: Same here, this needs to live in a cuda only library

Examples/Io/Root/CMakeLists.txt
```diff
@@ -50,6 +51,19 @@ target_link_libraries(
     PRIVATE ROOT::Core ROOT::Hist ROOT::Tree ROOT::TreePlayer
 )
 
+if(ACTS_ENABLE_CUDA)
```
paulgessinger: No coupling of the root target to cuda. If you want to read/write these via root, you'll need to add a neutral (=non cuda) data structure that can live in the general event data includes.

paulgessinger (changes requested):
This will need work to decouple the cuda parts from the rest of the code. None of the other libraries should conditionally depend on cuda or require preprocessor based guards for cuda-non cuda mode.

junggjo9 (dismissed):

pingponghero12:
@paulgessinger I updated and moved the code as we discussed. It is now inside `Examples/CudaMuonSegmentFitting`. 

It is decoupled from rest of code base. I added new compile flag `ACTS_BUILD_EXAMPLES_CUDA_MUON_SEGMENT_FITTING`, which triggers the build of it and of unit tests.

I got rid of reader, cuda container is now constructed from MuonSpacePointContainer.

Could you maybe look into it when you have a bit of time? Thanks a lot!

pingponghero12:
I'm thinking that maybe I should just build it when cuda is enabled to avoid the need for it to be added to CI.

Examples/CudaMuonSegmentFitting/include/ActsExamples/EventData/CudaMuonSpacePoint.hpp
```diff
@@ -0,0 +1,394 @@
...
+  element_type* operator->() const noexcept { return &m_proxy; }
+
+  /// Dereferences the proxy.
+  element_type& operator*() const noexcept { return m_proxy; }
+
+  /// Checks whether this pointer-like object is valid.
+  explicit operator bool() const noexcept { return m_valid; }
+
+ private:
+  mutable element_type m_proxy{};
```
paulgessinger: Mutable because you allow retrieving mutable refs through const accessor. Is that a design decision?

Examples/CudaMuonSegmentFitting/src/CudaUtilities.hpp
paulgessinger, line 1: depending on if you need them outside of the target `CudaMuonSegmentFitting` you could consider moving this header into the `src` directory (even though it's a header).

Examples/CudaMuonSegmentFitting/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+if(NOT CMAKE_CUDA_COMPILER)
+    enable_cuda()
+endif()
```
paulgessinger: Do you think we can combine that with:

https://github.com/acts-project/acts/blob/081ddad192f6153bb9ac51b3cacef48a38f9667e/CMakeLists.txt#L547-L548

so that there's only a single call to `enable_cuda()` and `find_package(CUDAToolkit REQUIRED)` (outside of traccc and detray)? Right now it's gated behing `ACTS_BUILD_PLUGIN_GNN`, but you could pull it out to just gate on `ACTS_ENABLE_CUDA`, which you're already toggling on if this here gets enabled, right?

pingponghero12: Yes, this is a better solution I changed to that in the new commit.

Examples/CudaMuonSegmentFitting/src/CudaUtilities.hpp
pingponghero12: This is very valid, those utilities are only useful for this part of code base, in rest of code the analogical utilities should be used from traccc or vecmem.

Examples/CudaMuonSegmentFitting/include/ActsExamples/EventData/CudaMuonSpacePoint.hpp
pingponghero12: This followed the logical-constness idea from before, but here it did not really add anything useful. I updated it to remove `mutable` and made the dereference operators non-const instead.

paulgessinger (dismissed):

paulgessinger (approved):

PR #5688, opened 2026-07-14
feat: Python track fitter performance writer

benjaminhuth:


Examples/Framework/include/ActsExamples/Validation/TrackFitterPerformanceCollector.hpp
```diff
@@ -0,0 +1,98 @@
...
+    int fitIterations = 3;
+  };
+
+  TrackFitterPerformanceCollector(Config cfg,
+                                  std::unique_ptr<const Acts::Logger> logger);
+
+  /// Fill histograms for one event.
+  ///
+  /// @note The caller must ensure exclusive access (e.g. hold a mutex).
+  ProcessCode fill(const Acts::GeometryContext& geoContext,
```
andiwand: - does this need a `ProcessCode` if it always succeeds?
- outside the sequence we usually use `Acts::Result`

Examples/Framework/include/ActsExamples/Validation/TrackFitterPerformanceCollector.hpp
```diff
@@ -0,0 +1,98 @@
...
+    /// mean/width fit.
+    int fitMinEntries = 10;
+    /// The range in sigma for the iterative Gaussian fit
+    double fitSigmaRange = 3.0;
+    /// The maximum number of iterations for the iterative Gaussian fit
+    int fitIterations = 3;
+  };
+
+  TrackFitterPerformanceCollector(Config cfg,
+                                  std::unique_ptr<const Acts::Logger> logger);
```
andiwand, line 50: I wonder if there can / should als be a `clear` function

andiwand (dismissed):
I think this is pretty nice! thanks for splitting this out

Examples/Framework/include/ActsExamples/Validation/TrackFitterPerformanceCollector.hpp
benjaminhuth: Good idea, but this would be a larger refactor I think, so I wouldn't add it in this PR.

andiwand (approved):

PR #5689, opened 2026-07-14
feat: Changes to improve material loading with JSON geometries

benjaminhuth:


paulgessinger:
I think we should phase out the use of the material decorator altogether. I'm on board with the tracking geometry json io reading and writing it, but the decorator should not be involved.

Plugins/Json/include/ActsPlugins/Json/TrackingGeometryJsonConverter.hpp
```diff
@@ -182,22 +186,30 @@ class TrackingGeometryJsonConverter {
...
+      const IMaterialDecorator* materialDecorator = nullptr) const;
 
   /// Reconstruct a tracking geometry from a deserialized JSON file
   ///
   /// @param gctx geometry context
   /// @param encoded Reference to the encoded JSON object
   /// @param options options for the conversion
+  /// @param materialDecorator optional decorator applied to every
+  ///        reconstructed surface and volume, e.g. to attach material that
+  ///        was not embedded in the JSON payload itself
```
paulgessinger: In this case you would apply it after the conversion, and not through the decorator class.

paulgessinger (changes requested):

benjaminhuth:
> I think we should phase out the use of the material decorator altogether. I'm on board with the tracking geometry json io reading and writing it, but the decorator should not be involved.

Is this a general design decision I have missed? what would be the new concept?

benjaminhuth:
Close - can be implemented with `trackingGeometry.apply`

benjaminhuth:
Okay, there are some remaining changes still to be merged

paulgessinger (approved):

PR #5690, opened 2026-07-14
fix: correct mom. assignment for G4 secondaries, also set parentParticleId

altsybee:
Two fixes to `Examples/Algorithms/Geant4/src/ParticleTrackingAction.cpp`, both affecting Geant4-produced secondary particles:

## 1. Use the momentum magnitude, not the kinetic energy
The secondary's absolute momentum was set from `track.GetKineticEnergy()`:
```cpp
const G4double p = convertEnergyToActs * track.GetKineticEnergy();
...
aParticle.setAbsoluteMomentum(p);
```
which is wrong since `setAbsoluteMomentum` expects `|p| = sqrt(E^2 - m^2)`, while `GetKineticEnergy()`
returns `T = E - m`.  Changed to:
```cpp
const G4double p = convertEnergyToActs * track.GetMomentum().mag();
```
QA check using K0S decays to π⁺π⁻:
image

## 2. Populate `parentParticleId` for Geant4 secondaries

The Geant4 sim builds secondary barcodes with `makeDescendant()` but never records the parent, so `parentParticleId()` was left at its default for every secondary, making mother→daughter relations (decay/secondary vertices) unrecoverable. The parent barcode is already available via `eventStore().trackIdMapping.at(GetParentID())`, so we now set it on the new particle (primaries have `parentId == 0` and keep the default). It is also re-applied in `PostUserTrackingAction` after `convert()` rebuilds the final state, so both states carry it.

## Impact

Only Geant4 secondaries are affected (input primaries unchanged). Enables truth mother/daughter + secondary-vertex reconstruction and gives physically correct secondary momenta. No API or configuration change.


andiwand (dismissed):
Very important fixes - thank you @altsybee !

andiwand:
@altsybee this requires a reference file update. you can find the new file here https://github.com/acts-project/acts/actions/runs/29352812173?pr=5690 (at the end of the page). the file that needs updating is this one https://github.com/acts-project/acts/blob/main/CI/physmon/reference/simulation/particles_geant4_hist.root

andiwand (approved):

PR #5696, opened 2026-07-15
fix: make the space-point grid phi-binning robust for large impactMax

altsybee:
`CylindricalSpacePointGridCreator` computes the number of phi bins from the maximum azimuthal deflection, which includes `asin(impactMax / rMin)` with `rMin = rMax - deltaRMax`. 
When `impactMax >= rMin` the `asin` argument exceeds 1 and returns `NaN`; the `deltaPhi <= 0` guard does not catch `NaN`, so the phi-bin count casts to 0 and the grid axis constructor throws `Axis: Invalid binning, at least one bin is needed.`

Geometrically a track with `|d0| >= r` cannot reach radius `r`, so the azimuthal deflection saturates at pi/2. Clamping the `asin` arguments to `[0, 1]` yields a valid (coarse) phi binning in this regime instead of a crash -- in the worst case the phi bins become maximally coarse (the deflection saturates at pi/2), so the neighbour search effectively covers the full 2pi (search-all-phi), which is slow but correct.

No behaviour change when `impactMax < rMax - deltaRMax`.
Enables large-`impactMax` / displaced-vertex seeding (e.g. to track V0 daughters).


andiwand (approved):
makes sense to me - thanks for the fix @altsybee !

PR #5699, opened 2026-07-16
chore: Handle discarded Acts::Result

benjaminhuth:
Non-breaking subset of cases that would be enforced by #5700

### DIFF
--- a/Core/include/Acts/TrackFitting/detail/GsfActor.hpp
+++ b/Core/include/Acts/TrackFitting/detail/GsfActor.hpp
@@ -202,7 +202,14 @@ struct GsfActor {
       // No hole before first measurement
       if (result.processedStates > 0 && surface.isSensitive()) {
         TemporaryStates tmpStates;
-        noMeasurementUpdate(state, stepper, surface, result, tmpStates, true);
+        Result<void> res = noMeasurementUpdate(state, stepper, surface, result,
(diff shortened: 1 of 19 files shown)


andiwand (approved):

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
```diff
@@ -202,7 +202,14 @@ struct GsfActor {
       // No hole before first measurement
       if (result.processedStates > 0 && surface.isSensitive()) {
         TemporaryStates tmpStates;
-        noMeasurementUpdate(state, stepper, surface, result, tmpStates, true);
+        Result<void> res = noMeasurementUpdate(state, stepper, surface, result,
+                                               tmpStates, true);
+        if (!res.ok()) {
+          if (m_cfg.abortOnError) {
+            std::abort();
```
paulgessinger, line 209: Really abort? That hard terminates the program

paulgessinger (approved):

Core/include/Acts/TrackFitting/detail/GsfActor.hpp
benjaminhuth: yeah so this also caught me again, forgot about that. but the option `abortOnError` is public API, so I cannot remove it right now...

PR #5701, opened 2026-07-16
refactor: Attach DigitizationDesign to surfaces

delitez:
This PR introduces a design abstraction layer for digitization configuration as preparation for a long-term goal: to move the digitization configuration from a standalone GeometryHierarchyMap lookup to a pointer stored directly on each Surface.


Core/include/Acts/Surfaces/Surface.hpp
```diff
@@ -254,6 +255,17 @@ class Surface : public virtual GeometryObject,
...
+  /// Assign a design object to this surface.
+  ///
+  /// The design is planned to describe the sensor independently of the
+  /// digitization algorithm.
+  ///
+  /// @param design The design object to attach
+  void assignDesign(std::shared_ptr<const IDesign> design);
+
+  /// Return the design object associated to this surface, or nullptr.
+  const IDesign* design() const { return m_surfaceDesign.get(); }
```
paulgessinger: I was wondering if we should push this into the `SurfacePlacement`/`DetectorElement`. What do you think?

paulgessinger: What do you think @delitez ?

Examples/Algorithms/Digitization/src/DigitizationAlgorithm.cpp
```diff
@@ -144,6 +145,16 @@ DigitizationAlgorithm::DigitizationAlgorithm(
...
   m_digitizers = Acts::GeometryHierarchyMap<Digitizer>(digitizerInput);
+
+  // Attach a DigitizationDesign to each surface that has a digitizer
+  for (const auto& [geoId, surfacePtr] : m_cfg.surfaceByIdentifier) {
+    auto digitizerItr = m_digitizers.find(geoId);
+    if (digitizerItr == m_digitizers.end()) {
+      continue;
+    }
+    auto design = std::make_shared<DigitizationDesign>(&(*digitizerItr));
+    const_cast<Acts::Surface*>(surfacePtr)->assignDesign(std::move(design));
```
paulgessinger: Can we avoid the const cast somehow?

Core/include/Acts/Surfaces/Surface.hpp
delitez: It depends on if we want to keep `design` general or not. Isn't `SurfacePlacement` only for sensors?

paulgessinger: Ahm, yes indeed. I guess I was thinking that only sensors get designs..

Core/include/Acts/Surfaces/SurfacePlacementBase.hpp
```diff
@@ -59,5 +62,17 @@ class SurfacePlacementBase {
   //         a detector surface
   /// @return True if this is a sensitive surface
   virtual bool isSensitive() const = 0;
+
+  /// @brief  Returns a pointer to the sensor design associated with this placement
+  ///         element. If the placement is not associated with a sensor design,
+  ///         it returns a nullptr.
+  /// @return Pointer to the sensor design or nullptr
+  virtual const ISensorDesign* sensorDesign() const { return nullptr; }
```
junggjo9, line 70: Can't the sensor design just inherit from the SurfacePlacement? That's how the readout geometry is setup in Athena

delitez: It can, and in that case each detector element would have its own desing object. But here the idea is to have a shared design pointer so the surfaces with the identical config can resolve the same (digicomponents)config entry.

junggjo9: Hmm, or you can just pipe it into the examples implementation?

delitez: Right but if I do so, I must link `ActsExamplesDigitization` against `ActsExamplesDetectordd4hep` , so i can use `DD4hepDetectorElementWithDesign`.

paulgessinger: The idea we discussed was that a detector element / surfaceplacement HAS a design, not that it IS one.

Core/include/Acts/Surfaces/SurfacePlacementBase.hpp
```diff
@@ -59,5 +62,17 @@ class SurfacePlacementBase {
...
+
+  /// @brief Assign a sensor design to this placement element.
+  ///        Default implementation is a no-op; override in subclasses
+  ///        that support design assignment.
+  virtual void assignSensorDesign(
+      const std::shared_ptr<const ISensorDesign>& /*design*/) const {}
```
paulgessinger, line 76: I think what we had discussed was to try to keep this method OUT of the `SurfacePlacementBase` and only in `DD4hepDetectorElementWithDesign`.

Examples/Algorithms/Digitization/src/DigitizationAlgorithm.cpp
```diff
@@ -144,6 +145,21 @@ DigitizationAlgorithm::DigitizationAlgorithm(
...
+    if (digitizerItr == m_digitizers.end()) {
+      continue;
+    }
+
+    const auto* placement = surfacePtr->surfacePlacement();
+    if (placement == nullptr) {
+      continue;
+    }
+
+    auto design = std::make_shared<DigitizationDesign>(&(*digitizerItr));
```
paulgessinger, line 160: Here you would just add a `dynamic_cast<const DD4hepDetectorElementWithDesign*>(placement)` and if that works you do the `assignSensorDesign` below.

Examples/Detectors/DD4hepDetector/include/ActsExamples/DD4hepDetector/DD4hepDetector.hpp
paulgessinger, line 133: Add a `@TODO` explaining that this is temporary for verification.

Examples/Detectors/DD4hepDetector/include/ActsExamples/DD4hepDetector/DD4hepDetectorElementWithDesign.hpp
```diff
@@ -0,0 +1,49 @@
...
+    return m_design.get();
+  }
+
+  void assignSensorDesign(
+      const std::shared_ptr<const Acts::ISensorDesign>& design) const override {
+    m_design = design;
+  }
+
+ private:
+  mutable std::shared_ptr<const Acts::ISensorDesign> m_design{nullptr};
```
paulgessinger, line 36:
```suggestion
  void assignSensorDesign(
      const std::shared_ptr<const Acts::ISensorDesign>& design) override {
    m_design = design;
  }

 private:
  std::shared_ptr<const Acts::ISensorDesign> m_design{nullptr};
```

Try to avoid mutable.

Examples/Algorithms/Digitization/src/DigitizationAlgorithm.cpp
delitez: If I add a `dynamic_cast`, I'll have to add `ActsExamplesDetectorDD4hep` to the cmake link of `ActsExamplesDigitization`. isn't it something we'd like to avoid?

paulgessinger: True. Other suggestion:

put another base next to `SurfacePlacementBase` but put it in detail: `detail::ISensorDesignHolder` (we can remove `detail` without breaking versions). This is **temporary** and goes away in the follow up work. Then, `DD4hepDetectorElementWithDesign` can just inherit from that one, and it's a neutral `dynamic_cast` target.

Makes sense?

delitez: Yes, I can try that!

paulgessinger (approved):

PR #5703, opened 2026-07-17
ci(traccc): Fix call to apt and add update

stephenswat:
This commit replaces a call to apt with a call to apt-get in the traccc CI. It also adds an apt-get update to avoid a persistent CI issue which has popped up recently.

### DIFF
--- a/.github/workflows/traccc.yml
+++ b/.github/workflows/traccc.yml
@@ -133,7 +133,7 @@ jobs:
           persist-credentials: false
       - name: Install dependencies
         run: |
-          apt install -y zstd wget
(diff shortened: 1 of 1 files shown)


paulgessinger:
Wondering if we can push this to the container images

stephenswat:
> Wondering if we can push this to the container images

Agree this would be the nicer option, but we use a few images that were removed in v83, so we cannot easily update those machines. Thinking of it, that's a pretty unsustainable development model, so maybe we should get rid of the CI jobs using those deprecated machines?

stephenswat:
Closing in favour of #5704.

PR #5712, opened 2026-07-21
feat: Update OpenDataDetector to v6.0.2

themrluke:
- Updated the ODD dependency from **v4.0.4 to v6.0.2**. v6 adds a calorimeter and muon system to the existing silicon tracker. Material remapping not necessary as the Volume IDs, surface counts and the digitisation configs are unchanged

- Added `buildTracker` / `buildCalorimeter` / `buildMuonSystem` to `getOpenDataDetector()` (default to `True`). These flags allow subsystems that aren't needed to be skipped.

- 12 ODD-dependent **reference hashes regenerated**: Geant4 simulation and material recording, CKF on the ODD, ODD Kalman track states, and ODD strip space points.

--- END COMMIT MESSAGE ---

- the GSF and refitting references were not modified. Their hashes did not match, but this happens on the generic detector too, which the ODD cannot influence, so the cause is not this change. I have left those for CI to regenerate in the reference environment.

@asalzburger


asalzburger:
Approved the workflow, so we can see what it does.

asalzburger:
Rebased, let's get this going ...

asalzburger:
Fixed the python import.

asalzburger:
Fixed MuonSystem to default: False (we will switch on later)

asalzburger (dismissed):
It may require a round of hashes update, but let's see.

asalzburger (approved):

asalzburger:
Closing that in favor of #5940 which is same content.

PR #5715, opened 2026-07-21
feat: implement dwalk and optimize cuda cc for GNN plugin

jaychan-hep:
- Added DWALK graph segmentation implementation for the GNN4ITk pipeline
- Optimized the ConnectedComponents implementation

--- END COMMIT MESSAGE ---

@benjaminhuth


Plugins/Gnn/src/DWalkTrackBuilding.cu
```diff
@@ -0,0 +1,1625 @@
...
+};
+
+struct DpCudaState {
+  float *bestScore = nullptr;
+  int *bestChild = nullptr;
+  unsigned char *sourceMask = nullptr;
+  std::size_t numNodes = 0;
+};
+
+void freeDeviceOrientedEdges(DeviceOrientedEdges &graph, cudaStream_t stream) {
```
benjaminhuth: Hmm as far as I see, these `free...` functions are in nested control flow below. I think it would be saver if we would add proper destructors to structs like `DeviceOrientedEdges`, that automatically free the memory if the type goes out of scope. (would require to forbid copy-construction/assignment to avoid double free)

Examples/Scripts/Python/gnn4itk_example.py
```diff
@@ -121,12 +130,28 @@ def runGNN4ITk(
     }
     graphConstructor = ModuleMapCuda(**moduleMapConfig)
 
+    if useEdgeLayerConnector:
+        trackBuilder = "edgeLayerConnector"
+    trackBuilder = {
```
benjaminhuth: Hmm jthis seems a bit redundant to me, can we just work with the lowercase strings?

Plugins/Gnn/include/ActsPlugins/Gnn/detail/ConnectedComponents.cuh
```diff
@@ -239,6 +248,9 @@ TLabel connectedComponentsCuda(std::size_t nEdges, const TEdges *sourceEdges,
   makeLabelMask<<<gridDim, blockDim, 0, stream>>>(nNodes, labels, tmpMemory);
   ACTS_CUDA_CHECK(cudaGetLastError());
 
+  TLabel nLabels = thrust::reduce(thrust::device.on(stream), tmpMemory,
```
benjaminhuth: Hmm is this really faster? Below, we just copy the label from the end of the `tmpMemory` array, but now we aare doing a full reduction...

benjaminhuth (commented):
Hi @jaychan-hep, thanks for the PR, I had a first look and have a few comments. In general, you could consider also adding some unit tests for your algorithmic parts similar as they exist for the connected components, but this is not mandatory.
Also, you should run `pre-commit` to make the linter test pass.

Plugins/Gnn/src/DWalkTrackBuilding.cu
jaychan-hep: Good point, thanks. I replaced the manual `freeDevice...` helpers with move-only RAII structs for the device graph/state buffers, with copy construction/assignment deleted to avoid double frees. I kept explicit `reset()` calls where we intentionally release memory before the end of the scope.

Examples/Scripts/Python/gnn4itk_example.py
jaychan-hep: Agreed. I simplified this to use lowercase track-builder strings internally and in the CLI choices: `junctionremoval`, `edgelayerconnector`, and `dwalk`

Plugins/Gnn/include/ActsPlugins/Gnn/detail/ConnectedComponents.cuh
jaychan-hep: Good point. I removed the `full thrust::reduce`. The component count is now computed from the last exclusive-scan value plus the last label-mask value, so it avoids the full reduction while still handling the case where the last node is itself a surviving component label. I rebuilt and tested this locally, and the D-WALK workflow looks fine (no visible different in timing).

jaychan-hep:
Hi @benjaminhuth, I pushed the requested updates.

I also ran `pre-commit` on the touched files locally, and it passes now.

It looks like the CI bridge did not trigger for my latest commit. Could you please rerun the CI bridge when you have a chance?

Thanks!

benjaminhuth (dismissed):
LGTM

jaychan-hep:
Hi @benjaminhuth, I pushed the lint fix and the lightweight checks are passing now. It looks like `policy-bot: main` is still pending (`0/1 rules approved`) and merge-sentinel still reports the required full CI groups as missing for the latest commit. Could you please approve/trigger the required checks when you have a chance?

benjaminhuth (dismissed):

benjaminhuth (approved):

PR #5716, opened 2026-07-22
fix: Force positive scale factor for `qopt` residuals in `ResPlotTool`

andiwand:
Otherwise we flip the sign and `reco - truth` becomes `truth - reco` and we mix them together which will be hard to interpret.

--- END COMMIT MESSAGE ---


benjaminhuth (approved):
LGTM

PR #5718, opened 2026-07-22
feat: configurable bin expansion for SurfaceArrayNavigationPolicy

paulgessinger:
## Summary

Adds a `binsExpansion` factor to `SurfaceArrayNavigationPolicy::Config` and support for auto-determining grid bin counts from the module layout.

A bin count of `0` in either direction now requests automatic sizing via `SurfaceArrayCreator::determineBinCount` (one bin per module row), multiplied by `binsExpansion`. The default `binsExpansion = 1.0` means existing configurations are unchanged.

`SurfaceArrayCreator::determineBinCount` is moved from `private` to `public` so navigation policies can query it.

## Motivation

Detector builders currently hard-code the acceleration-grid bin counts for each layer. This mirrors the Gen1 layer builder's `numberOfBinsFactor`, letting builders size the grid relative to sensor granularity instead. In the ATLAS ITk Gen3 blueprint this lets layers use module-derived grids (`bins = {0, 0}`, `binsExpansion = 5`), which markedly reduces the candidate set returned per `neighbors()` query and thus navigation cost, without hand-tuning bin counts per layer.

## Behavior / backward compatibility

- `bins = {N, M}` with `N, M > 0`: unchanged (explicit bin counts).
- `bins` with a `0` component: that axis is auto-sized from the module layout, scaled by `binsExpansion`.
- `binsExpansion` defaults to `1.0` → no change for existing users.

## Changes

- `SurfaceArrayNavigationPolicy::Config`: add `binsExpansion` (default `1.0`); document the `bins == 0` auto-determination.
- `SurfaceArrayNavigationPolicy` ctor: auto-determine per-axis bin counts (disc: R/phi, cylinder: phi/Z) when requested, scaled by `binsExpansion`.
- `SurfaceArrayCreator::determineBinCount`: `private` → `public`.


Core/include/Acts/Navigation/SurfaceArrayNavigationPolicy.hpp
```diff
@@ -27,8 +27,14 @@ class SurfaceArrayNavigationPolicy : public INavigationPolicy {
     /// The type of the layer
     LayerType layerType = LayerType::Cylinder;
     /// The number of bins in the local directions. The interpretation depends
-    /// on the layer type.
+    /// on the layer type. A value of 0 in either direction requests automatic
+    /// determination from the module layout (one bin per module), scaled by
+    /// @c binsExpansion.
     std::pair<std::size_t, std::size_t> bins;
+    /// Multiplier applied to auto-determined (0) bin counts, mirroring the
+    /// legacy Gen1 layer builder's numberOfBinsFactor. Finer grids return fewer
```
noemina: I would remove the reference to Gen1

Core/include/Acts/Navigation/SurfaceArrayNavigationPolicy.hpp
```diff
@@ -27,8 +27,14 @@ class SurfaceArrayNavigationPolicy : public INavigationPolicy {
...
+    /// surfaces per neighbour query, cutting navigation candidate counts.
+    double binsExpansion = 1.0;
```
andiwand:
```suggestion
    double numberOfBinsFactor = 1.0;
```
- in case we want to stick with the current naming
- expansion sounds to me like it can only grow while one can also use numbers smaller than 1
- should we check for non-negative input?

andiwand (dismissed):

andiwand (approved):

PR #5719, opened 2026-07-22
fix: Consistent minimum required python version

benjaminhuth:
Also:
* Use universal lock files for requirements (`uv pip compile --universal`)
* Format changes due to black reacting to `pyproject.toml`-specified version.


pyproject.toml
```diff
@@ -5,6 +5,10 @@ build-backend = "scikit_build_core.build"
 [project]
 name = "pyacts"
 dynamic = ["version", "readme"]
+# Single source of truth for the Python versions Acts supports. Drives the
+# cibuildwheel target matrix and the floor that the checked-in requirements.txt
+# lockfiles are compiled against (see CI/supported_python_versions.py).
+requires-python = ">=3.11,<3.15"
```
paulgessinger: I think generally upper bounding is frowned upon in python land. Do we need it?

paulgessinger (commented):
Bit awkard with the script to detemine the version, but ok. 
Where to the python format changes come from?

benjaminhuth:
black seems to react to the minimum python version that is given in `pyproject.toml`

pyproject.toml
benjaminhuth: Yeah I know that its not recommended by some people. however, I wouldn't know how to embed the upper bound for the version parsing otherwise...

benjaminhuth: Hmm alternatively: we could embed the upper version in `supported_python_versions.py`. maybe thats cleaner

benjaminhuth:
Solves at least the linux build: https://github.com/acts-project/acts/actions/runs/30115899541

paulgessinger (approved):

PR #5720, opened 2026-07-22
perf(navigation): reduce Gen3 ODD propagation overhead

paulgessinger:
## Summary

Gen3 geometry propagation through the Open Data Detector was ~14% slower than the equivalent Gen1 run, almost entirely in navigation. Profiling localized the gap to per-propagation heap churn plus per-call overhead in the navigation-policy machinery. This branch closes it in two passes — **Gen1 is unchanged except for one shared win (commit 9), which makes it faster too.**

Net effect (100k single-threaded ODD propagations, muon gun, `EigenStepper`, IO off):

| | heap alloc / propagation | instructions | run time |
|---|---|---|---|
| gen3 before | 99.3 | — | ~4.55 s |
| gen3 after pass 1 | 15.2 | 39.2e9 | ~4.10 s |
| **gen3 after pass 2** | **14.2** | **37.2e9** | **~3.80 s** |
| gen1 (reference) | 13.2 | 36.9e9 | ~4.0 s |

i.e. from ~14% slower than Gen1 to **faster in wall clock** and within ~0.5% in instructions.

## Changes (one per commit, each independently measured)

**Pass 1 — allocations and per-step overhead:**

1. **Remove per-propagation heap allocations from Gen3 navigation** (99.3 → 15.2 alloc/prop):
   - `NavigationStream::initialize`: replace the allocating `std::stable_sort` used for candidate de-duplication with an in-place first-occurrence scan (same "first wins" semantics, no `_Temporary_buffer`), and reuse the `additionalCandidates` scratch buffer.
   - `NavigationPolicyStateManager`: back the type-erased state stack with `Acts::AnyBase<128>` instead of `std::any`, so the small policy states are stored inline instead of heap-allocated. Hidden entirely in the manager — no policy changed.
2. **Store `MultiNavigationPolicy` child states linearly**: drop the stored vector of child-state handles in favour of index addressing (children live contiguously on the stack), via new `NavigationPolicyState::index()`/`atIndex()`.
3. **Skip the per-step `isValid()` when the policy state is the default sentinel** (−2.4% instructions): the always-true validity check was pure per-step overhead for policies that don't override it. A policy with a meaningful `isValid()` necessarily pushes a real state, so `NavigationPolicyState::isDefault()` (true for the default `EmptyState`) is a footgun-free signal for when the navigator can skip the call.

**Pass 2 — remaining hot spots from a fresh `perf annotate`:**

4. **`AnyBase`: no zero-fill of the small buffer on default construction** — the default member initializer memset 128 bytes on every `pushState` (`rep stos`, ~29% of `createState` self time). The buffer is only ever read through the handler; the default ctor becomes user-provided so `const AnyBase` stays default-initializable.
5. **Cache policy-state defaultness per volume transition** — `isDefault()` cost a static-guard check + PLT call to `typeHash<EmptyState>()` *every step*; it can only change at volume transitions, so the navigator caches it as a bool.
6. **Consume candidates directly off the navigation stream** — drop the per-rebuild copy-out into a separate `small_vector` (~112 B/candidate + heap spill); the near/far path-length window is applied lazily while advancing through the (path-length sorted) stream, which accepts exactly the same candidates in the same order. Wall ~4.06 → ~3.94 s (memory traffic, not instructions).
7. **Skip the candidate de-duplication pre-pass when the caller guarantees uniqueness** — duplicates can only come from external/free surfaces, which the navigator appends itself, so it knows when there are none (new defaulted `candidatesAreUnique` hint on `initialize`; post-sort unique remains as a unit-tested backstop).
8. **Skip `createState`/`popState` for stateless policies** — `INavigationPolicy` caches whether it pushes only default states, probed once per policy at the end of `Blueprint::construct` under a documented contract (state *defaultness* must not depend on context/arguments). Create and pop are skipped under the same immutable condition, so the state stack stays balanced by construction; debug builds always exercise the subsystem and assert the contract. −2.5% instructions, −1 alloc/prop.
9. **Inline the `checkPathLength` fast path** (shared win) — two comparisons were an out-of-line call with up to four verbose-filter checks, hot in both navigators; verbose diagnostics are delegated out of line. **gen1: 37.6 → 36.9e9 instructions (−1.7%)**; gen3 benefits equally.

A byte-arena rewrite of the state manager was prototyped and **reverted** — it measured slightly worse (the cost is per-call, not storage). Rationale and full profiling methodology are in `ANALYSIS.md` (happy to drop or relocate that doc).

## Testing

`ActsUnitTestNavigationStream`, `ActsUnitTestNavigationPolicy`, `ActsUnitTestNavigationPolicyState`, `ActsUnitTestNavigator`, `ActsUnitTestAny`, `ActsUnitTestIntersection`, `ActsUnitTestPropagator`, `ActsUnitTestEigenStepper`, and the Blueprint/TrackingGeometry suites pass. `ConeValidityPolicy` (a policy with a real `isValid()` and state) still forces re-resolutions unchanged — confirming the sentinel/stateless machinery is footgun-free. A few pre-existing cases are skipped locally only because the lean build has `ACTS_ENABLE_LOG_FAILURE_THRESHOLD=OFF`; unrelated to these edits.

## Notes

- Draft: opened for review of the approach; standard CI/labels not yet addressed.
- Measured against a material-off configuration to isolate geometry/navigation; the allocation and call-overhead deltas are independent of material.
- The stateless-policy probe (commit 8) runs in `Blueprint::construct`; any other Gen3 geometry source would need the same call — unprobed policies just take the normal (correct, slower) path.


ANALYSIS.md
andiwand, line 1: this should be dropped or moved

Core/include/Acts/Utilities/Intersection.hpp
```diff
@@ -300,16 +301,40 @@ static_assert(std::is_trivially_move_assignable_v<MultiIntersection2D>);
...
 /// @param nearLimit The minimum path length for an intersection to be considered
 /// @param farLimit The maximum path length for an intersection to be considered
 /// @param logger A optionally supplied logger which prints out a lot of infos
 ///               at VERBOSE level
-bool checkPathLength(double pathLength, double nearLimit, double farLimit,
-                     const Logger& logger = getDummyLogger());
+inline bool checkPathLength(double pathLength, double nearLimit,
+                            double farLimit,
+                            const Logger& logger = getDummyLogger()) {
+  if (logger.doPrint(Logging::VERBOSE)) [[unlikely]] {
```
andiwand, line 330: I think this is the first time we use `[[unlikely]]` in ACTS? I imagine it won't do much since the `if` is perfectly predictable after the first try

Core/src/Navigation/NavigationStream.cpp
```diff
@@ -20,26 +20,46 @@ namespace Acts {
...
-  m_candidates.erase(initialDuplicates.begin(), initialDuplicates.end());
+  // once in this pass, keeping the first occurrence (in insertion order). This
+  // reproduces the previous std::stable_sort + std::unique result (first-wins),
+  // but in place: it avoids the temporary buffer that std::stable_sort
+  // allocates on every call, which matters on this per-navigation hot path. The
+  // candidate count per volume is small, so the quadratic scan is cheap — but
+  // it is skipped entirely when the caller guarantees uniqueness. (Should a
+  // duplicate slip through regardless, the post-sort unique pass below still
+  // removes it; only the first-wins tolerance selection is then not enforced.)
+  if (!candidatesAreUnique) {
```
andiwand, line 38: I guess this optimization could also be done for Gen1? not that we should do it but if so we are comparing a non-optimized vs an optimized solution

Core/include/Acts/Navigation/NavigationStream.hpp
```diff
@@ -113,6 +113,10 @@ class NavigationStream {
   /// @param queryPoint holds current position, direction, etc.
   /// @param cTolerance is the candidate search tolerance
   /// @param onSurfaceTolerance is the tolerance for on-surface intersections
+  /// @param candidatesAreUnique the caller guarantees that no surface was
+  ///        added more than once, so the pre-intersection de-duplication pass
+  ///        can be skipped. Candidates that intersect the same surface twice
+  ///        (multi-intersections) are still handled correctly.
```
andiwand, line 119: I wonder if this could be externalized as a hard requirement. I guess not because `NavigationStream` is stateful? if so it would be nice because we could push the complexity out completely and deal with it on the caller site

Core/include/Acts/Utilities/Any.hpp
```diff
@@ -816,7 +829,11 @@ class AnyBase : public AnyBaseAll {
 #endif
       );
 
-  alignas(kMaxAlignment) std::array<std::byte, SbSize> m_data{};
+  // Deliberately not value-initialized: the buffer is only ever read through
+  // m_handler, which is set exactly when a value has been constructed in (or
+  // pointed to from) the buffer. Zero-filling it on every default construction
+  // would put a memset on hot paths that create AnyBase objects frequently.
+  alignas(kMaxAlignment) std::array<std::byte, SbSize> m_data;
   const Handler* m_handler{nullptr};
```
andiwand, line 837:
```suggestion
  const Handler* m_handler;
```
I guess in this case we can also skip zeroing this?

Core/include/Acts/Navigation/INavigationPolicy.hpp
```diff
@@ -237,6 +286,40 @@ class INavigationPolicy {
...
+  /// @note This relies on the contract that the *defaultness* of the state a
+  ///       navigation policy pushes does not depend on the geometry context
+  ///       or the navigation arguments: a policy either always pushes a real
+  ///       state (it has something to validate) or always pushes the default
+  ///       empty state.
+  ///
+  /// @param gctx The current geometry context object, e.g. alignment
+  /// @param logger Logger for debug output
+  void probeStatelessness(const GeometryContext& gctx,
+                          const Logger& logger = getDummyLogger()) {
```
andiwand: just from the name I expected this to be invariant. do you see any way to make this not dependent on the calling order?

Core/src/Geometry/Blueprint.cpp
```diff
@@ -213,6 +213,16 @@ std::unique_ptr<TrackingGeometry> Blueprint::construct(
   Acts::detail::AlignablePortalVisitor alignPortals{gctx, logger};
   world->apply(alignPortals);
 
+  // All navigation policies are attached at this point: probe each one for
+  // whether it pushes only default states, so the navigator can skip the
+  // per-volume-entry state creation for volumes with stateless policies.
+  world->apply([&](TrackingVolume &volume) {
+    if (INavigationPolicy *policy = volume.navigationPolicy();
+        policy != nullptr) {
+      policy->probeStatelessness(gctx, logger);
```
andiwand: this was the surprising stateful side-effect part triggering the other comment

Core/src/Navigation/Navigator.cpp
```diff
@@ -506,10 +525,17 @@ NavigationTarget Navigator::getNextTargetGen3(State& state,
...
-  bool isValid = state.currentVolume->navigationPolicy()->isValid(
-      state.options.geoContext, {.position = position, .direction = direction},
-      policyState, logger());
+  // Only consult isValid() when the policy state is non-default; a default
+  // (EmptyState) state signals the resolved sequence cannot be invalidated.
+  // The defaultness is cached in the navigator state since it only changes at
+  // volume transitions.
+  bool isValid = true;
+  if (!state.policyStateIsDefault) {
+    auto policyState = state.policyStateManager.currentState();
```
andiwand, line 534:
```suggestion
    const auto& policyState = state.policyStateManager.currentState();
```

Core/src/Navigation/Navigator.cpp
```diff
@@ -559,6 +598,28 @@ NavigationTarget Navigator::tryGetNextTarget(State& state,
   }
 }
 
+void Navigator::createPolicyState(State& state, const Vector3& position,
+                                  const Vector3& direction) const {
+  const TrackingVolume& volume = *state.currentVolume;
```
andiwand, line 605:
```suggestion
    assert(state.currentVolume != nullptr && "currentVolume is nullptr");

    const TrackingVolume& volume = *state.currentVolume;
```

andiwand (commented):
Left a couple of comments - looks overall good to me

One thing I wonder is if we are comparing apples to apples with the Gen1 vs Gen3 geometry navigation. I marked on case here, but similarly I think the cylinder policy puts Gen3 at an advantage as this is not optimized in Gen1

ANALYSIS.md
paulgessinger: oh yeah right, sorry

Core/include/Acts/Utilities/Intersection.hpp
paulgessinger: I didn't (have claude) measure the specific impact of the `[[unlikely]]` only. Is there any reason we should avoid adding it?

Core/src/Navigation/NavigationStream.cpp
paulgessinger: Yes, probably. I think the comparison wrt Gen1 is less "which one is faster" but "does it regress from what we're running now". I can apply the optimization to Gen1 if you want, though.

Core/include/Acts/Navigation/NavigationStream.hpp
paulgessinger: You mean externalize by saying this should always be the case? I think we were handling this because potentially nav policies can push surfaces multiple times, and external surfaces can potentially also produce duplicates?

Core/include/Acts/Utilities/Any.hpp
paulgessinger: Probably yes

Core/include/Acts/Utilities/Intersection.hpp
andiwand: no I don't think it hurts. just wanted to flag the potential paradigm shift in not having vs having this in ACTS

Core/include/Acts/Navigation/INavigationPolicy.hpp
paulgessinger: What do you mean by invariant? You think it's surprising that it mutates the policy? And I'm not sure I understand the calling order issue.

Core/src/Geometry/Blueprint.cpp
paulgessinger: Right, ok. This happens ahead of time, so that the policy doesn't latch this lazily when it's first called. I pushed for making this an initialization step rather than a lazy late initialization. Do you think raneming the probe method is enough?

Core/src/Navigation/NavigationStream.cpp
andiwand: I think at the end if observe similar compute time with marginal optimization difference it is fine. we can always try to optimize more. putting Gen1 on equal footing might move the target again and introduce more noise overall. I think we we want Gen3 for its flexibility and while getting there not buying into noticeably slowdowns

Core/include/Acts/Navigation/NavigationStream.hpp
andiwand: right - I fear the interface does not really allow putting this burden on the caller

Core/include/Acts/Utilities/Any.hpp
paulgessinger: Actually, no. `nullptr` is an actual sentinel here, that drived the `operator bool` and also internally what do to when assigning.

Core/src/Geometry/Blueprint.cpp
andiwand: yes I think that would be fair. I guess in the best case this could be determined during construction but I guess it can't? alternative name from top of my head `determineStatelessness`

or we go one step further and name it `initialize` which is a hard requirement to be called otherwise consecutive operations on the object will trigger failures

Core/src/Navigation/Navigator.cpp
paulgessinger: The state is a fairly small handle type that's returned by value, so the const ref probably doesn't do the right thing here.

Core/include/Acts/Navigation/INavigationPolicy.hpp
paulgessinger: Renamed to `initializeStatelessCache`.

Core/src/Geometry/Blueprint.cpp
paulgessinger: Renamed to `initializeStatelessCache`. If the initialization is skipped, the navigator will just fall back to the slow path where it assumes the policy is stateless, so not a correctness but only a performance concern. Note that it's automatically called when closing the geometry, so the only failure case is if the policy is reassigned after the geometry is closed.

andiwand (approved):

PR #5722, opened 2026-07-22
ci: smoke-test husk self-hosted runners (do not merge)

paulgessinger:
**Draft, for CI validation only — not for merge.**

End-to-end test of the [husk](https://github.com/acts-project) self-hosted runner pool (CERN OpenStack, AlmaLinux + rootless podman) against real ACTS CI.

### What changed

| Job | Before | After |
| --- | --- | --- |
| `linux_ubuntu` | `ubuntu-latest` | `[self-hosted, linux, x64, husk-size-standard]` |
| `linux_examples_test` | `ubuntu-latest` | `[self-hosted, linux, x64, husk-size-standard]` |
| `linux_physmon` | `ubuntu-latest` | `[self-hosted, linux, x64, husk-size-standard]` |
| `lcg_cvmfs` | *(new)* | `[self-hosted, linux, x64, husk-size-standard, cvmfs]` |

Everything else is untouched and still runs on GitHub-hosted runners.

### What it is actually testing

1. **`container:` on a podman slot.** All three moved jobs use `container:`, which the GitHub runner implements by shelling out to `docker`. On a husk slot that binary is the podman shim. Every heavy ACTS job is container-based, so if this does not work, nothing else moves — which is why these jobs were picked and not the `checks.yml` ones (those call `sudo apt-get` directly on the runner, and the slot is AlmaLinux).

2. **CVMFS end to end.** `lcg_cvmfs` mirrors the GitLab `lcg_108` / alma9 / gcc15 leg. The `cvmfs` label means the client is mounted on the slot and the configured repositories are bound into the job container — not that some LCG toolchain happens to be installed. It verifies the mount *before* sourcing the view, so a mount failure is distinguishable from a wrong view path.

3. **Pool churn.** The pool is `max_total = 2`, so these four jobs will queue and recycle slots between them rather than all starting at once. Wall-clock here is not representative of a sized fleet.

### On the `runs-on` selectors

`runs-on` matches a runner carrying *all* of the listed labels, so every dimension left out is a dimension that accepts any value. `[self-hosted, linux, x64]` alone would match GPU slots once a GPU pool exists; dropping `x64` would match arm64. Both arch and size class are pinned on purpose.

https://claude.ai/code/session_016XR8NwDNBbfruT3UxGHPWA


paulgessinger:
Superseded by #5723, which carries the same branch from a fork.

PR #5723, opened 2026-07-22
ci: run linux_ubuntu on husk self-hosted runners

paulgessinger:
Moves the `linux_ubuntu` job onto the husk self-hosted runner pool (CERN OpenStack, AlmaLinux + rootless podman), and adds the apt-mirror handling that makes doing so worthwhile.

Earlier revisions of this branch were a throwaway fixture with most of CI switched off. That scaffolding is gone: every workflow and job is restored byte-for-byte, and the diff is now five files.

### What changed

| File | Change |
| --- | --- |
| `.github/workflows/builds.yml` | `linux_ubuntu` → `[self-hosted, linux, x64, husk-size-standard]` — one line |
| `.github/actions/apt_mirror/action.yml` | new composite action |
| `.github/actions/dependencies/action.yml` | invokes it as its first step |
| `CI/dependencies/apt_mirror.sh` | new script |
| `CI/dependencies/opengl.sh` | comment only |

Only `linux_ubuntu` moves. `linux_examples_test` and `linux_physmon` stay on `ubuntu-latest` and consume the artifact `linux_ubuntu` builds on a husk slot, so the cross-fleet handoff is exercised without migrating them yet.

### Why the apt mirror

`archive.ubuntu.com` measured 48 kB/s from CERN against ~1.9 MB/s for the Swiss mirror, which made `apt-get` the longest step in `linux_ubuntu` — 4m39s to fetch 9.7 MB of package lists. A GitHub-hosted runner on the same PR measured 82–96 kB/s against the same URLs, so this is upstream slowness rather than anything about the slot.

The URL is deliberately **not** hardcoded, because the right answer depends on where the job runs and this repository cannot know that. The fleet exports `HUSK_UBUNTU_APT_MIRROR` into the job container; GitHub-hosted runners never set it, and the script then does nothing at all. The hosted path is unchanged by construction.

### Why a separate action rather than a script call

The mirror was previously applied from `opengl.sh`, which `setup.sh` reaches only *after* spack is installed, cached and configured — several minutes in. Anything a job apt-installed before that point stayed on the upstream archive. As a composite action it can run as the first step of `.github/actions/dependencies`, covering the whole job, and any future job can apply it before its own `apt-get` with a single `uses:` line.

The `opengl.sh` call is kept, because GitLab CI invokes `setup.sh` directly and never sees the action. The script is idempotent, so the double call costs nothing.

### Distro safety

An Ubuntu mirror is only valid for Ubuntu. Pointing a Debian image at one does not degrade to slow, it 404s, because the suite names do not overlap (`bookworm` vs `noble`). Debian was already unaffected here, but only because the rewrite patterns happened not to match `deb.debian.org` — nothing stated the intent, so broadening a pattern later would have silently broken it.

The script therefore gates explicitly on `ID=ubuntu` from `/etc/os-release`, and skips an image it cannot identify. `ports.ubuntu.com` (arm64) is deliberately left alone for the same reason: the country mirrors that carry the x86 archive frequently do not carry ports.

It also reports what actually happened instead of printing an unconditional success line. The case worth surfacing is a named mirror on an Ubuntu image that rewrote nothing — meaning the sources are in an unrecognised form and the job is silently still on the slow path.

### Verified locally

Run against real containers with a synthetic mirror URL:

| Condition | Result | Sources after |
| --- | --- | --- |
| `ubuntu:24.04` amd64 (deb822) | 1 file rewritten | on the mirror, incl. `-security` |
| `ubuntu:22.04` amd64 (`sources.list`) | 1 file rewritten | all `jammy*` on the mirror |
| `debian:bookworm` amd64 | skipped, logged | untouched |
| `ubuntu:24.04` arm64 (ports) | no rewrite | untouched — correct |
| self-hosted, no mirror named | logged | untouched |
| hosted runner, no mirror named | silent, exit 0 | untouched |

`shellcheck` clean. `zizmor --persona=regular` clean. `actionlint` reports the same 10 pre-existing shellcheck findings in `builds.yml` as `main` — no new ones.

Not verified: `ghcr.io/acts-project/ubuntu2404:87` itself, only the stock base image it derives from. If its apt sources are laid out differently the log will say so explicitly rather than failing quietly.

### On the `runs-on` selectors

`runs-on` matches a runner carrying *all* the listed labels, so every dimension left out is one that accepts any value. `[self-hosted, linux, x64]` alone would match GPU slots once a GPU pool exists; dropping `x64` would match arm64. Both arch and size class are pinned deliberately.

### Fork-PR caveats

The branch lives on a fork, so `GITHUB_TOKEN` is read-only and repository secrets are unavailable. ccache saves and anything secret-dependent behave differently from a branch PR, and timings should be read with that in mind.

### Follow-ups, not in this PR

- `lcg_cvmfs` — a husk mirror of the GitLab `lcg_108` / alma9 / gcc15 leg, exercising CVMFS bound into the job container. Parked on the `ci/husk-lcg-cvmfs` bookmark until this has a timing baseline.
- Migrating `linux_examples_test` and `linux_physmon`.

Supersedes #5722.


benjaminhuth:
Hmm the PR title does not really match with the changes I think, maybe better "Allow runner-dependent apt mirror for ubuntu" or so.

paulgessinger:
@benjaminhuth I think the headline feature is that the `linux_ubuntu` job moves to the self-hosted git runner.

paulgessinger:
@benjaminhuth it does happen through this commit here:

https://github.com/acts-project/acts/pull/5723/changes#diff-a76cf7978f0a981f911e8d68d2351a72a268977304612226433df4fb8203b06f

```yaml
runs-on: [self-hosted, linux, x64, husk-size-standard]
```

andiwand (approved):

PR #5724, opened 2026-07-22
fix: Navigator - Recover missed candidates if start is at portal

junggjo9:
If the navigation starts on a portal surface and the propagation points out of the volume. The start volume needs to be changed to the adjacent volume
--- END COMMIT MESSAGE ---

Tagging: @paulgessinger, @andiwand


Core/src/Navigation/Navigator.cpp
```diff
@@ -205,8 +205,70 @@ Result<void> Navigator::initialize(State& state, const Vector3& position,
 
       return Result<void>::failure(NavigatorError::NotOnExpectedSurface);
     }
-  }
+    // navigation is started on a portal
+    if (m_geometryVersion == GeometryVersion::Gen3 &&
+        state.currentVolume != nullptr &&
+        state.currentSurface->geometryId().boundary() != 0) {
```
andiwand: I think this whole block should be moved out of `if (state.currentSurface != nullptr) {` to separate the logic and concern. also removed one level of nesting

it is a pity that we have to rely on the geometry id here. I don't think we do this anywhere else in the navigator. alternatively we could check if the current surface is one of the boundary / portal surfaces of the current volume. what do you think?

Core/src/Navigation/Navigator.cpp
```diff
@@ -205,8 +205,70 @@ Result<void> Navigator::initialize(State& state, const Vector3& position,
...
+      ACTS_VERBOSE(volInfo(state) << " Start surface is a portal");
+      for (const Portal& portal : state.currentVolume->portals()) {
+        if (portal.surface().geometryId() !=
+            state.currentSurface->geometryId()) {
+          continue;
+        }
```
andiwand: here we actually look for the surface anyway so I think we can just cut the geometry ID dependency

Core/src/Navigation/Navigator.cpp
```diff
@@ -205,8 +205,70 @@ Result<void> Navigator::initialize(State& state, const Vector3& position,
...
 
+        auto res =
+            portal.resolveVolume(state.options.geoContext, position, direction);
+        if (!res.ok()) {
+          ACTS_ERROR(volInfo(state)
+                     << "Failed to resolve volume through portal: "
+                     << res.error().message());
+          return Result<void>::failure(NavigatorError::NotInsideExpectedVolume);
+        }
+        /// The portal does not change the volume
```
andiwand:
```suggestion
        // The portal does not change the volume
```

Core/src/Navigation/Navigator.cpp
```diff
@@ -205,8 +205,70 @@ Result<void> Navigator::initialize(State& state, const Vector3& position,
...
+        if (state.currentVolume == res.value()) {
+          return Result<void>::success();
+        }
```
andiwand: I think we should be carful with early returns in `initialize`. people might put code in the tail of the function and expect it to be called no matter what happens before

so in this case I would rather go for nesting `if (state.currentVolume != res.value()) {` or we just rely on what `res` gives us and initialize with the given volume in any case

Core/src/Navigation/Navigator.cpp
```diff
@@ -205,8 +205,70 @@ Result<void> Navigator::initialize(State& state, const Vector3& position,
...
+          policy->createState(state.options.geoContext,
+                              {.position = position, .direction = direction},
+                              state.policyStateManager, logger());
+        } else {
+          ACTS_VERBOSE(
+              volInfo(state)
+              << "No more volume to progress to, stopping navigation.");
+          state.navigationBreak = true;
+        }
+        break;
```
andiwand: here I feel like we are just undoing work that happened before. can we just delay the policy creation until the end of `initialize`?

Core/src/Navigation/Navigator.cpp
junggjo9: Hmm. We should have a protection whether the surface is associated with a portal. Was thinking of maybe adding a portal() method to the surface? Thah would replace on a long term the layer attribute. What do you think?

Core/src/Navigation/Navigator.cpp
junggjo9: If the navigation initialization fails above, it would have been bailed out alreay, no?

Core/src/Navigation/Navigator.cpp
junggjo9: True point

Core/src/Navigation/Navigator.cpp
junggjo9: Done with portal assignemtn

Core/src/Navigation/Navigator.cpp
junggjo9: Implemented portal

Core/include/Acts/Geometry/Portal.hpp
```diff
@@ -130,6 +130,7 @@ class Portal {
   /// ```
   /// @param gctx The geometry context
   /// @param args The struct containing the arguments
+  [[deprecated("Constructor seems to be unused.. Will be removed soon")]]
```
junggjo9:
```suggestion
```

Core/include/Acts/Surfaces/RegularSurface.hpp
```diff
@@ -115,6 +119,18 @@ class RegularSurface : public Surface {
...
   using Surface::isOnSurface;
+
+  /// Return the pointer to the portal holding this surface
+  /// @return Pointer to the portal holding this surface (might be nullptr)
+  const Portal* associatedPortal() const;
+
+  /// Associate the surface with a portal. The method
+  /// is called at the portal's construction
+  /// @param portal: Reference to the portal holding this surface
+  void associatePortal(const Portal& portal);
```
paulgessinger: I don't think this should be a property of the surface.

Core/src/Navigation/Navigator.cpp
```diff
@@ -164,6 +164,44 @@ Result<void> Navigator::initialize(State& state, const Vector3& position,
...
+      auto res = regularSurf->associatedPortal()->resolveVolume(
+          state.options.geoContext, position, direction);
+      if (!res.ok()) {
+        ACTS_ERROR(volInfo(state) << "Failed to resolve volume through portal: "
+                                  << res.error().message());
+        return Result<void>::failure(NavigatorError::NotInsideExpectedVolume);
+      }
+      // Update the current volume
+      state.currentVolume = res.value();
+    }
```
andiwand: I would still lift this outside the `if (state.currentSurface != nullptr) {` and put it into a self-contained logical block below. but I think that is more a matter of personal preference

andiwand (dismissed):
Thanks for the followup @junggjo9 ! LGTM

Core/src/Navigation/Navigator.cpp
```diff
@@ -164,6 +164,44 @@ Result<void> Navigator::initialize(State& state, const Vector3& position,
...
+                 << "We did not end up on the expected surface. surface = "
+                 << state.currentSurface->geometryId()
+                 << " position = " << position.transpose()
+                 << " direction = " << direction.transpose());
+
+      return Result<void>::failure(NavigatorError::NotOnExpectedSurface);
+    }
+    // navigation is started on a portal
+    if (const auto* regularSurf =
+            dynamic_cast<const RegularSurface*>(state.currentSurface);
```
paulgessinger: Can we avoid downcasts in the navigation loop?

paulgessinger (commented):
I don't think we should support the mechanism of mapping a surface back to a portal. We'll have to find another way to make this information available if this is something we need to support. We could to a slow start thing where we loop over all portals and check if the start surface is one of them, maybe?

Core/include/Acts/Surfaces/RegularSurface.hpp
junggjo9: It's kind of the replacement of the associatedLayer(), no?

Core/src/Navigation/Navigator.cpp
junggjo9: do you prefer a helper method which uses the surface type and then a static_cast?

Core/include/Acts/Surfaces/RegularSurface.hpp
junggjo9: To my opinion it's the cleanest way to find out whether the surface represents a portal. We could try to use the geometryIdentifier but that's an API mechanism on one wants to lock in one self either.  If that's blocked, one would need to loop over all portals at initialization and check whether the surface pointer belongs to the portal. Given that the muon system envelope has roughly 2k portals

Core/src/Navigation/Navigator.cpp
```diff
@@ -565,7 +589,8 @@ void Navigator::resolveCandidates(State& state, const Vector3& position,
     ACTS_VERBOSE(volInfo(state) << "No volume to resolve candidates.");
     return;
   }
-  ACTS_VERBOSE(volInfo(state) << "Searching for compatible candidates.");
+  ACTS_VERBOSE(volInfo(state) << "Searching for compatible candidates."
+                              << direction.transpose());
```
junggjo9:
```suggestion
  ACTS_VERBOSE(volInfo(state) << "Searching for compatible candidates.");
```

paulgessinger (dismissed):

Core/include/Acts/Propagator/Navigator.hpp
```diff
@@ -459,6 +459,8 @@ class Navigator final {
   TrackingGeometry::GeometryVersion m_geometryVersion{};
 
   std::shared_ptr<const Logger> m_logger;
+  /// Cache of the surfaces that are connected with portals
+  std::unordered_map<const Surface*, const Portal*> m_surfPortalMap{};
```
andiwand: I don't think the navigator object should carry any kind of state

Core/src/Navigation/Navigator.cpp
```diff
@@ -38,6 +38,16 @@ Navigator::Navigator(Config cfg, std::shared_ptr<const Logger> _logger)
...
+
+  m_cfg.trackingGeometry->apply([this](const Portal& portal) {
+    auto [insert, ok] =
+        m_surfPortalMap.insert(std::make_pair(&portal.surface(), &portal));
+    if (!ok && &portal != insert->second) {
+      throw std::invalid_argument(
+          std::format("Navigator Surface: {:} used for multiple portals",
+                      portal.surface().geometryId()));
+    }
+  });
```
andiwand: ok this seems to be done on navigator construction, fair enough, but why not put it into the tracking geometry in the first place?

tbh linking from the surface to the portal seems way better IMO

Core/src/Navigation/Navigator.cpp
```diff
@@ -164,6 +174,57 @@ Result<void> Navigator::initialize(State& state, const Vector3& position,
...
+    if (!locPos.ok()) {
+      ACTS_VERBOSE(volInfo(state)
+                   << "Navigation does not start on the surface = "
+                   << state.currentSurface->geometryId()
+                   << " position = " << position.transpose()
+                   << " direction = " << direction.transpose());
+      return Result<void>::failure(NavigatorError::NotOnExpectedSurface);
+    }
+    if (state.currentSurface->insideBounds(*locPos,
+                                           BoundaryTolerance::None())) {
```
andiwand, line 208: the logic here took me a while to digest - I think I got it now

only if we are inside the bounds we check on which side of the volume we are. in case we are outside we do the same as if we were on any other surface

generally starting from outside the bounds is risky because we could end up in a different volume than the surface is. this is the risk we currently take, just flagging that we do the same for starting from a boundary surface now

Core/include/Acts/Propagator/Navigator.hpp
paulgessinger: I asked for this, it's not really state, it's navigator local static information that's computed from the tracking geometry.

Core/src/Navigation/Navigator.cpp
paulgessinger: I asked for it to be local in the Navigator to not churn the tracking geometry interface. My argument was that this is the most local change and we can see if we need to deduplicate this by attaching it to the tracking geometry.

I think the `Surface` API should be unaware that it can be part of a portal, that's why I pushed back against the surface methods and member.

andiwand:
I checked in Athena where we create navigator objects and it looks like we would be mostly fine there as it almost always happens during `initialize`, but there are a few cases which do it during `execute`. ACTS Examples seems to be similar.

my main concern would be that the cost of creating a navigator object suddenly becomes non-negligible. if you think it is fine we can merge it as is and not touch the geometry @paulgessinger @junggjo9 

there are a few backlinks we currently have in our geometry, like `surface->layer->volume` but it seems more like a Gen1 artifact and it is not done consistently. so I would agree that linking `surface->portal` is rather outside of what we used to do. the next best option IMO would be putting it on the tracking geometry. either by a map like proposed here or with an high-level interface similar to `lowestTrackingVolume` (maybe an overload?) that gets `surface, position, direction` as input and returns the correct volume which might be behind a portal. what do you think?

paulgessinger:
I think "best practice" is not to create navigators in the fly. We probably do it in some cases but we should try to avoid it if possible. 

as discussed with @junggjo9, if this turns out to cause issues or be undesirable, we can still move it into the tracking geometry.

junggjo9:
I've setup an instance counter within athena. Thus far there are 18 navigator instances with 30.7k portals each. Given that we expect the number of  instances to grow and we're already at a footprint of roughly 30MB, I've put it into th tracking geometry.

junggjo9:
Superseeded by https://github.com/acts-project/acts/pull/5740

PR #5728, opened 2026-07-23
fix(traccc): don't treat in-tree Acts/Detray headers as system includes

paulgessinger:
Acts (and, transitively, Detray) are pulled into the standalone traccc build via FetchContent with the SYSTEM keyword, which marks their headers as -isystem for every consumer. That suppresses -Werror on those headers: diagnostics originating in Acts/Detray headers are shown but never promoted to errors, so warnings in that code slip through CI unnoticed.

Detray and Traccc now live in-tree and are the source of truth, so their headers should be treated as first-party. Drop the SYSTEM keyword from the Acts FetchContent declaration; Detray inherits the change via the directory-scoped SYSTEM property.

**Stack** (bottom to top):

- #5728 build/no-system  👈
- #5730 build/traccc-colocated-acts
- #5732 ci/traccc-detray-modern-deps
- #5757 chore/detray-tu-generation


Detray/core/include/detray/navigation/detail/intersection_kernel.hpp
```diff
@@ -77,8 +77,12 @@ struct intersection_initialize {
       }
       assert(mask_idx < mask_group.size());
 
-      result = intersector.point_of_intersection(
-          traj, ctf, mask_group[mask_idx], cfg.overstep_tolerance);
+      // Some mask group containers (e.g. device vectors) index with a 32 bit
+      // size type, so narrow explicitly rather than implicitly on subscript.
+      const auto idx = static_cast<typename mask_group_t::size_type>(mask_idx);
```
niermann999: Would it help to just declare the `mask_idx` above as `dindex`?

paulgessinger: Probably

paulgessinger: Not `mask_group_t::size_type`?

niermann999: That would also work. It is `std::size_t` for host and `dindex` for device, I think. The mask index from the single or range link that you are accessing here should be a `dindex` in either case, because it comes from the detector data state which obviously cannot switch size between host and device.

I would try to see if this works, because it is readable without knowledge about vecmem:
```
dindex mask_idx{dindex_invalid};
```

niermann999 (dismissed):

paulgessinger:
This becomes a bit wider now, since `-Wconversion` and `-Wshadow` trip in a few places in the ACTS code.

stephenswat (approved):
LGTM!

PR #5730, opened 2026-07-23
refactor(traccc): always build against colocated in-tree Acts

paulgessinger:
Remove the standalone-era machinery for consuming Acts as an external
dependency, now that Acts/Detray/Traccc are colocated and in-tree:
- extern/acts: FetchContent -> colocated add_subdirectory (no git fetch)
- drop the find_package/TRACCC_USE_SYSTEM_ACTS installed path
- drop the separate 'Build ACTS' + install steps in the GitHub CI, build
  Acts from source inline like the GitLab job already does
- build the in-tree Acts/Detray with warnings-as-errors
  (CMAKE_COMPILE_WARNING_AS_ERROR), matching Acts' own CI, so warnings in
  their own compilation fail the traccc build too

**Stack** (bottom to top):

- ~~#5728 build/no-system~~ (merged)
- #5730 build/traccc-colocated-acts  👈
- #5732 ci/traccc-detray-modern-deps
- #5757 chore/detray-tu-generation
- #5776 refactor/unified-warning-flags


stephenswat (dismissed):
This would be extremely useful and the code changes look good. 👍

stephenswat (dismissed):

stephenswat (approved):

PR #5731, opened 2026-07-23
refactor: extract CRTP base class for seeding space point grids

andiwand:
Moves the coordinate-independent functionality of the three seeding space point grids (`CylindricalSpacePointGrid`, `Experimental::SphericalSpacePointGrid`, `CartesianSpacePointGrid`) into a CRTP base class `Acts::detail::SpacePointGridBase<derived_t, grid_t>` in `Acts/Seeding/detail/SpacePointGridBase.hpp`.

The base owns the grid and the binned group and provides `clear`, `binIndex(Vector3)`, `insert(index, Vector3)`, `extend` (which dispatches to the derived `insert(proxy)` for the coordinate mapping), the bin/grid/group accessors and counters, plus protected `sortBinsBy` / `computeRange` helpers parameterized on a sort-key projection. The derived classes keep their `Config`, validation, axis construction, the named coordinate overloads (`insert(index, phi, z, r)` etc.), and their public sorting API (`sortBinsByR` / `sortBinsByCoord`), which now delegate to the base helpers with the appropriate projection.

The public API of all three grids is unchanged; net -100 lines while adding the base class documentation.

--- END COMMIT MESSAGE ---

Verified locally: all touched TUs and the `GridTripletSeedingAlgorithm` consumer compile warning-free, and the `SphericalSpacePointGrid` and phi-binning unit tests pass against the refactored code.


Core/include/Acts/Seeding/CylindricalSpacePointGrid.hpp
```diff
@@ -25,25 +25,23 @@ namespace Acts {
...
+               Axis<AxisType::Equidistant, AxisBoundaryType::Closed>,
+               Axis<AxisType::Variable, AxisBoundaryType::Open>,
+               Axis<AxisType::Variable, AxisBoundaryType::Open>>> {
  public:
-  /// Space point index type used in the grid.
-  using SpacePointIndex = std::uint32_t;
-  /// Type alias for bin container holding space point indices
-  using BinType = std::vector<SpacePointIndex>;
   /// Type alias for phi axis with equidistant binning and closed boundaries
   using PhiAxisType = Axis<AxisType::Equidistant, AxisBoundaryType::Closed>;
```
benjaminhuth: It would be nice, if one could query the axis type from the base class, such as `using PhiAxisType = std::tuple_element<GridType::multi_axis_t::AxisTuple, 0>`, etc.
That would enforce consistency.

Core/src/Seeding/CylindricalSpacePointGrid.cpp
```diff
@@ -12,9 +12,17 @@
 
 namespace Acts {
 
+namespace {
+
+float radiusProjection(const ConstSpacePointProxy& sp) {
```
benjaminhuth: Hmm this is implemented twice as free function. I think, given its shortness, a lambda would fit well here.

benjaminhuth (dismissed):
Looks good to me, two minor comments

benjaminhuth (approved):
LGTM

PR #5736, opened 2026-07-24
ci: store physmon references by content hash instead of in git

paulgessinger:
Moves the physmon reference histograms out of git. The repository keeps only
their hashes in `CI/physmon/reference.sha256`; the contents live as
content-addressed blobs in an OCI registry on GHCR.

--- END COMMIT MESSAGE ---

## Why

`CI/physmon/reference/` is **76% of all blob weight reachable from `main`** —
142 MB of history for 16 MB of working tree. ROOT files are internally
compressed, so git's delta compression buys nothing and each update writes full
new blobs.

The rate has also changed sharply. New reference bytes per year on `main`:

| Year | Added |
|---|---|
| 2023 | 11.8 MB |
| 2024 | 19.6 MB |
| 2025 | 19.7 MB |
| 2026 (through July) | **76.4 MB** |

That is not more frequent updates (46 in 2025 vs 18 so far in 2026) — the files
got bigger. The 2D/3D resolution profiles from #5196 / #5417 pushed
`performance_trackfitting.root` and `performance_fitting_ckf.root` to ~1.2–1.5 MB
each, against ~40 KB for a typical `*_hist.root`. Every reference update now
costs ~4 MB of permanent history.

This only stops the bleeding. Reclaiming the existing 142 MB needs a history
rewrite, which is a separate decision and deliberately not part of this PR.

## How it works

- `CI/physmon/reference.sha256` is `sha256sum` format, one line per file, sorted.
  `sha256sum -c ../reference.sha256` works on it directly from the reference
  directory.
- `phys_perf_mon.sh` populates the reference directory from the manifest before
  comparing, so local runs need no extra step. Blobs are cached under
  `~/.cache/acts/physmon-references`. `ACTS_PHYSMON_NO_FETCH=1` opts out.
- `CI/physmon/reference.py` (PEP 723 + typer, run with `uv run --no-project`)
  provides `pull`, `verify`, `candidate`, `import` and `update`.

### Updating references

Publishing is manual and uses the built-in `GITHUB_TOKEN` — no PAT:

1. Maintainer dispatches **Upload physmon references** from `main` with a run id.
2. It uploads the blobs and reports the manifest to commit.
3. The author commits that manifest and pushes.

Contributors on forks only ever commit a text file, so this works without giving
fork PRs any registry access.

**The changed set comes from the histcmp exit codes, not from comparing hashes.**
ROOT files embed a creation timestamp in their first `TKey`, so every run
produces a different hash for every file even when the physics is identical —
hash comparison would rewrite all forty entries on every update. `run_histcmp`
now records the reference and monitored paths (`histcmp_results.csv` gains two
columns; `summary.py` updated accordingly), and each run emits
`reference-candidate.sha256` into the physmon artifact.

A side benefit: a wholesale reference refresh is now visible in review as forty
changed lines rather than "binary files differ". Three of the last thirty
reference commits touched all forty files against a median of 4–10, which looks
like the whole output directory being copied.

## Before this can leave draft

- [x] Create the `physmon-references` package and seed it:
      `oras login ghcr.io && uv run --no-project CI/physmon/reference.py import`
- [x] Confirm `oras push` with an explicit `application/octet-stream` media type
      stores files raw (there is a hard assertion that fails loudly otherwise)
- [x] Do **not** set a retention policy that prunes untagged manifests, or
      checking out old commits breaks

**Physmon CI will fail on this PR until the package is seeded** — the fetch has
nothing to resolve against. Everything else was verified offline: manifest
generation, the `sha256sum -c` cross-check, candidate computation, the legacy
3-column CSV path, cache-backed pull, hash verification, idempotent re-pull and
corruption detection.

`.github/labeler.yml` needed repointing at the manifest; its
`CI/physmon/reference/**` glob would otherwise silently never match again.


CI/physmon/reference.py
```diff
@@ -0,0 +1,806 @@
...
+    for lineno, line in enumerate(path.read_text().splitlines(), start=1):
+        line = line.strip()
+        if not line or line.startswith("#"):
+            continue
+        try:
+            digest, relpath = line.split(None, 1)
+        except ValueError:
+            console.print(f"[red]{path}:{lineno}: malformed line: {line!r}[/red]")
+            raise typer.Exit(1)
+        digest = digest.lstrip("*")
```
andiwand:
```suggestion
        relpath = relpath.lstrip("*")
```

andiwand (dismissed):
looks overall good to me

andiwand (approved):

PR #5738, opened 2026-07-24
refactor: Describe axes with AxisSpec instead of ProtoAxis

andiwand:
A binning configuration knows some axis properties up front and others only once
the surface exists. `ProtoAxis` modelled that as a single "autorange" flag and
left `adjustBinUtility` to fill in the rest, down to guessing from a `> 1.9π`
range whether an azimuthal axis wraps.

`Acts::AxisSpec` makes each property optional on its own. Whatever the spec
leaves open is supplied at build time as `AxisSpec::Options`, under one rule per
property: it comes from exactly one side, or from both with the same value.
Supplying a property the spec already fixes therefore validates it instead of
overriding it. `Acts::MultiAxisSpec` bundles one spec per grid dimension, and
`resolveMultiAxis` builds a surface binning against that surface, taking range
and boundary type from its bounds.

This removes the last production use of `ProtoAxis`/`DirectedProtoAxis`, which
are deprecated along with their remaining consumers.

--- END COMMIT MESSAGE ---

### Behaviour changes

- full-circle phi material binning resolves to `Closed` instead of `Bound`
- DD4hep drops the `> 1.9π` closedness heuristic and no longer invents `Bound`
  when the XML gives no boundary type, leaving that property to the surface
- surface binning is strictly two-dimensional; use a single bin in one direction
  to bin along the other only
- `configureFace` (Experimental) no longer accepts `DirectedProtoAxis`

### Migrations

`ProtoGridSurfaceMaterial` becomes `ProtoSurfaceMaterialT<MultiAxisSpec2D>`;
`MultiWireVolumeBuilder` binning, DD4hep conversion (`convertAxisSpecs`),
JSON (`AxisSpecJsonConverter`) and the Python bindings follow the new types.
The Experimental blueprint API switches hard, everything else is deprecated.


asalzburger:
Getting back to that ...

Core/include/Acts/Geometry/SurfaceArrayCreator.hpp
```diff
@@ -326,16 +326,14 @@ class SurfaceArrayCreator {
   /// (currently possible: AxisPhi, AxisR, AxisZ)
   /// @param protoLayer Instance of @c ProtoLayer holding generic layer info
   /// @param transform is the (optional) additional transform applied
-  /// @return Instance of @c ProtoAxis containing determined properties
-  /// @note This only creates the @c ProtoAxis, this needs to be turned
-  ///       into an actual @c Axis object to be used
+  /// @return Type-erased @c IAxis containing the determined properties
```
asalzburger, line 329: That's then only a comment update I suppose ?

Core/include/Acts/Material/ProtoSurfaceMaterial.hpp
```diff
@@ -131,11 +131,11 @@ class ProtoSurfaceMaterialT : public ISurfaceMaterial {
...
-/// @brief Type alias for a prototype surface material using a grid of ProtoAxis
-/// A surface material implementation that uses a vector of ProtoAxis for
-/// grid-based binning
-using ProtoGridSurfaceMaterial =
-    ProtoSurfaceMaterialT<std::vector<DirectedProtoAxis>>;
+/// @brief Type alias for a prototype surface material using a multi-axis
+/// binning description
+/// A surface material implementation that carries a MultiAxisFactory whose
+/// deferred axes are resolved against the surface during material mapping
+using ProtoGridSurfaceMaterial = ProtoSurfaceMaterialT<MultiAxisFactory>;
```
asalzburger: I will probably de-template that then, since Gen2 is gone, there is no user anymore of the `ProtoSurfaceMaterialT<BinUtility>` - but this could go into a follow-up PR.

Core/include/Acts/Surfaces/SurfaceAxisResolution.hpp
```diff
@@ -0,0 +1,64 @@
+// This file is part of the ACTS project.
```
asalzburger: Let's try to change the naming here:

`AxisResolution` can be mistaken, I get what you want to say here,  can we make it the action rather the result, call it "SurfaceAxisResolver" and then return explicit a range and direction ?

Core/include/Acts/Surfaces/SurfaceAxisResolution.hpp
```diff
@@ -0,0 +1,64 @@
...
+/// returned in canonical direction order, each carrying its direction.
+///
+/// @param multiAxisFactory the binning description to resolve
+/// @param surface the surface to resolve against
+/// @throws std::invalid_argument if the directions cannot be matched or the
+///         surface is unsupported
+/// @throws std::domain_error if any description is fully specified instead
+///         of deferred
+/// @return the resolved axes, in canonical direction order
+std::vector<std::unique_ptr<IAxis>> resolveAxes(
```
asalzburger: Here I like the naming.

Core/include/Acts/Utilities/AxisFactory.hpp
```diff
@@ -0,0 +1,310 @@
...
+
+namespace Acts {
+
+/// @brief Per-axis resolution input supplied by the consumer of a deferred
+/// axis description.
+///
+/// A deferred @c AxisFactory only describes the binning structure of an axis;
+/// the range and the boundary type are determined by the consumer, e.g. from
+/// the bounds of the surface the axis is attached to.
+struct AxisResolution {
```
asalzburger: ```AxisResolution``` can be mistaken, as of "what resolution does this axis have ... not a good question, but one might fall into that".

Actually what you have here is a ```AxisBoundRange``` - a range that has been bound typed.

Core/include/Acts/Utilities/AxisSpec.hpp
```diff
@@ -0,0 +1,310 @@
+// This file is part of the ACTS project.
+//
```
asalzburger, line 2: I like the concept it will replace what I tried to do with the AxisHelper factory ...

Core/include/Acts/Utilities/BinningData.hpp
```diff
@@ -187,7 +187,9 @@ class BinningData {
   ///
   /// @param dpAxis is the ProtoAxis object
   ///
-  explicit BinningData(const DirectedProtoAxis& dpAxis)
```
asalzburger, line 190: Since we do get a warning from doxygen, we might have to add the ```@deprecated``` in the comments.

Core/src/Surfaces/SurfaceAxisResolution.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+std::string unsupportedSurfaceMessage(const Surface& surface) {
+  std::stringstream ss;
+  ss << "Axis resolution is not supported for this surface:\n"
+     << surface.toStream(GeometryContext::dangerouslyDefaultConstruct());
+  return ss.str();
+}
+
+AxisResolution resolveCylinder(const CylinderBounds& cBounds,
+                               AxisDirection aDir) {
```
asalzburger: Shall we enforce that cylinders need to have rPhi local coordinates ? (here it seems you allow phi as well).

Core/src/Surfaces/SurfaceAxisResolution.cpp
```diff
@@ -0,0 +1,183 @@
...
+  switch (aDir) {
+    case AxisDirection::AxisRPhi:
+      return {r * (avgPhi - halfPhi), r * (avgPhi + halfPhi), phiBoundaryType};
+    case AxisDirection::AxisPhi:
+      return {avgPhi - halfPhi, avgPhi + halfPhi, phiBoundaryType};
+    case AxisDirection::AxisZ:
+      return {-hz, hz, AxisBoundaryType::Bound};
+    default:
+      throw std::invalid_argument(
+          "Cylinder axis resolution must be along rphi, phi or z");
```
asalzburger: I would restrict to (rphi, z)

Core/src/Surfaces/SurfaceAxisResolution.cpp
```diff
@@ -0,0 +1,183 @@
...
+      throw std::invalid_argument(
+          "Trapezoid axis resolution must be along x or y");
+  }
+}
+
+}  // namespace
+
+Acts::AxisResolution Acts::surfaceAxisResolution(const Surface& surface,
+                                                 AxisDirection aDir) {
+  if (auto b = dynamic_cast<const CylinderBounds*>(&surface.bounds());
```
asalzburger: The bounds have an enum to avoid `dynamic_cast` operations.

Tests/UnitTests/Core/Utilities/ProfileEfficiencyTests.cpp
```diff
@@ -8,17 +8,18 @@
...
-#include "Acts/Utilities/ProtoAxis.hpp"
 
 using namespace Acts;
 using namespace Acts::Experimental;
 
 BOOST_AUTO_TEST_SUITE(ProfileEfficiencySuite)
 
 BOOST_AUTO_TEST_CASE(ProfileHistogram_BasicFill) {
-  ProtoAxis protoAxis(AxisBoundaryType::Bound, 0.0, 10.0, 10);
-  auto xAxis = BoostVariableAxis(protoAxis.getAxis().getBinEdges(), "x");
```
asalzburger, line 21: One other argument to sort this out earlier than later, the ProtoAxis usage is growing - let's fix this once for all.

asalzburger (changes requested):
Here are my comments, also I think you won't get around the doxygen ```@deprecated```.

Core/include/Acts/Surfaces/AnnulusBounds.hpp
```diff
@@ -115,8 +115,8 @@ class AnnulusBounds : public DiscBounds {
   /// Returns true for full phi coverage
   /// @return True if the annulus covers the full azimuthal range, false otherwise
   bool coversFullAzimuth() const final {
-    return (std::abs((get(eMinPhiRel) - get(eMaxPhiRel)) - std::numbers::pi) <
-            s_onSurfaceTolerance);
+    return std::abs((get(eMaxPhiRel) - get(eMinPhiRel)) -
+                    2 * std::numbers::pi) < s_fullAzimuthTolerance;
```
asalzburger, line 119: Makes much more sense than the `onSurfaceTolerance`

Core/src/Surfaces/ConeBounds.cpp
```diff
@@ -53,7 +53,7 @@ void ConeBounds::checkConsistency() noexcept(false) {
       std::abs(get(eMinZ) - get(eMaxZ)) < s_epsilon) {
     throw std::invalid_argument("ConeBounds: invalid z range setup.");
   }
-  if (get(eHalfPhiSector) < 0. || std::abs(eHalfPhiSector) > std::numbers::pi) {
+  if (get(eHalfPhiSector) < 0. || get(eHalfPhiSector) > std::numbers::pi) {
```
asalzburger, line 56: Good catch.

asalzburger (dismissed):
Looks good now.

There is still a CodeQuality error regarding null pointer.

asalzburger (approved):
Good to go.

PR #5741, opened 2026-07-25
ci: Authenticate GitHub dependency fetches in PyPI wheel builds

paulgessinger:
Two token-authentication fixes for the PyPI wheel job, both making dependency fetches use the available `GITHUB_TOKEN` instead of falling back to anonymous, rate-limited calls.

## 1. Forward `GITHUB_TOKEN` into Linux wheel builds

`CI/cibuildwheel.sh` forwarded the token via `CIBW_ENVIRONMENT`:

```sh
export CIBW_ENVIRONMENT="GITHUB_TOKEN=${GITHUB_TOKEN:-}"
export CIBW_ENVIRONMENT_PASS="CI"
...
export CIBW_ENVIRONMENT_LINUX="CMAKE_PREFIX_PATH=... LD_LIBRARY_PATH=..."
export CIBW_ENVIRONMENT_MACOS="CMAKE_PREFIX_PATH=... MACOSX_DEPLOYMENT_TARGET=26.0"
```

cibuildwheel's `CIBW_ENVIRONMENT_LINUX` / `CIBW_ENVIRONMENT_MACOS` **replace** `CIBW_ENVIRONMENT` rather than merging, so the `GITHUB_TOKEN=...` entry was discarded.

- **macOS** builds on the host and inherits `GITHUB_TOKEN` from the shell anyway → unaffected.
- **Linux** builds inside the manylinux container, where only `CIBW_ENVIRONMENT_PASS` variables cross the boundary → token ends up empty.

In the nightly logs the Linux `setup.sh` skips the "Setting GHCR credentials" step present on macOS, and `select_lockfile.py` prints empty request headers for its `api.github.com` call. Every GitHub/GHCR dependency fetch in the Linux wheel build (spack buildcache pulls, lockfile release-API query, release-asset download) therefore ran unauthenticated and subject to anonymous rate limits.

Fix: forward the token through `CIBW_ENVIRONMENT_PASS`, which is not overridden per-platform.

## 2. Always authenticate lockfile fetches

`CI/dependencies/select_lockfile.py` dropped the `Authorization` header on the **final** retry (`remaining_retries.get() > 0`), so the last attempt fell back to an anonymous request. Anonymous calls hit GitHub's much lower unauthenticated rate limit, making a rate-limit failure *more* likely on exactly the attempt meant to recover from it.

Fix: always attach the token when one is available. Also stop printing the raw `headers` dict, which would leak the bearer token into CI logs now that auth is attached on every attempt.

Smoke-tested locally: both the `api.github.com` release query and the release-asset download report `Authenticated: True` and the lockfile downloads successfully.

## Not in scope

- `pypi.yml` sets `permissions: {}` on the wheel job; GHCR reads currently succeed with the scopeless token, but granting `contents: read` + `packages: read` would be more explicit.
- The current nightly failures (`contourpy` needing Python ≥3.11 in the cp310 test venv; a macOS `delocate-wheel` error) are unrelated to token auth.


CI/cibuildwheel.sh
```diff
@@ -22,8 +22,14 @@ export CIBW_SKIP="*-musllinux* *-manylinux_i686"
...
 export CIBW_BEFORE_ALL_MACOS="brew install ninja ccache && ${SETUP_CMD}"
-export CIBW_ENVIRONMENT="GITHUB_TOKEN=${GITHUB_TOKEN:-}"
-export CIBW_ENVIRONMENT_PASS="CI"
+# GITHUB_TOKEN must be forwarded via CIBW_ENVIRONMENT_PASS, not CIBW_ENVIRONMENT:
+# the platform-specific CIBW_ENVIRONMENT_{LINUX,MACOS} below fully *replace*
+# CIBW_ENVIRONMENT (they are not merged), so a token set there would be dropped.
+# On macOS the build runs on the host and inherits GITHUB_TOKEN from the shell
+# regardless, but Linux builds run in a container where only *_PASS variables
+# cross the boundary, so without this the token is empty inside the container and
+# all GitHub/GHCR dependency fetches run unauthenticated.
```
benjaminhuth:
```suggestion
```
I think we can drop this very lengthy comment, I anyways don't really know anymore why I did not put it in `CIBW_ENVIRONMENT_PASS` which seems to be the natural choice.

benjaminhuth (dismissed):
Looks good, for my taste we can drop the very lengthy comment though.

benjaminhuth (approved):

PR #5742, opened 2026-07-25
ci: Fix macOS PyPI wheel delocate failure (Arrow libssl duplicate)

paulgessinger:
**Stacked on #5719.** Its commits appear in this diff until it merges; review this PR's own commits only, and it will be rebased onto `main` once #5719 lands. The remaining merge conflicts are all in `requirements.txt` lockfiles owned by #5719, not by this PR.

This PR's own contribution is two files: `CI/repair_wheel_macos.sh` and the three-line change wiring it up in `CI/cibuildwheel.sh`.

## Context

The PyPI nightly has two independent failures, both introduced by #5659:

1. **Linux** — cp310 test-requirements unresolvable. **Fixed by #5719** (drops Python 3.10, universal lockfiles).
2. **macOS** — `delocate-wheel` fails with `Already planning to copy library with same basename as: libssl.3.dylib`. Not addressed anywhere else; this PR targets it.

Failure #2 is Python-version-independent (it's about the wheel's linked dylibs), so #5719 alone leaves the macOS wheel red.

## Root cause

Confirmed from CI diagnostics (`delocate-listdeps --all --depending` on the un-repaired wheel). Enabling `ACTS_BUILD_PLUGIN_ARROW` + `ACTS_ARROW_ISOLATED` for wheels makes the macOS wheel link openssl **twice, from two different prefixes**:

| Consumer | openssl it links |
|---|---|
| `libActsPluginArrow.dylib` (static Arrow) | spack's — `.../spack/opt/spack/darwin-aarch64/openssl-3.6.1-u37gbox7.../lib/libssl.3.dylib` |
| `libthrift.0.22.0.dylib` (spack, an Arrow dependency) | the python.org framework's — `/Library/Frameworks/Python.framework/Versions/3.14/lib/libssl.3.dylib` |

`libthrift` is the *only* thing pulling in the framework copy. delocate cannot vendor two libraries sharing a basename, so it aborts.

Note this is **not** a spack lockfile problem: the macOS lockfile contains a single openssl. The second copy comes from spack's `thrift` having been linked against the framework openssl instead of its own.

Upgrading delocate does not help (0.13.0, latest, still fails) — these are genuinely two different files, not a symlink pair.

## The fix

`CI/repair_wheel_macos.sh`, wired up as `CIBW_REPAIR_WHEEL_COMMAND_MACOS`. Before delocating, it repoints any spack library referencing the framework openssl at the spack openssl, collapsing the graph to a single copy. Both are openssl 3.x, and ACTS' Parquet usage does not touch TLS.

It only mutates the spack store on the ephemeral CI runner; delocate re-signs whatever it vendors, so the invalidated store-lib signatures don't matter, and the wheel's tests then run against the self-contained wheel.

The script is deliberately conservative:

- derives the spack root from whatever spack path the wheel actually links, rather than hard-coding a layout;
- **no-ops** when the wheel doesn't link spack at all (e.g. Arrow disabled);
- **no-ops** when nothing references the framework openssl — so it self-disables once the proper fix lands, instead of rotting into a permanent hack;
- fails loudly if `install_name_tool` fails or the target openssl is missing, rather than masking it into a confusing delocate error;
- dumps the dependency tree if delocate still fails.

### This is a workaround

The proper fix belongs in **ci-dependencies**: build `thrift` against spack's own openssl, or drop openssl from that build entirely (Parquet needs no TLS). This unblocks the wheels until then, and reverts to a no-op automatically once that happens.

## Validation

End-to-end macOS wheel build on `macos-26` (cp311), confirming the whole chain:

```
spack root: /Users/runner/work/acts/acts/spack
  .../thrift-0.22.0-.../lib/libthrift.0.22.0.dylib:
    /Library/Frameworks/Python.framework/Versions/3.14/lib/libssl.3.dylib
      -> .../openssl-3.6.1-u37gbox7.../lib/libssl.3.dylib
  (same for libcrypto.3.dylib)
rewrote 2 framework openssl reference(s)
```

delocate then vendored exactly one copy of each:

```
Copying library .../openssl-3.6.1-u37gbox7.../lib/libssl.3.dylib    -> acts/.dylibs/libssl.3.dylib
Copying library .../openssl-3.6.1-u37gbox7.../lib/libcrypto.3.dylib -> acts/.dylibs/libcrypto.3.dylib
```

and the wheel's own test suite passed against the repaired wheel: `12 passed, 373 deselected`. That last point matters — it confirms the rewritten linkage actually loads at import time, not just that delocate stopped complaining.

The script's branches were additionally exercised locally against the real linkage topology captured from the CI artifact, with mocked `otool` / `install_name_tool` / `delocate-wheel`: it reproduces the original `DelocationError` without the fix, and no-ops correctly both when the wheel has no spack references and when `thrift` already points at spack's openssl.

The temporary push-triggered diagnostics workflow used to iterate on this has been removed.

## Before merging

- [x] End-to-end macOS wheel build green
- [x] Delete the temporary diagnostics workflow
- [x] Rebase onto `main` once #5719 lands


.github/workflows/update-pip-requirements.yml
```diff
@@ -28,6 +28,12 @@ jobs:
       # persisted credentials are required for the push step below
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
+      # Lowest Python version the lockfiles must install on, taken from
+      # `requires-python` in pyproject.toml so it is defined in exactly one place
+      - name: Determine supported Python floor
+        id: floor
+        run: echo "version=$(CI/supported_python_versions.py --floor)" >> $GITHUB_OUTPUT
+
```
benjaminhuth:
```suggestion
```

Duplicated with the lines below, can be removed

benjaminhuth (approved):

PR #5743, opened 2026-07-25
perf: cut peak compile memory of the heaviest translation units

paulgessinger:
## What

Reduces the **peak compiler memory** of several of the heaviest translation units in the build, with **no change to runtime behaviour or performance**. Motivation is CI/runner memory pressure: a handful of TUs dominate peak RSS, and each can OOM a constrained builder.

TUs were profiled with [`cmakeperf`](https://github.com/paulgessinger/cmakeperf) (per-process peak RSS) plus `clang -ftime-trace` to attribute the cost. The recurring pattern in every fix below: a heavy Eigen/template instantiation is moved behind a symbol that is compiled **once** (type-erasure, explicit instantiation, or a fitter-independent out-of-line function) instead of being re-instantiated in every caller.

## Changes

**1. `IMultiAxis::create` (3-axis) — `IMultiAxis.cpp` 1174 → 462 MB**
The 3-axis factory instantiates `MultiAxis<A1,A2,A3>` for all 6³ = 216 concrete axis-type combinations in one TU. Distributed across three per-boundary-type TUs via a declared-only helper.

**2. `calculateUnbiasedParametersCovariance` — type-erased overload**
This Eigen-heavy helper (~780 MB) was re-instantiated in every caller. Added a non-template overload over the type-erased `AnyConstTrackStateProxy`, compiled once in the core library; callers wrap their concrete proxy. The template is kept but `[[deprecated]]`.
- `RootTrackStatesWriter.cpp` 1467 → 768 MB
- `TrackHelpersTests.cpp` 1553 → 881 MB

**3. Fitter-independent alignment algebra — `AlignmentEngine.cpp`**
`Alignment<fitter_t>`'s per-track chi² algebra is heavy Eigen but does not depend on the fitter. `finaliseTrackAlignState` is de-`inline`d and `calculateAlignmentParameters(vector, result)`'s body extracted to `detail::solveAlignmentParameters`.
- `AlignmentAlgorithmFunction.cpp` 1392 → 1108 MB (−20%)
- `AlignmentTests.cpp` 1623 → 1299 MB (−20%)

**4. `Eigen::Quaternion::setFromTwoVectors` (JacobiSVD) — out-of-line**
`setFromTwoVectors` instantiates a fixed-size `JacobiSVD` costing ~0.5 GB per TU, paid in every TU building a 3D `Frustum` and in `GenericCuboidVolumeBounds`. Introduced `Acts::detail::quaternionFromTwoVectors`, explicitly instantiated for `float`/`double` (split per-scalar so neither TU is a new outlier).
- `GenericCuboidVolumeBounds.cpp` 944 → 505 MB (−46%)
- `BoundingBoxTest.cpp` 1204 → 786 MB (−35%)
- `FrustumTest.cpp` → 415 MB

## Correctness

Generated code is unchanged in every case. Verified by building the affected libraries and running the relevant unit tests (MultiAxis, TrackHelpers, Alignment kernel, BoundingBox, Frustum, GenericCuboidVolumeBounds) — all pass, plus an added arity-3 `IMultiAxis::create` smoke check.


Core/src/Utilities/detail/CreateMultiAxis3.hpp
benjaminhuth, line 1: Hmm shouldn't this go into `Core/include/detail/Utilities/CreateMultiAxis3.hpp?

Core/src/Utilities/CreateMultiAxis3Bound.cpp
benjaminhuth, line 1: Don't want to be to pedantic, but so far we never had undescores in filenames I think. Don't care too much though

benjaminhuth (dismissed):
Looks good, to minor comments

Core/src/Utilities/detail/CreateMultiAxis3.hpp
paulgessinger: That means it's installed, I don't think we need/want it installed.

Core/src/Utilities/detail/QuaternionFromTwoVectorsDouble.cpp
```diff
@@ -0,0 +1,27 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "QuaternionFromTwoVectors.ipp"
+
+// GCC (>=12) emits a spurious -Wmaybe-uninitialized from inside Eigen's SVD
+// code used by setFromTwoVectors; suppress it for this translation unit.
+#if defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 12
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
```
andiwand, line 14: could be a macro similar to what we have for deprecations but I guess we also don't want to make these ignores too prominent

Core/include/Acts/Utilities/detail/QuaternionFromTwoVectors.hpp
```diff
@@ -0,0 +1,34 @@
...
+/// means the SVD is instantiated once in the Acts core library instead of in
+/// every translation unit that needs it.
+///
+/// @tparam T the scalar type (float or double)
+/// @param a the source vector
+/// @param b the target vector
+/// @return the quaternion rotating @p a onto @p b
+template <typename T>
+Eigen::Quaternion<T> quaternionFromTwoVectors(const Eigen::Matrix<T, 3, 1>& a,
+                                              const Eigen::Matrix<T, 3, 1>& b);
```
andiwand: I guess this could also be simplified by overload if we only ever do `float` and `double`. also I wonder if we really need `float`

Core/include/Acts/Utilities/Frustum.ipp
```diff
@@ -50,7 +51,7 @@ Acts::Frustum<value_t, DIM, SIDES>::Frustum(const VertexType& origin,
   const VertexType lup = VertexType::UnitX();
 
   transform_type transform;
-  transform = (Eigen::Quaternion<value_type>().setFromTwoVectors(ldir, dir));
+  transform = detail::quaternionFromTwoVectors<value_type>(ldir, dir);
```
andiwand: outside this pr but like mentioned above I wonder if we want to support an open value parameter as we removed it in practically all the other cases in ACTS

Core/src/Utilities/CMakeLists.txt
```diff
@@ -13,5 +14,10 @@ target_sources(
         TransformComparator.cpp
         Histogram.cpp
         IMultiAxis.cpp
+        QuaternionFromTwoVectorsDouble.cpp
+        QuaternionFromTwoVectorsFloat.cpp
+        CreateMultiAxis3Open.cpp
+        CreateMultiAxis3Bound.cpp
+        CreateMultiAxis3Closed.cpp
```
andiwand: we could hav a `src/Utilities/detail` to collect these and the headers

Core/src/Utilities/IMultiAxis.cpp
```diff
@@ -34,15 +36,14 @@ std::unique_ptr<IMultiAxis2D> IMultiAxis::create(const IAxis& axis1,
 std::unique_ptr<IMultiAxis3D> IMultiAxis::create(const IAxis& axis1,
                                                  const IAxis& axis2,
                                                  const IAxis& axis3) {
```
andiwand, line 38: I didn't expect this to blow up so significantly but kind of makes sense since `MultiAxis` is not too small. my original thought was that we might need 3D for volume grids in the future

alternatively we could also remove the 3 axis factory or replace it by a dynamical type that just holds `std::vector<std::unique_ptr<const IAxis>>`

Core/src/Utilities/TrackHelpersUnbiased.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+std::pair<BoundVector, BoundMatrix> calculateUnbiasedParametersCovariance(
+    const AnyConstTrackStateProxy& trackState) {
+  // Delegates to the deprecated templated implementation. This is the one
+  // place it is meant to be instantiated, so silence the deprecation warning
+  // for the internal call only.
+  ACTS_PUSH_IGNORE_DEPRECATED()
+  return calculateUnbiasedParametersCovariance<AnyConstTrackStateProxy>(
+      trackState);
+  ACTS_POP_IGNORE_DEPRECATED()
```
andiwand: alternatively we could rotate everything in here already. implementation goes here and the templated version converts the track state into an any track state and calls the new helper

Core/include/Acts/Utilities/detail/QuaternionFromTwoVectors.hpp
paulgessinger: The way the frustum code is written at the moment, `float` is instantiated. But you're right that this is probably not needed.

Core/src/Utilities/CMakeLists.txt
paulgessinger: yeah why not

Core/src/Utilities/TrackHelpersUnbiased.cpp
paulgessinger: Yeah that also works.

Core/src/Utilities/IMultiAxis.cpp
paulgessinger: But the combinatorics have to come in somewhere, don't they? I think with this approach we can keep the 3D combinations for now, but happy to revisit in the future.

Core/src/Utilities/detail/QuaternionFromTwoVectorsDouble.cpp
paulgessinger: Can't easily use that because it produces warnings about the warning not existing on clang (i.e. the warning is not portable).

andiwand (approved):
thanks for optimizing this!

PR #5754, opened 2026-07-26
ci: public API surface report, break detection, and labeling

paulgessinger:
Tracks the ACTS **source-level** public API surface (ABI is out of scope) so it stops growing silently, and classifies each PR's effect on it.

### Scope
Core, Fatras, Alignment, and **all plugins** (including the Detray plugin). Examples and the top-level Detray/Traccc build-integration folders are excluded. Only documented, non-`detail`/`Experimental` entities in `Acts*::` count — and the docs job already enforces full documentation (`WARN_IF_UNDOCUMENTED` + `WARN_NO_PARAMDOC` + `FAIL_ON_WARNINGS`), so "documented" == the real public surface.

### What it measures
Types (class/struct/union/enum), free functions, aliases, variables, enums, concepts, **and public data members and method signatures** of documented types. Uses the docs job's pinned **Doxygen 1.15.0** (parses syntactically — no dependency headers or build needed).

Current snapshot (local): ~1680 documented names + ~4500 public methods + ~1660 public data members.

### What it does on a PR
- **`api_surface` job** (in `checks.yml`): measures this ref, diffs against the base, uploads a JSON + Markdown report to the job summary/artifacts. **Report-only, never fails.**
- **Signature-aware break detection**: removals, renames, removing an argument, adding a **non-defaulted** argument, retyping a parameter, changing a return type, and removing/retyping a public data member are **breaking**; adding a **defaulted** argument or a new member is not.
- **`api-surface-label.yml`** (`pull_request_target` labeler): applies bot-owned labels — `Public API` (additions) and `API breaking` (source breaks) — and posts a sticky comment listing the changed symbols. Only parses PR header text with a trusted Doxygen; never builds/executes PR code. Takes effect once on a base branch.

### Deprecation-docs gate
A C++ `[[deprecated]]` attribute only warns at compile time — it never reaches the rendered documentation. Doxygen's separate `@deprecated` command is what surfaces the deprecation in the published docs (the "Deprecated List" and a note on the page). The two are independent, so a symbol can be deprecated in code yet look current in the docs.

`CI/public_api/check_deprecated_docs.py` is a **dependency-free regex/heuristic** gate (no libclang, no compiler — deliberately, per discussion) that requires every `[[deprecated]]` declaration to carry a paired `@deprecated` / `\deprecated` command in its Doxygen block. It reads the same comment text Doxygen does. Wired into the `api_surface` job (as an actual **gate**, unlike the report), with a self-test. The 29 pre-existing unpaired sites are fixed in this PR, so it starts at zero violations.

### Files
- `CI/public_api/Doxyfile` — surface-metric config mirroring the versioning policy
- `CI/public_api/public_api_surface.py` — measure a ref → JSON inventory (+ signatures/fields)
- `CI/public_api/public_api_diff.py` — classify base→head as additions vs. breaking
- `CI/public_api/test_public_api_surface.py` — self-test for the surface tooling (Doxygen e2e)
- `CI/public_api/check_deprecated_docs.py` — gate: `[[deprecated]]` must have a `@deprecated` doc command
- `CI/public_api/test_check_deprecated_docs.py` — self-test for the deprecation-docs heuristic
- `.github/workflows/checks.yml` — the report/diff job + the deprecation-docs gate
- `.github/workflows/api-surface-label.yml` — the labeler


paulgessinger:
> [!NOTE]
> 🔬 **Synthetic preview — not a real analysis of this PR.** Posted manually to
> show how the `api-surface-label.yml` sticky comment renders. The data below is
> fabricated to exercise every classification bucket. The real bot uses the
> `<!-- api-surface-diff -->` marker and updates one comment in place; this
> preview uses a separate marker so it never collides. Safe to delete.

---

### Example A — a PR that both adds and breaks public API

In this case the labeler would apply the **`Public API`** and **`API breaking`** labels.

## Public API surface diff

**+8 added**, **6 breaking**.

### ⚠️ Breaking API changes (source-level)
<details><summary>Removed types / aliases / enums / variables (1)</summary>

- `type Acts::LegacySeedFinder`

</details>
<details><summary>Removed or changed call signatures (2)</summary>

- `Acts::Propagator::propagate(const Options&) const`
- `Acts::estimateTrackParamsFromSeed(const Vector3&, const Vector3&, const Vector3&, const Vector3&)`

</details>
<details><summary>Return-type changes (1)</summary>

- `Acts::Surface::center(const GeometryContext&) const: Vector3 -> Vector3D`

</details>
<details><summary>Removed public data members (1)</summary>

- `Acts::Navigator::Config::tolerance`

</details>
<details><summary>Retyped public data members (1)</summary>

- `Acts::Navigator::Config::maxSteps: unsigned int -> std::size_t`

</details>

### ➕ Added public API
<details><summary>New types / aliases / enums / variables / concepts (4)</summary>

- `aliases Acts::BoundMatrix`
- `concept Acts::TrackContainerBackend`
- `enums Acts::NavigationDirection`
- `type Acts::CylinderSurface`

</details>
<details><summary>New call signatures (incl. defaulted-arg overloads) (3)</summary>

- `Acts::CylinderSurface::radius() const`
- `Acts::CylinderSurface::radius(const GeometryContext&) const`
- `Acts::Propagator::propagate(const Options&, const Context&) const`

</details>
<details><summary>New public data members (1)</summary>

- `Acts::Navigator::Config::stepSizeCutOff`

</details>

---

### Example B — a PR with no public-API effect

Here no labels are applied and the comment collapses to a single line:

## Public API surface diff

No change to the public API surface. ✅

CI/public_api/Doxyfile
```diff
@@ -0,0 +1,42 @@
...
+#
+# This is NOT the documentation build (see docs/Doxyfile.in). It mirrors the
+# public/private policy from docs/pages/versioning.md so that the surface metric
+# matches what is considered public API:
+#   * EXTRACT_ALL = NO        -> only *documented* entities count
+#   * detail / Experimental excluded by path and by symbol
+# Only XML is generated; CI/public_api/public_api_surface.py parses it into a report.
+#
+# Run from the repository root:
+#   OUTPUT_DIRECTORY=<dir> doxygen CI/public_api/Doxyfile
```
andiwand, line 11: this could go out of sync in the future but I guess there is no good way of prevent that? or does Doxyfile somehow allow to include another config and override properties or provide some sharing?

andiwand (dismissed):
Looks pretty neat - looking forward to see it in action

CI/public_api/Doxyfile
paulgessinger: Right, I don't think there's a way to enforce this.

andiwand (approved):

PR #5760, opened 2026-07-27
ci: Group Dependabot updates and close pip-requirements gaps

paulgessinger:
## Dependabot grouping

Last week's run opened seven separate PRs (#5744–#5751), one per action/hook — each with its own CI run and merge-queue slot. This adds `groups` to both ecosystems so bumps batch into a single PR.

Majors get their own group rather than going in the same PR, so a breaking bump (last week's `actions/labeler 6.2.0 → 7.0.0`, for example) can't block the routine SHA refreshes behind it. At most two PRs per ecosystem instead of seven.

The 7-day `cooldown` was already in place from #5529 and is unchanged.

## `update-pip-requirements.yml`: latent footgun

`rm "$output_txt"` ran *before* the input-file check, and the missing-input branch only `echo`ed. So adding a folder without a `requirements.in` / `pyproject.toml` to `folder_list` deleted its `requirements.txt`, never regenerated it, and the job still went green.

The removal now happens only once an input is confirmed, and an unresolvable folder `exit 1`s. Removing the lockfile before compiling is still deliberate — it forces a fresh resolution instead of uv preferring the currently pinned versions — just no longer unconditional.

## Not included

**`Examples/Scripts/GsfDebugger`** was dropped from this PR. It has an unmanaged `requirements.txt` with stale pins and no `requirements.in`, but bringing it under this workflow means accepting an unreviewed major-version jump (`pandas 2.2.2 → 3.0.5`, plus `matplotlib`/`numpy` bumps) on the next scheduled run, which doesn't belong in a CI-hygiene PR. Left for a follow-up.

**`Detray/python/detray`** also has an unmanaged `requirements.txt`, but it's not a real gap: it declares `dependencies = []`, the file is a header with zero packages, and `uv.lock` holds only the virtual root. Adding it would also break the loop — `requires-python = ">=3.13"` against the 3.11 floor the workflow compiles at.

**Migrating requirements.txt to Dependabot** was considered and rejected. Dependabot's `uv` ecosystem does handle `uv pip compile` output (not just `uv.lock`) and preserves `--universal`/`--python-version` from the header, so it would work mechanically. But `directories` + `groups` produces [a PR per directory](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference) by default — six-plus PRs replacing today's one — and Dependabot would read the Python version back out of each file's header instead of deriving it from `requires-python`, so bumping the floor would silently leave the lockfiles compiling against the old one.

Inline PEP 723 script deps stay manual either way; [Dependabot doesn't support them](https://github.com/dependabot/dependabot-core/issues/11946). Real exposure there is four pins total: `gcovr==8.6`, `particle==0.24.0`, `histcmp==0.10.0`, `change-wheel-version==0.6.0`.

## Testing

Dry-ran the workflow's compile loop against all six folders at the real floor (3.11) — all resolve. `prek run` clean, zizmor passes.

Grouping behaviour itself can only be confirmed by a live Dependabot run.


andiwand (dismissed):

.github/workflows/update-pip-requirements.yml
```diff
@@ -56,22 +56,29 @@ jobs:
             docs/old
             Python/Examples/tests
             Examples/Scripts
+            Examples/Scripts/GsfDebugger
```
AJPfleger: do we test/run the gsf debugger in the CI? I think skipped the ipdate because it could just make it unusable over time.

AJPfleger (commented):
the or does 2 unrelated things for 2 different tools.
- the dependabot schedule changes look fine.
- the requirement updates seem like they could work (didn't test them)
- gsf debugger addition might be not what we want

.github/workflows/update-pip-requirements.yml
paulgessinger:
```suggestion
```

paulgessinger:
@AJPfleger removed the GsfDebugger requirements

AJPfleger (approved):

PR #5761, opened 2026-07-27
build: simplify FindTensorRT to the runtime libraries we actually use

paulgessinger:
The vendored `FindTensorRT` module searched for four components (`nvinfer`, `nvinfer_plugin`, `nvonnxparser`, `nvparsers`), but the GNN plugin only ever links the first two. ACTS exclusively deserializes pre-built `.engine` files — building an engine from ONNX is an offline step performed outside of ACTS — so the graph parsers are never needed, and `nvparsers` no longer ships as of TensorRT 10.

More importantly, the module only listed `TensorRT_INCLUDE_DIR` in `REQUIRED_VARS`. Since `find_package(TensorRT REQUIRED)` is called without `COMPONENTS`, `TensorRT_FIND_REQUIRED_<comp>` was never set, so `HANDLE_COMPONENTS` treated every component as optional. A configure would therefore succeed with only `NvInfer.h` present and fail later at link time with a confusing error.

This rewrites the module against the two libraries actually in use (174 → 106 lines):

- Both libraries are in `REQUIRED_VARS`, so a missing `libnvinfer` fails at configure time.
- `TensorRT_LIBRARY_DIRS` is set only on success, derived from the `nvinfer` location. Previously it was assigned from whichever component was processed first regardless of whether that lookup succeeded, which could cache an empty string. It stays a cache variable because `ActsCreateSetup` substitutes it from the top-level scope while `find_package(TensorRT)` runs down in `Plugins/Gnn`.
- Added `PATH_SUFFIXES lib lib64`, and `HINTS` rather than `PATHS` so a user-supplied `TensorRT_ROOT` outranks the system paths.
- Dropped `TensorRT_LIBRARIES`, which was always empty (the `set()` inside the helper function had no `PARENT_SCOPE`) and is not consumed anywhere.
- `UNKNOWN IMPORTED` with `IMPORTED_LOCATION` rather than `INTERFACE IMPORTED`. Imported targets treat their interface includes as SYSTEM by default, so the TensorRT headers stay warning-suppressed.

No change is needed in `Plugins/Gnn/CMakeLists.txt` — `find_package(TensorRT REQUIRED)` keeps working and now actually enforces the libraries.

--- END COMMIT MESSAGE ---

### Verification

`build_gnn_tensorrt` is the only consumer, and it needs the container, so I validated the module standalone against a mock TensorRT tree:

| case | before | after |
|---|---|---|
| complete install | found, version + both targets | same |
| `libnvinfer` removed | **`-- Found TensorRT`, configure proceeds** | `Could NOT find TensorRT (missing: TensorRT_nvinfer_LIBRARY)` |
| no `NvInferVersion.h` | found, empty version | same (unchanged behaviour) |

Confirmed nothing in the tree references `TensorRT_LIBRARIES`, `trt::nvonnxparser`, `trt::nvparsers`, or `TENSORRT_LIBRARY_DIR`.

The real check is the `build_gnn_tensorrt` GitLab job — worth a look at whether the TensorRT layout in `ubuntu2404_tensorrt` is still picked up (a system deb install under `/usr/lib/x86_64-linux-gnu` resolves via the default paths regardless).


cmake/FindTensorRT.cmake
```diff
@@ -1,174 +1,106 @@
...
-            FATAL_ERROR
-            "Fail to find TensorRT, please set TensorRT_ROOT. Include path not found."
-        )
-    endif()
-    return()
-endif()
-set(TensorRT_INCLUDE_DIRS ${TensorRT_INCLUDE_DIR})
-
-# Extract version of tensorrt
 if(EXISTS "${TensorRT_INCLUDE_DIR}/NvInferVersion.h")
```
benjaminhuth, line 36: Maybe we should fail if this is not found?

benjaminhuth (approved):
LGTM

cmake/FindTensorRT.cmake
paulgessinger: I believe if the `else` branch is taken here, `TensorRT_VERSION` is unset and `find_package_handle_standard_args` sets `TensorRT_FOUND` will communicate the error depending on `REQUIRED` being there.

PR #5762, opened 2026-07-27
ci: exclude Detray, Traccc, Examples and docs from coverage reports

paulgessinger:
Trims what the coverage report and the SonarCloud coverage metric look at.

**`CI/test_coverage.py`** (custom gcovr report: the HTML linked from the PR summary and the `cov.xml` handed to Sonar) — added to `EXCLUDE_PATHS`:

- `Detray/`, `Traccc/`
- `Examples/`, `docs/` (`docs/examples/*.cpp` are documentation snippets that get compiled)

**`sonar-project.properties`**:

- `sonar.coverage.exclusions` += `Detray/**/*,Traccc/**/*,docs/**/*` (`Examples/**/*` was already there)
- `sonar.exclusions` += `Detray/tests/**/*,Traccc/tests/**/*`

Note that `Detray/` and `Traccc/` stay in the Sonar *analysis* — their issues and hotspots keep showing up, only their test trees are excluded. The `github-ci-coverage` preset still compiles Detray and Traccc with `--coverage`, this only drops them from the reports.

On the current `main` analysis Detray sits at 15.2% coverage and Traccc at 0.0%, so dropping them from `cov.xml` moves the project from 38.6% to roughly 46.5%.


stephenswat:
Is there a way we can still get SonarCloud on traccc and detray like this? It's certainly useful for catching issues sometimes.

paulgessinger:
@stephenswat Good point. And in fact we probably want to eventually also include the unit test coverage for both projects. I'm not sure how easy this is to pull off though.

stephenswat (approved):
Great!

PR #5763, opened 2026-07-27
feat: Python-based visualization with IVisualization3D

yannikaMatt:
First version of Python 2D plotting and adaptation of trackingGeometry.visualize to take custom function with ViewConfig as return type

--- END COMMIT MESSAGE ---

@benjaminhuth @asalzburger @paulgessinger @andiwand 

Includes a basic default function that determines the ViewConfig based on the GeometryObject


Examples/Scripts/Python/plot2D.py
```diff
@@ -0,0 +1,78 @@
...
+        ax.set_xlabel('x')
+        ax.set_ylabel('y')
+
+    
+    for n_face, face in enumerate(self.surfaces):
+        if polyArea(face) < 1e-8:
+            ax.plot([item[0] for item in face], [item[1] for item in face], color=(self.faceColors[n_face][0]/255, self.faceColors[n_face][1]/255, self.faceColors[n_face][2]/255, 0.5), lw=1)
+
+
+def plotTrack(self, projection, lineColor=None, linewidth=None, linestyle = None, ax=None):
```
yannikaMatt: Planning to combine plotTrack() with plot()

Core/include/Acts/Visualization/PyVisualization.hpp
```diff
@@ -0,0 +1,195 @@
...
+  using VertexType = Eigen::Matrix<ValueType, 3, 1>;
+
+  /// Type of a line
+  using LineType = std::pair<std::size_t, std::size_t>;
+
+  /// Constructor that allows to set scalor and precision
+  /// @param prec The output precision with std::setprecision
+  /// @param scale An (optional) scaling for the writing out
+  explicit PyVisualization(unsigned int prec = 4, double scale = 1., std::string projection = "rz")
+      : m_outputPrecision(prec), m_outputScalor(scale), m_projection(projection) {}
```
yannikaMatt: Projection could also happen on the Python side

Core/include/Acts/Geometry/TrackingVolume.hpp
```diff
@@ -522,9 +522,7 @@ class TrackingVolume : public Volume {
   /// @param portalViewConfig View configuration for portals
   /// @param sensitiveViewConfig View configuration for sensitive surfaces
   void visualize(IVisualization3D& helper, const GeometryContext& gctx,
-                 const ViewConfig& viewConfig,
-                 const ViewConfig& portalViewConfig,
-                 const ViewConfig& sensitiveViewConfig) const;
+                 const std::function<ViewConfig(const GeometryObject &)> coloringFunction) const;
```
paulgessinger:
```suggestion
                 const std::function<ViewConfig(const GeometryObject &)> viewConfigFactory) const;
```
?

Also consider typedef'ing the function type.

Core/include/Acts/Visualization/EventDataView3D.hpp
```diff
@@ -307,6 +311,62 @@ struct EventDataView3D {
...
+
+  /// Helper method to draw a track from AnyTrackProxy
+  ///
+  /// @param helper [in, out] The visualization helper
+  /// @param track The track to be drawn
+  /// @param itrack the track index in the container
+  /// @param gctx The geometry context for which it is drawn
+  /// @param lineconfig The visualization options for the lines to be drawn
+
+  static void drawTrack(
```
paulgessinger: Can t his go to the `.cpp` file?

Core/include/Acts/Visualization/EventDataView3D.hpp
```diff
@@ -307,6 +311,62 @@ struct EventDataView3D {
...
+      Acts::TrackContainer<Acts::ConstVectorTrackContainer,
+                         Acts::ConstVectorMultiTrajectory, std::shared_ptr>;
+
+  using TrackIndexType = TrackContainer::IndexType;
+
+  using TrackProxy = TrackContainer::TrackProxy;
+  using ConstTrackProxy = ConstTrackContainer::ConstTrackProxy;
+
+  using TrackStateProxy = TrackContainer::TrackStateProxy;
+  using ConstTrackStateProxy = ConstTrackContainer::ConstTrackStateProxy;
```
paulgessinger: I think `ConstTrackProxy` is the only type you need in the header if you push to a `.cpp` file. In addition, can you consider using `AnyTrackProxy` on the API boundary? That would cleanly allow this to work with any track type, rather than just this one.

Core/include/Acts/Visualization/PyVisualization.hpp
```diff
@@ -0,0 +1,177 @@
...
+#include <vector>
+#include <array>
+#include <iostream>
+
+namespace Acts {
+
+/// This helper produces output for Python visualization. Note that colors are not
+/// supported in this implementation.
+///
+class PyVisualization : public IVisualization3D {
```
paulgessinger: Suggest to move as much as possible to a `.cpp` file. I see why you name this `PyVisualization`, but in the end it *only* accumulates the draw commands and vertices, right? Maybe `VisualizationBuffer` or `RecordVisualization`

Core/include/Acts/Visualization/PyVisualization.hpp
```diff
@@ -0,0 +1,177 @@
...
+  /// Type of a vertex based on the value type
+  using VertexType = Eigen::Matrix<ValueType, 3, 1>;
+
+  /// Type of a line
+  using LineType = std::pair<std::size_t, std::size_t>;
+
+  /// Constructor that allows to set scalor and precision
+  /// @param prec The output precision with std::setprecision
+  /// @param scale An (optional) scaling for the writing out
+  explicit PyVisualization(unsigned int prec = 4, double scale = 1., std::string projection = "rz")
```
paulgessinger: Is `m_projection` used anywhere? I don't see it.

Core/include/Acts/Visualization/PyVisualization.hpp
```diff
@@ -0,0 +1,177 @@
...
+  /// The output parameters
+  unsigned int m_outputPrecision = 4;
+  double m_outputScalor = 1.;
+
+  std::vector<Object> m_objects;
+
+  Vector3 m_vertex;
+  std::string m_projection;
+  std::vector<std::vector<std::array<double,3>>> m_surfaces;
+  std::vector<std::vector<std::array<double,3>>> m_segments;
```
paulgessinger: Why not store `Vector3` here as well?

Core/include/Acts/Visualization/PyVisualization.hpp
```diff
@@ -0,0 +1,177 @@
...
+  /// @param prec The output precision with std::setprecision
+  /// @param scale An (optional) scaling for the writing out
+  explicit PyVisualization(unsigned int prec = 4, double scale = 1., std::string projection = "rz")
+      : m_outputPrecision(prec), m_outputScalor(scale){}
+
+  /// @copydoc Acts::IVisualization3D::vertex()
+  void vertex(const Vector3& vtx, Color color = s_defaultColor) {
+    /// vertex in 2D, either rz- or xy-projection
+    m_vertexcolor = color;
+    m_vertex = Vector3{vtx[0], vtx[1], vtx[2]};
```
paulgessinger: This can only store a single vertex? Why not multiple?

Core/include/Acts/Visualization/ViewConfig.hpp
```diff
@@ -162,4 +164,11 @@ static const ViewConfig s_viewGrid = {.color = {220, 0, 0}};
 /// View configuration for line visualization
 static const ViewConfig s_viewLine = {.color = {0, 0, 220}};
 
+
+ViewConfig defaultColoring(const GeometryObject&); 
+
+///std::map<int, Color> as argument later
+std::function<ViewConfig(const GeometryObject &)> makeDefaultColoringFunction();
```
paulgessinger: Or maybe `viewConfigFactory`?

Core/src/Geometry/TrackingVolume.cpp
```diff
@@ -602,31 +602,31 @@ void TrackingVolume::addSurface(std::shared_ptr<Surface> surface) {
...
     Volume::visualize(helper, gctx, viewConfig);
   }
 
-  if (sensitiveViewConfig.visible && !surfaces().empty()) {
+  if (!surfaces().empty()) {
     helper.object(volumeName() + "_sensitives");
     for (const auto& surface : surfaces()) {
-      surface.visualize(helper, gctx, sensitiveViewConfig);
+      auto viewConfig = coloringFunction(surface);
+      surface.visualize(helper, gctx, viewConfig);
```
paulgessinger, line 637: I suppose `Surface::visualize` could internally also call `coloringFunction(*this)` rather than `TrackinVolume` pre-resolving before calling? Same for the two other cases below.

Core/src/Geometry/Volume.cpp
```diff
@@ -162,6 +162,7 @@ void Volume::visualize(IVisualization3D& helper, const GeometryContext& gctx,
   auto bSurfaces =
       volumeBounds().orientedSurfaces(localToGlobalTransform(gctx));
   for (const auto& bs : bSurfaces) {
+
```
paulgessinger:
```suggestion
```

Examples/Scripts/Python/geometry.py
```diff
@@ -28,9 +35,10 @@ def runGeometry(
     decorators,
     outputDir: Path,
     events=1,
-    outputObj=True,
-    outputCsv=True,
-    outputSurfacesJson=True,
+    projection='xy',
+    outputPy=True,
+    outputCsv=False,
+    outputSurfacesJson=False,
```
paulgessinger: Probably revert to the defaults when you're done testing.

Examples/Scripts/Python/geometry.py
```diff
@@ -2,10 +2,17 @@
 
 import os
 import json
+import numpy as np
+import matplotlib.pyplot as plt
+import time
+
+from matplotlib.patches import Polygon, Circle
+from matplotlib.collections import PatchCollection, LineCollection
```
paulgessinger: Can you move the imports to inside the `if outputPy` so that the script stays usable if you don't have matplotlib installed?

Examples/Scripts/Python/plot2D.py
paulgessinger, line 1: I think this is only imported but never used?

Examples/Scripts/Python/truth_tracking_kalman.py
```diff
@@ -6,6 +6,16 @@
...
 
+import numpy as np
+import matplotlib.pyplot as plt
+import time
+
+from matplotlib.patches import Polygon, Circle
+from matplotlib.collections import PatchCollection, LineCollection
+
+from geometry import runGeometry
+import plot2D
```
paulgessinger: Same here, guard the imports by moving them where it's clear they'll be needed.

Examples/Scripts/Python/truth_tracking_kalman.py
```diff
@@ -52,17 +63,32 @@ def runTruthTrackingKalman(
...
+    eventStore =acts.examples.WhiteBoard(name=f"EventStore#{ievt}", level=acts.logging.INFO)
+    ialg = 0
+    ithread = 0
+
+    context = acts.examples.AlgorithmContext(ialg, ievt, eventStore, ithread)
+    vis = acts.PyVisualization()
+    trackingGeometry.visualize(
+                vis,
+                context.geoContext,
+            )
```
paulgessinger: This should be conditional I would say.

Examples/Scripts/Python/truth_tracking_kalman.py
```diff
@@ -118,6 +144,7 @@ def runTruthTrackingKalman(
         field,
         digiConfigFile=digiConfigFile,
         rnd=rnd,
+        logLevel=acts.logging.DEBUG
```
paulgessinger: Revert once you're done testing. Same below.

Examples/Scripts/Python/truth_tracking_kalman.py
```diff
@@ -172,8 +200,29 @@ def runTruthTrackingKalman(
...
+
+       
+        def execute(self, context):
+            tracks = self.tracks(context.eventStore)
+            for track in tracks:
+                acts.EventDataView3D.drawTrack(vis, track) # draw track not a free function
+
+            return acts.examples.ProcessCode.SUCCESS
+
+    s.addAlgorithm(TrackVisualizerAlg("TrackVisualizerAlg", acts.logging.DEBUG))
```
paulgessinger: I suppose this algorithm is something we'd want to "ship" in the Examples somewhere, rather than just having this here in the script.

Python/Core/src/Visualization.cpp
```diff
@@ -59,5 +64,52 @@ void addVisualization(py::module& m) {
...
+        // one element in the array
+        auto r = arr.mutable_unchecked<2>(); // direct access of elements without internal checking of dimensions
+
+        for (size_t i = 0; i < colors.size(); ++i) {
+            r(i, 0) = colors[i].rgb[0];
+            r(i, 1) = colors[i].rgb[1];
+            r(i, 2) = colors[i].rgb[2];
+        }
+        return arr;
+      }
```
paulgessinger: The pattern used in `spanToNumpy1d` in `EventData.cpp` is a helper function that does this a single time. I then used a lambda returning another lambda to make the registration compact.

Python/Examples/tests/test_geometry.py
```diff
@@ -261,3 +263,26 @@ def run_prop(geo, out_dir):
...
         orig["pathLength"], rebuilt_data["pathLength"], rtol=1e-5
     )
+
+@pytest.mark.skipif(not dd4hepEnabled, reason="DD4hep not set up")
+@pytest.mark.odd
+def test_geometry_python_visualization(tmp_path):
+    from geometry import runGeometry
+    from matplotlib.collections import PatchCollection
+
+    with getOpenDataDetector(gen3=True) as detector:
```
paulgessinger: I recently reduced the pytest suite's time by reducing the number of time the geometry is built using fixtures (https://github.com/acts-project/acts/pull/5764). Can you use the pattern here?

Testing/Temporary/LastTest.log
paulgessinger, line 1: Delete

Tests/UnitTests/Core/EventData/TrackVisualizationTests.cpp
paulgessinger, line 1: Could this be in any existing unit test file?

paulgessinger (commented):
Thanks for this! I left a few comments, there's also some testing leftovers in the diff that would have to be removed.

Core/include/Acts/Visualization/EventDataView3D.hpp
benjaminhuth: I think we started with `ConstTrackProxy` because `AnyTrackProxy` did not allow iteration over track states back then. This is now implemented @paulgessinger , right?

paulgessinger: Exactly!

Core/include/Acts/Geometry/TrackingVolume.hpp
yannikaMatt: Where would be the best place to put in the typedef?
I would have placed it in ViewConfig.hpp but I am not sure.

Core/include/Acts/Visualization/PyVisualization.hpp
yannikaMatt: I have actually not used this function anywhere so far which is why I have not given it too much thought yet. But I can certainly implement that.

Core/include/Acts/Visualization/PyVisualization.hpp
yannikaMatt: I found this implementation easier to use at the later stage where I need to access the cooridnates of e.g. a vertex in Python to determine the projections for plotting in matplotlib. 
If you feel like these should be `Vector3` I can try to change.

Core/src/Geometry/TrackingVolume.cpp
yannikaMatt: I have also thought about this when writing the code. This would lead to quite a lot of code changes also in other files while, at least for me, there are no apparent gains from this. I might be wrong there, so if there are strong arguments for `Surface::visualize` calling `coloringFunction(*this)` internally, this could be changed. Otherwise I would prefer to keep it this way.

Tests/UnitTests/Core/EventData/TrackVisualizationTests.cpp
yannikaMatt: Actually this is quite outdated, I was just trying something. I must have missed this when comitting, sorry.

Core/include/Acts/Geometry/TrackingVolume.hpp
yannikaMatt: Renaming and typedefing has been performed

Core/include/Acts/Visualization/EventDataView3D.hpp
yannikaMatt: Is changed.

Core/include/Acts/Visualization/PyVisualization.hpp
yannikaMatt: Was still a relict from an earlier version and is not needed. I removed it.

Core/include/Acts/Visualization/PyVisualization.hpp
yannikaMatt: Changed both to take `Vector3`. Worked better this time than when I first tried

Core/src/Geometry/Volume.cpp
yannikaMatt: Done

Examples/Scripts/Python/geometry.py
yannikaMatt: Should be back to defaults now.

Examples/Scripts/Python/geometry.py
yannikaMatt: Done

Examples/Scripts/Python/plot2D.py
yannikaMatt: My mistake. It was removed

Examples/Scripts/Python/truth_tracking_kalman.py
yannikaMatt: Done

Examples/Scripts/Python/truth_tracking_kalman.py
yannikaMatt: Good idea. I implemented that

Examples/Scripts/Python/truth_tracking_kalman.py
yannikaMatt: Done

Examples/Scripts/Python/truth_tracking_kalman.py
yannikaMatt: There is a new file `visualization.py` now which includes the algorithm as well as the contents of pref. `plot2D.py`, which does not exist anymore.

Python/Examples/tests/test_geometry.py
yannikaMatt: Done

Testing/Temporary/LastTest.log
yannikaMatt: Done

Python/Core/src/Visualization.cpp
yannikaMatt: This is a nice idea. I implemented a similar function

Core/include/Acts/Visualization/PyVisualization.hpp
yannikaMatt: Done and also renamed to `VisualizationBuffer`

Core/include/Acts/Visualization/EventDataView3D.hpp
yannikaMatt: Done

Examples/Configs/odd-digi-smearing-config-gen3.json:Zone.Identifier
benjaminhuth, line 1: This should not be committed

obj/geometry.mtl
benjaminhuth, line 1: This should not be commited as well

obj/geometry.obj
benjaminhuth, line 1: This should not be commited as well

Core/include/Acts/Geometry/TrackingGeometry.hpp
```diff
@@ -238,10 +238,12 @@ class TrackingGeometry {
   /// @param portalViewConfig View config for portals
   /// @param sensitiveViewConfig View configuration for sensitive surfaces
   void visualize(IVisualization3D& helper, const GeometryContext& gctx,
-                 const ViewConfig& viewConfig = s_viewVolume,
-                 const ViewConfig& portalViewConfig = s_viewPortal,
-                 const ViewConfig& sensitiveViewConfig = s_viewSensitive) const;
+                 const ViewConfigFunc& viewConfig) const;
 
+  /// overload method to handle default value
```
benjaminhuth: Hmm I would rather provide a default argument to the above function, then introduce a new overload here.

Core/include/Acts/Geometry/TrackingGeometry.hpp
```diff
@@ -238,10 +238,12 @@ class TrackingGeometry {
   /// @param portalViewConfig View config for portals
   /// @param sensitiveViewConfig View configuration for sensitive surfaces
   void visualize(IVisualization3D& helper, const GeometryContext& gctx,
-                 const ViewConfig& viewConfig = s_viewVolume,
-                 const ViewConfig& portalViewConfig = s_viewPortal,
-                 const ViewConfig& sensitiveViewConfig = s_viewSensitive) const;
+                 const ViewConfigFunc& viewConfig) const;
```
benjaminhuth: You could rename the parameter, as in `TrackingVolume.hpp` for consistency.
```suggestion
                 const ViewConfigFunc& viewConfigFactory) const;
```

Core/include/Acts/Visualization/VisualizationBuffer.hpp
```diff
@@ -0,0 +1,126 @@
...
+        return m_vertexcolors;
+  }
+ private:
+
+  /// The output parameters
+  unsigned int m_outputPrecision = 4;
+  double m_outputScalor = 1.;
+  
+  std::vector<Vector3> m_vertices;
+  std::string m_projection;
```
benjaminhuth:
```suggestion
```
This is unused now, right?

Core/include/Acts/Visualization/ViewConfig.hpp
```diff
@@ -162,4 +164,12 @@ static const ViewConfig s_viewGrid = {.color = {220, 0, 0}};
 /// View configuration for line visualization
 static const ViewConfig s_viewLine = {.color = {0, 0, 220}};
 
+
+ViewConfig defaultColoring(const GeometryObject&);
```
benjaminhuth: Maybe rename to `defaultGeometryColoring`, also add doxygen comments (these `///` comments to document what this is.

Core/include/Acts/Visualization/ViewConfig.hpp
```diff
@@ -162,4 +164,12 @@ static const ViewConfig s_viewGrid = {.color = {220, 0, 0}};
 /// View configuration for line visualization
 static const ViewConfig s_viewLine = {.color = {0, 0, 220}};
 
+
+ViewConfig defaultColoring(const GeometryObject&); 
+
+using ViewConfigFunc = std::function<ViewConfig(const GeometryObject&)>;
+///std::map<int, Color> as argument later
```
benjaminhuth: Better not merge such a raw comment. Either write a bit more with a doxygen `@todo` tag, or just leave remove the comment.

Core/include/Acts/Visualization/ViewConfig.hpp
```diff
@@ -162,4 +164,12 @@ static const ViewConfig s_viewGrid = {.color = {220, 0, 0}};
 /// View configuration for line visualization
 static const ViewConfig s_viewLine = {.color = {0, 0, 220}};
 
+
+ViewConfig defaultColoring(const GeometryObject&); 
+
+using ViewConfigFunc = std::function<ViewConfig(const GeometryObject&)>;
```
benjaminhuth, line 170: Also here we need short `///` documentation comments

Examples/Scripts/Python/geometry.py
```diff
@@ -2,10 +2,10 @@
 
 import os
 import json
-from pathlib import Path
 
 import acts
 import acts.examples
+
```
benjaminhuth:
```suggestion
```

Examples/Scripts/Python/geometry.py
```diff
@@ -28,11 +28,11 @@ def runGeometry(
     decorators,
     outputDir: Path,
     events=1,
-    outputObj=True,
```
benjaminhuth, line 31: Hmm it seems like you remove the `outputObj` option and functionality. I think we should keep this.

Examples/Scripts/Python/pypi_finding_fitting_demo.py
benjaminhuth, line 1: Not sure, do we need to change this file?

Examples/Scripts/Python/truth_tracking_kalman.py
```diff
@@ -52,10 +54,15 @@ def runTruthTrackingKalman(
         addKalmanTracks,
     )
 
+    from acts.examples.visualization import (
+            TrackVisualizerAlg,
+            PyVisualization2D
+        )
+
     s = s or acts.examples.Sequencer(
-        events=100, numThreads=-1, logLevel=acts.logging.INFO
```
benjaminhuth, line 56: hmm we should not change the defaults here.

Examples/Scripts/Python/truth_tracking_kalman.py
```diff
@@ -64,6 +71,20 @@ def runTruthTrackingKalman(
 
     logger = acts.getDefaultLogger("Truth tracking example", acts.logging.INFO)
 
+    #for ievt in range(numParticles):
```
benjaminhuth: Hmm this part we need to refactor. I think we should just use `dangerouslyDefaultConstruct` for the geo context, and not use this workaround to get a geo context.

Python/Examples/tests/test_geometry.py
```diff
@@ -1,13 +1,13 @@
...
 import acts
 import functools
+import matplotlib.pyplot as plt
 from acts.examples import GenericDetector
 from acts.examples.odd import getOpenDataDetector
 import json
 
 from helpers import dd4hepEnabled
 
-
```
benjaminhuth, line 10:
```suggestion
```

Python/Examples/tests/test_geometry.py
```diff
@@ -34,6 +34,8 @@
         "odd",
     ],
 )
+
+
```
benjaminhuth:
```suggestion
```

review_test.png
benjaminhuth, line 1: Test files should not be commited

Python/Core/src/Visualization.cpp
```diff
@@ -59,5 +87,32 @@ void addVisualization(py::module& m) {
...
+
+  py::class_<EventDataView3D>(m, "EventDataView3D")
+    .def_static("drawTrack",
+      [] (IVisualization3D& helper, const AnyConstTrackProxy& track){
+        EventDataView3D::drawTrack(helper, track,  GeometryContext::dangerouslyDefaultConstruct());
+      });
+
+  py::class_<GeometryObject>(m, "GeometryObject");
+  m.def("viewConfigFactory()", &viewConfigFactory);
+  //m.def("defaultColoring", &defaultColoring);
```
benjaminhuth: Should this be commited, or not? If not, better remove the commented part

Examples/Scripts/Python/truth_tracking_kalman.py
```diff
@@ -218,18 +244,20 @@ def runTruthTrackingKalman(
         )
     )
 
-    return s
-
+    s.run()
+    vis.plot(projection=projection)
```
benjaminhuth:
```suggestion
    if pyVis:
    	vis.plot(projection=projection) 
```

Examples/Scripts/Python/truth_tracking_kalman.py
```diff
@@ -172,8 +195,11 @@ def runTruthTrackingKalman(
         reverseFilteringCovarianceScaling,
         linkForward=linkForward,
         useJosephFormulation=useJosephFormulation,
+        logLevel=acts.logging.INFO
     )
 
+    s.addAlgorithm(TrackVisualizerAlg("TrackVisualizerAlg", acts.logging.INFO, vis))
```
benjaminhuth:
```suggestion
    if pyVis:
    	s.addAlgorithm(TrackVisualizerAlg("TrackVisualizerAlg", acts.logging.INFO, vis))
```

Examples/Scripts/Python/truth_tracking_kalman.py
```diff
@@ -240,10 +268,16 @@ def runTruthTrackingKalman(
...
     runTruthTrackingKalman(
         trackingGeometry=trackingGeometry,
         field=field,
         digiConfigFile=digiConfigFile,
+        projection='xy',
         outputDir=Path.cwd(),
-    ).run()
+    )
+
+    plt.savefig("review_test.png")
```
benjaminhuth:
```suggestion
    plt.savefig("truth_tracking_kalman_visualization.png")
```
Should this be put into the function, and guarded by the boolean?

Examples/Scripts/Python/truth_tracking_kalman.py
```diff
@@ -52,10 +54,15 @@ def runTruthTrackingKalman(
         addKalmanTracks,
     )
 
+    from acts.examples.visualization import (
```
benjaminhuth: Hmm I think we can import this globally on the module top, as this is not conditional, right?

benjaminhuth (commented):
Hey, still a few comments, mostly minor though. A few larger points:
- The code contains many files that are test artifacts / images which should not be committed
- In the example script, some things are still a bit "rough" from local testing, that need to be cleaned up a bit
- It would be nice if you could also add a bit of documentation on the available features. Let's discuss later how to do this.

Core/include/Acts/Visualization/VisualizationBuffer.hpp
yannikaMatt: Yes, you are right. I have missed that.

Core/include/Acts/Geometry/TrackingGeometry.hpp
yannikaMatt: I have changed that

Core/include/Acts/Geometry/TrackingGeometry.hpp
yannikaMatt: Good point.

Core/include/Acts/Visualization/ViewConfig.hpp
yannikaMatt: I removed it for now

Examples/Scripts/Python/pypi_finding_fitting_demo.py
yannikaMatt: I do not think I have changed anything there. So I am not really sure how this ended up here.

Examples/Scripts/Python/truth_tracking_kalman.py
yannikaMatt: I missed that one. Changed it back to 100

Examples/Scripts/Python/truth_tracking_kalman.py
yannikaMatt: We could but I thought it would be nice to follow the pattern of the other example scripts here (like `acts.examples.simulation` etc.)

Python/Core/src/Visualization.cpp
yannikaMatt: That is also a remainder from when I was trying different things of how to bind this. I will remove it

Python/Examples/tests/test_geometry.py
yannikaMatt: Done

Python/Examples/tests/test_geometry.py
yannikaMatt: Done

Examples/Scripts/Python/truth_tracking_kalman.py
yannikaMatt: I can put this in the function and add an argument (and default) for the filename

Examples/Scripts/Python/truth_tracking_kalman.py
yannikaMatt: Done

Core/include/Acts/Visualization/ViewConfig.hpp
yannikaMatt: Done

Core/include/Acts/Visualization/ViewConfig.hpp
yannikaMatt: Done

Examples/Scripts/Python/geometry.py
yannikaMatt: Now it should be restored again. Sorry

Core/include/Acts/Geometry/TrackingVolume.hpp
```diff
@@ -522,9 +522,7 @@ class TrackingVolume : public Volume {
   /// @param portalViewConfig View configuration for portals
   /// @param sensitiveViewConfig View configuration for sensitive surfaces
   void visualize(IVisualization3D& helper, const GeometryContext& gctx,
-                 const ViewConfig& viewConfig,
-                 const ViewConfig& portalViewConfig,
-                 const ViewConfig& sensitiveViewConfig) const;
+                 const ViewConfigFunc& viewConfigFactory) const;
```
andiwand:
```suggestion
                 const ViewConfigFunc& func = viewConfigFactory()) const;
```
looking at `Core/include/Acts/Geometry/TrackingGeometry.hpp` - should this be synced?

Core/include/Acts/Visualization/EventDataView3D.hpp
```diff
@@ -307,6 +311,20 @@ struct EventDataView3D {
...
   }
+
+  /// Helper method to draw a track from AnyTrackProxy
+  ///
+  /// @param helper [in, out] The visualization helper
+  /// @param track The track to be drawn
+  /// @param itrack the track index in the container
+  /// @param gctx The geometry context for which it is drawn
+  /// @param lineconfig The visualization options for the lines to be drawn
+
```
andiwand:
```suggestion
```

Core/include/Acts/Visualization/EventDataView3D.hpp
```diff
@@ -307,6 +311,20 @@ struct EventDataView3D {
...
+  static void drawTrack(
+    IVisualization3D& helper, const AnyConstTrackProxy& track,
+      const GeometryContext& gctx =
+          GeometryContext::dangerouslyDefaultConstruct());
```
andiwand: I think it would be nice to point `drawMultiTrajectory` to this implementation. but I guess it will need another intermediate step to go from multi trajectory -> track. if this is too involved we should just drop `drawMultiTrajectory`

Core/include/Acts/Visualization/ViewConfig.hpp
```diff
@@ -162,4 +164,13 @@ static const ViewConfig s_viewGrid = {.color = {220, 0, 0}};
 /// View configuration for line visualization
 static const ViewConfig s_viewLine = {.color = {0, 0, 220}};
 
+/// Default ViewConfigFactory which colors sensitive surfaces green and boundarys orange 
+ViewConfig defaultGeometryColoring(const GeometryObject&); 
+
+/// Function that takes a GeometryObject as an input and outputs a ViewConfig
+/// This allows to use custom functions for determining the colour of GeometryObjects
+using ViewConfigFunc = std::function<ViewConfig(const GeometryObject&)>;
+ViewConfigFunc viewConfigFactory();
```
andiwand:
```suggestion

ViewConfigFunc defaultViewConfigFactory();
```

Core/include/Acts/Visualization/VisualizationBuffer.hpp
```diff
@@ -0,0 +1,125 @@
...
+  /// @param name The name of the object
+  void object(const std::string& name) {}
+
+  const std::vector<Vector3> &vertices3D() const{
+        return m_vertices;
+  }
+
+  const std::vector<std::vector<Vector3>> &surfaces() const {
+        return m_surfaces;
+      }
```
andiwand: formatting looks a bit off

Core/include/Acts/Visualization/VisualizationBuffer.hpp
```diff
@@ -0,0 +1,125 @@
...
+#include <vector>
+#include <array>
+#include <iostream>
+
+namespace Acts {
+
+/// This helper produces output for Python visualization. Note that colors are not
+/// supported in this implementation.
+///
+class VisualizationBuffer : public IVisualization3D {
```
andiwand: if this is python only, should it go to core?

Examples/Configs/odd-digi-smearing-config-gen3.json:Zone.Identifier
andiwand, line 1: should this be dropped from the branch?

Core/include/Acts/Geometry/TrackingVolume.hpp
yannikaMatt: It probably should

yannikaMatt: Though I am not really sure about the default value

Core/include/Acts/Visualization/EventDataView3D.hpp
benjaminhuth: I think we should try to unify the functionality of those fucntions in a next step. For now, I would keep it, also because it is part of the public API.

Core/include/Acts/Visualization/VisualizationBuffer.hpp
benjaminhuth: Could be done, but in principle this is more generic and could be used also by some other client that wants raw vertices/lines etc... No strong opinion.

Core/include/Acts/Geometry/TrackingVolume.hpp
yannikaMatt: I discussed with @benjaminhuth and we can sync this

Core/include/Acts/Visualization/VisualizationBuffer.hpp
yannikaMatt: In what sense? Should it be all in one line?

Examples/Configs/odd-digi-smearing-config-gen3.json:Zone.Identifier
yannikaMatt: Yes, I will do that

Core/include/Acts/Visualization/EventDataView3D.hpp
```diff
@@ -307,6 +311,18 @@ struct EventDataView3D {
...
+  ///
+  /// @param helper [in, out] The visualization helper
+  /// @param track The track to be drawn
+  /// @param itrack the track index in the container
+  /// @param gctx The geometry context for which it is drawn
+  /// @param lineconfig The visualization options for the lines to be drawn
+  static void drawTrack(IVisualization3D& helper,
+                        const AnyConstTrackProxy& track,
+                        const GeometryContext& gctx =
+                            GeometryContext::dangerouslyDefaultConstruct());
```
paulgessinger: Can we avoid the default here and ask t he user to call the geometry context factory?

Core/src/Geometry/TrackingVolume.cpp
paulgessinger: ok

Core/src/Visualization/ViewConfig.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Acts {
+
+ViewConfig defaultGeometryColoring(const GeometryObject& geoObj) {
+     if (geoObj.geometryId().boundary() != 0)
+    {
+        return ViewConfig{.color = {255, 165, 0}};
+    }
+  
+    //std::cout << geoObj.geometryId() << std::endl;
```
paulgessinger:
```suggestion
```

Core/include/Acts/Visualization/ViewConfig.hpp
```diff
@@ -162,4 +164,15 @@ static const ViewConfig s_viewGrid = {.color = {220, 0, 0}};
...
+/// Default ViewConfigFactory which colors sensitive surfaces green and
+/// boundaries orange
+ViewConfig defaultGeometryColoring(const GeometryObject&);
+
+/// Function that takes a GeometryObject as an input and outputs a ViewConfig
+/// This allows to use custom functions for determining the colour of
+/// GeometryObjects
+using ViewConfigFunc = std::function<ViewConfig(const GeometryObject&)>;
+
+ViewConfigFunc viewConfigFactory();
```
paulgessinger: So this is a factory factory? :) Does it not work to pass `defaultGeometryColoring` as a function itself to the factory arguments? (I believe it should)

Examples/Scripts/Python/geometry.py
```diff
@@ -111,13 +111,14 @@ def runGeometry(
 
 
 if "__main__" == __name__:
-    # detector = acts.examples.GenericDetector()
-    detector = getOpenDataDetector()
+    #detector = acts.examples.GenericDetector()
+    detector = getOpenDataDetector(gen3=True)
```
paulgessinger: Maybe this needs to be configurable? Otherwise this could stick to the gen1 default for now and the gen3+pyVis path is only tested through pytest?

Examples/Scripts/Python/geometry.py
```diff
@@ -111,13 +111,14 @@ def runGeometry(
...
     trackingGeometry = detector.trackingGeometry()
     decorators = detector.contextDecorators()
-
-    runGeometry(trackingGeometry, decorators, outputDir=Path.cwd())
-
+   
+    import matplotlib.pyplot as plt
+    runGeometry(trackingGeometry, decorators, projection='rz')
+    plt.savefig("geoDefault_rz")
```
paulgessinger: This should probably be gated on the python visualization actually running, right? Also, does it communicate with the active figure through the `plt` current figure/axes object only?

Python/Examples/python/visualization.py
```diff
@@ -0,0 +1,137 @@
...
+
+        return acts.examples.ProcessCode.SUCCESS
+
+
+class PyVisualization2D(acts.VisualizationBuffer):
+
+    def plot(self, projection, filename, linewidth=None, linestyle=None, ax=None):
+        import matplotlib.pyplot as plt
+
+        fig, ax = plt.subplots()
```
paulgessinger, line 60: Ah ok no, it produces a figure here. There's also the `ax` argument, but it's always ignored, I think. Another pattern (and I'm not sure if that's better necessarily), is to do `plt.gca()` to get the currently active global axes.

Python/Examples/src/EventData.cpp
```diff
@@ -497,6 +499,23 @@ void addEventData(py::module& mex) {
...
+
+  py::classh<Acts::AnyMutableTrackStateProxy>(mex, "AnyMutableTrackStateProxy")
+      .def(py::init<TrackStateProxy&>());
+  py::implicitly_convertible<TrackStateProxy,
+                             Acts::AnyMutableTrackStateProxy>();
+
+  py::classh<Acts::AnyConstTrackStateProxy>(mex, "AnyConstTrackState")
+      .def(py::init<ConstTrackStateProxy&>());
+  py::implicitly_convertible<ConstTrackStateProxy,
+                             Acts::AnyConstTrackStateProxy>();
```
paulgessinger, line 545: This was needed on top of the conversion changes in the other PR, correct? Maybe it makes sense to still break these out?

Python/Core/src/Visualization.cpp
```diff
@@ -21,6 +30,18 @@ using namespace Acts;
 
 namespace ActsPython {
 
+py::array_t<int> colorsToNumpy(std::span<const Color> colors,
+                               const py::object& base) {
+  std::vector<py::std::size_t> shape{
```
paulgessinger: I don't think `py::std::size_t` exists.

Tests/UnitTests/Core/EventData/CMakeLists.txt
```diff
@@ -16,5 +16,5 @@ add_unittest(ParticleHypothesis ParticleHypothesisTests.cpp)
 add_unittest(MultiTrajectoryHelpers MultiTrajectoryHelpersTests.cpp)
 add_unittest(SubspaceHelpers SubspaceHelpersTests.cpp)
 add_unittest(TrackParameterHelpers TrackParameterHelpersTests.cpp)
-add_unittest(SpacePointContainer SpacePointContainerTests.cpp)
-add_unittest(SeedContainer SeedContainerTests.cpp)
+add_unittest(SpacePointContainer2 SpacePointContainer2Tests.cpp)
+add_unittest(SeedContainer2 SeedContainer2Tests.cpp)
```
paulgessinger: I think this is a rebase artefact.

Examples/Scripts/Python/pypi_finding_fitting_demo.py
```diff
@@ -124,20 +124,11 @@ class PythonTrackFitter(acts.examples.IAlgorithm):
...
-            self.tracks = acts.examples.WriteDataHandle(
+            self.tracks = acts.examples.ReadDataHandle(
                 self, acts.examples.ConstTrackContainer, "Tracks"
             )
             self.tracks.initialize("fitted_tracks")
 
-            self.spacepoints = acts.examples.ReadDataHandle(
-                self, acts.SpacePointContainer, "Spacepoints"
-            )
-            self.spacepoints.initialize("spacepoints")
```
paulgessinger, line 140: Rebase artefact?

Core/src/Visualization/ViewConfig.cpp
paulgessinger, line 1: I think this needs the license header.

Tests/UnitTests/Core/Geometry/DiamondVolumeBoundsTests.cpp
```diff
@@ -59,8 +59,7 @@ BOOST_AUTO_TEST_CASE(DiamondVolumeBoundsCreation) {
       "TestDiamondVolume");
 
   Acts::ObjVisualization3D helper;
-  trackingVolume->visualize(helper, gctx, {.visible = true}, {.visible = true},
-                            {.visible = true});
+  trackingVolume->visualize(helper, gctx, {});
```
paulgessinger: I think the `{}` default will not work here, the function will be value-less and throw when called.

Examples/Scripts/Python/truth_tracking_kalman.py
```diff
@@ -218,7 +236,11 @@ def runTruthTrackingKalman(
         )
     )
 
-    return s
+    s.run()
```
paulgessinger: I think we'll still need to return the sequencer.

Examples/Scripts/Python/geometry.py
```diff
@@ -2,7 +2,6 @@
 
 import os
 import json
-from pathlib import Path
```
paulgessinger, line 5: This is still used I believe.

Examples/Scripts/Python/geometry.py
yannikaMatt: I actually forgot to change this in this file. By now there is a much cleaner way where the import and the saving of the image happens within the `plot()` function in `Python/Examples/python/visualization.py` directly. So both can be removed from here and `vis.plot()` gets an additional filename argument.

Python/Examples/python/visualization.py
yannikaMatt: You are right, the `ax` argument is not required anymore. I think it was @benjaminhuth who I was talking to about this, that `.gca()` can lead to some funny behaviour and it would be better to define `fig, ax = plt.subplots()` at the beginning of the function and later do `fig.save(filename)`.

Python/Core/src/Visualization.cpp
yannikaMatt: Yeah, this was a mistake of mine. Kinda weird though that it still worked. I think it has to be `std::size_t`

Tests/UnitTests/Core/EventData/CMakeLists.txt
yannikaMatt: Probably. I did not manually add them. I will change that back?

Examples/Scripts/Python/pypi_finding_fitting_demo.py
yannikaMatt: Yes, I will change it back

Examples/Scripts/Python/geometry.py
yannikaMatt: Added it

Examples/Scripts/Python/geometry.py
yannikaMatt: I think at the very beginning when I started here we had some issues when trying to run this for gen1. At the moment I cannot recall though what it was exactly.

Examples/Scripts/Python/truth_tracking_kalman.py
yannikaMatt: ok

Tests/UnitTests/Core/Geometry/DiamondVolumeBoundsTests.cpp
yannikaMatt: Removed `{}`

Examples/Scripts/Python/geometry.py
yannikaMatt: A test exists so we can definetly do that

Core/include/Acts/Visualization/EventDataView3D.hpp
yannikaMatt: Seems to work, so yes

Core/include/Acts/Visualization/ViewConfig.hpp
yannikaMatt: I think I have resolved that now. I had to change the python binding in `Geometry.cpp` to do so.

Core/src/Visualization/ViewConfig.cpp
yannikaMatt: Isn't that the very header at the top already?

Python/Examples/python/visualization.py
```diff
@@ -0,0 +1,135 @@
...
+                        self.faceColors[n_face][0] / 255,
+                        self.faceColors[n_face][1] / 255,
+                        self.faceColors[n_face][2] / 255,
+                        0.5,
+                    ),
+                    lw=1,
+                )
+
+        # Check if there is a track to be drawn
+        print(len(self.segments))
```
yannikaMatt, line 116:
```suggestion
    
```

Python/Examples/python/visualization.py
```diff
@@ -0,0 +1,135 @@
...
+                        self.faceColors[n_face][0] / 255,
+                        self.faceColors[n_face][1] / 255,
+                        self.faceColors[n_face][2] / 255,
+                        0.5,
+                    ),
+                    lw=1,
+                )
+
+        # Check if there is a track to be drawn
+
```
yannikaMatt:
```suggestion
```

paulgessinger (dismissed):
Just needs a rebase I think.

andiwand:
this is a breaking change right?

paulgessinger:
@andiwand technically correct that it's a breaking change, but could be made backwards compatible like this:

```diff
diff --git a/Core/include/Acts/Geometry/TrackingGeometry.hpp b/Core/include/Acts/Geometry/TrackingGeometry.hpp
index a116f1d072..460ce0b72f 100644
--- a/Core/include/Acts/Geometry/TrackingGeometry.hpp
+++ b/Core/include/Acts/Geometry/TrackingGeometry.hpp
@@ -288,6 +288,17 @@ class TrackingGeometry {
   /// @param viewConfig Global view config
   /// @param portalViewConfig View config for portals
   /// @param sensitiveViewConfig View configuration for sensitive surfaces
+  /// @deprecated Use the ViewConfigFunc overload instead
+  [[deprecated("Use the ViewConfigFunc overload instead")]]
+  void visualize(IVisualization3D& helper, const GeometryContext& gctx,
+                 const ViewConfig& viewConfig,
+                 const ViewConfig& portalViewConfig,
+                 const ViewConfig& sensitiveViewConfig) const;
+
+  /// Visualize a tracking geometry including substructure
+  /// @param helper The visualization helper that implement the output
+  /// @param gctx The geometry context
+  /// @param func Function determining the ViewConfig for each GeometryObject
   void visualize(IVisualization3D& helper, const GeometryContext& gctx,
                  const ViewConfigFunc& func = defaultGeometryColoring) const;
 
diff --git a/Core/include/Acts/Geometry/TrackingVolume.hpp b/Core/include/Acts/Geometry/TrackingVolume.hpp
index 9653b93fac..417367412c 100644
--- a/Core/include/Acts/Geometry/TrackingVolume.hpp
+++ b/Core/include/Acts/Geometry/TrackingVolume.hpp
@@ -529,6 +529,17 @@ class TrackingVolume : public Volume {
   /// @param viewConfig The view configuration
   /// @param portalViewConfig View configuration for portals
   /// @param sensitiveViewConfig View configuration for sensitive surfaces
+  /// @deprecated Use the ViewConfigFunc overload instead
+  [[deprecated("Use the ViewConfigFunc overload instead")]]
+  void visualize(IVisualization3D& helper, const GeometryContext& gctx,
+                 const ViewConfig& viewConfig,
+                 const ViewConfig& portalViewConfig,
+                 const ViewConfig& sensitiveViewConfig) const;
+
+  /// Produces a 3D visualization of this tracking volume
+  /// @param helper The visualization helper describing the output format
+  /// @param gctx The geometry context
+  /// @param func Function determining the ViewConfig for each GeometryObject
   void visualize(IVisualization3D& helper, const GeometryContext& gctx,
                  const ViewConfigFunc& func = defaultGeometryColoring) const;
 
diff --git a/Core/include/Acts/Visualization/GeometryView3D.hpp b/Core/include/Acts/Visualization/GeometryView3D.hpp
index 997016a961..7ec049c282 100644
--- a/Core/include/Acts/Visualization/GeometryView3D.hpp
+++ b/Core/include/Acts/Visualization/GeometryView3D.hpp
@@ -35,7 +35,7 @@ struct GeometryView3D {
   /// @param viewConfig The drawing configuration
   static void drawPolyhedron(IVisualization3D& helper,
                              const Polyhedron& polyhedron,
-                             const ViewConfig& viewConfig);
+                             const ViewConfig& viewConfig = s_viewVolume);
 
   /// Helper method to draw Surface objects
   ///
diff --git a/Core/src/Geometry/TrackingGeometry.cpp b/Core/src/Geometry/TrackingGeometry.cpp
index a52a9cb48d..833e654c45 100644
--- a/Core/src/Geometry/TrackingGeometry.cpp
+++ b/Core/src/Geometry/TrackingGeometry.cpp
@@ -388,6 +388,15 @@ TrackingGeometry::geoIdSurfaceMap() const {
   return m_surfacesById;
 }
 
+void TrackingGeometry::visualize(IVisualization3D& helper,
+                                 const GeometryContext& gctx,
+                                 const ViewConfig& viewConfig,
+                                 const ViewConfig& portalViewConfig,
+                                 const ViewConfig& sensitiveViewConfig) const {
+  highestTrackingVolume()->visualize(helper, gctx, viewConfig, portalViewConfig,
+                                     sensitiveViewConfig);
+}
+
 void TrackingGeometry::visualize(IVisualization3D& helper,
                                  const GeometryContext& gctx,
                                  const ViewConfigFunc& func) const {
diff --git a/Core/src/Geometry/TrackingVolume.cpp b/Core/src/Geometry/TrackingVolume.cpp
index 118c1ebac3..e55b2c9c17 100644
--- a/Core/src/Geometry/TrackingVolume.cpp
+++ b/Core/src/Geometry/TrackingVolume.cpp
@@ -605,6 +605,23 @@ void TrackingVolume::addSurface(std::shared_ptr<Surface> surface) {
   m_surfaces.push_back(std::move(surface));
 }
 
+void TrackingVolume::visualize(IVisualization3D& helper,
+                               const GeometryContext& gctx,
+                               const ViewConfig& viewConfig,
+                               const ViewConfig& portalViewConfig,
+                               const ViewConfig& sensitiveViewConfig) const {
+  visualize(helper, gctx,
+            [&](const GeometryObject& geoObj) {
+              if (geoObj.geometryId().boundary() != 0) {
+                return portalViewConfig;
+              }
+              if (geoObj.geometryId().sensitive() != 0) {
+                return sensitiveViewConfig;
+              }
+              return viewConfig;
+            });
+}
+
 void TrackingVolume::visualize(IVisualization3D& helper,
                                const GeometryContext& gctx,
                                const ViewConfigFunc& viewConfigFactory) const {

```

@yannikaMatt what do you think?

yannikaMatt:
I think this is a good suggestion.

Core/src/Geometry/TrackingGeometry.cpp
```diff
@@ -397,6 +397,12 @@ void TrackingGeometry::visualize(IVisualization3D& helper,
                                      sensitiveViewConfig);
```
paulgessinger, line 398: @yannikaMatt this one is deprecated now

Tests/UnitTests/Core/Geometry/BlueprintApiTests.cpp
```diff
@@ -391,7 +391,7 @@ BOOST_AUTO_TEST_CASE(NodeApiTestContainers) {
 
   ObjVisualization3D vis;
 
-  trackingGeometry->visualize(vis, gctx, {}, {});
+  trackingGeometry->visualize(vis, gctx, {});
```
paulgessinger:
```suggestion
  trackingGeometry->visualize(vis, gctx);
```

this would fix the failing unit test, I think.

Core/src/Geometry/TrackingGeometry.cpp
yannikaMatt: Should be fixed now

Examples/Scripts/Python/truth_tracking_kalman.py
```diff
@@ -14,13 +14,15 @@ def runTruthTrackingKalman(
...
     outputDir: Path,
+    pyVis=True,
     inputParticlePath: Optional[Path] = None,
     inputHitsPath: Optional[Path] = None,
     decorators=[],
     generatedParticleType: acts.PdgParticle = acts.PdgParticle.eMuon,
     reverseFilteringMomThreshold=0 * u.GeV,
     reverseFilteringCovarianceScaling=100.0,
     numParticles=1,
+    projection="rz",
```
benjaminhuth:
```suggestion
    pyVisProjection="rz",
```
maybe rename to make clear what this refers to

Examples/Scripts/Python/geometry.py
```diff
@@ -57,19 +59,14 @@ def runGeometry(
             )
             writer.write(context)
 
-        # The obj and material map outputs go to a single, event-independent
-        # file each, so writing them once (on the first event) is enough --
-        # every further event would only overwrite the same file.
         if outputObj and ievt == 0:
             vis = acts.ObjVisualization3D()
-            trackingGeometry.visualize(
```
benjaminhuth, line 65: Hmm we should maybe keep this, otherwise we don't test obj visualization anymore. you need to adapt to your new API though.

yannikaMatt: I think it should be fine to just keep the previous version of visualize here. 
We still have this version of `visualize` as an overload in `TrackingGeometry.hpp`

paulgessinger:
Looks like the visualize overloads became breaking again. I didn't look through the commits in detail, but did one of the defaults change?

Core/include/Acts/Geometry/TrackingGeometry.hpp
```diff
@@ -282,16 +282,24 @@ class TrackingGeometry {
...
   /// @param gctx The geometry context
   /// @param viewConfig Global view config
   /// @param portalViewConfig View config for portals
   /// @param sensitiveViewConfig View configuration for sensitive surfaces
+  /// @deprecated Use the ViewConfigFunc overload instead
+  [[deprecated("Use the ViewConfigFunc overload instead")]]
+  void visualize(IVisualization3D& helper, const GeometryContext& gctx,
+                 const ViewConfig& viewConfig,
+                 const ViewConfig& portalViewConfig,
+                 const ViewConfig& sensitiveViewConfig) const;
```
paulgessinger, line 295: The missing default on the second and third argument here is what's causing the API to be breaking

```
[[deprecated("Use the ViewConfigFunc overload instead")]]
void visualize(IVisualization3D& helper, const GeometryContext& gctx,
               const ViewConfig& viewConfig,
               const ViewConfig& portalViewConfig = s_viewPortal,
               const ViewConfig& sensitiveViewConfig = s_viewSensitive) const;
```

yannikaMatt: I fixed that. Could this also be the reason for the segmentation fault in the example tests?

benjaminhuth (approved):
LGTM!

yannikaMatt:
[Uploading truth_tracking_kalman_visualization.pdf…]()

yannikaMatt:
truth_tracking_kalman_visualization

PR #5765, opened 2026-07-28
test: Add GBTS seeding regression test

andiwand:
GBTS has no test coverage: nothing under `Tests/` references it and physmon
does not exercise it either. The only way to run it today is the ITk example
chain, which needs geometry and connector data files.

This adds a self-contained regression test. It builds a `GbtsGeometry` from an
in-memory connector table, fills a synthetic space point container with
straight tracks from the origin, and pins the output of
`GraphBasedTrackSeeder::createSeeds` for a sparse and a dense scenario.

The expected values are recorded from the current implementation rather than
derived independently. They exist to detect behaviour changes during the
refactoring that follows, not to validate GBTS physics.

--- END COMMIT MESSAGE ---

## Stack

This is **1/5** of a stacked series refactoring the GBTS seeding EDM. The
branches live on my fork, so every PR targets `main` and the diffs are
cumulative — **review only the last commit of each**.

1. **this PR** — regression test
2. GBTS acceptance cuts to config
3. node EDM replaced by container indices
4. Examples fills the node storage directly
5. documentation page

## Details

The setup is a toy detector of four pixel barrel layers, with the connector
table fed to `GbtsLayerConnectionMap::fromStream` as a string literal, so the
test needs no data files.

Two scenarios:

- **Separated tracks**, where the expected seed list is pinned exactly. The
  graph has 36 edges for 12 tracks, i.e. no cross-track candidates.
- **Dense tracks**, packed closely in phi *and* tau so neighbours fall inside
  each other's phi sliding window. That builds **1230 edges with 822 links**
  from 160 space points, so the graph combinatorics, the tracking filter and
  clone removal all do real work. Only the seed count and the structural
  invariants are asserted there — which candidates survive heavy branching
  depends on float rounding, so a pinned list would be fragile across
  platforms.


jpreston-cern:
Thanks for doing this! in term of layers tested, would it also be worth adding in some tracks for the endcaps as well (as i see that only 8x000 layers are given)? 

as some refactoring in the future will focus on seeing if the the graph building and filtering can be standardized to not have hard coded dependencies on the layer type, it would be good to detect behavior changes in the end caps as well

jpreston-cern (approved):

AJPfleger (approved):

PR #5766, opened 2026-07-28
refactor: Make GBTS acceptance cuts configurable

andiwand:
Three magic numbers in the GBTS seeding path become configuration options, and
the "no cut" tau sentinel becomes explicit. No behaviour change.

- `maxAbsTau` (was `36.0f`): largest |cot(theta)| accepted for a doublet, i.e.
  |eta| of about 4.3.
- `moduleHalfLengthY` (was `10.0f`) and `moduleEdgeTolerance` (was `0.3f`): the
  pixel module half-length and the distance to its edge below which a cluster
  may be shortened, so the machine learning lookup table's second pair of tau
  bounds applies.

The per-node tau window used +-100 as its "do not cut" value, which only worked
because the doublet loop already rejects |tau| > 36 -- an implicit coupling
between two unrelated constants. The window is only ever compared against,
never used in arithmetic, so +-infinity says "do not cut" independently of the
acceptance cut, which frees 36 to become a normal config knob. The dead
initialisation ahead of the lookup table branch is folded away with it.

--- END COMMIT MESSAGE ---

**2/5** of a stacked series on top of #5765 -- review only the last commit
(`refactor: Make GBTS acceptance cuts configurable`).

The regression test added in #5765 gives identical output before and after.

### DIFF
--- a/Core/include/Acts/Seeding/GbtsDataStorage.hpp
+++ b/Core/include/Acts/Seeding/GbtsDataStorage.hpp
@@ -122,7 +122,11 @@ class GbtsNodeStorage final {
   void sortByPhi();
   /// Initialize node attributes
   /// @param useMl Use machine learning features
-  void initializeNodes(bool useMl);
+  /// @param moduleHalfLengthY Half-length in local y of a pixel module
+  /// @param moduleEdgeTolerance Distance to the module edge below which a
+  ///        cluster may be shortened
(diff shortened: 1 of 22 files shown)


jpreston-cern:
this looks all good to me 👍

jpreston-cern (approved):

AJPfleger (approved):

PR #5767, opened 2026-07-28
refactor: Replace the GBTS node EDM with space point container indices

andiwand:
Every field of the 36-byte `GbtsNode` GBTS copied each space point into already
existed in the `Acts::SpacePointContainer` it was handed: `x`, `y`, `z`, `r`,
`phi` are known columns, `layer`, `pcw` and `locPosY` are the dynamic columns
GBTS already required, and `idx` is the container index itself.

A node is now a `GbtsNodeIndex` into `GbtsNodeStorage`, which orders the nodes
by (eta bin, phi) in a space point container of its own, holds the derived
per-node data in dynamic columns on it, and exposes each eta bin as a contiguous
`SpacePointIndexRange`. This is the layout the traccc device implementation
uses. Nodes go in through `GbtsNodeStorage::insert`, which takes plain scalars,
so a caller can fill it from its own space point EDM. Code outside the graph
builder reads nodes through `GbtsNodeView`, which yields a `GbtsNodeProxy` with
named accessors.

The parallel per-node arrays become `GbtsNodeParams` and `GbtsNodeEdgeInfo`,
each a single record because the innermost doublet loop reads all their fields
together. `GbtsEdge` stores node indices instead of pointers and caches the
outer node's GBTS layer id, so the innermost neighbour loop no longer chases the
node.

--- END COMMIT MESSAGE ---

On a synthetic 64k space point workload the Core seeding path goes from 55.9 ms
to 47.2 ms median, about 16% faster, for identical seed output. The regression
test from #5765 passes unchanged.

One deliberate behaviour change: nodes with exactly equal phi within an eta bin
are now tie-broken by insertion order, where before they were ordered by the
address of the node object. Arbitrary either way, but real data could reorder a
handful of equal-phi nodes.

### DIFF
--- a/Core/include/Acts/Seeding/GbtsDataStorage.hpp
+++ b/Core/include/Acts/Seeding/GbtsDataStorage.hpp
@@ -8,10 +8,15 @@
 
 #pragma once
 
+#include "Acts/EventData/SpacePointColumnProxy.hpp"
+#include "Acts/EventData/SpacePointContainer.hpp"
+#include "Acts/EventData/Types.hpp"
+
 #include <array>
 #include <cstdint>
 #include <limits>
 #include <memory>
+#include <optional>
 #include <span>
 #include <vector>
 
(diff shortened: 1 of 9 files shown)


Core/include/Acts/Seeding/GbtsDataStorage.hpp
```diff
@@ -8,168 +8,172 @@
...
+using GbtsNodeIndex = SpacePointIndex;
+
+/// Sentinel for an unset graph node index.
+static constexpr GbtsNodeIndex kGbtsNodeIndexInvalid = kSpacePointIndexInvalid;
+
+/// Per-node parameters used while building the graph.
+///
+/// All five values are read together in the innermost doublet loop, so they sit
+/// in one record rather than in separate arrays.
+struct GbtsNodeParams final {
```
jpreston-cern, line 49: if this has the r and z values for each node, is the packed `xyzr` column in the GBTS `SpacePointContainer` needed in this case?

Core/src/Seeding/GraphBasedTrackSeeder.cpp
```diff
@@ -41,58 +41,45 @@ GraphBasedTrackSeeder::GraphBasedTrackSeeder(
...
-    if (isPixel) {
-      nPixelLoaded += nodeStorage.loadPixelGraphNodes(
-          l, nodes, m_cfg.useMl, m_cfg.maxEndcapClusterWidth);
-    } else {
-      nStripLoaded += nodeStorage.loadStripGraphNodes(l, nodes);
-    }
-  }
-  ACTS_DEBUG("Loaded " << nPixelLoaded << " pixel space points and "
-                       << nStripLoaded << " strip space points");
+  nodeStorage.extend(spacePoints, layerColumn, clusterWidthColumn,
```
jpreston-cern, line 69: would it be better to call this function something like `addSpacepoints` or `addAndExtend` to make it explicit it is adding the spacepoints and then decorating the container with the extra cluster and layer information?

Core/include/Acts/Seeding/GbtsDataStorage.hpp
andiwand: this is an optimization of the innermost doublet loop. the data is duplicated but it is more cache efficient

Core/src/Seeding/GraphBasedTrackSeeder.cpp
andiwand: this is the same naming we use for the space point grids. the node storage and the grids align pretty well now. if we want to change the naming we should do it for all of the grids

jpreston-cern (approved):

paulgessinger (approved):

PR #5768, opened 2026-07-28
refactor: Fill the GBTS node storage directly from the input space points

andiwand:
`GraphBasedSeedingAlgorithm` built a second `Acts::SpacePointContainer` per
event, copying every space point across purely so that GBTS could read the
`layerId`, `clusterWidth` and `localPositionY` columns off it, and then mapped
the resulting seed indices back through `copiedFromIndex`.

`GbtsNodeStorage::insert` takes plain scalars, so that intermediate container
is no longer needed. The algorithm now resolves the GBTS layer index per space
point and inserts straight into the node storage, and the seeds come back
indexed into the input container, which removes the remapping loop as well.
Besides the copy this drops roughly `8 columns x N` virtual `emplace_back`
calls per event; the per-space-point `std::map` geometry-id lookup is inherent
to the layer mapping and simply moves.

Note this path is covered neither by the GBTS unit test, which drives Core
directly, nor by physmon, which does not exercise GBTS. It builds and keeps the
algorithm's structure, but diffing the seeds of `full_chain_itk_Gbts.py` before
and after needs ITk geometry data that was not available here.

### DIFF
--- a/Core/include/Acts/Seeding/GbtsDataStorage.hpp
+++ b/Core/include/Acts/Seeding/GbtsDataStorage.hpp
@@ -8,10 +8,15 @@
 
 #pragma once
 
+#include "Acts/EventData/SpacePointColumnProxy.hpp"
+#include "Acts/EventData/SpacePointContainer.hpp"
+#include "Acts/EventData/Types.hpp"
+
 #include <array>
 #include <cstdint>
 #include <limits>
 #include <memory>
+#include <optional>
 #include <span>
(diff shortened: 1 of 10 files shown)


andiwand:
Rebased on `main` after #5854. The GBTS series is no longer stacked: this PR now applies to `main` on its own and can be reviewed and merged independently of the others.

jpreston-cern (approved):

PR #5769, opened 2026-07-28
docs: Add a GBTS documentation page

andiwand:
GBTS had no prose documentation. This adds `docs/groups/pattern_recog/gbts.md`
under the seeding group, covering the four stages of the algorithm: sorting space
points into eta/phi-ordered graph nodes, building and linking the doublet
graph, propagating chain levels through a cellular automaton, and extracting
seeds with the Kalman-like tracking filter. It also covers the geometry model
GBTS uses instead of the ACTS tracking geometry, the connector file that
defines which layer pairs may form edges, the optional cluster-width
acceptance, and the configuration options grouped by the stage they affect.

The EDM structs are pulled in as doxygen snippets from the headers rather than
duplicated, following `docs/SNIPPETS.md`, which is why the headers gain marker
comments. The graph EDM now lives in a detail header, which doxygen does not
index, so the page names those types without linking them.

PR #5772, opened 2026-07-28
test: Add an ITk-scale GBTS seeding benchmark

andiwand:
Sixth of a stack. **Review only the last commit** — the branch lives on a fork, so GitHub can only base it on `main` and the diff is cumulative over the ones below.

Stack: #5765 → #5766 → #5767 → #5768 → #5769 → **this** → (perf follow-up)

---

Adds a throughput benchmark for the GBTS seeder, plus a synthetic event generator kept deliberately separate from it so other seeders can be profiled on the same event.

`SyntheticItkEvent.hpp` is a coarse stand-in for the ITk pixel detector — five barrel cylinders and nine endcap disks per side split into rings — filled from helical tracks with a soft-dominated 1/pT spectrum plus a displaced component standing in for secondaries and combinatorial background. The default lands near 200k space points, the order an ITk ttbar event at pileup 200 produces. Space points carry the generating track index, so seeding efficiency can be evaluated without a truth-matching stage.

`GbtsSeedingBenchmark.cpp` adds only the GBTS-specific geometry on top: the logical layer numbering and the layer connection table.

```
layers=41 etaBins=289 binGroups=289 tracks=33000 spacePoints=196732
seeds=15903  ~287 ms/event   (3.5M edges)
```

### Caveat on the edge ceiling

The synthetic event is noise free, so more geometric doublet candidates survive the cuts than in a real event and the graph outgrows the 2M edge default. Benchmarking against a saturated ceiling would hide any work done past it, so `--max-edges` defaults to 6M here. Pass `--max-edges 2000000` to measure the truncated production regime instead.

Whether a real ITk event sits above or below the 2M default is worth a check by someone with the data — if it saturates in production too, that is itself interesting.


andiwand:
Superseded by the synthetic benchmark in #5808.

PR #5773, opened 2026-07-28
perf: Speed up the GBTS graph building

andiwand:
Two micro-optimizations in the innermost loop of `buildTheGraph`: test the tau
ratio before the third node's layer is loaded and classified, and compute the
tau reciprocal as an addition instead of a division. Graph building gets ~5%
faster on an ITk PU200 event, ~11% at PU400.

The first is bit-identical. The second moves candidates sitting exactly on the
tau ratio cut, so a few seeds change without changing the efficiency, and is
kept as its own commit so it can be dropped. `DerivedConfig` now rejects a
negative `tauRatioCorr`, which the precomputed threshold assumes cannot happen.

Numbers, commit breakdown and the alternatives that were measured and rejected
are in the comment below.


andiwand:
Rebased on `main` after #5854. The GBTS series is no longer stacked: this PR now applies to `main` on its own and can be reviewed and merged independently of the others.

andiwand:
## Measurements

GBTS benchmark of #5808, each variant built and run back to back on a quiet
machine, median of ten runs. Each column is relative to the state before it.

| event | triplet loop | tau reciprocal |
| --- | --- | --- |
| ITk PU200, 221k space points | -5.4% | -0.6% |
| ODD PU200 | -3.4% | -0.4% |
| ITk PU400 | -11.1% | ±0 |

The gain scales with edge density, which is what separates the rows: the PU200
event makes 779k edges against the 2M production `nMaxEdges`, PU400 makes 3.2M.
The tau reciprocal sits at the run-to-run floor everywhere and should be read
as sub-1%, not as a number. Order bias was checked by re-running the sweep
backwards; the triplet loop reproduces within 0.5%.

## Commit breakdown

**`perf: Speed up the GBTS triplet matching loop`** - behaviour preserving,
seeds bit-identical.

- Test the tau ratio first. The adaptive correction can only *loosen* that cut,
  so comparing against the loosest threshold it could take rejects most
  candidates before the third node's layer is loaded and the barrel/gap
  classification is worked out. This is nearly all of the gain.
- Drop `at()` on the edge storage in the three loops that walk it; the index
  comes from the storage's own bookkeeping.
- Give the sliding window the source bin's node list, count and layer directly
  instead of a pointer to the bin, so the innermost loop does not chase the bin
  per candidate. Only non-empty bins get a window, and the windows are
  allocated once rather than per bin group.

**`perf: Compute the GBTS tau reciprocal without a division`** - ⚠️ not
bit-for-bit behaviour preserving, kept as its own commit so it can be dropped
independently.

`(sqrt(1+t^2) - t)(sqrt(1+t^2) + t)` is exactly 1, so the reciprocal is an
addition instead of a division, once per created edge. The two forms differ in
the last bits, so candidates sitting exactly on the tau ratio cut can fall
either way: on the ITk event the seeds come out identical, on ODD 2680 -> 2682
and at PU400 14814 -> 14815, efficiency unchanged in all three.

## Tried and rejected, for the record

All measured back to back against the committed state; none are in this PR.

- **Hoisting the edges' tau into a parallel array**, so the scans walk 4 B per
  edge instead of a 52 B stride: no change. That loop is compute- and
  branch-bound, not bandwidth-bound - which also argues against splitting
  `GbtsEdge` into hot/cold arrays, so that was not attempted.
- **Rewriting the tau cut as a precomputed range test on `p[0]`**: consistently
  ~2% *slower*.
- **Keeping the sliding window position in a local** across the scan: noise.
- **Reusing the phi-sort buckets across eta bins**: noise; the allocator is
  only ~3% of the event.

## Not covered

Neither change is exercised by physmon, which does not run GBTS. The Core unit
test from #5765 covers them, and the benchmark of #5808 is what produced the
numbers above.

Core/src/Seeding/GraphBasedTrackSeeder.cpp
```diff
@@ -17,13 +17,21 @@
...
 
 GraphBasedTrackSeeder::DerivedConfig::DerivedConfig(const Config& config)
     : Config(config) {
+  // buildTheGraph pre-computes the loosest tau ratio threshold it can apply,
+  // which assumes the correction only ever widens the cut.
+  if (config.tauRatioCorr < 0) {
+    throw std::invalid_argument(
+        "GraphBasedTrackSeeder: tauRatioCorr must not be negative");
+  }
+
```
jpreston-cern: is this check necessary as it is not derived by GBTS at runtime?

andiwand: I think this should go into `GraphBasedTrackSeeder::GraphBasedTrackSeeder`

jpreston-cern (approved):

PR #5774, opened 2026-07-28
ci: Add ALPAKA_HIP to CI

StewMH:
Add alpaka HIP backend to CI. Requires an update to 7.2.0 for some newer alpaka code with fancier C++ features.

--- END COMMIT MESSAGE ---


StewMH:
The sanity checks from #5835 are not compiling, e.g.:
```
/__w/acts/acts/Traccc/device/hip/src/clusterization/../sanity/../utils/global_index.hpp:17:10: error: use of undeclared identifier 'blockIdx'
   17 |   return blockIdx.x * blockDim.x + threadIdx.x;
      |          ^
/__w/acts/acts/Traccc/device/hip/src/clusterization/../sanity/../utils/global_index.hpp:17:23: error: use of undeclared identifier 'blockDim'
   17 |   return blockIdx.x * blockDim.x + threadIdx.x;
      |                       ^
/__w/acts/acts/Traccc/device/hip/src/clusterization/../sanity/../utils/global_index.hpp:17:36: error: use of undeclared identifier 'threadIdx'
   17 |   return blockIdx.x * blockDim.x + threadIdx.x;
      |  
```

and
```
In file included from /__w/acts/acts/Traccc/device/hip/src/clusterization/clusterization_algorithm.hip:9:
/__w/acts/acts/Traccc/device/hip/src/clusterization/../sanity/contiguous_on.hpp:136:3: error: no matching function for call to 'is_contiguous_on_compress_adjacent'
  136 |   hipLaunchKernelGGL(kernels::is_contiguous_on_compress_adjacent<CONTAINER>,
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  137 |                      (n + block_size - 1) / block_size, block_size, 0,
      |                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  138 |                      hip_stream, projection, view, iout);
      |                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/rocm-6.1.0/lib/llvm/bin/../../../include/hip/amd_detail/amd_hip_runtime.h:244:73: note: expanded from macro 'hipLaunchKernelGGL'
  244 | #define hipLaunchKernelGGL(kernelName, ...)  hipLaunchKernelGGLInternal((kernelName), __VA_ARGS__)
      |                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/rocm-6.1.0/lib/llvm/bin/../../../include/hip/amd_detail/amd_hip_runtime.h:241:9: note: expanded from macro 'hipLaunchKernelGGLInternal'
  241 |         kernelName<<<(numBlocks), (numThreads), (memPerBlock), (streamId)>>>(__VA_ARGS__);         \
      |         ^~~~~~~~~~
/__w/acts/acts/Traccc/device/hip/src/clusterization/clusterization_algorithm.hip:38:10: note: in instantiation of function template specialization 'traccc::hip::is_contiguous_on<traccc::edm::silicon_cell<vecmem::edm::device<vecmem::edm::schema<vecmem::edm::type::vector<const unsigned int>, vecmem::edm::type::vector<const unsigned int>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const unsigned int>>, vecmem::edm::container<traccc::edm::silicon_cell, vecmem::edm::type::vector<unsigned int>, vecmem::edm::type::vector<unsigned int>, vecmem::edm::type::vector<float>, vecmem::edm::type::vector<float>, vecmem::edm::type::vector<unsigned int>>::interface_type>>, traccc::cell_module_projection, vecmem::edm::view<vecmem::edm::schema<vecmem::edm::type::vector<const unsigned int>, vecmem::edm::type::vector<const unsigned int>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const unsigned int>>>>' requested here
   38 |   return is_contiguous_on<edm::silicon_cell_collection::const_device>(
      |          ^
/__w/acts/acts/Traccc/device/hip/src/clusterization/../sanity/contiguous_on.hpp:41:17: note: candidate template ignored: constraints not satisfied [with CONTAINER = traccc::edm::silicon_cell<vecmem::edm::device<vecmem::edm::schema<vecmem::edm::type::vector<const unsigned int>, vecmem::edm::type::vector<const unsigned int>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const unsigned int>>, vecmem::edm::container<traccc::edm::silicon_cell, vecmem::edm::type::vector<unsigned int>, vecmem::edm::type::vector<unsigned int>, vecmem::edm::type::vector<float>, vecmem::edm::type::vector<float>, vecmem::edm::type::vector<unsigned int>>::interface_type>>, P = traccc::cell_module_projection, VIEW = vecmem::edm::view<vecmem::edm::schema<vecmem::edm::type::vector<const unsigned int>, vecmem::edm::type::vector<const unsigned int>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const unsigned int>>>, S = unsigned int]
   41 | __global__ void is_contiguous_on_compress_adjacent(
      |                 ^
/__w/acts/acts/Traccc/device/hip/src/clusterization/../sanity/contiguous_on.hpp:39:44: note: because substituted constraint expression is ill-formed: no matching function for call to 'declval'
   39 |                                   decltype(std::declval<CONTAINER>().at(0))> &&
      |                                            ^
```

I didn't immediately see a fix, so I leave it to @krasznaa if he has time.

krasznaa:
:thinking: Fixing the `blockIdx` issue is easy, and absolutely necessary.

But the others are a bit more questionable. :frowning: Those are issues that I can indeed reproduce with ROCm 6.2. But with 7.2 they just don't show up.
  - I see an issue with ROCm 6.X not willing to use `std::optional` in device code. But since that's used fairly heavily in the clusterization code, I'm not necessarily keen on simplifying that code for ROCm 6.X's benefit.
  - The issue you list here, about the compiler not finding the templated kernels in the sanity checking code, come from the non-trivial `require` statements put onto those kernels.
    * By just removing those concepts/conditions, the compilation succeeds with ROCm 6.2. But I'm not keen on either removing those, or even hiding them behind preprocessor macros. :thinking:

So... How would you feel about a ROCm 7.0 minimum requirement for this project @StewMH?

StewMH:
Happy to move to 7.2, but will need to update the CI containers.

paulgessinger:
I'm also looking into this build with spack dependencies now, and see at least the same blockIdx issue. This will then also allow us to not have to update the CI containers anymore.

krasznaa:
> I'm also looking into this build with spack dependencies now, and see at least the same blockIdx issue. This will then also allow us to not have to update the CI containers anymore.

Please approve https://github.com/acts-project/acts/pull/5857 then, so that we can sort out the more interesting part. 😉

paulgessinger:
@krasznaa right I discovered that one too. I'm trying to work through the other compiler failures on HIP 6.4.4 now

krasznaa:
> Happy to move to 7.2, but will need to update the CI containers.

Also note that for ATLAS we are preparing ROCm 7.2.4 to become part of out LCG layers. So focusing on older versions, purely for ATLAS, is not much of a priority.

At the same time, the ROCm versioning is beyond ridiculous see: https://rocm.docs.amd.com/en/latest/release/versions.html Also, I tried to install 7.14.0 with their "runfile installer", but failed so far. So right now 7.0.Y - 7.2.Y are the versions that I can actually install in a relocatable way, with normal user rights.

StewMH:
OK, now we need #5856

StewMH:
Hits an internal compiler error (ICE), same as I reported locally https://github.com/acts-project/acts/issues/5877

StewMH:
OK, by switching to RelWithDebInfo (i.e. O2 to avoid crash in O3 optimisation) we're down to
```
/__w/acts/acts/build/bin/traccc_test_hip: error while loading shared libraries: libamdhip64.so.7: cannot open shared object file: No such file or directory
```
in traccc_test_hip. I did have a quick look at this but no luck yet.

krasznaa:
> OK, by switching to RelWithDebInfo (i.e. O2 to avoid crash in O3 optimisation) we're down to
> 
> ```
> /__w/acts/acts/build/bin/traccc_test_hip: error while loading shared libraries: libamdhip64.so.7: cannot open shared object file: No such file or directory
> ```
> 
> in traccc_test_hip. I did have a quick look at this but no luck yet.

This seems to just be library loading 101.

`/opt/rocm/lib` is not part of the library lookup paths by default. Even when you install ROCm using a DEB/RPM package. (I think one of the very top level packages may set it up, but when only installing the packages that you actually need/want, it definitely doesn't get set up.)

As with anything else, there are 2 ways of making the system look for libraries under that directory:
  - Add it to `$LD_LIBRARY_PATH`, let's say as part of `ci_setup.sh`;
  - Add a file like `/etc/ld.so.conf.d/rocm.conf` to the image, with the content:

```
/opt/rocm/lib
```

Though I admit I don't have hands-on experience with using `ld.so` inside of an image. It may be enough to just put this file in place. Though on a real machine you also need to `sudo ld.so` after such a configuration change, for the change to become visible to all users.

So long story short, you probably just want to add

```sh
export LD_LIBRARY_PATH=/opt/rocm/lib${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}
```

somewhere before the Alpaka+HIP build would start.

.github/workflows/traccc.yml
```diff
@@ -113,6 +113,15 @@ jobs:
               enable_cuda: true
               enable_hip: false
             build: Release
+          - platform:
+              name: "ALPAKA_HIP"
+              container: ghcr.io/acts-project/ubuntu2404_rocm:90
+              options: --preset alpaka-fp32 -Dalpaka_ACC_GPU_HIP_ENABLE=ON -DCMAKE_PREFIX_PATH=/opt/rocm/lib/cmake/ -Dalpaka_DISABLE_VENDOR_RNG=ON -DTRACCC_USE_ROOT=OFF -DTRACCC_SETUP_ROCTHRUST=ON -DTRACCC_BUILD_HIP=ON -DCMAKE_HIP_COMPILE_FEATURES=hip_std_20 -DDETRAY_BUILD_VALIDATION_TOOLS=ON
```
krasznaa: I believe

```
-DCMAKE_PREFIX_PATH=/opt/rocm
```

would be a little better. CMake can find the `lib/cmake` postfix by itself.

As for the C\+\+ standard, I myself was using `-DCMAKE_HIP_STANDARD=20` instead. Though this probably doesn't make much of a difference for the CI. 🤔

StewMH: Fixed, thank you. I thought there was some trick I was missing.

paulgessinger (approved):

PR #5779, opened 2026-07-28
feat: expose readJson in TGeo pybind

osbornjd:
This exposes the readJson function into the pybind TGeo implementation so that the python scripts can build detector geometries using TGeo with an input JSON configuration file.


paulgessinger (approved):

osbornjd:
@paulgessinger do I need to do anything else to add this to the merge queue?

paulgessinger:
@osbornjd no, my bad!

PR #5785, opened 2026-07-29
refactor: Allocate GSF smoothed parameters in the backward pass

andiwand:
The GSF allocated the smoothed parameters of the combined state during the
forward pass and filled them with a `-2` sentinel until the backward pass
computed them. States whose surface the backward pass never reached were
demoted to outliers and had the component unset again.

Allocate the smoothed parameters where they are computed instead. This removes
the sentinel writes and the cleanup pass, and makes `hasSmoothed()` line up with
"the backward pass produced smoothed parameters here" for the whole duration of
the fit.

Same class of change as the filtered parameters in #5784, see #5777.

--- END COMMIT MESSAGE ---

Stacked on top of #5784, please review/merge that one first.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5786, opened 2026-07-29
chore: Harden track state component allocation

andiwand:
Three inconsistencies around `addTrackStateComponents`:

- `MutableMultiTrajectoryBackend` did not require it, even though every fitter
  now calls `TrackStateProxy::addComponents()`. A third-party backend satisfied
  the concept and only failed at the first instantiation.

- `VectorMultiTrajectory` decided whether a component was present from the
  stored `allocMask`, while the Podio backend uses the component index.
  `shareFrom_impl` assigns an index without touching `allocMask` and
  `unset_impl` clears an index without touching it either, so the two can
  disagree: sharing a component and then adding it replaced the shared index,
  and unsetting a component and then adding it allocated nothing. Decide from
  the index like Podio does, and keep `allocMask` for the storage this track
  state actually owns, which is what `statistics()` reports.

- `AnyTrackStateProxy` had the handler entry but never surfaced a public
  `addComponents()`.

Covered by a new shared backend test exercising both sequences.

--- END COMMIT MESSAGE ---

Stacked on top of #5784 and #5785, please review/merge those first.

### DIFF
--- a/Core/include/Acts/EventData/AnyTrackStateProxy.hpp
+++ b/Core/include/Acts/EventData/AnyTrackStateProxy.hpp
@@ -727,6 +727,15 @@ class AnyTrackStateProxy
     mutableHandler()->unset(mutableContainerPtr(), m_index, target);
   }
 
+  /// Add additional components to the track state.
+  /// @param mask Property mask describing which components to allocate.
+  void addComponents(TrackStatePropMask mask)
+    requires(!ReadOnly)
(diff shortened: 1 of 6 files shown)


junggjo9 (approved):

andiwand:
Closing — this was opened from a branch pushed to the upstream repository by mistake. Reopening from my fork.

PR #5788, opened 2026-07-29
fix(traccc): gpu-gbts small fixes

m-brettell:
Some fixes mostly for sparse (single particle) events, no behaviour change on events that ran correctly before.
- use reverse instead resize before push_back.
- Count number of bin pairs exactly and avoid uint underflow for empty bins.
- Update dropout curvature comments.
- Add inv_max_curvature cut to the cuts scaled by user pT config.

--- END COMMIT MESSAGE ---


andiwand:
please indicate that this PR is about GBTS traccc and not GBTS ACTS CPU

m-brettell:
> please indicate that this PR is about GBTS traccc and not GBTS ACTS CPU

Renamed to fix(traccc): gpu-gbts small fixes, thanks.

stephenswat (approved):

PR #5790, opened 2026-07-29
refactor: Look up the start volume from scratch on re-initialization

andiwand:
If a navigation state is re-initialized, `Navigator::initialize` had a fast path that reused `state.startVolume` from the previous navigation whenever the start surface did not provide an associated layer (`resetForRenavigation` deliberately keeps the start* members).

Re-initializing means the navigation got lost and has to start over, so carrying state over can poison the new lookup: the cached volume does not have to contain the current position, and the start layer is then searched in the wrong volume. The volume is now always resolved from the position, unless the start surface hands us the layer (and with it the volume) directly.

The navigator unit test that set a start volume up front now sets a *stale* one, i.e. a volume that does not contain the start position, and checks that the lookup still resolves the volume at the position. It fails without this change.

--- END COMMIT MESSAGE ---

This path is reachable from outside Core: an earlier attempt to remove it (as part of #5740) changed physmon output, so this is split out into its own PR to see the physmon effect in isolation.

### DIFF
--- a/Core/src/Navigation/Navigator.cpp
+++ b/Core/src/Navigation/Navigator.cpp
@@ -150,13 +150,6 @@ Result<void> Navigator::initialize(State& state, const Vector3& position,
 
     state.startLayer = state.startSurface->associatedLayer();
     state.startVolume = state.startLayer->trackingVolume();
-  } else if (state.startVolume != nullptr) {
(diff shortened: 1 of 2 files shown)


andiwand:
after looking at the diff I don't think this is really relevant - closing

PR #5793, opened 2026-07-29
ci: move the long every-PR jobs to husk, cancel superseded push runs

paulgessinger:
Follows #5723 (`linux_ubuntu`) and #5756 (LCG matrix): moves the five long jobs that run on every PR onto husk.

**Needs on the fleet side:** `sw-nightlies.hsf.org` bound into the job container on `cvmfs`-labelled slots, for `external_key4hep-nightlies`. Without it that job fails at its `Check CVMFS` step with a named message. Nothing else here depends on fleet changes.

### Why

Queue delay, not runtime, is what these jobs cost. Wait time (`started_at - created_at`) across successful `checks.yml` jobs:

| day | n | median | p90 |
| --- | --- | --- | --- |
| Jul 23 | 160 | 23.4 m | 131 m |
| Jul 24 | 235 | 26.9 m | 80 m |
| Jul 27 | 175 | **94.1 m** | 175 m |
| Jul 28 | 260 | **69.7 m** | 217 m |

Husk over the same window: 0–9 min. A 90 s slot recycle is noise at that scale. Note husk is not *faster* per job — `linux_ubuntu` on upstream runs is 14.1 min median there vs 10.6 min hosted (n=12) — this buys latency, not throughput.

### What moves

| job | median | p90 |
| --- | --- | --- |
| `build_debug` (analysis.yml) | 59.1 m | 61.0 m |
| `linux_physmon` | 28.6 m | 29.7 m |
| `external_key4hep-nightlies` | 20.5 m | 23.4 m |
| `linux_examples_test` | 16.0 m | 17.1 m |
| `linux_ubuntu_extra` | 8.4 m | 70.0 m (ccache cold) |

Four are a one-line `runs-on`. `linux_examples_test` and `linux_physmon` are the follow-ups #5723 named; both already consume an artifact built on a husk slot.

`linux_physmon` additionally applies `./.github/actions/apt_mirror` itself, because it apt-installs `time` before the dependencies action gets there, and the upstream Ubuntu archive measures 48 kB/s from CERN (#5723).

### `external_key4hep-nightlies`

Runs in `ghcr.io/acts-project/alma9-base:88` and sources `/cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh` directly, using the `cvmfs` label — no CVMFS client install, no nested container, neither of which works under rootless podman. The compile payload is unchanged apart from quoting `$(pwd)` and `$(nproc)`, which shellcheck can now see. Its ccache uses the same sha + prefix-restore keys as the other jobs, so it shares with the parent commit instead of rotating daily, and `Check CVMFS` asserts the one repository the job actually reads rather than the three the previous action demanded.

The check now builds against the **alma9** nightly rather than ubuntu24. The stack is concretized for both — key4hep's own matrix is `[alma9, ubuntu24]` — `setup.sh` picks the flavour matching the container, and el9 is what the CERN side of the user base runs. It also means a `cvmfs` slot pulls one image instead of two, since the LCG legs already use `alma9-base`. That image carries no ccache, hence the one `dnf install`; key4hep's own build images install it at image level too, so the stack does not supply it.

### Images through the CERN registry cache

Every husk job — including `linux_ubuntu` and the LCG legs, which were already there — now pulls `registry.cern.ch/ghcr.io/acts-project/…` instead of `ghcr.io/…`. The slots sit at CERN behind a shared egress, where anonymous ghcr.io pulls run into the rate limit; `.gitlab-ci.yml` uses the same proxy from inside CERN. GitHub-hosted jobs are untouched. `CI/bump_versions.py` already treats the `registry.cern.ch/` prefix as optional, so tag tracking is unaffected.

### Cancelling superseded push runs

Second commit, same motivation: `github.head_ref` is only set on `pull_request`, so on a push the concurrency group fell back to `github.run_id` — unique per run, so nothing was ever cancelled and back-to-back merges each held a full set of runners to completion. Keying on `github.ref` groups per branch, so a merge cancels the run of the commit it supersedes. `traccc.yml` also picks up the `github.workflow` prefix the other workflows have, without which ref-based groups from different workflows would collide.

Consequence: on `main`, `linux_physmon_perf_report` records a metrics point for the newest commit of a burst rather than for every commit.

### Caveats

- **Capacity:** ~5 more husk jobs per PR on top of `linux_ubuntu` and the 4 LCG legs. If the pool can't absorb that, husk queueing just replaces GitHub queueing.
- **physmon metrics:** `linux_physmon_perf_report` pushes memory/runtime maxima to `runtime_metrics` on `main`, so the series gets a one-time step at the hardware change. Probably an improvement — fewer unknown neighbours — but it should be on the record.
- **Unverified until this runs:** that the key4hep stack overrides the `CXX` the alma9 image exports (`gcc-toolset-15`). The LCG legs are evidence it does — the clang19 leg's CMake picks the view's `clang++` over the same image env — but the key4hep stack is a different setup script.

### Not moved

**SonarCloud:** 0.1 min median queue (`workflow_run` doesn't contend) and 6 min runtime, so there is nothing to save — and it is the one job holding `SONAR_TOKEN` while checking out an attacker-influenced merged tree.

**`external_eic-shell`** (4 min) and the sub-3-minute `checks` jobs are too short to be worth a slot. **traccc** (8 legs × 18–31 min) and **detray** `host-container` (25 min) amortize well but are path-filtered and would arrive as an 8-slot burst with multi-GB CUDA/oneAPI pulls — worth revisiting once the load from this PR is known.

### Checked locally

`prek run` clean incl. zizmor. `actionlint`: 15 shellcheck findings, identical to `main`; the new `runner-label` warnings are the usual unknown-custom-label ones already produced by `linux_ubuntu` and the LCG legs.

https://claude.ai/code/session_0121kZNBbiwQo4ZGDu8SUB48

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -31,7 +31,7 @@ permissions:
 
 jobs:
   build_debug:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, linux, x64, husk-size-standard]
     container: ghcr.io/acts-project/ubuntu2404:87
     steps:
 
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -13,7 +13,7 @@ on:
       - "docs/**"
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
 
 env:
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -32,7 +32,8 @@ permissions:
 jobs:
   build_debug:
     runs-on: [self-hosted, linux, x64, husk-size-standard]
-    container: ghcr.io/acts-project/ubuntu2404:87
+    # Pull-through cache, as for the husk jobs in builds.yml.
+    container: registry.cern.ch/ghcr.io/acts-project/ubuntu2404:87
     steps:
 
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -837,9 +837,10 @@ jobs:
     runs-on: [self-hosted, linux, x64, husk-size-standard, cvmfs]
     permissions:
       contents: read
-    # Ubuntu 24.04 to match what the view is built against; ours rather than
-    # aidasoft's so CI/bump_versions.py keeps the tag current.
-    container: ghcr.io/acts-project/ubuntu2404:87
+    # AlmaLinux 9, same image as the LCG legs: the nightly stack is built for
+    # almalinux9 as well as ubuntu24 and setup.sh picks the flavour matching the
+    # container, so this is the one the CERN side of the user base runs.
+    container: ghcr.io/acts-project/alma9-base:88
     env:
       KEY4HEP_VIEW: /cvmfs/sw-nightlies.hsf.org/key4hep
       # Every run compiles a fresh checkout, so ctime/mtime differ even where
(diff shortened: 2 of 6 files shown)


andiwand (dismissed):

paulgessinger:
I need to mount the key4hep repo on the runners

andiwand (dismissed):

paulgessinger:
@andiwand it's working now.

andiwand (approved):

PR #5795, opened 2026-07-29
fix: Navigation initialization with gen3 geometry

benjaminhuth:
In case `startVolume` is set, and the navigation is re-initialized (e.g., branching in CKF), gen3 navigation would silently take a gen1-only path. This is mitigated by explicitly enforcing gen1 for gen1 passes.


benjaminhuth:
@paulgessinger @andiwand

andiwand (approved):

PR #5796, opened 2026-07-29
feat: Track reading for ColliderML, CI with small dataset sample

benjaminhuth:
Adds option to read in the published tracks of the ColliderML dataset. This allows: 
- Evaluation plots using ACTS tools
- Refit using ACTS algorithms. No covariances for final parameters are available, so pull plots will fail.

Also updates the CI to use a small sample of the real dataset.


Python/Examples/tests/test_colliderml_reader.py
```diff
@@ -6,185 +8,57 @@
...
-
-@pytest.fixture(scope="session")
-def colliderml_fatras_sample(tmp_path_factory):
-    """Generate a small ColliderML-compatible parquet dataset using FATRAS + ODD.
+sys.path.insert(0, str(_srcdir / "Examples/Scripts/Python"))
+from colliderml_truth_tracking import runColliderMLTruthTracking
 
-    Runs a minimal sequencer: ParametricParticleGenerator → FATRAS → Digitization
-    → ArrowParticleOutputConverter + ArrowSimHitOutputConverter → ParquetWriter.
+_COLLIDERML_DATA_ENV_VAR = "COLLIDERML_DATA"
```
paulgessinger: The env var is how the CI config communicates where this is downloaded?

Python/Examples/tests/test_colliderml_reader.py
```diff
@@ -216,23 +90,50 @@ def test_colliderml_truth_tracking(tmp_path, colliderml_fatras_sample, reco_geo)
...
+        if perfWriter is not None:
+            s.addWriter(perfWriter)
         s.run()
 
     if ctx is not None:
         with ctx:
             _run()
     else:
         _run()
+
```
paulgessinger: Not in the diff here, but what does this do? Why not just a nullcontext?

paulgessinger (commented):
Looks reasonable. We'll want to take a look at how long the job runs just to make sure it's sensible.

Python/Examples/tests/test_colliderml_reader.py
benjaminhuth: Yeah, similar to the GNN model stuff. But probably, we could hardcode this and don't need variables for that

Python/Examples/tests/conftest.py
```diff
@@ -96,10 +95,11 @@ def fn(key: str, file: Path):
         gkey = f"{request.node.name}__{key}"
         act_hash = helpers.hash_root.hash_root_file(file)
         if not gkey in root_file_exp_hashes:
-            warnings.warn(
-                f'Hash lookup key "{key}" not found for test "{request.node.name}"'
+            check.equal(
+                act_hash,
+                "[MISSING]",
+                msg=f'Hash lookup key "{key}" not found for test "{request.node.name}"',
```
paulgessinger, line 101: What does this do?

benjaminhuth: The changevthat made warnings fatal in the CI has the implication, that hashes of missing files are not printed anymore, because it aborts immediatly. Removing UserWarning here should be fine I think, the job fails anyways

.github/workflows/builds.yml
```diff
@@ -138,6 +138,7 @@ jobs:
       # locations resolve, so the test_fpe.py masking tests run instead of being
       # skipped. Replaces the FPE coverage previously unique to GitLab.
       ACTS_SEQUENCER_FAIL_ON_UNMASKED_FPE: 1
+      COLLIDERML_DATA: ${{ github.workspace }}/ci_colliderml
```
andiwand: instead of wiring an ENV couldn't we just pass that as an argument to the download script?

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
```diff
@@ -141,6 +146,20 @@ ColliderMLRelease1InputConverter::particleSchema() {
...
+  return ActsPlugins::ArrowUtil::ArrowSchemaHandle{arrow::schema({
+      arrow::field("d0", arrow::list(arrow::float32()), false),
+      arrow::field("z0", arrow::list(arrow::float32()), false),
+      arrow::field("phi", arrow::list(arrow::float32()), false),
+      arrow::field("theta", arrow::list(arrow::float32()), false),
+      arrow::field("qop", arrow::list(arrow::float32()), false),
+      arrow::field("majority_particle_id", arrow::list(arrow::uint64()), false),
+      arrow::field("hit_ids", arrow::list(arrow::list(arrow::uint32())), false),
+      arrow::field("track_id", arrow::list(arrow::uint16()), false),
+  })};
```
andiwand, line 160: should this rather be upstreamed into the Arrow plugin or a potential new colliderml plugin?

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
```diff
@@ -530,6 +564,104 @@ ProcessCode ColliderMLRelease1InputConverter::execute(
     if (m_outputMeasParticlesMap.isInitialized()) {
       m_outputMeasParticlesMap(ctx, std::move(measParticlesMap));
     }
+
+    if (m_outputTracks.isInitialized()) {
```
andiwand, line 568: the body might be worth its own function if the output is easy to capture

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
```diff
@@ -530,6 +564,104 @@ ProcessCode ColliderMLRelease1InputConverter::execute(
...
+
+      auto trackContainer = std::make_shared<Acts::VectorTrackContainer>();
+      auto trackStateContainer =
+          std::make_shared<Acts::VectorMultiTrajectory>();
+      TrackContainer tracks(trackContainer, trackStateContainer);
+
+      // Shared reference surface for every published track: ColliderML
+      // reports perigee parameters w.r.t. the detector origin.
+      auto perigee = Acts::Surface::makeShared<Acts::PerigeeSurface>(
+          Acts::Vector3::Zero());
```
andiwand, line 598: I think we do this in a couple of places but should we rather make this a config property?

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
```diff
@@ -530,6 +564,104 @@ ProcessCode ColliderMLRelease1InputConverter::execute(
...
+              track.appendTrackState(Acts::TrackStatePropMask::None);
+          trackStateProxy.typeFlags().setIsMeasurement();
+          trackStateProxy.setReferenceSurface(hitSurface->getSharedPtr());
+          trackStateProxy.setUncalibratedSourceLink(
+              Acts::SourceLink(sourceLink));
+        }
+
+        track.nMeasurements() = static_cast<std::uint32_t>(nTrackHits);
+        track.nHoles() = 0;
+        track.nOutliers() = 0;
```
andiwand, line 652: these are not recorded at all?

Python/Examples/tests/test_colliderml_reader.py
```diff
@@ -216,23 +90,44 @@ def test_colliderml_truth_tracking(tmp_path, colliderml_fatras_sample, reco_geo)
...
+        "colliderml_tracks", name="check_colliderml_tracks", level=acts.logging.WARNING
+    )
+
+    perfFile = tmp_path / "performance_colliderml_truth_tracking.root"
+    perfWriter = None
+    if reco_geo == "gen1":
+        RootTrackFinderPerformanceWriter = acts.examples._tryImportRoot(
+            "RootTrackFinderPerformanceWriter"
+        )
+        perfWriter = RootTrackFinderPerformanceWriter(
```
andiwand: I guess this can be replaced by the new python writer?

Python/Examples/tests/test_colliderml_reader.py
```diff
@@ -216,23 +90,44 @@ def test_colliderml_truth_tracking(tmp_path, colliderml_fatras_sample, reco_geo)
...
         ctx = detector
 
-    def _run():
+    check_alg = AssertCollectionExistsAlg(
+        "colliderml_tracks", name="check_colliderml_tracks", level=acts.logging.WARNING
+    )
+
+    perfFile = tmp_path / "performance_colliderml_truth_tracking.root"
+    perfWriter = None
+    if reco_geo == "gen1":
```
andiwand, line 94: why only gen1?

Python/Examples/tests/test_colliderml_reader.py
```diff
@@ -6,185 +9,56 @@
...
+        reason=(
+            f"ColliderML CI sample not found; set COLLIDERML_DATA to a "
+            f"directory containing the '{_SAMPLE}' sample"
+        ),
+    ),
+]
 
 
+@pytest.mark.root
 @pytest.mark.parametrize("reco_geo", ["gen1", "gen3"])
```
andiwand, line 43: is there an enum we bind to python for gen1 / gen3?

Python/Examples/tests/test_colliderml_reader.py
```diff
@@ -6,185 +9,56 @@
...
 
-    Returns (base_dir, sample_name).
-    """
-    from acts.arrow import particleSchema, simHitSchema
-    from acts.examples.arrow import (
-        ArrowParticleOutputConverter,
-        ArrowSimHitOutputConverter,
-        ParquetWriter,
+_colliderml_data_dir = os.environ.get("COLLIDERML_DATA")
+if _colliderml_data_dir is not None:
```
andiwand: ok I see the ENV kicks in here. can we encapsulate that similar to how it is done for the ODD? something like `getColliderMLData` and a mechanism that tells you if it is around or not other than direct ENV access

Python/Examples/tests/conftest.py
paulgessinger: Maybe worth making as a standalone change rather than bundling. 

I think originally I used warnings because that prints the whole list at the end and is easy to copy

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
benjaminhuth: Hmm I would rather have it local, as it is really exactly only needed in this reader, and should be independent of our track schema we have in the examples framework (which might evolve over time, this here should not).

.github/workflows/builds.yml
benjaminhuth: So this is the pattern we established for the GNNs, where I went from an argument based pattern to environment variable based pattern, I think as suggested by @paulgessinger back then.
In principle happy to swap back, but we should stay consistent...

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
benjaminhuth: Hmm, I think this should not be configurable, as this is a property of the datset that the data of the tracks are expressed here, right?

Python/Examples/tests/test_colliderml_reader.py
benjaminhuth: yeah, but would like to take advantage of the hash check, which does not work with python as of now.

Examples/Io/Arrow/src/ColliderMLRelease1InputConverter.cpp
benjaminhuth: nope

Python/Examples/tests/test_colliderml_reader.py
benjaminhuth: In principle, I want to run the gen3 test in the pypi build test, but there we don't have root writer, and thus no hash checks.

Python/Examples/tests/test_colliderml_reader.py
benjaminhuth: Hmm I think string is fine for pytest, right? this does not enter c++ land I think

Python/Examples/tests/test_colliderml_reader.py
benjaminhuth: yeah, maybe a good idea, I think about it.

Python/Examples/tests/test_colliderml_reader.py
andiwand: yeah not crucial - I only faintly remembered that we have this enum https://github.com/acts-project/acts/blob/5111f12398037cddb6016cf1204fcab1fb21d52e/Core/include/Acts/Geometry/TrackingGeometry.hpp#L297

andiwand (dismissed):

.github/workflows/builds.yml
paulgessinger: Command line argument is hard to get into pytest

andiwand (dismissed):

andiwand (approved):

PR #5797, opened 2026-07-29
refactor!: pass navigation start and target surfaces to initialize

andiwand:
The navigation start and target surface used to live in the navigator options, which are bound when the navigation state is created. Navigation states are reused for further runs though - the combinatorial track finder re-initializes one on every branch jump - so per-run inputs bound at state construction go stale.

They now travel in a new `NavigatorInitializeArguments` struct handed to `Navigator::initialize`, together with the position, direction and an optional start volume hint. What remains in the options is only what is invariant across every run a state serves.

This also fixes a latent bug: `initialize` read `state.startVolume` as a fast-path input while the reset never cleared it. It now resolves the start information into locals, and the new `resetForInitialization` drops it explicitly - in contrast to `resetForRenavigation`, which runs mid-propagation and has to keep the start surface that actors rely on.

On the propagator side the target surface moves from `makeState` to `initialize` as well. Only the aborter type is needed to build the state, so the two `makeState` overloads collapse into one: the target aborter type is a template parameter defaulting to the new `NoTargetAborter` tag, and `initialize` hands the surface to both the aborter and the navigator.

This is breaking for code calling `Navigator::initialize`, `Propagator::initialize` or `Propagator::makeState` directly, but not user facing - these are driven through the propagator. Every break is a compile error at the call site, never a silent behaviour change, and all in-tree callers are updated.


Core/include/Acts/Propagator/DirectNavigator.hpp
```diff
@@ -215,28 +222,25 @@ class DirectNavigator {
   /// This function initializes the navigator for a new propagation.
   ///
   /// @param state The navigation state
-  /// @param position The start position
-  /// @param direction The start direction
-  /// @param propagationDirection The propagation direction
+  /// @param args The initialization arguments of this navigation run
   /// @return Always returns success result as DirectNavigator initialization cannot fail
-  [[nodiscard]] Result<void> initialize(State& state, const Vector3& position,
```
benjaminhuth, line 222: Hmm so one option would be to deprecate this initialize function and forward it to the new one until the next breaking release. That would avoid breaking entirely.

benjaminhuth (commented):
Looks good. Not sure if we should deprecate rather then remove here...

andiwand:
### Why the options were the wrong home

Config is fixed for the lifetime of an object, options may change per use. The navigator had grown a third, unnamed tier, because the navigation state deliberately outlives a single run - the candidate buffers, the `reserve(50)` on the Gen3 stream and the policy state manager all exist so a state can be recycled. Once the storage outlives the run, "bound at state construction" stops meaning "per run".

The steppers already got this right, and that is the model followed here: `StepperPlainOptions` holds only invariant configuration, the per-run start condition is an argument to `stepper.initialize`. Hence no equivalent bug there despite the same state reuse. The rule: options are what a reset keeps, initialize arguments are what a reset invalidates.

### Regression test

`Navigator_reinitialize_does_not_reuse_start_state` initializes one state twice, the second time with no start surface and a position in a different volume. Reproduces on Gen1, so no Gen3 fixture is needed. Verified to fail before the `initialize` change and to pass after.

### Note for reviewers

The GSF backward pass was the one in-tree user of the two-argument `makeState` outside `propagate`, and it needs the target both for its `MultiStepperSurfaceReached` aborter and for the navigator. Making the target parameter of `Propagator::initialize` non-defaulted was deliberate: a defaulted parameter would have let that call site compile with an implicit `nullptr` and silently stop the backward propagation from terminating at its target. The seven call sites with no target now pass an explicit `nullptr`.

Merged main, which brings in the boundary-aware start volume resolution from #5795: the start surface hint and the `isOnSurface` precondition check now run on the locals instead of the state.

Local `ctest`: 375/375 passing.

benjaminhuth (approved):
LGTM

PR #5801, opened 2026-07-30
chore: Harden track state component allocation

andiwand:
Three inconsistencies around `addTrackStateComponents`:

- `MutableMultiTrajectoryBackend` did not require it, even though every fitter
  now calls `TrackStateProxy::addComponents()`. A third-party backend satisfied
  the concept and only failed at the first instantiation.

- `VectorMultiTrajectory` decided whether a component was present from the
  stored `allocMask`, while the Podio backend uses the component index.
  `shareFrom_impl` assigns an index without touching `allocMask` and
  `unset_impl` clears an index without touching it either, so the two can
  disagree: sharing a component and then adding it replaced the shared index,
  and unsetting a component and then adding it allocated nothing. Decide from
  the index like Podio does, and keep `allocMask` for the storage this track
  state actually owns, which is what `statistics()` reports.

- `AnyTrackStateProxy` had the handler entry but never surfaced a public
  `addComponents()`.

Covered by a new shared backend test exercising both sequences.

--- END COMMIT MESSAGE ---

Supersedes #5786, which was accidentally opened from a branch in this
repository instead of my fork. Unlike the previous version this is no longer
stacked on #5784 and #5785 — it touches a disjoint set of files and can be
reviewed on its own.


Core/src/EventData/VectorMultiTrajectory.cpp
```diff
@@ -92,49 +92,56 @@ void VectorMultiTrajectory::addTrackStateComponents_impl(
...
   IndexData& p = m_index[istate];
-  PropMask currentMask = p.allocMask;
+
+  // Whether a component is present is decided by its index, not by `allocMask`:
+  // `shareFrom_impl` assigns an index without owning storage and `unset_impl`
+  // drops an index without releasing it, so the two can disagree. Sharing a
+  // component and then adding it must not silently replace the shared index.
+  // `allocMask` only tracks storage owned by this track state, which is what
+  // `statistics()` reports.
+  PropMask allocated = PropMask::None;
```
andiwand:
```suggestion
  PropMask allocated = PropMask::None;
```

Core/include/Acts/EventData/detail/MultiTrajectoryTestsCommon.hpp
benjaminhuth, line 1: Hmm not in the diff, but I'm a bit surprised to find this code here in Core...

Core/include/Acts/EventData/AnyTrackStateProxy.hpp
```diff
@@ -727,6 +727,15 @@ class AnyTrackStateProxy
     mutableHandler()->unset(mutableContainerPtr(), m_index, target);
   }
 
+  /// Add additional components to the track state.
+  /// @param mask Property mask describing which components to allocate.
+  void addComponents(TrackStatePropMask mask)
```
benjaminhuth, line 732: Hmm this is only used in the test, right?

benjaminhuth (approved):
Looks good, I have only a few minor question out of curiosity

PR #5802, opened 2026-07-30
fix: validate the BinnedGroup navigation array

andiwand:
The `navigation` array handed to `BinnedGroup` lists, per axis, which local bins to visit and in which order. Those indices are **1-based** — the default `BinnedGroup` fills in is a `std::iota` from `1ul`, and `GridBinFinder` looks up `val[locPosition[i] - 1ul]` — but nothing enforced it.

Getting it wrong failed quietly. Index `0` is the underflow bin, which is never filled, so listing it just drops a slice of the detector from the seeding loop with no error or warning. The only guard was an `assert` in `GridBinFinder::getSizePerAxis`, compiled out in exactly the optimized builds where seeding runs. An index past the last bin, or a repeated one, was equally unreported.

This validates the array in both constructors, throwing `std::invalid_argument` naming the axis, the bin and the valid range. The empty-vector default is unchanged, and the fill-in logic is factored into one helper.

--- END COMMIT MESSAGE ---

### Compatibility

All in-tree configs (`itk.py` pixel, pixel high-occupancy, strip) are already valid 1-based, so nothing that works today starts throwing. Same for the six seeding configurations that feed this array from Athena's `GridTripletSeedingTool` — though Athena's own guard is `i >= zBinEdges.size()`, which accepts `0`, so a matching fix is going in there separately.

### Note for reviewers

Rejecting **duplicates** is the one judgement call here rather than an obvious bug — say the word if you'd rather that stayed permissive and I'll drop it to out-of-range checking only.

### Tests

`BinnedGroup` had no unit test coverage, so this adds `Tests/UnitTests/Core/Seeding/BinnedGroupTests.cpp`: the three rejection cases, the mask-constructor path, and two visiting-order tests.


Core/include/Acts/Seeding/BinnedGroup.ipp
```diff
@@ -11,9 +11,58 @@
...
+/// Validate a navigation array and fill in the axes that were left undefined.
+///
+/// Entries are 1-based local bin indices; 0 is the underflow bin, which is
+/// never filled. An empty vector is replaced by 1, 2, ..., nBins.
+///
+/// @param navigation The navigation array, modified in place
+/// @param numLocalBins Number of local bins per axis
+/// @throw std::invalid_argument if an entry is out of range or repeated
+template <std::size_t DIM>
+void validateAndCompleteNavigation(
```
benjaminhuth:
```suggestion
void validateAndCompleteNavigationArray(
```
I think that naming would be clearer

benjaminhuth (dismissed):
LGTM, small naming suggestion

benjaminhuth (approved):
LGTM

PR #5803, opened 2026-07-30
ci: Trim the gnn_cpu build and skip the GNN jobs when unaffected

paulgessinger:
`gnn_cpu` configured with the `gitlab-ci-gnn` preset, which inherits `common` and
therefore builds the ODD, the DD4hep and ROOT examples, the integration tests and
all 340 unit test executables -- with `-w`, and without ever running a test. Of
the 944 translation units it compiled, 20 were GNN related; the rest duplicates
`linux_ubuntu`, which builds the same code with warnings as errors and runs its
tests.

Add a `github-ci-gnn-cpu` preset that switches off everything outside the GNN
build closure, which brings the job down to ~550 translation units and ~45
executables, and use the freed budget to actually run the CPU GNN unit tests
(only the CUDA ones need GitLab's GPU runner).

Skipping the Core unit tests needs a new `ACTS_BUILD_CORE_UNITTESTS` option, a
sibling of the existing `ACTS_BUILD_EXAMPLES_UNITTESTS`.

Both GNN jobs also gain a path filter, since each pulls a multi-GB
torch/CUDA/TensorRT image that most changes cannot reach. It is a job-level
condition rather than a `paths-ignore:` on a separate workflow so the checks keep
reporting under `Builds / *`, where merge-sentinel already requires them, as
`skipped` -- a state it demonstrably accepts. The filter lists what the two
presets actually compile, including the build system, which is the only way that
list can go stale: a change that grows the closure has to touch
`CMakePresets.json`, `CMakeLists.txt` or `cmake/`, so it runs the jobs and can fix
the list in the same breath. Over the last 80 merged PRs it would have skipped 24
of them. Push runs stay unfiltered to keep the shared ccache warm.

--- END COMMIT MESSAGE ---

### Measurements

Cold `gnn_cpu` run before this change (ninja log of run 30510735992): 944 C++ TUs,
353 executables linked of which 340 are unit tests, 8 python stub generations,
21 min of build in a 29 min job. Warm run: 108 s of build in a 9 min job, the rest
being image pull and dependency setup.

After, measured by configuring the equivalent option set locally: ~550 TUs, ~45
executables. Expected cold build ~11-12 min, warm ~50 s.

### On coverage

Everything dropped from the build is still covered elsewhere: `linux_ubuntu`
builds it with warnings as errors and runs the tests, and GitLab's `build_gnn`
still uses the unmodified `gitlab-ci-gnn` preset on every PR, so the full GNN
compile surface (CUDA + ModuleMap + torch + ONNX) is unchanged.

Two silent-failure modes closed along the way:

- `ctest` runs the whole registered set rather than a `-R` filter, so a GNN test
  added later cannot go unrun, plus `--no-tests=error` so an empty selection
  fails instead of passing green.
- Python stub generation stays on: importing `acts.gnn` is the only load-time
  check that the CPU-only module resolves its symbols.

### On the filter

Checked against the last 80 merged PRs (2026-07-07 to 07-30): it skips 24 of them.
Every one is verifiably outside the GNN build -- twelve touch only other workflow
files, three only `.pre-commit-config.yaml`, the rest `Detray/`, `Traccc/`,
`CI/physmon`, `CODEOWNERS`, `sonar-project.properties`. An equivalent deny list
("run unless everything in the change is on this list") was tried first and only
reached 8 of 80, because most changes mix one irrelevant directory with a file
nobody thought to list; it also needed 40 entries to the current 18.

The residual risk is the list going stale if either preset later gains a
component. Three things bound it: the build-system entries mean the change that
grows the closure always runs these jobs, unfiltered pushes to the long-lived
branches catch a stale entry on the next commit, and GitLab's unfiltered
`build_gnn` builds the superset on every PR.


.github/workflows/builds.yml
```diff
@@ -353,11 +353,77 @@ jobs:
...
+  # and the job's own definition. Those last entries are what make the list
+  # self-defending: the closure can only grow by editing CMakePresets.json,
+  # CMakeLists.txt or cmake/, so the change that grows it always runs these jobs
+  # and can update the list in the same breath.
+  #
+  # These are job-level `if:` conditions rather than a `paths-ignore:` on a
+  # separate workflow, so the checks keep reporting under `Builds / *` (as
+  # `skipped`) and merge-sentinel keeps requiring them.
+  # ─────────────────────────────────────────────────────────────────────
+  changes:
```
andiwand, line 371:
```suggestion
  gnn_relevant_changes:
```

andiwand (approved):
fine by me - pinging @benjaminhuth as this touches GNN

.github/workflows/builds.yml
paulgessinger: I think the job itself being called just "changes" makes sense, there's currently only one action that flags gnn changes, but I'll want to experiment with using this to drop builds potentially.

PR #5805, opened 2026-07-30
refactor: Use likelihood fit for resolutions

benjaminhuth:


benjaminhuth:
Seems not too relevant now...

PR #5810, opened 2026-07-30
feat: added a user data field in the traccc detector conditions

flg:
feat: added a user data field in the traccc detector conditions

--- END COMMIT MESSAGE ---

This is the field to store the Athena module ID hash so that we can access it during GPU measurement conversion.


flg:
Draft until I make the Athena side of things work.

stephenswat (dismissed):
I like it.

andiwand (approved):

PR #5815, opened 2026-07-31
fix(detray): allow homogeneous material on a sparse set of surfaces

paulgessinger:
## What

Lets the Detray conversion produce material on a **sparse** set of surfaces, and removes the workaround that was needed because it could not.

## Why

`homogeneous_material_factory::operator()` indexed its own data by `surface index - smallest surface index`:

```cpp
std::size_t sf_offset{*std::ranges::min_element(m_indices)};
for (auto [i, sf] : detray::views::pick(surfaces, m_indices)) {
  std::size_t sf_idx{i - sf_offset};
  const material<scalar_type> &mat = m_materials.at(sf_idx);
```

This silently assumes the surfaces carrying material form a contiguous block. Since `max(sf_idx) = max(m_indices) - min(m_indices) >= n-1`, with equality only when contiguous, any gap runs past the end of `m_materials` and throws `out_of_range`. So in practice only detectors where *every* surface carried material could be built.

The `DetrayPayloadConverter` worked around this by padding every volume with a dummy slab per surface (`// @HACK: Detray does not like homoegeneous material only on SOME surfaces.`). The filler could not be anything inert: `material_slab::operator bool()` rejects vacuum, zero thickness, `thickness == max()`, and zero density alike, and `consistency_checker` throws on any slab it rejects. So the padding used `thickness = 42, mat = {42, ...}` — roughly a radiation length of fictitious material on every surface that ACTS considered material-free, including surfaces whose real material is a grid.

## Changes

- **`homogeneous_material_factory`**: index the factory data by position in the sequence. The n-th entries of `m_indices`, `m_materials`, `m_thickness` and `m_links` all describe the same material instance, while `m_indices` holds the surface it belongs to. Surfaces carrying material no longer have to be contiguous.
- **`DetrayPayloadConverter`**: drop the dummy-slab padding. Also stop pinning `index_in_coll` to the surface index — that field is the position in the detector's material collection, and forcing the two to match made `insert_in_container` size the collection to the largest surface index and default-fill the gaps with invalid material. Left unset, the collection packs in payload order.
- **`DetrayGeometryConverter`**: only instantiate the homogeneous material reader when the target metadata supports slabs or rods (it hard-fails a `static_assert` otherwise), and check slab and rod support separately against what the payload actually contains. Only allocate the homogeneous payload when a volume contributes material, so a null payload means "none" rather than requiring callers to inspect contents.
- **`itk_metadata`**: enable homogeneous material, which the generated ITk metadata otherwise lacks (it supported only grid-based maps, so `DetrayGeometryConverter::convert()` could not be instantiated for it).

## Tests

New regression test `homogeneous_material_on_sparse_surfaces`: five surfaces, material on #1 and #4 — a gap, not starting at the first surface, not a trailing block.

- Without the factory fix: `vector::_M_range_check: __n (which is 3) >= this->size() (which is 2)`
- With it: passes.

`DetrayPayloadConverterTests` passes, including `check_consistency`. On the same geometry the slab count goes from 2517 (one per surface, i.e. all padding) to 2500 — 17 surfaces now correctly carry no homogeneous material, and detray accepts the sparse result.

Its assertions encoded the padding (`volumes.size() == 6`, "exactly one slab per surface") and were replaced with structural invariants: every registered volume exists and is non-empty, at most one entry per surface, each surface link in range and unique, `index_in_coll` unset.

https://claude.ai/code/session_015ZfKcVzPeWvRKHFGzVzXqc


paulgessinger:
@niermann999 what do we do with this?

Detray/core/include/detray/builders/homogeneous_material_factory.hpp
```diff
@@ -328,62 +328,63 @@ class homogeneous_material_factory final
...
-
-    DETRAY_DEBUG_HOST("-> sf_offset=" << sf_offset);
+    // Add the material to the surfaces that the data links against.
+    //
+    // The n-th entries of @c m_indices, @c m_materials, @c m_thickness and
+    // @c m_links all describe the same material instance, so the factory data
+    // is indexed by the position in that sequence, while @c m_indices holds
+    // the index of the surface the material belongs to. The surfaces that
+    // carry material therefore do not have to be contiguous in the volume.
+    for (const auto [n, i] : detray::views::enumerate(m_indices)) {
```
niermann999: Can we rename this, so that it becomes clearer which is the running index and which is the data? So, e.g.

```suggestion
    for (const auto [i, sf_idx] : detray::views::enumerate(m_indices)) {
```
and then current `i`->`sf_idx` and `n`->`i`

Detray/core/include/detray/builders/homogeneous_material_factory.hpp
```diff
@@ -328,62 +328,63 @@ class homogeneous_material_factory final
     }
     DETRAY_DEBUG_HOST("-> Indices: " << DETRAY_LOG_VECTOR(m_indices));
 
-    // Correctly index the data in this factory
-    std::size_t sf_offset{*std::ranges::min_element(m_indices)};
-
-    DETRAY_DEBUG_HOST("-> sf_offset=" << sf_offset);
+    // Add the material to the surfaces that the data links against.
+    //
+    // The n-th entries of @c m_indices, @c m_materials, @c m_thickness and
```
niermann999:
```suggestion
    // The i-th entries of @c m_indices, @c m_materials, @c m_thickness and
```

Detray/tests/integration_tests/cpu/builders/homogeneous_material_builder.cpp
```diff
@@ -179,6 +179,85 @@ GTEST_TEST(detray_builders, decorator_homogeneous_material_builder) {
...
 }
 
+/// Integration test: homogeneous material on a sparse subset of surfaces
+///
+/// Regression test: the material factory used to index its own data by
+/// (surface index - smallest surface index), which silently assumed that the
+/// surfaces carrying material form a contiguous block. For material on a
+/// non-contiguous set of surfaces that index ran past the end of the factory's
+/// material vector, so only detectors where every surface carried material
+/// could be built.
```
niermann999: Not really relevant to know in the future?
```suggestion
```

Detray/tests/integration_tests/cpu/builders/homogeneous_material_builder.cpp
```diff
@@ -179,6 +179,85 @@ GTEST_TEST(detray_builders, decorator_homogeneous_material_builder) {
...
+GTEST_TEST(detray_builders, homogeneous_material_on_sparse_surfaces) {
+  using transform3 = typename detector_t::transform3_type;
+  using material_id = typename detector_t::material::id;
+
+  using rectangle_factory = surface_factory<detector_t, rectangle2D>;
+  using mat_factory_t = homogeneous_material_factory<detector_t>;
+
+  vecmem::host_memory_resource host_mr;
+  detector_t d(host_mr);
```
niermann999, line 191: Can we add a dummy volume with a few surfaces here, so that we make sure the local surface indexing works?

Detray/tests/integration_tests/cpu/builders/homogeneous_material_builder.cpp
```diff
@@ -179,6 +179,85 @@ GTEST_TEST(detray_builders, decorator_homogeneous_material_builder) {
...
+      if (idx == 1u) {
+        EXPECT_EQ(slab.get_material(), silicon<scalar>());
+        EXPECT_NEAR(slab.thickness(), 1.f * unit<scalar>::mm, tol);
+      } else {
+        EXPECT_EQ(slab.get_material(), tungsten<scalar>());
+        EXPECT_NEAR(slab.thickness(), 2.f * unit<scalar>::mm, tol);
+      }
+    } else {
+      // Surfaces that were not given material must not have any
+      EXPECT_NE(mat_link.id(), material_id::e_material_slab);
```
niermann999: How about checking the same access that the navigator uses?
```suggestion
      EXPECT_FALSE(sf_desc.has_material());
```

niermann999 (dismissed):
Looks good from the detray side, just some minor comments

paulgessinger:
All done @niermann999

Detray/tests/integration_tests/cpu/builders/homogeneous_material_builder.cpp
```diff
@@ -179,6 +179,125 @@ GTEST_TEST(detray_builders, decorator_homogeneous_material_builder) {
...
+  // factory indices are volume local and have to be translated accordingly
+  constexpr std::size_t n_dummy_surfaces{3u};
+  constexpr std::size_t n_dummy_slabs{1u};
+  {
+    auto dummy_vbuilder =
+        std::make_unique<volume_builder<detector_t>>(volume_id::e_cylinder);
+    auto dummy_mat_builder =
+        homogeneous_material_builder<detector_t>{std::move(dummy_vbuilder)};
+
+    auto dummy_factory = std::make_shared<rectangle_factory>();
```
niermann999, line 205: I think we don't actually need a dummy factories, since e.g. `rect_factory.clear()` should reset the factory for reuse with the next volume

niermann999 (approved):

Detray/tests/integration_tests/cpu/builders/homogeneous_material_builder.cpp
paulgessinger: meaning the factory is reusable?

niermann999: At least that was the idea at some point, yes. Not sure I ever actually made use of it anywhere, so famous last words...

PR #5817, opened 2026-08-01
fix(cmake): avoid actsvg C++ standard cache leak

### LINKED ISSUE #4952: bug: cmake -DACTS_BUILD_PLUGIN_ACTSVG=on requires -DCMAKE_CXX_STANDARD=20
```
cmake -B build -S . -DACTS_BUILD_PLUGIN_ACTSVG=on
```
works fine, but if I rerun it on the same build directory, I get:
```
cmake -B build -S . -DACTS_BUILD_PLUGIN_ACTSVG=on

CMake Error at cmake/ActsCompilerOptions.cmake:47 (message):
  CMAKE_CXX_STANDARD=17, but ACTS requires C++ >=20
Call Stack (most recent call first):
  CMakeLists.txt:227 (include)
```

This can be fixed by adding an explicit `-DCMAKE_CXX_STANDARD=20`. Other CMake options I tried didn't affect this issue. I am using `LCG_108/x86_64-el9-gcc15-opt`, but saw the same issue with previous versions.

shuv-amp:
ActSVG caches `CMAKE_CXX_STANDARD=17`, which causes a second configure
in the same build directory to fail. Remove the cache entry after loading
ActSVG if it was not present before, leaving an existing value unchanged.

Add a CI check that configures ACTS with ActSVG twice.

Closes #4952

Tested repeated configuration with the default standard and with C++23,
and ran the ActSVG converter tests.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -53,6 +53,15 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           env_file: .env
 
+      - name: Check ActSVG reconfiguration
+        run: |
+          CI/dependencies/run.sh .env cmake -B build-actsvg-reconfigure -S . \
+            -GNinja \
+            -DACTS_BUILD_PLUGIN_ACTSVG=ON
+          CI/dependencies/run.sh .env cmake -B build-actsvg-reconfigure -S . \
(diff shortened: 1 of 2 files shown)


paulgessinger:
I think this should be fixed in actsvg, not here.

PR #5825, opened 2026-08-04
feat: Add TrackProxy conversion changes

yannikaMatt:
Implicit conversion from ConstTrackProxy to AnyConstTrackProxy and similiar types in pybind.
--- END COMMIT MESSAGE ---

@benjaminhuth 

Pybind11 cannot automatically convert a ConstTrackProxy into AnyConstTrackProxy, which most C++ functions in acts use. This conversion has to be specified explicitely.


Core/include/Acts/EventData/AnyTrackStateProxy.hpp
```diff
@@ -529,7 +529,7 @@ class AnyTrackStateProxy
                              AnyTrackStateProxy<true>> &&
              !std::is_same_v<std::remove_cv_t<track_state_proxy_t>,
                              AnyTrackStateProxy<false>>)
-  explicit AnyTrackStateProxy(track_state_proxy_t& ts)
+  explicit AnyTrackStateProxy(track_state_proxy_t ts)
```
paulgessinger: This means it makes a copy to then copy the internals. Is this required?

paulgessinger (commented):
I suppose the alternative would be to make the conversions explicit like in c++?

Core/include/Acts/EventData/AnyTrackStateProxy.hpp
benjaminhuth: Hmm, so in my opinion, a proxy is a small object that can be copied around. The lvalue reference leads to compile error with the python constructor of `AnyTrackStateProxy` from a vector-track-state-proxy.

Also, this is now symmetric with the `AnyTrackProxy` (also constructor by value)

benjaminhuth: I think the alternative would be `&&` or `const &`, but only `&` feels really odd here I think...

paulgessinger: It's more than a single pointer so there's a runtime overhead, although it's likely small. 

If the const ref works with the constness then that's probably it

benjaminhuth: Hmm as soon as you use `AnyTrackStateProxy`, I would say we are not on a hot path anyway, so construction by value is fine. (for reference, I think the thing has 3x64bit). 

If we think we need to optimze this, const ref would be fine I think, but for consistency, this should then be changed for `AnyTrackProxy` as well @paulgessinger

paulgessinger: Ok but then why not just leave it with the ref as is? You're copying to make another copy right after. I don't really get it

benjaminhuth: Because it does not compile with the mutable ref. A workaround would be (I think) to do something like
```c++
.def(py::init<ConstTrackStateProxy>([](auto proxy) {
	return AnyTrackStateProxy(proxy);
});
```
Then I think the lvalue ref would work, but it would just move the copy to the pybind side. But maybe @yannikaMatt  can post the compile error if you are interested.

The thing is: I really don't see why we want to have a mutable lvalue ref in this place. This suggests that the proxy can be modified by the constructor.

andiwand: I suppose all of this will be inlined anyways?

since the parameter is templated I think a universal reverence would be most generic

paulgessinger: If universal reference works that's probably the best. Const ref does not work?

yannikaMatt: So the actual compile error is quite long. But I think just the beginning and end are most helpful: 
`error: no matching constructor for initialization of 'Acts::AnyTrackStateProxy<false>' 83 | return new Class{std::forward<Args>(args)...};` ,
then some notes in between and finally: 
`note: candidate constructor [with track_state_proxy_t = Acts::TrackStateProxy<Acts::VectorMultiTrajectory, 6, false>] not viable: expects an lvalue for 1st argument 532 | explicit AnyTrackStateProxy(track_state_proxy_t& ts)`

yannikaMatt: But I will try the const ref to see if it works then.

benjaminhuth: Okay, const ref does not work, actually makes sense, because the proxy do not follow pointer semantics.
We'll try universal reference.

paulgessinger: I'm still not sure: did you try writing the construction explicitly in python?

```python
tp = getTrackState()
atp = AnyTrackStateProxy(tp)
```

benjaminhuth: Okay, turns out `py::classh<Acts::AnyMutableTrackStateProxy>(mex, "AnyMutableTrackStateProxy")
    .def(py::init<TrackStateProxy &>());` works with lvalue ref.
This would allow us to keep Core untouched. Is this a viable solution @paulgessinger ?

From the design point-of-view I would say that implicit conversion is desirable in python land (and is anyways othogonal to the reference issue).

benjaminhuth (approved):
LGTM

PR #5826, opened 2026-08-04
perf: Speed up the CKF stayOnSeed candidate filter

andiwand:
`stayOnSeed` checked every candidate on a surface by walking the seed's space points and comparing type erased source links, and copied the candidate vector to apply the result. On ttbar pu200 there are ~23 candidates per surface, which made this the largest single item in track state creation.

Resolve the seed to measurement indices once in `setSeed` and partition the candidates in place. Selection is unchanged: the measurement selector sorts by chi2 afterwards.

--- END COMMIT MESSAGE ---

Examples only, no API change.

Profiled on ODD ttbar pu200: `MeasurementSelector::select` drops from 6.6% to 2.1% of the track finding time. Found tracks, selected tracks and stopped branches are unchanged.


benjaminhuth (approved):
LGTM!

paulgessinger (approved):

